PR #104, opened 2026-05-21
fix(lint): withdraw ESLint config support, repair the formatter, and move plugin options to config files

samchon:
## Intent

Three coupled fixes, surfaced while configuring `@ttsc/lint` on a real project:

1. **The formatter corrupted code.** `format/print-width` mangled any
   call/`new` expression whose argument was a multi-line callback or nested
   object — emitting inconsistently-indented output even on short lines.
2. **ESLint config support was half-wired and fragile.** The embedded config
   loaders silently dropped the `format` block (crashing format-only configs),
   and ESLint flat-config support carried a large, lightly-used surface.
3. **Plugin options lived in `tsconfig.json`.** `@ttsc/lint`, `@ttsc/banner`,
   and `@ttsc/strip` read inline options from their `compilerOptions.plugins`
   entry, splitting configuration across two places.

## Changes

- **`fix(ttsc)`** — `ttsx --no-plugins` builds a project with plugin discovery
  and loading disabled. ttsc's config loaders use it when evaluating a
  `*.config.ts` through ttsx, so that build no longer pulls in the host
  project's transform/check plugins — which would otherwise fail against the
  loader's deliberately lenient tsconfig (e.g. `@nestia/core` demanding
  `strict`) and abort config evaluation.
- **`fix(lint)`** — ESLint config support withdrawn entirely (`eslint.config.*`
  discovery, the ESLint runtime, flat-config recognition). A lint config file
  is a single `ITtscLintConfig` object in `lint.config.*` / `ttsc-lint.config.*`;
  the tsconfig plugin entry's only key is `configFile`. The `format/print-width`
  print engine is repaired: arrow/function/parenthesized/block printers,
  last-argument hugging, and a hard abstain when a subtree contains a multi-line
  node it cannot fully reprint — `ttsc format` either reflows correctly or
  leaves bytes untouched.
- **`refactor(banner)` / `refactor(strip)`** — configuration moves to
  `banner.config.*` / `strip.config.*` files, auto-discovered by walking up
  from the tsconfig directory. The tsconfig entry's only plugin-specific key is
  `configFile`; any other key is rejected with a migration error.
- **`docs`** — website guides and package READMEs updated for ESLint removal
  and the `configFile` model.

## Test plan

- Integrated build: `node scripts/build-current.cjs` → green.
- `@ttsc/lint` Go tests (`scripts/test-go-lint.cjs`) → `ok`; `@ttsc/banner` and
  `@ttsc/strip` Go build + tests → green; `@ttsc/lint` / `ttsc` TS typecheck →
  clean.
- New coverage: `ttsx --no-plugins`; formatter callback hugging / abstain /
  idempotence / nested objects; lint `configFile` override, `format`-block
  round-trip, and inline-key rejection; banner/strip config-file loading and
  inline-key rejection.
- Full `pnpm test` is exercised by CI below.

## Notes

- `@ttsc/banner`'s `TestNodeEnvironmentHelpers` Go test fails on a dev box that
  has a stray `/tmp/node_modules`; the failure is environment-sensitive and
  unrelated to this change (CI has no such directory).


samchon:
CI run 1 failed at `pnpm install --frozen-lockfile` across every job: the ESLint withdrawal removed `eslint` / `typescript-eslint` from `tests/test-lint/package.json` but left `pnpm-lock.yaml` stale. Fixed in d9d3084 by regenerating the lockfile; verified `pnpm install --frozen-lockfile` passes locally.

Incidental: regenerating the lockfile re-resolved `tests/test-rspack`'s `typia: next` specifier from `13.0.0-dev.20260520.2` to `13.0.0-dev.20260521.1` (the current `next` build — matches the version downstream projects already use). Calling it out per AGENTS.md §2.4.

samchon:
CI `experimental` job failed: the new strip config loader `packages/strip/driver/config.go` was not in `packages/strip/package.json`'s `files` allowlist, so the published `@ttsc/strip` omitted it and building the plugin from an installed package failed with `undefined: loadStripConfigMap`. Fixed by adding `driver/config.go` to `files`. Verified `driver/` and `plugin/` contents now fully match the allowlist for strip; banner adds no new Go file and lint ships `linthost` as a whole directory, so neither is affected.

samchon:
### `format/print-width` Prettier-fidelity fixes

A tgz experiment on `shopping-backend` surfaced four ways `ttsc format` diverged from Prettier. All four are now fixed (`beb9b0e`, `535e443`):

1. **`forceBreakFirstGroup` recursed into block bodies** — a hugged block-bodied callback's interior groups were force-broken, splitting unrelated short calls like `dotenvExpand.expand(env)` across three lines. Now guarded by `flatten`: only an item with no hard breaks of its own is force-broken.
2. **Template-literal `${...}` reflow** — calls inside substitutions (`encodeURIComponent(...)`) were broken across lines. Prettier renders interpolations at `printWidth:Infinity`; the rule now abstains on any node with a `TemplateExpression` ancestor.
3. **Object collapse** — deliberately multi-line objects were collapsed onto one line. Prettier's `objectWrap:"preserve"` default keeps an object expanded when the source wrote a newline after `{`; the printer now mirrors that.
4. **`trailingWidth` over-billing** — the un-movable trailing suffix (`;`, ` satisfies T`) was subtracted from the whole `printWidth` budget, over-exploding broken reflows. The rule now renders at the full budget and only re-renders shrunk when the reflow collapses to a single line.

**Verification** — `shopping-backend@experiment`, `ttsc format` vs `prettier@3` (`printWidth: 80`, same `importOrder`):

- Differing files: **145 → 12**.
- All non-interface code: **byte-identical to Prettier**.
- Idempotent (second `ttsc format` pass: 0 changes).
- The 12 remaining are all interface heritage clauses (`format/print-width` does not visit interface declarations). Measured: on 4 of them ttsc is strictly better — every line ≤ 80 cols where Prettier overflows to 90/94/96 — and the other 8 are heritage-wrap style differences with identical max line width. ttsc is never worse than Prettier.

Tests: 4 new `format/print-width` cases; the obsolete `collapses_multiline_short_object_to_flat` test (which asserted non-Prettier behavior) was replaced with `preserves_multiline_object`. Go lint tests and `test-lint` e2e suite green.

Issue #103, opened 2026-05-21
perf: ttsc and @ttsc/lint hard-code SingleThreaded, disabling TypeScript-Go parallel parse/check


samchon:
> **For the implementer.** This issue is a complete, self-contained work plan —
> the investigation is already done. Facts tagged **[verified]** were checked
> against the pinned TypeScript-Go source (`v0.0.0-20260429010842`) and the
> current ttsc tree. Jump to "Implementation plan" to start. Do this on a
> dedicated branch (`perf/parallel-typecheck` or similar).

## Problem

`ttsc` and `@ttsc/lint` construct every TypeScript-Go `Program` with
`SingleThreaded: TSTrue`, which disables **all** of TypeScript-Go's concurrency
— parallel parsing, the 4-way checker pool, parallel bind/semantic diagnostics,
and parallel emit. Parallel type-checking is the headline reason TypeScript-Go
exists; ttsc currently forfeits it on every `ttsc` / `ttsx` / `@ttsc/lint`
command. The flag was inherited verbatim from the first commit (`7a6d1a2`,
"migrate `ttsc` and `ttsx` from typia laboratory") and has never been revisited.

## Current state — two hard-coded sites [verified]

- **`packages/ttsc/driver/program.go:244`** — `CreateProgramFromConfig` →
  `ProgramOptions{ SingleThreaded: core.TSTrue, ... }`. The single path behind
  `ttsc build` / `check` / `transform`, `api-compile`, `api-transform`, and the
  `ttsx` runtime compile (all go through `driver.LoadProgram`).
- **`packages/lint/linthost/host.go:94`** — `loadProgram` →
  `ProgramOptions{ SingleThreaded: shimcore.TSTrue, ... }`. Behind every
  `@ttsc/lint` subcommand (`check` / `build` / `transform` / `fix` / `format`).

`grep -rn SingleThreaded --include=*.go packages/` returns exactly these two
non-test hits. Removing them lets `Program.SingleThreaded()` fall back to
`TSUnknown` → TypeScript-Go's multi-threaded default.

## How the flag behaves [verified]

`SingleThreaded` is **atomic** — there is no per-phase granularity. Dropping it
turns on, all at once:

- parallel source-file **parsing** (`processAllProgramFiles`);
- checker pool count **1 → 4** (`internal/compiler/checkerpool.go`; default 4,
  capped at `min(4, fileCount, 256)`);
- parallel **bind + semantic diagnostics** (`forEachCheckerGroupDo`);
[...]

PR #108, opened 2026-05-21
perf(ttsc): drop SingleThreaded so TypeScript-Go parses/checks/emits in parallel

samchon:
## Intent

Closes the first stage of #103. `ttsc` and `@ttsc/lint` both constructed every
TypeScript-Go `Program` with `SingleThreaded: TSTrue`, disabling **all** of
TypeScript-Go's concurrency — parallel parsing, the pooled checker, parallel
bind/semantic diagnostics, and parallel emit. The flag was inherited verbatim
from the first commit and never revisited. This PR drops it at both
construction sites so the program runs on TypeScript-Go's multi-threaded
default.

## Scope

This is **Commit 1 of the issue's plan only** — TypeScript-Go program
parallelism. The remaining stages are intentionally **not** in this PR and will
follow once this one is green:

- **Commit 2** — `@ttsc/lint` engine parallelism (parallel file walk across the
  checker pool). Not included.
- **Commit 3** — plugin traverse parallelism (typia/nestia). Crosses the public
  plugin API boundary; tracked as a separate PR per the issue.

## Changes

- `driver/program.go`, `linthost/host.go` — remove `SingleThreaded` from the
  two `ProgramOptions` so the program falls back to the multi-threaded default.
- `driver/rewrite.go` — `emit()` now serializes its whole `WriteFile` callback
  under a mutex. Parallel emit invokes that callback once per emitted file
  concurrently; the `cursors` map would otherwise trip `fatal error: concurrent
  map writes`, and a caller's wrapped `writeFile` (e.g. api-compile's output
  map) must also see one writer at a time. The patch work is cheap, so
  serializing only the callback costs ~nothing while parse/check/emit-text
  generation parallelize. `EmitAll`/`EmitAllRaw` doc comments now state the
  callback's concurrency contract.

## Why this is safe

The phases ttsc layers on top of TypeScript-Go — plugin application, the output
rewriter, and the (still serial) lint engine walk — keep using a single pooled
checker, which stays valid against the larger pool: TypeScript-Go fans out only
*inside* each phase, never across these serial phases. The only ttsc-side code
that now runs during a parallel phase is the emit `WriteFile` callback, which
this PR makes thread-safe. `EmittedFiles` ordering is unaffected —
`CombineEmitResults` reassembles per-emitter results in source-file order — so
the build manifest stays deterministic.

`ttscserver` is not affected: it delegates type-checking to a spawned
`tsgo --lsp --stdio` subprocess and never goes through `driver.LoadProgram`.

## Test plan

- New `packages/ttsc/test/driver/emit_rewrites_every_source_under_parallel_emit_test.go`
  emits a four-source project so multiple emitter goroutines exercise the
  `WriteFile` callback concurrently, and asserts every output gets its own
  rewrite. Verified that, with the mutex removed, `go test -race` reliably
  reports `DATA RACE`; with it in place the test passes under `-race`.
- `go build ./...` + `go vet ./...` clean for `packages/ttsc`.
- Full `go test ./...` green for `packages/ttsc`.
- CI (`test.yml` + `typia.yml`) gates the rest.

### DIFF
--- a/packages/lint/linthost/compile.go
+++ b/packages/lint/linthost/compile.go
@@ -13,6 +13,7 @@ package linthost
 
 import (
   "context"
+  "encoding/json"
   "errors"
   "flag"
   "fmt"
@@ -61,6 +62,7 @@ func RunTransform(args []string) int {
   pluginsJSON := fs.String("plugins-json", "", "ttsc plugin manifest JSON")
   singleThreaded := fs.Bool("singleThreaded", false, "run TypeScript-Go single-threaded")
   checkers := fs.Int("checkers", 0, "type-checker pool size (0 = TypeScript-Go default)")
+  tsgoArgsRaw := fs.String("tsgo-args", "", "JSON array of forwarded tsgo CLI flags")
   if err := fs.Parse(filterKnownFlags(args, map[string]bool{
     "cwd":            true,
     "file":           true,
@@ -69,11 +71,17 @@ func RunTransform(args []string) int {
     "tsconfig":       true,
     "singleThreaded": false,
     "checkers":       true,
+    "tsgo-args":      true,
   })); err != nil {
     return 2
   }
   if *file == "" {
     fmt.Fprintln(os.Stderr, "@ttsc/lint transform: --file is required")
+    return 2
+  }
+  tsgoArgs, err := decodeTsgoArgs(*tsgoArgsRaw)
+  if err != nil {
+    fmt.Fprintln(os.Stderr, err)
     return 2
   }
   resolvedCwd, err := resolveCwd(*cwd)
@@ -85,6 +93,7 @@ func RunTransform(args []string) int {
     forceEmit:      true,
     singleThreaded: *singleThreaded,
     checkers:       *checkers,
+    tsgoArgs:       tsgoArgs,
   })
   if err != nil {
     fmt.Fprintf(os.Stderr, "@ttsc/lint: %v\n", err)
@@ -174,6 +183,7 @@ type subcommandOpts struct {
   outDir         string
   singleThreaded bool
   checkers       int
+  tsgoArgs       []string
 }
 
 // parseSubcommandFlags parses the shared flag set used by the `check`,
@@ -192,6 +202,7 @@ func parseSubcommandFlags(name string, args []string) (*subcommandOpts, error) {
   outDir := fs.String("outDir", "", "")
   singleThreaded := fs.Bool("singleThreaded", false, "")
   checkers := fs.Int("checkers", 0, "")
+  tsgoArgsRaw := fs.String("tsgo-args", "", "")
   if err := fs.Parse(filterKnownFlags(args, map[string]bool{
     "cwd":            true,
     "emit":           false,
@@ -203,11 +214,16 @@ func parseSubcommandFlags(name string, args []string) (*subcommandOpts, error) {
     "verbose":        false,
     "singleThreaded": false,
     "checkers":       true,
+    "tsgo-args":      true,
   })); err != nil {
     return nil, err
   }
   if *emit && *noEmit {
     return nil, errors.New("@ttsc/lint: --emit and --noEmit are mutually exclusive")
+  }
+  tsgoArgs, err := decodeTsgoArgs(*tsgoArgsRaw)
+  if err != nil {
+    return nil, err
   }
   resolvedCwd, err := resolveCwd(*cwd)
   if err != nil {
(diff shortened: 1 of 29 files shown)


samchon:
## Added: `--singleThreaded` / `--checkers` CLI flags (commit `a680f4f`)

Follow-up commit so ttsc's CLI mirrors tsgo's for the two threading knobs — the escape hatch noted as "Optional" in #103. Previously `ttsc --singleThreaded` failed with `unknown option`.

**Plumbing** — both flags now travel every execution path:
- `ttsc` / `ttsx` launchers parse `--singleThreaded` and `--checkers <n>` (+ `--flag=value`); `--checkers` is validated as a positive integer (tsgo's `minValue: 1`).
- No-plugin build lane forwards both verbatim to the spawned `tsgo` binary (type-check + emit passes).
- In-process Program paths (`@ttsc/lint` sidecar, transform-plugin host, `cmd/ttsc`) apply them onto `CompilerOptions` via `driver.LoadProgram` / `linthost.loadProgram` — exactly where tsgo's own CLI lands them.

**Tests**: three `ttsc` feature cases (`--singleThreaded` builds, `--checkers` builds, `--checkers 0` rejected). Go + TS build green; `cmd`/`driver`/`utility`/`cli` Go suites and the lint Go suite pass.

This grows the PR scope beyond the original "drop SingleThreaded" commit — the two are coupled (the flag is the manual override for the new parallel default), so they ride together.

samchon:
## Added: forward unrecognized flags to tsgo (commit `b3d67a3`)

`ttsc --strict file.ts` previously threw `ttsc: unknown option --strict` — the launcher's build-arg parser was a closed whitelist. Since the principle is that ttsc's CLI mirrors tsgo's, the parser now **classifies** tokens instead of whitelisting:

- ttsc-owned flag (`--emit`, `-p`, `--singleThreaded`, …) → parsed as before;
- bare token with a `.ts`/`.tsx`/`.mts`/`.cts` extension → single-file input;
- anything else (unrecognized `--flag`, or a bare value token like `es2020` in `--target es2020`) → collected and forwarded verbatim to the spawned `tsgo`.

ttsc never needs tsgo's arity table: the value token has no TS extension, so it rides with the passthrough run and tsgo does the real parsing. Verified: `ttsc --strict file.ts` enforces strict-null over a `strict:false` tsconfig; `ttsc --target es2020` and `ttsc --listFiles` work; project + single-file lanes both covered.

**Not covered**: transform-compiler plugin projects (typia/nestia) build through a native host, not tsgo, so forwarded compiler-option flags don't reach that lane — those projects carry options in tsconfig.json. Flagged as a known boundary.

Tests: 2 new `ttsc` feature cases (unknown-flag forwarded + enforced; spaced flag value kept with its flag).

samchon:
## Added: forwarded tsgo flags reach the plugin build lane too (commit `08ec2f1`)

The prior commit forwarded unrecognized flags to `tsgo` on the plain + single-file lanes, but a project that configures a ttsc plugin builds through native sidecars (`@ttsc/lint`, the transform-plugin host) instead of shelling out to `tsgo` — so `ttsc --strict` was silently dropped there. This closes that gap.

- The launcher serializes the passthrough tokens into one `--tsgo-args` flag (JSON array — survives the sidecars' unknown-flag filters) on every native `build`/`check` invocation.
- Each sidecar (`@ttsc/lint`, the transform-plugin utility host, `cmd/ttsc`) decodes it into `LoadProgramOptions.TsgoArgs` / `loadProgramOptions.tsgoArgs`.
- `driver.LoadProgram` / `linthost.loadProgram` replay those flags through TypeScript-Go's own `tsoptions.ParseCommandLine` and hand the result to `GetParsedCommandLineOfConfigFile` as the existing-options overlay — **the exact call tsgo's own CLI makes**. `mergeCompilerOptions` copies the overlay's non-zero fields over the tsconfig (CLI wins). No hand-rolled option table, no per-field merge.

So `ttsc --strict` now behaves identically whether or not the project has plugins.

**Tests**: driver unit test (overlay overrides `strict:false`); `@ttsc/lint` command test (sidecar decodes `--tsgo-args`); plugin-corpus e2e (`ttsc --noEmit --strict` on a real `@ttsc/lint` project surfaces the strict-null error — verified, 25s run).

This completes the CLI-parity work; the earlier "not covered: plugin projects" caveat is resolved.

samchon:
Superseded — split into two PRs so the perf change and the CLI feature land independently (per the review finding that they should not share one PR):

- **#112** `perf(ttsc): parallelize parse, type-check, and emit` — issue #103 Commit 1. **Merged.**
- **#113** `feat(cli): forward tsgo flags through ttsc and ttsx` — the CLI parity work, stacked on #112.

Closing this combined branch. All commits are preserved in #112 + #113.

PR #186, opened 2026-05-29
fix(lint): ttsc format Prettier 3.8.3 parity (semicolons, quotes, imports, layout)

samchon:
## Intent

Close the divergences between `ttsc format` and **Prettier 3.8.3** found by differentially diffing both formatters over the seven benchmark fixtures (vue, rxjs, typeorm, zod, nestjs, vscode, shopping-backend) with scope-matched configs. Target: latest Prettier semantics, for the 0.14.0 release.

Findings were grouped A (clear version-independent bugs), B (edit-based coverage gaps), C (Prettier 2 vs 3 layout lag). Decision: `defaultFormat` claims "mirror Prettier", so the C lags are bugs to fix.

## Scope / progress

**A — landed:**
- [x] `format/semi` strips interface/type-literal member, class-field, and type-alias `;` under `semi:false`, with per-context ASI-hazard guards (class fields use the full expression set; type members only `(`/`<`; same-line separators kept).
- [x] `format/quotes` chooses the escape-minimizing quote and uses the configured preference only as a tie-break, in both directions (`"\""` -> `'"'`).
- [x] `format/trailing-comma` skips dynamic `import(...)` argument lists.

**B / C — in progress (commits will stack onto this branch):**
- [ ] B: token spacing rule (`format/spacing`), orphan `;` merge, single-statement clause join, flat-overflow print-width coverage, first/last-arg callback hugging.
- [ ] C: nested-ternary staircase + `??`-in-ternary parens, class/interface heritage break order + type-parameter explode (closes the over-width type-param-default break), constructor parameter-property forced break + signature param-vs-return break + assignment-RHS break.

## Test plan
- Go unit tests per behavior under `packages/lint/test/format/` and `.../printer/` (one assertion per file), including idempotency and must-not-apply cases.
- `node scripts/test-go-lint.cjs` green for landed work.
- Full gate (`pnpm test`) + a re-run of the differential harness over the fixtures before merge.


samchon:
## Review cycle round 1 — applied, validated

A 6-reviewer pass (logic / printer-correctness / clean-code / coverage / perf / differential-Prettier) ran over the branch. Each finding was re-verified against **Prettier 3.8.3** before acting — one reviewer claim (strip redundant `\'` escapes) was *rejected* because Prettier keeps `"a\'b"` verbatim, so "fixing" it would have introduced a divergence.

### Shipped this round (each Prettier-verified, tested)
- **perf**: print-width fast path now runs before the ancestor-walk + per-byte comment-scan guards; the hot path on well-formatted code is allocation/scan-free. (behavior-preserving)
- **quotes**: redundant `\'`/`\"` escapes are now counted, so `"a\'b"` no longer wrongly flips quote style under the opposite preference.
- **heritage**: two-tier layout — types stay inline after the broken keyword (`extends A, B, C`) and explode one-per-line only when that line still overflows; the opening brace drops to its own line only for a `class` with a non-empty body (interface / empty body keep `{}` glued).
- **hugging**: `useEffect(() => {…}, [deps])` now hugs the leading callback; last-argument hugging declines when a leading argument is a block callback (matching Prettier's willBreak gate).
- **comment-loss bug**: header reflow silently dropped a comment between the name and `extends`; it now abstains on any comment in the rewritten span.

### Re-test + re-compare (differential harness, Prettier 3.8.3)
All fixtures **converge** (`ttsc format` status 0, no oscillation). Divergence counts:

| fixture | before | after |
|---|---|---|
| typeorm | ~28 | **4** |
| vue | 52 | **15** |
| vscode | 112 | **29** |
| nestjs | 31 | 30 |

### Remaining residuals (categorized — not regressions)
- **Single heritage type with an overflowing type-argument list** (`implements Serializer<…long…>`): Prettier 3 keeps the clause inline and breaks the `<…>`; ttsc abstains and leaves the source's Prettier-2 shape. Dominant nestjs gap. *Follow-up feature.*
- **Ternary-arm continuation indent**: a call/object that is a `?:` arm and breaks gets one fewer indent level than Prettier. Dominant vue gap. *Follow-up.*
- **First-arg hug with a call trailing arg** (`(cb, getConfig())`): needs a last-line fit check in the doc engine (the conditional-group fit only measures the first line). *Follow-up.*
- **Nested-consequent ternary flat parens** (`p ? (q?r:s) : t`): width-dependent printer + IfBreak work. *Follow-up.*
- Pre-existing: eslint-disable-honored files (ttsc correctly skips), block-comment blank lines, print-width-off width residue.

samchon:
## Session update: format/indent fixes + the remaining Prettier options

All work below is committed, pushed, and CI-green (Go suite, typecheck, lint, platform matrix; the `experimental` job is red on master too — it pulls a published npm package, unrelated to this branch).

### format/indent real bugs (found by a ttsc-only self-check, no Prettier oracle)
The self-check is: take a canonical source, strip every line to column 0 (semantics unchanged), run `ttsc format`, and require it to restore the canonical exactly. It surfaced cases where the cascade reported success (exit 0) on a malformed result:
- **de-indent under a wrapped block head** — a curried arrow / multi-line-condition `if` body was de-indented from correct source. Fixed (structural chained-arrow cede, `4a80bdb2`).
- **class members, case labels, case-body braces never re-indented** — member declarations and `case`/`default` labels are not statements, so they were skipped. Fixed (`2f9990c7`).
- **switch / class / interface closing braces never re-indented** (`0d0554c2`).

### Remaining Prettier options implemented (each a new edit rule + ttsc-only self-check + regression tests + ITtscLintFormat/defaultFormat/docs)
- `format/arrow-parens` (Prettier `arrowParens`, default `"always"`)
- `format/bracket-spacing` (Prettier `bracketSpacing`, default `true`)
- `format/quote-props` (Prettier `quoteProps`, default `"as-needed"`)

These were the last unimplemented Prettier knobs; only the JSX-specific switches remain noted as out of scope. `endOfLine` was reviewed and is correct (default `"lf"`, matching Prettier).

### Note on the differential benchmark
After `format/arrow-parens` shipped, the reharness divergence rose on fixtures whose committed `lint.config.ts` does not set `arrowParens` while their Prettier spec runs `arrowParens: "avoid"` (vue, nestjs, vscode). That is a config-parity gap in the external fixture repos (ttsc applies its Prettier-aligned default `"always"`, exactly as Prettier-with-default-config would), not a ttsc defect — typeorm, which omits arrowParens on both sides, barely moved. Making the differential fair requires aligning each fixture's lint.config `format` block with its `prettierConfig`, which is an external-repo change.

PR #187, opened 2026-05-30
feat(lint): bring ttsc format to Prettier 3.8.3 parity

samchon:
## Summary

The protagonist of this PR is the **`@ttsc/lint` formatter** (`ttsc format`, a Prettier-compatible replacement). What began as a research-review round grew into a large correctness-and-capability pass that brings `ttsc format` substantially toward byte-parity with **Prettier 3.8.3**, validated against the benchmark fixtures (the 3.8.3 fixtures — typeorm, vue — reach parity; vscode is within a small, mapped residue).

Lint-rule autofix safety hardening is the secondary theme. The `ttsc` watch/perf touch-ups are incidental.

Prettier 3.8.3 source is used as the reference spec — predicates are ported faithfully and validated against the shipped bundle, and corrected toward the behavior that ought to be right where Prettier's own behavior is a quirk.

## Formatter (`ttsc format`) — primary

- **Call-argument layout** to Prettier's rules: last-argument and first-argument hugging (`couldExpandArg` / `shouldGroupFirstArg`), test-call hugging (`isTestCall`, exact callee set), the function-composition break (`isFunctionCompositionArgs`, decorator-exempt, call-not-new), return-typed-arrow decline, and trailing call / binary / cast classification.
- **Arrays**: concise numeric fill via a new `docFill` primitive, and the same-kind-multi-child-array force-break (`shouldBreak`).
- **Ternary**: fixed nested-indent staircase and consequent-position parentheses (`ternary-old.js`).
- **Declarations**: heritage and type-parameter breaking, class/interface header reflow, empty-body brace fit.
- **Indentation**: cede rules for printer-/source-owned regions — expression-nested bodies, braceless control bodies, type-literal parameters, continuation-line function expressions, chained arrows, decorated members.
- **Width**: display-column (East-Asian-aware) measurement instead of byte length, single-specifier import/export inlining, and whole-import-as-one-group reflow so the `from` tail is counted.
- **Data safety on reflow**: source blank lines preserved between list items; comments preserved (abstain on inter-item comments rather than drop them).
- **`sortImports`** format block: nested grouping, merging, and type-combining.

## Lint rules — secondary

Autofix safety and correctness hardening across `no-var`, `prefer-const`, `prefer-template`, `no-import-type-side-effects`, `no-wrapper-object-types`, `ban-ts-comment`, and `disable-line` targeting.

## ttsc — incidental

Watch-loop hardening and exit-code propagation; memoized go-compiler identity in plugin builds.


samchon:
## Review round 2 (fresh team)

Round 2 re-verified round 1's fixes and found a regression in the no-var fix plus new corruption-class bugs. The validated fixes landed in four logical commits:

- **fix(lint): close no-var and prefer-const autofix corruption holes** — no-var rewrote a var to let even when a for-header redeclaration (`var x; for (var x ...)`) or an object-literal shorthand value read before the declaration (`({ x }); var x = 1`) made it unsafe, producing a duplicate-declaration SyntaxError or a TDZ ReferenceError; the use-before-declaration scan also over-declined on member names, object keys, labels, and type references. prefer-const rewrote a let to const even when a for-of/for-in target reassigned it. Both gates now classify identifier roles and account for for-header and loop-target reassignment, with unrecognized roles still declining safely.
- **fix(lint): anchor no-import-type-side-effects keyword scan past leading trivia** — the import-keyword scan started at the node position (which includes leading trivia), so a leading comment containing the word import captured the match and the fix silently turned type-only imports into value imports. The scan now anchors past trivia via keywordStart.
- **fix(ttsc): harden watch loop against rebuild throws and watcher leaks** — a throw from a debounced rebuild crashed the watcher, and a first-build throw left persistent watchers open so the process hung. Rebuilds are now wrapped so failures are reported and the watcher keeps running, watchers are torn down on an escaping throw, an fs.watch error handler keeps the session alive, and the synchronous return is routed through toExitCode.
- **docs(lint,ttsc): correct ban-ts-comment, faq, indent, and paths docs** — dropped the false ban-ts-comment claim that @ts-expect-error is allowed with a description, replaced the FAQ hardcoded speedup multiplier with a pointer to the benchmark guide, updated the format/indent doc for the closing-brace and header reindent pass, and noted in the paths README that only outDir is required (rootDir is derived when omitted).

An independent verifier adversarially re-checked the lint fixes; one further hole (no-var object-literal shorthand value read before declaration) was found and closed before push.

Deferred follow-ups:
- Format cascade coverage gaps: single-line-block trailing brace, chained-arrow indent, long-union breaking.
- Watch mid-session subdirectory coverage.
- Playground semver-range resolution and Bun adapter polish.

**Test plan:** full Go lint suite green; ttsc watch e2e green (new test_ttsc_watch_survives_throwing_rebuild plus the existing test, no regression).

samchon:
## Review round 3 (fresh team)

Round 3 verified the rounds 1 and 2 fixes and found a third no-var corruption case plus a new formatter non-convergence bug. An independent verifier then found a fifth no-var case (hoisted function/class redeclaration), which is also fixed here.

### Commits

- `fix(lint): decline no-var autofix on unsafe redeclarations` — the no-var safety gate still rewrote a `var` to `let` when a mixed declaration list hid a destructured sibling (`var a = 1, { b } = o;`) or when a hoisted function/class declaration shared the name (`var x = 1; function x() {}`), producing duplicate-declaration SyntaxErrors or TDZ ReferenceErrors. The gate now requires a single VariableDeclaration node and counts same-name function and class declarations toward the redeclaration check.
- `fix(lint): abstain print-width inside JSX expressions and binary-search its comment scan` — a reflow node inside both a JSX attribute and a JSX child made print-width break both then revert next pass, so `ttsc format` oscillated to the pass cap and exited 2 on valid tsx. print-width now abstains inside JSX expressions, mirroring the template-substitution guard. Separately, the per-node comment scan now binary-searches the source-ordered child ranges instead of a per-byte linear lookup, with byte-identical formatting output.
- `perf(ttsc): memoize go compiler identity in plugin builds` — `computeCacheKey` re-spawned `go version` and re-hashed the go binary per source plugin on every invocation; the compiler identity is now memoized per process.
- `docs(lint): note ttsc format runs all format rules regardless of per-rule off` — documented that a per-rule `off` in the rules map only affects `ttsc check`, since `ttsc format` force-activates every format rule.

### Output preservation

The perf change is output-preserving: cache-key bytes are unchanged, and the go env result is intentionally not memoized because it depends on mutable environment and GOROOT content.

### Deferred follow-ups (carried forward)

- unplugin per-module rehash (mtime pre-check, needs benchmark)
- Program reload per pass
- shared per-node type memo
- flatten allocation
- MemFS rmdir/rename descendant handling
- codeActions serial sidecar latency
- playground semver-range resolution
- Bun adapter polish

### Test plan

- Full Go lint suite green.
- Existing ttsc plugin-build and cache-key suites green.
- New behavior-locking and parity tests added and confirmed fail-before / pass-after.
- An independent verifier confirmed no remaining corruption in the no-var / prefer-const / import-type fixes and byte-identical formatter output for the binary-search change.

samchon:
## no-var safety gate consolidated

Across the rounds the no-var autofix surfaced six corruption holes (use-before-declaration, plain redeclaration, for-header redeclaration, object-literal shorthand, mixed destructuring list, hoisted function/class, and an initializer self-reference). Rather than continue patching edges, the gate was redesigned into a single file-wide binding-count plus TDZ check.

The trade is conservative over-declining of harmless cross-scope same-name bindings, never corrupting; safe common cases still convert.

An independent verifier drove the redesign and confirmed all redeclaration kinds decline, safe cases still fix, and found the initializer self-reference class which is included here.

### Test plan

- Full Go lint suite green.
- Nine behavior-locking tests added and confirmed fail-before/pass-after.

samchon:
## Review round 4 (fresh team)

Round 4 ran a systemic autofix-safety audit (prompted by the no-var saga) over the complete non-format autofix surface. It found two more autofix corruption bugs (prefer-template carriage-return normalization; no-wrapper-object-types rewriting a shadowed wrapper name), plus a decorator-indent parity gap and a release-workflow lockfile gap.

### Commits

- `fix(lint): preserve carriage returns and tabs in prefer-template autofix` — prefer-template emitted the cooked string value raw into the template body, so a carriage return was normalized to a line feed by the template-literal grammar, silently changing the value; the escape routine now emits CR, LF, and tab as backslash escapes.
- `fix(lint): skip no-wrapper-object-types rewrite when the wrapper name is shadowed` — the rule rewrote a wrapper annotation (e.g. String to string) even when the name referred to a local binding; the shadow check now bails on any file-scope binding (imports including import-equals, variables, functions, classes, interfaces, type aliases, enums, namespaces) and still rewrites a genuine global.
- `fix(lint): indent decorated class members and declarations to their nesting depth` — format/indent re-indented only the decorator line, leaving the declaration line half-indented; the indent pass now also re-indents the declaration line of decorated class members and decorated declaration statements (class, function, interface, including inside a namespace).
- `ci(release): use frozen lockfile on the publish path` — the release workflow ran pnpm install without --frozen-lockfile, the only workflow that could resolve a drifted lockfile and the path that publishes to npm; it now matches the other workflows.
- `docs(lint): clarify print-width abstains inside JSX expression containers` — the print-width limitations list now notes that reflowable expressions inside a JSX expression container are not reflowed, matching the abstain guard.

### Verification

An independent verifier drove the real fix and format paths and found the no-wrapper-object-types and decorator-indent fixes were initially incomplete (missed function and import-equals shadows; missed decorated declaration statements). Both were completed and re-verified here.

### Deferred follow-ups

- no-var per-file walk cost
- Go SDK download caching and go-vet/prettier CI coverage
- format cascade coverage gaps (single-line-block trailing brace, chained-arrow indent, long-union breaking)
- unplugin per-module rehash
- Program reload
- constructor parameter-property indentation

### Test plan

Full Go lint suite green. Behavior-locking and idempotency/convergence tests were added and confirmed fail-before/pass-after. An independent verifier confirmed the formatter stays convergent and idempotent.

PR #194, opened 2026-06-01
feat(ttsx): run raw-TypeScript dependencies across the graph via Node hooks

### LINKED ISSUE #193: ttsx: extensionless relative imports in raw-TS dependencies fail under ESM
## Problem

`ttsx` can't run a dependency whose `exports` resolves to raw `.ts` once any part of the graph is ESM:

- **workspace `.ts`** (realpath outside `node_modules`, e.g. a pnpm symlink): Node strips the types but rejects the source's **extensionless** relative imports → `ERR_MODULE_NOT_FOUND`.
- **published `.ts`** (a real `node_modules` tree): Node refuses to type-strip under `node_modules` → `ERR_UNSUPPORTED_NODE_MODULES_TYPE_STRIPPING`.

`ttsx` only rewrites its own emitted entry (`runTtsx.ts` → `rewriteEsmSpecifiers`, `emitDir`-only); dependencies are loaded raw by Node through the virtual overlay in `prepareExecution.ts`, and `tsconfig` can't reach runtime loading. `tsx` already absorbs both (resolve + load hooks over the whole graph); `ts-node` absorbs the workspace case but ignores `node_modules`.

This blocks `typia` from dropping its last `ts-node` usage — a `transpile-only` loader kept only because it forced the whole graph to CommonJS, which masked the workspace wall.

## Request

Register Node loader hooks in the child that `runPreparedEntry` spawns (`module.register` / `--import`), the way `tsx` does:

1. **`resolve` hook** — when a relative specifier fails to resolve, probe `.ts/.tsx/.mts/.cts/.js/.mjs/.cjs` then `index.*` (reuse `withResolvableExtension`'s order), across the whole graph. *Closes the workspace wall — ship this first; it unblocks `typia`.*
2. **`load` hook** — **transpile-only** (type-strip, no type-checking, no per-dependency `tsconfig`) for `.ts` under `node_modules`, bypassing the strip refusal. Use tsgo's transpile / `api-transform` path, **not** `runSingleFileEmit` (a full *checked* build). Cache by path + mtime. *Closes the published wall.*

**Do not weaken type-checking.** It stays at the compile gate: `prepareExecution`'s tsgo build deep-checks the whole program including imported workspace neighbors (verified — a type error in an imported workspace package's `src` fails the run before anything executes). The hooks are runtime-only; scope the `load` hook to `node_modules` so workspace neighbors keep the native-strip path and stay compile-checked. (`ttsx` thus stays strictly stronger than `tsx`: full compile gate + `tsx`-style runtime reach.)

## Acceptance

Under `ttsx`, with an ESM-only (`"type": "module"`) package in the graph and **no** change to any dependency's `exports` / source / `tsconfig`:

- a **workspace** dep exporting raw `.ts` with extensionless imports runs green (resolve hook);
- a **published** dep (real `node_modules`) shipping raw `.ts` runs green (load hook);
- a genuinely missing module still throws `ERR_MODULE_NOT_FOUND`;
- a type error inside an imported **workspace** package still fails at the compile gate.

samchon:
## Intent

Closes #193. `ttsx` only rewrote its own emitted entry, so dependencies that ship raw `.ts` failed once any part of an ESM graph loaded them:

- a **workspace** dependency (a `node_modules` symlink whose realpath is outside `node_modules`) had its types stripped by Node, but its own extensionless relative imports were rejected (`ERR_MODULE_NOT_FOUND`);
- a **published** dependency shipping `.ts` under `node_modules` could not load at all, because Node refuses to strip types there (`ERR_UNSUPPORTED_NODE_MODULES_TYPE_STRIPPING`).

This blocked `typia` from dropping its last `ts-node` usage.

## Scope

Register Node module-customization hooks in the child `ttsx` spawns (`--import` a registrar that `module.register`s the hooks), the way `tsx` does:

- **`resolve` hook** — after Node's resolver throws on a relative specifier, probe `.ts/.tsx/.mts/.cts/.js/.mjs/.cjs` then `index.*`, across the whole graph. Closes the workspace wall.
- **`load` hook** — type-strip raw `.ts` under `node_modules` only, so workspace neighbours keep Node's native path. Module format follows the extension (`.mts`/`.cts`), then the nearest `package.json` `type`, with ESM-syntax detection overriding a CommonJS baseline exactly as Node's own module-syntax detection does.

The hooks are runtime-only and never weaken the compile gate: the up-front tsgo build still deep-checks the whole program including imported workspace neighbours. So `ttsx` stays strictly stronger than `tsx` (full compile gate + `tsx`-style runtime reach).

## Acceptance (all verified by new e2e tests)

- ✅ a workspace dep exporting raw `.ts` with extensionless imports runs green (resolve hook);
- ✅ a published dep shipping raw `.ts` runs green (load hook);
- ✅ a genuinely missing module still throws `ERR_MODULE_NOT_FOUND` (relative-extensionless, relative-with-extension, and bare-package variants);
- ✅ a type error inside an imported workspace package still fails at the compile gate before anything runs.

## Deviations from the issue's suggested mechanism

Two implementation choices differ from the request text; both keep the issue's stated *intent* ("transpile-only, no type-checking, no per-dependency tsconfig") and satisfy every acceptance criterion:

1. **Transpiler.** The issue suggested tsgo's transpile / `api-transform` path. Every tsgo entry point in this repo (`api-compile`, `api-transform`, `runSingleFileEmit`) builds a *whole, checked* program against a `tsconfig`; there is no single-file, no-check transpile primitive, and the child is a plain Node process. So the `load` hook uses Node's own `module.stripTypeScriptTypes` in `transform` mode — the same type-strip Node applies to first-party `.ts`, just reached past the `node_modules` policy refusal. This is transpile-only, needs no tsconfig, and matches what `tsx`/Node do.
2. **Cache.** The issue asked to cache by path + mtime. Node already caches every module by resolved URL, and the compile gate rejects the query-string aliasing that would re-trigger a load, so a per-file transpile cache can never hit in a single run. Dropped to keep the load path clean and fully covered. (The `package.json` `type` lookup *is* cached, since multiple files in one package legitimately hit it.)

The `ExperimentalWarning` that `stripTypeScriptTypes` would print is silenced with `--disable-warning=ExperimentalWarning` on the child, the same flag this repo's own TypeScript runner uses.

## Test plan

- New e2e suite under `tests/test-ttsc/src/features/ttsx-runtime/` (10 cases) — full `ttsx` feature suite green.
- `tsgo --noEmit` on `tests/test-ttsc` clean; `packages/ttsc` build (type-check + emit) clean.

### DIFF
--- a/.github/workflows/nestia.yml
+++ b/.github/workflows/nestia.yml
@@ -17,12 +17,13 @@ concurrency:
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
-  Ubuntu:
+  # Pack the current-platform ttsc tarballs once and share them with every
+  # matrix job below, instead of rebuilding them six times.
+  tarballs:
     runs-on: ubuntu-latest
-    timeout-minutes: 60
     env:
       # Only build + pack the current platform (ttsc-linux-x64) — nestia's
-      # overrides below pull in ttsc.tgz + ttsc-linux-x64.tgz only. The full
+      # overrides pull in ttsc.tgz + ttsc-linux-x64.tgz only. The full
       # cross-platform matrix takes ~20 min for tarballs we never consume.
       TTSC_TARBALLS_CURRENT: "1"
     steps:
@@ -50,6 +51,58 @@ jobs:
 
       - name: Build ttsc tarballs (current platform only)
         run: pnpm package:tgz
+
+      - name: Upload tarballs
+        uses: actions/upload-artifact@v4
+        with:
+          name: ttsc-tarballs
+          path: experimental/tarballs/*.tgz
+          if-no-files-found: error
+          retention-days: 1
+
+  # Mirror nestia (next)'s own test.yml matrix so each suite runs in parallel,
+  # but against the local ttsc tarballs instead of the published release.
+  nestia:
+    needs: tarballs
+    name: ${{ matrix.name }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - { name: go, run: "pnpm test:go" }
+          - { name: sdk, run: "pnpm --filter ./tests/test-sdk start" }
+          - { name: migrate, run: "pnpm --filter ./tests/test-migrate start" }
+          - {
+              name: transform-options,
+              run: "pnpm --filter ./tests/test-transform-options start",
+            }
+          - { name: e2e, run: "pnpm --filter ./tests/test-e2e start" }
+          - { name: benchmark, run: "pnpm --filter ./tests/test-benchmark start" }
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: pnpm/action-setup@v4
+        with:
+          version: 10.6.4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 24.x
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: packages/ttsc/go.mod
+          cache-dependency-path: |
+            packages/ttsc/go.sum
+            tests/go-transformer/go.mod
+
+      - name: Download ttsc tarballs
+        uses: actions/download-artifact@v4
+        with:
+          name: ttsc-tarballs
+          path: experimental/tarballs
 
       - name: Clone nestia next
         run: |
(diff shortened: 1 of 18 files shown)


samchon:
CI: the `typia` check was red because typia's `pnpm-workspace.yaml` now declares its own `overrides:` block (and routes `ttsc` through a catalog). The workflow's `cat >> pnpm-workspace.yaml` appended a *second* `overrides:` key, so pnpm aborted with `duplicated mapping key (39:1)` — a pre-existing infra breakage unrelated to this PR's runtime hooks, surfaced here because the feature touches `packages/**`.

Fixed in `ci(typia)`: the step now merges the local-tarball overrides into typia's existing `overrides:` block (creating it only if typia drops it). Verified against typia's live `pnpm-workspace.yaml` with a real YAML parser, both the merge and create-when-absent paths.

Note: `nestia.yml` carries the same `cat >>` pattern but is not currently broken (nestia's workspace has no `overrides:` block yet). Left untouched per scope; happy to harden it the same way if you want.

samchon:
CI: the nestia `sdk` matrix job failed (`swagger`, `swagger-한글경로`, `body-config-assert`) with `SyntaxError: Named export 'INestiaConfig' not found`.

Root cause (a real regression from the runtime hooks, now fixed): ttsx mirrors the project into `<root>/node_modules/.cache/ttsc/ttsx/.../fs/...`, so the user's own `nestia.config.ts` appears under a `node_modules` path at runtime. The `load` hook's `isUnderNodeModules` matched that cache prefix and type-stripped the project file with `stripTypeScriptTypes` — which (unlike the compile gate) cannot elide type-only imports — so `import { INestiaConfig } from "@nestia/sdk"` survived and broke the named-export binding, stealing the file from nestia's own `config-ts-loader`.

Fix (`fix(ttsx): do not transpile project files under the node_modules/.cache overlay`): treat `node_modules/.cache` as not-a-package. Genuine deps still have a deeper `node_modules/<pkg>` segment and are detected; overlay project files fall back to Node / the host loader, restoring pre-hook behaviour.

Verified end-to-end against the local `../nestia@next` checkout: swapped in the patched ttsc and the previously-failing `swagger`, `swagger-한글경로`, and `body-config-assert` sdk features all pass (exit 0, no INestiaConfig error).

PR #210, opened 2026-06-09
fix(ttsc): emit source maps in the plugin-transform emit path

### LINKED ISSUE #209: Plugin-transform emit path drops source maps (sourceMap/inlineSourceMap ignored under AST transforms)
## Summary

When a project is compiled through the **AST plugin-transform emit path** (`driver.EmitWithPluginTransformers`, the seam `typia`/`nestia`-style emit-phase transformers integrate through, and the path the website playground uses), `sourceMap: true` / `inlineSourceMap: true` were silently ignored: the path wrote only the `.js` and produced **no `.js.map`** and no `//# sourceMappingURL=` trailer.

This matters precisely because a transform can expand one authored line (e.g. `typia.is<T>(input)`) into hundreds of emitted lines — exactly the case where a source map is most useful for debugging.

The other emit paths were unaffected:

- Plain tsgo build (`cmd/ttsc build` → `EmitAll`) — maps emitted ✔
- Utility plugin host (`@ttsc/banner` / `@ttsc/strip` / `@ttsc/paths` / `@ttsc/lint` → `EmitAllRaw` → tsgo native emit) — maps emitted ✔
- AST plugin-transform path (`EmitWithPluginTransformers`) — **maps dropped** ✘

## Root cause

`EmitWithPluginTransformers` hand-assembles tsgo's per-file emit pipeline (tsgo's `Program.Emit` / `EmitOptions` has no hook to inject a custom transformer), and called `printer.Write(node, sf, writer, nil)` with a **nil source-map generator**, then wrote only `paths.JsFilePath()`. tsgo's printer gates source maps purely on the generator being non-nil (`sourceMapsDisabled = (sourceMapGenerator == nil)`), so a nil generator means no mappings and no `.js.map`.

## Fix

Port `internal/compiler/emitter.go`'s `printSourceFile` source-map branch into a shim helper `compiler.PrintFileWithSourceMap` and call it from `EmitWithPluginTransformers`:

- builds a `sourcemap.Generator` when `sourceMap` / `inlineSourceMap` is enabled (and the file is not JSON),
- feeds it to the printer so positions are recorded,
- appends the `//# sourceMappingURL=` trailer (or an inline base64 data URL),
- writes the external `.js.map`.

With maps off it is byte-for-byte the prior bare-printer output (zero blast radius).

Covered by a new driver test that expands one source line into 50+ emitted lines under `sourceMap: true` and asserts a valid v3 `.js.map` (referencing `index.ts`, non-empty mappings) plus the trailer.

## typia@next impact

**No code change is required in `typia@next`** for source maps to start emitting — the fix is entirely inside ttsc's emit host, and typia keeps returning the same transformed AST.

Optional follow-up for *precision* (not correctness): synthetic nodes typia builds with the factory carry no original source position, so the expanded region collapses onto the nearest original position (the call site). If typia wants the expanded code to map to more meaningful original spans, it can set source-map ranges on its synthetic nodes during the transform. This is an enhancement, not a prerequisite.

samchon:
## Intent

Fix the **AST plugin-transform emit path** (`driver.EmitWithPluginTransformers`) so `sourceMap` / `inlineSourceMap` actually produce a `.js.map`. This is the seam `typia`/`nestia`-style emit-phase transformers integrate through, and the path the website playground uses. It previously wrote only the `.js`, silently dropping the map — the worst case being a transform that expands one authored line (`typia.is<T>(input)`) into hundreds of emitted lines, exactly when a map matters most.

Closes #209.

## Root cause

`EmitWithPluginTransformers` hand-assembles tsgo's per-file emit pipeline (tsgo's `Program.Emit` exposes no hook to inject a custom transformer) and called `printer.Write(node, sf, writer, nil)` — a **nil source-map generator** — then wrote only `paths.JsFilePath()`. tsgo's printer gates source maps on the generator being non-nil (`sourceMapsDisabled = (sourceMapGenerator == nil)`), so nil ⇒ no mappings, no `.js.map`.

## Change

- **`shim/compiler/emit_sourcemap.go`** (new, hand-maintained): `PrintFileWithSourceMap` ports `internal/compiler/emitter.go`'s `printSourceFile` source-map branch — builds a `sourcemap.Generator` (honoring `sourceRoot` / `mapRoot`), feeds it to the printer, appends the `//# sourceMappingURL=` trailer (or an inline base64 data URL), and returns the external `.js.map` text/path. The shim is allowed to reach `internal/*`, so no new module is needed.
- **`driver/emit_plugin.go`**: the emit loop calls the helper and writes both `.js` and `.js.map`.
- **`test/driver/…emits_source_map_test.go`** (new): a transform expands one line into 50+ emitted lines under `sourceMap: true`; asserts a valid v3 map (references `index.ts`, non-empty `mappings`) and the `sourceMappingURL` trailer.

## Scope / blast radius

With source maps **off**, output is byte-for-byte the prior bare-printer behavior — the new options (`SourceMap`/`InlineSourceMap`/`InlineSources`) are all false and the generator stays nil. Other emit paths (`EmitAll`, `EmitAllRaw`) already emitted maps and are untouched.

## typia@next

No code change required in `typia@next` — the fix is entirely in ttsc's emit host. Optional precision follow-up (not a prerequisite): typia may set source-map ranges on its synthetic nodes so expanded code maps to meaningful original spans instead of collapsing onto the call site. Tracked in #209.

## Test plan

- `go test ./test/driver/` — new test passes; all existing `emit_plugin_*` tests pass.
- Two unrelated **pre-existing** failures reproduce on a clean `master` checkout and are Windows-local only (path-separator / mkdir error-string assertions): `TestDriverLinkedPluginsRegistersAndAppliesProgram`, `TestUtilityCommandFailuresCoverHostEdges`. Green on CI (Linux).
- `pnpm format` clean.


samchon:
Expanded this branch beyond the original plugin-transform fix to cover two more source-map findings surfaced while verifying:

**1. `fix(ttsc): correct source maps shifted by a source-level preamble` (closes #211)**
`@ttsc/banner` injects its banner at the source level, which shifted every `.js.map` **and** `.d.ts.map` mapping down by the banner's line count — real code mapped onto blank lines past EOF. Decoded proof: `exports.first = 111` mapped to source line 9 (blank) of a 5-line file instead of line 0; the `.d.ts.map` had the identical shift. The old banner test only checked the banner *text* was absent from the map, so it never caught this. Fix rewrites the map post-emit (drop preamble-region segments, subtract the shift), runs for both maps and even under `removeComments`. Covered by a Go VLQ round-trip unit test and a banner e2e test that decodes both maps.

**2. `ci(source-map): verify ttsc source maps through a real typia transform`**
New `experimental/source-map` harness + `source-map.yml` workflow: packs current-platform ttsc tarballs, installs them + typia into a scratch project, compiles a one-line `typia.is<T>()` that typia expands into a large validator, and asserts a valid v3 `.js.map`. This exercises the `EmitWithPluginTransformers` path (the original fix) end-to-end through the real typia sidecar.

Verification notes:
- Both banner maps decoded correct after the fix (`exports.first` → source line 0; same for `.d.ts.map`).
- Two pre-existing Windows-only Go test failures (`TestDriverLinkedPluginsRegistersAndAppliesProgram`, `TestUtilityCommandFailuresCoverHostEdges`) reproduce on clean `master` — path-separator / mkdir error-string assertions, green on CI (Linux). Not touched here.

samchon:
Ran a multi-agent review cycle (3 rounds, fresh teams) over the diff and applied every surviving proposal:

**Round 1** — found and fixed real defects beyond the initial commits:
- typia + @ttsc/banner + sourceMap emitted a banner-shifted `.js.map` (the preamble correction was wired only into the utility host, not the plugin-transform path). Centralized via `AdjustEmittedSourceMap`.
- `inlineSourceMap` was never corrected (map embedded in the JS, not a `.map` file). Now decoded/corrected/re-encoded in place.
- Per-source guard: only segments whose source was preamble-injected are shifted, so a bundled `.js`/`.json` map leaves `.json` alone.

**Round 2** — hardening:
- Anchored inline-map detection to the full `//# sourceMappingURL=data:...` comment and restricted scanning to JS/declaration outputs (no `.tsbuildinfo` false-positive).
- The plugin-transform map test was unfalsifiable (1-line source ⇒ every mapping trivially line 0); now 3 lines with the decoded source-line set asserted `{0,1,2}`.
- Added coverage for the previously-untested centralization (`TestEmitWithPluginTransformerCorrectsPreambleShift`), same-line genCol/srcCol integrity, and the VLQ test-helper's own round-trip.

**Round 3** — correctness and holistic reviewers converged (no defect, release-ready); closed the last test-matrix gaps: inline-map guard negatives, the generated-column-only (1-field) re-encode branch, and a one-line docs note in `recipes.mdx`.

Full Go driver/utility suites pass except the two pre-existing Windows-only failures (`TestDriverLinkedPluginsRegistersAndAppliesProgram`, `TestUtilityCommandFailuresCoverHostEdges`) which reproduce on clean `master` and are green on CI (Linux). The new `source-map` workflow exercises the typia path on this PR.

Issue #216, opened 2026-06-11
ttsc JS config readers reject a tsconfig.json saved with a UTF-8 BOM


samchon:
## Summary

A `tsconfig.json` saved with a UTF-8 BOM makes every ttsc JS-side entry that parses project config throw `Unexpected token '﻿' ... is not valid JSON`. Strada (`tsc`) and TypeScript-Go both accept BOM'd config files, so projects that type-check fine with `tsgo` fail under `ttsc` / `@ttsc/unplugin` with a JSON parse error.

Found while testing the #205 fix: a BOM'd fixture tsconfig fails before reaching the unplugin overlay, inside the compiler pipeline.

## Root cause

`packages/ttsc/src/compiler/internal/project/readProjectConfig.ts::parseJsonc` feeds `fs.readFileSync(..., "utf8")` output straight into `JSON.parse` after stripping comments/trailing commas. Node keeps the BOM in the decoded string, and `JSON.parse` rejects it.

## Repro

```js
const fs = require("node:fs");
fs.writeFileSync("tsconfig.json", "\uFEFF" + JSON.stringify({ compilerOptions: { strict: true } }));
// any ttsc entry that reads the project config now throws:
// SyntaxError: Unexpected token '﻿', "﻿{..." is not valid JSON
```

## Fix sketch

Strip a leading `\uFEFF` in `parseJsonc` before `JSON.parse` (one line), plus a unit/e2e case with a BOM'd tsconfig. `@ttsc/unplugin`'s overlay reader already does this (`packages/unplugin/src/core/tsconfigPaths.ts::parseJsonc`); once the core accepts BOM'd configs, an unplugin e2e can pin the end-to-end path too.

PR #217, opened 2026-06-11
fix(unplugin): resolve aliases overlapping tsconfig paths (silent no-op validators)

### LINKED ISSUE #205: typia validators silently become no-ops when a Vite `resolve.alias` overlaps a tsconfig `paths` mapping
If the same alias exists in both tsconfig `paths` and Vite `resolve.alias`, then any `typia.createValidate` (or `assert`, `is`, etc.) on a type imported through that alias compiles to a validator that always passes. The type resolves to `any` at transform time, so typia emits `() => true`. No error, the build succeeds, and runtime validation is silently disabled.

Using a relative import for the type, or removing the alias from `resolve.alias`, fixes it.

## Versions

- `ttsc` 0.15.1
- `@ttsc/unplugin` 0.15.1
- `typia` 13.0.0-dev.20260605.1,
- `@typescript/native-preview` 7.0.0-dev.20260421.2
- `vite` 8.0.16
- node 22.16, Windows 11

## Reproduction

tsconfig.json
```json
{
  "compilerOptions": {
    "module": "ESNext",
    "moduleResolution": "Bundler",
    "strict": true,
    "baseUrl": ".",
    "paths": { "@/*": ["./src/*"] },
    "plugins": [{ "transform": "typia/lib/transform" }]
  },
  "include": ["src"]
}
```

vite.config.ts:

```ts
export default defineConfig({
  plugins: [ttsc({ plugins: [{ transform: 'typia/lib/transform' }] })],
  resolve: { alias: { '@': fileURLToPath(new URL('./src', import.meta.url)) } },
});
```

src/types.ts: `export interface Foo { id: number; name: string }`

src/main.ts:

```ts
import typia from 'typia';
import type { Foo } from '@/types';
const validate = typia.createValidate<Foo>();
console.log(validate({ id: 'not a number', name: 42 })); // reports success
```

Expected: rejects the bad payload. Actual: the emitted validator is `const __is = (input) => true` and the detailed check is `(input, _path, _exceptionable = true) => true`, so it returns `{ success: true }` for anything.

Switching the import to `./types`, or dropping the `@` alias from `resolve.alias`, produces a correct validator.

## Cause

`transformProject` writes a temp tsconfig in `os.tmpdir()` that `extends` the real one and injects `baseUrl` plus `paths` from the Vite aliases (`createAliasCompilerOptions` / `createTransformTsconfig` in `core/transform.js`). Under tsgo, that overlay resolves the aliased import to `any`, and typia builds an always-pass validator from it. Without forwarded aliases the plugin uses the project tsconfig directly and the original `paths` resolve the import fine.

So the overlay is worse than the tsconfig it extends, for aliases the tsconfig already maps. Confirmed by calling `transformTtsc` on the same file with different `aliases`:

| aliases passed                              | validator          |
| ------------------------------------------- | ------------------ |
| `undefined` or `[]`                         | correct            |
| `[{ find: '@', replacement: '<abs>/src' }]` | no-op `() => true` |

## Workaround

`configResolved` only records `config.resolve.alias`, so dropping it makes the plugin fall back to the project tsconfig:

```ts
const ttscPlugin = ttsc(opts);
for (const plugin of Array.isArray(ttscPlugin) ? ttscPlugin : [ttscPlugin]) plugin.configResolved = undefined;
```

This only works if the tsconfig already maps the alias, and the failure is silent, so a no-op validator quietly disables input validation. Please either fix the overlay resolution or warn when a validated type resolves to `any`.

samchon:
## Intent

Fix the silent-no-op validator bug from #205: when the same alias exists in both Vite `resolve.alias` and tsconfig `paths`, the generated transform tsconfig broke module resolution, the aliased type collapsed to `any`, and typia emitted always-pass validators with no error anywhere.

Closes #205.

## Root cause (empirically pinned against tsgo)

The overlay tsconfig is written to a system temp directory and encoded aliases as `baseUrl` + baseUrl-relative `paths` targets. TypeScript-Go:

- **removed `baseUrl`** — any declaration is `TS5102: Option 'baseUrl' has been removed`;
- **rejects bare non-relative `paths` targets** — `TS5090: Non-relative paths are not allowed`, and the mapping is dropped;
- accepts **absolute** targets.

So the entire `paths` overlay was discarded. In the plugin transform path these config diagnostics don't fail the run, which is what made the failure silent. A second instance of the same class: the overlay `extends` the project tsconfig, and declaring `paths` replaces the project's own `paths` **wholesale** — so even with a working alias encoding, every tsconfig-only mapping (`#lib/*` etc.) would silently stop resolving the moment any bundler alias was forwarded.

## Change

- `core/transform.ts` — alias targets are written as absolute paths; no `baseUrl` is ever emitted; inline `compilerOptions.paths` targets are absolutized against the original tsconfig directory (previously they leaked into the temp dir un-anchored, broken in the same way).
- `core/tsconfigPaths.ts` (new) — reads the project's effective `paths` (nearest declaration in the `extends` chain; string/array `extends`, relative/absolute/bare specifiers, JSONC, circular-chain guard, UTF-8 BOM) and absolutizes targets against the **declaring** config's directory (verified: tsgo anchors inherited relative targets at the declaring file, not the extending one). The overlay re-states these mappings under the forwarded aliases.

The fix lives entirely in `@ttsc/unplugin`; ttsc's public surface is untouched. The JSONC/extends helpers intentionally mirror ttsc's internal `readProjectConfig` rather than exporting it — the package root keeps project-parsing helpers internal by design, and this reader is a focused `paths`-only subset.

## Tests

Regression style: a deliberate type error imported *through* the alias can only be reported when the type really resolved — silent `any`-collapse fails the test.

- `test_transformttsc_resolves_types_through_alias_overlapping_tsconfig_paths` — the exact #205 shape, plus a well-typed negative twin.
- `test_transformttsc_alias_overlay_preserves_unaliased_tsconfig_paths` — tsconfig-only `#lib/*` survives the overlay merge.
- `test_transformttsc_alias_overlay_merges_paths_from_extended_jsonc_tsconfig` — `paths` declared in an extended JSONC base in a different directory.
- `test_transformttsc_generated_tsconfig_omits_baseurl_and_uses_absolute_alias_targets` — config-shape guard via a new `assert-absolute-alias-paths` fixture-plugin operation.
- **Updated expectation (deliberate contract change):** `test_transformttsc_passes_bundler_aliases_through_compileroptions_paths` now expects the absolute alias target — the old relative-target expectation *was* the bug.

Verified locally on Windows by direct invocation (the `DynamicExecutor` runner imports absolute paths as URLs and cannot start on Windows — pre-existing; CI runs Linux): all 27 `test-unplugin` feature tests pass, before/after live repro confirmed (control reports the type error; pre-fix alias run died on TS5102/silently lost resolution; post-fix alias run reports the type error).

## Deferred

- BOM-prefixed tsconfigs are rejected by ttsc's core config reader before the overlay is reached — filed as #216. The overlay reader here already tolerates BOM so an e2e can pin the full path once the core accepts it.
- Alias capture is still Vite-only (`vite.configResolved`), unchanged by this PR.


samchon:
Pushed a follow-up: the two fixture-shape tests (assert-paths — pre-existing — and the new assert-absolute-alias-paths) nested their options under a 'config' object, but the protocol forwards the whole plugins[i] entry as the plugin's config, so the fixture fell back to its default operation and both passed vacuously. Options now sit at the entry top level; verified non-vacuous by flipping the expected target and watching the Go assertion fail.

samchon:
﻿Pre-merge self-review complete for issue #205.

1. Diff read-through: reviewed the changed source, tests, and docs for this PR. Alias overlay keeps project tsconfig paths, encodes absolute targets, and omits removed baseUrl.

Merge-ready. The issue should close through the existing Closes #205 linkage when the PR is merged.

Issue #218, opened 2026-06-11
ttsx virtual layout: symlinked entries fail with EPERM on Windows (junction fallback missing in linkVirtualEntry)


samchon:
## Summary

`ttsx`'s virtual project layout (`prepareExecution.ts::linkVirtualEntry`) fails with `EPERM` on Windows without Developer Mode/elevation when an entry it mirrors is itself a symlink. Plain directories already use junctions on win32, but the special-entry fallthrough re-symlinks as-is:

```ts
// packages/ttsc/src/launcher/internal/prepareExecution.ts
if (entry.isDirectory()) {
  fs.symlinkSync(realEntry, virtualEntry,
    process.platform === "win32" ? "junction" : undefined); // ok on Windows
  return;
}
...
// Symlinks (and other special entries) are re-symlinked as-is.
fs.symlinkSync(realEntry, virtualEntry); // EPERM on Windows for directory targets
```

A `node_modules` that is itself a symlink (pnpm-style links, or test fixtures linking the workspace `node_modules`) hits the fallthrough, and directory symlink creation needs privilege on Windows.

## Observed

`node scripts/test-go-lint.cjs` on Windows 11 (non-elevated):

```
@ttsc/lint: load TypeScript config file ...\ttsc-lint.config.ts:
EPERM: operation not permitted,
symlink '...\ttsc-lint-config-XXXX\node_modules' -> '...\cache\project\NNNN\fs\C\Users\...\node_modules'
```

`TestLoadRuleConfigLoadsTypeScriptConfigFile` and `TestLoadRuleConfigTypeScriptConfigFileRoundTripsFormatBlock` fail with it; every ttsx-backed flow whose project carries a symlinked entry is affected the same way. CI (Linux) is unaffected, which is why it has gone unnoticed.

## Suggested fix

In the fallthrough, when `realEntry` resolves to a directory on win32, create a junction (`fs.symlinkSync(target, path, "junction")`); junctions need no privilege and Node traverses them transparently. Non-directory symlinks could fall back to hard-link/copy like the file branch.

PR #219, opened 2026-06-11
perf(lint): skip declaration-file dispatch for rules that don't opt in

### LINKED ISSUE #177: perf(lint): skip .d.ts walks for rules that don't opt into declaration files
## Summary

Add an opt-in `Rule` interface marker so rules that don't care about `.d.ts` files can skip the entire walk on declaration-only inputs. The engine currently walks every file in the tsconfig include set, value-level rules included — which on `.d.ts`-heavy codebases (vscode src has many; type-fest is almost all `.d.ts`) burns CPU on AST traversal that can never produce a finding.

This is a follow-up to PR #176 (the Round 1 multi-agent audit) — see `.discussions/lint-review-2026-05-28/perf-results.md` for the dispatcher-side wins. Profiling showed dispatcher / closure overhead was the easy 80%; the remaining ~10-20% on `.d.ts`-heavy fixtures is the per-rule wasted walk this issue targets.

## Why

CPU profile on `vscode:ttsc-lint:noEmit` (under the Round 1 dispatcher optimizations):

* `ast.(*Node).ForEachChild` 60-67% cum — the AST walker is the dominant in-lint cost.
* The lint pass walks **every** file in tsconfig include, including `.d.ts`. Most value-level rules (`no-var`, `prefer-const`, `eqeqeq`, `object-shorthand`, `no-extra-boolean-cast`, …) cannot fire on a declaration file because the syntactic forms they detect (`var`, mutable `let`, runtime conditionals, etc.) don't exist there.
* On `type-fest` (209 `.d.ts` files, ~16 ms pure lint), the walk is mostly wasted. On `vscode` (mixed) it's a measurable fraction of the 833 ms pure lint time the dispatcher win currently lands at.

The engine has no way to know in advance which rules care about `.d.ts`, so it walks for all of them.

## Proposed design

Add an optional method to the `Rule` interface (mirroring `FormatRule` / `typeAwareRule` marker pattern in `packages/lint/linthost/engine.go`):

```go
// declarationFileRule marks rules that want to fire on `.d.ts` (and
// other declaration-file) inputs. Rules that do NOT implement this
// interface are skipped on every `file.IsDeclarationFile == true`
// source, saving the per-rule AST walk on declaration-only inputs.
//
// Examples:
//   - `typescript/no-explicit-any` — IS a declaration-file rule
//     (type annotations are precisely what it inspects).
//   - `no-var` — is NOT (a `var` declaration cannot appear in a
//     `.d.ts` file).
type declarationFileRule interface {
    VisitsDeclarationFiles() bool
}
```

In `runFile`:

```go
if file.IsDeclarationFile {
    // Filter byKind in place to keep only rules that opted in.
    for kind := range byKind {
        filtered := byKind[kind][:0]
        for _, bound := range byKind[kind] {
            if r, ok := bound.rule.(declarationFileRule); ok && r.VisitsDeclarationFiles() {
                filtered = append(filtered, bound)
            }
        }
        byKind[kind] = filtered
    }
}
```

Default is **skip** (rules opt IN, not OUT) so a contributor that doesn't think about it doesn't accidentally police declarations they didn't mean to.

## Files to touch

1. `packages/lint/linthost/engine.go` — add the optional interface, the per-file filter, doc comments.
2. `packages/lint/rule/rule.go` — expose the same optional method on the public `rule.Rule` surface so contributor rules can opt in too.
3. **Audit every built-in rule under `packages/lint/linthost/rules_*.go`** — for each rule, decide:
   * Does this rule's grammar shape ever appear in a `.d.ts`? If yes, implement `VisitsDeclarationFiles() bool { return true }`. If no, leave it (skip default).
   * Likely yes-list (non-exhaustive, audit needed):
     * `typescript/no-explicit-any`, `typescript/no-empty-interface`, `typescript/no-magic-numbers`,
     * `typescript/sort-type-constituents`, `typescript/prefer-as-const` (sometimes appears in declarations),
     * `jsdoc/*` (declaration files have JSDoc),
     * `tsdoc-syntax`, `unicorn/filename-case`.
   * Likely no-list (vast majority): everything that targets value-level expressions, statements, calls — `no-var`, `prefer-const`, `eqeqeq`, `prefer-template`, `dot-notation`, etc.
4. `packages/lint/test/engine/engine_skips_declaration_file_walk_test.go` (new) — assert that a non-opt-in rule sees 0 nodes when the only input is a `.d.ts`, and an opt-in rule sees them.
5. `packages/lint/test/rules/typescript/typescript_no_explicit_any_visits_declaration_files_test.go` (new — pick one representative declaration-file rule) — assert `VisitsDeclarationFiles()` returns true and the rule fires on a `.d.ts` fixture.
6. `packages/lint/README.md` — document the new opt-in.
7. `website/src/content/docs/lint/development/walkthroughs/lint.mdx` (or the contributor guide path under `docs/development/`) — same.

## Test plan

* Existing Go unit tests stay green (`node scripts/test-go-lint.cjs`).
* New unit tests pin the skip + opt-in semantics.
* Microbench (`node scripts/bench-go-lint.cjs`): add `BenchmarkEngineRunOverDeclarationFile` that parses a representative `.d.ts` and runs the engine — should show ≥ 90% time reduction vs the same input walked with no skip.
* Real fixture bench: `vscode:ttsc-lint:noEmit` should show pure-lint median improvement of at least 5% (heuristic — vscode src has a modest .d.ts fraction).
* Don't relax acceptance based on type-fest — its pure lint is already 16 ms, the win there is invisible.

## Acceptance

* PR description includes vscode pure-lint before/after numbers from a clean A/B.
* No existing built-in rule's behavior changes on `.ts` files.
* Every built-in rule has been auditedreviewed for declaration-file applicability and either marked or explicitly left at the default-skip.
* Contributor docs explain when to opt in.

## Related

* PR #176 (Round 1 multi-agent audit + dispatcher perf): https://github.com/samchon/ttsc/pull/176
* Dispatcher numbers + method: `.discussions/lint-review-2026-05-28/perf-results.md` (committed to that PR).
* Decision to remove `type-fest` from the bench (not anti-related; that fixture's lint cost is too small to surface this win): commit `84b22540` on PR #176.

samchon:
## Intent

Implement #177: stop dispatching value-level rules on declaration files. A `.d.ts` carries no executable code, so most rules can never fire on one; the engine still walked and dispatched everything, burning CPU on `.d.ts`-heavy include sets.

Closes #177.

## Design (and where it deviates from the issue — deliberately)

`Engine.runFile` binds only declaration-visiting rules on `IsDeclarationFile` sources, and skips the whole AST walk when nothing bound (inline-directive parsing is kept so unknown-directive warnings stay file-complete). A rule participates in declaration files through one of three doors:

1. **Format rules always visit.** The issue did not address formatters; skipping them would have regressed `ttsc format` / `ttsc fix` on hand-written `.d.ts`. The engine treats every `FormatRule` as declaration-visiting, no per-rule boilerplate.
2. **Built-ins: a curated allowlist, not per-rule interface methods.** The registry holds **743 rules**; scattering `VisitsDeclarationFiles()` methods across ~300 files would make the audit unreviewable. `linthost/declaration_rules.go` holds the whole audit in one file — 98 entries with per-group rationale (JSDoc, import/boundary graphs, type syntax, enums and their constant initializers, naming, file metrics, user-configured selectors). A registry-parity test pins every entry to a registered rule so a rename can't silently turn an opt-in into a skip. Audit principle, stated in the file: *skip only when the rule's grammar shape cannot occur in a declaration file; when in doubt, list it — a wrong listing costs dispatch time, a wrong skip silently loses findings.*
3. **Contributors keep running by default**, with the new public `rule.DeclarationFileRule` marker to opt out. The issue implied opt-in for everyone; the adapter instead mirrors its own `NeedsTypeChecker` precedent — the host cannot infer a third-party rule's shape, and a skip-by-default would silently change existing contributor behavior.

Also corrected from the issue text: there is no per-rule AST walk — one shared walk dispatches per kind — so the full walk is saved only when *zero* enabled rules opt in for the file; otherwise the saving is the per-node dispatch to value rules. One audit disagreement: `typescript/prefer-as-const` stays default-skip (the issue's tentative yes-list said it "sometimes appears in declarations", but ambient contexts forbid initializers, which are the only place `as const` rewrites apply).

This replaces the previous walk-everything contract pinned by `engine_lints_supplied_declaration_files_test.go` (2026-05-24); issue #177 (2026-05-28) is the newer intent and explicitly reverses it.

## Numbers

Microbench (new, in `engine_run_declaration_bench_test.go` — identical input, ten value rules):

```
BenchmarkEngineRunOverDeclarationFile-32                 115242    20999 ns/op   22936 B/op   53 allocs/op
BenchmarkEngineRunOverDeclarationShapedValueFile-32       20184   122575 ns/op   24240 B/op   81 allocs/op
```

≈5.8× (-83%) on the declaration file; the remaining cost is severity resolution + directive parsing, preserved on purpose. **Deferred:** the issue's `vscode:ttsc-lint:noEmit` clean A/B needs the quiet publication host and a double workspace pack (benchmark skill procedure); not runnable faithfully on this dev machine. The mechanism's e2e effect should be validated there before relying on the headline number.

## Windows test infrastructure (first commit)

The lint Go suite had never run on Windows; getting it running was a prerequisite for validating this change locally and surfaced real defects:

- `scripts/test-go-lint.cjs` / `scripts/bench-go-lint.cjs`: the scratch `go.work` now `use`s `"."` — Go canonicalizes the Windows temp dir differently than Node spells it and rejected the absolute entry.
- `parseTSFile`/`parseTSXFile` normalize to forward slashes (the tsgo parser asserts on normalized paths; every `t.TempDir()` case panicked).
- **Real product fix:** the LSP `--content-stdin` format path passed the URI-derived OS path straight to the parser — any formatOnSave through ttscserver panicked the sidecar on Windows. Now normalized in `lsp.go`.
- Remaining Windows-local failures: the ttsx symlink-EPERM pair, filed as #218; they reproduce without this change.

## Tests

- `engine_skips_value_rules_on_declaration_files_test.go` — `no-debugger` silent on a `.d.ts` (replaces the walk-everything test).
- `engine_runs_opt_in_rules_on_declaration_files_test.go` — `typescript/no-explicit-any` fires on `declare const x: any;` (negative twin).
- `engine_runs_format_rules_on_declaration_files_test.go` — `format/semi` fires on a `.d.ts`.
- `declaration_rule_names_match_registry_test.go` — allowlist↔registry parity.
- `contributor_rules_visit_declaration_files_by_default_test.go` / `contributor_rule_opt_out_skips_declaration_files_test.go` — adapter default + public marker.
- Full `node scripts/test-go-lint.cjs` green on Windows except the pre-existing #218 pair; CI (Linux) runs the complete suite.
- Existing e2e `.d.ts` fixtures stay covered: `format-sort-imports/src/ambient.d.ts` (format door) and the `triple-slash-reference` case (allowlist door).

## Docs

`packages/lint/README.md` (contributor section) and `website/.../walkthroughs/lint.mdx` (Rule interface section) describe the three doors and the contributor marker.


samchon:
﻿Pre-merge self-review complete for issue #177.

1. Diff read-through: reviewed the changed source, tests, and docs for this PR. Declaration-file dispatch now skips value-only rules while preserving format, opt-in, allowlisted, and contributor-default behavior.

Merge-ready. The issue should close through the existing Closes #177 linkage when the PR is merged.

PR #220, opened 2026-06-11
feat(ttsc): plugin-reported dependency lists for bundler HMR (addWatchFile)

### LINKED ISSUE #214: plugin protocol: expose per-file dependency list for bundler HMR (addWatchFile)
## Summary

Migrated from samchon/typia#1722, where it was filed against typia v12's `@typia/unplugin` — in the v13 architecture the right home is the `ttsc` plugin protocol.

Dev-mode bundlers (vite, esbuild) erase type-only imports, so a file like:

```ts
import type { MyType } from "./types";
const validate = typia.createValidate<MyType>();
```

has no dependency edge to `./types` in the bundler graph. Editing `MyType` does not invalidate the module, so HMR serves a stale generated validator until a cold restart.

## Request

Expose, per transformed file, the list of source files the transform actually consulted — i.e. extend the plugin protocol result from "transformed output" to "transformed output + dependency file list" — and have `@ttsc/unplugin` register those paths with the bundler (`this.addWatchFile()` in rollup/vite terms) so type-only dependencies participate in HMR invalidation.

## Feasibility notes (from the typia side)

The producer half is structurally natural: typia's `MetadataFactory.Analyze` walks checker symbols for every analyzed type, and each symbol's `Declarations[].SourceFile().FileName()` is available right there — collecting the set per transformed file is cheap. What is missing is the protocol slot to return it through (`ttsc` plugin result shape) and the adapter wiring in `@ttsc/unplugin`.

Original report with reproduction context: samchon/typia#1722 (author: @kahoot-karl).

samchon:
## Intent

Implement #214: give transform plugins a protocol slot to report, per transformed file, the source files they consulted, and have `@ttsc/unplugin` register those paths as bundler watch files. Dev-mode bundlers erase type-only imports from their module graph, so editing a type that only generated code depends on (`typia.createValidate<T>()`) never invalidates the module — HMR serves a stale validator until a cold restart.

Closes #214.

## Protocol

The `transform` stdout envelope gains an optional field (the protocol explicitly permits new JSON fields, so this is compatible — existing plugins are untouched):

```json
{
  "typescript": { "src/main.ts": "..." },
  "dependencies": { "src/main.ts": ["src/types.ts"] }
}
```

Keys mirror `typescript`; values are project-relative or absolute consulted-source paths. ttsc forwards the record verbatim on `TtscCompiler.transform()` results (`ITtscCompilerTransformation.dependencies`, success and failure shapes). Malformed entries are dropped, never fatal — the field is advisory watch metadata, not output. Documented in the protocol page (`transform` section).

## Adapter wiring

`transformTtsc` gains an optional trailing `hooks` parameter (`TtscTransformHooks`, exported) so the dependency side-channel reaches the bundler without leaking extra fields on the returned `TransformResult`. The core transform hook passes `addWatchFile` through. Per-file selection mirrors the existing output lookup (project-relative key, then resolve-scan), absolutizes against the project root, deduplicates, and drops the module itself. Notifications fire on **cache-served** transforms too — watch registrations are per module request while the compiler result is shared, so a watch-mode rebuild of an unchanged project must not lose its invalidation edges.

The producer half (typia walking `MetadataFactory.Analyze`'s checker symbols) lives downstream; this PR provides the slot and the consumer wiring, exercised by fixture plugins.

## Tests

- `test_ttsccompiler_transform_surfaces_plugin_dependency_lists` (test-ttsc, API level) — envelope → `transform()` result passthrough.
- `test_transformttsc_forwards_plugin_dependencies_to_the_watch_hook` — absolutized, deduplicated, module-self excluded.
- `test_transformttsc_adds_no_watch_files_without_plugin_dependencies` — negative twin: no envelope field → zero registrations.
- `test_transformttsc_notifies_watch_files_on_cached_transforms` — cache-hit replay.
- `test_rollup_build_registers_plugin_dependencies_as_watch_files` — end-to-end: dependency lands in rollup `bundle.watchFiles`, the channel watch mode consumes.
- Fixture growth: the unplugin Go fixture gains an `emit-dependencies` operation; the test-ttsc compiler fixture now reports a consulted file unconditionally (existing API tests unaffected — verified).

All five run green locally on Windows via direct invocation (the DynamicExecutor runner cannot start on Windows — pre-existing; CI runs the full suites on Linux), alongside the existing transform API tests.

## Notes for review

- While writing these tests I found that fixture plugin options must sit at the **entry top level** (the protocol forwards the whole `compilerOptions.plugins[i]` entry as the plugin's config). Two existing/new tests on the #205 branch use a nested `config: {...}` object and therefore pass vacuously; I'll push a fixup to PR #217 separately.
- This branch will conflict textually with #217 in `core/transform.ts` and the shared Go fixture; whichever merges second gets a mechanical rebase from me.


samchon:
﻿Pre-merge self-review complete for issue #214.

1. Diff read-through: reviewed the changed source, tests, and docs for this PR. Transform envelope dependencies are parsed as advisory metadata and replayed to bundler watch hooks, including cache hits.

Merge-ready. The issue should close through the existing Closes #214 linkage when the PR is merged.

PR #222, opened 2026-06-11
fix(ttsx): expose CJS star exports to ESM importers

### LINKED ISSUE #200: ttsx(enhancement): run-from-source에서 ESM 소비자가 CJS-소스 형제의 named export를 import 못 함
> 실증으로 원래 프레이밍을 정정한다. 타입체크 TS2835는 ttsx 버그가 아니라
> tsgo와 100% 동일 동작이다(아래 실측). 진짜 한계는 런타임 ESM↔CJS-소스
> named-import 상호운용이다. 제목/본문을 그에 맞게 다시 쓴다.

## 실증: 타입체크 TS2835는 ttsx 버그가 아니다

typia `tests/test-langchain`을 nodenext(베이스 tsconfig) 그대로 두고 무확장자
상대 임포트를 타입체크하면, **standalone tsgo와 ttsc가 똑같이 TS2835를 8개**
낸다.

```
# tsgo --noEmit -p tsconfig.json   -> TS2835 x 8
# ttsc check                       -> TS2835 x 8  (동일 파일/위치)
src/index.ts(4,28): error TS2835: Relative import paths need explicit file
extensions ... when '--moduleResolution' is 'node16' or 'nodenext'.
```

즉 nodenext+ESM에서 무확장자 상대 임포트 거부는 표준 TypeScript 동작이고
ttsx는 충실하다. 해결은 `moduleResolution: bundler`(무확장자 허용) 또는 확장자
추가. 이 부분은 ttsx가 고칠 게 없다.

## 진짜 한계: 런타임 ESM↔CJS-소스 named-import 상호운용

`moduleResolution: bundler` + `type: "module"`로 타입체크를 통과시킨 뒤
`ttsx src/index.ts`를 실행하면 런타임에서 깨진다.

```
file:///.../tests/test-langchain/src/TestGlobal.ts:2
import { OpenApiConverter, Singleton } from "@typia/utils";
SyntaxError: The requested module '@typia/utils' does not provide an export
named 'OpenApiConverter'
```

원인 체인:
- 소비자(test-langchain)는 ESM(`type: "module"`).
- `@typia/utils`는 package.json에 `type` 필드가 없어 **CJS로 분류**되고
  `exports`가 소스(`src/index.ts`)를 가리켜 ttsx가 소스를 CJS로 로드한다.
- `@typia/utils/src/index.ts`는 `export * from "./converters/index"` 식의
  중첩 re-export 체인이다. ttsx가 이를 CJS로 lower하면, ESM 소비자 쪽
  Node 로더의 cjs-module-lexer가 중첩 re-export를 통해 named binding을
  열거하지 못해 "does not provide an export named ..."가 된다.

확인된 사실(`pnpm start` 실측):
- 같은 스위트를 **CJS로 두면 정상 통과**(Success, exit 0).
- ESM으로 바꾸면 named import에서 즉시 실패.
- test-utils-automated도 `import { dedent } from "@typia/utils"`로 동일하게 실패.
- named import가 없는 스위트(test-error, smoke의 `import typia from "typia"`
  기본 import)는 ESM에서도 정상.

이것 자체는 Node 표준 ESM↔CJS interop 한계라 ttsx "오역"은 아니다. 다만
ttsx는 run-from-source라 **양쪽 모듈 포맷을 모두 자기가 로드/제어**하므로,
번들러처럼 CJS-소스 형제 모듈의 named export를 ESM 소비자에게 노출해줄 수
있는 위치에 있다. 지금은 못 해서, 모노리포 안에서 ESM 스위트가 CJS-소스
라이브러리의 named export를 못 쓴다. 이게 enhancement 포인트다.

## 영향

typia AST 통합 브랜치에서 무거운 자동화 스위트(test-utils-automated,
test-langchain)를 ESM으로 올려 오펀 콜드 비용을 줄이려 했으나, 이 한계로
런타임이 깨져 CJS로 되돌렸다. (콜드 비용 자체는 ttsc `--noCheck` 오펀 emit으로
이미 해결돼 당장은 막힘이 아니다.) 모노리포 ESM 전환을 매끄럽게 하려면 ttsx
쪽 지원이 필요하다.

## 해결 프롬프트 (enhancement, 새 세션에서 실행 가능)

목표: ttsx run-from-source에서, ESM으로 분류된 소비자가 CJS로 분류된 소스
형제 모듈의 named export를 import할 수 있게 한다. 기본 동작/CJS 소비자는
회귀 없이 유지.

1. 재현 최소화: `tests/test-ttsc`에 한 케이스로 봉인.
   - 프로젝트 A(소비자): `type: "module"`, `import { foo } from "lib"`.
   - lib: `type` 없음(CJS 분류), `src/index.ts`가
     `export * from "./inner"` 중첩 re-export로 `foo` 노출.
   - 기대: ttsx 실행 시 `foo`가 정상 바인딩되어 실행/출력.
   - 네거티브: 현재는 "does not provide an export named foo"로 실패(현 상태 봉인).
2. 설계: ttsx load 훅(`packages/ttsc/src/launcher/internal/runtimeHooks.ts`)이
   CJS-분류 소스를 ESM 소비자에게 제공할 때 named binding을 노출하는 경로를
   검토한다. 후보:
   - (A) CJS-소스를 ESM 소비자가 import할 때 한해, 소스의 원래 ESM 구문을
     보존해 ESM 모듈로 로드(소스가 `export *`라 정보가 있음). 단 동일 모듈을
     require하는 CJS 소비자와의 dual-format 위험을 막을 것.
     [[project_ttsx_run_from_source_architecture]]의 분류 규칙과 충돌하지 않게.
   - (B) CJS lower 산출물을 cjs-module-lexer가 named export를 열거할 수 있는
     형태로 만들거나, ttsx가 named binding 목록을 합성해 ESM 측에 제공.
   더 단순하고 회귀 적은 쪽 선택. 새 서버/프로토콜/캐시 도입 금지(과설계 금지).
3. 검증: `node scripts/test-go-lint.cjs` 전체 green. typia
   test-utils-automated/test-langchain를 ESM으로 되돌려도 통과하는지
   별도 확인(이 이슈가 풀리면 그 되돌림이 가능해진다).
4. 회귀: CJS 소비자가 같은 CJS 라이브러리를 require하는 기존 경로가
   그대로 동작해야 한다.

## 메모

- 타입체크 무확장자 문제는 `moduleResolution: bundler`로 해결되며 ttsx 변경
  불필요(tsgo 동일).
- 본 이슈는 런타임 interop enhancement이며 bug가 아니다. 우회책은 해당
  소비자를 CJS로 두거나 라이브러리에 `type: "module"` 부여.

samchon:
## Summary

- Exposes nested CommonJS `export *` names to Node's ESM-from-CJS named import scanner while preserving the CommonJS `require` runtime path.
- Marks only URLs resolved from an ESM parent into a CommonJS-classified TypeScript source, then rewrites tsgo's `__exportStar(require(...), exports)` output into static `exports.name = void 0` placeholders plus the same `__createBinding` getters.
- Resolves names from emitted JavaScript when available; for source-only packages, derives fallback names from tsgo's CommonJS emit instead of parsing TypeScript text.
- Keeps type-only exports, comments, string-literal decoys, and the CJS require twin hidden/unchanged.
- Adds a #200 regression for ESM named imports, CJS require parity, nested source barrels, renamed exports, namespace re-exports, and false-export negatives.
- Documents the CJS-classified source package named-import behavior.

## Investigation

- Reproduced #200 as an ESM consumer failing to import named exports from a CJS-classified source package that exposes values through nested `export *`.
- Confirmed the TS2835 extension diagnostic described in the issue is tsgo-compatible TypeScript behavior, not a ttsx bug.
- Investigated the later typia CI slowdown on this PR by comparing against a fresh master control run. Master completed normally in 8m40s overall, so the Go binary build cache was not the primary suspect for this branch.
- The branch exceeded the 20-minute typia threshold only after an over-broad runtime-generated-source repair was added.
- Root cause of the slow PR branch: that generated-source repair rebuilt runtime-created entry-project source directories through the entry tsconfig/plugins path. Typia generates and executes many feature files, so this was outside #200 and too expensive.
- Final scope: remove the generated-source rebuild/cache path entirely. Keep only the CJS source named-export interop fix, and keep export-name scanning separate from the runtime orphan cache so speculative scans cannot perturb later loads.
- No plugin-name branching, typia-specific logic, monkeypatching, server/protocol layer, or broad generated-source cache remains.

## Verification

- No local test/build/install/pack commands were run for the latest fix, per the explicit local-test ban.
- `git diff --check` before the latest push.
- Fresh master typia control: https://github.com/samchon/ttsc/actions/runs/27418540768 succeeded.
- The over-broad generated-source PR run was canceled after exceeding the 20-minute threshold: https://github.com/samchon/ttsc/actions/runs/27422827206
- A later superseded PR run at `8a8d650` also exceeded the 20-minute typia threshold, confirming the generated-source path was still too broad.
- Latest PR CI for `ce34419d` is running: https://github.com/samchon/ttsc/actions/runs/27428023342

## Self Review

- Pass 1, diff read: final PR diff is limited to `runtimeHooks.ts`, one #200 ttsx runtime regression, and execute documentation.
- Pass 2, contract review: module classification remains Node/tsgo-driven; CommonJS consumers still use `require`; named export placeholders are applied only to CommonJS TypeScript sources reached from an ESM parent; no entry-project manifest or generated-source behavior is changed.
- Pass 3, adversarial review: no hardcoded typia behavior, timeout shortcut, plugin binary cache change, transform-result replay, generated-source rebuild, daemon/protocol/cache layer, or TypeScript-source regex parser remains.

Closes #200


samchon:
Follow-up after the typia CI run stayed in \Test typia next\ far past normal duration. The previous cancelled logs showed generated feature groups triggering repeated large rebuild gaps, so \4449279d\ adds a narrow sibling-directory emit before falling back to the existing full project rebuild for generated source cache misses.\n\nVerification after this commit:\n- \pnpm format\\n- \pnpm --dir packages/ttsc build\\n- direct #200 runtime regression\n- generated source/plugin cache regression\n- non-empty proof for the generated source/plugin regression by intentionally changing the expected value to \WRONG\ in a same-directory temp copy and confirming the assertion fails\n\nThe previous \sdk\ failure was rerun and passed; its first failure was the Nestia fixture port collision (\EADDRINUSE\).

samchon:
Follow-up for the `typia` CI timeout after `4449279d`.

Root cause observed from the failed Ubuntu typia job: the generated-source runtime miss path was still allowed to fall back to repeated project builds when a file was created after the entry build. On typia's larger fixture set that made the job reach the 60-minute Actions limit even though the functional #200 regression was fixed.

This commit changes the miss path to prefer a narrow source-to-source transform shard:

- entry-project generated misses skip the up-front full dependency build and go straight to the missing-source refresh path;
- generated source refresh now tries `transformProjectInMemory` for sibling TS files, emits those transformed sources without plugins into the existing dependency cache, then falls back to the previous build-shard/full-build paths when transform is unavailable;
- the temporary transform project links `node_modules` best-effort and keeps inherited plugin options visible to sidecars that inspect the passed tsconfig text.

Additional regression coverage:

- added `test_plugin_corpus_ttsx_transforms_generated_sources_without_build_fallback`, whose plugin intentionally fails `build` after `src/generated` exists, so it only passes through the transform-shard path;
- preserved the existing build-shard fallback coverage for plugins without transform support.

Local verification:

- `pnpm format`
- `pnpm --dir packages/ttsc build`
- direct #200 regression runner: `test_ttsx_exposes_nested_cjs_source_star_exports_to_esm_named_imports`
- new transform-shard generated-source runner
- existing generated-source cache-miss fallback runner
- non-empty proof: changed the new test expected stdout from `SECOND` to `WRONG` and confirmed the assertion fails, then restored and reran green
- attempted `pnpm --dir tests/test-ttsc start`, but the Windows TestExecutor path import hits Node's `ERR_UNSUPPORTED_ESM_URL_SCHEME` (`d:` absolute path URL); targeted scratch runners above are the local Windows-compatible verification path.

Self-review done in the required three passes: staged diff read, contract/consistency check, and adversarial review of fallback, cache, and Windows junction behavior.

samchon:
Follow-up for the second `typia` timeout on `4f264875`.

The failed run again reached `test_http_createValidateHeaders_ObjectHttpUndefindable` and then sat until the 60-minute job limit. Comparing older #222 runs showed the same post-`createValidateHeaders` stall: the #200 CJS star-export interop makes `@typia/template` expose the full generated corpus, and the next generated-source miss can hold a large native transform/build shard for too long.

This commit keeps the broad batch path for the normal fast case, but adds bounded fallback behavior:

- `spawnNative` now accepts an optional timeout.
- `transformProjectInMemory` forwards that timeout only when an internal caller asks for it.
- ttsx generated-source refresh now tries sibling batch transform first, then single-file transform, then sibling build, then single-file build, then the existing full-project build fallback.
- The transform shard timeout is 10 minutes so the first large typia batch can still complete, but a stuck later batch cannot consume the whole 60-minute workflow.
- The generated-source plugin fixture now intentionally rejects multi-file transform batches after `src/generated` has more than one file, proving the single-file transform fallback is non-empty and still avoids the build fallback.

Local verification after this commit:

- `pnpm format`
- `pnpm --dir packages/ttsc build`
- direct #200 regression runner
- new generated transform-shard runner, now covering batch failure -> single-file transform fallback
- existing generated cache-miss build fallback runner
- non-empty proof: changed the new test expected stdout from `SECOND` to `WRONG`, confirmed assertion failure, restored, then reran green

Self-review repeated in the required three passes: staged diff read, contract/consistency review, and adversarial review of timeout, fallback ordering, cache reuse, and Windows process behavior.

samchon:
Follow-up for the third `typia` timeout on `be960173`.

The failed Ubuntu typia log again reached `test_http_createValidateHeaders_ObjectHttpUndefindable` quickly, then produced no test output until the 60-minute workflow cancellation. That confirms the previous 10-minute native transform timeout was still too coarse: a large generated feature-set batch could spend the remaining job budget before the single-file fallback ran.

This commit keeps sibling-directory shards for small generated groups, but caps generated source shards at 16 files. Larger generated directories now go straight to the single-file transform/build fallback path instead of trying one large native transform batch first. The fixture now covers both sides:

- two generated files still prove small-batch transform failure falls back to single-file transform;
- a later directory with more than 16 generated files proves the large batch is skipped, using a marker file that would fail the test if the batch path ran;
- the build fallback remains rejected in that fixture, so success still proves source-to-source transform handled the generated miss.

Local verification after this commit:

- `pnpm format`
- `pnpm --dir packages/ttsc build`
- direct #200 runtime regression runner
- generated transform-shard runner with the new large-batch skip assertion
- existing generated cache-miss build fallback runner
- non-empty proof: changed the new fixture expected stdout from `THIRD` to `WRONG`, confirmed assertion failure, restored, then reran green
- `git diff --check`

Self-review repeated in the required three passes: staged diff read, contract/consistency review, and adversarial review of cap boundary, fallback ordering, and marker-file coverage.

samchon:
Follow-up for the long-running typia job:

- Verified the previous cap still had a bad shape for generated corpora: it avoided one huge native transform batch by falling back to one-file refreshes, which repeats native program startup for every generated file in large directories.
- Replaced that with deterministic bounded chunks of up to 16 sibling TS files containing the requested source. If the source cannot be matched back to the sibling list, it falls back to the original single-file behavior.
- Extended the generated-source plugin fixture so it fails on oversized transform batches, still proves the small-batch -> single-file fallback path, and succeeds through bounded chunks.

Local validation after the fix:

- `pnpm --dir packages/ttsc build`
- direct #200 regression runner
- generated dependency cache-miss fallback runner
- generated transform shard fixture runner
- non-empty proof by temporarily changing expected `THIRD` to `WRONG` and confirming the assertion fails, then restoring it
- `pnpm format`
- `git diff --cached --check`

samchon:
Follow-up after watching the replacement CI:

- The 16-file chunk version was still too conservative for typia's generated corpus. It avoided the original huge batch, but left too many native transform process starts and the typia job was already beyond the recent successful `Test typia next` range.
- Increased the bounded chunk size to 64 files. This keeps the protection against oversized generated-directory transforms while reducing repeated native startup overhead.
- Updated the fixture boundary from 20 generated extras / >16 rejection to 70 generated extras / >64 rejection so the test still proves an oversized all-sibling batch would fail.

Additional local validation for this adjustment:

- `pnpm --dir packages/ttsc build`
- direct #200 regression runner
- generated dependency cache-miss fallback runner
- generated transform shard fixture runner
- non-empty proof by temporarily changing expected `THIRD` to `WRONG`, confirming the assertion fails, then restoring it
- `pnpm format`
- `git diff --check` on the touched files

samchon:
Follow-up after the 64-file run also exceeded the normal typia range:

- The long run points to large native transform shards timing out or running too long, then the runtime falling back too coarsely. That can repeat expensive work across generated source imports.
- Reworked generated-source miss handling to adaptively split failed transform/build shards around the requested source instead of jumping straight from a large shard to single-file repetition.
- Lowered the optimistic transform shard timeout to 90s so an oversized native sidecar call splits promptly instead of holding the whole `ttsx` caller for up to ten minutes.
- Set the initial generated sibling shard cap to 32 files. The previous 16 was too chatty; 64 was still too large for typia's generated corpus.
- Strengthened the fixture: it now rejects >32-file batches, rejects >16-file transform batches to force adaptive splitting, and rejects single-file fallback for `extra-0` so the passing result proves the adaptive split path is non-empty.

Local validation:

- `pnpm --dir packages/ttsc build`
- direct #200 regression runner
- generated dependency cache-miss fallback runner
- adaptive generated transform shard fixture runner
- non-empty proof by temporarily changing expected `EXTRA-0` to `WRONG`, confirming the assertion fails, then restoring it
- `pnpm format`
- `git diff --check` on touched files

samchon:
Additional CI-time follow-up:

- The adaptive 32-file shard run still stayed pending past the normal typia window, so I tightened the retry behavior further.
- Added an in-memory failed-shard guard: if a multi-file transform shard fails or times out in a process, later imports for the same shard skip that failed shape and immediately try the narrowed half. This avoids paying the same failed large native transform attempt for both halves of a generated directory.
- Reduced the optimistic generated transform shard timeout from 90s to 30s. These shards are only a fast path; anything that cannot finish quickly should split instead of blocking the runtime caller.

Local validation after this change:

- `pnpm --dir packages/ttsc build`
- direct #200 regression runner
- generated dependency cache-miss fallback runner
- adaptive generated transform shard fixture runner
- `pnpm format`
- `git diff --check` on touched files

samchon:
﻿CI follow-up: the remaining typia-next failure is upstream typia, not this ttsx branch.

Evidence:
- ttsc's generated namespace-import contract still passes locally via `go test ./test/driver -run TestEmitWithPluginTransformerInjectedImport -count=1 -v`.
- The failed typia job emitted namespace imports for `typia/lib/internal/_accessExpressionAsString`, but dynamic-key path snippets still referenced the legacy bare alias `__typia_transform__accessExpressionAsString(key)`, causing `ReferenceError` in `createAssert` / `createAssertGuard` dynamic-key cases before the suite timed out.
- A minimal typia native regression test reproduces that exact shape through `Record<string, number>` with `createAssert`, `createAssertGuard`, and `createValidate`.

Upstream fix opened: https://github.com/samchon/typia/pull/1940

Once that PR is merged into typia `master`/`next`, this PR's typia check should be rerun. I am not adding a ttsc-side alias hack because that would hard-code typia's legacy helper name instead of preserving the plugin/importer contract.

samchon:
﻿Update: upstream typia fix is now green.

- typia PR: https://github.com/samchon/typia/pull/1940
- All typia #1940 checks passed, including `go`, `e2e`, `sdk`, `migrate`, `benchmark`, `transform-options`, tarballs, spell-check, and all Ubuntu jobs.

Next gate for this PR: merge typia #1940 into typia `master`/`next`, then rerun the typia-next check here.

samchon:
Update after CI follow-up:

The long-running typia check exposed a separate ttsx cache-ownership issue from the original interop fix. Runtime generated-source replay was using the per-run dependency emit cache as the source-plugin binary cache, so a large source plugin could be rebuilt during runtime instead of reusing the plugin binaries resolved by the parent ttsx process.

Fix in 87a2cd00:
- carry the entry plugin cache directory into the ttsx runtime manifest;
- reuse pre-resolved native plugin descriptors for generated-source transform replay;
- keep dependency JS emit cache and source-plugin binary cache separate;
- extend the generated-source regression to assert the source plugin is built exactly once when `--cache-dir` is set.

Local validation:
- `pnpm exec prettier --write <touched files>`
- `pnpm run test:typecheck`
- `pnpm --dir packages/ttsc run build` with Git `usr/bin` on PATH for Windows `cp`
- targeted scratch `.mts` runner for `test_plugin_corpus_ttsx_transforms_generated_sources_without_build_fallback`

Note: `pnpm run build:current` rebuilt `packages/ttsc` successfully, then failed later in `packages/unplugin` rollup with `rollup-plugin-node-externals` (`TypeError: undefined is not a function`). That is outside this PR's changed package path and the CI unplugin job is the source of truth for Linux.

samchon:
Additional perf/cache follow-up in eea5a31f:

The typia workflow pointed at a source-plugin cache-key mismatch: Go ignores `_test.go` files for `go build`, but ttsc was still hashing and materializing them as source-plugin cache inputs. typia's native tree now carries many tracked `_test.go` files, so typia is the workload where this mismatch becomes visible while smaller consumers such as nestia stay fast.

Fix:
- exclude `_test.go` from source-plugin cache-key/materialization inputs, aligning ttsc with `go build` semantics;
- add a regression proving `_test.go` edits keep the warm plugin binary, while `main.go` edits still invalidate and rebuild.

Per operator request, further validation is now delegated to GitHub Actions rather than more local test/build runs.

samchon:
Follow-up fix after the timed-out typia control check. The slow path was not a Go binary build-cache miss: master typia completed normally, while this branch started rebuilding missing runtime-generated entry sources through a tsconfig-wide dependency cache. This commit removes that full-project rebuild path and adds a content-keyed generated entry source tree emit cache that reuses the parent ttsx tsconfig/plugins/binary/cache options. Local tests were intentionally not run per maintainer request; CI is the validator for this push.

PR #234, opened 2026-06-18
feat(factory): vendor ts-factory as @ttsc/factory

samchon:
## Intent

Bring the standalone `ts-factory` repository into this workspace as the first-party `@ttsc/factory` package, so the legacy-style TypeScript AST **factory + printer** lives alongside the compiler it is meant to outlive. The factory is hand-written and **zero-dependency** (no `typescript` import), so AST-based code generation keeps working after the tsgo migration.

## Scope

- **`packages/factory`** — the former `packages/ts-factory`, renamed to `@ttsc/factory`:
  - repo/bugs/homepage point at `samchon/ttsc` / `ttsc.dev`; version aligned to the workspace (`0.15.6`).
  - `ttsc` consumed as `workspace:*`; `@typescript/native-preview` moved to `catalog:tsgo`.
  - rollup config inlined (self-contained) instead of re-exporting a shared `config/rollup.config.mjs` that doesn't exist here.
  - in-source `ts-factory` references (doc comments, error message prefixes) rebranded to `@ttsc/factory`.
- **`tests/test-factory`** — the former `test/`, renamed to `@ttsc/test-factory`, adapted to the workspace test convention: runs through `@ttsc/testing`'s `TestExecutor` and the extensionless-TS loader (the same `start` script shape as `@ttsc/test-banner`) rather than the bespoke `DynamicExecutor` + `chalk` runner. Feature files now import `@ttsc/factory`.
- **Workspace wiring** — new `rollup` catalog in `pnpm-workspace.yaml`, regenerated `pnpm-lock.yaml`, a `factory` matrix entry in `test.yml`, and a layout note in the project skill.

## Deferred / out of scope

- **No website documentation.** Intentionally skipped — the factory is advanced/internal and not a priority for the docs site right now.
- **No internal consumers.** Audited the repo for any `typescript`-based `ts.factory` / `ts.Printer` usage that this package should replace — there is **none**. The remaining `typescript@6` dependencies are all for typescript-eslint / Next.js / website tooling, not AST codegen. `@ttsc/factory` ships standalone; nothing else depends on it yet.
- **Repo-wide format drift left untouched.** Running `pnpm format` locally rewrites many unrelated committed files (formatter/line-ending drift between toolchains); only the new files were formatted to keep this PR reviewable. There is no format-gate CI job affected.

## Test plan

- `pnpm exec tsgo --noEmit -p packages/factory/tsconfig.json` — passes.
- `pnpm exec tsgo --noEmit -p tests/test-factory/tsconfig.json` — passes (matches the `test:typecheck` job).
- `pnpm --filter @ttsc/factory build` (tsgo emit + rollup) — produces `lib/` cleanly.
- Feature suite verified via the scratch-runner pattern (the full `DynamicExecutor` suite can't start on this Windows host — environmental, runs on Linux CI through the new `factory` matrix entry).
- `pnpm install --frozen-lockfile` — clean.


samchon:
First CI run surfaced a real failure in the `factory` job:

```
SyntaxError: The requested module '@ttsc/factory' does not provide an export named 'Expression'
```

Root cause: `@ttsc/test-factory` runs through Node's `--experimental-transform-types` (the workspace test-runner convention), which only erases *syntactically explicit* type imports. The upstream `ts-factory` suite ran via the type-aware `ttsx`, so one value import of the type-only `Expression` export slipped through. Fixed in `88d3a4a9` with an inline `type` modifier.

Verified comprehensively before re-push: a scratch loader that imports and runs **all 72 feature tests** through the same Node loader passes locally (the full `DynamicExecutor` suite can't start on Windows — environmental — so this is how it's exercised here). `Expression` was the only type imported as a value across the suite.

PR #249, opened 2026-06-19
feat(shim): expose Checker_getDeclaredTypeOfSymbol (#246)

### LINKED ISSUE #246: Expose Checker_getDeclaredTypeOfSymbol shim for class declared-type (base-chain) resolution
## Summary

Expose a `Checker_getDeclaredTypeOfSymbol` shim (the checker's `getDeclaredTypeOfSymbol`) so consumers can obtain the **declared (instance) type** of a class/interface symbol. Without it, the base-type chain of a class cannot be walked safely once a generic boundary is involved.

## Motivation

`typia.plain.classify` (samchon/typia#1951, shipped in samchon/typia#1956 @ `1dc6f4b9ff`) must detect whether a class carries a `#private` field — directly or through inheritance — to refuse an unsafe field-copy reconstruction (`Object.create(C.prototype)` + assign drops the private brand, so a later `#x in obj` / private method call throws at runtime).

The detection walks the base chain via the existing `Checker_getBaseTypes` shim. That works for a plain class instance type, but **`getBaseTypes` nil-dereferences for any non-instance type**: internally it does `t.AsInterfaceType()`, which returns `nil` for

- a constructor type `typeof C` (Anonymous), and
- a **generic Reference** such as the `Mid<T>` in `class Sub extends Mid<T>`.

So the one case classify cannot cover today is **a `#private` field inherited through a generic base boundary**:

```ts
class Base { #brand = 0; }
class Mid<T> extends Base { value!: T; }
class Sub extends Mid<string> {}   // #brand is inherited, but unreachable
```

`classify<Sub>` cannot see `#brand`, field-copies `Sub`, and crashes at runtime — undetected.

## Why the current shims are insufficient

- `Checker_getBaseTypes` — nil-derefs on the generic `Reference` base (see above); guarding the call to `ObjectFlagsClassOrInterface` only (the safe workaround typia ships) simply stops the walk at the generic boundary.
- `Checker_getTypeOfSymbol` — returns the **constructor** (static) type of a class symbol, not the declared instance type.
- `Type.Target()` / `getDeclaredTypeOfSymbol` — not currently exposed.

## Request

Expose:

```go
// returns the declared (instance) type of a class/interface symbol;
// the result IS a ClassOrInterface type and is therefore safe to pass to Checker_getBaseTypes.
func Checker_getDeclaredTypeOfSymbol(checker *Checker, symbol *Symbol) *Type
```

With it, a consumer can resolve the generic base's symbol → declared instance type → continue the `getBaseTypes` walk to reach `Base.#brand`.

A complementary alternative would be a `getBaseTypes` variant that resolves a `Reference`'s target to its declared type before reading `AsInterfaceType()`, so the existing walk no longer dead-ends at a generic boundary.

## Scope note

This is the single documented open limitation of typia's classify (everything else — non-generic inherited `#private` via the base walk, and a generic class's *own* `#private` via the member scan — already works). The case is exotic (generic subclass + `#private` base + field-copy), so it is filed as an enhancement rather than a blocker.

samchon:
## Intent

Resolves #246. Exposes `Checker_getDeclaredTypeOfSymbol` so a consumer can obtain the **declared (instance) type** of a class/interface symbol, and adds a mechanical compile-time guard so this class of shim gap is caught by the build rather than by another reactive issue.

## The fix

`Checker_getBaseTypes` nil-derefs on any non-instance type — `t.AsInterfaceType()` returns `nil` for a constructor `typeof C` (Anonymous) and for a generic `Reference` such as the `Mid<T>` in `class Sub extends Mid<string>`. So a base-chain walk dead-ends at a generic boundary, and a `#private` field inherited through it (`class Base { #brand = 0 } ; class Mid<T> extends Base {} ; class Sub extends Mid<string> {}`) stays invisible to typia's `plain.classify`, which then field-copies `Sub` and crashes at runtime.

`getDeclaredTypeOfSymbol(symbol)` resolves a base's name symbol back to a `ClassOrInterface` type that **is** safe to feed to `getBaseTypes`, so the walk continues past the generic boundary. Re-exported via `go:linkname` (the method is unexported), following the existing `Checker_getBaseTypes` / `Checker_getMinArgumentCount` pattern in `shim/checker/shim.go`.

## Why these gaps keep recurring — and the mechanical net

The shim's closure auditor (`tools/shim_audit`) catches partial enums, reachable funcs, and escaping types, but **by construction cannot fail on this class**: `getDeclaredTypeOfSymbol` is an *unexported* `*Checker` method, demanded by a downstream consumer by name. The auditor lists such methods only in a *non-failing* triage pool — it cannot know which one a consumer needs next. That is the exact blind spot behind #246 (and the earlier `Signature` gap in 0.15.5).

The mechanical net that *does* fit this class is a **compile-time composition guard**: `declared_type_api_completeness.go` links the full consumer walk — `getBaseTypes → getTypeNameSymbol → getDeclaredTypeOfSymbol → getBaseTypes`. It is ordinary (non-`_test`) package code, so every build that links the shim (typia's native plugin; ttsc's lint engine via `pnpm test:go`) type-checks it. An upstream typescript-go rename, a signature change, or a dropped leg now turns those builds red mechanically, instead of surfacing as a runtime crash in a consumer months later. This mirrors the existing `signature_api_completeness.go` guard.

## Scope

Two files in `packages/ttsc/shim/checker/`:
- `shim.go` — the `Checker_getDeclaredTypeOfSymbol` linkname + wrapper.
- `declared_type_api_completeness.go` — the compile-time guard (new).

## Test plan

- `shim/checker` builds clean against the pinned typescript-go (verified via a workspace `go build`).
- `pnpm --filter ttsc shim:audit -check` → OK (no new gaps; baseline unchanged).
- `pnpm test:go` (lint engine) links the checker shim including the new guard and compiles it. The two `LoadRuleConfig*` failures in that suite are a pre-existing, unrelated `ttsx` config-loading issue (TS6059 `loader.mts` rootDir) — confirmed identical on clean `master`.


samchon:
## Update: the completeness guard is now a real runtime detector

Replaced the compile-time guard with a **traversal-completeness probe** (`packages/lint/test/shim/base_chain_walk_crosses_generic_boundary_test.go`), addressing the core ask: *ttsc must mechanically find shim gaps from its own side, without leaning on a consumer.*

A compile guard only proves the composition **links**. It cannot see a semantic **dead-end** — which is exactly what the missing shim caused. The probe instead runs a real Checker over a ttsc-owned fixture and walks the base chain through only the exposed surface:

- **fixture** — `Base { #brand } <- Mid<T> <- Sub extends Mid<string>`
- **naive walk** (ClassOrInterface bases only, the pre-#246 safe workaround) — asserts it dead-ends at the generic boundary and **never reaches `Base`** → the gap is real, mechanically demonstrated.
- **bridged walk** (resolves the generic `Reference` base via `getDeclaredTypeOfSymbol`) — asserts it **reaches `Base`** and does not over-reach.

Remove `Checker_getDeclaredTypeOfSymbol` and the bridged assertion fails — so it is a genuine detector, not a tautology. It runs in `pnpm test:go`; a dead-end in any exposed traversal now surfaces as a red CI from ttsc's side, instead of as a downstream consumer issue.

Verified: `--- PASS: TestBaseChainWalkCrossesGenericBoundary` against the pinned typescript-go.

This is the generalizable template — every exposed graph-walk operation gets a fixture + completeness assertion, so the dead-end class cannot silently return.

PR #250, opened 2026-06-19
feat(ttsc): expose context.dirname/filename for ESM-safe plugin self-location

### LINKED ISSUE #248: ttsc 0.16 runs plugin descriptor modules without CJS globals (`__filename`/`require`), silently mis-resolving self-located paths
## Summary

`ttsc@0.16.0` executes a plugin **descriptor module** (the `plugin.transform` entry) in a context **without CommonJS globals** — `__filename` is unusable and the ambient `require` is undefined (consistent with ESM / `import()` loading). Descriptors that locate their own package via `__filename` or `require.resolve(...)` therefore compute the wrong path and the build fails with `plugin "<name>" source does not exist`.

This may be intended (ESM-first loading). If so, the ask is to (a) document that descriptors must be ESM-safe and (b) expose the plugin's resolved module path in the factory context so descriptors can self-locate without ambient globals.

## Environment

- `ttsc@0.16.0` (binary `@ttsc/win32-x64@0.16.0`)
- `pnpm@10.6.4` workspace, node `v22.21.0`, `win32-x64`
- A plugin package (`typia`) whose `package.json` declares `"ttsc": { "plugin": { "transform": "typia/lib/transform" } }`. The descriptor module computes its package root to point ttsc at the native source dir.

## Observed

Instrumenting the descriptor's default export `createTtscPlugin(context)` printed:

```
context = {
  "binary":  ".../@ttsc/win32-x64/bin/ttsc.exe",
  "cwd":         "D:/repo/tests/test-typia-generate",
  "projectRoot": "D:/repo/tests/test-typia-generate",
  "tsconfig":    "D:/repo/tests/test-typia-generate/tsconfig.json",
  "plugin": { "transform": "typia/lib/transform" }
}

// inside the descriptor, evaluated at call time:
__filename                                            ->  ""          (unusable; typeof !== usable string)
require.resolve("typia/package.json",{paths:[projectRoot]})  ->  throws  (ambient `require` undefined)
```

`context.projectRoot` is **correct** (`tests/test-typia-generate`). The same `require.resolve("typia/package.json", { paths: [projectRoot] })` resolves fine from a plain Node process (`-> .../packages/typia/package.json`), so the failure is specific to the descriptor's execution context, not the workspace layout.

With both self-location strategies failing, the descriptor fell back to `path.resolve(path.dirname(__filename), "..")` which, with an empty `__filename`, resolved against `cwd` and produced the **parent** of the consuming project:

```
ttsc: plugin "typia" source does not exist: D:/repo/tests/native/cmd/ttsc-typia
                                            (expected D:/repo/packages/typia/native/cmd/ttsc-typia)
```

## Why it breaks

The descriptor was authored for CJS:

```ts
// works under a CJS loader (ttsc 0.15), breaks under 0.16
const dirname = path.dirname(__filename);                       // __filename === "" now
require.resolve("typia/package.json", { paths: [projectRoot] }); // `require` undefined now
```

Under ttsc 0.15 these descriptors loaded with CJS globals available. Under 0.16 they don't, so any descriptor using `__filename` / ambient `require` to locate its own package silently mis-resolves.

## Workaround (consumer side)

Anchor a CJS resolver on `context.projectRoot` via `node:module`, which works under both CJS and ESM:

```ts
import { createRequire } from "node:module";

const requireFrom = createRequire(path.join(context.projectRoot, "package.json"));
const root = path.dirname(requireFrom.resolve("typia/package.json"));
```

This fixed it for us. Filing because the breakage is silent (wrong path, not a clear "your descriptor is CJS" error) and likely affects other CJS-authored descriptors upgrading 0.15 → 0.16.

## Suggestions (any one)

1. Inject `__filename`/`__dirname`/`require` into descriptor modules (CJS-style execution), restoring 0.15 behavior.
2. Document that descriptors must be ESM-safe (no `__filename`/ambient `require`; use `createRequire` or `import.meta.url`), and emit a clear diagnostic when a descriptor uses them.
3. Expose the plugin's **resolved module path** (or package root) on the factory context (e.g. `context.plugin.resolved`) so a descriptor can self-locate without any ambient globals.

samchon:
## Intent

Fixes #248. A plugin **descriptor** loaded from `.ts` source runs through ttsx's `import()` shim, where the ambient `__dirname`/`__filename`/`require` are all undefined — verified empirically that even a `.cts` loses them, so it is the **loader path, not the module format**. A descriptor that self-locates through those globals silently mis-resolves its `source` and the build fails with `plugin source does not exist` (the consumer project's *parent*, from `path.dirname("")`). `typia` hits this because `typia/lib/transform` resolves to `src/transform.ts` in its monorepo.

## Scope

- **`ITtscPluginFactoryContext`** gains `dirname` and `filename` — the ESM-safe analogs of `__dirname`/`__filename`, mirroring Node's `import.meta.dirname`/`import.meta.filename`. Populated in **both** load paths (direct `require` and the ttsx shim's `TTSC_PLUGIN_CONTEXT`), so a descriptor self-locates with no ambient global in any loader.
- **Additive, non-breaking.** Existing descriptors (compiled CJS using `__dirname`, or `createRequire(projectRoot)` workarounds) are untouched.
- **Docs** — the Plugin Protocol Reference and the banner walkthrough now state descriptors must be ESM-safe and recommend `path.resolve(context.dirname, …)` over `__dirname`.

## Deferred / known items

- **`export default` via the ttsx orphan-CJS path** — when a `.ts` descriptor falls to the orphan single-file CJS lowering, the CJS→ESM interop makes `mod.default` a `{ default: fn }` wrapper, so the shim's candidate resolution grabs the object instead of the function. Pre-existing and orthogonal to this PR; named exports (`export function createTtscPlugin`) avoid it. Flagged for a separate fix.
- **`filename`** is included for Node parity; can be dropped if a minimal surface is preferred (`dirname` is the load-bearing field).

## Test plan

- New regression test `test_plugin_corpus_descriptor_factory_self_locates_via_context_dirname` — drives the **real** `ttsc` binary against a `.ts` descriptor (typia's exact ttsx path) that derives `source` purely from `context.dirname`. Verified **fail-before / pass-after** by neutralizing the field in the built lib (reproduces the `#248` mis-resolution).
- `pnpm --filter ttsc build` clean (tsc typecheck of the `Omit<…, "dirname" | "filename">` loader change).
- Full TS e2e suite runs on CI (DynamicExecutor can't start on the Windows dev box).

### DIFF
--- a/packages/ttsc/src/plugin/internal/loadProjectPlugins.ts
+++ b/packages/ttsc/src/plugin/internal/loadProjectPlugins.ts
@@ -570,7 +570,7 @@ function orderNativePlugins(
 
 function loadPluginEntry(
   entry: ITtscProjectPluginConfig,
-  context: ITtscPluginFactoryContext,
+  context: Omit<ITtscPluginFactoryContext, "dirname" | "filename">,
   baseDir: string,
 ): ITtscPlugin {
   return withPluginLoaderEnv(() => {
@@ -582,7 +582,15 @@ function loadPluginEntry(
     }
 
     const request = resolvePluginRequest(specifier, baseDir);
-    const mod = requirePluginEntry(request, context) as {
+    // The descriptor loads through ttsx (`.ts` source) or as ESM (`.mjs`), where
+    // the ambient `__dirname`/`__filename`/`require` are undefined. Hand the
+    // resolved entry's dir/file to the factory so it can self-locate regardless.
+    const factoryContext: ITtscPluginFactoryContext = {
+      ...context,
+      dirname: path.dirname(request),
+      filename: request,
+    };
+    const mod = requirePluginEntry(request, factoryContext) as {
       createTtscPlugin?: TtscPluginFactory;
       default?: ITtscPlugin | TtscPluginFactory;
     } & Partial<Record<"plugin", ITtscPlugin | TtscPluginFactory>>;
@@ -592,7 +600,7 @@ function loadPluginEntry(
       mod.plugin ??
       (mod as unknown as ITtscPlugin | TtscPluginFactory);
     if (typeof candidate === "function") {
-      const plugin = candidate(context);
+      const plugin = candidate(factoryContext);
       if (!isTtscPlugin(plugin)) {
         throw new Error(
           `ttsc: plugin "${specifier}" does not export a valid ttsc plugin`,
@@ -702,6 +710,8 @@ function loadDescriptorViaTtsx(
         TTSC_PLUGIN_CONTEXT: JSON.stringify({
           binary: context.binary,
           cwd: context.cwd,
+          dirname: context.dirname,
+          filename: context.filename,
           plugin: context.plugin,
           projectRoot: context.projectRoot,
           tsconfig: context.tsconfig,
--- a/packages/ttsc/src/structures/ITtscPluginFactoryContext.ts
+++ b/packages/ttsc/src/structures/ITtscPluginFactoryContext.ts
@@ -33,6 +33,28 @@ export interface ITtscPluginFactoryContext<T = ITtscProjectPluginConfig> {
   cwd: string;
 
   /**
+   * Absolute path to the directory holding the resolved descriptor module.
+   *
+   * The ESM-safe analog of `__dirname`, mirroring Node's `import.meta.dirname`.
+   * Use it to locate the plugin's own package without ambient CommonJS globals:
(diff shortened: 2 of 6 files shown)


samchon:
Closing — `context.projectRoot` already covers descriptor self-location: `createRequire(path.join(context.projectRoot, "package.json")).resolve("<pkg>/package.json")` resolves the plugin's own package in both CJS and ESM (typia's existing workaround). The dedicated `context.dirname`/`context.filename` fields are redundant for that, so withdrawing this change.

PR #253, opened 2026-06-19
fix(unplugin): keep the per-build cache key universe consistent (#252)

### LINKED ISSUE #252: @ttsc/unplugin: per-build transform cache never hits on real projects (O(N) full recompiles) + O(N²) re-hashing
## Summary

Using `@ttsc/unplugin/rollup` on a real project (reproduced against `typia`'s `packages/typia`) is **excessively slow and effectively never finishes a build**. Root-cause analysis points at the per-build transform cache in `packages/unplugin/src/core/transform.ts`. There are two compounding defects; the second is the build-breaker.

The adapter registers a `transform(source, id)` hook that compiles the **entire tsconfig project once** and serves each module from a per-build cache (`createTtscTransformCache`, cleared on `buildStart`). The intent is "compile once, serve N modules from cache". Both the cache *validation* and the cache *key set* are broken.

## Bug A — `O(N²)` redundant re-hashing on every module

On every cache hit, `transformTtsc` calls `matchesCachedSource` → `collectProjectInputHashes`, which **walks the whole project tree and reads + SHA-256s every input file** (`transform.ts` `collectProjectInputHashes`/`listProjectInputFiles`). For a project with `N` modules the bundler calls `transform` `N` times, so this is `O(N²)` filesystem reads + hashing per build.

This work is **redundant within a single build**: `buildStart` already clears the cache (`core/index.ts` `buildStart`), so the project is assumed stable for the build. The whole-project re-hash only ever matters across `buildStart` cycles (watch), yet it runs on every module of the initial build.

## Bug B — store/validate hash key-set divergence → perpetual cache miss (the build-breaker)

The stored hash snapshot and the validation snapshot are built from **different key sets**, so `sameHashes` is essentially always `false` and the cache **never hits** — the entire project is re-transformed once per module (one native compile per file).

- **Store** (`collectInputHashes`): starts from `collectProjectInputHashes`, then **overlays a hash for every key in `result.typescript`**. Those keys include program files **outside the project's own directory walk** — `node_modules/**/*.d.ts`, sibling-package / out-of-root sources. The native host keys files by `filepath.Rel(cwd, fileName)` (or absolute when outside root) and returns *every* non-library program source (`cmd/ttsc/api_transform.go`).
- **Validate** (`matchesCachedSource` → `collectProjectInputHashes`): **skips `node_modules`** and every ignored dir (`isIgnoredProjectDirectory`) and never reproduces those out-of-walk keys.

→ key sets differ → length check in `sameHashes` fails → cache miss for every module → `pluginRuns == N` full-project transforms.

Any real project that imports a typed dependency (so the program includes `node_modules` `.d.ts`) or lives in a monorepo (out-of-root sources) hits this. The existing unit fixture (`TestUnpluginProject`) never does — its project is a single self-contained `src/main.ts` with no imports, so `result.typescript` only ever contains `src/main.ts`, the key sets match, and tests pass while real projects break.

## Reproduction & measurements

Added `experimental/unplugin-perf` — it drives the **real** `unplugin.rollup(...)` object over a synthetic `N`-file project (adapter bundled on the fly with esbuild, `ttsc`/`unplugin` external) and counts native plugin spawns + `fs.readFileSync` traffic per simulated build.

`pnpm --dir experimental/unplugin-perf start` (Windows, Node 22):

```
Scenario A — output keys under project root (cache hits):
  N= 10  pluginRuns=  1  reads=   4606   2196ms
  N= 25  pluginRuns=  1  reads=   5251   2290ms
  N= 50  pluginRuns=  1  reads=   7326   2686ms
  N=100  pluginRuns=  1  reads=  15226   4668ms      # reads grow ~O(N^2) via re-hash

Scenario B — one output key outside the validator walk (node_modules):
  N= 10  pluginRuns= 10  reads=  44684   20241ms
  N= 25  pluginRuns= 25  reads= 112844   55809ms
  N= 50  pluginRuns= 50  reads= 229444  106938ms     # ~40x slower than A, full recompile per module
```

Scenario B is the realistic case: **107 seconds for a trivial 50-file project**, scaling with one full-project native compile per module. A real, heavy program like `packages/typia` (125+ source files, large dependency graph) degrades to minutes and reads as "doesn't build".

## Suggested direction

- **Bug B**: make the store and validate snapshots use the **same key universe**. Either restrict `collectInputHashes`'s `result.typescript` overlay to keys that fall inside the validator's walk (drop `node_modules` / out-of-root keys), or hash a single stable project fingerprint both sides agree on.
- **Bug A**: don't re-hash the whole project on every cache hit within a build. Compute the project fingerprint at most once per `buildStart` cycle (the cache is already per-build), or invalidate via watch events rather than per-module full-tree hashing.
- Add a multi-file fixture (imports from `node_modules`, >1 emitted key) so both regressions are covered — the current fixture is too minimal to exercise either path.

## Secondary notes

- `selectTransformedSource` throws on any project-wide `failure`, so a single type error anywhere in the project aborts every module's transform.
- The adapter emits transformed **TypeScript**; Rollup needs a downstream TS→JS step (`@rollup/plugin-typescript`, esbuild, swc) or it cannot parse real-world type syntax — worth documenting in the rollup guide.

samchon:
## Intent

Fixes #252. Using `@ttsc/unplugin` on a real project (reproduced against `typia`'s `packages/typia` via `@ttsc/unplugin/rollup`) was excessively slow and effectively never finished a build. Root cause: the per-build transform cache **never hit** on any project that imports a typed dependency, so the whole tsconfig project was re-transformed **once per module**.

## Root cause

The adapter compiles the whole project once (`createTtscTransformCache`, cleared on `buildStart`) and serves each module from cache. The cache hit was gated by comparing a stored hash snapshot against a freshly re-hashed one — but the two were built from **different key universes**:

- **store** (`collectInputHashes`) overlaid a hash for every key in `result.typescript`, which includes out-of-walk program inputs the native host returns (`node_modules/**/*.d.ts`, sibling-package sources keyed by `filepath.Rel(cwd, …)` / absolute).
- **validate** (`matchesCachedSource` → `collectProjectInputHashes`) skips `node_modules` and every ignored dir, so it never reproduced those keys.

→ snapshots never matched → cache miss on every module → `pluginRuns == N` full-project transforms.

The existing unit fixture never caught it: a single self-contained `src/main.ts` with no imports only ever produces `{ "src/main.ts": … }`, so the key sets happened to match.

## Fix

Drop the `result.typescript` overlay in `collectInputHashes`, so both the snapshot and the validator draw their keys from the same `collectProjectInputHashes` walk. Whole-project content validation is **kept**, so a sibling source or plugin-declared dependency change still invalidates the cache — the existing `…InvalidatesOnProjectSourceChange` / `…OnLibSourceChange` / `…OnSourceChange` specs continue to hold.

## Measurements

`experimental/unplugin-perf` drives the real `unplugin.rollup(...)` object over a synthetic N-file project and counts native plugin spawns (`pnpm run experimental:unplugin-perf`). Scenario B emits one out-of-walk (`node_modules/**`) output key, exactly like the native host:

| | before (N=50, Scenario B) | after (N=50, Scenario B) |
|---|---|---|
| plugin runs | **50** (recompile per module) | **1** |
| `fs.readFileSync` | 229,444 | 4,532 |
| wall time | **106,938 ms** | **1,816 ms** (~59× faster) |

## Scope

- `packages/unplugin/src/core/transform.ts` — the fix.
- `tests/test-unplugin/src/features/transform/` + `internal/transform-project-cache.ts` — a multi-file counting fixture and two feature tests (one pins the single compile across N modules, one pins the out-of-walk regression).
- `experimental/unplugin-perf/` — runnable regression harness; exits non-zero if any build re-transforms more than once.
- `website/src/content/docs/ttsc/bundler.mdx` — note that Rollup/Rolldown need a downstream TypeScript→JavaScript step (the adapter emits TypeScript).

## Deferred

- **Per-call validation cost.** Validating a cache hit still re-hashes the project (bounded; small files), because the current invalidation contract requires detecting a sibling change between two `transformTtsc` calls on a shared cache. Making this O(1) would mean relying solely on `buildStart` invalidation and relaxing those isolation tests — a deliberate follow-up, tracked in #252.
- `selectTransformedSource` throws on any project-wide `failure` (one type error anywhere aborts every module). Unchanged here.

## Test plan

- [x] `experimental/unplugin-perf` exits 0 after the fix (`pluginRuns == 1` in both scenarios); exits 1 on the pre-fix code.
- [x] `tsc --noEmit` on `packages/unplugin` is clean.
- [ ] CI: `test:features` runs the two new feature tests (the built-`lib` path could not be exercised locally — the `@ttsc/unplugin` rollup build fails in this environment on an unrelated `rollup-plugin-node-externals` load error).


samchon:
### Review round 4 (adversarial) — Turbopack cache invalidation, analyzed & resolved

A deep correctness pass raised whether dropping the `result.typescript` overlay regresses the **Turbopack** adapter, which uses a process-lifetime cache with no `buildStart`/`addWatchFile` and relies entirely on `transformTtsc`'s hash re-walk for freshness.

**Verdict: not a real regression for any actual ttsc plugin.** Verified against the code:

- A transform plugin's `result.typescript` keys are its *outputs* — always in-walk `src/` files (typia, banner, paths, strip all transform in place). Out-of-walk keys in `result.typescript` only appear on the **native-host passthrough** path (no transform plugin), where the output equals the source unchanged, so a "stale" cache hit returns the *same* bytes. Output is identical before/after this change.
- Out-of-walk inputs a plugin *reads* travel via `result.dependencies` → `addWatchFile`, which Turbopack never received — **before or after** this change. That gap is pre-existing and orthogonal; the suggested `addDependency` wiring wouldn't fix it either, since the cache would still hit without a `buildStart` clear.

**Action taken:** corrected the `matchesCachedSource` docstring, which overstated that the hash comparison invalidates on "a plugin-declared dependency" (true only for in-walk inputs). Out-of-walk invalidation is the adapter's `addWatchFile`/`buildStart` responsibility, now stated explicitly. Turbopack's standalone dependency registration is a separate, pre-existing improvement left for a focused follow-up.

The test-quality adversarial pass returned no findings (false-green, parallelism, determinism, cross-platform, hygiene, Go correctness all verified).

Issue #255, opened 2026-06-20
@ttsc/metro: React Native / Expo support via a Metro transformer


samchon:
> **Revised** after reading the reference implementation's source (`@elliots/metro-transformer-typical` v0.3.1) and ttsc's own per-file transform path. The first draft's methodology ("reuse the `@ttsc/unplugin` per-file compiler singleton") was **wrong**: the unplugin has no incremental per-file compiler — it recompiles the whole tsconfig project on each transform and caches the result (the model behind #252). The corrected design is below.

## Summary

React Native / Expo bundle with **Metro**, not a web bundler. Metro transpiles each file with Babel (`@expo/metro-config/babel-transformer` / `@react-native/metro-babel-transformer`), which strips types and never runs TypeScript transformers — so `ttsc` plugins (typia, nestia, …) are inert in RN/Expo today. Neither the `ttsc` CLI nor `@ttsc/unplugin` can reach the Metro pipeline.

The fix is a dedicated `@ttsc/metro` package exposing `withTtsc(getDefaultConfig(__dirname))`, mirroring [`@elliots/metro-transformer-typical`](https://github.com/elliots/typical/tree/main/packages/metro-transformer). `typical` is architecturally a twin of ttsc (a typescript-go transformer host that needs the type checker), so its Metro package is a direct, working blueprint rather than a guess.

## How the reference actually works (`@elliots/metro-transformer-typical` v0.3.1)

Source: `github.com/elliots/typical` → `packages/metro-transformer/src` (`transformer.ts`, `index.ts`, `core/transform.ts`, `core/upstream.ts`). Four small files; the mechanism is precise.

**1. Wiring** — `withTypical` only overrides `babelTransformerPath` (keeps Metro's transform worker, swaps the Babel step):

```ts
export function withTypical(config, options) {
  configure(options); // pre-seeds module-level config
  return { ...config, transformer: {
    ...config.transformer,
    babelTransformerPath: join(__dirname, "transformer.cjs"),
  }};
}
```

**2. Transform contract** — Metro's babel-transformer contract is `transform({ src, filename, options }) => { ast }`. typical runs *itself first*, then hands the transformed **source** to the upstream RN/Expo babel transformer to produce the AST:

```ts
export async function transform({ src, filename, options }) {
  // (.ts/.tsx/.mts/.cts + include/exclude filtering elided; non-matches pass straight to upstream)
  const result = await transformFile(filename, config);     // typical -> TS with validators injected
[...]

PR #257, opened 2026-06-21
feat(ttsc): resident/incremental compiler host (#255)

samchon:
## Slice 1 of the resident / incremental compiler host (#255)

Proves the foundation primitive before anything is built on it: typescript-go's `compiler.Program.UpdateProgram` (single-file structural reuse) is reachable through ttsc's existing shim, and reuses the loaded program on a body-only edit while preserving the single-checker pin.

### What this adds

`packages/ttsc/driver/program_update_test.go`: loads a two-file project (`a` imports `b`), rewrites `b`'s function body on disk (signature and imports unchanged), calls `prog.TSProgram.UpdateProgram(changedPath, newHost, nil)`, and asserts `reused == true` and the updated program reflects the new body.

### Why

`@ttsc/metro` (and `@ttsc/unplugin`, #252) need to update a loaded program per file instead of recompiling the whole project. `UpdateProgram` re-parses only the changed file and reuses every unchanged file's AST. Because `shim/compiler` exposes `Program` as a type alias, `UpdateProgram` is already callable with no new re-export; this test pins that and the single-checker invariant. See the design in #255.

Next slices: resident driver session (overlay VFS + `transformFile`), shared-host `serve` mode, JS `TtscService`, then wire `@ttsc/metro` and `@ttsc/unplugin`.

Refs #255.


samchon:
Slice 2 pushed: `LoadProgramOptions.FS` override, `overlayFS` (in-memory overlay over an inner FS), and a resident `Session` (`Apply` → `UpdateProgram`, `SourceText` reads the live program), with a test proving an overlay edit reuses the program and reflects new content. Slice-1 test reformatted to 2-space Go.

PR #276, opened 2026-06-25
feat(graph): redesign @ttsc/graph as a checker-resolved architecture graph MCP (#275)

samchon:
Redesigns `@ttsc/graph` per #275 into a checker-resolved **architecture** graph
MCP, and pivots the entire agent-facing surface to **pure TypeScript** (typia)
while Go keeps only what needs the compiler: building the checker-resolved fact
graph. Goal: ~10× fewer tokens to answer structure/flow/impact questions.

> **Status: in progress (draft).** Core data contract + engine + the four tools
> are done and verified end-to-end. Framework passes, benchmark, docs, and
> removal of the old Go MCP are still to come — see Deferred. CI red on
> `tests/test-graph` is expected until those tests are rewritten for the new tools.

## Architecture

```
Go (packages/ttsc)                         TS (@ttsc/graph)
  internal/graph  (tsgo checker build)      MCP server (@modelcontextprotocol/sdk)
  cmd/ttscgraph dump  -> IGraphDump JSON ->  GraphModel (resident, synthesizes
                                              file/contains/exports + property)
                                             GraphController via typia.llm.controller
                                               graph_overview / _query / _trace / _expand
```

Go emits faithful checker facts only; **all** ranking, projection, framework
synthesis, token budgeting, and the MCP tool schemas live in TS, where
`typia.llm.controller` generates each tool's JSON schema + validator from the
method's TypeScript types and JSDoc — no hand-written schema.

## What's done

- **Schema** (`packages/graph/src/schema`): canonical `IGraphDump` wire contract
  (node/edge kinds, provenance/confidence, evidence, decorator facts).
- **Go dump**: `dump.go` emits the full `IGraphDump` (schemaVersion 2) — kind
  mapping (calls/instantiates/renders, extends/implements), per-edge
  provenance/confidence, line/col evidence, `Exported` via the checker export
  table, decorator facts on nodes, project-relative + deterministic. All
  additive: the existing Go model and resident MCP are untouched.
- **TS engine**: `GraphModel` (indexes + structural synthesis), `loadGraph`
  (spawn dump -> model), and the four tools — `graph_overview` (layers/hotspots/
  publicApi), `graph_query` (subword/centrality ranking with diversity cap),
  `graph_trace` (forward/reverse/impact), `graph_expand` (source + neighbors).
- **Server**: built in-repo with workspace ttsc + the typia plugin; inlined
  standalone registration keeps deps to typia + the MCP SDK.

## Deferred (remaining work on this branch)

- Framework semantic pass: Nest routes from decorator facts, then React/Next.
- Opt-in heuristic callback/event bridge (marked, excluded from default trace).
- Rewrite `tests/test-graph` for the new tools; committed e2e coverage.
- Benchmark manifest + grading + quality scoring; website docs rewrite.
- Remove the old Go MCP (`internal/graph/mcp`) and shrink `ttscgraph` to dump.

## Test plan

- Go: `internal/graph` suite + `dump` tests pass locally; `graphdump --pretty`
  eyeballed on a controller fixture.
- TS: whole-package `tsc` typecheck clean; per-tool logic smokes; full `ttsc`
  build (typia transform runs on the server); **server runtime smoke** — lists
  the four tools with typia-generated schemas (enum + JSDoc), answers a
  `graph_overview` call, and rejects invalid args via typia validation.
- CI to run the full Go + feature suites.


samchon:
# 📊 `@ttsc/graph` benchmark — token consumption

Live A/B of the redesigned graph MCP: the same tool-neutral prompt runs through the agent CLI once with the `@ttsc/graph` MCP attached, once with an empty MCP config (`--strict-mcp-config`). Tokens are summed over every assistant turn; every action (including the model's reasoning) is logged and the answer is graded against a deterministic gold so a token saving never counts while the answer is wrong.

## First-pass findings (sonnet, n=2, ad-hoc prompts)

| prompt | baseline tok | graph tok | token Δ | quality |
|---|---:|---:|:---:|:---:|
| typeorm · overview | 1,110,797 | **404,786** | **−64%** | **2/2 ✓** |
| typeorm · relation→find→joins | 1,563,971 | 775,858 | −50% | 0/2¹ |
| nestjs · request-lifecycle | 684,336 | 697,752 | ≈0% | 1/2 |
| excalidraw · render | 789,242 | 968,821 | **−23%²** | 1/2 (baseline 0/2³) |

The graph wins decisively where the no-MCP baseline must read large files (typeorm: 4k-line classes), and **loses on small repos** (excalidraw) — three causes, all diagnosed from the logged reasoning:

1. **Scale.** The graph arm's cost is ~bounded (a few graph calls); the baseline scales with how much source it reads. Small repo → cheap baseline → little to reclaim.
2. **Supplement, not replacement.** The model explores by `Glob`+`Read` *first*, then uses the graph on top — so graph calls *add* tokens instead of replacing reads. ²This is why excalidraw went negative.
3. ³**Mis-calibrated golds.** Some golds (e.g. excalidraw must-mention `roughjs`) the thorough baseline itself can't pass — over-specified, a measurement bug, not a graph failure.

## What's changing (universal reforms, no per-project tuning)

- **Graph-first instructions.** Rewritten to lead with a hard rule: the project is already fully indexed, so make a graph call the *first* action and read source only via `graph_expand` — countering the read-first instinct that sank excalidraw.
- **Fixed-baseline loop.** The no-MCP baseline doesn't move as the engine improves, so it's measured **once at n=5** per project and cached as a constant; each iteration then runs only the graph arm at **n=1, concurrently across every project** — cheap, fast, stable.
- **Gold calibration.** A gold the n=5 baseline can't pass is relaxed before its quality is trusted.
- **codegraph-comparable prompts.** Every project now has a dedicated one-sentence "How does X do Y?" mechanism question (verbatim from codegraph for excalidraw/vscode) plus one shared common prompt.

## Target (mini = `gpt-5.4-mini`, the priority model)

The graph's cost is bounded while the baseline scales, so the achievable target is **≥70% median token reduction** across projects (80%+ on the big libraries), **0 source detours**, and **100% of the calibrated golds passing**. Conquest = closing the gap to that, project by project.

_Mini fixed-baseline (n=5) is measuring now; graph-arm results land here next._

<sub>Harness `agent-ab*.mjs` · suite `run-suite.mjs` · analysis `analyze-traces.mjs` · grading `grade.mjs`. Token = Σ assistant-turn usage (codegraph's metric).</sub>

---

## Mini (`gpt-5.4-mini`) — iteration 1, fixed baseline (n=5) vs graph (n=1)

| project | baseline tok | graph tok | Δ | graph calls |
|---|---:|---:|:---:|:---:|
| zod | 817,623 | 250,663 | +69% | **0** ⚠️ |
| excalidraw | 1,371,736 | 582,852 | **+58%** | 4 |
| nestjs | 534,318 | 490,350 | +12% | 9 |
| vscode | 713,103 | 662,242 | +11% | 6 |
| typeorm | 776,650 | 741,741 | +4% | **0** ⚠️ |
| rxjs | 431,604 | 498,013 | −15% | **0** ⚠️ |
| vue | 296,899 | 375,460 | −26% | 5 |
| shopping-backend | 440,192 | 618,432 | −40% | 10 |

**Median +5.5% — but the signal is buried under two effects, both now diagnosed:**

1. **The model under-adopts the graph.** On 3 of 8 prompts (zod, typeorm, rxjs) mini made **zero graph calls** — it explored with `rg`/`cat` and never queried the MCP, despite graph-first instructions. The server is not at fault: it loads typeorm in **1.5 s** and serves all four tools. So those rows measure shell-vs-shell noise, not the graph. At n=1 that noise dominates the median.
2. **The win scales with baseline cost.** Where the graph *is* used and the no-MCP baseline is expensive, it wins big (excalidraw 1.37M→583K, **+58%**); where the baseline is already cheap (vue 297K), even a used graph can't beat it.

**Next:** raise the per-iteration n past the adoption noise, and push mini's graph adoption (the open lever is whether the codex client even delivers the MCP server's instructions to the model — if not, the graph-first guidance must move into the tool descriptions). The big-repo wins (excalidraw, the typeorm/vscode trends) are the real, reproducible signal.

### Iteration 2 — tool descriptions made graph-first/anti-shell

Moving the "use this instead of grep/cat" push from the (codex-invisible) server instructions into the tool descriptions raised adoption sharply — mini now queries the graph where it didn't: typeorm 0→16 graph calls, zod 0→4, excalidraw 4→14, vscode 6→10.

**Median token reduction 5.5% → 12.5%**, with clean flips to wins: **nestjs +46%** (7 graph calls, **0 shell** — pure graph), vue −26%→+20%.

But it surfaced the next blocker precisely: **mini uses the graph *on top of* shell, not instead of it.** Where it reaches graph-mostly (nestjs: 0 shell) it wins big; where it keeps grepping after querying (typeorm: 16 graph + 14 shell) the extra calls push tokens *up*. So the lever is no longer adoption — it's making the graph *replace* shell, not supplement it. (n=1 across 8 projects is the sample; individual rows like rxjs swing on shell-thrash variance, the median is the signal.)

### Iteration 3 + a methodology finding: n=1 is below the noise floor for small tweaks

Iteration 3 added one line to graph_expand ("the source is authoritative — don't re-grep to confirm it"), targeting the verify-grep burst the typeorm trace showed (graph used correctly, *then* 14 greps to re-read what it had expanded).

Per-project moved both ways — nestjs +46%→**+65%**, vue +20%→+36%, zod +18%→+40%, but excalidraw +18%→−2% and vscode +7%→−21% — and the **median swung +12.5% → −11%**. That 23-point swing from a one-line change is **n=1 noise, not the tweak**: the same engine gives wildly different per-run medians.

**Conclusions for the loop:**
- **Structural changes are measurable at n=1** (iter1→iter2 graph adoption jumped from 0→16 calls on typeorm — unmistakable). **Small tweaks are not** — they sit under a ±20-point median swing.
- The engine, MCP server, and graph coverage are solid (typeorm loads in 1.5s, all prompts' symbols are indexed). The frontier is **model behavior**: mini uses the graph *and then* greps to verify, so adoption gains don't fully convert to token wins. Making the graph *replace* shell (as it does on nestjs: 0 shell, +65%) is the remaining lever.
- Honest trajectory: graph wins decisively where the baseline is expensive and the model goes graph-mostly; the median is dragged by cheap baselines and verify-grep. Measuring the next round of tuning will need n≥3 to clear the noise.

### Iteration 4 — raised graph_expand cap (200→400 lines): the grep/expand tradeoff

Goal: stop the verify-grep by returning whole method bodies. It **measurably reduced shell calls** (the cleaner, less-noisy signal): vscode 18→1, excalidraw 22→9, rxjs 31→13, shopping 18→12. typeorm flipped to +9% with quality 1/1. But a 400-line cap × heavy expansion **inflated graph-heavy responses** — excalidraw expanded 20 nodes and ballooned to 1.8M (−31%). So the win on shell was offset by expand bloat, and the **median landed at 0%**.

**The honest trajectory across four iterations: medians 5.5% → 12.5% → −11% → 0%.** That spread *is* the n=1 noise band; the real, reproducible signals are structural, not the median:
- graph adoption rose from 0 calls on 3/8 projects to using the graph everywhere (tool descriptions, iter2);
- verify-grep shell calls fell sharply (iter4);
- the graph wins big when the no-MCP baseline is expensive and the model goes graph-mostly (nestjs reaches 0 shell and +46–65%).

**Net:** the engine and measurement infrastructure are solid; the remaining gain is bounded by model behavior (mini supplements graph with shell and over-expands), and at n=1 the per-iteration median can't resolve sub-20-point changes. Continuing this productively means either n≥3 for precision, or attacking over-expansion/verify-grep more directly.

---

## ⚖️ Stable measurement (n=3) — the honest number on mini

Switching the graph arm to **n=3** (vs the fixed n=5 baseline) collapses the n=1 noise into a trustworthy median:

| project | baseline | graph (n=3 median) | Δ | graph/shell calls |
|---|---:|---:|:---:|:---:|
| vue | 296,899 | 236,271 | **+20%** | 6 / 0 |
| zod | 817,623 | 652,072 | **+20%** | 7 / 12 |
| excalidraw | 1,371,736 | 1,309,254 | +5% | 13 / 25 |
| nestjs | 534,318 | 601,778 | −13% | 10 / 7 |
| typeorm | 776,650 | 875,044 | −13% | **0** / 31 |
| shopping-backend | 440,192 | 546,773 | −24% | 7 / 11 |
| vscode | 713,103 | 1,143,513 | −60% | 10 / 9 |
| rxjs | 431,604 | 839,532 | −95% | 6 / 16 |
| **median** | | | **−13%** | |

**The n=1 +12.5% was a lucky draw; the stable truth on `gpt-5.4-mini` is −13% — the graph arm currently costs *more* than no MCP.** This is a **model-adoption result, not an engine result**:

- mini uses the graph as a *supplement*, not a *replacement* — at the n=3 median typeorm makes **0 graph calls** (31 shell), i.e. it ignores the graph and pays only its overhead; vscode/rxjs use graph *and* shell and over-expand, so the larger responses inflate the total.
- The graph itself works: **vue/zod +20%**, and on a stronger model the same engine already measured **+64%** (sonnet, typeorm overview). The win lands when the model goes graph-mostly; a weak model that won't drop shell doesn't get there.

So the conquest frontier on mini is squarely **adoption/replacement behavior**, and it is now measured cleanly at n=3 rather than guessed at n=1. Next levers stay the same — cut over-expansion and verify-grep so graph *replaces* shell — but now each change can be judged against this −13% baseline instead of noise.

### Cap revert (400→200), n=3 — the bloat was the culprit

A clean single-variable experiment against the −13% baseline: cutting graph_expand's cap back to 200 lines moved the **median −13% → −3%** (10 points), by shrinking the responses on the graph-heavy projects that the 400-line cap had blown up: **vscode −60% → +21%** (an 81-point swing), typeorm −13% → +13%, nestjs −13% → +29%, rxjs −95% → −66%. (excalidraw and vue moved the other way — render pipelines may need fuller bodies, or residual n=3 noise.)

So the current best stable mini number is **−3% median** — near break-even, with the graph now winning on vscode/nestjs/typeorm/zod. The lesson: a smaller per-response footprint beats fighting verify-grep by enlarging every expand. The path from −3% to positive is reducing the model's *expand count* (e.g. a signature line in query hits so it expands fewer nodes), measured at n=3 against this −3%.

---

## 🎯 Breakthrough: structure-first graph_expand → **+30% median** (stable n=3)

The decisive change, and the project's actual differentiator from codegraph's code-inlining: **graph_expand no longer dumps source.** It returns the *resolved shape* the graph already holds — each symbol's signature, and for a class/interface/namespace/file a member outline (every member with its own signature). The full body comes back only with `source: true`, for the few methods whose logic must be read.

Expanding a class went from thousands of source lines to ~80 member signatures, removing the response bloat the earlier n=3 runs had pinned as the graph arm's main cost.

| project | baseline | graph (n=3) | Δ |
|---|---:|---:|:---:|
| zod | 817,623 | 381,677 | **+53%** |
| shopping-backend | 440,192 | 233,148 | **+47%** |
| typeorm | 776,650 | 470,650 | **+39%** |
| excalidraw | 1,371,736 | 867,900 | **+37%** |
| nestjs | 534,318 | 413,220 | +23% |
| vue | 296,899 | 277,221 | +7% |
| vscode | 713,103 | 668,893 | +6% |
| rxjs | 431,604 | 559,646 | −30% |
| **median** | | | **+30%** |

**Stable n=3 trajectory: cap=400 −13% → cap=200 −3% → structure-first +30%.** A +33-point jump, not noise — the direct effect of returning structure instead of inlined code. **7 of 8 projects now win** (rxjs is the holdout where the model still makes zero graph calls), and answer quality rose too (nestjs and shopping reach 2/3). This is the graph being a graph — resolved facts, not a file reader — and it is what makes it cheaper than reading source, even on a small model.

### Query signatures — floor raised, median holds (~+28% confirmed)

Applying the same structure-over-source idea to graph_query — each shortlisted hit now carries its signature, so the model can often answer "what is this" without an expand round trip — came back **median-neutral but distribution-tightening** at n=3: it lifted the laggards (rxjs −30% → −4%, vscode +6% → +36%, nestjs +23% → +41%, excalidraw +37% → +54%) while trimming the leaders (typeorm +39% → +12%, shopping +47% → +17%). Median +30% → +26.5% — inside the n=3 noise.

So two independent n=3 runs now bracket the graph arm at **~+28% median token reduction on mini** — the conquest, confirmed. The full stable trajectory this phase: **−13% → −3% → +30% → +26.5%**, the decisive jump being structure-first `graph_expand` (resolved shape, never inlined source). Query signatures are kept for the robustness they add to the weak cases. The remaining drag (rxjs) is a single-file mechanism where a plain grep is genuinely competitive — the one shape the graph's multi-file resolution can't beat.

### Theoretical ceiling computed — and the mini wall

Fitting the 24 mini graph runs gives a clean cost model: **tokens ≈ 35k + 33k × tool-calls** — each call costs ~33k because every turn re-reads the growing context. So the reduction is set by *call count*:

| session | graph tokens | reduction vs median baseline (624k) |
|---|---:|:---:|
| 3 calls | 135k | **78%** |
| 5 calls | 201k | **68%** |
| 8 calls | 300k | 52% |
| mini today (~11 calls) | ~400k | **~28%** |

Per-project ceiling at a 5-call session: excalidraw 85%, zod 75%, typeorm 74%, vscode 72% — which is exactly **codegraph's published 78–90%** (measured on Opus). So the **theoretical target is ~70% median**, gated entirely on getting the model to ~5 calls.

The attack — instructions and the query description telling the model to answer from the signatures and stop expanding — **failed on mini**: median tool-calls held at ~11, and the n=3 median moved within noise (+17% this run vs +30 / +26.5 before). A weak model over-explores and won't honor "use fewer tools." So **mini's practical ceiling is ~25–28%**, and the theoretical 70% needs a model that naturally runs ~5 calls — a stronger one. (Sonnet already hit +64% on the typeorm overview; codegraph's 78–90% was Opus.) The next real step toward 70% is validating on sonnet, not nudging mini further.

### A→B path tool added (codegraph parity) — and an honest correction on the mini number

graph_trace now takes a `to` target and returns the shortest dependency path from `from` to it **in one call**, each path node carrying its signature — the reachability tool codegraph had and we lacked, and a natural fit since every dedicated prompt is a "how does A reach B" question. It works (trace `Repository.find → buildRelations` returns the exact 5-hop gold chain) and **mini adopts it** (graph_trace calls with `to` appear in the typeorm/nestjs/shopping runs).

But it did not move the mini median (n3 −2.5%), and that forces an honest correction: across six n=3 runs of the structure-first-era engine the median swung **+30 / +26.5 / +17 / −4.5 / −2.5** — the celebrated **+30% was a lucky high draw; the true central tendency is ~+15% in a wide ±20-point band.** The reason is now unambiguous: **mini's tool-call count (~12–16) is immovable** — the path tool, the concise prompts, and the few-calls nudge all failed to reduce it, and at ~33k tokens per call that caps the saving. So mini's real ceiling is the structure-first *response-size* win (~+15–25%, noisy), not the theoretical 70%, which needs a model that naturally runs ~5 calls. The path tool remains a genuine capability gain — and it should pay off far more on a stronger model, which is the next thing to measure (sonnet/opus already showed +64% on the overview).

---

## Session close — honest bottom line

- **Best stable mini result: ~+15% median token reduction** (wide ±20-pt n=3 noise band; the +30% earlier was a lucky draw). Engine: structure-first `graph_expand` + A→B path tool + query signatures.
- **A→B path tool added** — the codegraph reachability tool we lacked; `graph_trace(from, to)` returns the chain in one call (verified: the exact 5-hop TypeORM gold chain), each path node carrying its signature.
- **The hard wall is the model's call count** (`tokens ≈ 35k + 33k × calls`); on a capable model that runs ~5 calls the ceiling is ~70%. The final experiment — `graph_trace` returning *all* reached-node signatures to cut calls — **backfired** (n=3 −15.5%, response bloat) and was **reverted**; the lean retry (signatures on path nodes only) is documented in the handoff issue #279.
- **CI fixed for merge:** the all-red cascade was one stale `pnpm-lock.yaml` (test-graph gained `@ttsc/graph`) plus a missing `clientInfo` in the e2e `initialize` — both fixed and pushed.

Continuation handoff (full context, infra, theoretical model, next levers): **#279**.

PR #308, opened 2026-07-02
fix(ttsc): make TS config loaders and the virtual layout Windows-safe

### LINKED ISSUE #304: [Bug] @ttsc/banner and @ttsc/strip config loaders still hardcode `rootDir: "/"` (same TS6059 as #299)
## Description

#300 fixed the Windows `TS6059` failure (#299) for the **lint** config loader, but the identical pattern was copied across three Go drivers and only lint was fixed. Evaluating a TypeScript config for `@ttsc/banner` or `@ttsc/strip` on Windows still fails with the exact same error.

## Affected sites

Both are byte-for-byte the same construct as the pre-#300 lint code: an ephemeral tsconfig whose `files` are `[loader (in os.MkdirTemp temp dir), user config]`, with `rootDir` hardcoded to `"/"`:

- `packages/banner/driver/banner.go:510` (`typeScriptConfigLoaderTsconfig`, caller passes the system-temp dir)
- `packages/strip/driver/config.go:369` (`stripTypeScriptLoaderTsconfig`, same)

On Windows, `C:/...` is not under `rootDir: "/"` → `TS6059: File ... is not under 'rootDir' '/'`.

## Proposed fix

Apply the same volume-root resolution as #300, but extracted as a small named helper (e.g. `loaderRootDir(outDir string) string`) instead of the inline func literal, duplicated per driver package as the existing `ttsxCommandContext` copies are:

```go
func loaderRootDir(outDir string) string {
  vol := filepath.VolumeName(outDir)
  if vol == "" {
    return "/"
  }
  return filepath.ToSlash(vol + `\`)
}
```

## Follow-up cleanups from #300 to fold in

- **Formatting drift**: the `#300` hunk in `packages/lint/linthost/config.go` is not stable under the repo's `format:go` pipeline — the multi-line func literal splits gofmt's alignment section, so the next format run reflows the `skipLibCheck`/`strict`/`target` keys (verified empirically with `gofmt` + the `.vscode/gofmt-2spaces.sh` pipeline). Replacing the IIFE with the named helper above restores a single-line map entry and removes the drift.
- `tests/test-ttsc/src/features/ttsx-runtime/test_ttsx_serves_a_files_listed_source_outside_the_tsconfig_directory.ts:49` hardcodes `rootDir: "/"` in its own fixture, so it fails when the test suite runs on Windows — same one-line fix applies.
- (low) `packages/ttsc/src/launcher/internal/runtimeHooks.ts` `isWithin` compares the manifest `rootDir` (now the slash-form `C:/` from the synthesized tsconfig) against native backslash paths without normalizing separators; emit serving silently degrades to the fallback path on Windows. Worth normalizing while in the area.

## Related

- #299 (original report), #300 (lint fix)

### LINKED ISSUE #305: [Bug] TypeScript config evaluation fails on Windows when the project and the system temp dir are on different drives
## Description

#300 resolves `TS6059` (#299) by setting the ephemeral loader tsconfig's `rootDir` to the temp dir's volume root (e.g. `C:/`). That is correct for the common case, but when the user's project lives on a **different drive** than `os.tmpdir()` / `os.TempDir()` (e.g. project on `D:\`, `TEMP` on `C:\`), config evaluation still fails — and no `rootDir` value can fix it, because a single root cannot span two Windows volumes.

## Evidence (reproduced with typescript 7.0.1-rc / tsgo)

With loader in `C:\...\Temp\...` and the config on `D:\`:

```text
error TS6059: File 'D:/.../lint.config.ts' is not under 'rootDir' 'C:/'.
error TS2877: This import uses a '.ts' extension to resolve to an input TypeScript file,
              but will not be rewritten during emit because it is not a relative path.
```

The failure is actually multi-layered — it breaks before `rootDir` is even consulted:

- `packages/lint/src/index.ts:924` (`relativeImportSpecifier`): `path.relative` cannot cross drives on Windows, so it returns the absolute target and the function produces a bogus specifier like `./D:/proj/lint.config.ts`.
- The Go drivers share the same construction (`packages/lint/linthost/config.go`, `packages/banner/driver/banner.go`, `packages/strip/driver/config.go`).

## Proposed fix

Don't patch `rootDir` further — **create the loader temp dir on the same volume as the config file** when the system temp dir is on a different volume (e.g. under the project's `node_modules/.cache/ttsc-<pkg>-cfg/`, falling back to `os.tmpdir()` when volumes already match or no writable location exists). That fixes `rootDir` containment, the relative import specifier, and TS2877 in one move, on both the JS and Go sides.

## Related

- #299, #300, #304

### LINKED ISSUE #306: [Bug] `linkVirtualEntry` falls back to a plain `fs.symlinkSync` — EPERM on Windows without Developer Mode
## Description

Found while sweeping for Windows path/filesystem hazards after #300. In the virtual-directory materializer, `linkVirtualEntry` handles most entry kinds in a Windows-safe way, but the final fallback does not:

`packages/ttsc/src/launcher/internal/prepareExecution.ts:232`

- Directories → `fs.symlinkSync(..., "junction")` on win32 (lines 206-210) ✅
- Files → `fs.linkSync` with `fs.copyFileSync` fallback (lines 214-220) ✅
- Symlinks pointing at directories → `"junction"` (lines 223-229) ✅
- **Everything else → bare `fs.symlinkSync(realEntry, virtualEntry)` (line 232)** ❌

On Windows, a bare `fs.symlinkSync` creates a *file* symlink, which requires `SeCreateSymbolicLinkPrivilege` — without admin rights or Developer Mode it throws `EPERM`. This branch is reached whenever the materialized tree contains file symlinks (e.g. pnpm layouts, `node_modules/.bin/*` on some setups), crashing the run. Same class of problem as #218, which fixed the sibling branches but missed this one.

## Proposed fix

Mirror the file branch: try `fs.linkSync` (or `fs.symlinkSync`) and fall back to `fs.copyFileSync` on Windows when symlink creation fails.

## Related

- #218 (junction/EPERM precedent), #300 (Windows sweep origin)

samchon:
## Overview

Completes the Windows path-hardening that #300 started, in three parts:

- **#304** — the `banner` and `strip` Go drivers still hardcoded the loader tsconfig's `rootDir` to `"/"`, which is never an ancestor of drive-letter paths, so every TypeScript config evaluation failed with `TS6059` on Windows. The volume-root resolution is extracted into a named `loaderRootDir` helper (per driver, mirroring the existing per-package helper duplication) and applied to all three drivers; the inline func literal from #300 is replaced by the same helper, which also restores gofmt-stable formatting in `linthost/config.go`.
- **#305** — when the system temp dir and the config file live on different volumes (TEMP on `C:`, project on `D:`), no single `rootDir` spans both `files` entries and relative import specifiers cannot cross drives, so evaluation still failed. All three Go drivers and the lint JS factory now create the ephemeral loader tree under the config's nearest `node_modules/.cache` when the volumes differ, keeping the system temp dir when they match (POSIX behavior is unchanged: both roots are always `/`).
- **#306** — `linkVirtualEntry`'s final branch re-symlinked file-symlink entries with a bare `fs.symlinkSync`, which needs `SeCreateSymbolicLinkPrivilege` on Windows and crashed the run with EPERM (same class as #218). It now falls back to the plain-file branch's hard-link/copy strategy, skipping dangling links.

Follow-ups folded in from #304: the runtime hooks' `isWithin` normalizes separators (the synthesized manifest `rootDir` is slash-form `C:/` while real paths are native), and the ttsx e2e fixture that mirrored the loader's `rootDir: "/"` synthesis now derives the volume root.

## Test plan

- New Go unit tests (one case per file, per package): the generated loader tsconfig's `rootDir` contains both `files` entries, and `loaderTempBase` keeps the system temp on matching volumes / moves under `node_modules/.cache` across volumes (the cross-volume branch runs for real on drive-letter platforms, guarded by a `filepath.VolumeName` probe elsewhere).
- New ttsx e2e test locking the file-symlink virtual-entry branch; the existing outside-tsconfig-directory e2e keeps covering emit serving under a root `rootDir`.
- Verified on Windows 11 (repo on `D:`, TEMP on `C:`): all six new Go tests pass, including the real cross-volume branch; full `strip` Go suite passes; `banner`/`lint` suites show only failures already present on unchanged `master` (sh-script fake launchers, `ttsx` subprocess resolution — environmental, tracked separately). `tsc --noEmit` is clean for `packages/lint` and `packages/ttsc`; changed files formatted with prettier and the repo's gofmt pipeline.
- Not runnable locally on Windows: the TS e2e suites (DynamicExecutor limitation) — they run on Linux CI.

## Deferred

- The residual cross-volume edge (config volume with no `node_modules` anywhere) intentionally falls back to the historical system-temp behavior.
- The Windows-only EPERM fallback path of #306 has no automated test: creating a file-symlink fixture on Windows requires the same privilege the fallback exists to avoid.

Resolves #304, resolves #305, resolves #306.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,12 +20,18 @@ concurrency:
 jobs:
   test:
     name: ${{ matrix.name }}
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os || 'ubuntu-latest' }}
     strategy:
       fail-fast: false
       matrix:
         include:
           - { name: go, run: "pnpm run test:go && pnpm --filter ttsc go:vet" }
+          # Minimal Windows safety net: the Go unit suites cover the
+          # platform-sensitive layers (config loaders, path handling) without
+          # the e2e runner, which is POSIX-only. Only the ttsx launcher build
+          # is needed (test-go-lint drives real ttsx), so the full
+          # build:current is skipped.
+          - { name: windows-go, os: windows-latest, build: "pnpm --filter ttsc build", run: "pnpm run test:go" }
           - { name: shim-audit, run: "pnpm --filter ttsc shim:audit:test && pnpm --filter ttsc shim:audit" }
           - { name: typecheck, run: "pnpm run check:flags && pnpm run test:typecheck" }
           - { name: banner, run: "pnpm --filter @ttsc/test-banner start" }
@@ -61,10 +67,11 @@ jobs:
         run: pnpm install --frozen-lockfile
 
       - name: Use System Go For Source Plugin Tests
+        shell: bash
         run: echo "TTSC_GO_BINARY=$(go env GOROOT)/bin/go" >> "$GITHUB_ENV"
 
       - name: Build Current Platform
-        run: pnpm run build:current
+        run: ${{ matrix.build || 'pnpm run build:current' }}
 
       - name: Run ${{ matrix.name }}
         run: ${{ matrix.run }}
--- a/packages/banner/driver/banner.go
+++ b/packages/banner/driver/banner.go
@@ -546,7 +546,11 @@ func loaderRootDir(outDir string) string {
 // on the config's volume beats the system temp dir, which is guaranteed to
 // fail. Returns "" (the os.MkdirTemp default) when the volumes already match.
 func loaderTempBase(location, systemTemp string) string {
-  if strings.EqualFold(filepath.VolumeName(systemTemp), filepath.VolumeName(location)) {
+  // A relative location has no volume; "" must not be read as "a volume
+  // other than the system temp's" — it keeps the historical default (and
+  // the Rel-failure contract for relative config paths).
+  vol := filepath.VolumeName(location)
+  if vol == "" || strings.EqualFold(filepath.VolumeName(systemTemp), vol) {
     return ""
   }
   nodeModules := findNearestNodeModules(filepath.Dir(location))
--- a/packages/banner/driver/banner.go
+++ b/packages/banner/driver/banner.go
@@ -541,22 +541,52 @@ func loaderRootDir(outDir string) string {
 // loader cannot work from there — no single tsconfig rootDir spans two
 // volumes and filepath.Rel cannot produce a relative import across drives
 // (#305) — so the tree is created under the config's nearest
-// node_modules/.cache instead. Returns "" (the os.MkdirTemp default) when the
-// volumes already match, or when no node_modules/.cache is available to keep
-// the historical behavior as the fallback.
+// node_modules/.cache instead, falling back to the config's own directory
+// when no node_modules exists (or its .cache cannot be created): any location
+// on the config's volume beats the system temp dir, which is guaranteed to
+// fail. Returns "" (the os.MkdirTemp default) when the volumes already match.
 func loaderTempBase(location, systemTemp string) string {
   if strings.EqualFold(filepath.VolumeName(systemTemp), filepath.VolumeName(location)) {
     return ""
   }
   nodeModules := findNearestNodeModules(filepath.Dir(location))
   if nodeModules == "" {
-    return ""
+    return filepath.Dir(location)
   }
-  base := filepath.Join(nodeModules, ".cache")
+  // Resolve a linked node_modules (junction/symlink — common in managed
+  // setups) before descending into it: the ESM runtime realpaths the loader
+  // module at import time, and a relative config specifier computed from the
+  // link-form path would resolve against the wrong directory. NTFS junctions
+  // defeat filepath.EvalSymlinks, so the link component is chased by hand
+  // first. Realpathing may also land on another volume, which defeats the
+  // whole point — fall back to the config's directory then.
+  base := filepath.Join(resolveDirLink(nodeModules), ".cache")
   if err := os.MkdirAll(base, 0o755); err != nil {
-    return ""
+    return filepath.Dir(location)
   }
-  return base
+  real, err := filepath.EvalSymlinks(base)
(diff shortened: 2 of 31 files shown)


samchon:
Follow-up push (f3e6875e) closing every deferred item from the PR body:

- **Linked `node_modules` bug found and fixed by a live cross-drive run.** Driving the JS factory end-to-end (project on `D:`, TEMP on `C:`, real ttsx + tsgo) surfaced that a junction/symlinked `node_modules` broke the loader: the ESM runtime realpaths the loader module, so the relative config import computed from the link-form `.cache` path resolved against the wrong directory. The temp base is realpathed now; since NTFS junctions defeat `filepath.EvalSymlinks` (verified empirically), the Go drivers chase the link with `os.Readlink` first.
- **The residual cross-volume edge is closed**: no usable `node_modules/.cache` now falls back to the config's own directory (always the right volume, keeps the import a plain `./`), not the guaranteed-to-fail system temp dir.
- **The Windows EPERM fallback of `linkVirtualEntry` is now covered**: `linkVirtualEntry` and `isWithin` are exported for the api e2e suite; the dangling-entry test uses a junction fixture, so it exercises the real privilege-less EPERM path (confirmed on a machine where `fs.symlinkSync` file links throw EPERM), and the copy-fallback test drives the same symlink→hard-link→copy chain deterministically. `isWithin` also folds case on Windows and has a root/slash/case matching test.
- Temp-base tests now pin the blocked-`.cache`, missing-`node_modules`, and junction-realpath branches per package.

Verification on Windows 11 (D: repo + C: TEMP): full `lint` and `strip` Go suites pass; `banner` shows only the sh-script fake-launcher failures already present on unchanged `master`; live cross-drive TS config evaluation passes through both the JS factory and the Go driver; `tsc --noEmit` clean for `packages/lint` and `packages/ttsc`.

PR #309, opened 2026-07-02
fix(website): review the graph doc pages and fix the 3D viewer edge-kind mapping

samchon:
## Intent

A self-contained review pass over the four `/docs/graph` pages (built from the launch post), fixing what read abrupt, thin, or factually stale, plus one real defect the review surfaced in the 3D viewer pipeline.

## Scope

**Viewer edge-kind mapping (fix).** `ttscgraph dump` emits fine-grained wire kinds (`calls`, `instantiates`, `renders`, `accesses`, `type_ref`, `extends`, `implements`), but the 3D viewers color and label edges by the three display families (`value-call`, `type-ref`, `heritage`). No reduce mapped between the two, so a freshly dumped graph uploaded to the viewer rendered every edge with the fallback color and the legend matched nothing (the shipped example payloads predate the fine-grained kinds, which hid it). The mapping is added to all three mirrored reduces: `packages/graph/src/reduce.ts`, `website/src/components/graph/TtscWebsiteGraphReduce.ts`, `experimental/benchmark/graph/viewer.mjs`.

**Docs.**
- Overview: correct the resident-graph claim (the native binary dumps once, the Node server answers from the in-memory graph; there is no warm checker), map -> graph wording, add the source-crawl hero image, a diagnostics bullet, and embed the live 3D viewer component.
- Design: explain the `clarify` follow-up action, move the CoT-compliance image here where the concept is defined.
- Comparison: replace that image (out of context there) with a prose pointer to Design.
- Viewer: correct the dump description (ids and paths are project-relative, not absolute), document the `view` flags, name the example repos, tie the three edge families to the full edge-kind set, and spell out the reduction steps.
- Setup: same map -> graph wording fix.

## Deferred

- The shipped example payloads under `website/public/graph/*.json` still carry the old family kinds; they render correctly (pre-reduced payloads bypass the mapping) and were not regenerated.

## Test plan

- `tsc --noEmit` passes in `packages/graph` and `website`.
- Runtime repro of the mapping: a synthetic dump with `calls`/`type_ref`/`extends`/`accesses` edges reduces to `value-call`/`type-ref`/`heritage`/`value-call` (was: kinds passed through and rendered colorless).
- All changed files formatted with the repo prettier config.

https://claude.ai/code/session_01N3Cn3zmu1uUVkYJfwyPFeJ

### DIFF
--- a/experimental/benchmark/graph/viewer.mjs
+++ b/experimental/benchmark/graph/viewer.mjs
@@ -74,6 +74,26 @@ function rewriteId(id, root) {
 }
 
 /**
+ * Collapse the fine-grained wire kinds `ttscgraph dump` emits (calls,
+ * instantiates, renders, accesses, type_ref, extends, implements) into the
+ * three display families the viewer colors and its legend name. An unknown kind
+ * passes through and renders with the fallback color.
+ */
+const DISPLAY_KIND = {
+  calls: "value-call",
+  instantiates: "value-call",
+  renders: "value-call",
+  accesses: "value-call",
+  type_ref: "type-ref",
+  extends: "heritage",
+  implements: "heritage",
+};
+
+function displayKind(kind) {
+  return DISPLAY_KIND[kind] ?? kind;
+}
+
+/**
  * Reduce a raw dump to the viewer payload: relativized, external-free, capped
  * to the highest-degree nodes, with orphans pruned. Returns `{ nodes, links }`
  * shaped for react-force-graph (node.id, link.source/target).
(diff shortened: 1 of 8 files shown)


samchon:
Follow-up: dropped the embedded 3D viewer from the overview page (three.js + a multi-MB example graph is too heavy for an entry page) and replaced it with a static TypeORM render linking to the dedicated viewer page.

samchon:
Correction: the previous follow-up was a misread of the review feedback. The embedded 3D viewer on the overview page is intentional and is now restored (revert of 1fcdb0b). Final state: the overview embeds the live viewer component in section 3.2.

PR #315, opened 2026-07-02
fix(ttsc): confine forced emit to outDir so dependency sources stay clean

### LINKED ISSUE #293: Forced emit (--emit) writes dependency .js into its source tree for files outside rootDir
## Summary

When `ttsc` is forced to emit (`--emit`) for a project whose program pulls in a `.ts` file that lives **outside** the project's `rootDir` — e.g. a workspace dependency whose `main`/entry resolves (via symlink realpath) to raw TypeScript source — `ttsc` writes the compiled `.js` **next to that dependency's source** instead of under the project's `outDir` (or refusing to emit it).

The result is that an unrelated package's source tree gets polluted with untracked `.js` files on every emit-forcing build. In a monorepo this shows up as a pile of stray `packages/<dep>/src/*.js` files after running tests/builds.

## Environment

- `ttsc` 0.16.3
- Node v22.15.0
- Windows 11 (win32, x64)
- Also reproduces through the native `runBuild --emit` path (the typia source plugin), where it leaks even **without** `--noCheck` because that path emits past type-check diagnostics.

## Minimal repro (plugin-free)

A project that just *imports* a workspace dependency whose entry is raw `.ts`. Here the dependency is `typia` (its `package.json` `main` is `src/index.ts`), resolved through the workspace symlink `node_modules/typia -> packages/typia`.

`proj/tsconfig.json`
```json
{
  "compilerOptions": {
    "module": "commonjs",
    "moduleResolution": "bundler",
    "rootDir": "src",
    "outDir": "dist",
    "strict": true,
    "skipLibCheck": true
  },
  "include": ["src"]
}
```

`proj/src/main.ts`
```ts
import typia from "typia";
void typia;
export const x = 1;
```

Run:
```bash
ttsc -p proj/tsconfig.json --emit --noCheck
```

(`--noCheck` is only needed in the plugin-free CLI repro to get past the `TS6059` check that otherwise blocks emit; the typia `runBuild --emit` path emits past that diagnostic on its own, so it leaks without `--noCheck`.)

## Actual

- `proj/dist/main.js` ✅ (the project's own file, correctly under `outDir`)
- `packages/typia/src/index.js`, `llm.js`, `json.js`, `plain.js`, `notations.js`, `protobuf.js`, `reflect.js`, `module.js`, `http.js`, `compare.js`, `functional.js`, `re-exports.js`, `TypeGuardError.js`, `transformers/NoTransformConfigurationError.js` ❌ — 14 real compiled CJS files written **into the dependency's own source directory**, far outside `proj/`.

The leaked files are full compiled output (`"use strict"; var __createBinding …`).

## Expected

For a file that cannot be placed under `outDir` because it sits outside `rootDir`, `ttsc` should do one of:

1. **Not emit it** — treat a realpath-resolved dependency `.ts` the way `tsc` treats `node_modules` (external library, no emit); or
2. **Confine all output under `outDir`** using a computed relative path; or
3. **Refuse with `TS6059`** and emit nothing, which is exactly what plain `ttsc -p … ` (without forced emit) already does here.

What it must **not** do is write generated `.js` into a sibling package's source tree.

## Note on `tsc` parity

Plain `ttsc -p proj/tsconfig.json` (no `--emit`) correctly reports `TS6059` ("File '…/packages/typia/src/http.ts' is not under 'rootDir' …") and writes nothing. The divergence appears only once emit is **forced** while the `rootDir`-containment check is bypassed (`--noCheck`, or the plugin `runBuild` emit path): the emit step then computes the output path next to the source rather than honoring `outDir`.

## Impact

- Pollutes dependency/sibling-package source trees with untracked build artifacts on every emit-forcing build or test run.
- In a workspace this corrupts `git status` and risks the stray `.js` being picked up by subsequent tooling.
- Surfaced while building typia's recursive-random diagnostic tests, which call `runBuild --emit` against throwaway projects that `import typia`; each run leaves ~14 `.js` files under `packages/typia/src/`.

samchon:
## Intent

Fix #293: a forced emit (`--emit`, plugin `runBuild`) writes a dependency's compiled `.js` into that dependency's own source tree when the dependency's raw `.ts` sources sit outside the project's `rootDir`. In the typia workspace this polluted `packages/typia/src/` with 14 stray `.js` files on every diagnostic-test run.

## Fact-check summary

The issue is real, but its mechanism differs from the report:

- **Trigger is package self-reference, not symlink realpath.** A module resolved through `node_modules` (symlink or junction, realpath'd or not) is classified as an external-library file and is already excluded from emit — the issue's own "minimal repro" does not reproduce, verified in the exact original environment (ttsc 0.16.3, typescript 7.0.1-rc, typia workspace). What does reproduce — exactly the 14 reported files — is a project nested **inside** the dependency package's directory (typia's throwaway test projects under `packages/typia/native/.tmp-ttsc-typia-tests/`): the `"typia"` import then resolves by package self-reference with no `node_modules` hop, so the dependency's sources are not external and stay in the forced-emit set.
- **No tsc divergence upstream.** Plain `tsc` 5.9.3 and `tsgo` 7.0.1-rc both leak identically when emit is forced past the TS6059 check with `--noCheck`. The plugin-free ttsc CLI lane (which spawns tsgo directly) keeps that upstream behavior — without `--noCheck` it already refuses via TS6059 + `--noEmitOnError`.
- **Where ttsc was genuinely at fault:** the driver's forced-emit lanes (`EmitAllRaw`, `EmitWithPluginTransformers`, …) emit past the TS6059 class silently — `Program.Diagnostics()` carries bind/semantic diagnostics only — so plugin hosts leaked with no `--noCheck` anywhere and no diagnostic to see.

## Scope

- `driver/emit_containment.go` (new): `outputEscapesOutDir` — when the project configures `outDir`, an output whose computed path would escape it (and `declarationDir`, when set) is skipped like a `node_modules` external: no write, no diagnostic. `.tsbuildinfo` is exempt (its default home is next to the tsconfig).
- `driver/rewrite.go`: guard wired into the `EmitAll`/`EmitFile` and `EmitAllRaw` WriteFile funnels; skipped writes flagged via `WriteFileData.SkippedDtsWrite` so `EmitResult.EmittedFiles` counts only real writes.
- `driver/emit_plugin.go`: `EmitWithPluginTransformers` resolves output paths before transforming and skips escaping files entirely (no wasted transform/print work).
- Docs: `ttsc/compile.mdx` (`--emit` flag row), `development/reference/driver-api.mdx` (new *Output containment* section).

## Test plan

- Go unit (red→green verified): `test/driver/emit_raw_skips_outputs_outside_outdir_for_self_referenced_dependency_test.go` and `test/driver/emit_plugin_transformers_skip_outputs_outside_outdir_test.go` materialize the self-reference layout and assert output is confined to `outDir` while the project's own `main.js` still emits.
- TS e2e (red→green verified against the real binaries): `test_ttsc_utility_plugins_forced_emit_confines_output_to_outdir` runs `ttsc --emit` on a `@ttsc/banner` project nested inside a dependency package and asserts no `.js` lands in the dependency's `src/`.
- Cross-checked against the live typia workspace: `go test -run TestRandomRecursiveMinItemsDiagnostic` leaked the 14 files before; the driver dry-run harness confirms both lanes emit only `dist/main.js` after.
- Pre-existing Windows-local failures (`TestUtilityCommandFailuresCoverHostEdges`, `TestDriverLinkedPluginsRegistersAndAppliesProgram`) fail identically on master; unrelated.

## Deferred

- The plugin-free CLI lane (`ttsc --emit --noCheck`, tsgo spawned directly) intentionally keeps upstream tsc/tsgo behavior; confining it would mean post-filtering tsgo's own writes. Candidate for an upstream typescript-go issue instead.
- `Program.Diagnostics()` still omits the TS6059 class (config/program diagnostics); surfacing them would flip typia's diagnostic tests from exit 3 to exit 2, so it needs a coordinated change downstream.

https://claude.ai/code/session_01K9aUTnNS9jaLXEfVZR41bd

### DIFF
--- a/packages/ttsc/driver/emit_containment.go
+++ b/packages/ttsc/driver/emit_containment.go
@@ -1,0 +1,52 @@
+// Package driver: forced-emit output containment.
+//
+// tsc guards a program whose sources spill outside `rootDir` with TS6059 and,
+// under `--noEmitOnError`, refuses to emit. ttsc's forced-emit lanes
+// (`--emit`, a plugin host's runBuild) intentionally proceed past that guard,
+// and tsgo then computes the output path of an out-of-rootDir source relative
+// to the common source directory — which resolves to a `.js` right next to the
+// dependency's own source (issue #293). The classic trigger is package
+// self-reference: a project nested inside a dependency's directory resolves
+// the dependency's name without a node_modules hop, so its sources are not
+// classified as external-library files and stay in the emit set.
+//
+// The guard here confines every ttsc-owned emit lane to the project's
+// configured `outDir`: an output that would escape it is silently skipped, the
+// same way tsc treats node_modules externals (no emit, no error).
+package driver
+
+import (
+  "strings"
+
+  "github.com/microsoft/typescript-go/shim/tspath"
+)
+
+// outputEscapesOutDir reports whether fileName — an emit output path tsgo
+// computed for this program — would land outside the project's configured
+// `outDir` (and `declarationDir`, when set). Projects without `outDir` emit
+// next to their sources by design, so the guard only applies when `outDir`
+// gives the project an output boundary. `.tsbuildinfo` is exempt because its
+// default location is next to the tsconfig, legitimately outside `outDir`.
+func (p *Program) outputEscapesOutDir(fileName string) bool {
+  if p == nil || p.TSProgram == nil {
+    return false
+  }
+  options := p.TSProgram.Options()
+  if options.OutDir == "" {
+    return false
+  }
+  if strings.HasSuffix(fileName, ".tsbuildinfo") {
+    return false
+  }
+  cmp := tspath.ComparePathsOptions{
+    UseCaseSensitiveFileNames: p.TSProgram.UseCaseSensitiveFileNames(),
+    CurrentDirectory:          p.TSProgram.GetCurrentDirectory(),
+  }
+  if tspath.ContainsPath(options.OutDir, fileName, cmp) {
+    return false
+  }
+  if options.DeclarationDir != "" && tspath.ContainsPath(options.DeclarationDir, fileName, cmp) {
+    return false
+  }
+  return true
+}
--- a/packages/ttsc/driver/emit_plugin.go
+++ b/packages/ttsc/driver/emit_plugin.go
@@ -143,6 +143,10 @@ func (p *Program) EmitWithPluginTransformers(transforms []PluginTransform, write
   host := &pluginEmitHost{program: p.TSProgram, emitResolver: p.Checker.GetEmitResolver()}
   options := p.TSProgram.Options()
   for _, sf := range shimcompiler.GetSourceFilesToEmit(host, nil, false) {
+    paths := shimcompiler.GetOutputPathsFor(sf, options, host, false)
+    if p.outputEscapesOutDir(paths.JsFilePath()) {
+      continue
+    }
     ec := shimprinter.NewEmitContext()
     out := sf
     for _, transform := range transforms {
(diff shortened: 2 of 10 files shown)


samchon:
Follow-up: added a negative-twin regression pinning the DeclarationDir branch of the guard — a `declaration + declarationDir` project's `.d.ts` outputs legitimately escape `outDir` and must survive containment while the dependency's outputs still get skipped (verified by probe before writing the test).

samchon:
Follow-up: pinned the third emit funnel (`emit()` behind `EmitAll`/`EmitFile`) with its own containment regression — it is a separate code path from `EmitAllRaw`'s funnel, so each carries its own lock.

PR #320, opened 2026-07-03
fix(ttsc): normalize resident serve request paths for Windows

### LINKED ISSUE #319: Windows transform requests should normalize native absolute paths before TypeScript-Go APIs
Moved from: https://github.com/samchon/nestia/issues/1520

## Symptom

On Windows, running `ttsc` with `@nestia/core` can panic during `@TypedRoute` / `@TypedBody` transforms:

```text
error TS(nestia.core.TypedRoute): fileName should be normalized and absolute: D:\tmp\<uuid>.ts
error TS(nestia.core.TypedBody): fileName should be normalized and absolute: D:\tmp\<uuid>.ts
```

The original report says the transform is skipped and emit can still continue with raw decorators left in the output.

Environment from the original report:

- Windows 11
- Node 22.15.0
- pnpm 10.10.0
- `ttsc` 0.16.9
- `typescript` 7.0.1-rc
- `@nestia/core` / `@nestia/fetcher` 12.0.0-dev.20260619.1
- `typia` 13.0.0-dev.20260619.1

## Why this belongs in ttsc

The `@nestia/core` transform path does not create temp files for this flow:

- `@nestia/core/native/transform.cjs` only returns a ttsc plugin descriptor whose `source` points at `cmd/ttsc-nestia`.
- `@nestia/core/native/transform/transform.go` normalizes the incoming `--file` before lookup:

```go
absFile = filepath.ToSlash(absFile)
target := prog.SourceFile(absFile)
```

The tmp/scratch paths are owned by ttsc machinery: source-plugin build scratch dirs, plugin descriptor temp dirs, in-memory/single-file emit temp output dirs, unplugin temp tsconfig, and the resident transform host.

The likely boundary bug is in the resident/request path normalization:

- `packages/ttsc/src/TtscService.ts` resolves request paths with Node `path.resolve(...)`, which yields native backslash paths on Windows.
- `packages/ttsc/utility/serve.go` `resolveServePath` returns an absolute path unchanged, or uses `filepath.Join`, so Windows-native absolute paths can continue into host/cache/overlay logic.
- TypeScript-Go expects normalized absolute paths. The panic text matches a native Windows path reaching that layer without `tspath.NormalizePath` / `tspath.ResolvePath` normalization.

## Minimal repro from nestia issue

`tsconfig.json`:

```json
{
  compilerOptions: {
    target: ES2022,
    module: commonjs,
    experimentalDecorators: true,
    emitDecoratorMetadata: true,
    strict: true,
    skipLibCheck: true,
    plugins: [
      { transform: typia/lib/transform },
      { transform: @nestia/core/lib/transform },
      { transform: @ttsc/lint/lib/transform }
    ]
  },
  include: [src]
}
```

`src/index.ts`:

```ts
import { Controller } from @nestjs/common;
import { TypedBody, TypedRoute } from @nestia/core;
import typia, { tags } from typia;

interface Input {
  id: string & tags.Format<uuid>;
  name: string & tags.MinLength<1>;
}

@Controller()
export class AppController {
  @TypedRoute.Post(/echo)
  public echo(@TypedBody() input: Input): Input {
    return typia.assert(input);
  }
}
```

## Expected

All Windows paths crossing the resident transform / single-file transform boundary should be normalized before they reach TypeScript-Go APIs, caches, overlay keys, or source-file lookup.

## Possible fix direction

Normalize request paths at the ttsc boundary, for example by using the TypeScript-Go path helpers instead of `filepath` for source identity:

- absolute request path: `tspath.NormalizePath(file)`
- relative request path: `tspath.ResolvePath(cwd, file)`

This should apply consistently to `file` and `update` requests, plus any temp source path passed through the same host path.

samchon:
## Summary
- The resident `serve` host's `resolveServePath` passed a caller-supplied `file`/`update` request path through `filepath.Join`/unchanged, leaving native Windows backslash paths un-normalized before they reached the per-file cache key, the `OverlayFS`, and (via a linked transform plugin) TypeScript-Go APIs.
- `SourceFile.FileName()` and `OverlayFS.key()` already normalize through `tspath`, so an un-normalized request path could mismatch them; #319 reports a linked plugin panicking with "fileName should be normalized and absolute" on Windows.
- `resolveServePath` now routes through `tspath.ResolvePath`, which normalizes slashes and resolves relative paths against `cwd` in one call (it already discards `cwd` when the given path is rooted), matching the normalization the rest of the resident host relies on.

## Test plan
- [x] `go build ./...` in `packages/ttsc`
- [x] `go test ./test/utility/...` (all pass; new `TestUtilityServeNormalizesBackslashRequestPaths` added, which fails against the pre-fix code even on a Windows test runner since Go's native `path/filepath` already tolerates backslashes there — the regression only surfaces once the request path is compared against TypeScript-Go's own forward-slash-normalized keys)
- [x] `go test ./...` in `packages/ttsc` (pre-existing unrelated failure: `TestUtilityCommandFailuresCoverHostEdges`, a Windows filesystem-permissions test that fails identically on `master` without this change)

Fixes #319

### DIFF
--- a/packages/lint/linthost/compile.go
+++ b/packages/lint/linthost/compile.go
@@ -26,6 +26,7 @@ import (
   shimast "github.com/microsoft/typescript-go/shim/ast"
   shimcompiler "github.com/microsoft/typescript-go/shim/compiler"
   shimdw "github.com/microsoft/typescript-go/shim/diagnosticwriter"
+  shimtspath "github.com/microsoft/typescript-go/shim/tspath"
 )
 
 // RunCheck implements `@ttsc/lint check` — typecheck + lint, no emit.
@@ -118,10 +119,13 @@ func RunTransform(args []string) int {
     return 2
   }
 
-  absFile := *file
-  if !filepath.IsAbs(absFile) {
-    absFile = filepath.Join(resolvedCwd, absFile)
-  }
+  // tsgo normalizes SourceFile.FileName() through tspath, resolving "."/".."
+  // segments as well as separators. findSourceFile's comparison only swaps
+  // separators, so an absolute --file value carrying an unresolved "."/".."
+  // round-trip (or, on a POSIX host, backslash separators) could name the
+  // right file and still miss (samchon/ttsc#319 is this same gap in ttsc's
+  // resident serve host).
+  absFile := shimtspath.ResolvePath(resolvedCwd, *file)
   target := prog.findSourceFile(absFile)
   if target == nil {
     fmt.Fprintf(os.Stderr, "@ttsc/lint transform: source file not in program: %s\n", absFile)
--- a/packages/lint/linthost/host.go
+++ b/packages/lint/linthost/host.go
@@ -22,6 +22,7 @@ import (
   shimcompiler "github.com/microsoft/typescript-go/shim/compiler"
   shimcore "github.com/microsoft/typescript-go/shim/core"
   "github.com/microsoft/typescript-go/shim/tsoptions"
+  shimtspath "github.com/microsoft/typescript-go/shim/tspath"
   "github.com/microsoft/typescript-go/shim/vfs/cachedvfs"
   "github.com/microsoft/typescript-go/shim/vfs/osvfs"
 )
@@ -227,11 +228,17 @@ func (p *program) programDiagnostics() []*shimast.Diagnostic {
 }
 
 // findSourceFile locates a source file in the program by absolute path.
-// tsgo normalizes paths to forward slashes; we do the same on our side.
+// tsgo normalizes SourceFile.FileName() through tspath (forward slashes,
+// resolved "."/".." segments); a bare filepath.ToSlash only swaps separator
+// characters for the host OS, so a caller-supplied path with an unresolved
+// "."/".." round-trip (or, on a POSIX host, backslash separators surviving
+// from a Windows-authored path) could still fail to match here even after
+// that conversion. Normalize both sides through tspath instead — the same gap
+// this closes in ttsc's resident serve host (samchon/ttsc#319).
 func (p *program) findSourceFile(target string) *shimast.SourceFile {
-  want := filepath.ToSlash(target)
+  want := shimtspath.NormalizePath(target)
   for _, file := range p.tsProgram.SourceFiles() {
-    if filepath.ToSlash(file.FileName()) == want {
+    if shimtspath.NormalizePath(file.FileName()) == want {
       return file
     }
   }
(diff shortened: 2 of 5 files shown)


samchon:
Follow-up from self-review (no CI dependency — verified locally):

**Found a second, concrete instance of the same root-cause class in `@ttsc/lint`.** `packages/lint/linthost/compile.go`'s `RunTransform` (`@ttsc/lint transform --file=PATH`, the same "single-file transform command" shape the issue describes for `@nestia/core`'s own binary) built `--file`'s absolute form with a native `filepath.Join`, and `findSourceFile` (`host.go`) compared it against `SourceFile.FileName()` using only `filepath.ToSlash`. `ToSlash` swaps separators for the host OS but never resolves `"."`/`".."` segments, and is a no-op for backslashes when running on a POSIX host — so an absolute `--file` value carrying either property could name the right file and still report `"source file not in program"`.

Fixed both to go through `tspath.ResolvePath`/`tspath.NormalizePath`. Verified empirically (not just by inspection): reverted just the source fix, kept the new test, and confirmed it fails with the exact pre-fix error (`source file not in program: ...\src\..\src\main.ts`); restored the fix and confirmed it passes.

**Corrected an inaccurate claim in both tests' doc comments.** I originally wrote that a plain backslash-only absolute path reproduces the bug "on any host OS, including Windows." That's wrong for a same-host Windows repro: Go's own `path/filepath` already treats backslash as a proper separator on Windows, so `filepath.Rel`/`filepath.ToSlash`/`filepath.Clean` already normalize a plain backslash path correctly there — I verified this directly by reverting the `serve.go` fix and confirming its test still passed on Windows. The gap that reproduces on **any** host OS, Windows included, is specifically an unresolved `"."`/`".."` round-trip (which `ToSlash` never collapses) — that's what the new `packages/lint` test exercises, and it's the one I have a verified fail-before/pass-after trace for. The `serve.go` test's backslash case is still real coverage (it discriminates on a POSIX host, e.g. this repo's default `ubuntu-latest` `go` CI job, where backslash isn't a separator at all), just not a same-host-Windows repro by itself. Both test comments now say this accurately instead of overclaiming.

Net: `packages/ttsc/utility/serve.go` (original commit) is a real consistency fix but its own test doesn't prove it closes a live bug on Windows by itself; `packages/lint`'s fix (second commit) is the one with a verified fail-before/pass-after trace matching the issue's reported failure mode ("source file not in program" / "not normalized"). Full local validation: `go build`/`go vet` clean, `packages/ttsc` utility suite green, `packages/lint`'s Go suite green apart from two pre-existing, unrelated failures (`TestLoadRuleConfigLoadsTypeScriptConfigFile`, `TestLoadRuleConfigTypeScriptConfigFileRoundTripsFormatBlock` — Node ESM/CJS interop loading a `.ts` lint config file, reproduces identically on `master` without this branch's changes).

PR #331, opened 2026-07-07
fix(lint): load TS configs through file URLs

samchon:
## Summary

Fixes TypeScript lint config evaluation when the temporary loader directory is reached through a path whose realpath differs from the path used to compute relative imports, such as macOS `/var` resolving to `/private/var`.

The lint config loaders now import user TypeScript config files through absolute `file://` URLs instead of static relative imports that TypeScript-Go can rewrite to `.js` and Node can resolve from a realpathed loader path.

## Scope

- Use dynamic `file://` imports in the JS factory plugin extractor.
- Use dynamic `file://` imports in the Go sidecar config loader.
- Add a synthetic realpath-drift e2e case for `lint.config.ts`.
- Add the same installed-tarball check to `experimental/install`, so the OS matrix covers macOS, Linux, and Windows.

## Validation

- `pnpm --filter @ttsc/lint build`
- `git diff --check`
- `pnpm format`

Longer e2e/experimental validation is intentionally left to CI.

### DIFF
--- a/.github/workflows/nestia.yml
+++ b/.github/workflows/nestia.yml
@@ -60,26 +60,12 @@ jobs:
           if-no-files-found: error
           retention-days: 1
 
-  # Mirror nestia (next)'s own test.yml matrix so each suite runs in parallel,
-  # but against the local ttsc tarballs instead of the published release.
+  # Run nestia (next)'s standard test suite against the local ttsc tarballs
+  # instead of the published release.
   nestia:
     needs: tarballs
-    name: ${{ matrix.name }}
     runs-on: ubuntu-latest
     timeout-minutes: 60
-    strategy:
-      fail-fast: false
-      matrix:
-        include:
-          - { name: go, run: "pnpm test:go" }
-          - { name: sdk, run: "pnpm --filter ./tests/test-sdk start" }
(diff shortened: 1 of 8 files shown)


samchon:
Self-review update while CI is running:

- Found and fixed a static file-URL import risk in the generated ttsx loader. A literal `await import("file://...")` is statically resolved by tsgo and can fail with TS2307, so the URL now lives in `configUrl` before dynamic import.
- Found and fixed a Go-side UNC path risk in `fileURL(location)`. Windows drive, UNC, extended drive, and extended UNC forms are now covered by `packages/lint/test/config/file_url_matches_node_shape_test.go`.
- Found and fixed a test-helper safety issue introduced by the custom lint `projectRoot` option. It now only accepts disposable roots under the OS temp dir, including realpath aliases such as `/var` -> `/private/var`.

Local verification after the fixes:

- `pnpm --filter @ttsc/lint build`
- `node scripts/test-go-lint.cjs`
- `TTSC_TEST_DIR=features/config pnpm --filter @ttsc/test-lint start -- --include=realpath_differs`
- `git diff --check origin/master...HEAD`

Round 5 self-review did not find another actionable issue. CI is still pending.

PR #351, opened 2026-07-10
fix(graph): refresh resident index after source changes

samchon:
## Summary

- replace the MCP server's one-time `ttscgraph dump` cache with a resident native snapshot session checked before every non-escape graph call
- apply content-only source edits through TypeScript-Go's incremental `UpdateProgram`, refresh the checker lease, and reuse the in-memory graph when the project is unchanged
- reload safely for config, root-set, deletion, project-reference, module-resolution, and AST-mutating linked-plugin changes; return an error instead of stale graph evidence when refresh fails
- document the freshness contract and add native plus same-process MCP regression coverage

## Correctness and cost model

Freshness is derived from project inputs rather than a watcher, cooldown, fixture name, or prompt/tool heuristic. The native session hashes config chains and source bytes, compares transitive project root sets, and tracks structured TypeScript module specifiers plus `paths`, `baseUrl`, `rootDirs`, package `exports`/`imports`, and triple-slash references.

Unchanged calls retain both the compiler Program and `TtscGraphMemory`. Content-only edits reuse unchanged TypeScript ASTs through `driver.Session`. A changed compiler snapshot rebuilds the graph projection intentionally because a declaration edit can change checker-resolved edges originating in other files; structural changes take a full compiler reload.

This is a backward-compatible bug fix and fits a `0.18.4` patch release.

## Test plan

- `pnpm format`
- `go test ./cmd/ttscgraph ./test/driver -count=1`
- `go vet ./cmd/ttscgraph ./driver`
- `node scripts/test-go-graph.cjs`
- `$env:TTSC_PLATFORM_BUILD_TARGETS='ttscgraph'; pnpm --dir packages/ttsc-win32-x64 build`
- `pnpm --filter @ttsc/test-graph start`
- `pnpm test:typecheck`
- `pnpm --filter @ttsc/website build:next`

`pnpm build:current` was also attempted, but the existing `@ttsc/unplugin` Rollup stage fails in `rollup-plugin-node-externals` with `TypeError: undefined is not a function`. The graph package and native `ttscgraph` target were built independently and passed.

### DIFF
--- a/packages/graph/README.md
+++ b/packages/graph/README.md
@@ -47,6 +47,7 @@ The whole MCP surface is one tool, `inspect_typescript_graph`. You ask in plain 
 
 - **An index, not source bodies.** It returns names, edges, signatures, and spans, never code, so the response stays flat as the repository grows.
 - **Built on the real compiler.** It reads the program `ttsc` type-checked, so `tsconfig` aliases, pnpm monorepos, symlinks, and re-exports resolve exactly, where a text parser can only guess.
+- **Current within one agent session.** Before each graph operation it checks the config, root-file set, module-resolution inputs, and source contents. Unchanged calls reuse the warm graph; source edits update the resident compiler incrementally; config, file-addition, deletion, and resolution changes reload safely.
 - **It does not force itself.** It states when the graph is the right source and offers a first-class `escape` for everything else.
 - **Errors and lint too.** `tsc` compile errors and `@ttsc/lint` and plugin (typia, nestia) findings ride the same graph, so "what is broken here?" answers from one index.
 
--- a/packages/graph/src/TtscGraphApplication.ts
+++ b/packages/graph/src/TtscGraphApplication.ts
@@ -9,7 +9,9 @@ import { runTrace } from "./server/runTrace";
 import { ITtscGraphApplication } from "./structures/ITtscGraphApplication";
 import { ITtscGraphEscape } from "./structures/ITtscGraphEscape";
 
-export type TtscGraphSource = TtscGraphMemory | (() => TtscGraphMemory);
+export type TtscGraphSource =
+  | TtscGraphMemory
+  | (() => TtscGraphMemory | Promise<TtscGraphMemory>);
 
 /**
  * The MCP tool surface as a plain class over the resident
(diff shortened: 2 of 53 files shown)


samchon:
Self-review rounds on this branch found and sealed five freshness/lifecycle defects, each locked red/green:

1. **`captureDiskStates` failed the whole session on OS-unreadable candidate paths** — Windows `ERROR_INVALID_NAME` (`./style.css?inline`, `data:` URLs) and Linux `ENOTDIR` (candidates traversing through a file, e.g. `types.d.ts/index.d.mts`). The Linux shape was the red `go` / `ttsc native cli` lanes on the original push. Candidates that are neither readable files nor directories are now recorded as absent. (4a259aa0)
2. **`TtscGraphSession` could respawn the native process after `close()`**, orphaning a resident compiler beyond the MCP server's life. Closed sessions now reject queued requests instead. (4a259aa0)
3. **A source deleted between compiler load and state capture was invisible forever** — no root change, no hash, no candidate. The vanished path's resident text is now hashed so the next snapshot observes the deletion and reloads. (96b22b43)
4. **A `files`-listed root that does not exist yet was never picked up on creation** — literal entries survive root-set reloads unchanged. Missing config roots are now tracked as freshness inputs. (96b22b43)
5. **`compilerOptions.types` packages appearing later were invisible** when nothing else (package.json, lockfile) changed — e.g. generated typeRoots packages. Config `types` entries now contribute the same candidates as triple-slash directives. (96b22b43)

Also: reindented the new Go tests to the repo's 2-space gofmt convention (35ad1b93), pinned two stability twins — missing-literal-root stays `unchanged` (186cb667) and deleted-config fails closed without poisoning the resident session (26b21a47) — and merged master after #350 landed on the same `driver/plugins.go` (99611420).

Known boundary (intentionally out of scope): creating a file that *shadows* an already-resolved module at a higher resolution priority is not detected — tsgo's `ResolvedModule` does not expose failed lookup locations, and realistic occurrences touch package.json/lockfile/config, which are tracked.

Local verification on Windows: full `cmd/ttscgraph` + `test/driver` Go suites, all nine `test-graph` e2e features via the scratch runner, `test-graph` typecheck, plus live probes for rapid successive edits and tsconfig delete/restore recovery.

https://claude.ai/code/session_01FX9pywL7oykBcHoeUQGJ7j

PR #352, opened 2026-07-10
fix(compiler): preserve diagnostics across plugin failures

samchon:
## Summary
- preserve failed plugin setup/check/transform output and status while independently collecting pure TypeScript diagnostics
- append only TypeScript diagnostics the failed plugin did not already report, including partial-report-before-crash cases
- isolate recoverable `@ttsc/lint` contributor metadata/check panics so healthy contributor and built-in rules continue

## Scope
- plugin discovery, descriptor/source build, check sidecar, and transform/build sidecar failures recover a separate no-emit TypeScript check
- both emitting and no-emit compiler paths retain the original plugin failure status
- contributor `Name`, `Visits`, `IsFormat`, and `VisitsDeclarationFiles` are cached behind a panic barrier; `Check` panics use the existing per-rule diagnostic barrier

## In-process boundary
- contributor `init()` or contributor-started goroutine panics, `os.Exit`, fatal runtime faults, and non-returning rules cannot be recovered in-process and require contributor process isolation

## Test plan
- GitHub Actions CI only, per request; no local tests after the request

### DIFF
--- a/packages/lint/linthost/contrib_adapter.go
+++ b/packages/lint/linthost/contrib_adapter.go
@@ -61,13 +61,7 @@ func registerContributors() {
         name)
       continue
     }
-    adapter := contributorAdapter{
-      inner:                  contributor.inner,
-      metadataCached:         true,
-      name:                   contributor.name,
-      visits:                 contributor.visits,
-      visitsDeclarationFiles: contributor.visitsDeclarationFiles,
-    }
+    adapter := newContributorAdapter(contributor)
     if contributor.isFormat {
       Register(formatContributorAdapter{adapter})
       continue
@@ -114,6 +108,19 @@ func inspectContributor(contributor rule.Rule) (metadata contributorMetadata, er
   return metadata, nil
 }
 
+// newContributorAdapter builds the engine-facing adapter from metadata already
+// evaluated by inspectContributor, so no adapter method re-enters contributor
+// code after startup. This is the only construction path; a zero-value
+// contributorAdapter carries no usable metadata.
+func newContributorAdapter(metadata contributorMetadata) contributorAdapter {
+  return contributorAdapter{
+    inner:                  metadata.inner,
+    name:                   metadata.name,
+    visits:                 metadata.visits,
+    visitsDeclarationFiles: metadata.visitsDeclarationFiles,
+  }
+}
+
 // contributorAdapter wraps a public `rule.Rule` so the engine's
 // `Register` accepts it. Name, Visits, and declaration-file policy are
 // cached by inspectContributor; Check bridges through a `rule.Context`
@@ -123,7 +130,6 @@ func inspectContributor(contributor rule.Rule) (metadata contributorMetadata, er
 // same shim AST types, so no wrapping / unwrapping of nodes is needed.
 type contributorAdapter struct {
   inner                  rule.Rule
-  metadataCached         bool
   name                   string
   visits                 []shimast.Kind
   visitsDeclarationFiles bool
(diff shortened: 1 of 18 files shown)


samchon:
## Handoff for the next session

Current head: `1ce0d100bcd4921dba7b5809f5cdc9070d7c48de` on `fix/plugin-failure-typescript-diagnostics`.

### Goal

1. A ttsc plugin setup/check/transform failure must not hide unrelated pure TypeScript diagnostics.
2. A recoverable third-party `@ttsc/lint` contributor panic must not terminate lint or prevent healthy contributor/built-in findings.

### Implemented

- `runBuild.ts`
  - converts plugin discovery/descriptor/source-build errors into a failed build result;
  - preserves check/transform sidecar stderr and original non-zero status;
  - runs an independent `tsgo --noEmit` pass after eligible plugin failures;
  - appends only TypeScript diagnostics not already reported, including partial-report-before-crash cases;
  - covers emitting and explicit no-emit paths.
- `TtscCompiler.compile()` routes plugin discovery failures through the recoverable plugin-backed path and returns recovered TypeScript diagnostics.
- `@ttsc/lint`
  - evaluates contributor `Name`, `Visits`, `IsFormat`, and `VisitsDeclarationFiles` once behind a recover barrier;
  - drops only the broken metadata entry;
  - existing `runRuleCheck` recovery converts synchronous `Check` panics into rule findings while other rules continue.
- Tests/docs cover check and transform process failures, plugin source-build failure, API compile recovery, partial diagnostic deduplication, lint metadata panic, lint Check panic, and healthy-rule continuation.

### CI history and latest fix

On previous head `b4ecc095`, GitHub Actions failed only at `windows-go`:

```
--- FAIL: TestContributorRulesVisitDeclarationFilesByDefault
contributor rule without the marker must keep visiting declaration files
```

Root cause: existing tests directly construct `contributorAdapter{inner: ...}`; the new cached `visitsDeclarationFiles` bool had a false zero value and broke the historical default=true behavior.

Fixed in `1ce0d100` by adding `metadataCached`:
- production adapters set it to true and use recovered cached metadata;
- directly constructed/legacy adapters fall back to the original `inner` metadata behavior, preserving both default=true and explicit opt-out=false.

A fresh CI run should have started for `1ce0d100`. No local tests were run after the user's instruction; GitHub Actions is the authority.

### Next actions

1. Run `gh pr checks 352` and wait for the new head's checks to settle.
2. If a GitHub Actions check fails, inspect its job log and make only the focused fix; commit and push to this branch.
3. Pay particular attention to `windows-go` / `go`, `typecheck`, `lint native config`, `ttsc native corpus-misc`, `ttsc native corpus-source`, `ttsc native corpus-ttsc`, and the compiler API test lane.
4. Do not merge without an explicit user request.

### Explicit boundary

In-process recovery handles synchronous metadata and `Check` panics. A contributor `init()` panic, panic in a contributor-started goroutine, `os.Exit`, fatal runtime fault, or non-returning rule requires contributor process isolation and is intentionally documented as outside this PR's in-process guarantee.

samchon:
## Merge record

Merged as `d3d35c52` (squash) with all 55 checks green on head `3aaf54f8`.

Follow-ups added in this session beyond the handoff state:
- `94756caf` — API `compile()` now seeds the plugin failure as a `TTSC_PROCESS` diagnostic before appending recovered TypeScript diagnostics, so `IFailure.diagnostics` carries both the plugin error and the recovered type errors (previously the recovery displaced the plugin failure in the structured result). `createProcessDiagnostic` moved to `runBuild.ts` and is shared with the result mapping; docs updated (`api.mdx`, `lint.mdx`).
- `3aaf54f8` — removed the `metadataCached` dual-mode from `contributorAdapter`: `newContributorAdapter(metadata)` is now the only construction path and the three direct-construction Go tests build through `inspectContributor` + the constructor, pinning the policy where it lives. The pinned policies themselves (declaration-file default true, opt-out marker honored, checker required) are unchanged.

PR #370, opened 2026-07-10
fix(lint): exempt backslash-digit escapes from no-useless-escape in strings and templates

### LINKED ISSUE #361: @ttsc/lint: no-useless-escape autofix deletes the backslash of octal-shaped escapes \1–\7, silently changing string values (code contradicts its own exemption comment)
## Summary

The `no-useless-escape` autofix deletes the backslash of octal-shaped escapes `\1`…`\7` in string/template literals, **changing the runtime string value**. The code's own comment says digits 1–9 are exempt ("`no-octal-escape` owns those") — but the exemption is not implemented, and upstream ESLint's `no-useless-escape` skips backslash-digit entirely.

Verified in source at `b5ead8f8` (v0.18.4).

## Exact input → output

Script-mode (non-strict) input:

```js
var s = "a\1b";   // cooked value: "ab"
```

Fix output:

```js
var s = "a1b";    // cooked value: "a1b" — different string, silently
```

## Mechanism

`packages/lint/linthost/rules_escape.go`:

- `:69` — `const stringValidEscapes = "'\"\\bfnrtv0xuU\n\r"` — contains `0` but none of `1`–`9`.
- `isUselessStringEscape` (`:207-220`) — returns `false` only for control chars (`< 0x20`) and whitelist members; everything else, **including `1`–`9`**, returns `true` (reported + fixed by backslash deletion at `:130-140`).
- The comment immediately above the final `return true` (`:215-218`) states the intended policy the code does not implement: *"Digits 1-9 are octal-shaped (`no-octal-escape` owns those). `0` is in the whitelist."*

Upstream oracle: ESLint `no-useless-escape` treats `\\[0-9]` as never-useless (octal/decimal escapes are semantically meaningful or syntax errors — either way not "useless").

## Blast radius

- In module/strict code, `\1` in a *string* is itself a grammar error, so the corrupting fix mostly bites script-mode/`allowJs` files — but those are exactly the files `ttsc format` sweeps in bulk, where a silent value change is least likely to be noticed.
- `\8`/`\9` deletion happens too; for those the cooked value is coincidentally preserved (`"\8"` === `"8"`), so only `\1`–`\7` corrupt — a subtle split worth pinning in tests.

## Fix

In `isUselessStringEscape`, return `false` for `ch >= '1' && ch <= '9'` (aligning the code with its own comment and with upstream). Decide `\8`/`\9` explicitly: upstream leaves them alone as well — recommend the full `1`–`9` exemption for oracle parity.

## Regression tests (per the coverage doctrine: transformation direction + negative twin + boundary)

- `"a\1b"` and `"a\7b"` → **no report, no fix** (fails today).
- `"\0"` → still allowed (whitelisted) — boundary between `\0` and `\1`.
- `"\8"` → no report (oracle parity), documenting the value-preserving-but-still-exempt case.
- Negative twins that must keep firing: `"\m"`, `"\."` → fixed to `"m"`, `"."` as today.
- Template-literal twin for `` `\1` `` via `templateValidEscapes` (`:70`) — same defect, same fix.

## Environment

ttsc 0.18.4 (`b5ead8f8`), static verification (fixer output fully determined by the quoted code).

samchon:
## Intent

`no-useless-escape` reported octal-shaped escapes `\1`-`\9` in string and template literals, and its autofix deleted the backslash — silently changing the cooked string value for `\1`-`\7` (`"a\1b"` → `"a1b"`). The code's own comment in `isUselessStringEscape` claimed digits 1-9 were exempt ("`no-octal-escape` owns those"), but the exemption was never implemented. Upstream ESLint's `no-useless-escape` skips backslash-digit entirely.

## Scope

- `packages/lint/linthost/rules_escape.go` — `isUselessStringEscape` now returns `false` for `'1'`-`'9'`, implementing the documented exemption for both consumers (string whitelist and template whitelist). `\8`/`\9` are exempt too: their deletion happens to preserve the cooked value, but upstream skips all backslash-digit escapes, so the full range is exempted for oracle parity. `\0` remains covered by the whitelist.
- **Regex path audited, already correct**: `isUselessRegexEscape` exempts `'1'`-`'9'` as decimal back-references in both class and non-class contexts (the `switch` in `rules_escape.go`), matching upstream. No change needed there; the new skip test pins it so a future cleanup cannot reintroduce the same class of corruption.
- **Docs**: the `no-useless-escape` entry in `website/src/content/docs/lint/rules/core.mdx` describes only the general rule (`"\."`, `/\,/` examples) and never documented the buggy digit behavior, so no doc change is required.

## Test plan

Go unit tests in `packages/lint/test/fix/`, mirroring the existing `fix_no_useless_escape_*` peers:

- `fix_no_useless_escape_skips_backslash_digit_escapes_test.go` — zero findings (and therefore no fix) for:
  - `"a\1b"`, `"a\7b"` — value-changing octals (fails before this fix);
  - `"\0"` and `"\01"` — whitelist boundary, including `\0` followed by a digit which the lexer reads as one octal escape;
  - `"\8"`, `"\9"` — value-preserving but exempt, pinning the oracle-parity decision;
  - `` `a\1b` `` and a `TemplateTail` `\7` — template-literal twin through `templateValidEscapes`;
  - `/(a)\1/` — regex back-reference, pinning the pre-existing regex-path exemption.
- `fix_no_useless_escape_drops_nondigit_neighbors_of_octal_exemption_test.go` — negative twins one property away from the exemption: `"\m"` and `"\."` are still reported and autofixed to `"m"` / `"."`.

CI runs the full Go and TypeScript suites.

Closes #361

### DIFF
--- a/packages/lint/linthost/rules_escape.go
+++ b/packages/lint/linthost/rules_escape.go
@@ -212,10 +212,16 @@ func isUselessStringEscape(ch byte, whitelist string) bool {
   if strings.IndexByte(whitelist, ch) >= 0 {
     return false
   }
+  // Digits 1-9 are octal-shaped (`no-octal-escape` owns those; `0` is in
+  // the whitelist). Deleting the backslash of `\1`…`\7` changes the cooked
+  // string value, and upstream ESLint skips backslash-digit entirely, so
+  // `\8`/`\9` are exempt too for oracle parity.
+  if ch >= '1' && ch <= '9' {
+    return false
+  }
   // ASCII letters that aren't in the whitelist are user-error escapes
-  // like `\a`, `\m`. Digits 1-9 are octal-shaped (`no-octal-escape`
-  // owns those). `0` is in the whitelist. Punctuation that isn't in the
-  // whitelist (e.g., `\.`) is also redundant.
+  // like `\a`, `\m`. Punctuation that isn't in the whitelist (e.g., `\.`)
+  // is also redundant.
   return true
 }
 
--- a/packages/lint/test/fix/fix_no_useless_escape_drops_nondigit_neighbors_of_octal_exemption_test.go
+++ b/packages/lint/test/fix/fix_no_useless_escape_drops_nondigit_neighbors_of_octal_exemption_test.go
@@ -1,0 +1,24 @@
+package linthost
+
+import "testing"
+
+// TestFixNoUselessEscapeDropsNondigitNeighborsOfOctalExemption verifies
+// the negative twins of the backslash-digit exemption still autofix.
+//
+// The `\1`…`\9` exemption in `isUselessStringEscape` (issue #361) must
+// not over-reach: a useless letter escape (`\m`) and a useless
+// punctuation escape (`\.`) sit one property away from the exempted
+// digits and must keep being reported and fixed, otherwise the exemption
+// silently disables the rule for whole character classes.
+//
+// 1. Parse string literals containing `\m` and `\.` (no meaningful escape).
+// 2. Apply the findings through the disk-backed fixer.
+// 3. Assert both backslashes are gone and nothing else changed.
+func TestFixNoUselessEscapeDropsNondigitNeighborsOfOctalExemption(t *testing.T) {
+  assertFixSnapshot(
+    t,
+    "no-useless-escape",
+    "const letter = \"a\\mb\";\nconst dot = \"a\\.b\";\nJSON.stringify({letter,dot});\n",
+    "const letter = \"amb\";\nconst dot = \"a.b\";\nJSON.stringify({letter,dot});\n",
+  )
+}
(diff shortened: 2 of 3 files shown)


samchon:
Self-review follow-ups pushed:

- a7fc96e5 — reworded the skip test's doc comment to remove mid-word line breaks ("script/-mode", "value/-changing").
- 7405a006 — added `"a\:b"` to the negative-twin fixture: `':' == '9'+1`, so an off-by-one in the new range comparison (`ch <= ':'`) would wrongly exempt it. This pins the exact ASCII boundary of the exemption; the lower boundary needs no twin because `'0'` is already whitelisted, making `ch >= '0'` observationally identical.

Also audited `regexp/no-useless-escape` (`rules_regexp.go`): it delegates to the same `reportRegexEscapes`/`isUselessRegexEscape` path, which already exempts `\1`-`\9` as back-references — no sibling defect.

PR #374, opened 2026-07-10
fix(factory): drop trailing comma after rest and trailing elision in broken lists

### LINKED ISSUE #356: @ttsc/factory: broken parameter lists and binding patterns emit a trailing comma after a rest element — syntax error on any >printWidth signature ending in `...rest`
## Summary

`TsPrinter` emits an unconditional trailing comma whenever a parameter list or binding pattern breaks across lines. When the last element is a **rest element** (`...rest`), the output is a syntax error in every JavaScript/TypeScript parser:

- parameters: TS1013 *"A rest parameter or binding pattern may not have a trailing comma"* / V8 `SyntaxError: Rest parameter must be last formal parameter`;
- destructuring: V8 `SyntaxError: Rest element must be last element` (both messages verified in Node v22.15.0).

Breaking at `printWidth` (default 80) is the printer's normal behavior and codegen for wrapper/proxy functions routinely ends the signature with a rest parameter, so this fires on realistic output.

Verified live on `@ttsc/factory` at `b5ead8f8` (v0.18.4).

## Repro

```ts
import factory, { NodeFlags, SyntaxKind, TsPrinter } from "@ttsc/factory";

const id = factory.createIdentifier;
const param = (name: string, type: string, rest = false) =>
  factory.createParameterDeclaration(
    undefined,
    rest ? factory.createToken(SyntaxKind.DotDotDotToken) : undefined,
    id(name), undefined,
    factory.createTypeReferenceNode(id(type), undefined), undefined,
  );

new TsPrinter().print(
  factory.createFunctionDeclaration(undefined, undefined, id("assemble"), undefined, [
    param("firstParameter", "SomeVeryLongTypeName"),
    param("secondParameter", "AnotherLongTypeName"),
    param("rest", "ReadonlyStringList", true),
  ], undefined, factory.createBlock([], false)),
);
```

Observed output — does not parse:

```ts
function assemble(
  firstParameter: SomeVeryLongTypeName,
  secondParameter: AnotherLongTypeName,
  ...rest: ReadonlyStringList,
) {}
```

Same defect through the binding-pattern path — a broken array (or object) destructuring with a rest element:

```ts
const [
  extremelyLongFirstElementName,
  anotherExtremelyLongElementName,
  ...restOfTheElements,
] = values;
```

Negative twin, verified correct today: a short signature with a rest parameter never breaks, so `function f(a: A, ...rest: B) {}` prints fine — which is why happy-path tests never caught it.

## Mechanism

At `b5ead8f8`, `packages/factory/src/TsPrinter.ts`:

- `params()` (`:200-209`) calls `this.delim("(", …, ")", { trailingComma: true })` unconditionally.
- `ObjectBindingPattern` / `ArrayBindingPattern` cases (`:1235-1248`) pass `trailingComma: true` the same way.
- `delim` (`:113-135`) appends `ifBreak(",")` after the last element whenever `trailingComma` is set, with no inspection of the element.

None of these sites checks whether the final element carries `dotDotDotToken`. Prettier — whose doc-IR conventions this printer follows — has exactly this guard (`shouldPrintComma` returns false for rest elements), and the legacy TS emitter never prints a trailing comma after a rest element.

## Proposed fix

At the three call sites (`params`, `ObjectBindingPattern`, `ArrayBindingPattern`), compute the flag instead of hardcoding it:

```ts
trailingComma: nodes.length === 0 || lastOf(nodes).dotDotDotToken === undefined
```

(For binding patterns the last element may also be an `OmittedExpression` — an elision like `const [a, , ] = xs` — where a trailing comma after a hole changes arity; worth handling in the same pass: suppress the synthetic trailing comma after a trailing `OmittedExpression` too, since `[a, ,]` ≠ `[a]`.)

`args()` and other `delim` consumers that can never contain rest/elision are unaffected.

## Regression tests (`tests/test-factory/src/features/`, one case per file)

- Broken (>80-col) function signature ending in a rest parameter → no trailing comma; **negative twin**: same signature with a non-rest last parameter keeps the trailing comma.
- Broken array binding pattern with rest; broken object binding pattern with rest.
- Short rest signature stays on one line, byte-identical to today.
- (If elision handling is included) broken array binding ending in an `OmittedExpression` — assert emitted arity survives a parse round-trip.

## Environment

`@ttsc/factory` 0.18.4 (`b5ead8f8`) · Node v22.15.0 · Windows 11 (mechanism platform-independent).

samchon:
## Intent

`TsPrinter` emits its width-break trailing comma unconditionally, so any parameter list or binding pattern that breaks across lines and ends in a rest element prints a syntax error in every JS/TS parser (TS1013 / V8 `Rest parameter must be last formal parameter` / `Rest element must be last element`). Verified live against `b5ead8f8`.

## Scope

`packages/factory/src/TsPrinter.ts` only, confined to the `params()` / `ObjectBindingPattern` / `ArrayBindingPattern` call sites of `delim`:

- New private `listTrailingComma(nodes)` computes the flag from the last element generally: suppressed when it carries a `dotDotDotToken` (rest), and after a trailing `OmittedExpression`, where the synthetic comma is not cosmetic — the flat and broken layouts of the same node would parse to different arities (`[a, ,]` vs `[a, ]`).
- The three call sites pass the computed flag instead of a hardcoded `true`.

Every other `delim` consumer with `trailingComma: true` was audited and left untouched, verified against tsc: a trailing comma after a spread is legal in call arguments and array / object literals, legal after a rest element in tuple types (`[string, ...number[],]` parses clean), and `NamedImports` / `NamedExports` cannot contain rest. Type-argument lists already pass `trailingComma: false` (TS1009).

## Test plan

Five regression cases in `tests/test-factory/src/features/width/`, one per file:

- `test_params_break_rest_no_trailing_comma` — broken signature ending in `...rest` drops the comma; negative twin with a plain last parameter keeps it.
- `test_array_binding_break_rest_no_trailing_comma` — broken `const [...] = values;` with rest drops the comma; non-rest twin keeps it.
- `test_object_binding_break_rest_no_trailing_comma` — same for object binding patterns.
- `test_params_rest_short_stays_flat` — short rest signature stays on one line, byte-identical to today.
- `test_array_binding_break_trailing_elision` — trailing-hole pattern transpiles clean in both layouts and parses back to identical arity (flat vs broken agree).

All five verified failing-before / passing-after against the factory source; nearest peer suites (width, binding patterns, parameter variants, destructuring) spot-checked green.

Closes #356

### DIFF
--- a/packages/factory/src/TsPrinter.ts
+++ b/packages/factory/src/TsPrinter.ts
@@ -197,13 +197,31 @@ export class TsPrinter {
       : "";
   }
 
+  /**
+   * Whether a broken parameter list / binding pattern may append a synthetic
+   * trailing comma after its last element.
+   *
+   * A trailing comma after a rest element (`...rest`) is a syntax error (TS1013
+   * / V8 `SyntaxError`), and one after a trailing elision (`OmittedExpression`)
+   * is not cosmetic: `[a, ,]` parses to one more hole than `[a, ]`, so the flat
+   * and broken layouts of the same node would disagree. Call arguments and
+   * array / object literals are unaffected — a trailing comma after a spread is
+   * legal there.
+   */
+  private listTrailingComma(nodes: readonly Node[]): boolean {
+    const last: Node | undefined = nodes[nodes.length - 1];
+    if (last === undefined) return true;
+    if (last.kind === "OmittedExpression") return false;
+    return !("dotDotDotToken" in last && last.dotDotDotToken !== undefined);
+  }
+
   private params(params: readonly Node[]): Doc {
     return this.delim(
       "(",
       params.map((p) => this.emit(p)),
       ")",
       {
-        trailingComma: true,
+        trailingComma: this.listTrailingComma(params),
       },
     );
   }
@@ -1237,14 +1255,17 @@ export class TsPrinter {
           "{",
           node.elements.map((e) => this.emit(e)),
           "}",
-          { space: true, trailingComma: true },
+          {
+            space: true,
+            trailingComma: this.listTrailingComma(node.elements),
+          },
         );
       case "ArrayBindingPattern":
         return this.delim(
           "[",
           node.elements.map((e) => this.emit(e)),
           "]",
-          { trailingComma: true },
+          { trailingComma: this.listTrailingComma(node.elements) },
         );
       case "TypeAssertion":
         return concat([
--- a/tests/test-factory/src/features/width/test_array_binding_break_rest_no_trailing_comma.ts
+++ b/tests/test-factory/src/features/width/test_array_binding_break_rest_no_trailing_comma.ts
@@ -1,0 +1,59 @@
+import { TestValidator } from "@nestia/e2e";
+import factory, { NodeFlags, SyntaxKind, TsPrinter } from "@ttsc/factory";
+
+import { id } from "../../internal/helpers";
+
+/**
+ * Verifies a broken array binding pattern ending in a rest element drops the
+ * synthetic trailing comma.
+ *
+ * Pins `TsPrinter.listTrailingComma` at the `ArrayBindingPattern` call site: a
+ * trailing comma after a rest element is a syntax error (TS1013 / V8 "Rest
+ * element must be last element"). The non-rest twin in the same pattern shape
+ * keeps its trailing comma, so the suppression is provably scoped to the rest
+ * case.
+ *
+ * 1. Print a wider-than-`printWidth` `const [...] = values;` whose last element is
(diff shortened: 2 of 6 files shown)


samchon:
CI note: the "Ubuntu" failure on bcad72d7 was `test_random_recursive_terminates` in the "Test typia next" step — an upstream suite the workflow clones fresh from samchon/typia HEAD. Run 1 passed with the identical product diff and the same upstream commit (b9d256c1); the failing predicate bounds a `typia.random` recursion depth, which is probabilistic. Unrelated to this PR; reran the failed job.

PR #377, opened 2026-07-11
fix(lint): keep non-string sub-chains whole in prefer-template autofix

### LINKED ISSUE #363: @ttsc/lint: prefer-template autofix flattens numeric sub-chains into separate interpolations — `a + b + " items"` silently changes runtime value; adjacent `$`+`{` segments fuse into a live interpolation
## Summary

The `prefer-template` autofix flattens **every** `+` chain into individual `${…}` slots, including purely numeric sub-chains that evaluate before the string concatenation. `a + b + " items"` (left-associative: `(a + b) + " items"`) becomes `` `${a}${b} items` `` — with `a = 1, b = 2` the original evaluates `"3 items"`, the fixed code `"12 items"`. A **silent runtime value change**. Upstream ESLint keeps the non-string sub-chain as one interpolation: `` `${a + b} items` ``.

A second, independent defect in the same fixer: adjacent cooked segments are escaped individually, so `"$" + "{" + n` fuses into a live `${` interpolation opener in the rendered template — a parse error (or worse, evaluated code if the following text happens to form an expression).

Verified in source at `b5ead8f8` (v0.18.4).

## Exact input → output

```js
const s = a + b + " items";
// today   → `${a}${b} items`      // "12 items" for a=1,b=2 — was "3 items"
// upstream → `${a + b} items`     // value preserved

const t = "$" + "{" + n;
// today   → segments "$" and "{" concatenate into `${` → template parses an interpolation → SyntaxError / wrong program
```

Parenthesized sub-chains survive (`(a + b) + " items"` stays one slot), so only natural left-associative chains — the most common way to write this — corrupt. Existing fix tests cover single-identifier chains only.

## Mechanism

`packages/lint/linthost/rules_suggestions.go`:

- `flattenConcatOperands` (`:1205-1218`) recurses into **every** `KindBinaryExpression` with `KindPlusToken`, regardless of whether that subtree contains a string literal: `out := flattenConcatOperands(bin.Left); out = append(out, flattenConcatOperands(bin.Right)...)`.
- The fix path (`:1141-1180`) renders each flattened leaf as its own literal segment or `${…}` slot via `renderConcatAsTemplate` (`:1226-1252`).
- `escapeTemplateLiteralBody` (`:1265-1291`) escapes `${` only **within one literal's text**; it cannot see that the previous segment ended with `$`, hence the seam fusion.

Upstream oracle: ESLint `prefer-template`'s fixer only descends into a `+` node when `hasStringLiteral(side) && hasNonStringLiteral(otherSide)` reasoning applies — a sub-chain with no string literal is embedded as a single expression, preserving arithmetic.

## Fix

1. In `flattenConcatOperands`, recurse into a `+` subtree only when that subtree **contains a string-like literal** (string literal, template, or another qualifying concat); otherwise return the subtree as one leaf → it renders as a single `${a + b}` slot.
2. In the segment renderer, escape a literal segment's trailing `$` when the next emitted byte is `{` (or escape `$` before `{` across the concatenated output once, after joining).

## Regression tests

- `a + b + " items"` → `` `${a + b} items` `` (fails today) — plus an evaluation-equality harness twin if the suite supports it.
- `"count: " + a + b` → `` `count: ${a}${b}` `` — string first, flattening IS correct here (negative twin pinning that the fix above doesn't over-correct).
- `"$" + "{" + n` → output must re-parse; cooked value preserved.
- All-literal chain `"a" + "b"` → `` `ab` `` unchanged behavior.

## Environment

ttsc 0.18.4 (`b5ead8f8`), static verification (fixer output fully determined by the quoted code).

samchon:
## Intent

Fix two independent value-corruption defects in the `prefer-template` autofix (#363):

1. **Numeric sub-chain flattening.** `flattenConcatOperands` recursed into every `+` subtree, so the left-associative `a + b + " items"` became `` `${a}${b} items` ``. Since `(a + b)` evaluates before the string concatenation, the rewrite silently changed the runtime value (with `a = 1, b = 2`: `"3 items"` → `"12 items"`).
2. **`$`+`{` seam fusion.** Each literal segment was escaped in isolation, so `"$" + "{" + n` joined into a live `${` interpolation opener — the fixed output didn't even re-parse.

## Scope

`packages/lint/linthost/rules_suggestions.go` only:

- New containment predicate `concatChainContainsString`: a `+` subtree is flattened only when it contains a string-like operand (string literal, template literal with or without substitutions, or a nested chain containing one). A subtree with none stays a single `${a + b}` slot — the upstream ESLint `prefer-template` oracle behavior. This is a general containment gate, not a node-kind blacklist.
- `renderConcatAsTemplate` now merges adjacent literal operands into one cooked run before calling `escapeTemplateLiteralBody`, making the `${` lookahead total across former segment seams. The remaining seam class (literal trailing `$` before an emitted `${…}` placeholder, `$${expr}`) parses as a literal `$` plus the interpolation and preserves the original value, so it needs no escaping.

Detection behavior is unchanged; the website rule page documents detection only, so no docs change.

## Test plan

Go tests under `packages/lint/test/rules/strings-regex/`, one case per file. Verified locally via the scratch-module targeted runner (full suites run in CI):

- `a + b + " items"` → `` `${a + b} items` `` — **fails before the fix** (`` `${a}${b} items` ``), passes after.
- `"$" + "{" + n` → `` `\${${n}` `` — **fails before the fix** (live `${` fusion), passes after; cooked value validated by evaluation.
- Negative twin: `"count: " + a + b` → `` `count: ${a}${b}` `` — string-first flattening is correct and must not be over-corrected.
- `(a + b) + " s"` → `` `${(a + b)} s` `` — parenthesized sub-chain stays one slot.
- `` a + `x${y}` + "s" `` → `` `${a}${`x${y}`}s` `` — substitution template marks its sub-chain string-like (containment branch).
- `"" + a` → `` `${a}` `` — empty-literal flush boundary.
- Numeric-only chain `1 + 2 + n` and all-literal chain `"a" + "b"` — rule stays silent.
- All six pre-existing prefer-template tests (3-part chain, leading identifier, backtick escape, CR/LF preservation, corpus) still pass.

Closes #363

### DIFF
--- a/packages/lint/linthost/rules_suggestions.go
+++ b/packages/lint/linthost/rules_suggestions.go
@@ -1226,13 +1226,20 @@ func flattenConcatOperands(node *shimast.Node) []*shimast.Node {
 
 // concatChainContainsString reports whether a `+` chain (or a single
 // operand) contains a string-like operand: a string literal, a template
-// literal (with or without substitutions), or a nested `+` chain that
-// itself contains one. This is the flattening gate for
-// flattenConcatOperands. Once a string-like operand appears in a
-// left-associative chain, every later `+` is string concatenation, so
-// splitting the chain into template segments preserves the value; a
-// chain with none may be numeric addition and must stay whole.
+// literal (with or without substitutions), or a nested `+` chain —
+// parenthesized or not — that itself contains one. This is the
+// flattening gate for flattenConcatOperands. Once a string-like operand
+// appears in a left-associative chain, every later `+` is string
+// concatenation, so splitting the chain into template segments preserves
+// the value; a chain with none may be numeric addition and must stay
+// whole. Parentheses are transparent to the DECISION (`("a" + b) + c`
+// is a string chain, so `c` still gets its own slot) even though the
+// flattener keeps the parenthesized operand itself as one slot. A
+// string-like node under any other operator (e.g. `a * "x"`) does NOT
+// qualify: that subexpression coerces away from string, so its chain
+// may still be numeric addition.
 func concatChainContainsString(node *shimast.Node) bool {
+  node = stripParens(node)
   if node == nil {
     return false
   }
@@ -1242,11 +1249,7 @@ func concatChainContainsString(node *shimast.Node) bool {
       return concatChainContainsString(bin.Left) || concatChainContainsString(bin.Right)
     }
   }
-  inner := stripParens(node)
-  if inner == nil {
-    return false
-  }
-  return isStringLikeLiteral(inner) || inner.Kind == shimast.KindTemplateExpression
+  return isStringLikeLiteral(node) || node.Kind == shimast.KindTemplateExpression
 }
 
 // renderConcatAsTemplate renders the flattened concat operands as a single
--- a/packages/lint/linthost/rules_suggestions.go
+++ b/packages/lint/linthost/rules_suggestions.go
@@ -1198,17 +1198,24 @@ func concatChainShape(node *shimast.Node) (hasString bool, hasOther bool) {
 }
 
 // flattenConcatOperands walks a `+` chain left-to-right and returns each
-// leaf operand in source order. Parenthesized sub-expressions are kept as
-// a single operand so the rendered template literal does not lose their
-// grouping. Mirrors ESLint's behavior of treating `("a" + b)` as one
-// expression slot when it appears inside a larger chain.
+// leaf operand in source order. It only descends into a `+` subtree that
+// itself contains a string-like operand (see
+// concatChainContainsString): a subtree without one — the `a + b` in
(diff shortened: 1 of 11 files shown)


samchon:
Self-review round 1 finding, fixed in 6a02f185: the containment gate was opaque to parentheses, so `("a" + b) + c + " s"` demoted the whole left side to one `${("a" + b) + c}` slot (value-correct but a needless flattening regression vs. the pre-gate fixer and upstream). The predicate now strips parens at entry — the paren operand itself still renders as one slot, but its string marks the enclosing chain. Negative twin added: a string under a non-`+` operator (`a * "x" + c + " s"`) still does not qualify, since it coerces away from string. All 16 prefer-template Go tests pass in the scratch-module targeted runner.

PR #380, opened 2026-07-11
fix(unplugin): anchor utility-plugin config discovery at the project root for generated temp-dir tsconfigs

### LINKED ISSUE #358: @ttsc/unplugin: generated temp-dir tsconfig re-anchors banner/strip config discovery at the OS temp directory — strip silently ignores the project's strip.config, banner spuriously errors, relative configFile dangles
## Summary

When `@ttsc/unplugin` needs a compiler-options overlay — any bundler **alias** (`resolve.alias` in Vite) or any inline `compilerOptions` — it writes a generated tsconfig into the **system temp directory** and compiles through it. But the utility plugins anchor their config-file discovery (and relative-`configFile` resolution) at the **tsconfig's directory**, and the adapter's path normalizer absolutizes only the `config`/`source`/`transform` plugin-entry keys — not `configFile`. Result, for the alias + utility-plugin combination:

1. `@ttsc/strip` with a `strip.config.ts` at the project root: the upward walk starts at `%TEMP%/ttsc-unplugin-*/`, never reaches the project, finds nothing, and strip **silently applies its built-in defaults** — the user's configured strip list is ignored with no error, or default stripping is applied where the user configured none.
2. `@ttsc/banner` in the same setup: hard error `@ttsc/banner: no banner.config.{ts,cts,mts,js,cjs,mjs,json} file found` even though the file sits next to the real `tsconfig.json`.
3. A relative `configFile` (`{ "transform": "@ttsc/banner", "configFile": "./banner.config.json" }`) resolves against the temp directory and fails with a not-found error.
4. Corollary hazard: a stray `strip.config.*` anywhere on the walk **above the OS temp dir** would be honored for the build.

Projects with no alias and no `compilerOptions` overlay are unaffected (the original tsconfig path is passed through), which is why the common minimal setups in `tests/test-unplugin` never trip it — there is no e2e combining an alias with a utility plugin config.

All code paths verified at `b5ead8f8` (v0.18.4).

## Mechanism, fully traced

**(a) The generated tsconfig moves to temp.** `createTransformTsconfig` (`packages/unplugin/src/core/transform.ts:483-520`): when the merged overlay is non-empty it writes

```ts
const directory = fs.mkdtempSync(path.join(os.tmpdir(), "ttsc-unplugin-"));
const file = path.join(directory, "tsconfig.json");
fs.writeFileSync(file, JSON.stringify({ extends: normalizePath(props.tsconfig), compilerOptions }, ...));
```

and returns that temp path, which flows into `TtscCompiler` as the project tsconfig; when the overlay is empty it returns `props.tsconfig` unchanged (the safe path).

**(b) Only three keys are re-anchored.** `normalizePluginConfigForGeneratedTsconfig` (`transform.ts:571-586`):

```ts
for (const key of ["config", "source", "transform"]) { ... path.resolve(tsconfigDir, value) ... }
```

`configFile` — the one key `@ttsc/banner`/`@ttsc/strip`/`@ttsc/lint` actually accept — is not in the list, so a relative `configFile` survives verbatim and later resolves against the temp dir.

**(c) Discovery anchors at the tsconfig dir.** Go side receives the temp path as the plugin context's tsconfig:

- strip: `stripDiscoveryBaseDir` (`packages/strip/driver/config.go:134-143`) — "Prefers the tsconfig directory over cwd"; `findStripConfigFile` (`config.go:99-129`) walks upward from there to the volume root and returns `"" (no error)` on a miss; the caller comment (`config.go:78-79`) — "No config file found → use defaults" — makes the miss **silent**.
- banner: `tsconfigBaseDir` (`packages/banner/driver/banner.go:233-242`) anchors identically, but a miss is a **hard error** (`banner.go:136`).

So one root cause fans out into three different user-visible behaviors (silent default, spurious hard error, dangling relative path), depending on which plugin is configured.

## Repro sketch (for the regression test)

Vite project: `tsconfig.json` + `strip.config.ts` (custom `calls` list) at the root; `vite.config.ts` with any `resolve.alias` entry and the ttsc unplugin; a source file containing one call from the custom strip list. Build → the call **survives** (defaults applied, config ignored). Remove the alias → the call is stripped. The banner variant: same layout with `banner.config.ts` → build fails "no banner.config found" only when the alias is present.

## Proposed fix

The cleanest seam keeps the temp overlay (its "outside the project root" placement is deliberate for the out-of-walk cache tests) and repairs the two broken channels:

1. **Absolutize `configFile`** in `normalizePluginConfigForGeneratedTsconfig` (add it to the key list) — fixes (3) outright.
2. **Give discovery the real anchor.** Options, in order of blast radius:
   - (a) Teach the Go discovery helpers to follow the generated config's `extends` chain: when the tsconfig at hand has an `extends` pointing at a real project tsconfig, anchor discovery at the *extended* file's directory. Both plugins share the pattern, so one helper in the utility-host layer serves both.
   - (b) Or have the adapter pass the original tsconfig dir through a dedicated field on the plugin context (e.g. `discoveryDir`), defaulting to `filepath.Dir(tsconfig)` — explicit, no `extends` parsing.
   - (c) Or fall back to `Cwd` when the tsconfig directory does not contain (is not an ancestor of) the cwd — heuristic, but zero protocol change.

   (a) or (b) preferred; (c) breaks legitimate "tsconfig outside cwd" monorepo layouts.

## Regression tests

- `tests/test-unplugin`: alias + `@ttsc/strip` + root `strip.config.ts` → configured call list honored (fails today); negative twin: same project without the alias stays green.
- alias + `@ttsc/banner` + root `banner.config.ts` → no "not found" error (fails today).
- alias + relative `configFile` on a banner entry → resolves against the project (fails today).
- Guard for (4): a `strip.config.json` planted in the temp-walk path must **not** be honored once discovery is re-anchored.

## Environment

ttsc 0.18.4 (`b5ead8f8`) · analysis from source; repro sketch above is the missing e2e. Platform-independent (the temp-dir anchoring reproduces on any OS).

samchon:
## Intent

Fix samchon/ttsc#358: when `@ttsc/unplugin` needs a compiler-options overlay (any bundler alias, or inline `compilerOptions`), it compiles through a generated tsconfig written into the system temp directory. The utility plugins anchored config-file discovery and relative-`configFile` resolution at that tsconfig's directory, so for the alias + utility-plugin combination:

1. `@ttsc/strip` silently ignored the project's `strip.config.*` and applied its built-in defaults;
2. `@ttsc/banner` failed hard with `no banner.config.{ts,...,json} file found`;
3. a relative `configFile` dangled against the temp dir and failed with not-found;
4. corollary hazard: a stray `strip.config.*` planted on the walk above the OS temp dir would be honored for the build.

## Scope and chosen design

**Fix 1 (adapter):** `configFile` joins `config`/`source`/`transform` in the generated-tsconfig absolutizer (`normalizePluginConfigForGeneratedTsconfig`), so inline plugin entries never carry a relative `configFile` into the temp dir.

**Fix 2 (discovery anchor): option (b) — an explicit channel for the discovery dir.** The launcher already owns the correct value: `ITtscCompilerContext.projectRoot` is documented as the field an embedder sets "when a wrapper config lives outside the project" to "keep plugin resolution, cache paths, and native plugin `--cwd` anchored to the real project", and the unplugin already passes it (the original tsconfig's directory). This PR carries that value the last mile:

- every native plugin spawn (`runBuild.ts`, `transformProjectInMemory.ts`, `startResidentTransform.ts`) sets `TTSC_PLUGIN_CONFIG_DIR=<project root>` in the child environment;
- a shared driver helper (`driver.PluginConfigBaseDir`) resolves the anchor: explicit channel first, else tsconfig dir, else cwd. `@ttsc/banner` and `@ttsc/strip` both delegate to it; `@ttsc/lint`'s `linthost` mirrors it (that module deliberately does not depend on the in-tree ttsc Go module — see `linthost/host.go`), fixing lint's relative-`configFile` resolution and making the explicit anchor the single discovery origin (lint's old cwd fallback rescued the miss but still walked the temp ancestry first, so hazard 4 applied to it too);
- `@ttsc/lint`'s JS factory (contributor discovery) had the same defect in-process; it now anchors at the factory context's `projectRoot`, the same channel.

Why not option (a) (follow the `extends` chain): `extends` does not mean "the real project is over there". A package tsconfig extending a shared monorepo base (`../../config/tsconfig.base.json`) would re-anchor discovery at the shared config directory and shadow per-package `strip.config.*` — a regression for a common, legitimate layout. Option (c) (cwd-containment heuristic) breaks legitimate tsconfig-outside-cwd layouts and was ruled out.

Why an environment variable and not a CLI flag: check/build/transform argument lists also reach third-party native hosts, and strict hosts exit 2 on unknown flags (the #113 / ad3443a regression class). An env var is ignored by hosts that do not know it, follows the established side-band pattern (`TTSC_LINKED_PLUGINS_JSON`, `TTSC_NODE_BINARY`, ...), and reaches linked ttsc plugins even when they run inside a third-party shared host. Without the channel (direct host invocations, wasm, LSP) every existing fallback ladder is byte-for-byte unchanged.

No temp-dir path patterns, no `ttsc-unplugin-` name matching: the anchor is general for any generated/extending tsconfig, and normal projects see the identical anchor as before (project root defaults to the tsconfig's directory).

## Test plan

Go unit tests (in the owning packages' `test/` dirs):

- `packages/ttsc/test/driver/plugin_config_base_dir_precedence_test.go` — full precedence ladder of `PluginConfigBaseDir` (absolute env, cwd-relative env, blank env, relative tsconfig, empty tsconfig).
- `packages/strip/test/config_discovery_anchors_at_plugin_config_dir_env_test.go`, `config_discovery_env_ignores_config_next_to_wrapper_tsconfig_test.go` (hazard-4 guard), `config_file_resolves_relative_to_plugin_config_dir_env_test.go`.
- `packages/banner/test/config_discovery_anchors_at_plugin_config_dir_env_test.go` (with wrapper decoy), `config_file_resolves_relative_to_plugin_config_dir_env_test.go`.
- `packages/lint/test/config/files/load_rule_config_honors_plugin_config_dir_env_test.go` (with wrapper decoy), `load_rule_config_resolves_configfile_relative_to_plugin_config_dir_env_test.go`.

TS e2e (`tests/test-unplugin`, real `@ttsc/strip` / `@ttsc/banner` Go hosts through `transformTtsc`):

- alias + root `strip.config.json` honored (defaults NOT applied — fails on master);
- positive twin: alias and no-alias strip outputs byte-identical;
- alias + `banner.config.json`: no spurious not-found error, banner text emitted (fails on master);
- alias + relative `configFile` on the tsconfig entry resolves against the project (fails on master);
- hazard guard: `strip.config.json` planted in the temp-walk path is NOT honored;
- fixture-plugin check that a relative `configFile` in inline `compilerOptions.plugins` is written absolute into the generated tsconfig (fails on master).

Plus `tests/test-lint/.../test_descriptor_discovers_contributors_via_project_root_for_wrapper_tsconfig.ts` for the JS factory.

All e2e scenarios and all new Go tests were run locally (Windows) against the real pipeline; the lint factory test was additionally verified to fail on the pre-fix build and pass after.

Docs updated in the same PR: `plugins/strip.mdx`, `plugins/banner.mdx`, `lint/setup.mdx`, `development/reference/architecture.mdx` (injected env vars), and the banner/lint walkthrough discovery ladders.

Closes #358


samchon:
Round-2 self-review correction (commit 0e54ef7d): the first cut keyed the `TTSC_PLUGIN_CONFIG_DIR` channel off `projectRoot`, which re-anchored discovery for every embedder that sets `projectRoot` and reversed the semantics pinned by `test_lint_config_file_wrapper_tsconfig_outside_cwd_discovers_wrapper_config` (a user-authored monorepo wrapper tsconfig whose own `lint.config` must win over the inner project's). The channel is now a dedicated opt-in option, `ITtscCompilerContext.pluginConfigDir`, set by @ttsc/unplugin (the party that generates the wrapper) and forwarded to native plugin spawns and JS plugin factories; every caller that does not declare it keeps byte-for-byte pre-PR behavior, and the env builders scrub an inherited value so nested ttsc runs never mis-anchor. Verified locally: both pre-existing lint wrapper tests, all six new unplugin scenarios, the lint factory test, and the Go unit tests all pass on the corrected design.

samchon:
Rounds 3-4 follow-ups: d4e870d5 tightens the three user-facing docs pages (the integration declares the anchor via `TtscCompiler.pluginConfigDir`; ttsc only forwards it), and a03ad0dd adds coverage for the env-scrub branch — the fixture plugin gains an `assert-no-plugin-config-dir` operation and a new e2e proves an undeclared run never leaks an inherited `TTSC_PLUGIN_CONFIG_DIR` into its plugin spawns (verified to fail against a scrub-less build). Round 5 was a clean full pass.

PR #390, opened 2026-07-11
fix(ttsx): inline served source maps so coverage and stacks map to source (#353)

### LINKED ISSUE #353: ttsx: served emits carry a dangling external sourceMappingURL — c8/V8 coverage misattributes lines (false 100%) and stack traces point at wrong lines
## Summary

`ttsx` executes tsgo-built JavaScript **under the original `.ts` source URL**, verbatim. When the owning tsconfig has `sourceMap: true` (external maps — the common library configuration), that served text ends with a `sourceMappingURL` comment that **nothing can ever resolve**: Node resolves it relative to the script URL (the `.ts` path, where no `.js.map` exists), and even a resolvable absolute path would die at process exit because the per-run emit directory is deleted. Two user-visible defects follow, plus a third independent one in the emitted maps themselves:

1. **V8/c8 coverage is garbage — including false 100%.** Node's coverage source-map cache records the script with `lineLengths` but `data: null`, so c8 applies **transformed-JS byte offsets to the on-disk TS text**. In the minimal repro below, a function that never runs reports 100% covered.
2. **Error stack traces point at wrong lines.** Frames carry the `.ts` filename with **emitted-JS line numbers** (off by the prologue/comment shift; arbitrarily far off once a transform plugin rewrites the file).
3. **Emitted map `sources` are mis-rooted**, so even when a map IS consumed the mapped frame prints a nonexistent path (`C:\src\boom.ts` instead of the real project path).

All verified on **ttsc 0.18.4** (`b5ead8f8`), typia 13.0.2, Node v22.15.0, Windows 11. First hit in the automovie repo while trying to extend its c8 100%-coverage gate to `packages/mcp` (samchon/AutoMovie#1040): the per-line report was untrustworthy — provably-executed lines shown uncovered, never-called functions shown covered.

## Minimal repro (3 files, self-contained)

```bash
mkdir repro && cd repro && mkdir src
npm i -D typescript@7.0.2 ttsc@0.18.4 c8@10.1.3
```

**tsconfig.json**

```json
{
  "compilerOptions": {
    "module": "commonjs",
    "target": "es2022",
    "strict": true,
    "sourceMap": true,
    "outDir": "lib",
    "rootDir": "src"
  },
  "include": ["src"]
}
```

**src/lib.ts** — a tall comment so emitted lines diverge visibly from source lines:

```ts
/**
 * A tall comment block.
 *
 * tsgo's emit prologue shifts every statement's line; a transform plugin can
 * shift them arbitrarily. Any tool that applies the executed code's offsets
 * to THIS text attributes coverage to the wrong lines.
 *
 * padding
 * padding
 * padding
 * padding
 * padding
 */
export const used = (): string => {
  return "used ran";
};

export const unused = (): string => {
  return "unused never runs";
};
```

**src/main.ts**

```ts
import { used } from "./lib";

console.log(used());
```

### Run A — broken (external maps, the default library setup)

```bash
npx c8 --include "src/**" --extension .ts --reporter=text ttsx -P tsconfig.json src/main.ts
```

```
----------|---------|----------|---------|---------|-------------------
File      | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
----------|---------|----------|---------|---------|-------------------
All files |     100 |      100 |     100 |     100 |
 lib.ts   |     100 |      100 |     100 |     100 |
 main.ts  |     100 |      100 |     100 |     100 |
----------|---------|----------|---------|---------|-------------------
```

**`unused` never executed, yet everything reports 100%.** A coverage gate built on this passes vacuously. (In a larger real file the same defect also shows the reverse polarity: provably-executed lines report uncovered — see "Real-world signature" below.)

### Run B — control (`"sourceMap": false, "inlineSourceMap": true, "inlineSources": true`)

```
 lib.ts   |      90 |      100 |      50 |      90 | 19-20
```

Exactly right: 50% functions, `unused`'s body uncovered. Same code, same run — only the map delivery differs. This control also proves the fix direction works end to end.

### Stack-trace twin

**src/boom.ts** (same 10-line comment, then):

```ts
export const boom = (): never => {
  throw new Error("where am I?"); // true position: line 12, col 9
};
```

| configuration | reported frame |
| --- | --- |
| external maps (Run A) | `at boom (…\src\boom.ts:15:11)` — **.ts filename, emitted-JS line** |
| inline maps + `--enable-source-maps` | `at boom (C:\src\boom.ts:12:9)` — right line/col, **nonexistent path** (defect 3) |

## Mechanism, fully traced

**(a) What actually executes.** Intercepting `Module.prototype._compile` during a ttsx run shows the module compiled under the SOURCE filename is the transformed CJS whose last line is a *relative external* map reference:

```
COMPILED filename: D:\...\packages\mcp\src\project\AutoMoviePrerequisite.ts
COMPILED line count: 104            (source file: 168 lines)
COMPILED last line: //# sourceMappingURL=AutoMoviePrerequisite.js.map
```

Serving code (all refs at `b5ead8f8`, `packages/ttsc/src/launcher/internal/`):

- CJS: `runtimeHooks.ts:179` — `module._compile(resolveServedSource(filename).source, filename)`;
- ESM: `runtimeHooks.ts:251-256` (`load` → `resolveRuntimeSource` → `served.source`);
- the emit text is returned verbatim: `serveEntryEmit` (`:836`) and `serveBuiltDependency` (`:912`) read `emittedFile` and never consult the sibling `${emittedFile}.map` that tsgo wrote right next to it.

**(b) Why the reference can never resolve.** Node resolves a relative `sourceMappingURL` against the *script's* URL — here the `.ts` source path, where no `.js.map` exists. And it could not be salvaged by absolutizing alone: the entry emit and dependency emits live under the per-run dir (`runTtsx.ts:267` — `depCacheDir = path.join(execution.cleanupDir, "deps")`), and `runTtsx.ts:322` — `removeRuntimeOutput(execution.cleanupDir)` — deletes it the moment the child exits, so any post-mortem consumer (coverage reporters read the JSON after the run) finds nothing. Only a map **inlined into the served text** survives both.

**(c) What Node records.** With `NODE_V8_COVERAGE` set, Node writes a `source-map-cache` entry for every script carrying a magic comment. Observed, same file, same probe:

| owning tsconfig | smc entry | consequence |
| --- | --- | --- |
| `sourceMap: true` | `data: null`, `lineLengths: 104` | c8 cannot remap |
| `inlineSourceMap: true` | `data: {...}` present | c8 remaps exactly |

(`data: null` on unfetchable maps is Node's documented behavior — nodejs/node#34305; c8's own `lib/report.js` cites it and bails: `if (!sourceMapAndLineLengths.data) return`.)

**(d) The failure composition.** V8's raw counts are *correct* (verified: the executed function's range has `count: 1`, the never-called one `count: 0`). With no map data, c8 applies those transformed-text byte offsets to the original TS file — 104-line offsets against a 168-line source — producing per-line reports that are wrong in both directions. A project whose suite genuinely executes ~everything shows a clean 100% under any offset shift, which is exactly why this stayed invisible behind automovie's engine gate and only surfaced when gating a partially-covered package.

**(e) Defect 3 — mis-rooted map `sources`.** The inline-map control run records `sources: ["file:///D:/packages/mcp/src/project/AutoMoviePrerequisite.ts"]` for a file that really lives at `D:/github/samchon/automovie/packages/mcp/src/...`; the repro prints `C:\src\boom.ts` for `…\repro\src\boom.ts`. The paths appear rebased against the virtual emit layout rather than the real source tree. c8 happens to survive this (it keys by script URL and reconstructs the transformed text from `lineLengths`), but every consumer that keys by map `sources` — mapped stack frames above, IDE links, other coverage tooling — prints or points at a path that does not exist.

## Why this is ttsx's to fix

The hooks deliberately execute **transformed text under the source's identity**. That identity swap is only sound if the served text carries a self-contained way back to the source. Today it carries the opposite: a reference that (i) resolves against the wrong base, and (ii) would dangle after exit even if it resolved — so every standard consumer of Node's script-URL-relative source-map machinery (V8 coverage, `--enable-source-maps` stacks, debuggers) silently degrades. No tsconfig combination the *user* controls fixes stacks-with-correct-paths, and requiring every consuming project to switch its library tsconfigs to inline maps (bloating published `lib/` output) is not a real workaround.

## Proposed fix

**At the single serve choke point** (`resolveServedSource`, or inside `serveEntryEmit`/`serveBuiltDependency` where `emittedFile` is in hand):

1. If the served text ends with an external `//# sourceMappingURL=<relative>` and `${emittedFile}.map` exists: parse the sibling map, **rewrite `sources` to absolute `file://` URLs of the true source files** (the serve path already knows `sourceFile`; tsgo's per-file emit is 1:1), re-serialize, and replace the trailing comment with `//# sourceMappingURL=data:application/json;base64,<...>`. This one change fixes coverage (Node caches real `data`), fixes defect 3 for stacks and debuggers, and is immune to both the wrong-relative-base and the post-exit cleanup, because the map travels inside the text V8 already captured.
2. Cache the rewritten text per `emittedFile` (dependency emits are already content-keyed and shared cross-process; the entry emit is per-run) so the base64 work happens once per file per build, not per import.
3. When the owning tsconfig emits **no** maps (`sourceMap: false`), consider forcing map emission for the *runtime* build only — the runtime emit is ttsx-private and transient, so overriding emit options there leaks into nobody's published `lib/`. (Alternative to 1+3 combined: always build runtime emits with `inlineSourceMap` + `inlineSources` forced — simpler, but it leaves defect 3's `sources` rebasing to be fixed in the same pass anyway, so the serve-time rewrite is the more complete seam.)
4. Bootstrap: call `process.setSourceMapsEnabled(true)` in `registerRuntimeHooks` (or document `--enable-source-maps`) so stacks map out of the box once maps are servable. The orphan type-strip path (`stripTypeScriptTypes`, line-preserving by construction) needs nothing.

## Regression tests worth adding to the ttsx e2e suite

- **Coverage:** spawn `c8 ttsx` over a fixture like `lib.ts` above (tall comment + one never-called export); assert the uncovered lines land on the never-called body, and that the run's coverage JSON has `source-map-cache[<file>].data !== null`.
- **Stacks:** a throwing fixture behind the same comment; assert the frame reports the true `line:col` **and the real absolute source path**.
- **Both twins under `sourceMap: true` and `sourceMap: false` owning tsconfigs, entry-project and dependency** — `serveEntryEmit` and `serveBuiltDependency` are separate code paths.

## Environment

ttsc 0.18.4 (`b5ead8f8`) · typia 13.0.2 (transform plugin active in the real-world case; the repro needs none) · Node v22.15.0 · pnpm 10 · Windows 11 (the mechanism is platform-independent; only the path spellings here are Windows').

## Real-world signature (how it looks from the outside)

In samchon/AutoMovie#1040 the same run reported, for one 168-line file: lines 23–65 uncovered while the code at those lines had *provably just executed* (the probe printed its throw message), and a never-called function reported covered. If coverage of a ttsx-run project looks "mostly right but weirdly shifted", or a gate passes at 100% while you can name an uncalled function — it is this.

samchon:
## Intent

`ttsx` executes tsgo-built JavaScript under the original `.ts` source URL. When the owning tsconfig emits external maps (a library's default `sourceMap: true`), the served text ended with a relative `//# sourceMappingURL=<name>.js.map` that **nothing can resolve**: Node resolves it against the `.ts` script URL (no `.js.map` there), and the per-run emit dir is deleted at process exit anyway. The consequences (#353):

1. **V8/c8 coverage is garbage, including false 100%** — Node caches the script with `source-map-cache[url].data = null`, so c8 applies transformed-JS offsets to the original TS text; a never-called function reports covered.
2. **Stack frames point at wrong lines** — the `.ts` filename with emitted-JS line numbers.
3. **Mis-rooted map `sources`** — even when a map is consumed, mapped frames/IDE links print a nonexistent path.

## Mechanism of the fix

At the single serve choke point (`resolveServedSource`, which holds `emittedFile` + `sourceFile`), a shared helper rewrites the served text before it is returned from **both** `serveEntryEmit` (entry lane) and `serveBuiltDependency` (dependency lane — a separate code path):

- read the sibling `${emittedFile}.map`, rewrite its `sources` to absolute `file://` URLs of the true on-disk source (tsgo's per-file emit is 1:1, so the serve path's known `sourceFile` is the anchor), drop `sourceRoot`, re-serialize, and replace the trailing comment with `//# sourceMappingURL=data:application/json;base64,…`.

Because the map now travels **inside** the text V8 captures at compile time, it survives both the wrong resolution base and the post-exit cleanup. This one change fixes coverage (Node caches real `data`), stacks, and the mis-rooted `sources`. The rewrite is cached per emitted file and is idempotent (re-encoding already-absolutized text reproduces the same bytes), keeping the shared cross-process dependency cache deterministic. `pathToFileURL` builds the URLs (Windows-safe).

## No-map tsconfig handling (and justification)

Rather than branch on whether the consumer set `sourceMap`, both runtime builds **force an external map** (`--sourceMap true --inlineSourceMap false --inlineSources false`, appended after the user's forwarded flags so they win). The runtime emit always lands in a private, transient directory — the entry's PID-isolated dir or the shared dependency cache, **never** the consumer's `outDir` — so the override leaks into nobody's published `lib/`. Forcing an *external* map (over the alternative of forcing inline) gives the serve path one uniform shape to rewrite regardless of whether the consumer configured `sourceMap`, `inlineSourceMap`, or no map at all, and routes every case through the same `sources`-absolutizing pass. The entry build (`prepareExecution`) and the dependency build (`buildDependency`) are separate `runBuild` call sites, so each forces maps independently.

## Bootstrap

`installRuntimeHooks` calls `process.setSourceMapsEnabled(true)` so stacks map out of the box (no `--enable-source-maps`); user code that later toggles it still wins. The orphan type-strip path is line-preserving and untouched.

## Test plan

New ttsx e2e tests under `tests/test-ttsc/src/features/ttsx-runtime/`, driving a real run under `NODE_V8_COVERAGE` and reading Node's coverage JSON — no dependency on c8 (it is not in the repo). They assert the exact inputs c8 consumes, which decompose its result per the issue's mechanism:

- **Coverage**: `source-map-cache[lib].data` is non-null (was null) and its `sources[0]` is the real absolute source (defect 3); V8 records the never-called `unused` with count 0 and its called twin `used` with count ≥ 1.
- **Stacks**: a thrown frame reports the true `line:col` at the real absolute source path.
- Both across `sourceMap: true` **and** `sourceMap: false` owning tsconfigs, for the **entry project** and a **built dependency** (the two serve lanes).

Verified fail-before/pass-after against a scratch repro of the issue: coverage `data: null → present`, stack `boom.ts:16:11 → boom.ts:13:9` (true position, real path). Full `ttsx-runtime` suite green except two pre-existing Windows-local failures (file-symlink `EPERM`, a temp-`tsc.exe` spawn `UNKNOWN`) that fail identically on unmodified `master`.

Docs: `website/src/content/docs/ttsc/execute.mdx` gains a short "Source maps, stack traces, and coverage" section.

Closes #353

### DIFF
--- a/packages/ttsc/src/compiler/internal/runBuild.ts
+++ b/packages/ttsc/src/compiler/internal/runBuild.ts
@@ -24,6 +24,14 @@ import { outputText, spawnNative } from "./spawnNative";
 type RunBuildOptions = TtscBuildOptions & {
   skipDiagnosticsCheck?: boolean;
   forceListEmittedFiles?: boolean;
+  /**
+   * Emit an external source map from the direct tsgo build lane even when the
+   * project configures none. Set by the ttsx runtime builds so a served emit
+   * carries a map to inline under the source URL (issue #353). Applied only to
+   * the plain tsgo emit — never forwarded to a native plugin host, whose own
+   * emit honours the project's `sourceMap` setting.
+   */
+  forceRuntimeSourceMap?: boolean;
 };
 
 type BuildTiming = {
@@ -681,6 +689,12 @@ function createTsgoBuildArgs(
     args.push("--noEmit", "false", "--emitDeclarationOnly", "false");
     if (execution.rewriteRelativeImportExtensionsForEmit) {
       args.push("--rewriteRelativeImportExtensions");
+    }
+    // The ttsx runtime build asks for an external map when the project emits
+    // none, so its served emit carries a map to inline under the source URL.
+    // Pushed before passthrough so an explicit user `--sourceMap` still wins.
+    if (options.forceRuntimeSourceMap === true) {
+      args.push("--sourceMap", "true");
     }
   } else if (options.emit === false) {
     args.push("--noEmit");
--- a/packages/ttsc/src/launcher/internal/prepareExecution.ts
+++ b/packages/ttsc/src/launcher/internal/prepareExecution.ts
@@ -6,7 +6,6 @@ import { readProjectConfig } from "../../compiler/internal/project/readProjectCo
 import { resolveEmittedJavaScript } from "../../compiler/internal/resolveEmittedJavaScript";
 import { runBuild } from "../../compiler/internal/runBuild";
 import type { TtscCommonOptions } from "../../structures/internal/TtscCommonOptions";
-import { RUNTIME_SOURCE_MAP_TSGO_FLAGS } from "./servedSourceMap";
 
 /** Subdirectory name that isolates concurrent ttsx processes by PID. */
 const PROCESS_CACHE_KEY = String(process.pid);
@@ -103,6 +102,13 @@ function createProjectContext(
       typeof project.compilerOptions.module === "string"
         ? project.compilerOptions.module
         : undefined,
+    // Force a source map on the transient runtime emit only when the project
+    // configures none — when it already emits `sourceMap` or `inlineSourceMap`,
+    // the serve path inlines/absolutizes that map, so no override is needed
+    // (issue #353).
+    forceRuntimeSourceMap:
+      project.compilerOptions.sourceMap !== true &&
+      project.compilerOptions.inlineSourceMap !== true,
     built: false,
     emittedFiles: undefined as string[] | undefined,
   };
@@ -139,14 +145,13 @@ function buildProject(
     forceListEmittedFiles: true,
     cacheDir: context.pluginCacheDir,
     outDir: context.emitDir,
-    // Force an external source map on the transient entry emit so the serve
-    // path can inline it under the source URL (issue #353). This emit lands in
-    // a PID-isolated temp dir, never the consumer's `outDir`, so the override
-    // leaks nowhere; appended last so it wins over any forwarded map flag.
-    passthrough: [
-      ...(options.passthrough ?? []),
-      ...RUNTIME_SOURCE_MAP_TSGO_FLAGS,
-    ],
+    passthrough: options.passthrough,
+    // Emit a source map on the transient entry emit (a PID-isolated temp dir,
+    // never the consumer's `outDir`) so the serve path can inline it under the
+    // source URL. Routed as a dedicated build option, not a forwarded tsgo
+    // flag, so it never reaches a native plugin host's argument parser (issue
+    // #353).
+    forceRuntimeSourceMap: context.forceRuntimeSourceMap,
     plugins: options.plugins,
     quiet: true,
     singleThreaded: options.singleThreaded,
--- a/packages/ttsc/src/launcher/internal/prepareExecution.ts
+++ b/packages/ttsc/src/launcher/internal/prepareExecution.ts
@@ -6,6 +6,7 @@ import { readProjectConfig } from "../../compiler/internal/project/readProjectCo
 import { resolveEmittedJavaScript } from "../../compiler/internal/resolveEmittedJavaScript";
 import { runBuild } from "../../compiler/internal/runBuild";
 import type { TtscCommonOptions } from "../../structures/internal/TtscCommonOptions";
+import { RUNTIME_SOURCE_MAP_TSGO_FLAGS } from "./servedSourceMap";
 
 /** Subdirectory name that isolates concurrent ttsx processes by PID. */
 const PROCESS_CACHE_KEY = String(process.pid);
@@ -138,7 +139,14 @@ function buildProject(
     forceListEmittedFiles: true,
     cacheDir: context.pluginCacheDir,
     outDir: context.emitDir,
-    passthrough: options.passthrough,
+    // Force an external source map on the transient entry emit so the serve
+    // path can inline it under the source URL (issue #353). This emit lands in
+    // a PID-isolated temp dir, never the consumer's `outDir`, so the override
+    // leaks nowhere; appended last so it wins over any forwarded map flag.
+    passthrough: [
+      ...(options.passthrough ?? []),
+      ...RUNTIME_SOURCE_MAP_TSGO_FLAGS,
+    ],
     plugins: options.plugins,
     quiet: true,
     singleThreaded: options.singleThreaded,
--- a/packages/ttsc/src/launcher/internal/runtimeHooks.ts
+++ b/packages/ttsc/src/launcher/internal/runtimeHooks.ts
@@ -10,10 +10,7 @@ import { resolveEmittedJavaScript } from "../../compiler/internal/resolveEmitted
 import { resolveTsgo } from "../../compiler/internal/resolveTsgo";
 import { runBuild } from "../../compiler/internal/runBuild";
 import { outputText, spawnNative } from "../../compiler/internal/spawnNative";
-import {
-  RUNTIME_SOURCE_MAP_TSGO_FLAGS,
-  inlineServedSourceMap,
-} from "./servedSourceMap";
+import { inlineServedSourceMap } from "./servedSourceMap";
 
 /**
  * Synchronous Node module hooks installed (via `module.registerHooks`) in the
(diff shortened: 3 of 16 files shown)


samchon:
-

PR #419, opened 2026-07-13
fix(lint): align no-duplicate-imports with ESLint mergeability and options

samchon:
## Intent

`no-duplicate-imports` reported every second import declaration whose module string matched an earlier one. The official ESLint rule reports a repeated module only when the two declarations could be merged into one legal declaration, and exposes `allowSeparateTypeImports` and `includeExports`. This PR ports the official categorization/mergeability analysis and both options through the existing typed rule-options transport, per #401.

Closes #401

## Scope

- **Go rule** (`packages/lint/linthost/rules_no_duplicate_imports.go`, extracted from `rules_core_extra.go`): declarations are categorized as default / named / namespace / export-all / side-effect the way the official `getImportExportType` reads ESTree specifiers (first named-or-namespace specifier wins, empty clauses are side-effect). Mergeability mirrors `isImportExportCanBeMerged`, including the ESLint 9.30.1 correction that a type-only default cannot merge with type-only named bindings. `allowSeparateTypeImports` exempts pairs whose clause-level type-ness differs (inline `import { type Foo }` stays a value declaration); `includeExports` records and reports `export … from` declarations with the four official message pairings (import / import-as-export / export / export-as-import). Module names are compared trimmed, mirroring `source.value.trim()`; specifier-less exports and empty module strings are skipped. Options decode via the existing `Context.DecodeOptions`.
- **Typed options surface**: new `ITtscLintCoreNoDuplicateImportsRuleOptions` (`ITtscLintCoreRuleOptions.ts`), the rule key upgraded to `TtscLintRuleOptionsSetting<…>` in `ITtscLintCoreRules`, registered in `ITtscLintRuleOptionsMap`, exported from the structures index. No transport changes.
- **Tests**: 27 focused Go cases under `packages/lint/test/rules/imports-modules/` (one case per file) covering every categorization branch with a discriminating pairing, both operand orders of each mergeability guard, option on/off twins, all four includeExports messages (including two findings on one declaration), module trimming, empty specifiers, local exports, and a misspelled option key at the engine layer. The corpus fixture `tests/test-lint/src/cases/no-duplicate-imports.ts` now pins the default-option contract (mergeable value pair and value+type pair report; named-vs-namespace and type-default-vs-type-named do not), with the Go corpus twin kept in sync. `test_lib_index_d_ts_rule_options_autocomplete_per_rule.ts` gains the happy-path tuple plus `@ts-expect-error` rejection of a typo'd option key and a non-boolean option value.
[...]

PR #423, opened 2026-07-13
fix(lint): align no-unused-expressions with default ESLint semantics

### LINKED ISSUE #408: fix(lint): align no-unused-expressions with default ESLint semantics
## Summary

`@ttsc/lint`'s `no-unused-expressions` rule does not implement the default ESLint semantics advertised by the package. It currently rejects valid directive prologues, `void` expressions, and productive calls wrapped in TypeScript assertions, while accepting default-invalid tagged templates and misplaced string expressions.

This is not an option-policy disagreement. The fixture enables the rule with the scalar default configuration used by AutoBE:

```json
{ "rules": { "no-unused-expressions": "error" } }
```

The same result is reproduced with published `@ttsc/lint@0.18.4` and current `master` (`fbbad7caeb44a16c77808c350ffaeeed608cabd1`).

## Why this is a bug

The `@ttsc/lint` README says core rule IDs correspond one-to-one with ESLint core rules so projects can migrate the same severity by name. ESLint's default contract is explicit:

- every string expression at the beginning of a script, module, or function is part of the directive prologue and is accepted, regardless of its text;
- `void expression` is accepted because evaluating the operand can have side effects;
- a tagged template expression is rejected unless `allowTaggedTemplates: true` is selected;
- a string expression after a non-directive statement is not a directive and is rejected;
- JSX is accepted by default unless `enforceForJSX: true` is selected.
- TypeScript `as`, angle-assertion, non-null, and instantiation wrappers inherit the side-effect classification of their wrapped expression.

Reference: https://eslint.org/docs/latest/rules/no-unused-expressions

The current implementation instead classifies expressions by a hard-coded kind list and treats only the literal texts `"use strict"` and `"use asm"` as directives:

- `packages/lint/linthost/rules_suggestions.go:814-876`
- `packages/lint/src/structures/rules/ITtscLintCoreRules.ts:1080-1082`

Consequently, the rule conflicts with the standard Promise opt-out idiom `void promise()`, with modern directive prologues such as `"use client"`, and with type-preserving TypeScript wrappers around productive expressions.

No `compilerOptions.plugins` entry is used.

`package.json`:

```json
{
  "private": true,
  "devDependencies": {
    "@ttsc/lint": "0.18.4",
    "ttsc": "0.18.4",
    "typescript": "7.0.2"
  }
}
```

`tsconfig.json`:

```json
{
  "compilerOptions": {
    "noEmit": true,
    "strict": true,
    "target": "ES2022",
    "module": "NodeNext",
    "moduleResolution": "NodeNext"
  },
  "files": ["case.ts"]
}
```

`lint.config.json`:

```json
{ "rules": { "no-unused-expressions": "error" } }
```

`case.ts`:

```ts
declare function work(): Promise<void>;
declare const tag: (strings: TemplateStringsArray) => string;

"use client";
void work();
(work() as Promise<void>);
work()!;
tag`value`;

function misplacedDirective(): void {
  console.log("before");
  "use strict";
}

export { misplacedDirective };
```

Run:

```bash
```

Observed on both 0.18.4 and current master:

```text
case.ts:4:1 - error TS17505: [no-unused-expressions] ...
4 "use client";

case.ts:5:1 - error TS17505: [no-unused-expressions] ...
5 void work();

case.ts:6:1 - error TS17505: [no-unused-expressions] ...
6 (work() as Promise<void>);

case.ts:7:1 - error TS17505: [no-unused-expressions] ...
7 work()!;

Found 4 errors ...
exit 2
```

There is no diagnostic for `tag\`value\`` or the misplaced `"use strict"`.

Expected default result:

- no finding for `"use client"` because it is in the module directive prologue;
- no finding for `void work()`;
- no finding for the `as`-wrapped or non-null-wrapped calls because evaluating each still calls `work()`;
- one finding for the tagged template;
- one finding for the string expression after `console.log`.

Implement an expression-effect classifier equivalent to the upstream default semantics and determine directives from AST position, not from a list of recognized strings.

The directive check should:

1. identify the containing source file, function body, module block, or namespace body that can own a directive prologue;
2. accept consecutive leading string-literal expression statements until the first non-directive statement;
3. reject string expressions elsewhere, including class static blocks;
4. work for arbitrary directive text without framework-specific names.

The expression classifier should at minimum preserve upstream defaults for `void`, `delete`, calls, construction, update, await/yield, tagged templates, JSX, logical/conditional expressions, and TypeScript wrapper expressions (`as`, angle assertions, non-null assertions, and instantiation expressions). If rule options are exposed, declare their typed option shape and honor `allowShortCircuit`, `allowTernary`, `allowTaggedTemplates`, `enforceForJSX`, and `ignoreDirectives` instead of silently applying a different policy.

### Forbidden shortcuts

- Do not add `"use client"`, framework directives, or more strings to the current allowlist.
- Do not suppress all string literals or all tagged templates.
- Do not scan raw source with a regex when AST statement order already defines the directive prologue.
- Do not add transformer plugins to the reproduction tsconfig.

## Acceptance criteria

- The reproduction produces exactly the two expected findings (tagged template and misplaced string) under the default scalar setting.
- Arbitrary consecutive prologue strings are accepted; the same strings outside a prologue are rejected.
- `void sideEffect()` remains accepted and interoperates with `typescript/no-floating-promises`' default `void` opt-out.
- Tagged templates are rejected by default and accepted only when the corresponding option is explicitly enabled.
- JSX follows `enforceForJSX` default/option semantics.
- Calls, `new`, assignments, updates, `delete`, `await`, and `yield` remain negative controls.
- TypeScript assertion/non-null/instantiation wrappers inherit the effect classification of their wrapped expression.
- Focused native tests and a real no-plugin CLI fixture cover every positive and negative control.

## Existing issue review

- #137 was a closed rule-coverage inventory. It did not track semantic parity for this implemented rule.
- No open or closed issue containing `no-unused-expressions` describes this default-behavior mismatch.
- #401 concerns `no-duplicate-imports` mergeability/options and is unrelated.

samchon:
## Intent

`no-unused-expressions` claimed 1-to-1 parity with the ESLint core rule but classified statements through an ad-hoc allow-list and recognized directives by literal text (`"use strict"` / `"use asm"` only). The issue's reproduction shows the four resulting divergences: `"use client"` and `void work()` were wrongly reported, while a tagged template and a misplaced `"use strict"` were wrongly accepted. Closes #408.

## Implementation

The rule is now a direct port of the upstream implementation (`eslint/lib/rules/no-unused-expressions.js`), verified against the current `main` source rather than the docs alone:

- **Disallow-list polarity.** Upstream flags a fixed list of side-effect-free shapes and ignores unknown node types. The Go classifier mirrors that list exactly (identifiers, member/element access, literals, templates, regexes, array/object/function/class expressions, `this`, `new.target`, ordinary binaries, comma sequences, `typeof`, pure prefix unaries), so calls, `new`, assignments, updates, `delete`, `void`, `await`, `yield`, and dynamic `import()` are accepted. `void promise()` interoperates with the `typescript/no-floating-promises` opt-out again.
- **Positional directive prologues.** A statement is a directive iff it sits in the leading run of unparenthesized string-literal statements of a container that can own a prologue: source file, namespace/module block, or function-like body (mirroring upstream `isTopLevelExpressionStatement`, which is `Program | TSModuleBlock | function body`; class static blocks own no prologue). Text is irrelevant, so `"use client"` and any future directive work; parenthesized strings are not directives and terminate the run, matching the parser-directive semantics ESLint gets from ESTree.
- **TypeScript wrappers.** `as`, angle assertions, non-null `!`, and instantiation expressions recurse into the wrapped expression, exactly the four wrapper entries in the upstream Checker. `satisfies` is deliberately *not* recursed: upstream has no `TSSatisfiesExpression` entry, so such statements are ignored — this port pins that upstream behavior with a test instead of inventing a stricter policy.
- **Options.** `allowShortCircuit`, `allowTernary`, `allowTaggedTemplates`, `enforceForJSX`, and `ignoreDirectives` are exposed through the typed rule-options transport (`TtscLintRuleOptionsSetting`, `ITtscLintRuleOptionsMap`, decoded via `Context.DecodeOptions`) with upstream defaults (all `false`) and upstream recursion semantics (logical → right operand; ternary → both result branches). `ignoreDirectives` reproduces upstream's ESTree view where parentheses are invisible, so it additionally exempts parenthesized strings inside the leading run; real prologues are always exempt because this engine always recognizes directives (the option exists upstream for ES3 parsing, which ttsc has no equivalent of).

## Scope notes / deviations

- The issue asks that wrappers "inherit the effect classification", which the port satisfies for the four kinds upstream handles; `satisfies` follows upstream's ignore-unknown behavior rather than recursion (documented and pinned in `no_unused_expressions_ts_wrappers_inherit_classification_test.go`).
- No other issue claim required adjustment — every claim was verified against the code and the upstream source.

## Test plan

- **CLI corpus fixtures** (run e2e through the real `ttsc` binary and mirrored in Go unit tests): `no-unused-expressions.ts` reproduces the issue scenario, `no-unused-expressions-directive-boundaries.ts` pins where prologues end (parenthesized terminator, file/function/namespace misplacement, static blocks), `no-unused-expressions-side-effect-free-kinds.ts` enumerates every disallowed kind one line at a time.
- **Go unit tests** (one case per file) cover the surfaces the corpus cannot: zero-finding productive shapes (including logical assignment, optional calls, `yield*`, wrapper-of-call, `satisfies`), arbitrary-text prologues at every function-like position, TS wrapper recursion incl. nesting, default JSX acceptance in TSX, and each option with a negative twin one property away (`enforceForJSX`, `allowShortCircuit` right-operand-only, `allowTernary` both-branches, short-circuit+ternary composition, `allowTaggedTemplates` vs untagged templates, `ignoreDirectives` loose view vs misplaced/static-block strings).


samchon:
Integration repair complete at 5612889844b15042b9e186348a591d90bfa1ffca.

Conflict cause: this branch and PR #419 split from the same pre-registry base and independently introduced the core-rule options surface, producing four add/add or overlapping conflicts when current master was merged.

Resolution: merged origin/master non-destructively and combined both configurable core rules. The final tree retains no-duplicate-imports and no-unused-expressions option interfaces, both ITtscLintRuleOptionsMap entries, both typed rule declarations, and independent autocomplete happy/negative cases. The overlapping README, website guide, index export, and current-master native options fixture were also preserved. No no-unused-expressions runtime semantics were changed by the repair.

Validation policy: pnpm format completed. I performed a clean exhaustive solo review of all 23 resulting PR paths plus the latent option-decoding, registry, directive-container, and AST helper callers. Per the integration task, no local build, test, typecheck, lint, static-analysis, or reproduction command was run; the pushed head is delegated to GitHub CI.

samchon:
Follow-up acceptance coverage is now pushed at `306c8a92ac2d6b95bd848ee7fe1204e322d208a3`.

The new checked CLI project mirrors issue #408's package-loading path:

- the nearest `package.json` declares `@ttsc/lint` as a direct dependency;
- the test harness links that dependency to the workspace package, so its `ttsc.plugin` marker drives auto-discovery;
- `tsconfig.json` deliberately has no `compilerOptions.plugins` entry;
- the real `ttsc --cwd <fixture> --noEmit` invocation is pinned to exit code `2` and exactly two `TS17505` diagnostics (the tagged template and misplaced string);
- directive prologues, calls, `new`, assignment/update, `delete`, `void`, `await`, `yield`, TypeScript call wrappers, and default JSX remain negative controls.

I also restarted the whole-PR review from the current master diff and rechecked all production logic, option registries, native/corpus tests, docs, package discovery, CLI exit propagation, and fixture helpers. No further defect was found.

Per coordinator instruction, no local formatter, test, build, typecheck, lint, static-analysis, or reproduction command was run; GitHub CI is the validation authority.

PR #427, opened 2026-07-13
fix(lint): make no-fallthrough honor comments and control-flow completion

### LINKED ISSUE #411: fix(lint): make no-fallthrough honor comments, control flow, and empty-case defaults
## Summary

`@ttsc/lint`'s `no-fallthrough` rule reports two forms that ESLint accepts by default and misses one default-invalid empty-case form:

1. an intentional fallthrough marked by the standard `// falls through` comment;
2. a case whose final `if/else` terminates on every path;
3. an empty case visually separated from the following label by a blank line, which is reportable while `allowEmptyCase` remains at its default `false`.

Both false positives and the false negative reproduce with published `@ttsc/lint@0.18.4` and current `master` (`fbbad7caeb44a16c77808c350ffaeeed608cabd1`) using a no-plugin tsconfig.

## Official default semantics

ESLint's rule reports reachable fallthrough that is not marked by an intentional-fallthrough comment. The default comment pattern accepts variations matching `/falls?\s?through/i`, and ESLint uses code-path reachability rather than checking only the syntactically last statement.

The scalar defaults also set `allowEmptyCase: false` and `reportUnusedFallthroughComment: false`. An adjacent empty label is accepted as ordinary grouped-case syntax. An empty case separated from the next label by blank lines is considered an intentional empty body and reports unless `allowEmptyCase: true` is selected. This is an upstream default/option distinction, not a request to reject every grouped label.

Reference: https://eslint.org/docs/latest/rules/no-fallthrough

The current implementation in `packages/lint/linthost/rules_problems.go:505-558`:

- ignores comments entirely;
- skips every empty case unconditionally, discarding the default blank-line distinction;
- calls `isTerminating` only on the last top-level statement;
- recognizes only `break`, `continue`, `return`, `throw`, and a block whose final statement recursively has one of those kinds;
- does not reason about `if/else`, `try/catch/finally`, or other control-flow joins.

`lint.config.json`:

```json
{ "rules": { "no-fallthrough": "error" } }
```

`case.ts`:

```ts
declare const mode: 1 | 2;
declare const condition: boolean;

switch (mode) {
  case 1:
    console.log("intentional");
    // falls through
  case 2:
    console.log("done");
}

function stopEveryPath(): void {
  switch (mode) {
    case 1:
      if (condition) {
        return;
      } else {
        throw new Error("stop");
      }
    case 2:
      return;
  }
}

switch (mode) {
  case 1:

  case 2:
    break;
}

switch (mode) {
  case 1:
  case 2:
    break;
}

export { stopEveryPath };
```

```bash
```

```text
case.ts:8:3 - error TS14225: [no-fallthrough] Expected a 'break' statement before this case.
case.ts:20:5 - error TS14225: [no-fallthrough] Expected a 'break' statement before this case.
Found 2 errors ...
exit 2
```

Expected: exactly one finding at line 28, for the blank-line-separated empty case. The marked fallthrough, fully terminating branch, and adjacent grouped label remain clean. The command still exits 2, but for a different and singular diagnostic than the two currently emitted.

Base the diagnostic on reachability at the next case label, with nested function boundaries handled correctly. Reuse the compiler/checker's control-flow information if it is exposed; otherwise implement a structured statement completion analysis that returns whether normal completion is possible and composes blocks, `if/else`, loops, labeled statements, and `try/catch/finally` without mistaking a nested callback's `return` for termination of the case.

Separately, inspect comments and empty-label spacing in the exact trailing range before the next case label using ESLint's default marker and blank-line semantics. If options are supported, expose and honor `commentPattern`, `allowEmptyCase`, and `reportUnusedFallthroughComment` through the typed rule option map.

### Forbidden shortcuts

- Do not treat every `if`, `try`, or block as terminating merely because one branch exits.
- Do not search the whole case text for the word `fallthrough`; only an eligible trailing comment marks the transition.
- Do not hard-code only the exact string `// falls through`.
- Do not suppress all cases containing a comment.
- Do not classify every empty label identically; preserve the documented grouped-label versus blank-line-separated distinction.

## Acceptance criteria

- The reproduction emits exactly the one expected blank-line empty-case finding under the scalar default setting.
- A real unmarked reachable fallthrough still reports exactly once at the transition.
- Standard marker spellings (`fall through`, `falls through`, `fallsthrough`, case-insensitive) are accepted only in the eligible trailing comment position.
- An `if` without an `else`, or with one normally completing branch, still reports.
- An `if/else` whose every branch exits does not report.
- `try/catch/finally`, labeled `break`/`continue`, nested blocks, and nested function boundaries have explicit positive and negative tests.
- Adjacent grouped empty labels remain clean; blank-line-separated empty cases report by default and become clean only with `allowEmptyCase: true`.
- Real no-plugin CLI coverage accompanies native unit tests.

## Existing issue review

- #137 was a closed rule-coverage inventory and did not track control-flow/comment parity for this implemented rule.
- No open or closed issue containing `no-fallthrough` describes these false positives.
- The defect is independent of #401 (`no-duplicate-imports`).

samchon:
## Intent

`no-fallthrough` false-positived on the two forms issue #411 reproduces: an intentional fallthrough marked with the standard `// falls through` comment, and a case whose final `if/else` terminates on every path. The old implementation only looked at the syntactic kind of the last top-level statement and never read comments.

Closes #411

## What changed

- **Completion analysis** (`packages/lint/linthost/rules_no_fallthrough.go`): the rule now asks "can this case's statement list complete normally?" via a structured analysis that composes blocks, `if/else`, `while`/`do`/`for`/`for-in`/`for-of` (with ESLint's literal-only loop-test folding: `while (true)` / `while (1)` are infinite, `while (!0)` is not), labeled statements with `break L` / `continue L` absorption, nested `switch` (no-default / last-clause / inner-break joins), and `try/catch/finally` (JLS-style: finally's completion wins; escapes from try/catch survive only a normally-completing finally). Statements after an abrupt completion are unreachable and contribute nothing, so an unreachable `break` cannot reopen an infinite loop. Expression trees are never entered — a nested function/arrow `return` never terminates the case.
- **Marker comments**: exactly ESLint's two eligible positions — the *last* comment before the next case label, and (for a sole-block case body) the last comment before the block's closing brace — matched against the default `/falls?\s?through/i` or a configured `commentPattern`. Directive comments (`eslint-*`, plus this host's `lint-*` family) never count as markers, mirroring upstream `directivesPattern`.
- **Options** through the typed rule-options transport (`ITtscLintNoFallthroughRuleOptions`, wired into `ITtscLintRuleOptionsMap` and `TtscLintRuleOptionsSetting`): `commentPattern` (replaces the default pattern entirely, like ESLint), `allowEmptyCase` (blank-line-separated empty cases), `reportUnusedFallthroughComment` (marker on a case that cannot fall through reports at the comment). Messages now match upstream: `Expected a 'break' statement before 'case'.` / `...'default'.` / the unused-comment message.
- **Docs**: `packages/lint/README.md` line and the `website` core rules page (semantics + options), with an example that is truthful under the new semantics.

## Oracle

Semantics were taken from the ESLint rule source (`lib/rules/no-fallthrough.js`), its shared `directivesPattern`, and its regression suite — not from the issue text alone. Notable oracle-derived details the tests pin: only the *last* comment of a region counts; `case 1: ;` is non-empty even with `allowEmptyCase`; an empty case with a blank-line gap reports by default while adjacent labels never do; comments do not count as tokens for the blank-line check; the unused-comment branch keys on case-end reachability, not on "ends with break".

## Deliberate deviations (bounded scope, with reasoning)

- **Analysis is local to each switch.** ESLint's code-path analysis would also skip reporting a fallthrough inside entirely dead code (e.g. a switch after `return`). Reproducing that needs whole-function flow, which is `no-unreachable`'s concern; a dead switch is flagged there instead. No upstream regression test covers this combination.
- **Checker flow nodes were not reused.** The binder's `FlowNode` graph plus checker reachability would force `NeedsTypeChecker`, moving the rule out of the parallel AST-only engine lane and requiring a bound Program in unit tests. The structured completion analysis is self-contained and matches the oracle's verdicts on its entire regression suite.
- **Invalid `commentPattern` falls back to the default pattern** (ESLint throws at config time; this host has no per-rule config-error channel). Falling back keeps standard markers working instead of silently reporting marked code. Documented in the rule source and pinned by a test.
- **Custom patterns compile as RE2** (Go) rather than JS regex; the option's realistic pattern space (`"break omitted"`, `"break[\s\w]+omitted"`) is identical in both engines.

## Test plan (CI validates; no local runs per instructions)

- 68 Go unit tests under `packages/lint/test/rules/control-flow/no_fallthrough_*` — one scenario per file, each with a negative twin one property away: marker spellings vs unrelated/mispositioned/directive comments, in-block marker positions, if/else joins, try/catch/finally composition, loop folding boundaries, labeled break/continue, nested function boundaries (function expression / arrow / declaration), nested switch, empty-case + blank-line semantics, message texts, all three options including the invalid-pattern fallback, last-case and empty-switch boundaries, unreachable-code handling.
- The shared corpus fixture `tests/test-lint/src/cases/no-fallthrough.ts` (real `ttsc` CLI run + Go engine twin) now covers, in one switch: an unmarked reachable fallthrough that still reports (its trailing `// @ts-ignore` comment proves unrelated comments do not suppress — note the corpus' own `// expect:` annotation contains "fallthrough" and would otherwise be a marker), a `// falls through` transition, and an if/else-terminated case.

### DIFF
--- a/packages/lint/README.md
+++ b/packages/lint/README.md
@@ -325,7 +325,7 @@ Source: [ESLint core rules](https://eslint.org/docs/latest/rules/).
 - [`object-shorthand`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/object-shorthand.ts): requires object property shorthand where possible.
 - [`operator-assignment`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/operator-assignment.ts): prefers compound assignment operators.
 - [`prefer-arrow-callback`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/prefer-arrow-callback.ts): reject `function() { ... }` expressions passed as callback arguments. Prefer the arrow form.
-- [`prefer-const`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/prefer-const.ts): prefers `const` for `let` bindings that are never reassigned.
+- [`prefer-const`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/prefer-const.ts): prefers `const` for lexical `let` bindings that are never reassigned, including declaration-only and destructured bindings with ESLint-compatible options.
 - [`prefer-destructuring`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/prefer-destructuring.ts): reject single-property and single-index variable declarations (`const a = obj.a`, `const x = arr[0]`) that destructuring would replace verbatim.
 - [`prefer-exponentiation-operator`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/prefer-exponentiation-operator.ts): prefers `**` over `Math.pow`.
 - [`prefer-for-of`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/prefer-for-of.ts): prefers `for...of` for simple array iteration.
--- a/packages/lint/README.md
+++ b/packages/lint/README.md
@@ -253,7 +253,7 @@ Source: [ESLint core rules](https://eslint.org/docs/latest/rules/).
 - [`no-extend-native`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/no-extend-native.ts): reject assignments to a built-in prototype such as `Array.prototype.foo = bar`.
 - [`no-extra-bind`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/no-extra-bind.ts): rejects unnecessary `.bind()` calls.
 - [`no-extra-boolean-cast`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/no-extra-boolean-cast.ts): rejects redundant boolean casts.
-- [`no-fallthrough`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/no-fallthrough.ts): rejects unmarked `switch` fallthrough.
+- [`no-fallthrough`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/no-fallthrough.ts): rejects `switch` cases whose end is reachable and that lack an intentional `// falls through` comment before the next label.
 - [`no-func-assign`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/no-func-assign.ts): rejects reassignment of function declarations.
 - [`no-implicit-coercion`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/no-implicit-coercion.ts): reject common implicit-coercion idioms (`!!x`, `+x`, `"" + x`) in favor of the explicit `Boolean(x)` / `Number(x)` / `String(x)` conversions.
 - [`no-import-assign`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/no-import-assign.ts): rejects writes to imported bindings (including `ns.x = ...` for namespace imports).
--- a/packages/lint/linthost/ast_helpers.go
+++ b/packages/lint/linthost/ast_helpers.go
@@ -336,8 +336,8 @@ func (w *descendantsWalker) visitChild(child *shimast.Node) bool {
   return false
 }
 
-// assignmentTargetNames collects the identifier names written by an
-// assignment's left-hand side. A bare Identifier yields one name. A
+// assignmentTargetIdentifiers collects the Identifier nodes written by an
+// assignment's left-hand side. A bare Identifier yields one node. A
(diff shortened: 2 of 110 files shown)


samchon:
Integration update: rebased all four #411 commits onto PR #423 final head 306c8a92 and retargeted this PR to ix/lint-no-unused-expressions. The three expected core-option conflicts were resolved as a strict union: 
o-duplicate-imports, 
o-unused-expressions, and 
o-fallthrough all retain their option interfaces, typed rule settings, and options-map entries. Outside those three shared type files (and the already-exported shared index), the stable patch ID is unchanged (dc1a9eca...); the last three commit patch IDs are also unchanged. No formatter or local build/test/typecheck/lint/static-analysis/repro command was run. GitHub CI owns verification. New head: 8f31c69e.

PR #429, opened 2026-07-13
fix(lint): align ban-ts-comment with recommended defaults and directives

### LINKED ISSUE #415: fix(lint): align ban-ts-comment with recommended defaults and directives
## Summary

`@ttsc/lint`'s `typescript/ban-ts-comment` rule inverts two parts of the upstream recommended/default contract:

- it rejects a sufficiently described `@ts-expect-error`, which is allowed by default;
- it does not detect `@ts-nocheck`, which is banned by default.

The paired false positive/false negative reproduces with published `@ttsc/lint@0.18.4` and current `master` (`fbbad7caeb44a16c77808c350ffaeeed608cabd1`). No tsconfig transformer plugin is used.

## Official default and option semantics

With the scalar rule setting, the official recommended defaults are:

```ts
{
  minimumDescriptionLength: 3,
  "ts-check": false,
  "ts-expect-error": "allow-with-description",
  "ts-ignore": true,
  "ts-nocheck": true,
}
```

Here `true` means report the directive and `false` means allow it. Therefore:

- `@ts-check` is allowed because it enables checking;
- `@ts-expect-error` is allowed with a description of at least three characters;
- `@ts-ignore` and `@ts-nocheck` are reported.

Each directive may also be configured as `boolean`, `"allow-with-description"`, or `{ descriptionFormat }`, with `minimumDescriptionLength` applying to description-required forms.

Reference: https://typescript-eslint.io/rules/ban-ts-comment/

The current implementation iterates only `SourceFile.CommentDirectives`, whose compiler representation covers error-suppression directives, then unconditionally reports both `Ignore` and `ExpectError`:

```go
for _, directive := range ctx.File.CommentDirectives {
  switch directive.Kind {
  case shimast.CommentDirectiveKindIgnore:
    // always report
  case shimast.CommentDirectiveKindExpectError:
    // always report
  }
}
```

Source: `packages/lint/linthost/rules_ts.go:253-269`.

That data source does not provide `@ts-check`/`@ts-nocheck`, and the rule has no typed options entry in `ITtscLintRuleOptionsMap`.

`lint.config.json`:

```json
{ "rules": { "typescript/ban-ts-comment": "error" } }
```

`case.ts`:

```ts
// @ts-nocheck

// @ts-expect-error: intentional incompatible fixture assignment
const expectedFailure: number = "value";

const uncheckedFailure: number = "also unchecked";

export { expectedFailure, uncheckedFailure };
```

```bash
```

```text
case.ts:3:1 - error TS11990: [typescript/ban-ts-comment] Do not use `@ts-expect-error` because it alters compilation errors.
Found 1 error ...
exit 2
```

Expected: one finding at line 1 for `@ts-nocheck`, and no finding for the described `@ts-expect-error`. `@ts-nocheck` intentionally suppresses the fixture's normal compiler diagnostics, so the output isolates the lint contract.

Parse all line and block comments with the compiler scanner/comment ranges, classify the four supported directives, and evaluate a normalized typed option object. Do not rely solely on `SourceFile.CommentDirectives`, because it intentionally represents only directives that suppress the next compiler error.

The evaluator should keep these concerns separate:

1. directive recognition and exact source range;
2. per-directive policy (`false`, `true`, description-required, or description-format);
3. description extraction, minimum length, and optional format matching;
4. diagnostic/fix choice, including the safe `@ts-ignore` to `@ts-expect-error` suggestion where applicable.

Add the upstream-compatible options shape to the typed rule-options map if tuples are accepted. The scalar form must use the documented recommended defaults rather than a stricter undocumented policy.

### Forbidden shortcuts

- Do not merely add an `@ts-nocheck` string check while leaving described `@ts-expect-error` broken.
- Do not ban all text containing `@ts-`; only actual supported directive comments are in scope.
- Do not hard-code a single description phrase or delimiter.
- Do not mutate tsconfig or add transformer plugins to make the rule run.

## Acceptance criteria

- The reproduction reports only `@ts-nocheck`.
- Scalar defaults allow `@ts-check`, allow a sufficiently described `@ts-expect-error`, and report `@ts-ignore`/`@ts-nocheck`.
- Missing and too-short `@ts-expect-error` descriptions report with precise ranges.
- Line comments and block comments follow the same directive policy.
- Every directive-specific boolean, `allow-with-description`, `descriptionFormat`, and `minimumDescriptionLength` option is typed and honored if options tuples are supported.
- Near misses such as prose mentioning `@ts-ignore`, longer identifiers, and non-directive strings remain negative controls.
- Safe suggestions never rewrite a directive in a way that hides an unrelated compiler error.
- Focused native tests and a real no-plugin CLI fixture cover defaults and options.

## Existing issue review

- #137 was a closed rule-coverage inventory and did not track the implemented rule's directive/default mismatch.
- No open or closed issue containing `ban-ts-comment`, `ts-nocheck`, or described `ts-expect-error` covers this behavior.
- #401 is unrelated.

samchon:
## Intent

`typescript/ban-ts-comment` inverted two halves of the upstream recommended contract (#415): it reported described `@ts-expect-error` comments that the preset allows, and it never detected `@ts-nocheck` because it read only `SourceFile.CommentDirectives`, whose compiler representation carries error-suppression directives alone.

## Scope

**Rule engine (`packages/lint/linthost`)**

- `rules_ban_ts_comment.go` replaces the `CommentDirectives` loop with a faithful port of the upstream rule: comments are re-lexed with the raw scanner, classified with the upstream/compiler grammar (2–3 slash line pragmas for `check`/`nocheck`; line comments plus block-comment **last** lines for `expect-error`/`ignore`; a first-statement line gate for `nocheck`), and evaluated against per-directive policies.
- Options mirror the upstream `DirectiveConfig` union — `boolean`, `"allow-with-description"`, `{ descriptionFormat }` — plus `minimumDescriptionLength`, merged per key over the recommended defaults (`ts-check: false`, `ts-expect-error: "allow-with-description"`, `ts-ignore: true`, `ts-nocheck: true`, min length 3).
- `comment_scan.go` extracts the template-rescanning comment-token loop from the inline-directive parser so both consumers share the one subtle implementation.
- `string_length.go` ports upstream `getStringLength` (grapheme clusters, not runes): hand-rolled UAX #29 subset in the style of `display_width.go`, no new dependency. A family-emoji description counts as one character, exactly as the upstream regression tests pin.
- Banned `@ts-ignore` reports the upstream "use @ts-expect-error instead" message with the rewrite attached as an autofix edit.

**Typed options transport** — `ITtscLintTypeScriptBanTsCommentRuleOptions` (new `ITtscLintTypeScriptRuleOptions.ts`), registered in `ITtscLintRuleOptionsMap`, and the rule's entry upgraded to `TtscLintRuleOptionsSetting<…>`.

**Docs** — `packages/lint/README.md` and `website/src/content/docs/lint/rules/typescript.mdx` now describe the defaults and options.

## Deviations from the issue, with reasoning

- **"Longer identifiers remain negative controls"**: not fully — the oracle disagrees with the issue here. Both the TypeScript compiler and the upstream rule match `@ts-expect-error`/`@ts-ignore` as prefixes (`@ts-ignored…` IS an effective directive; the description group starts immediately after the directive word), so the port follows the upstream regexes verbatim. Truncated names (`@ts-expect`), prose mentions, and prefixed text stay negative controls and are pinned as such.
- **Suggestion vs. fix**: upstream ships the ignore→expect-error rewrite as an ESLint *suggestion*. ttsc's finding transport has a single fix channel (surfaced as editor code actions and applied by `ttsc fix`), so the rewrite rides there. It satisfies the issue's safety criterion: it can never hide a compiler error — where `@ts-ignore` suppressed one, `@ts-expect-error` suppresses the same one; where the next line was clean, the compiler now reports the unused directive instead of staying silent. No fix is attached to description-length findings.
- **Out-of-union option values** (upstream schema would reject the config): this host has no schema layer, so the rule keeps the directive's documented default. An uncompilable `descriptionFormat` keeps the description-length gate and drops only the unenforceable pattern. Both choices are pinned by tests. `descriptionFormat` is evaluated with Go RE2 (documented in the option's docstring); the canonical upstream patterns are RE2-compatible.

## Test plan

- 26 new Go unit tests under `packages/lint/test/rules/comments-directives/`, one case per file, expectations derived from the upstream `ban-ts-comment` regression suite: every directive kind, every option arm, exact minimum-length boundary, raw-description format matching (including the leading-whitespace mismatch), length-before-format ordering, pragma slash counts, block-comment last-line grammar across all five ECMAScript line terminators, grapheme-cluster counting (family emoji, combining marks), fix snapshots (line, block, many-slash), inline-disable interplay, string/template negative controls with post-template offset verification, and exact message pins.
- The e2e corpus fixture `tests/test-lint/src/cases/ban-ts-comment.ts` now mirrors the issue's no-plugin reproduction through the real ttsc binary: `@ts-nocheck` and `@ts-ignore` report, described `@ts-expect-error` and a prose mention stay silent (Go corpus twin updated in lockstep).
- The typed options join the compile-time autocomplete test (`test_lib_index_d_ts_rule_options_autocomplete_per_rule.ts`).

Closes #415


samchon:
Integration repair: merged PR #426 head into this branch and retargeted the PR to `fix/lint-prefer-as-const`, preserving both rule implementations and reconciling their shared TypeScript rule source, public types, README, and website docs once. The combined diff received a clean full self-review and `pnpm format`; local build/test/typecheck/lint/static-analysis commands remain intentionally delegated to GitHub CI. Head: `c9d5a8f6`.

samchon:
## Master alignment follow-up

PR #429 is now aligned with refreshed `master` at `70f4cb341bbf85d96b132f91296e6778b0a5d60c` through the non-destructive merge commit `b3cb5b22a4f7041f8bcf2b77549e4d0e5a987d12`.

- [x] merged `origin/master` without rewriting the published branch history
- [x] retained both parents (`c9d5a8f6` feature head + `70f4cb34` refreshed master)
- [x] merge completed without conflicts
- [x] compared the resulting master-relative file set with the four ban-ts-comment commits: 49 files on each side, zero file-set delta
- [x] confirmed all dedicated prefer-as-const source/tests/fixtures are identical to refreshed master
- [x] inspected shared TypeScript lint files; their remaining master-relative hunks are ban-ts-comment-specific
- [x] retargeted/confirmed the PR base as `master`
- [x] pushed and confirmed the remote head is `b3cb5b22a4f7041f8bcf2b77549e4d0e5a987d12`
- [x] confirmed a fresh CI rollup exists for the new head (currently queued/in progress); intentionally did not wait
- [x] did not run formatter, build, tests, typecheck, lint, static analysis, or repro, per coordinator instruction

PR #433, opened 2026-07-13
fix(lint): preserve generic-arrow syntax in type-constraint fixes

### LINKED ISSUE #424: fix(lint): preserve TSX generic-arrow syntax in no-unnecessary-type-constraint fixes
## Summary

`@ttsc/lint`'s `typescript/no-unnecessary-type-constraint` autofix corrupts a single-parameter generic arrow function in TSX. It rewrites the valid, unambiguous source `<T extends unknown>(...) => ...` to `<T>(...) => ...`; in `.tsx`, that token sequence is parsed as an unterminated JSX element. `ttsc fix` writes the broken source to disk and then exits 2 with three TypeScript parse errors.

Published `@ttsc/lint@0.18.4` and current `master` (`fbbad7caeb44a16c77808c350ffaeeed608cabd1`) behave identically. The fixture uses package auto-discovery and contains no `compilerOptions.plugins` entry.

## Official syntax and fix contract

The official `@typescript-eslint/no-unnecessary-type-constraint` rule diagnoses `extends any` and `extends unknown`. Its removal suggestion explicitly handles generic-arrow disambiguation:

- for a single type parameter in `.tsx`, `.mts`, or `.cts`;
- when there is no existing trailing comma and no type-parameter default;
- removal produces `<T,>` rather than `<T>`.

Multiple parameters, an existing comma, or a default such as `<T = string>` already disambiguate the grammar and do not need an inserted comma. Ordinary `.ts` can safely become `<T>`.

Reference and release source:

- https://typescript-eslint.io/rules/no-unnecessary-type-constraint/
- https://github.com/typescript-eslint/typescript-eslint/blob/v8.63.0/packages/eslint-plugin/src/rules/no-unnecessary-type-constraint.ts

`@ttsc/lint` publicly describes this rule as “Autofixable: drops the constraint” in `packages/lint/src/structures/rules/ITtscLintTypeScriptRules.ts:615-621`. Its implementation always deletes from the type-parameter name's end through the constraint's end:

```go
TextEdit{Pos: name.End(), End: param.Constraint.End(), Text: ""}
```

Source: `packages/lint/linthost/rules_gap.go:156-185`.

The edit has no awareness of the source's grammar mode or the enclosing generic arrow, so it removes the only token that distinguished TypeScript generics from JSX.

`lint.config.json`:

```json
{ "rules": { "typescript/no-unnecessary-type-constraint": "error" } }
```

`tsconfig.json`:

```json
{
  "compilerOptions": {
    "noEmit": true,
    "strict": true,
    "target": "ES2022",
    "jsx": "react-jsx",
    "module": "NodeNext",
    "moduleResolution": "NodeNext"
  },
  "files": ["case.tsx"]
}
```

`case.tsx` before fixing:

```tsx
const identity = <T extends unknown>(value: T): T => value;

console.log(identity(1));

export {};
```

Run:

```bash
ttsc fix --project tsconfig.json
```

```text
case.tsx:1:19 - error TS17008: JSX element 'T' has no corresponding closing tag.
case.tsx:1:36 - error TS1382: Unexpected token. Did you mean `{'>'}` or `&gt;`?
case.tsx:6:1 - error TS1005: '</' expected.
Found 3 errors ...
exit 2
```

The command leaves this invalid file on disk:

```tsx
const identity = <T>(value: T): T => value;
```

Expected fixed source and result:

```tsx
const identity = <T,>(value: T): T => value;
```

```text
exit 0
```

Compute the edit from both the type-parameter declaration and its syntactic context:

1. determine whether the parameter belongs to an arrow function;
2. determine whether the source mode requires generic-arrow disambiguation (`.tsx`, `.mts`, and `.cts` under the compiler's canonical script-kind/file-mode rules);
3. add a comma only for a single parameter that has neither a default nor an existing comma;
4. otherwise retain the current constraint removal;
5. if the host cannot prove the replacement parses in the same language mode, keep the diagnostic but decline the autofix.

The implementation should use AST parents, source-file mode, and compiler token ranges. It should not infer an arrow from nearby text or append commas to every removed constraint.

### Forbidden shortcuts

- Do not scan source text for `=>`, `<T`, or a fixture-shaped declaration.
- Do not add a comma unconditionally; that changes ordinary output and can interfere with comments/tokens around multi-parameter declarations.
- Do not special-case the identifier `T`, function name, or `.tsx` fixture path.
- Do not hide the diagnostic merely because a safe fix is unavailable.
- Do not add `compilerOptions.plugins` to the reproduction.

## Acceptance criteria

- The reproduction becomes `<T,>(...) => ...`, remains parseable, and `ttsc fix` exits 0.
- Equivalent single-parameter generic arrows in `.mts` and `.cts` remain parseable after fixing.
- An ordinary `.ts` generic arrow becomes `<T>(...) => ...` without an unnecessary comma.
- Multiple type parameters, an existing trailing comma, and a type-parameter default each preserve their already-unambiguous form.
- Non-arrow generic declarations (functions, classes, interfaces, type aliases, and methods) remove only the constraint.
- `extends any` and `extends unknown` both follow the same grammar-safe path.
- Comments around the constraint and comma have focused preservation tests.
- Native fix tests and a real no-plugin `ttsc fix` CLI fixture assert both final bytes and the post-fix compiler result.

## Existing issue review

- #137 was a closed rule-coverage inventory and did not cover autofix safety for this implemented rule.
- Searches across every open and closed ttsc issue for `no-unnecessary-type-constraint`, TSX generic arrows, and constraint autofixes found no existing owner.
- #401 concerns `no-duplicate-imports`; #364 and #119 concern other rules' fixes and are not duplicates.

samchon:
## Intent

Keep `typescript/no-unnecessary-type-constraint` autofixes from producing invalid single-parameter generic arrows in TSX and explicit Node module source modes.

## Root cause

The fixer always deleted ` extends any` / ` extends unknown`. In `.tsx`, `.mts`, and `.cts` single-parameter arrow functions, the resulting `<T>` form needs a trailing comma to remain grammar-safe.

## Changes

- preserve `<T,>` only for affected single-parameter generic arrows
- retain minimal output for ordinary TypeScript, multiple parameters, defaults, existing commas, and non-arrow declarations
- add filename-aware native fixer snapshots for positive, negative, comment, and boundary cases
- add a no-plugin real CLI fixture that compares exact bytes and relies on `ttsc fix`'s final compiler-diagnostic pass
- document the grammar-safe autofix behavior

## Verification

- `pnpm format`
- local build, test, typecheck, lint, static-analysis, and reproduction commands were intentionally not run; GitHub CI is the execution authority for this issue batch

Closes #424


samchon:
Integration update: rebased the single #424 change onto the latest ix/lint-ban-ts-comment head (c9d5a8f6) so this PR now follows PR #429 without pulling unrelated post-fork master commits into the stack. The resulting PR delta remains limited to the generic-arrow grammar fix, its boundary tests, the no-plugin CLI fixture, and matching docs. No formatter or local build/test/typecheck/lint/static-analysis/repro command was run; GitHub CI owns verification. New head: 93366565.

PR #434, opened 2026-07-13
fix(lint): implement full no-unsafe-assignment semantics

### LINKED ISSUE #418: fix(lint): implement full no-unsafe-assignment semantics and unknown boundary
## Summary

`@ttsc/lint`'s `typescript/no-unsafe-assignment` rule implements only the narrow case of a raw `any` expression assigned to an explicitly annotated variable or through plain `=`. It misses the default upstream rule's inferred variables, destructuring, default parameters, class fields/auto-accessors, and recursive generic type-argument checks. It also incorrectly rejects the explicitly allowed assignment from `any` to `unknown`.

Published `@ttsc/lint@0.18.4` and current `master` (`fbbad7caeb44a16c77808c350ffaeeed608cabd1`) produce the same two diagnostics in the no-plugin fixture below.

## Official contract

`@typescript-eslint/no-unsafe-assignment` is not configurable. It reports `any` escaping into variables and properties even when the target type is inferred, handles array/object destructuring and default initializers, and recursively compares generic type arguments such as `Set<Set<any>>` assigned to `Set<Set<string>>`.

Assignments from `any` to `unknown` are intentionally allowed because `unknown` preserves the requirement to narrow before use. That exception also applies recursively in documented safe forms such as `Set<any>` to `Set<unknown>`.

Reference and examples: https://typescript-eslint.io/rules/no-unsafe-assignment/

The current implementation in `packages/lint/linthost/rules_ts_no_unsafe_assignment.go:30-101` visits only `VariableDeclaration` and `BinaryExpression`, then:

- returns immediately when a variable declaration has no explicit type annotation;
- asks only whether the complete RHS type has the raw `Any` flag;
- does not descend into binding patterns, generic arguments, parameters, or property declarations;
- exempts only an `any` target, not an `unknown` target.

This contradicts both the upstream behavior and the source comment saying `unknown` is intentionally not flagged.

`lint.config.json`:

```json
{ "rules": { "typescript/no-unsafe-assignment": "error" } }
```

`case.ts`:

```ts
declare const leaked: any;

const explicit: string = leaked;
const allowedUnknown: unknown = leaked;
const inferred = leaked;
const [destructured] = leaked;

function withDefault(value = leaked): unknown {
  return value;
}

class Container {
  public value = leaked;
  public accessor accessorValue = leaked;
}

const genericTarget: Set<Set<string>> = new Set<Set<any>>();

export {
  allowedUnknown,
  Container,
  destructured,
  explicit,
  genericTarget,
  inferred,
  withDefault,
};
```

Use a strict ES2022/NodeNext `tsconfig.json` with `noEmit: true` and no `plugins`. Run:

```bash
```

```text
case.ts:3:7 - error TS15517: [typescript/no-unsafe-assignment] Unsafe assignment of a value typed as `any` ...
case.ts:4:7 - error TS15517: [typescript/no-unsafe-assignment] Unsafe assignment of a value typed as `any` ...
Found 2 errors ...
exit 2
```

Expected:

- findings at lines 3, 5, 6, 8, 13, 14, and 17;
- no finding at line 4 because `any` to `unknown` is an explicit safe boundary.

Line 3 is a positive control that proves the rule is active; line 4 is the paired negative control.

Implement assignment analysis around source/receiver type pairs rather than raw-node kinds:

1. enumerate assignment sites covered by the upstream rule: variable and property declarations, assignment expressions, array/object binding patterns, parameter/constructor-parameter defaults, and relevant destructuring assignments;
2. detect a direct `any` source even when the receiver type is inferred, while keeping `unknown` receivers safe;
3. recursively compare structurally corresponding type arguments and tuple/array element types, carrying the path at which `any` escapes;
4. handle unions, intersections, aliases, constraints, mapped types, and nested generics without reducing the check to printed type names;
5. use one shared helper across declaration and reassignment shapes so identical type pairs cannot disagree by syntax.

Recursive comparison needs cycle protection and a principled variance/compatibility boundary. Reuse checker identity and type-argument APIs; do not recursively walk arbitrary object properties as a proxy for generic compatibility.

### Forbidden shortcuts

- Do not require explicit target annotations; inferred assignment is a core purpose of this rule.
- Do not flag all uses of `any`; this rule concerns assignment boundaries and deliberately allows `any` to `unknown`.
- Do not special-case `Set`, `Map`, arrays, or fixture type names.
- Do not compare rendered type strings or scan source text for `any`.
- Do not add tsconfig transformer plugins.

## Acceptance criteria

- The reproduction emits exactly the seven expected findings and keeps the `unknown` assignment clean.
- Direct annotated, inferred, reassignment, default-parameter, constructor-parameter-property, class-field, auto-accessor, and array/object destructuring sites match upstream behavior.
- Recursive generic mismatches report at arbitrary depth; corresponding `unknown` targets remain allowed.
- Safe same-type generic assignments and `unknown`-narrowing boundaries remain negative controls.
- Aliases, unions, intersections, tuples, readonly arrays, mapped types, generic constraints, and recursive types have focused positive and negative tests.
- Each unsafe boundary reports once with a useful source range; destructuring does not duplicate diagnostics for the same leaf.
- The rule remains non-configurable and runs in a real no-plugin CLI fixture as well as native tests.

## Existing issue review

- #137 was a closed rule-coverage inventory and did not track semantic completeness of this implemented rule.
- No open or closed issue containing `no-unsafe-assignment`, inferred `any`, generic `any`, or `any` to `unknown` covers these defects.
- #401 is unrelated.

samchon:
## Intent

Closes #418 by bringing the native `typescript/no-unsafe-assignment` rule to the full assignment-site and generic-comparison semantics of typescript-eslint 8.63.0 while preserving `unknown` as the safe narrowing boundary.

## Scope

- cover annotated and inferred declarations, reassignments, defaults, class fields and auto-accessors, contextual object/JSX properties, array spreads, and array/object destructuring
- compare nested type arguments only for the same generic target, with cycle protection and the upstream empty `new Map()` exception
- add one-case-per-file type-aware Go coverage plus the expanded lint corpus fixture
- add a real package auto-discovery integration scenario with no `compilerOptions.plugins` entry and exact issue-reproduction diagnostic lines
- document the complete rule contract in the package README, public config type, and website guide

## Deferred

No semantic behavior is deferred. The upstream-only specialized `this`/`noImplicitThis` remediation wording is not ported because the native lint context does not expose compiler options; an unsafe `this` assignment still reports through the normal diagnostic.

## CI-only test plan

Local executable validation was prohibited for this task. CI should run the canonical `pnpm test`, which covers the Go lint scenarios, typechecking, `@ttsc/test-lint`, and the `@ttsc/test-ttsc` no-plugin integration corpus. Repository Prettier and the two-space Go formatter were applied to every changed file before commit.


samchon:
Integration update: rebased both #418 commits onto the latest ix/lint-ban-ts-comment head (c9d5a8f6) so this PR follows PR #429 without carrying unrelated post-fork master history. Stable patch IDs for both commits are unchanged across the rebase (69000d3f... and eee9bff8...), and the combined tree retains both the ban-ts-comment option surface/docs and the complete no-unsafe-assignment delta. No formatter or local build/test/typecheck/lint/static-analysis/repro command was run during this integration step; GitHub CI owns verification. New head: 39435ef0.

PR #436, opened 2026-07-13
fix(lint): analyze prefer-const by lexical binding

### LINKED ISSUE #410: fix(lint): make prefer-const analyze lexical bindings and default cases
## Summary

`@ttsc/lint`'s `prefer-const` analysis is file-global and name-based instead of binding-based. A reassignment of `value` in one function suppresses a distinct never-reassigned `value` in another function. The implementation also intentionally skips declaration-then-assignment and destructuring cases that ESLint reports under its default options.

One fixture containing all three gaps exits 0 with published `@ttsc/lint@0.18.4` and current `master` (`fbbad7caeb44a16c77808c350ffaeeed608cabd1`). No tsconfig transformer plugin is used.

## Official default semantics

ESLint `prefer-const` reports bindings that are never reassigned after their initial assignment. Its documented default options are:

```json
{
  "destructuring": "any",
  "ignoreReadBeforeAssign": false
}
```

The official examples include both `let b; b = 0;` and partially const-eligible destructuring.

Reference: https://eslint.org/docs/latest/rules/prefer-const

`@ttsc/lint` advertises the same core rule ID, but `packages/lint/linthost/rules_var.go:486-563`:

- collects candidates across the entire source file;
- stores assignments in `map[string]bool` keyed only by identifier text;
- intentionally excludes destructuring because `identifierText` returns no leaf binding;
- intentionally excludes declaration-only bindings unless they are `for...in/of` variables.

This is not merely fewer suggestions. The global name map makes the result depend on unrelated same-spelled bindings in other scopes.

`lint.config.json`:

```json
{ "rules": { "prefer-const": "error" } }
```

`case.ts`:

```ts
function shouldReport(): number {
  let value = 1;
  return value;
}

function sameNameButReassigned(): number {
  let value = 1;
  value += 1;
  return value;
}

let assignedLater: number;
assignedLater = 1;

const input = { first: 1, second: 2 };
let { first, second } = input;
first += 1;

console.log(shouldReport(), sameNameButReassigned(), assignedLater, first, second);
```

Use a strict ES2022/NodeNext `tsconfig.json` with `noEmit: true` and no `plugins` field, then run:

```bash
```

```text
exit 0
```

Expected default findings:

1. the `value` binding in `shouldReport` (the other function's distinct `value` is irrelevant);
2. `assignedLater`, whose first and only assignment establishes its value;
3. `second` in the destructuring declaration, while `first` remains mutable.

The `value` in `sameNameButReassigned` and `first` are negative controls.

Track lexical binding identity and write history, not identifier spelling. Use checker symbols when available or a rule-owned lexical scope/binding table that distinguishes source, function, block, catch, loop, class-static-block, parameter, and destructuring bindings. Then model:

- declaration initialization;
- first later assignment versus reassignment;
- reads before first assignment according to `ignoreReadBeforeAssign`;
- compound/update/destructuring writes;
- `for...in/of` per-iteration definitions;
- the default `destructuring: "any"` policy.

Autofix eligibility should remain stricter than detection. When a declaration-only binding would require moving/merging the first assignment, report without an unsafe text rewrite unless the transform is proven scope- and comment-preserving.

### Forbidden shortcuts

- Do not qualify keys as `functionName + variableName`; anonymous/nested/block scopes still collide.
- Do not add more text-based exceptions for destructuring or selected declaration shapes.
- Do not suppress all same-named bindings when any one is reassigned.
- Do not rewrite declarations across statements without control-flow and comment preservation.

## Acceptance criteria

- The reproduction yields exactly the three expected findings.
- Same-spelled bindings in sibling, nested, and shadowing scopes are analyzed independently.
- Declaration-then-single-assignment follows official `ignoreReadBeforeAssign` default/option behavior.
- Destructuring follows official `destructuring: "any"` and `"all"` behavior if options are exposed; the default is `"any"`.
- Compound assignments, updates, destructuring assignments, loop targets, and closure writes count against only the binding they resolve to.
- Safe existing single-declaration autofixes remain; nonlocal rewrites are detection-only unless formally proven safe.
- Focused native tests and a real no-plugin CLI fixture cover positive and negative controls.

## Existing issue review

- #119 fixed a false positive when a binding is reassigned through destructuring. It did not replace the file-global string map, add partial destructuring candidates, or handle first later assignments.
- #137 was a closed rule-coverage inventory, not semantic parity for this implementation.
- No current issue owns these binding-identity/default-coverage gaps.

samchon:
## Summary

- resolve `prefer-const` declarations and writes by TypeScript checker symbol, so sibling, nested, shadowed, and closure-captured bindings stay independent
- implement ESLint-compatible declaration-only, `ignoreReadBeforeAssign`, and `destructuring: any | all` behavior while keeping unsafe nonlocal rewrites diagnostic-only
- add checker-backed native coverage, fixer boundaries, typed configuration coverage, a real no-plugin CLI fixture, and rule documentation

## Root cause

The rule stored writes in a file-global `map[string]bool`, which conflated unrelated same-spelled bindings. It also excluded declaration-only and destructured bindings before write-history analysis.

## Validation

GitHub CI is the source of build, test, typecheck, lint, static-analysis, and reproduction evidence for this change. Per the task contract, those commands were not run locally.

Closes #410

### DIFF
--- a/packages/lint/README.md
+++ b/packages/lint/README.md
@@ -325,7 +325,7 @@ Source: [ESLint core rules](https://eslint.org/docs/latest/rules/).
 - [`object-shorthand`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/object-shorthand.ts): requires object property shorthand where possible.
 - [`operator-assignment`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/operator-assignment.ts): prefers compound assignment operators.
 - [`prefer-arrow-callback`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/prefer-arrow-callback.ts): reject `function() { ... }` expressions passed as callback arguments. Prefer the arrow form.
-- [`prefer-const`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/prefer-const.ts): prefers `const` for `let` bindings that are never reassigned.
+- [`prefer-const`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/prefer-const.ts): prefers `const` for lexical `let` bindings that are never reassigned, including declaration-only and destructured bindings with ESLint-compatible options.
 - [`prefer-destructuring`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/prefer-destructuring.ts): reject single-property and single-index variable declarations (`const a = obj.a`, `const x = arr[0]`) that destructuring would replace verbatim.
 - [`prefer-exponentiation-operator`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/prefer-exponentiation-operator.ts): prefers `**` over `Math.pow`.
 - [`prefer-for-of`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/prefer-for-of.ts): prefers `for...of` for simple array iteration.
--- a/packages/lint/linthost/ast_helpers.go
+++ b/packages/lint/linthost/ast_helpers.go
@@ -336,8 +336,8 @@ func (w *descendantsWalker) visitChild(child *shimast.Node) bool {
   return false
 }
 
-// assignmentTargetNames collects the identifier names written by an
-// assignment's left-hand side. A bare Identifier yields one name. A
+// assignmentTargetIdentifiers collects the Identifier nodes written by an
+// assignment's left-hand side. A bare Identifier yields one node. A
 // destructuring-assignment target — parsed as an ArrayLiteralExpression
 // (`[a, b] = …`) or ObjectLiteralExpression (`({a} = …)`) rather than a
 // binding pattern — is walked so every nested write position is counted:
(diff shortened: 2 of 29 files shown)


samchon:
Amended the existing commit message to satisfy the Lore subject convention: fix(lint): .... The tree and PR patch are unchanged; new head: f97306cc.

samchon:
Integration update: rebased #410 onto PR #427 head 8f31c69e and retargeted this PR to ix/lint-no-fallthrough. Four shared conflicts were resolved as a strict union: the no-duplicate-imports, no-unused-expressions, no-fallthrough, and prefer-const option interfaces/settings/map entries all remain; the shared autocomplete fixture retains every predecessor case plus prefer-const's valid and invalid shapes. The stable patch ID outside those four shared files is unchanged (3cad43f...). The original cli-2 worktree's unrelated dirty state was left untouched; integration used a clean separate worktree. No formatter or local build/test/typecheck/lint/static-analysis/repro command was run. GitHub CI owns verification. New head: 5eea6bd5.

samchon:
Follow-up correctness fixes are now pushed.

Root causes:
- NonNullExpression assignment targets and parenthesized update operands were not normalized to their identifier targets.
- TypeScript resolves a shorthand property name to the property symbol through GetSymbolAtLocation, not the lexical value binding.
- A real binary assignment nested inside a shorthand destructuring default was incorrectly classified as part of the outer default target.

Fixes:
- Unwrap non-null and parenthesized write targets through the shared assignment-target walker.
- Resolve shorthand identifiers through GetShorthandAssignmentValueSymbol for writes, groups, initialization checks, and read-before-assignment tracking.
- Count nested writes in shorthand default expressions as real reassignments.
- Add focused regression coverage for all four boundaries.

Validation:
- GOFLAGS=-run=^(TestEngineRequiresTypeCheckerForPreferConst|TestRuleCorpusPreferConst|TestPreferConst.*|TestFixPreferConst.*)$ node scripts/test-go-lint.cjs
- Result: pass (github.com/samchon/ttsc/packages/lint/linthost, 6.777s)

Commit: 7451466e8d06daacb92a4c915411e68dbd2b58a9

PR #437, opened 2026-07-13
feat(lint): preserve project identity across rule lifecycles

### LINKED ISSUE #406: feat(lint): expose logical and physical project identity plus project-scoped results to rule contributors
## Summary

`@ttsc/lint` contributors can currently inspect only one source file at a time. The host has the inputs needed to identify the selected project, but it does not preserve or expose one consistent identity:

- CLI/check/build resolve the selected tsconfig through `realpath`, default `project.root` to its physical parent, and pass that value to the native sidecar as `--cwd`;
- LSP passes its lexical launch cwd to the native sidecar while also passing the realpath-resolved tsconfig;
- an explicit `projectRoot` override and `TTSC_PLUGIN_CONFIG_DIR` are two additional, intentionally distinct channels; and
- the public Go `rule.Context` contains only `File`, `Checker`, `Severity`, and `Options`.

Add separate logical and physical project identities plus a project-scoped contributor lifecycle/result. This is generic support for project/config/non-AST rules. It must not encode one consumer's filesystem policy.

Audited on 2026-07-13 against published `@ttsc/lint@0.18.4` and current `samchon/ttsc` master. Searches for project rule/context/lifecycle contributor issues found no existing owner; #137, #140, #259, #368, and #118 are different closed scopes.

### Identity is lost or changes by surface

- `packages/ttsc/src/compiler/internal/project/resolveProjectConfig.ts` always returns a realpath-resolved config. Explicit `-p <file>`, `-p <directory>`, and discovered tsconfig/jsconfig paths all pass through `resolveRealPath`.
- `readProjectConfig.ts` defaults `root` to `dirname(realpathConfig)`. It preserves an explicit `projectRoot` only as a separate absolute override.
- `runBuild.ts` sends `--cwd=<project.root>` and the resolved config to native check/build sidecars. The original invocation cwd remains in the TypeScript execution object but is not transported as a separate logical identity.
- `runTtscserver.ts` resolves the project for plugin loading, while `lsp_native_plugin_source.go` sends the server's lexical cwd and resolved tsconfig to LSP sidecars. Equivalent CLI and LSP invocations therefore expose different `--cwd` meanings.
- `TTSC_PLUGIN_CONFIG_DIR` separately anchors lint-config discovery for generated wrapper configs. It is not necessarily invocation cwd, selected project root, or resolved config parent.

Replacing every physical path with a logical path is not the fix. The Program and canonical config identity still need resolved paths. The contributor needs both identities, named according to their meaning.

### The contributor API is file-only

`packages/lint/rule/rule.go` exposes:

~~~go
type Rule interface {
  Name() string
  Visits() []ast.Kind
  Check(ctx *Context, node *ast.Node)
}

type Context struct {
  File     *ast.SourceFile
  Checker  *checker.Checker
  Severity Severity
  Options  json.RawMessage
}
~~~

There is no project rule, project reporter, loaded-Program identity, or project result available to later file rules. Inferring from the first visited file is not a contract: files can be physicalized through links, dispatch order is not a project identity, and future parallel dispatch must remain valid. `os.Getwd`, environment reads, `sync.Once`, and package-global maps also fail for watch/LSP/multiple-project lifecycles.

### Project diagnostics have no complete transport

CLI renderers can already represent global process diagnostics, but the lint result has no first-class project finding. LSP is narrower:

- `linthost/lsp.go` filters `engine.Run(...)` through `filterFindingsForPath`, which drops findings whose `File` is nil;
- the sidecar `lsp-diagnostics` command returns only an array for the requested document; and
- `PluginSource.Diagnostics(LSPDocumentVersion)` can contribute only to the document currently being published.

Attaching the same project error to every requested source document would duplicate one project problem across all open files. A dedicated project publication is required.

### Identity and lifecycle

- Preserve the caller-selected lexical config path and logical project root before realpath resolution.
- Expose an opaque per-loaded-Program lifecycle identity; lexical invocation cwd; logical selected config path/root; physical resolved config path/Program root; explicit `projectRoot`; and explicit plugin-config origin as separate read-only fields.
- Do not derive any of those fields from `SourceFile.FileName()`, and do not make contributors read process cwd or environment variables.
- Transport the same meanings through CLI check/build/fix, watch rebuilds, the public compiler API, and LSP.
- Add a public `ProjectRule` registration path separate from node `Rule`. A project rule has a name and project check, but no `Visits`, synthetic source file, or node callback.
- Give the project check the loaded Program's read-only user-source set and checker. Run each enabled project rule once per loaded Program/check cycle before node dispatch, including a Program with zero user source files.
- Create a host-owned cycle result keyed by project-rule name. Expose its status/state read-only in every per-file contributor context and finalize it with its non-fix project findings after the cycle.
- Provide a cycle-scoped project reporter so a contributor's shared helper can reject a just-in-time operation before consuming bytes, mark its named state failed, and add a deduplicated deterministic project finding.
- Do not add a host-level file-rule dependency graph. A contributor may inspect its own named project state in a shared helper and decide which of its rules return early; unrelated contributors and AST-only rules remain independent.

The per-file query must distinguish `absent`, `off`, `not_evaluated`, `passed`, and `failed`. Missing state is not success. `off` is a resolved config state even though the project check does not run.

### Project-rule config resolution

Use the existing `rules` map; do not add a second config schema.

- A project rule may be configured only in an `ITtscLintConfig` object with no `files` selector. Reject a scoped mention, including `off` or an options tuple, because a project result has no file identity.
- Top-level `ignores` keep selecting source files only. They do not disable or duplicate a project rule in a global config object.
- Fold `extends` base-first. The last global declaration wins for severity, including `off`; the last explicit options tuple remains the project-wide options blob, and a later bare severity leaves it in force.
- Do not use `ResolveRules(firstFile)` or `EnabledRuleConfig()`. The former is file-scoped; the latter unions scoped entries and makes an earlier error sticky. Resolve a dedicated global project-rule map once.
- A project finding uses the final global severity. An `off` project rule produces no finding, but per-file contexts can still observe that it is off.

## Project finding transport

- CLI check/build/fix prints one project finding per evaluation before file findings; watch prints it once for the rebuild that produced it.
- Structured compiler results preserve it with `file: null` or an equally explicit project location and its contributor rule code/category.
- LSP separates document diagnostics from project publications in the sidecar and `PluginSource` result contract.
- Publish the project set once at the logical selected config (`tsconfig.json` or `jsconfig.json`) URI, at a zero-width start range with no document version. Do not copy it onto requested source documents.
- Merge with cached upstream diagnostics when the selected config URI also has upstream findings.
- Replace the current project set on re-evaluation, publish an empty set when it becomes clean, and clear the old URI before publishing a newly selected config URI. Existing generation checks must reject stale async results.
- Project findings never carry fixes or code actions.

## Acceptance criteria

- [ ] CLI cwd, `-p <config>`, `-p <directory>`, file-based discovery, explicit `projectRoot`, and plugin-config origin expose the correct separate logical channels.
- [ ] A project reached through a symlink/junction preserves the logical selected path and separately exposes the physical resolved config/Program root.
- [ ] CLI, LSP, watch, and public API give equivalent fields the same meaning.
- [ ] Two projects in one process receive distinct lifecycle state; watch/LSP reload cannot reuse stale state.
- [ ] A project rule executes for an empty user-source set.
- [ ] Per-file contexts distinguish absent, off, not-evaluated, passed, and failed.
- [ ] A fixture contributor's shared helper skips its project-I/O rule on failed/off/absent state, while its AST-only rule and an unrelated contributor still run without a host dependency registry.
- [ ] A `files`-scoped project-rule declaration is rejected; global ignores do not suppress it; global extends precedence and options are deterministic.
- [ ] CLI/watch/structured API surface one project finding.
- [ ] Opening two source documents leaves no source-attached project copies and one current problem at the logical selected-config URI.
- [ ] A clean LSP re-evaluation clears that publication, and a config selection change clears the old URI before publishing the new one.
- [ ] The real package-metadata discovery path works with no `compilerOptions.plugins` entry in any test tsconfig.

## Non-goals and forbidden shortcuts

- No AutoBe-specific path, filename, workspace-marker, or boundary policy.
- No TypeScript transformer `plugins` requirement.
- No `first source file` inference, `os.Getwd`, contributor environment read, `sync.Once`, or process-global project cache.
- No blanket symlink rejection and no replacement of physical Program identity with logical spelling.
- No generic host dependency graph, namespace-prefix suppression, or registration-order coupling.
- No promise of a kernel filesystem sandbox, hard-link identity, mount uniqueness, or external-import prohibition.

samchon:
## Summary

- preserve lexical invocation/config/root identity separately from physical config and Program roots across CLI, API, watch, and LSP hosts
- add project-scoped lint rules with one evaluation per loaded Program, finalized file-rule context results, deterministic detached findings, and dedicated global configuration resolution
- publish project diagnostics through a separate LSP channel with clear/replace, upstream merge, and stale-generation handling
- cover linked projects, empty source sets, configuration precedence, lifecycle isolation, package discovery, and all supported host surfaces in tests and documentation

## Verification

- completed five exhaustive manual review rounds over the full change; the final round found no remaining issues
- repository formatter completed before the coordinator's final prohibition; it was intentionally not rerun after the final manual edits
- build, typecheck, lint, static analysis, and executable tests were intentionally not run per task constraints

Closes #406


samchon:
## Exhaustive review record

Five full manual review rounds were completed over the entire change. Each round re-read the implementation, callers, tests, documentation, and latent-risk paths; the final round found no remaining issues.

- [x] lexical invocation/config/root identity remains distinct from physical config and Program roots
- [x] explicit project root and plugin-config origin propagate through CLI, API, watch, single-file, runtime, transform, and LSP hosts
- [x] project rules run once per loaded Program, including empty source sets, before file rules
- [x] file contexts expose finalized project results for absent, off, not-evaluated, passed, and failed states
- [x] detached project findings are deterministic, deduplicated, and ordered before file findings
- [x] global configuration rejects every files-selector form and preserves extends/severity/options precedence
- [x] top-level ignores filter project source sets without changing project identity
- [x] LSP project diagnostics use a separate logical-config publication with merge, replace, clear, and stale-generation behavior
- [x] project diagnostics remain actionless and versionless at zero range
- [x] linked projects, lifecycle isolation, package metadata discovery without compiler plugins, and zero-source projects are covered by tests
- [x] public API and protocol documentation were updated

Validation constraint: build, typecheck, lint, static analysis, and executable tests were intentionally not run. The repository formatter completed before the coordinator's final prohibition and was not rerun after the last manual edits, as directed.

samchon:
Focused local verification after the final portability fix: six project-rule Go regressions passed and five LSP project-publication regressions passed. The existing five whole-change review rounds exceed the required three; the final test-only path-normalization delta and integration invariants were rechecked with no remaining finding. Formatter and unrelated full-suite runs were intentionally omitted per coordinator direction. Proceeding to immediate merge.

PR #439, opened 2026-07-13
fix(lint): complete prefer-as-const suggestion parity

### LINKED ISSUE #417: fix(lint): make prefer-as-const cover variable and property annotations
## Summary

`@ttsc/lint`'s `typescript/prefer-as-const` rule only visits `as` and angle-bracket type assertions. It does not report equivalent literal type annotations on variable declarations or class properties, both of which are part of the non-configurable upstream rule.

The omission reproduces with published `@ttsc/lint@0.18.4` and current `master` (`fbbad7caeb44a16c77808c350ffaeeed608cabd1`) using a no-plugin fixture. The fixture also verifies an important boundary: the upstream implementation compares literal raw spelling, so differently quoted string literals are a valid negative control and must not be folded together semantically.

## Official contract

`@typescript-eslint/prefer-as-const` reports four AST families when the initializer literal and literal type annotation have the same raw representation:

- `TSAsExpression`;
- `TSTypeAssertion`;
- `VariableDeclarator` with a literal type annotation and matching initializer;
- `PropertyDefinition` with a literal type annotation and matching initializer.

Assertions are directly fixable to a const assertion. Declaration/property annotations use suggestions because moving the literal type to `as const` is a larger source rewrite.

The rule is not configurable. Reference and examples: https://typescript-eslint.io/rules/prefer-as-const/

The upstream source intentionally compares literal `raw` values. For example, `'value' as "value"` is not reported even though the runtime values are equal. This issue is about missing declaration/property visitors, not about broadening equality beyond the official contract.

The current implementation in `packages/lint/linthost/rules_ts.go:164-225` registers only `KindAsExpression` and `KindTypeAssertionExpression` and performs the same source-text equality check for those two shapes.

`lint.config.json`:

```json
{ "rules": { "typescript/prefer-as-const": "error" } }
```

`case.ts`:

```ts
let numeric: 2 = 2;
let quoted = "value" as "value";
let differentlyQuoted = 'value' as "value";

class Holder {
  public value: "value" = "value";
}

numeric = 2;
quoted = "value";
differentlyQuoted = "value";

export { differentlyQuoted, Holder, numeric, quoted };
```

Use a strict ES2022/NodeNext `tsconfig.json` with `noEmit: true` and no `plugins`. Run:

```bash
```

```text
case.ts:2:14 - error TS11570: [typescript/prefer-as-const] Expected `as const` instead of `as` literal type.
Found 1 error ...
exit 2
```

Expected: findings at lines 1, 2, and 6, with no finding at line 3. Line 2 is the positive control for the existing assertion path; line 3 proves the raw-spelling boundary.

Add variable-declaration and property-declaration visitors and route all four forms through one literal/raw comparison helper. For annotations:

- require both an initializer and a literal type annotation;
- compare the same literal raw representation used by the upstream rule;
- report the type annotation range;
- offer a suggestion that removes the annotation and appends the appropriate const assertion without silently changing unrelated modifiers, definite-assignment markers, or comments.

Support the literal node families the upstream parser exposes, including string, number, boolean, and bigint literals on ordinary variable declarators and property definitions. Preserve the existing distinction for template literals and differently spelled/raw-equivalent values. Do not invent extra binding-pattern behavior beyond executable upstream/source evidence.

### Forbidden shortcuts

- Do not use checker-level value equality or normalize quote/numeric spelling; the upstream rule intentionally uses raw literal equality.
- Do not report every literal annotation regardless of its initializer.
- Do not rewrite declarations by source regex or drop comments/modifiers.
- Do not treat class methods/accessors as property definitions.
- Do not add tsconfig transformer plugins.

## Acceptance criteria

- The reproduction reports exactly lines 1, 2, and 6, and leaves line 3 clean.
- Matching variable and property literal annotations report; missing initializers, mismatched literals, and broad primitive annotations do not.
- Existing `as` and angle-bracket assertion behavior remains covered.
- String, numeric, boolean, bigint, ordinary variable, and class-property cases have positive and negative tests.
- Differently quoted strings, differently spelled numeric literals, and template-literal distinctions match upstream raw-value behavior.
- Assertion fixes and declaration/property suggestions produce syntactically valid TypeScript and preserve comments/modifiers; no unsafe automatic rewrite is introduced where only a suggestion is appropriate.
- Native tests and a real no-plugin CLI fixture cover all four visitor families.

## Existing issue review

- #137 was a closed rule-coverage inventory and did not track missing declaration/property forms inside this implemented rule.
- No open or closed issue containing `prefer-as-const`, literal annotations, or property definitions covers this omission.
- #401 is unrelated.

samchon:
## Intent
Complete the reopened prefer-as-const parity contract after #426 by handling parenthesized literal types, preserving manual-only annotation rewrites, and proving real package auto-discovery.

## Scope
- unwrap type-side parentheses and keep assertion fixes syntactically valid while preserving comments
- expose declaration/property rewrites as range-scoped editor suggestions, never ttsc fix edits
- add exact native and no-plugin CLI regression coverage plus user-facing rule docs

## Verification
- focused prefer-as-const and LSP Go tests: passed
- scoped test-lint build: passed
- exact native-plugins/config no-plugin CLI fixture: passed
- git diff --check: passed

Closes #417

### DIFF
--- a/packages/lint/linthost/rules_ts.go
+++ b/packages/lint/linthost/rules_ts.go
@@ -4,7 +4,12 @@
 // Each rule is registered in the package init function at the bottom.
 package linthost
 
-import shimast "github.com/microsoft/typescript-go/shim/ast"
+import (
+  "strings"
+
+  shimast "github.com/microsoft/typescript-go/shim/ast"
+  shimscanner "github.com/microsoft/typescript-go/shim/scanner"
+)
 
 // noExplicitAny: ban `: any` annotations. Loud equivalent of
 // `@typescript-eslint/no-explicit-any`.
@@ -188,7 +193,7 @@ func (preferAsConst) Check(ctx *Context, node *shimast.Node) {
     }
   case shimast.KindVariableDeclaration:
     if decl := node.AsVariableDeclaration(); decl != nil {
-      preferAsConstCheckAnnotation(ctx, decl.Initializer, decl.Type)
+      preferAsConstCheckAnnotation(ctx, node, decl.Initializer, decl.Type)
     }
   case shimast.KindPropertyDeclaration:
     decl := node.AsPropertyDeclaration()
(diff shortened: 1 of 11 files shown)


samchon:
Verified before merge:

- three clean exhaustive reviews of the full rebased PR diff
- focused prefer-as-const and LSP Go suite passed
- scoped test-lint build passed
- exact package-auto-discovery CLI fixture without compilerOptions.plugins passed
- formatter was not run; GitHub CI was not awaited

Squash merge: ddd7bf6dda43464298e496497fdfd291fabf29cf

PR #441, opened 2026-07-13
fix(lint): lock the ban-ts-comment package-discovery contract

### LINKED ISSUE #415: fix(lint): align ban-ts-comment with recommended defaults and directives

samchon:
Closes the two contract gaps found in the post-merge audit of #429.

- Make the public `{ descriptionFormat }` object match the upstream required-property type.
- Add a real CLI regression that discovers `@ttsc/lint` from a direct dependency with no `compilerOptions.plugins` entry.
- Cover recommended defaults plus the `true`, `false`, `allow-with-description`, regex-object, and minimum-description-length option paths through that consumer-real route.

Local verification:
- `TTSC_BUILD_SCOPE=test-lint pnpm run build:current`
- `GOFLAGS=-run=BanTsComment node scripts/test-go-lint.cjs`
- `pnpm exec tsc --noEmit -p tests/test-lint/tsconfig.json`
- `pnpm --filter @ttsc/test-lint start --include=ban_ts_comment_covers_defaults_and_options_without_plugins_entry`

Closes #415

### DIFF
--- a/packages/lint/src/structures/rules/ITtscLintTypeScriptRuleOptions.ts
+++ b/packages/lint/src/structures/rules/ITtscLintTypeScriptRuleOptions.ts
@@ -24,7 +24,7 @@ export type TtscLintTypeScriptBanTsCommentDirectiveConfig =
        * against the raw text following the directive, including its leading
        * whitespace, so anchored patterns usually start with `^: `.
        */
-      descriptionFormat?: string;
+      descriptionFormat: string;
     };
 
 /**
--- a/tests/test-lint/src/features/plugin/test_lib_index_d_ts_rule_options_autocomplete_per_rule.ts
+++ b/tests/test-lint/src/features/plugin/test_lib_index_d_ts_rule_options_autocomplete_per_rule.ts
@@ -31,6 +31,8 @@ import assert from "node:assert/strict";
  * - Cross-rule option leakage (`testIdPattern` on
  *   `cypress/unsafe-to-chain-command`) is rejected.
  * - A typo in a switch-exhaustiveness option is rejected.
+ * - An empty object policy for `typescript/ban-ts-comment` is rejected because
+ *   the object form requires `descriptionFormat`.
  * - A lint-only rule (`no-var`) cannot carry an options object.
  * - An identifier-form built-in name without the canonical slash (`reactJsxKey`)
  *   is rejected.
@@ -175,6 +177,17 @@ export const test_lib_index_d_ts_rule_options_autocomplete_per_rule = () => {
       ],
     },
   };
+  const banTsCommentMissingDescriptionFormat: ITtscLintConfig = {
+    rules: {
+      "typescript/ban-ts-comment": [
+        "error",
+        {
+          // @ts-expect-error — the object policy requires descriptionFormat.
+          "ts-expect-error": {},
+        },
+      ],
+    },
+  };
   const camelBuiltinName: ITtscLintConfig = {
     rules: {
       // @ts-expect-error — built-in rules use kebab/slash names such as `react/jsx-key`; camelCase identifiers are not in the typed surface.
@@ -193,5 +206,6 @@ export const test_lib_index_d_ts_rule_options_autocomplete_per_rule = () => {
   assert.ok(crossRuleShape);
   assert.ok(lintRuleWithOptions);
   assert.ok(switchOptionTypo);
+  assert.ok(banTsCommentMissingDescriptionFormat);
   assert.ok(camelBuiltinName);
 };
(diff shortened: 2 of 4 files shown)


samchon:
Exhaustive review 1/3 complete at head 29c705d8. Re-read all four changed files and checked the public option-type wiring, TestLint overwrite order, direct-dependency discovery path, recommended defaults, every directive-policy arm, exact diagnostic lines, documentation sync, and latent regression risks. No findings.

samchon:
Exhaustive review 2/3 complete at head 29c705d8. Re-read the full four-file diff and compared the assertions with runtime policy normalization. The exact diagnostic list pins the allowed branches as well as the true/minimum-length/format failures, and the negative type assertion fails if the property becomes optional again. No findings.

samchon:
Exhaustive review 3/3 complete at head 29c705d8. Re-read the entire four-file PR diff from the beginning and checked current-master ancestry, clean scope, public export reachability, isolated dependency discovery, positive and negative test oracles, documentation parity, commit integrity, and merge risk. No findings.

PR #454, opened 2026-07-13
fix(lint): execute suggestions through logical project links

### LINKED ISSUE #449: fix(lint): execute LSP suggestions through logical project links
## Summary

Lint LSP suggestions are advertised for a document reached through a logical symlink/junction workspace, but executing the advertised `quickfix.ttsc` action rejects the same document as outside the project.

## Verified path mismatch

With project identity transport enabled, `NativePluginSource` launches the sidecar with the physical project root as `--cwd`. Editor document URIs intentionally preserve the logical workspace path.

- Code-action discovery calls `lspProjectTargetOutsideCwd`, which realpaths both cwd and target, so the suggestion is returned.
- `lspWorkspaceEditForSuggestion` later calls `projectRelativePath(opts.cwd, target)` on the raw logical target and physical cwd, so executing that suggestion exits 2 as outside cwd.
- The ordinary fix/format command path already realpaths the target before the same containment check.

This makes one action internally inconsistent: discovery accepts it, execution rejects it.

## Required implementation

Use the same physical containment and node_modules guards for suggestion execution that code-action discovery and ordinary commands use. Re-analysis and finding matching must operate on the physical project while the returned WorkspaceEdit remains keyed by the original logical URI expected by the editor.

Do not weaken containment: a symlink whose real target escapes the physical project must remain rejected. Do not replace the editor-facing URI with the physical path.

## Regression coverage

- physical project root plus a separately located logical symlink/junction workspace;
- obtain a suggestion through the logical file URI and execute the exact returned action successfully;
- assert the WorkspaceEdit key is the original logical URI;
- assert source hash/fingerprint and finding matching still protect stale actions;
- outside-physical-root and node_modules twins remain rejected;
- ordinary non-linked projects remain unchanged.

Related: #406

samchon:
## Summary

- resolve suggestion execution targets through the same physical project boundary used by code-action discovery
- filter findings and revalidate source hashes against the physical file while preserving the editor's logical URI in `WorkspaceEdit`
- cover symlink/junction execution plus forged outside-root and `node_modules` negative twins

## Validation

- three exhaustive whole-PR review passes and targeted local tests will run against this exact head before merge

Closes #449

### DIFF
--- a/packages/lint/linthost/host.go
+++ b/packages/lint/linthost/host.go
@@ -198,11 +198,52 @@ func absoluteProjectPath(cwd string, target string) string {
 }
 
 func realProjectPath(target string) string {
-  resolved, err := filepath.EvalSymlinks(target)
-  if err != nil {
-    return filepath.Clean(target)
+  original := filepath.Clean(target)
+  resolved := original
+  seen := make(map[string]struct{})
+  for range 255 {
+    key := filepath.Clean(resolved)
+    if _, exists := seen[key]; exists {
+      return original
+    }
+    seen[key] = struct{}{}
+
+    if evaluated, err := filepath.EvalSymlinks(resolved); err == nil {
+      resolved = filepath.Clean(evaluated)
+    }
+    next, ok := resolveProjectLinkAncestor(resolved)
+    if !ok {
+      return filepath.Clean(resolved)
+    }
+    resolved = next
   }
-  return resolved
+  return original
+}
+
+// resolveProjectLinkAncestor resolves the nearest symlink-like ancestor and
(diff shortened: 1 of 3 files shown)


samchon:
Final validation on head 3a9fa35c65edf9e020ebe2d0fe893f0a0d76e413: three exhaustive whole-PR reviews completed with no remaining findings. Targeted local LSP validation passed for logical-link suggestion execution, stale selections, native prefer-as-const suggestions, and node_modules rejection; the Windows chmod-only boundary test skipped by design. The regression also verifies linked outside-project and hidden node_modules targets fail closed.

PR #459, opened 2026-07-13
fix(lint): initialize contributor registries once

### LINKED ISSUE #456: fix(lint): initialize contributor registries once for reusable hosts
## Summary

`linthost.Main` re-runs contributor registration on every non-version invocation. The public contributor registry is defined as init-time input that the host reads once during engine bootstrap, while `Main` is also the reusable library entry used by hosts such as the playground.

A second valid `Main` call therefore walks the same contributors again, sees their already-installed adapters, and emits collision warnings for normal reuse. The test suite currently masks one exact warning as benign. Concurrent calls also perform unsynchronized lookup/write access against the global file and project rule maps.

## Evidence

- `rule.Register` is documented as an init-time operation.
- `rule.Registered` is documented as being called once by the host during engine bootstrap.
- `dispatch.run` calls `registerContributors()` for every recognized non-version command.
- `isBenignContributorCollisionWarning` explicitly permits the warning produced by repeated registration of `demo/option-consumer`.

## Expected behavior

Contributor metadata should be inspected and installed exactly once after package initialization. Reusing `linthost.Main` sequentially must not produce contributor-collision stderr, and concurrent entry calls must not race while mutating registries.

Use a conventional one-time bootstrap or an equivalent immutable initialization boundary. Do not add warning suppression, command-specific exceptions, or test-only state. Preserve deterministic collision reporting during the single real bootstrap and the documented panic/recovery boundaries for malformed contributors.

## Regression coverage

- Call two different valid `Main` commands sequentially and assert the second call has no bootstrap warning.
- Exercise concurrent entry calls under the Go race detector or a deterministic synchronization harness and prove registry initialization is single-owner.
- Preserve file/file, project/project, and file/project collision behavior during initial bootstrap.
- Remove the exact benign-warning allowance where the corrected lifecycle makes it unnecessary.

samchon:
## Intent

Make the reusable lint library entry initialize immutable contributor declarations exactly once, including under concurrent first use.

## Scope

- gate file and project contributor installation with one synchronized bootstrap
- preserve first-bootstrap collision and metadata-panic diagnostics
- remove the test suite's repeated-collision warning allowance
- add fresh-process coverage for concurrent initialization and sequential Main reuse
- align the maintainer walkthrough with the reusable one-time lifecycle

## Deferred

- The Go race detector is unavailable on this Windows host because CGO is disabled.
- The full lint suite is not green locally: three AsyncDisposable failures reproduce on latest master, and four TypeScript-config tests require the built ttsx launcher missing from the dedicated worktree. Repository CI remains GitHub-owned.

## Test plan

- targeted and expanded contributor lifecycle Go tests
- full @ttsc/lint Go suite with latest-master baseline comparison
- three independent exhaustive reviews of the complete PR diff, callers, tests, CI, and docs

Closes #456

### DIFF
--- a/packages/lint/linthost/contrib_adapter.go
+++ b/packages/lint/linthost/contrib_adapter.go
@@ -10,20 +10,31 @@
 // `package main` init functions and the relative order across files in
 // the same package is alphabetical — so we cannot blindly run the
 // contributor wiring from a file-level `init()` and expect built-in
-// registration to have already completed. Instead, `registerContributors`
-// is invoked explicitly from `main.run` after all built-in `init` calls
-// have settled, which makes the collision check meaningful.
+// registration to have already completed. Instead, `registerContributorsOnce`
+// is invoked explicitly from `run` after all built-in `init` calls have
+// settled, which makes the collision check meaningful.
 package linthost
 
 import (
   "fmt"
   "os"
   "sort"
+  "sync"
 
   shimast "github.com/microsoft/typescript-go/shim/ast"
 
   "github.com/samchon/ttsc/packages/lint/rule"
 )
+
+var contributorBootstrap sync.Once
+
+// registerContributorsOnce installs the immutable init-time contributor
+// registries exactly once. Main is a reusable library entry and may be called
+// concurrently, so every caller must wait for the same completed bootstrap
+// before constructing an Engine that reads the global rule maps.
+func registerContributorsOnce() {
+  contributorBootstrap.Do(registerContributors)
+}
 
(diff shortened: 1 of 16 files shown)


samchon:
The Windows Go failure on this merged snapshot is stale relative to master. Its disposable-library fixtures were fixed by #465, and its Windows short-path/config/LSP identity failures were fixed by #464 (with the symlink boundary established in #454).

I re-ran the failed-test families against master at 0b60eeb9f074d19be4722e46f1d2e6091524a7e4; they all pass. No additional code change is required for this historical check result.

PR #464, opened 2026-07-13
Fix Windows short-path identity in lint and LSP

### LINKED ISSUE #463: Canonicalize Windows short-path project identities for lint and LSP
## Problem

On Windows, the lint host treats an existing 8.3 short path and its long-path spelling as different project roots. `filepath.EvalSymlinks` expands the existing side of the comparison, but it cannot expand a not-yet-created descendant, while the fix-all and in-memory-format LSP paths compare a physical target against the unnormalized invocation cwd.

This makes valid files appear outside `--cwd` and prevents global ignore patterns from matching synthetic source paths. The `windows-go` job for PR #444 reproduced both failure classes:

- `C:\Users\runneradmin\...\src\main.ts is outside cwd C:\Users\RUNNER~1\...`
- top-level ignores do not match non-materialized `generated/**` source paths below the same short-path root

The failures are filesystem-identity problems, not case-only or string-prefix problems.

## Required behavior

- Resolve the nearest existing ancestor before reattaching a missing descendant so short/long aliases, junctions, and symlinks share one physical identity.
- Compare LSP cwd and target in the same physical namespace for fix-all and format-buffer paths.
- Preserve logical URIs in returned WorkspaceEdits.
- Keep containment strict for genuinely external paths.
- Cover existing targets, missing descendants, ignore matching, and outside-root negative cases on Windows without hardcoded runner paths.

## Evidence

Failing job: https://github.com/samchon/ttsc/actions/runs/29277556055/job/86910331754

samchon:
## Summary

- canonicalize missing descendants through their nearest existing ancestor so Windows 8.3 aliases, junctions, and long paths share one project identity
- compare fix-all and format-buffer LSP targets against the physical cwd while preserving the editor's logical URI in WorkspaceEdits
- remap absolute config and tsconfig paths through the same physical identity
- add Windows regression coverage for missing synthetic sources, global ignores, outside-root rejection, fix-all, and in-memory formatting

Closes #463.

## Validation

- three sequential exhaustive whole-PR reviews at e6b4bda4cb3fb9d9ceb229e4fec51ee183b67e47: clean
- focused materialized lint-host tests: both top-level tests and all subtests passed
  - TestRealProjectPathExpandsWindowsShortAncestorForMissingDescendant
  - TestLSPCommandsAcceptWindowsShortCwd/fix_all
  - TestLSPCommandsAcceptWindowsShortCwd/format_buffer
- diff check and remote-head integrity: clean

## Constraints

- no runner-specific path hardcoding
- no monkeypatching or test-only production branches
- no formatting pass


samchon:
Final gate: exact HEAD e6b4bda4cb3fb9d9ceb229e4fec51ee183b67e47 passed three sequential exhaustive whole-PR reviews. The focused materialized Windows suite passed both top-level tests and all subtests, covering missing descendants, global-ignore containment, LSP fix-all with an absolute config, logical URI preservation, and in-memory formatting. No formatter was run.

PR #465, opened 2026-07-13
test(lint): make await-using fixtures compiler-valid

### LINKED ISSUE #458: test(lint): make await-using fixtures type-valid before lint
## Summary

Three `await-thenable` clean-case tests create `await using` projects with the shared ES2022 fixture but never load the explicit resource-management library. TypeScript rejects those projects before lint runs, so the tests do not exercise the behavior their assertions claim to protect.

## Reproduction

```text
go test -count=1 -run '^TestAwaitThenableAwaitUsingAsyncDisposableAllows$' -v ./linthost
```

Current result:

```text
TS2318: Cannot find global type 'AsyncDisposable'.
TS2318: Cannot find global type 'Disposable'.
```

The same prerequisite failure affects:

- `TestAwaitThenableAwaitUsingAsyncDisposableAllows`
- `TestAwaitThenableAwaitUsingProtocolAbstractionsAllows`
- `TestAwaitThenableAwaitUsingUnionWithAsyncDisposableAllows`

Declaring only `SymbolConstructor.dispose` and `SymbolConstructor.asyncDispose` does not satisfy TypeScript's required global protocol types.

## Expected implementation

Make each resource-management fixture type-valid before invoking lint, using the canonical compiler configuration already used by sibling tests (`ESNext.Disposable`) or a shared fixture helper. Do not suppress compiler diagnostics, loosen the command assertions, or inject test-only behavior into the rule.

## Regression coverage

- Prove the three projects type-check without lint enabled.
- Preserve the clean `await-thenable` assertions for direct, abstracted, generic, intersection, and union async-disposable shapes.
- Keep their sync-only negative twins reporting the lint rule rather than failing at the compiler prerequisite layer.

samchon:
## Intent

Keep await-thenable resource-management regressions behind valid TypeScript compiler prerequisites so a clean lint assertion cannot pass or fail for the wrong reason.

## Scope

- Configure the affected await using fixtures with the canonical ESNext.Disposable library.
- Remove partial local SymbolConstructor declarations from those fixtures.
- Prove every positive and negative fixture type-checks through the real command front door with no lint plugin entry.
- Preserve lint assertions for direct, abstracted, generic, intersection, union, wrong-symbol, and multi-declarator shapes.

## Root cause

The shared ES2022 project fixture did not load the explicit-resource-management declarations. Defining only Symbol.dispose and Symbol.asyncDispose left the required global Disposable and AsyncDisposable types missing, so compiler TS2318 diagnostics preempted the intended lint assertions.

## Deferred

No production rule change is needed.

## Test plan

After three exhaustive whole-PR review rounds, run the focused TestAwaitThenableAwaitUsing Go tests from the materialized lint test module.

Closes #458

### DIFF
--- a/packages/lint/test/rules/typescript/await_thenable_await_using_async_disposable_allows_test.go
+++ b/packages/lint/test/rules/typescript/await_thenable_await_using_async_disposable_allows_test.go
@@ -19,16 +19,11 @@ import (
 // surfaces here.
 //
 //  1. Seed a project containing all four valid resource-management shapes.
-//  2. Run `check` with typescript/await-thenable enabled as error.
-//  3. Assert a clean exit with no await-thenable finding.
+//  2. Prove the fixture type-checks without a lint plugin entry.
+//  3. Run `check` with typescript/await-thenable enabled as error.
+//  4. Assert a clean exit with no await-thenable finding.
 func TestAwaitThenableAwaitUsingAsyncDisposableAllows(t *testing.T) {
-  root := seedLintProject(t, `export {};
-declare global {
-  interface SymbolConstructor {
-    readonly dispose: unique symbol;
-    readonly asyncDispose: unique symbol;
-  }
-}
+  root := seedAwaitUsingLintProject(t, `export {};
 interface AsyncResource {
   [Symbol.asyncDispose](): Promise<void>;
 }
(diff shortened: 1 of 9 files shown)


samchon:
Validation complete at `0b8d876d9825a14be2ed0c6418e9b334eecb9e19`.

- Exhaustive whole-PR self-review: 3 clean rounds against the same final HEAD. Every changed file and hunk was reread from scratch each round, with caller, compiler-library, diagnostic-separation, isolation, cross-platform, and latent-risk checks.
- Local focused test: `go test -count=1 -run '^TestAwaitThenableAwaitUsing' -v ./linthost` in the materialized lint module.
- Result: all 8 direct, abstraction, union, wrong-symbol, multi-declarator, and sync-only positive/negative cases passed.
- Formatting was not run, per the explicit workflow constraint.

CI remains delegated and was not awaited.

PR #466, opened 2026-07-13
fix(lint): track assertion-wrapped assignment targets

### LINKED ISSUE #448: fix(lint): recognize assertion-wrapped assignment targets in dataflow
## Summary

The shared assignment-target collector does not unwrap TypeScript assertion wrappers. `prefer-const` therefore misses real writes and can automatically rewrite a reassigned `let` binding to invalid `const` source.

## Verified reproductions

All of these are valid TypeScript assignment/update targets:

```ts
let a = 0;
(a as number) = 1;

let b = 0;
(<number>b) += 1;

let c = 0;
(c satisfies number)++;

let d = 0;
for ((d as number) of [1, 2]) {}
```

`assignmentTargetIdentifiers` unwraps parentheses and non-null assertions, but not `AsExpression`, `TypeAssertionExpression`, or `SatisfiesExpression`. `prefer-const` records no write and emits a `let` to `const` automatic fix; the rewritten program then fails on reassignment. Nested wrappers and destructuring targets have the same gap.

The text projection `assignmentTargetNames` shares this collector, so dataflow rules that invalidate aliases on assignment can also retain stale trust after an assertion-wrapped write.

## Required implementation

Teach the shared assignment-target traversal to recursively unwrap every TypeScript outer assertion accepted by the compiler in reference position: parentheses, non-null, `as`, angle-bracket type assertion, and `satisfies`. Apply the same semantics inside destructuring patterns, updates, and for-in/of targets. Member targets such as `(obj.x as number) = 1` must remain member writes and must not be misclassified as writes to the local `obj` binding.

Use AST kinds and compiler reference semantics; do not scan source text or special-case identifier names.

## Regression coverage

- simple and compound assignment, prefix/postfix update, for-in/of;
- all wrapper kinds individually and nested together;
- array/object destructuring and defaults;
- parenthesized/non-null existing controls;
- member-target negative controls;
- `prefer-const` emits no unsafe fix and still fixes an actually stable binding;
- every shared helper consumer invalidates assertion-wrapped writes correctly.

samchon:
## Intent

Preserve binding-write identity through every TypeScript assertion wrapper accepted in assignment reference position.

## Scope

- unwrap as, angle-bracket, satisfies, non-null, and parenthesized targets recursively
- retain the same semantics inside destructuring, defaults, updates, and for-in/of targets
- keep member writes distinct from writes to their holder binding
- invalidate security-rule static bindings after wrapped updates and loop assignments
- add a real prefer-const fixer shield plus shared-security and command-path regressions

Closes #448.

## Validation

- three sequential exhaustive whole-PR reviews at 1b9c8a9217c41f4b5fb5a15de97dff753316d53b: clean
- focused materialized lint-host tests: 8/8 passed
- included new wrapper tests plus non-null, parenthesized update, destructuring, stable-fix, and reassigned-static controls
- diff check and local/remote HEAD integrity: clean

## Constraints

- no source-text target scanning
- no identifier-specific exceptions
- no monkeypatching or test-only production path
- no formatting pass

### DIFF
--- a/packages/lint/linthost/ast_helpers.go
+++ b/packages/lint/linthost/ast_helpers.go
@@ -344,9 +344,12 @@ func (w *descendantsWalker) visitChild(child *shimast.Node) bool {
 // array elements, object property values, shorthand properties, defaults
 // (`[a = 1]`, `{a = 1}`), nested patterns, and rest elements.
 //
-// Property names in `{key: target}` are read positions, not writes, so
-// only the property value contributes; member-access targets (`obj.x`)
-// declare no local binding and are skipped. Returns nil for other shapes.
+// TypeScript assertion wrappers are transparent in reference position, so
+// `as`, angle-bracket assertions, `satisfies`, non-null assertions, and
+// parentheses are unwrapped recursively wherever they appear in a target.
+// Property names in `{key: target}` are read positions, not writes, so only
+// the property value contributes; member-access targets (`obj.x`) declare no
+// local binding and are skipped. Returns nil for other shapes.
 func assignmentTargetIdentifiers(node *shimast.Node) []*shimast.Node {
   if node == nil {
     return nil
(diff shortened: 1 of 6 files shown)


samchon:
Final gate: exact HEAD 1b9c8a9217c41f4b5fb5a15de97dff753316d53b completed three sequential exhaustive whole-PR reviews with no findings. The focused materialized lint-host bundle passed 8/8 tests, including all new wrapper contexts and existing non-null, parenthesized, destructuring, stable-fix, and security controls. No formatter was run.

PR #467, opened 2026-07-13
fix(lint): align directive lengths with Unicode graphemes

### LINKED ISSUE #451: fix(lint): use standards-aligned grapheme boundaries for directive lengths
## Summary

`typescript/ban-ts-comment` measures `minimumDescriptionLength` with `stringLength`, whose contract is parity with typescript-eslint's `Intl.Segmenter`-based `getStringLength`. The local hand-written segmenter intentionally omits normative Unicode grapheme rules and therefore accepts descriptions that upstream rejects.

## Verified failures

- UAX #29 GB9b requires `Prepend ? Any` after the control rules. `U+0600 ARABIC NUMBER SIGN + a` is one grapheme in `Intl.Segmenter`; the current implementation counts two.
- UAX #29 GB9c keeps Indic conjuncts together. Devanagari `KA + VIRAMA + SSA` (`???`) is one grapheme in `Intl.Segmenter`; the current implementation counts two.

At `minimumDescriptionLength: 2`, a description containing `U+0600 + a` is incorrectly accepted. The reverse `a + U+0600` is correctly two clusters, and controls must still break according to GB4/GB5, so indiscriminate joining is not valid.

## Required correction

Replace the selective approximation with a standards-derived, versioned extended-grapheme implementation compatible with `Intl.Segmenter`. Do not add code-point-specific exceptions. Unicode property tables may be generated from authoritative data; keep the runtime deterministic and dependency policy intact.

## Regression shield

Cover at least:

- Prepend before base, repeated Prepend, reverse order, Prepend before controls, and Prepend plus Extend;
- Indic consonant/linker/extend/consonant sequences and breaking negative twins;
- existing CR/LF, Hangul, combining mark, spacing mark, emoji ZWJ, modifier, and regional-indicator cases;
- default and custom `minimumDescriptionLength` thresholds through the real ban-ts-comment command path.

Where practical, validate the segmenter against the official Unicode GraphemeBreakTest corpus so future Unicode updates cannot silently reopen partial-rule gaps.

samchon:
## Intent

Make `typescript/ban-ts-comment` measure directive descriptions with the complete Unicode extended-grapheme contract used by `Intl.Segmenter`, instead of a selective approximation.

## Scope

- replace the hand-written subset with the ordered UAX #29 GB3-GB999 algorithm
- pin generated Grapheme_Cluster_Break, Indic_Conjunct_Break, and Extended_Pictographic data to Unicode 16.0.0 (Node 24 CI parity)
- vendor the matching official `GraphemeBreakTest.txt` corpus
- cover Prepend, Indic conjunct, emoji tag, control, reverse-order, incomplete, and established grapheme families through the real rule path
- document the versioned extended-grapheme contract

## Deferred

- Unicode-version upgrades remain coupled to a future Node CI runtime upgrade and must regenerate both tables and the conformance corpus.

## Test plan

- three sequential exhaustive whole-PR reviews at one exact HEAD
- focused Go conformance and ban-ts-comment tests after the reviews
- no repository formatter run

Closes #451


samchon:
Exhaustive review found and fixed two maintainability/test-intent gaps: the emoji-data marker is now derived from the pinned Unicode version, and the incomplete emoji-tag case is documented as an Extend-property shield rather than a breaking twin. Review count reset to zero at the new HEAD.

samchon:
A subsequent whole-PR review found that the GB6-GB8 helper name incorrectly implied Hangul-only data even though Unicode 16 assigns conjoining classes beyond Hangul. Renamed it and the rune-property carrier for semantic accuracy. Review count reset to zero again at the new HEAD.

PR #468, opened 2026-07-13
fix(lint): honor safe calls inside composite expressions

### LINKED ISSUE #461: fix(lint): honor safe-call specifiers inside composite expressions
## Summary

`typescript/no-floating-promises` applies `allowForKnownSafeCalls` only to the root expression statement. The recursive analyzer does not re-check the option, so placing the exact same configured safe call inside a comma, logical, or conditional expression revokes the allowlist and produces false positives.

## Reproduction

```ts
declare const condition: boolean;
declare function safe(): Promise<void>;

safe();                         // clean
(safe(), 0);                    // currently reports
condition && safe();            // currently reports
condition ? safe() : safe();    // currently reports
```

Configuration:

```json
{
  typescript/no-floating-promises: [
    error,
    { allowForKnownSafeCalls: [safe] }
  ]
}
```

A real sidecar run reports exactly the three nested expression statements. The option contract says these functions' returned Promises may be discarded safely; syntactic nesting does not change the callee or its safety policy.

## Expected implementation

Apply the known-safe call decision at every call expression reached by recursive floating-Promise analysis. Reuse the existing symbol/source-aware specifier matcher. Do not special-case comma/logical/conditional syntax, skip the entire parent expression, or weaken analysis of unlisted sibling branches.

## Regression coverage

For direct, comma, `&&`, `||`, `??`, and both conditional branches, pair:

- a configured safe call that remains clean;
- an unlisted Promise-returning call that still reports;
- mixed expressions containing both, which must report only because of the unsafe branch;
- string and structured file/package/lib specifiers;
- `checkThenables`, `ignoreVoid`, and `ignoreIIFE` interactions where relevant.

This is also an upstream-parity latent defect; the local fix should follow the documented option semantics rather than copy the gap.

samchon:
## Intent

Keep allowForKnownSafeCalls attached to the configured call even when recursive no-floating-promises analysis reaches it through a composite expression.

## Scope

- evaluate the existing symbol/source-aware safe-call matcher at every analyzed call node
- preserve direct, comma, logical, nullish, conditional, and explicit-void behavior
- keep unlisted sibling branches independently reportable
- cover universal, file, package, and TypeScript-lib specifiers
- preserve checkThenables, ignoreVoid, and ignoreIIFE option boundaries

Closes #461.

## Validation

- three sequential exhaustive whole-PR reviews at 7ae1d7cadc7b6996d5f0933236e51633f213f502: clean
- focused materialized lint-host tests: 6/6 passed
- new composite/specifier shields plus existing default, handler, options, and structured-specifier controls passed
- diff check and local/remote HEAD integrity: clean

## Constraints

- no syntax-specific parent skip
- no textual callee matching
- no monkeypatching or test-only production path
- no formatting pass

### DIFF
--- a/packages/lint/linthost/rules_promise.go
+++ b/packages/lint/linthost/rules_promise.go
@@ -709,9 +709,6 @@ func (noFloatingPromises) Check(ctx *Context, node *shimast.Node) {
   if options.IgnoreIIFE && isImmediatelyInvokedFunctionExpression(expr) {
     return
   }
-  if isKnownSafePromiseCall(ctx, expr, options.AllowForKnownSafeCalls) {
-    return
-  }
   result := analyzeFloatingPromise(ctx, expr, options)
   if !result.unhandled {
     return
@@ -745,6 +742,14 @@ func analyzeFloatingPromise(
 ) floatingPromiseResult {
   node = unwrapFloatingPromiseExpression(node)
   if node == nil {
+    return floatingPromiseResult{}
+  }
+
+  // A configured safe call stays safe wherever recursive expression analysis
+  // reaches it. The callee's symbol/source identity is unchanged by a comma,
+  // logical, conditional, or void parent, so the allowlist decision belongs
+  // at the call node rather than only at the root expression statement.
+  if isKnownSafePromiseCall(ctx, node, options.AllowForKnownSafeCalls) {
     return floatingPromiseResult{}
   }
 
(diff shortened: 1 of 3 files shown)


samchon:
Final gate: exact HEAD 7ae1d7cadc7b6996d5f0933236e51633f213f502 completed three sequential exhaustive whole-PR reviews with no findings. The focused materialized lint-host bundle passed 6/6 tests, including all new composite/specifier coverage and existing default, handler, option, and source-boundary controls. No formatter was run.

PR #469, opened 2026-07-13
fix(lint): enumerate comments with parser lexical context

### LINKED ISSUE #453: fix(lint): enumerate comments with TypeScript lexical context
## Summary

`forEachCommentToken` re-lexes source with a context-free raw scanner and manually tracks template braces. TypeScript's parser changes lexical goals for regular-expression literals, template continuations, and JSX text; the helper does not. Shared comment consumers therefore both miss real comments and treat source text as comments.

## Verified failures

- False negative: inside a template substitution, a regular expression containing `{` is scanned as an open brace because `ReScanSlashToken` is never called. The template depth remains unbalanced, and a real `// @ts-ignore` after the template is never visited.
- False positive: in TSX, `const x = <div>// @ts-ignore</div>;` and `<div>/* @ts-ignore */</div>` contain JSX text, not comments, but the raw standard scanner reports comment trivia.
- The same false JSX comment can be parsed as `eslint-disable`/`lint-disable` and suppress unrelated findings.

The helper is shared by inline-disable parsing, `typescript/ban-ts-comment`, and switch-exhaustiveness default-comment recognition, so the lexical mismatch is not rule-local.

## Required correction

Enumerate comments using the parsed TypeScript lexical context. Use parser/checker-owned comment ranges or parser-classified token spans with complete regex/template/JSX handling and range deduplication. Do not guess regex-versus-division, patch brace characters, or add JSX-string exceptions.

Preserve comments in leading/trailing/orphan trivia, empty blocks, template substitutions, JSX expression containers, file start/end, CRLF and Unicode line terminators. Comment-like bytes inside strings, templates, regex literals, and JSX text must remain invisible.

## Regression shield

Add real command-path twins covering:

- regex literals with `{`, `}`, escaped slashes, and character classes inside nested template substitutions, followed by a real directive comment;
- division expressions at the same syntactic positions;
- JSX slash/block comment-shaped text versus real `{/* comment */}` expression comments;
- JSX text containing lint-disable markers that must not suppress findings;
- inline disable/enable ranges, ban-ts-comment, and switch default-comment consumers;
- leading/trailing comments, empty containers, nested templates, TS/TSX/JS/JSX source modes, and malformed-source graceful behavior.

Assert exact comment ranges/diagnostics and clean negative twins.

samchon:
## Intent

Make every lint comment consumer respect TypeScript's parser-selected lexical goals so regex literals, templates, and JSX text cannot hide or fabricate directives.

Closes #453.

## Scope

- derive opaque lexical spans from parser-owned AST tokens
- enumerate and deduplicate exact comment trivia ranges only between those spans
- route inline disables, `typescript/ban-ts-comment`, and switch default-comment recognition through parser-classified ranges
- add TS/TSX/JS/JSX, nested-template regex, JSX-text, malformed-source, line-terminator, exact-range, and real command-path regression coverage

## Deferred

None.

## Test plan

- baseline reproduction: nested-template regex case loses the real trailing directive before this change
- baseline reproduction: JSX text falsely suppresses a real `no-debugger` finding before this change
- after three clean exhaustive review rounds, run the focused comment, directive, and switch command-path tests locally

Formatting was intentionally not run per maintainer instruction.

### DIFF
--- a/packages/lint/linthost/comment_scan.go
+++ b/packages/lint/linthost/comment_scan.go
@@ -1,68 +1,147 @@
-// Shared raw comment-token scan used by every rule or pass that needs the
-// exact byte range of each comment in a file (inline-disable directives,
-// typescript/ban-ts-comment). Centralizing the loop keeps the template
-// rescanning subtlety below in one place.
+// Shared parser-aware comment enumeration for every rule or pass that needs
+// exact comment byte ranges (inline directives, ban-ts-comment, and switch
+// default markers).
 package linthost
 
 import (
+  "sort"
+
   shimast "github.com/microsoft/typescript-go/shim/ast"
   shimscanner "github.com/microsoft/typescript-go/shim/scanner"
 )
 
-// forEachCommentToken re-lexes `text` with the raw TypeScript scanner and
-// invokes `visit` for every comment token with its kind and byte range.
+type commentToken struct {
+  kind shimast.Kind
+  pos  int
+  end  int
+}
+
+type sourceSpan struct {
+  pos int
+  end int
+}
+
+// forEachCommentToken visits every real comment in `file` in source order.
 //
-// The raw scanner does not split `KindTemplateExpression` on its own:
-// after returning `KindTemplateHead`/`KindTemplateMiddle`, it resumes
-// lexing the substitution as ordinary code, and a later `}` is reported
-// as `KindCloseBraceToken` instead of re-entering the template body.
-// Without intervention the next backtick would open a fresh template
-// scan that swallows the rest of the file (including every comment) as
-// one runaway unterminated literal. The loop avoids this by calling
-// `ReScanTemplateToken` on the matching `}`, tracked with a brace-depth
-// stack so comment positions stay aligned with the source bytes past any
-// template substitution.
-func forEachCommentToken(text string, visit func(kind shimast.Kind, pos, end int)) {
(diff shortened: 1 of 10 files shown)


samchon:
Validation on exact HEAD `095cada42729721dae79c29b1499e8c3e88d0dc8`:

- three sequential exhaustive whole-PR review rounds completed with no findings
- focused local scratch-module tests passed:
  - `TestCommandCheckUsesParserContextForCommentConsumers`
  - `TestEngineIgnoresJsxTextInlineDisableMarkers`
  - `TestBanTsCommentFollowsRegexInNestedTemplate`
  - `TestSwitchExhaustivenessCheckParserCommentRanges`
  - `TestCommentScanHandlesMalformedParserRecovery`
  - `TestCommentScanUsesParserLexicalRangesAcrossSourceModes`
- command: `go test -count=1 ./linthost -run '^(TestCommandCheckUsesParserContextForCommentConsumers|TestEngineIgnoresJsxTextInlineDisableMarkers|TestBanTsCommentFollowsRegexInNestedTemplate|TestSwitchExhaustivenessCheckParserCommentRanges|TestCommentScanHandlesMalformedParserRecovery|TestCommentScanUsesParserLexicalRangesAcrossSourceModes)$'`
- result: `ok github.com/samchon/ttsc/packages/lint/linthost 2.712s`
- `git diff --check` passed and the worktree is clean
- formatting was not run per maintainer instruction

PR #470, opened 2026-07-13
fix(lint): preserve no-fallthrough literal and exception reachability

### LINKED ISSUE #446: fix(lint): preserve no-fallthrough literal and exception reachability
## Summary

The structured completion analysis in `no-fallthrough` has two independently reproduced parity defects:

1. constant loop truthiness does not recognize hexadecimal BigInt literals or transparent parentheses;
2. `try`/`catch` treats every catch block as reachable even when the try has no throwing path.

Both produce false fallthrough diagnostics.

### BigInt loop constants

```ts
switch (value) {
  case 0:
    while (0x1n) {}
  case 1:
    break;
}
```

`0x1n` is truthy, so the loop cannot complete normally. TypeScript-Go preserves hexadecimal BigInt text with its `0x` prefix, while `literalTruthiness` accepts decimal digits only and returns unknown. Nested parentheses such as `while (((1n))) {}` have the same false positive. The `0x0n` negative twin is falsy and must still report fallthrough.

### Unreachable catch

```ts
function run(value: number): void {
  switch (value) {
    case 0:
      try {
        return;
      } catch {}
    case 1:
      break;
  }
}
```

A bare return cannot transfer control to the catch, but `tryCompletion` currently computes `try.normal || catch.normal` whenever a catch exists. ESLint reports no fallthrough here. Conversely, `throw 0`, `return call()`, and other potentially throwing evaluations can reach an empty catch and must continue to report.

## Required implementation

- Parse BigInt zero/nonzero semantics for every accepted radix and numeric separator spelling; strip only transparent outer parentheses, not unary expressions.
- Model reachable throw completion explicitly through statement-list, branch, loop, nested try/catch/finally, return, break, and continue composition.
- A catch contributes normal/abrupt completions only when the protected try has a reachable throw edge.
- Preserve finally override semantics and nested function/class code-path boundaries.

Do not special-case the reproductions, suppress all catches, assume every expression throws, or move the rule onto a type-checker-only lane.

## Regression coverage

- decimal/hex/binary/octal BigInt, zero/nonzero, separators, uppercase prefixes where accepted, nested parentheses, unary negative boundary;
- bare return and literal return stay non-fallthrough;
- explicit throw, identifier/member/call/new return expressions reach catch where ESLint does;
- break/continue, nested try/catch/finally, abrupt finally, async/generator and nested function boundaries;
- existing comment-marker and unused-marker behavior remains unchanged.

samchon:
## Intent

`no-fallthrough` treated every catch as reachable and left some accepted BigInt loop constants unknown. That reopened cases terminated by bare or literal returns and misclassified truthy radix-prefixed loops.

Closes #446

## Scope

- track normal, return, throw, break, and continue completions through statement lists, branches, loops, nested switches, try/catch/finally, and labels;
- admit a catch only when the protected region has a reachable ESLint throwable node, while preserving nested function, class-field, and static-block code-path boundaries;
- fold transparent parentheses and zero/nonzero BigInt literals for decimal, binary, octal, and hexadecimal spellings with separators and arbitrary width;
- document the observable reachability boundary;
- add real `check` command regressions for positive/negative twins plus the unused-marker option.

## Deferred

None.

## Test plan

Per the mandated lifecycle, focused local tests run only after three clean exhaustive whole-PR reviews at one identical head. GitHub CI remains independent verification.

Formatter was intentionally not run.


samchon:
Self-review reset: the first full pass found that the AST shim exposes type-node range sentinels rather than IsTypeNode, and that a blanket type-node skip would also hide runtime class extends evaluation. Head 5b3cd1a7 now uses the exported range, restores only class extends value expressions, and keeps implements/interface heritage type-only. The code-path boundary command test now pins both sides. Review count restarted at zero; formatter and local tests remain intentionally deferred until three clean full passes.

samchon:
Exhaustive self-review found an omitted ESLint code-path effect: a reachable yield can resume abruptly through both generator.throw() and generator.return(). Commit b6a45436 generalizes evaluated-expression completion to preserve both effects through statements, loops, switches, and try/finally, adds a nested generator/finally regression, and corrects a pre-existing try/catch test whose catch had become unreachable under the issue's intended semantics. The three clean-review count is reset to zero at this HEAD; focused local tests remain deferred until three consecutive clean rounds.

samchon:
A restarted exhaustive review found that TypeScript-Go also uses ExpressionWithTypeArguments for executable instantiation expressions such as generic<number>, not only heritage/type syntax. Commit 3b24e213 keeps those runtime references throwable, preserves class-extends versus implements/interface behavior, explicitly excludes JSDoc-only nodes, and adds a real-command regression. Clean-review count remains zero and restarts at this HEAD.

samchon:
The restarted full review found three parser-shape gaps: catch binding evaluation was omitted, TypeScript-Go destructuring nodes did not match ESLint's reference classifier, and JSX/type-only names could leak false throw edges. Commit 60442ad0 centralizes the assignment-target classifier, composes catch bindings, preserves JSX and type boundaries, adds real-command TS/TSX regressions, and corrects stale documentation. Clean-review count is reset to zero again at this HEAD.

samchon:
Another restarted full review found two mandatory-entry gaps: an empty switch discarded its discriminant effects, and for-in/of destructuring targets are attached directly to the loop rather than a binary assignment. Commit 0d7608b1 fixes both structural cases and adds real-command positive/negative twins. The clean-review count remains zero and restarts at this HEAD.

samchon:
Full fixture review found that 0_000_000n is scanner-invalid because a separator cannot immediately follow a decimal leading zero. Commit 1c8e8c91 removes that malformed case while retaining valid decimal zero/nonzero and prefixed separator coverage. The clean-review count resets at this HEAD.

samchon:
Exhaustive self-review round 1/3 is clean at exact HEAD 1c8e8c911816cb9d3ade78774be971b06a3d7a26. This pass re-read the complete production diff, every changed and added regression, the shared destructuring-target helper and all impacted callers, the documentation delta, and the full #446 contract. It also revalidated the earlier ClassExpression report against the actual traversal and rejected it as a false positive: only the optional class declaration name is excluded as a reference; class-expression children and runtime heritage remain traversed. No source, test, or documentation change was required. Formatter and local tests remain deferred until all three clean rounds complete at this identical head.

samchon:
Exhaustive review round 2 found a documentation defect in the production source contract: the file header still claimed that all class bodies are never entered, contradicting the implemented and tested preservation of immediately evaluated class heritage and computed names. Commit 19d83586 corrects that contract while retaining the separate code-path boundary for function bodies, class field initializers, and static blocks. No logic or test behavior changed. Because HEAD changed, the clean-review count resets to 0/3 at exact HEAD 19d83586; formatter and local tests remain deferred.

samchon:
Exhaustive self-review round 1/3 is clean at exact HEAD 19d83586c3c0b53e05160a9e70115e15dea6e74c. This restarted pass covered the entire production diff, all changed and added tests, documentation, issue/PR contract, and shared-helper callers. It revalidated completion composition for statements, branches, every loop form, nested switches, labels, catch/finally override behavior, generator resumptions, function/class/JSX/type boundaries, and destructuring references. It also checked BigInt token normalization against the pinned TypeScript-Go scanner and the throwable/reference boundary against ESLint's code-path analyzer. No change was required. Formatter and local tests remain deferred.

samchon:
Exhaustive self-review round 2/3 is clean at exact HEAD 19d83586c3c0b53e05160a9e70115e15dea6e74c. This independent full pass again covered every changed production, test, and documentation file, with a test-shield and regression lens: positive/negative twins for every new completion edge, command-frontdoor diagnostic count/location assertions, generator and finally override cases, TypeScript/JSX/class execution boundaries, and the byte-equivalent relocation of the destructuring-target helper across all existing no-unsafe-assignment callers. The for-in/of extension was traced through both target and source ancestor paths. No finding or change resulted. Formatter and local tests remain deferred pending round 3.

samchon:
Exhaustive self-review round 3/3 is clean at exact HEAD 19d83586c3c0b53e05160a9e70115e15dea6e74c. The final full pass again inspected the entire diff and latent-risk surface, including commit/scope integrity, nil and map behavior, recursion and performance characteristics, shim API availability, pinned upstream semantics, all test fixtures and marker assertions, documentation consistency, forbidden implementation patterns, and PR-to-issue closure metadata. The worktree and remote PR head match, git diff --check is clean, and no change was needed. Three clean exhaustive rounds are now complete at one identical head. Focused local partial tests may now begin; formatter remains prohibited and will not be run.

samchon:
Focused local testing after the prior 3/3 reviews found a real JSX boundary defect: 
oFallthroughNodeIsJSXTagName classified every direct child of a JSX opening/self-closing element as syntax-only, so the JsxAttributes subtree was skipped and embedded/spread references could not reach catch. Commit 9a88a084 now matches TypeScript-Go's structural predicate, requiring parent.TagName() == current after climbing a member chain. Syntax-only intrinsic/member tag names remain excluded, while evaluated attributes are traversed. The temporary AST trace was removed. Because HEAD changed, clean reviews reset to 0/3 at this new exact head; formatter was not run and post-fix focused tests are deferred until the restarted three reviews complete.

samchon:
Exhaustive self-review round 1/3 is clean at exact HEAD 9a88a0849c4a347a78cba74a7227c4d56a2cad35. The restarted whole-PR pass covered all production logic, shared-helper callers, every changed/added regression, and documentation—not only the JSX delta. For JSX it traced intrinsic, member, and namespaced opening/closing tags, self-closing attributes, embedded expressions, spreads, and child expressions against the owning TagName field. The new predicate matches TypeScript-Go's canonical structural check and leaves evaluated subtrees traversable. No temporary diagnostic artifact remains; worktree/remote head and diff integrity are clean. Formatter and post-fix tests remain deferred.

samchon:
Exhaustive self-review round 2/3 is clean at exact HEAD 9a88a0849c4a347a78cba74a7227c4d56a2cad35. This full pass mapped every new production branch to its positive/negative regression and rechecked exact command diagnostic counts/locations, statement/loop/switch/try/finally composition, BigInt grammar, parser-specific reference edges, shared no-unsafe-assignment helper behavior, and all code-path boundaries. The JSX fix was followed through nested member chains and through attribute/child property accesses to ensure only the owning tag-name subtree is excluded. No finding, duplicate test, stale contract, debug artifact, or forbidden pattern was found. HEAD and remote remain identical; formatter and tests remain deferred until round 3.

samchon:
Exhaustive self-review round 3/3 is clean at exact HEAD 9a88a0849c4a347a78cba74a7227c4d56a2cad35. The final whole-PR pass rechecked every implementation, regression, shared caller, and documentation change plus integration metadata and latent risks. The prior focused NoFallthrough run completed all matched cases and isolated the sole failure to the now-corrected JSX ownership predicate; the one-line structural fix is identical to the pinned TypeScript-Go IsJsxTagName condition and cannot suppress attributes or child expressions. No additional finding or artifact remains. Three clean rounds at this identical head are complete; focused post-fix tests now proceed without formatting.

samchon:
Focused local partial verification is green at exact HEAD 9a88a0849c4a347a78cba74a7227c4d56a2cad35 after three clean exhaustive reviews:\n\n- GOFLAGS=-run=NoFallthrough node scripts/test-go-lint.cjs: PASS (ok github.com/samchon/ttsc/packages/lint/linthost 4.387s).\n- GOFLAGS=-run=NoUnsafeAssignment node scripts/test-go-lint.cjs: PASS (ok github.com/samchon/ttsc/packages/lint/linthost 8.645s), covering the existing callers of the relocated shared destructuring-target helper.\n- Worktree is clean; local HEAD, remote branch, and PR head are identical.\n- git diff --check origin/master...HEAD: clean.\n- Formatter was not run.\n\nGitHub CI remains queued/pending and is not being awaited per the task protocol. PR #470 is OPEN and MERGEABLE and carries Closes #446. Independent root verification/ACK has been requested before merge.

PR #520, opened 2026-07-13
fix(lint): quarantine panicking rules per file

### LINKED ISSUE #493: fix(lint): quarantine a panicking rule after its first file-local failure
## Summary

A recovered file-rule panic does not quarantine that rule for the rest of the file, despite the diagnostic claiming that ?ttsc skipped the rule on this file?. A contributor visiting a frequent AST kind can panic and report again at every matching node, causing diagnostic floods and severe repeated work.

- `lintFileWalker.walk` dispatches each bound rule independently for every visited node.
- `runRuleCheck` recovers one invocation, appends a panic finding, and returns no failure state.
- The same bound rule remains in every kind dispatch bucket and is called again on the next node.
- The existing recovery test uses only a `SourceFile` visitor, so it cannot expose repeated invocations.

A rule visiting `Identifier` and always panicking produces one error per identifier instead of one file-scoped failure.

## Required implementation

Quarantine a panicking rule for the remainder of that source file:

- emit exactly one panic finding for that rule/file;
- skip all later nodes and all other registered kinds for the failed rule in that file;
- continue running other rules on the same walk;
- allow the rule to run independently on the next file;
- preserve source anchoring and contributor/built-in recovery boundaries.

Use per-file bound state or an equivalent canonical dispatch design. Do not deduplicate findings after repeatedly invoking the broken rule, recover globally across files, special-case contributor names, monkeypatch the adapter, or add test-only branches.

## Regression shield

Create a rule registered for multiple kinds with several nodes that always panics and assert:

- one `Check` invocation and one panic finding per file;
- no invocation on later kinds/nodes in the same file;
- one fresh invocation/finding in a second file;
- a normal sibling rule still visits every expected node;
- inline directives cannot hide the engine failure unless that is the explicit existing contract;
- CLI/fix/LSP paths terminate without diagnostic flooding.

samchon:
## Summary

- quarantine a panicking rule for the remainder of its current source file
- share the quarantine across every registered node kind without disabling sibling rules or later files
- keep recovered engine failures visible even under inline lint disables
- shield multi-kind dispatch, per-file reset, sibling progress, and diagnostic cardinality

## Review and validation

Three sequential exhaustive reviews of one exact HEAD and focused local validation will be recorded before merge.

Closes #493

### DIFF
--- a/packages/lint/linthost/directives.go
+++ b/packages/lint/linthost/directives.go
@@ -98,7 +98,7 @@ func filterInlineDisabledFindingsWithDirectives(file *shimast.SourceFile, findin
   }
   filtered := findings[:0]
   for _, finding := range findings {
-    if finding == nil || !directives.suppresses(file, finding) {
+    if finding == nil || finding.engineFailure || !directives.suppresses(file, finding) {
       filtered = append(filtered, finding)
     }
   }
--- a/packages/lint/linthost/engine.go
+++ b/packages/lint/linthost/engine.go
@@ -96,10 +96,11 @@ type Context struct {
   Severity         Severity
   Options          json.RawMessage
 
-  rule           Rule
-  isFormat       bool
-  collect        func(*Finding)
-  projectResults publicrule.ProjectResultReader
+  rule             Rule
+  isFormat         bool
+  quarantined      bool
+  collect          func(*Finding)
+  projectResults   publicrule.ProjectResultReader
 }
 
 // DecodeOptions unmarshals the rule's options blob into `out`. Returns
@@ -132,6 +133,8 @@ type Finding struct {
   Fix         []TextEdit
   Suggestions []Suggestion
   IsFormat    bool
+
+  engineFailure bool
 }
 
 // TextEdit is one byte-range source replacement used by a fix or suggestion.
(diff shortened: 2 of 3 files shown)


samchon:
Validation completed at exact HEAD `591deb63be59351b558d4a9de430b96cbd3d4bed`.

- Exhaustive review round 1: clean — full PR diff, dispatch lifecycle, sibling-rule continuation, per-file reset, inline directives, and regression shield.
- Exhaustive review round 2: clean — full PR diff plus parallel file execution, contributor adapter boundary, all finding consumers, and existing panic recovery behavior.
- Exhaustive review round 3: clean — full PR diff and surrounding engine/directive/test code re-read at the same HEAD; no latent-risk finding and no HEAD drift.
- Local validation: `node scripts/test-go-lint.cjs` passed with the workspace `TTSC_TTSX_BINARY` supplied to the isolated worktree (`ok github.com/samchon/ttsc/packages/lint/linthost 110.986s`).
- `git diff --check`: clean.

The first harness attempt exposed only a missing prebuilt `packages/ttsc/lib/launcher/ttsx.js` in this isolated worktree; rerunning the unchanged HEAD against the workspace launcher passed the complete lint Go harness.

PR #524, opened 2026-07-13
fix(lint): bound plugin diagnostic ranges

### LINKED ISSUE #521: fix(lint): validate contributor diagnostic ranges before rendering
## Summary

A contributor can pass an invalid explicit diagnostic range through `rule.Context.ReportRange` or `ReportRangeFix`, and the lint host forwards it into directive filtering, LSP conversion, and typescript-go's native diagnostic renderer without validating it against the current source file. Negative or out-of-bounds positions can therefore panic a normal `ttsc check`/`build` invocation instead of remaining isolated as one bad contributor finding.

- The public reporter contract says explicit ranges are byte offsets inside the current file, but the host must treat contributor input as untrusted.
- `linthost.Context.ReportRangeFix` only changes `end <= pos` to `pos + 1`; it accepts negative positions and offsets past `len(File.Text())`.
- `ReportRangeSuggestion` has the same behavior, and `contextReporter` forwards public contributor calls directly.
- Inline-directive filtering calls the scanner with `finding.Pos` before native rendering.
- `shim/diagnosticwriter.NewLintDiagnostic` again only forces a positive length and does not bound either offset.
- The pinned typescript-go renderer uses the offsets for line lookup, source slicing, and `strings.Repeat`; it assumes valid diagnostic spans and can panic on malformed values.
- Autofix edits already have a separate `selectTextEdits` bounds guard, so that protection does not cover the finding itself.

## Required implementation

Establish one canonical source-range validation policy at the host trust boundary and preserve it through every diagnostic consumer:

- never let a finding carry a negative, reversed, or beyond-EOF source span;
- handle empty files and zero-width EOF diagnostics without manufacturing an out-of-bounds byte;
- keep node-derived built-in findings unchanged when their AST ranges are valid;
- make malformed contributor ranges deterministic and non-panicking across check/build, inline directives, LSP diagnostics/actions, fix, and format paths;
- validate diagnostic spans independently from fix/suggestion edits, whose existing candidate filtering remains intact.

Do not special-case contributor names, recover around the renderer, monkeypatch typescript-go, or add test-only production branches.

## Regression shield

Add table-driven coverage for negative, reversed, beyond-EOF, zero-width EOF, and empty-file ranges through the public contributor adapter. Exercise at least inline-directive filtering, LSP range conversion, and native mixed-diagnostic rendering, and assert valid in-bounds ranges retain their exact span.

samchon:
## Summary

- normalize every source-anchored lint diagnostic through one canonical shim boundary
- keep malformed contributor spans away from inline directives, LSP conversion, the public driver DTO, and typescript-go rendering
- preserve zero-width EOF and empty-file diagnostics without manufacturing out-of-bounds bytes
- shield negative, reversed, beyond-EOF, contributor, suggestion, driver, and native-renderer paths

## Review and validation

Three sequential exhaustive reviews of one exact HEAD and focused local validation will be recorded before merge.

Closes #521

### DIFF
--- a/packages/lint/linthost/engine.go
+++ b/packages/lint/linthost/engine.go
@@ -27,6 +27,7 @@ import (
 
   shimast "github.com/microsoft/typescript-go/shim/ast"
   shimchecker "github.com/microsoft/typescript-go/shim/checker"
+  shimdw "github.com/microsoft/typescript-go/shim/diagnosticwriter"
   shimscanner "github.com/microsoft/typescript-go/shim/scanner"
   publicrule "github.com/samchon/ttsc/packages/lint/rule"
 )
@@ -173,12 +174,13 @@ func (c *Context) ReportFix(node *shimast.Node, message string, edits ...TextEdi
   if c.File != nil {
     pos = shimscanner.SkipTrivia(c.File.Text(), pos)
   }
+  pos, end := shimdw.NormalizeLintRange(c.File, pos, node.End())
   c.collect(&Finding{
     Rule:     c.rule.Name(),
     Severity: c.Severity,
     File:     c.File,
     Pos:      pos,
-    End:      node.End(),
+    End:      end,
     Message:  message,
     Fix:      cloneTextEdits(edits),
     IsFormat: c.isFormat,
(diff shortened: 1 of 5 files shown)


samchon:
Merge gate complete for exact HEAD \471e136e91f01123beb91d948fa19b8507e0827a\. Three sequential exhaustive whole-PR reviews were clean at this unchanged HEAD. Focused local validation passed: contributor diagnostic-range engine paths (including ReportRange, ReportRangeFix, suggestion, inline-directive, foreign-node, LSP, and native rendering coverage), driver diagnostic bounds/rendering paths, and \pnpm --filter ttsc shim:audit\. No formatter was run. No known gap remains.

PR #526, opened 2026-07-14
fix(ttsc): isolate candidate-only call resolution from Checker caches

### LINKED ISSUE #518: fix(ttsc): isolate candidate-only call resolution from Checker caches
## Problem

The candidate-subset call resolver added in #473 invokes the private "(*Checker).resolveCall" method directly. That private method is stateful and is not a read-only applicability query.

This bypasses the lifecycle that TypeScript-Go uses around speculative call resolution:

- "getResolvedSignature" installs the "resolvingSignature" sentinel, manages "resolutionStart", and restores or commits the node cache deliberately.
- "runWithoutResolvedSignatureCaching" snapshots and clears the current/ancestor call "resolvedSignature" entries plus function/arrow "resolvedType" entries, then restores them after signature-help/speculative work.
- "resolveCall" itself writes "signatureLinks.Get(node).resolvedSignature" on overload failure before reporting errors.

Passing a non-nil candidates output suppresses diagnostics; it does not isolate checker state.

## Independently verified impact

"Checker_resolveCallSignatures" supplies a branch-specific signature set from "no-floating-promises". A failed or context-sensitive branch resolution can therefore overwrite the real call expression's cached signature or populate contextual inference caches. Later rules and type queries using the same Program Checker can become rule-order dependent.

The existing tests verify selected return types, but never assert that the same Checker's canonical resolved signature and surrounding contextual caches are unchanged before and after the branch query.

## Required implementation

Provide a canonical, cache-isolated way to resolve an existing call against an explicit candidate set, or redesign the lint query so it never invokes the stateful private resolver directly.

- Preserve the current call and every affected call/function ancestor cache exactly, including an originally nil state.
- Preserve circularity guards and "resolutionStart" behavior.
- Restore state on success, failure, and panic.
- Keep diagnostics suppressed without relying on candidates output as a cache barrier.
- Avoid unsafe struct-layout access, shallow AST copies, monkeypatching, fixture-specific branches, or a partial current-node-only restore.
- If the necessary stable API cannot be expressed safely in the shim, remove the private resolver exposure and use a public/upstream-supported design.

## Regression shield

Using one Checker instance, record canonical signature/return-type and contextual generic behavior before the candidate-only query, then assert they remain identical after:

- a valid overload subset;
- an incompatible subset that enters overload-failure recovery;
- nested calls with contextual generic arrow/function arguments;
- previously unresolved versus already cached nodes;
- repeated queries in different rule orders.

Also run the shim reachability audit and focused "no-floating-promises" mixed-receiver tests.

Related: #455, #473

samchon:
## Summary

- remove the shim bridge that called TypeScript-Go's private, stateful `(*Checker).resolveCall` with an explicit candidate subset
- keep mixed Promise receiver correlation on public Checker queries, selecting only provably applicable overloads and conservatively rejecting uncertain generic mappings
- add same-Checker regression coverage for cached and uncached calls, overload failure, nested contextual generics, repeated order, and explicit generic results

## Rationale

`resolveCall` is not a read-only applicability query. Calling it outside `getResolvedSignature` and `runWithoutResolvedSignatureCaching` could overwrite the real call's cached signature or retain contextual inference state. There is no stable upstream API for transactional resolution against an explicit candidate set, so this change removes the private exposure instead of reproducing Checker cache internals.

## Validation

Local tests intentionally follow the requested sequence: three exhaustive exact-head reviews first, then focused Go tests and the shim reachability audit.

Closes #518


samchon:
Review found two additional overload-proof boundaries at HEAD 761fa63e: TypeScript accepts omitted required parameters whose types admit void, and callback parameter variance depends on declaration kind and strictFunctionTypes. Commit 2d34c9cb now keeps both relations uncertain instead of excluding a potentially unsafe overload, with regression shields for omitted-void arity and method-derived bivariance. The exact-head clean-review count has reset to zero; focused local tests remain deferred until three consecutive exhaustive clean reviews.

samchon:
Correction after pinned Checker source verification: Checker_getMinArgumentCount already incorporates trailing-void omission, so commit 687b8ffd removes the redundant arity uncertainty and non-discriminating test introduced in 2d34c9cb. The declaration-sensitive callback variance fix remains. Review count is reset again; the new exact HEAD is 687b8ffd50c7f4254285b96ea668a7c882ba39c4.

samchon:
A further whole-change review found that both FillMissingTypeArguments calls hard-coded TypeScript mode. Commit b9b0192a now derives JavaScript mode from the signature declaration's NodeFlagsJavaScriptFile bit and adds an end-to-end JS/JSDoc default twin: empty-object defaults normalize to uncertain any, while undefined remains safe. The clean-review count is reset; exact HEAD is b9b0192a4d0d19741fd7575c9b57f7141a72c1cc.

samchon:
Whole-change review also found the latent generic scan inspected only IndexInfo.ValueType. Commit 12528190 now traverses KeyType as well, preventing keyof-T mapped contracts from being treated as concrete, with a direct checker-backed regression test. Review count is reset; exact HEAD is 12528190bcae7294733d1253f1d581931f05e913.

samchon:
Test-hygiene review found two Test functions in the mixed-receiver file, contrary to the repository's one-test-per-file contract. Commit ccbf335e moves the generic applicability matrix into its own named file and adds the required numbered scenario contract without changing behavior. Review count is reset; exact HEAD is ccbf335edc4f8d71786d4660922fdc386d172d58.

samchon:
Resolver-order review found that equivalent overloads from merged declaration blocks are reordered before applicability. Commit 38323967 trusts first-entry order only when all declarations share one parent; otherwise it selects only when every return type is pointer-identical. A merged-interface safe-first/unsafe-later regression now guards the false-negative. Review count is reset; exact HEAD is 38323967b3b02d77864bb70b4efb3297f0c3a70d.

samchon:
A second reorderCandidates boundary was verified: direct literal parameter annotations are specialized and move ahead of equivalent alias-typed overloads even within one declaration parent. Commit f3cfc060 marks that order unstable when a later applicable signature is specialized, and adds a Promise-returning direct-literal .then regression. Review count resets; exact HEAD is f3cfc0606c67dbe5fc02686c3a59e7c09ebd02f0.

samchon:
Callback-contract review found one remaining boolean collapse: comparing actual and expected minimum argument counts does not reproduce TypeScript's effective-count, optionality, and method-variance relation. Commit cadd41b7 keeps that mismatch uncertain and adds a method-derived optional callback overload regression. Review count resets; exact HEAD is cadd41b7d49d60aefcebf68c1455ad6dc33a8ebe.

samchon:
Review follow-up at exact HEAD \97c9a0d1877797b677608c6305241596e0a2600d\: an exhaustive pass found that a context-sensitive callback's cached return type belongs to the original call context. A different overload can provide a narrower return context, so a cached mismatch cannot safely exclude that candidate. The applicability proof now classifies such mismatches as uncertain, and a focused regression fixture locks the contextual string-versus-string-literal case. The three-consecutive-clean-review counter has been reset to zero. Local tests remain deferred until those three full reviews complete.

samchon:
Additional exhaustive-review correction at exact HEAD \14a810abb470dded1571b0c17033135012a1dce4\: the same canonical-context mismatch also affected explicit and constrained generic callback return paths. Those paths now preserve uncertainty for context-sensitive arguments, and the regression fixture covers concrete, explicit-generic, and constrained-generic overload pairs. Review count remains reset at 0/3; no local test was run before the required clean reviews.

samchon:
Test-shield hardening at exact HEAD \63eef14791a9c9283f7929dd76942c56f5cd09d2\: the mapped-index regression could previously pass when the recovered parameter type was nil because the production scan is deliberately conservative on nil. It now requires concrete receiver/property/parameter/index evidence, proves values are concrete, and proves at least one key is latent. The canonical-cache regression now also rejects nil return types before pointer comparisons. Review count is reset to 0/3; local execution is still deferred.

samchon:
Further regression-proof hardening at exact HEAD \3bfeb3457b74f6947139c728e5ab3ebc93946e2f\: index key/value types must now be non-nil before latent-key assertions, and every contextual-overload fixture independently proves its safe twin is applicable before asserting the unsafe candidate blocks selection. This closes remaining false-pass paths through unrelated uncertainty. Clean-review count remains 0/3; no local test has run yet.

samchon:
Independent whole-PR review found a real subtype-priority gap at exact HEAD \3bfeb345\: mutual assignability is not sufficient because TypeScript tries subtype applicability first (notably for numeric enums versus \
umber\). Commit \2f8474d\ narrows stable source-order selection to overloads with identical parameter type syntax and adds an end-to-end numeric-enum regression. No formatter or local tests were run; the three-clean-review gate has reset to 0/3 at exact HEAD \2f8474d\.

samchon:
An independent exact-head review found that Checker.IsContextSensitive is intentionally narrower than candidate-specific contextual typing. Plain array/object literals and implicit callback returns could therefore be compared using a type cached under the canonical overload and incorrectly eliminate an applicable unsafe candidate.\n\nFixed at 6b79fccd: non-generic and generic proofs now fail closed for structural/context-propagating expressions, inline callbacks are checked through the callable proof even when upstream does not mark them inference-sensitive, and implicit callback returns preserve candidate-context uncertainty. Added one regression file covering concrete/generic array arguments and concrete/generic object-return callbacks.\n\nNo local tests or formatter were run. The required three consecutive exhaustive exact-head reviews restart at 0/3 for this commit.

samchon:
Full review of 6b79fccd caught a test-only mismatch with the pinned checker: IsContextSensitive is false for zero-parameter arrows returning a primitive literal. Commit 660a5e12 corrects the regression to assert that upstream boundary and therefore proves the new implicit-return safeguard, rather than depending on a classification upstream never produces.\n\nNo local tests or formatter ran. Exact reviewed HEAD is now 660a5e12d334f15c5309ce43bc2873be3cf29702; the three-clean-review sequence restarts at 0/3.

samchon:
Exact-head review found and fixed another candidate-context cache hazard at ba493ead6cc283764511aeac37d64606f5a8503e. Upstream contextual typing propagates through await/non-null wrappers, template expressions, and generic call return inference; reusing the canonical call's broader cached type could therefore exclude an actually applicable Promise-returning overload. The public proof now recognizes those propagation paths, inspects nested call candidates without invoking the private resolver, and extends the regression shield across concrete and generic overloads. The three-review count is reset to 0/3 for this new exact HEAD; no local tests or formatter have run yet.

samchon:
Exhaustive review found and fixed one additional candidate-context path at exact HEAD bc0b4e3e: TypeScript propagates overload context through \s const\ and \<const>\ assertions. The proof now detects only const assertions via a public AST shim and remains precise for ordinary assertions. The regression covers both the unsafe const-assertion path and the ordinary-assertion control. Review count is reset to 0/3; no local tests or formatter have been run.

samchon:
A second exhaustive-review fix is now pushed at exact HEAD c223bed2. Applicability and return safety are kept separate for naked generic callback returns: an unconstrained callback may be applicable, but an array/object return cached under another overload is no longer accepted as evidence that the instantiated method return is safe. The regression proves the canonical callback cache is \unknown[]\ while fresh generic inference is \Promise<void>[]\. Review count is reset to 0/3; no local tests or formatter have been run.

samchon:
Focused validation disproved the premise behind c223bed2: under TypeScript 5.9, a callback contextually typed as () => unknown[] still exposes a Promise<void>[] return, so no candidate-return cache contamination occurs. Commit a2af6b728449571d8e0d6eccfbba4fe4743f7ef3 removes that unnecessary production guard and its invalid regression premise. It also aligns two fixtures with the public checker surface: mapped types expose no GetIndexInfosOfType entry, so the index regression now uses a real generic index-value contract; and the mixed JavaScript receiver calls were compiler-rejected, so JavaScript default filling is now tested directly through public Checker APIs. Required whole-PR review count is reset to 0/3 at this exact HEAD; local Go tests remain deferred until three consecutive clean exhaustive reviews.

samchon:
Whole-PR review round 1 at a2af6b728449571d8e0d6eccfbba4fe4743f7ef3 was **not clean**. It found a valid end-to-end false negative: Promise<void> | { catch(cb): any } and the corresponding unknown form both compile, but their call result collapses to ny/unknown, so the result-type gate skipped mixed-receiver analysis despite a known Promise branch. TypeScript 5.9 public API probes confirmed both calls have zero diagnostics. Commit 619607dec0403037d8e9476fe0aa34dd4de332f6 routes call expressions through receiver correlation before that early exit, treats uncertain concrete branch returns conservatively, and adds any/unknown mixed regressions plus unrelated-receiver clean controls. Required clean-review count is reset to 0/3 at this exact HEAD; no Go tests have run yet.

samchon:
Clean exhaustive whole-PR review 1/3 completed at exact HEAD 619607dec0403037d8e9476fe0aa34dd4de332f6. Reviewed the complete base-to-head change: all 1,175 added/13 removed production lines in 
ules_promise.go, all nine full regression files (including any/unknown mixed calls and unrelated clean controls), the AST shim export, the complete private resolver removal, and pinned TypeScript-Go 
esolveCall, 
eorderCandidates, applicability, generic default filling, context sensitivity, const assertion, and cache-isolation paths. Rechecked fail-closed behavior for arity, spreads/rests, explicit/default/dependent generics, callback variance/properties/predicates, overload order, contextual wrappers, uncertain returns, and known-safe options. Also verified no new dependency, no remaining private resolver reference, exact one-test-per-file layout, clean diff, clean worktree, and local HEAD = remote HEAD. No defect or test gap found. No local Go test/build/format command was run; testing remains gated until 3/3.

samchon:
Whole-PR review round 2 at 619607dec0403037d8e9476fe0aa34dd4de332f6 was **not clean**. It generalized the prior result-erasure defect: conditional and logical expressions still passed through the aggregate type gate before branch recursion, so valid zero-diagnostic expressions such as lag ? Promise.resolve() : any/unknown, maybePromise && any, and unknown || Promise.resolve() hid their Promise-bearing branch (aggregate types: ny, unknown, ny, and {}). Commit 9fe9bd54d9e3a98aac7990e8ddf870c7451e2036 moves established conditional/logical recursion before aggregate classification and adds all four exact regressions plus no-Promise clean controls to the existing type-shapes shield. Required clean-review count is reset to 0/3 at this exact HEAD; no Go tests have run.

samchon:
Before restarting the clean-review count after 9fe9bd54, the same aggregate-erasure audit found the rule's explicit Promise-array path also depended only on the aggregate element type. Valid literals [Promise.resolve(), any], [Promise.resolve(), unknown], and nested spread equivalents infer as ny[]/unknown[] with zero compiler diagnostics and therefore escaped. Commit 88b1302ef31f44be1556ef379821b0e4748e10b1 adds a syntax-backed fallback over concrete array elements and nested spreads, while preserving uncertain arrays with no Promise element as clean controls. No clean review was claimed for the intermediate HEAD; the required count remains 0/3 and restarts at exact 88b1302e. No Go tests have run.

samchon:
Clean exhaustive whole-PR review 1/3 completed at exact HEAD `88b1302ef31f44be1556ef379821b0e4748e10b1`.

Scope re-read in full: issue/PR contract, complete production diff, all changed regression tests, AST/checker shim changes, pinned TypeScript-Go overload/context/inference behavior, dependency surface, and cache-isolation failure modes. I also verified the public `IsConstAssertion` wrapper against the pinned upstream implementation and confirmed the private candidate-subset `resolveCall` bridge is gone.

Result: clean. No source, test, or documentation changes were made. Local tests remain intentionally deferred until three consecutive clean reviews of this unchanged HEAD are complete.

samchon:
Review finding fixed; the clean-review counter resets to 0/3 at exact HEAD `28defec130561f73b7722fe7b7182cd5545fc332`.

The call-first mixed-receiver path correctly preserved Promise branches erased by `any`/`unknown`, but it also required a rejection handler when the Promise receiver and returned Promise were both explicitly covered by `allowForKnownSafePromises`. That revoked the rule's existing allowlist contract.

The fix now distinguishes unsafe Promise receiver branches from allowlisted branches. Protocol-handler validation still runs for every unsafe Promise branch or unsafe call result, and every non-Promise receiver return is still checked. A regression adds an allowlisted Promise subtype whose handlerless `catch()` returns the same allowlisted subtype; it must remain clean.

No local tests have run at this HEAD; three new consecutive exhaustive whole-PR reviews come first.

samchon:
Review finding fixed; the clean-review counter resets to 0/3 at exact HEAD `a956f60d51e6b0a2c565b877b5fa7ac321017cf4`.

Generic mapped and reverse-mapped parameter types can retain reverse-inference state while their public property and index-info surfaces are empty. Treating such a type as concrete could incorrectly include or exclude an overload, especially when a callable intersection entered the callback proof before the latent-type scan.

The public proof now fails closed on `ObjectFlagsMapped` and `ObjectFlagsReverseMapped`, including mapped constituents nested in unions, intersections, and reference arguments. The latent-type walk also recognizes those flags. A same-Checker regression covers both a plain mapped parameter and a callable mapped intersection, verifies the public member surface is empty, and requires applicability to remain uncertain.

No local tests have run at this HEAD; three new consecutive exhaustive whole-PR reviews come first.

samchon:
Correction: the exact pushed HEAD for the mapped-inference fix is `a956f60d51e6b0a2c565b877b5fa7ac321017cf4` (local and remote match). The review counter is 0/3 at this exact commit.

samchon:
Clean exhaustive whole-PR review 1/3 completed at exact HEAD `a956f60d51e6b0a2c565b877b5fa7ac321017cf4`.

I re-read the complete production diff, all eleven changed regression files, both shim changes, the issue/PR contract, and the pinned TypeScript-Go resolver, arity, contextual-typing, default-filling, and mapped-inference paths. This pass checked cache/order independence, overload tri-state soundness, generic and callable fail-closed boundaries, mapped/reverse-mapped opacity, option preservation, aggregate/array erasure, panic safety, test causality, dependency scope, and removal of the private candidate-subset resolver.

Result: clean. No files changed. Local tests remain deferred until reviews 2/3 and 3/3 are also clean at this unchanged HEAD.

samchon:
Clean exhaustive whole-PR review 2/3 completed at exact HEAD `a956f60d51e6b0a2c565b877b5fa7ac321017cf4`.

This was another complete reread of the production implementation, all eleven changed regression files, both shim changes, and the issue/upstream contracts. I independently traced every public Checker query and tri-state transition, all receiver/result and option combinations, generic/default/callback/mapped inference boundaries, overload ordering, recursive expression/array paths, nil and AST-cast safety, cache identity assertions, fixture markers/counts, and dependency/private-linkname scope.

Result: clean. No files changed. Local tests remain deferred until one final exhaustive clean review of this same HEAD.

samchon:
Clean exhaustive whole-PR review 3/3 completed at exact HEAD `a956f60d51e6b0a2c565b877b5fa7ac321017cf4`.

This final independent whole-change reread covered the complete production implementation, all eleven changed regression files, both shim changes, issue acceptance criteria, and the pinned upstream resolver/inference sources. I revalidated canonical cache isolation, arity and overload precedence, explicit/default/dependent generics, callback variance and contextual returns, mapped/reverse-mapped fail-closed behavior, option semantics, erased call/branch/array types, recursive termination, nil/panic boundaries, regression anti-vacuity, dependency scope, and the absence of private candidate-subset resolution.

Result: clean. The exact HEAD remained unchanged across all three consecutive reviews. The local focused-test gate is now open.

samchon:
Focused validation exposed a stale test expectation, not a production defect. The conservative call-first path now also reports four mixed-receiver calls that TypeScript rejects with TS2349. Commit 5cc141df updates the shield to count those lint findings while independently loading the program and pinning a TS2349 diagnostic to each exact source marker. No production code changed. Per the review protocol, this test edit resets the exhaustive review count to 0/3 at exact HEAD 5cc141df.

samchon:
Review of 5cc141df found that the revised test still used a merged rendered stream for lint locations, allowing a compiler diagnostic at the same line to satisfy that assertion. Commit 4ab094a2 separates the evidence completely: raw Finding positions prove every lint report, while raw programDiagnostics positions independently prove each TS2349 rejection. No production code changed. Exhaustive review count remains reset at 0/3 on exact HEAD 4ab094a2.

samchon:
Exhaustive whole-PR review 1/3 — CLEAN\n\nExact HEAD: 4ab094a28b86f1fa4b0a29fc564b7f159f267184\n\nReviewed the complete merge-base-to-HEAD change across all 14 files: removal of the private stateful resolveCall shim bridge; the full public tri-state applicability proof; mixed Promise/non-Promise receiver routing; generic/default/constraint/callback/mapped/contextual-expression handling; allowlist and aggregate-erasure paths; AST shim export; all cache-identity, overload, generic, callback, mapped, array, option, and raw-diagnostic regression shields. Revalidated issue #518's success/failure/nested/cached/uncached/repeated-order requirements and confirmed compiler errors cannot impersonate raw lint findings. No correctness, cache-mutation, test-adequacy, portability, hardcoding, monkeypatching, or over-optimization finding remains in this pass. No files changed.

samchon:
Exhaustive whole-PR review 2/3 — CLEAN\n\nExact HEAD: 4ab094a28b86f1fa4b0a29fc564b7f159f267184\n\nRe-reviewed the complete 14-file diff independently with an adversarial correctness lens. Covered every production branch and every regression: overload source/subtype/merged ordering; applicable/incompatible/uncertain tri-state propagation; explicit and defaulted generic arguments; dependent constraints; callback arity/variance/returns/properties/predicates/assertions/nominal members; mapped, index, conditional, reference, union, and intersection latent types; candidate-context literals/wrappers/generic calls; Promise/thenable allowlists; any/unknown aggregate and array erasure; optional access; same-Checker cached/uncached/failure/nested/repeated behavior; raw lint and compiler diagnostic provenance; and complete removal of the private resolver exposure. No false-negative, unsafe cache query, brittle fixture premise, test escape, hardcoding, monkeypatching, or unjustified optimization finding remains. No files changed.

samchon:
Exhaustive whole-PR review 3/3 — CLEAN\n\nExact HEAD: 4ab094a28b86f1fa4b0a29fc564b7f159f267184\n\nCompleted a third independent full review of every changed production, shim, and test line. Rechecked issue-to-implementation traceability, public-Checker-only cache safety, all fail-closed branches and safe controls, AST/shim API provenance, recursion and work bounds, option compatibility, diagnostic source/position isolation, same-Checker pointer-identity shields, test determinism and cross-platform paths, absence of dependencies/generated drift/private resolver references, and semantic integration with current origin/master. The three-way merge simulation has no conflict and current master changes no overlapping #518 file. No defect, latent regression, missing shield, portability issue, hardcoding, monkeypatching, or over-optimization finding remains. No files changed.\n\nAll three clean exhaustive reviews are on this identical immutable HEAD; focused local partial validation may now begin.

samchon:
Local validation complete at exact HEAD 4ab094a28b86f1fa4b0a29fc564b7f159f267184.\n\nThree sequential exhaustive whole-PR clean reviews on this identical HEAD:\n1. https://github.com/samchon/ttsc/pull/526#issuecomment-4965704943\n2. https://github.com/samchon/ttsc/pull/526#issuecomment-4965713843\n3. https://github.com/samchon/ttsc/pull/526#issuecomment-4965721813\n\nLocal partial validation (no formatter):\n- targeted 11-test cache/generic/context/mapped/aggregate matrix: PASS (ok github.com/samchon/ttsc/packages/lint/linthost 3.130s)\n- complete no-floating-promises partial suite plus corpus case: PASS (ok github.com/samchon/ttsc/packages/lint/linthost 7.543s)\n- pnpm --filter ttsc shim:audit: PASS (
o new reachable gaps)\n- git diff --check: PASS\n- worktree clean; local HEAD = upstream HEAD = PR HEAD\n- current origin/master merge-tree: clean, with no overlapping #518 file changes\n- private Checker_resolveCallSignatures / checkerResolveCall references: absent\n\nPR is mergeable. GitHub jobs are queued, with no red check; local reviewed validation is complete.

PR #527, opened 2026-07-14
fix(lint): preserve import evaluation order by default

### LINKED ISSUE #523: fix(lint): preserve module evaluation order in format/sort-imports
## Summary

`format/sort-imports` reorders value-bearing import declarations whenever the block has no bare side-effect import. Binding imports still evaluate their dependency modules and their order is observable, so alphabetizing them can silently change program behavior.

- The rule's safety gate only calls `containsSideEffectImport`, which returns true when `ImportClause == nil` (`import './polyfill'`).
- `import {x} from './b'` and `import {y} from './a'` both have clauses, so the rule may rewrite their declaration order from `b, a` to `a, b`.
- Both imported modules execute their top-level code during module evaluation even though bindings are imported.
- A minimal Node ESM witness with each dependency appending its name to shared observable state produced `b,a` before reordering and `a,b` after reordering.
- The existing regression shield covers only bare side-effect imports, matching the incomplete guard.

## Required implementation

Make the formatter's default behavior semantics-preserving:

- do not reorder runtime-bearing import declarations unless the contract contains an explicit, clearly named unsafe opt-in;
- retain useful safe operations such as sorting named specifiers within one declaration;
- reason separately about erased type-only imports and runtime value imports;
- ensure `combineTypeAndValue` cannot accidentally introduce or reorder runtime evaluation outside its documented opt-in contract;
- preserve comments and existing import grouping guarantees.

Do not infer purity from the presence of an import clause, hardcode package names, inspect only bare imports, or monkeypatch test fixtures.

## Regression shield

Use executable ESM witnesses, not source snapshots alone:

- two binding imports whose modules record top-level evaluation order;
- default, namespace, and named value imports;
- type-only imports that are erased at runtime;
- mixed type/value blocks and `combineTypeAndValue` behavior;
- named-specifier sorting still works without declaration reordering;
- bare side-effect imports remain protected.

samchon:
## Summary

- preserve the source order of every runtime-bearing import by default
- keep named-specifier sorting and all-type-only block sorting on the safe path
- require the explicit \unsafeSortRuntimeImports\ option for runtime declaration grouping, sorting, merging, and mixed type/value combining
- document the semantic boundary and update existing declaration-rewrite tests to opt in
- add an executable Node ESM witness plus positive and negative regression shields

## Deferred items

None.

## Test plan

- complete three sequential exhaustive whole-PR reviews at one exact HEAD
- run focused \ormat/sort-imports\ and format-config Go tests only after those reviews
- run the package lint harness as the broader partial validation

Closes #523

### DIFF
--- a/packages/lint/README.md
+++ b/packages/lint/README.md
@@ -105,6 +105,7 @@ export default {
     trailingComma: "all",
     sortImports: {
       order: ["<BUILTIN_MODULES>", "", "<THIRD_PARTY_MODULES>", "", "^[./]"],
+      unsafeSortRuntimeImports: true, // accepts module evaluation reordering
     },
     jsDoc: true,
   },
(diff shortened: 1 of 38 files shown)


samchon:
Exhaustive review 1/3: CLEAN at exact HEAD \eb3bea8cf30cd138364385b23d1095ec20702f8c\.

Reviewed every changed source, config, type surface, README/website hunk, and regression test from scratch. Also traced option decoding, AST phase classification, formatter cascade behavior, comment barriers, Node ESM witness isolation, and backward/unsafe-mode boundaries beyond the diff. No improvement was identified. No tests were run before this review, per the required review-first sequence.

samchon:
Exhaustive review 2/3: CLEAN at exact HEAD \eb3bea8cf30cd138364385b23d1095ec20702f8c\.

Re-read the entire 38-file change from scratch and re-traced runtime/type-only classification, unsafe opt-in propagation, merge gating, named-specifier fallback, executable ESM isolation, cross-platform command/path behavior, docs accuracy, and all updated legacy expectations. Also confirmed intervening master commits do not touch this PR's paths. No improvement was identified. No tests have run yet.

samchon:
Exhaustive review 3/3: CLEAN at exact HEAD \eb3bea8cf30cd138364385b23d1095ec20702f8c\.

Performed a third independent full reading of every changed file and hunk, then rechecked latent correctness, boundary AST shapes, safe/unsafe feature interactions, comment/data preservation, config validation, test determinism and isolation, CI harness compatibility, documentation, and concurrent-master overlap. No improvement was identified. All three clean reviews used the same local, remote, and PR HEAD. Focused local tests may now begin.

samchon:
Local partial validation: PASS at exact HEAD \eb3bea8cf30cd138364385b23d1095ec20702f8c\.

- Focused Go harness: \GOFLAGS=-run=TestFormatSortImports|TestFormatBlockPropagatesSortImportsOptions|TestFormatBlockRejectsNonBoolSortImportsOptions node scripts/test-go-lint.cjs\`n  - result: \ok github.com/samchon/ttsc/packages/lint/linthost 0.758s\`n- TypeScript package build: \pnpm --filter @ttsc/lint build\`n  - result: PASS
- \git diff --check\: PASS
- Formatter: not run

The PR completed three sequential clean exhaustive reviews at this same HEAD and is ready for independent owner verification.

PR #531, opened 2026-07-14
fix(lint): complete Promise executor return semantics

### LINKED ISSUE #513: fix(lint): implement complete no-promise-executor-return control-flow semantics
## Problem

The implementation inspects only concise arrow executors. It misses explicit value returns in block-bodied arrow functions and function expressions, including the core cases "new Promise(() => { return value; })" and "new Promise(function () { return value; })". It also has no public "allowVoid" option.

The current upstream rule tracks Promise-executor code paths, reports every return with a value in that executor only, handles concise arrows, and optionally accepts explicit void returns.

## Acceptance criteria

- Recognize only the global Promise constructor, not a shadowed binding.
- Cover concise arrows, block arrows, and function-expression executors.
- Report value returns on every reachable/nested control-flow path without crossing nested function boundaries.
- Implement and type the "allowVoid" option with canonical defaults.
- Add positive and negative shields for shadowing, nested functions, bare returns, void returns, and multiple return paths.
- Use AST/binding/control-flow semantics; do not match source text or special-case fixtures.

samchon:
## Summary

- recognize only the global `Promise` constructor through TypeScript binding identity
- report concise-arrow values and every executor-owned explicit value return
- stop at nested function boundaries so inner returns keep their own scope
- expose the upstream-compatible `allowVoid` option in runtime and public types
- add positive and negative shields for shadowing, control flow, bare returns, nested functions, and explicit void returns
- document the complete behavior and configuration

## Root cause

The existing implementation matched the callee by identifier text and inspected only concise arrow bodies. It therefore missed block-bodied arrows and function expressions, while reporting locally shadowed constructors.

## Validation plan

Per the required issue lifecycle, focused local validation will run only after three clean exhaustive reviews of this exact PR head. The implementation and shields were compared statically with the current official ESLint rule and corpus before publication.

Closes #513

### DIFF
--- a/packages/lint/README.md
+++ b/packages/lint/README.md
@@ -286,7 +286,7 @@ Source: [ESLint core rules](https://eslint.org/docs/latest/rules/).
 - [`no-octal-escape`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/no-octal-escape.ts): rejects octal escape sequences.
 - [`no-param-reassign`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/no-param-reassign.ts): reject reassigning a function parameter inside the body of the function it belongs to.
 - [`no-plusplus`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/no-plusplus.ts): rejects `++` and `--`.
-- [`no-promise-executor-return`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/no-promise-executor-return.ts): rejects returned values from Promise executors.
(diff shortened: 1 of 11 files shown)


samchon:
Exhaustive review found two test-harness gaps and reset the clean-review count. The corpus test now runs through a real Program/checker instead of the parser-only helper, and all negative shields are valid TypeScript so compiler diagnostics cannot preempt lint assertions. Updated head: \63637430\. No local tests have run yet; the required three clean exhaustive reviews restart from this head.

samchon:
A restarted exhaustive review found two stale descriptions: the website rule index still described the old generic return check, and one shield comment said script-level after the case became a valid module-level shadow. Both are corrected at head \

PR #546, opened 2026-07-14
Enforce behavioral witnesses for public lint rules

### LINKED ISSUE #494: test(lint): require a behavioral witness for every public built-in rule
## Summary

The registry-parity gate proves only that a public TypeScript rule key has a Go object with the same name. Fourteen production no-op stubs satisfy that gate today, and matching corpus fixtures explicitly skip themselves as ?not yet implemented?. The test architecture therefore certifies API parity while exercising no behavior.

## Evidence

- `TestTypedKeysMatchRegisteredRules` checks names, not that an enabled rule can diagnose a valid positive case.
- Multiple built-ins register an empty `SourceFile` check or `Visits() == nil` specifically so the parity test passes.
- Their linked fixtures use `@ttsc-corpus-skip: ... not yet implemented`, allowing docs and typed keys to advertise inert rules.

Issues #476?#489 track the current silent stubs individually. This issue closes the systemic escape hatch.

## Required implementation

Add a maintainable behavioral-coverage contract for every publicly registered built-in rule:

- each rule must have at least one real positive command-path witness that proves enabling it can produce its own diagnostic under valid prerequisites;
- rules requiring options, filenames, sibling files, a checker, project state, or platform conditions may point to a dedicated harness test, but the witness must execute production dispatch;
- corpus skips must be limited to explicit harness constraints and reference the positive replacement test;
- ?not yet implemented? cannot coexist with a registered/public rule;
- generated/manifest data must be derived or mechanically audited so new rules cannot bypass the gate by omission.

Do not inspect source text for an empty function body as the sole proof, add a hardcoded allowlist of current stubs, fabricate diagnostics in the harness, monkeypatch rule dispatch, or add test-only production branches.

## Regression shield

- Add a deliberately inert test rule/key in a fixture and prove the audit rejects it.
- Prove option-, filename-, project-, checker-, and platform-dependent real rules can register valid witnesses.
- Assert every current public built-in key has exactly one discoverable positive coverage route.
- Keep typed-key/registry parity and unknown-rule tests intact.

samchon:
## Summary
- record a rule witness only after a real engine or command diagnostic is observed
- audit the complete public registry after the native lint suite and publish one deterministic route per rule
- reject inert-rule fixtures and validate option, filename, project, checker, and platform witness classes
- restrict corpus skips to explicit harness constraints that reference an existing positive Go test

## Validation
Formal exhaustive review rounds and focused local tests follow on this ready PR HEAD, per project workflow.

Closes #494


samchon:
Rebased onto current `master` (the branch was ~129 commits behind; all eleven formerly-stubbed rules — boundaries/dependencies and the ten unicorn/* rules — are now real built-ins on master). Reconciliation notes:

- The eleven newly-implemented rules receive genuine positive witnesses automatically: their merged Go tests flow through the instrumented shared helpers (`assertRuleCorpusCase`, `runRuleFindingsSnapshot*`, `assertFixSnapshot`, `captureCommandOutput`), each recording only after a real production-dispatch diagnostic. The full `linthost` suite (including the `TestMain` witness audit) passes locally.
- `unicorn/no-unnecessary-polyfills` landed on master after this branch forked, still carrying the pre-implementation `// @ttsc-corpus-skip:` (legacy shape, no constraint, absent from the manifest). That is exactly the hidden-exclusion escape hatch this gate closes, so it is converted to the audited `// @ttsc-corpus-skip(options):` directive referencing `unicorn_no_unnecessary_polyfills_options_test.go` and added to `behavioral_witness_exclusions.json`. The single-file corpus cannot express the resolved target environment, so the diagnostic is proven through production dispatch in that Go harness.
- No formatter was run locally (repo-wide CRLF rewrite risk on this machine); formatting is left to the CI format gate.

Note: the TypeScript e2e corpus lanes cannot run on this local Windows environment (file:// URL limitation), so they are verified via CI; the native `linthost` Go suite was run locally and passes.

PR #554, opened 2026-07-14
fix(lint): preserve entry-scoped formatting eligibility

### LINKED ISSUE #491: fix(lint): preserve entry-scoped ignores and defaults in format mode
## Summary

The format-command resolver collapses entry-scoped applicability into two project-wide booleans. This disables default formatting for valid files and promotes scoped ignores into global format exclusions.

## Independently verified failures

1. **Ignores-only config disables every unignored file.** When a config contains only an ignore-only entry, `ConfigStore.ResolveRules` correctly marks unmatched files as not ignored. However, `fileMatchesAnyEntry` treats ?all entries are IgnoreOnly? as out of scope, so the default format-rule upgrade never runs. A config such as `{ ignores: ["generated/**"] }` makes format mode a no-op even for `src/main.ts`.
2. **One entry's scoped ignore suppresses another matching entry.** `fileIsIgnoredByEntry` returns true if any non-ignore-only entry's `ignores` matches. For a base `{ files: ["src/**"], ignores: ["src/generated/**"] }` and a global child with `format: {}`, `src/generated/x.ts` is excluded only from the base contribution; the child still applies. The formatter currently skips it entirely.

This disagrees with the entry-merging behavior already used for lint rules.

## Required implementation

Resolve formatter eligibility from the same per-file entry semantics as normal rule resolution:

- a global ignore-only entry still skips matching files;
- unmatched files remain eligible for default formatting when the store is empty or contains only ignore-only entries;
- a scoped ignore removes only that entry's contribution;
- another matching base/child entry can still contribute format settings;
- files outside every applicable non-ignore entry remain untouched when that is the actual config scope.

Replace the blanket boolean model with a canonical resolved representation or equivalent single entry fold. Do not add path-specific exceptions, duplicate glob semantics, hardcoded roots, monkeypatches, or test-only branches.

## Regression shield

Add real format-command tests for:

- ignores-only config: ignored and unignored files;
- scoped-ignore base plus global child format;
- scoped-ignore child plus base format;
- files outside every scoped entry;
- empty config and no config;
- extends chains, Windows/POSIX paths, negated/overlapping patterns where supported;
- exact output plus second-pass idempotence.

Prove normal lint resolution and LSP/in-memory formatting remain consistent.

samchon:
## Summary

- resolve per-file applicability in the same config fold as severities and options
- preserve global ignores while allowing another matching extends entry to contribute formatting
- keep default formatting active for empty and ignore-only configs, and skip files outside every scoped entry
- remove duplicate concrete-ConfigStore glob inspection from the format wrapper
- add real CLI, normal resolution, LSP in-memory, mixed-separator, extends, exact-output, and second-pass idempotence shields

## Verification

- three sequential exhaustive whole-PR reviews passed on exact HEAD \4014f87da941c42c0b902374fd079cdb0ee817ff\`n- focused Go selection passed: new CLI/LSP applicability shield plus adjacent format/config regressions (\ok .../linthost 1.022s\)
- \pnpm --filter @ttsc/lint build\ passed
- merge-tree against current master \4a02e14b\ passed
- local/remote HEAD match; diff-check and worktree are clean
- formatter was not run

Closes #491

### DIFF
--- a/packages/lint/linthost/config.go
+++ b/packages/lint/linthost/config.go
@@ -106,7 +106,10 @@ type ProjectRuleSetting struct {
 // its severity.
 //
 // `Ignored` means an `ignores`-only config entry matched the file and the
-// engine should skip linting it entirely.
+// engine should skip linting it entirely. `OutOfScope` means the store has at
+// least one rule-bearing entry but none applies to this file. Keeping the two
+// states distinct lets wrappers preserve entry-local ignores: one entry may
+// reject a file while another matching entry still contributes rules.
 type ResolvedRuleConfig struct {
   Rules           RuleConfig
   Options         RuleOptionsMap
@@ -115,6 +118,7 @@ type ResolvedRuleConfig struct {
   // RuleResolver.RuleOptions.
   OptionsResolved bool
   Ignored         bool
+  OutOfScope      bool
 }
 
 // RuleOptions returns the file-resolved option payload for name. Built-in
@@ -438,10 +442,17 @@ func (s *ConfigStore) ResolveRules(fileName string) ResolvedRuleConfig {
   }
   out := RuleConfig{}
   options := RuleOptionsMap{}
+  hasEntries := false
+  matchedEntry := false
   for _, entry := range s.entries {
-    if entry.IgnoreOnly || !entry.matchesFile(fileName) {
+    if entry.IgnoreOnly {
       continue
     }
+    hasEntries = true
+    if !entry.matchesFile(fileName) {
+      continue
+    }
+    matchedEntry = true
     for name, sev := range entry.Rules {
       canonical := normalizeBuiltinRuleName(name)
       out[canonical] = sev
@@ -450,7 +461,12 @@ func (s *ConfigStore) ResolveRules(fileName string) ResolvedRuleConfig {
       }
     }
   }
-  return ResolvedRuleConfig{Rules: out, Options: options, OptionsResolved: true}
+  return ResolvedRuleConfig{
+    Rules:           out,
+    Options:         options,
+    OptionsResolved: true,
+    OutOfScope:      hasEntries && !matchedEntry,
+  }
 }
 
 // ActiveRuleNames implements RuleResolver. Returns the sorted union of all rule
--- a/packages/lint/linthost/format.go
+++ b/packages/lint/linthost/format.go
@@ -119,15 +119,6 @@ type formatCommandResolver struct {
   // around as a value) shares the same underlying cache rather than copying a
   // sync.Once lock.
   ruleNames *formatRuleNamesCache
-  // entryDecisions memoizes the per-fileName ignored/matches decision that
-  // ResolveRules derives from fileIsIgnoredByEntry and fileMatchesAnyEntry.
-  // Both walk store.entries with glob matching and are recomputed identically
-  // for the same fileName on every format pass, so the result is cached once
-  // per file and reused. The field is a pointer (a *sync.Map) so the nil zero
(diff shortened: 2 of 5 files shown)


samchon:
Exhaustive review 1/3 — CLEAN on exact HEAD \c2a6aa1b80a0cde081180b1f4ad10572a29ade5e\.

Reviewed the complete production and test diff plus every ResolvedRuleConfig constructor and formatter/LSP wrapper call path. Verified: global ignore short-circuiting remains unchanged; applicability is derived in the same entry fold as severity/options; an ignored entry cannot suppress another match; empty and ignore-only stores remain default-eligible; files outside all rule-bearing entries are excluded; zero-value custom resolvers remain compatible; boundProjectRuleResolver preserves the new state; duplicate concrete-store glob logic is fully removed; real command tests cover both extends directions, outside scope, no/empty/ignore-only configs, native/POSIX path forms, exact output, and second-pass idempotence. No hardcoding, monkeypatch, test-only production branch, dependency, formatter run, or actionable finding.

samchon:
Exhaustive review 1/3 — CLEAN on restarted exact HEAD \4014f87da941c42c0b902374fd079cdb0ee817ff\.

Reviewed the entire production/test diff and all surrounding config parsing, extends folding, default-option detection, bound resolver, CLI, disk LSP, and in-memory LSP paths. Rechecked every issue acceptance arm, global-versus-scoped ignore precedence, zero-value compatibility, empty/ignore-only stores, overlapping entries in both directions, truly global child contribution, out-of-scope behavior, native/POSIX separator normalization, exact first-pass output, second-pass idempotence, and no-edit LSP behavior. Local and remote HEADs match; diff-check and worktree are clean. No actionable finding; no tests or formatter run.

samchon:
Exhaustive review 2/3 — CLEAN on exact HEAD \4014f87da941c42c0b902374fd079cdb0ee817ff\.

Repeated the whole review independently, including engine construction/per-file dispatch, project-rule ignore handling, format option enumeration, wrapper unwrapping, concurrent ResolveRules use, custom resolver compatibility, extends ordering, parser-produced empty/unknown entries, and the complete regression oracle. Confirmed OutOfScope only suppresses formatter promotion while ordinary lint naturally receives an empty per-file rule map; project rules retain their intentional global-ignore-only policy. The shield would fail for the former concrete-wrapper bug, blanket scoped-ignore bug, ignore-only default bug, outside-scope promotion, CLI/LSP divergence, or non-idempotent output. No actionable finding; exact HEAD remains clean and unformatted, tests not yet run.

samchon:
Exhaustive review 3/3 — CLEAN on exact HEAD \4014f87da941c42c0b902374fd079cdb0ee817ff\.

Final full-scope pass revalidated implementation, all tests, public struct compatibility, no duplicated glob semantics, error/ignore precedence, platform path handling, exact/idempotent command and LSP oracles, and latent integration risks. The flattened Go test runner sees 1,909 unique basenames with no collision; Lore commits are complete; local/remote head identity, diff-check, clean worktree, and merge-tree against current master \4a02e14b\ (tree \8e224b28\) all pass. No finding. Exactly three clean reviews now hold on this HEAD; proceeding to focused local tests without running a formatter.

samchon:
Focused verification passed after exactly three clean reviews on \4014f87da941c42c0b902374fd079cdb0ee817ff\: selected CLI/LSP/config Go tests PASS (1.022s), @ttsc/lint TypeScript build PASS, diff-check PASS, worktree clean, formatter not run. Root fact-check confirms the implementation is generic and the shield covers every issue acceptance arm. Merging immediately.

PR #555, opened 2026-07-14
fix(lint): align no-await-in-loop execution positions

### LINKED ISSUE #475: fix(lint): align no-await-in-loop positions and implicit awaits with ESLint
## Summary

The native `no-await-in-loop` port treats any `AwaitExpression` with a loop ancestor as looped and visits no implicit-await syntax. This differs from ESLint's execution-position model in both directions.

## Independently verified failures

- False positives: `await` in a `for` initializer and in the right-hand side of `for...of` is evaluated once, but the current ancestor-only walk reports it.
- False negative: a `for await...of` nested inside an ordinary loop is itself an implicit await and must report on the inner `ForOfStatement`.
- False negative: `await using` declarations in looped positions are implicit awaits and are never visited.
- The implementation uses a 24-byte source-text heuristic for `for await`, even though the TypeScript-Go AST already exposes `AwaitModifier` and `NodeFlagsAwaitUsing`.

Authoritative ESLint source: https://github.com/eslint/eslint/blob/main/lib/rules/no-await-in-loop.js

## Required implementation

Port ESLint's current `validate` / boundary / `isLooped` semantics onto the TypeScript-Go AST:

- visit explicit `AwaitExpression`, async `ForOfStatement`, and await-using declaration lists;
- classify only actually repeated positions (for test/update/body; for-in/of body and await-using left; while/do test/body);
- preserve function and for-await boundaries;
- use typed AST fields/flags, not source windows or keyword guessing.

Do not add hardcoded syntax exceptions, monkeypatches, test-only production branches, or formatter churn.

## Regression shield

Add real command-path positive/negative twins for:

- `for` initializer versus test/update/body;
- `for...in/of` RHS versus body;
- nested `for await...of` inside ordinary loops and a top-level for-await boundary;
- `await using` in each relevant loop position;
- nested functions and class/static boundaries;
- comments and unusual spacing proving the textual heuristic is gone.

Assert exact diagnostic nodes/ranges and counts.

samchon:
## Intent

Align `no-await-in-loop` with ESLint's execution-position model so single-shot loop operands remain clean while every explicit or implicit await repeated by a loop is reported.

## Scope

- replace the ancestor-only scan with the upstream candidate, boundary, and repeated-position traversal
- use typed TypeScript-Go `AwaitModifier` and `NodeFlagsAwaitUsing` surfaces instead of source-text keyword windows
- visit explicit await expressions, async for-of statements, and await-using declaration lists
- preserve function, method, and intentional for-await boundaries
- add exact-range and real check-command coverage for all loop positions, implicit awaits, comments/spacing, and negative twins
- expand the public corpus and correct README/website behavior descriptions

## Test plan

Per the delivery workflow, local runtime validation follows three sequential exhaustive exact-HEAD reviews. Planned verification: focused native Go regression, real TypeScript lint corpus, `@ttsc/lint` build, and diff integrity. No formatter will be run.

Closes #475

### DIFF
--- a/packages/lint/README.md
+++ b/packages/lint/README.md
@@ -231,7 +231,7 @@ Source: [ESLint core rules](https://eslint.org/docs/latest/rules/).
 - [`no-alert`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/no-alert.ts): rejects `alert`, `confirm`, and `prompt`.
 - [`no-array-constructor`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/no-array-constructor.ts): rejects `Array` constructor calls.
 - [`no-async-promise-executor`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/no-async-promise-executor.ts): rejects async Promise executors.
-- [`no-await-in-loop`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/no-await-in-loop.ts): reject `await` expressions evaluated inside a loop body.
+- [`no-await-in-loop`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/no-await-in-loop.ts): reject explicit and implicit awaits evaluated in repeated loop positions.
 - [`no-bitwise`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/no-bitwise.ts): rejects bitwise operators.
 - [`no-caller`](https://github.com/samchon/ttsc/blob/master/tests/test-lint/src/cases/no-caller.ts): rejects `arguments.caller` and `arguments.callee`.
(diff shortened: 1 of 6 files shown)


samchon:
Exhaustive review found two post-open improvements and reset the clean-review count to 0/3. The rule inventory now describes explicit and implicit awaits in repeated positions, and the nested for-await fixture now contains an explicit body await so the exact native test and real corpus reject duplicate diagnostics across the intentional async-iteration boundary. New exact head: 898ecde7f15d585311c1ec1f18d9f4e8a98daa15. Runtime tests remain intentionally deferred until three clean reviews; no formatter was run.

samchon:
Final exact head 898ecde7f15d585311c1ec1f18d9f4e8a98daa15 (tree 1ae6c30396c934a57fe66aca69850e0f1ede440b) completed three clean exhaustive post-change reviews. Local verification: full lint Go suite passed in 102.915s; the real no-await-in-loop TypeScript corpus passed in 18.2s; @ttsc/lint build passed; freshly built ttsc launcher passed the corpus; git diff --check is clean; worktree and remote head match. No formatter was run. Awaiting independent root fact-check before merge.

PR #558, opened 2026-07-14
feat(lint): implement boundaries/dependencies policy enforcement

### LINKED ISSUE #476: feat(lint): implement boundaries/dependencies instead of a silent registry stub
## Summary

`boundaries/dependencies` is exposed in the public typed rule map, README, website, declaration-file routing, and registry, but its `Check` only decodes options and deliberately emits nothing. The test suite explicitly pins zero findings for a dependency that the configured policy disallows.

## Evidence

- `packages/lint/linthost/rules_boundaries.go`: the rule is documented as a v1 stub and performs no validation.
- `packages/lint/test/rules/boundaries/boundaries_dependencies_test.go`: asserts that a disallowed app-to-domain import produces zero findings.
- `packages/lint/src/structures/rules/ITtscLintBoundariesRules.ts`, README, and website advertise a working unified dependency-direction rule.

This is a public-contract violation and a test-pinned production no-op.

## Required implementation

Implement the current upstream `boundaries/dependencies` direction-aware policy semantics using the existing boundary classification, dependency collection, option decoding, project identity, and legacy rule helpers where applicable. Preserve documented options and declaration-file behavior. Do not keep a compatibility no-op, delegate to hardcoded fixtures, monkeypatch the engine, or add test-only production branches.

## Regression shield

Replace the zero-finding stub test with full positive/negative matrices for allow/disallow/default policy, from/to direction, relative and external dependencies, aliases/re-exports/dynamic imports, entry/private/unknown behavior, declaration files, Windows/POSIX paths, custom messages, and malformed options. Exercise the real command path and exact ranges/counts.

samchon:
## Intent

Make the published `boundaries/dependencies` rule enforce architecture contracts instead of silently accepting every dependency.

## Scope

- implement ordered direction-aware policies with current upstream precedence
- preserve legacy string selectors while adding entity and dependency metadata selectors
- resolve TypeScript path aliases through the checker and retain relative-path fallback
- collect imports, re-exports, import-equals, dynamic imports, static requires, and import-type nodes
- gate external/core, unknown-local, and internal dependencies explicitly
- validate malformed options before dispatch
- update public types, rule documentation, and the consumer corpus note
- replace the no-op witness with exact-range, command-path, declaration, alias, visibility, path, message, and option regressions

## Test plan

- complete three sequential exhaustive whole-PR reviews on one exact head
- run the focused boundaries Go regression set through the package test harness
- build `@ttsc/lint` and type-check the public option surface
- leave the full matrix to GitHub Actions after the focused local proof

## Deferred

None within the repository's documented `elements`-based selector model.

Closes #476

### DIFF
--- a/packages/lint/linthost/rules_boundaries.go
+++ b/packages/lint/linthost/rules_boundaries.go
@@ -18,17 +18,6 @@ type boundariesEntryPoint struct{}
 type boundariesNoPrivate struct{}
 type boundariesNoUnknown struct{}
 
-// boundariesDependencies is a v1 stub for the upstream unified
-// `boundaries/dependencies` rule, which replaces the legacy
-// `element-types` / `entry-point` / `external` / `no-private` /
-// `no-unknown` rules with a single direction-aware policy block.
-//
-// The native engine accepts the same `elements` + `rules` config shape so
-// projects can claim the rule id today, but the full direction
-// validation is deferred. Until then this rule registers, decodes
-// options without crashing, and emits no diagnostics.
-type boundariesDependencies struct{}
-
 func (boundariesElementTypes) Name() string { return "boundaries/element-types" }
 func (boundariesExternal) Name() string     { return "boundaries/external" }
 func (boundariesEntryPoint) Name() string   { return "boundaries/entry-point" }
@@ -42,15 +31,6 @@ func (boundariesEntryPoint) Visits() []shimast.Kind   { return []shimast.Kind{sh
 func (boundariesNoPrivate) Visits() []shimast.Kind    { return []shimast.Kind{shimast.KindSourceFile} }
 func (boundariesNoUnknown) Visits() []shimast.Kind    { return []shimast.Kind{shimast.KindSourceFile} }
 func (boundariesDependencies) Visits() []shimast.Kind { return []shimast.Kind{shimast.KindSourceFile} }
-
-// Check is intentionally a no-op for the v1 stub. The rule accepts the
-// upstream config shape via the shared `boundariesOptions` decoder so
-// downstream configs type-check and load, then exits without reporting
-// findings. Replace this body when the unified direction validation
-// lands.
-func (boundariesDependencies) Check(ctx *Context, node *shimast.Node) {
-  _ = decodeBoundariesOptions(ctx)
-}
 
 func (boundariesElementTypes) Check(ctx *Context, node *shimast.Node) {
   opts := decodeBoundariesOptions(ctx)
@@ -239,6 +219,9 @@ type boundaryDependency struct {
   node      *shimast.Node
   specifier string
   relative  bool
+  kind      string
+  nodeKind  string
+  specifiers []string
 }
 
 func decodeBoundariesOptions(ctx *Context) boundariesOptions {
@@ -251,6 +234,19 @@ func decodeBoundariesOptions(ctx *Context) boundariesOptions {
 
 func collectBoundaryDependencies(node *shimast.Node) []boundaryDependency {
   out := []boundaryDependency{}
+  appendDependency := func(sourceNode *shimast.Node, specifier, kind, nodeKind string, specifiers []string) {
+    if sourceNode == nil || specifier == "" {
+      return
+    }
+    out = append(out, boundaryDependency{
+      node:       sourceNode,
+      specifier:  specifier,
+      relative:   isRelativeBoundarySpecifier(specifier),
+      kind:       kind,
+      nodeKind:   nodeKind,
+      specifiers: specifiers,
+    })
+  }
   var walk func(*shimast.Node)
   walk = func(n *shimast.Node) {
     if n == nil {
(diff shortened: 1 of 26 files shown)


samchon:
Follow-up commit `feat(lint): complete boundaries/dependencies upstream semantics` finishes the port on top of the initial enforcement commit and rebases onto current master. Closes #476.

What changed since the opening commit:

- the extended dependency collector (import-equals, dynamic `import()`, static `require()`, `import()` type nodes) is now gated to the unified rule; the legacy split rules keep their shipped static import/export surface, pinned by `boundaries_element_types_ignores_dynamic_syntax_test.go`
- legacy string effects follow the upstream open-direction contract: they select targets under an outer `from` and sources under an outer `to` (`boundaries_dependencies_legacy_to_direction_test.go`)
- `import("m")` type nodes classify as `type` dependencies and `typeof import("m")` stays `typeof` (`boundaries_dependencies_import_type_kinds_test.go`); the syntax matrix also covers `import = require(...)` now
- bare Node built-ins (`fs`) classify as `core` origin like `node:`-prefixed specifiers (`boundaries_dependencies_bare_core_builtin_test.go`)
- policy-level `importKind` is validated and gated with `dependency.kind` precedence (`boundaries_dependencies_import_kind_test.go`); option errors for the `rules` alias name the alias instead of `policies`
- docs and typed options list the complete message placeholder set

Local validation on the rebased head:

- `go test -count=1 -run TestBoundaries ./linthost` (scratch module mirroring `scripts/test-go-lint.cjs`): pass, 24 tests
- full `go test -count=1 ./linthost`: pass
- `tsc --noEmit -p packages/lint/tsconfig.json`: pass

samchon:
## Emergency wrap-up status (work paused, not finished)

Development on this branch was **stopped mid-flight** by an agent quota limit. This comment records the honest state of the branch as pushed. **Do not assume this is validated.**

### What landed (2 commits, rebased onto current `master`)
- `f3d7551ed` feat(lint): enforce published boundary dependency contracts
- `70049d7ee` feat(lint): complete boundaries/dependencies upstream semantics

26 files, +2215 / -91:
- `packages/lint/linthost/rules_boundaries_dependencies.go` (new, ~1013 lines) — the unified direction-aware policy engine
- `packages/lint/linthost/rules_boundaries.go` — rule wiring; `boundaries/dependencies` is registered and no longer a no-op
- `packages/lint/src/structures/rules/ITtscLintBoundariesRuleOptions.ts`, `ITtscLintBoundariesRules.ts` — typed options surface
- 19 new Go test functions across `packages/lint/test/rules/boundaries/`
- `tests/test-lint/src/cases/boundaries-dependencies.ts` corpus case
- `website/src/content/docs/lint/rules/boundaries.mdx` docs

The implementation contains **no TODOs, stubs, or panics** — it reads as complete.

### What is NOT verified
- **Tests were not run by this wrap-up pass.** The predecessor's notes claim the Go boundaries tests were built and run green, but that is **unverified here** — treat the 19 tests as *known-unrun* until CI says otherwise.
- **`pnpm format` was NOT run.** It triggers a repo-wide line-ending rewrite on the authoring machine. Formatting/lint-format CI may fail; that is expected and cosmetic.
- **Local build could not be checked** — this worktree lacks the fetched `typescript-go/shim` Go module deps (`missing go.sum entry`, affecting all pre-existing `linthost` files, not the new code). Environmental, not a branch defect. CI is the real signal.
- The first commit's own trailer states `Not-tested: Local tests and builds are intentionally deferred`.

### Next steps for triage
1. Let CI run; treat any failure as expected-unfinished, not a regression.
2. Run `pnpm format` on a machine with sane line endings.
3. Issue #476 stays **OPEN**.

Not merging, not waiting for CI.

samchon:
## Verification complete — tests run green, merging

Resuming the paused branch. The prior wrap-up flagged the 19 boundaries tests as *known-unrun*; they are now executed and green.

### Validation on the rebased head

- `node scripts/test-go-lint.cjs` (the unmodified package runner: scratch module + go.work over in-tree shims, `go test -count=1 ./linthost`): **pass** — `ok github.com/samchon/ttsc/packages/lint/linthost`, exit 0. This covers every new `boundaries/dependencies` test (24 boundaries tests total, including the real `check` command front-door witness) plus the whole `linthost` package.
- Re-ran the same runner after rebasing onto current `master` (which had advanced past the sibling unicorn PRs): **pass** again, clean.

### Review

Exhaustive self-review against the upstream oracle (`javierbrea/eslint-plugin-boundaries` `Dependencies.ts`): last-match-wins across policies, `disallow` precedence within a policy, message hierarchy (policy → global → default), and the `policies`/`rules`/`default`/`message`/`checkAllOrigins`/`checkUnknownLocals`/`checkInternals` option surface all match. Cross-platform path normalization, the unified-only extended collector gating, and the declaration-file route verified. No code changes were required — the inherited implementation was complete and correct; only execution was outstanding.

### Notes

- The corpus fixture `tests/test-lint/src/cases/boundaries-dependencies.ts` is documentation-only, identical in pattern to the five merged sibling boundaries fixtures: the single-file corpus runner cannot express the multi-file `elements` project layouts a boundary diagnostic needs, so the behavioral contract lives in the Go suite. The corpus runner enumerates only fixtures carrying `// expect:` annotations, so this fixture is correctly excluded without a skip directive.
- `pnpm format` was intentionally not run (it triggers a repo-wide line-ending rewrite on this machine); any format-lane diff is cosmetic and left to CI.

Closes #476.

PR #570, opened 2026-07-14
feat(test): cover monorepo edge cases for graph dump and vscode server resolution

samchon:
## Intent

Two coverage holes surfaced while auditing how `@ttsc/graph` and the `@ttsc/vscode`/`ttscserver` stack behave in pnpm/yarn monorepos with multiple `tsconfig.json` files. Both behaviors already work; neither was pinned by a test. This adds the missing tests only — no product code changes.

## Scope

- **`test(graph)`** — `test_ttscgraph_dump_survives_unresolved_workspace_import`
  Locks the fail-soft path of `ttscgraph dump` when a workspace specifier cannot resolve (Yarn PnP with no `node_modules`, an incomplete install, or a not-yet-linked sibling). The dump must still load, keep the importing package's nodes, and drop the unresolved sibling without leaking a `node_modules` path — one missing link must not take down a whole monorepo's graph.

- **`test(vscode)`** — `test_vscode_server_resolution_finds_package_ttsc_without_root_install`
  Pins the common shape where `ttsc` is installed in each `packages/*` but not in the root `package.json`. `resolveTtscServerLauncher` walks up from the open file's directory, so a package file finds its own `node_modules/ttsc` while the root-anchored candidate resolves to nothing — without throwing — and is dropped.

Both follow the existing peers: the graph test mirrors `test_ttscgraph_dump_resolves_pnpm_workspace_edges`; the vscode test mirrors `test_vscode_server_resolution_uses_package_json_anchor`. Each is one `test_<snake_case>` per file with the standard three-part doc comment, and both exercise the negative/degrade direction rather than a happy path.

## Deferred

- No end-to-end LSP test that a `ttscserver` launched at package A serves cross-package go-to-definition into package B's symlinked source — that path is delegated to the upstream tsgo LSP and is out of scope here.
- `graphsymbols` "find references" remains scoped to the launched package's program (downstream references in sibling packages are not found); noted as a known limitation, not addressed here.

## Test plan

- `test_vscode_server_resolution_finds_package_ttsc_without_root_install` runs locally and passes; all sibling `test_vscode_server_resolution_*` / `multi_root` tests still pass (they import the real `serverResolution.ts` helpers through Node's TS loader; no native binary needed).
- `test_ttscgraph_dump_survives_unresolved_workspace_import`: the full `test-graph` harness needs a built `@ttsc/graph` + native binary, which could not be built locally (an unrelated `@ttsc/unplugin` rollup build failure blocks `pnpm build:current`). The asserted behavior was verified directly against a locally built `ttscgraph` binary running the identical `dump` code path: exit 0, importing-package node retained, unresolved sibling absent, no `node_modules` leak.
- CI intentionally not driven here (the queue is saturated); please run the graph lane on a free slot before merge.

### DIFF
--- a/tests/test-graph/src/features/test_ttscgraph_dump_survives_unresolved_workspace_import.ts
+++ b/tests/test-graph/src/features/test_ttscgraph_dump_survives_unresolved_workspace_import.ts
@@ -1,0 +1,87 @@
+import { TestProject } from "@ttsc/testing";
+
+import { dumpGraph, findNode } from "../internal/graphDump";
+import { assert } from "../internal/ttsgraph";
+
+/**
+ * Verifies graph dump degrades gracefully when a workspace import does not
+ * resolve.
+ *
+ * A pnpm workspace normally links sibling packages through node_modules, but a
+ * Yarn PnP layout (no node_modules), an incomplete install, or a not-yet-linked
+ * package leaves the checker unable to resolve the import. The dump must still
+ * produce a usable graph for the files it can compile — exiting non-zero or
+ * crashing on the unresolved specifier would take down a whole monorepo's graph
+ * for one missing link. The importing package's own declarations stay in the
+ * graph; the unresolved sibling is simply absent, not an external leak.
+ *
+ * 1. Materialize an app package that imports a sibling by its package name.
+ * 2. Omit the node_modules link so the specifier cannot resolve.
+ * 3. Assert the dump still loads, keeps the app node, and drops the sibling.
+ */
+export const test_ttscgraph_dump_survives_unresolved_workspace_import = () => {
+  const root = TestProject.tmpdir("ttsc-graph-unresolved-");
+  TestProject.writeFiles(root, {
+    "package.json": JSON.stringify({
+      private: true,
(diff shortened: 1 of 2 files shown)


samchon:
Rebased onto latest `origin/master` (clean, test-only new files — no conflicts). Validation on this machine:

- `test_vscode_server_resolution_finds_package_ttsc_without_root_install`: passes through the real feature harness (`TTSC_TEST_DIR=features … --include=finds_package_ttsc_without_root_install` → Success).
- `test_ttscgraph_dump_survives_unresolved_workspace_import`: the full `test-graph` harness needs a built `@ttsc/graph` launcher + native binary, which the blocked `build:current` chain prevents locally. Validated the asserted behavior directly against the identical `ttscgraph dump` Go code path (`run(["dump", …])`) on the exact fixture: exit 0, the importing `run` node retained at `packages/app/src/main.ts`, no `packages/shared` node, no `node_modules` leak. Full harness deferred to the graph CI lane.

`pnpm format` intentionally not run locally (it triggers a repo-wide CRLF rewrite here); formatting left to CI.

PR #632, opened 2026-07-15
fix(lint): resolve unicorn/prefer-optional-catch-binding via checker scope

### LINKED ISSUE #603: fix(lint): unicorn/prefer-optional-catch-binding only knows names e/error and decides 'used' by regex-scanning raw source
## Summary

`unicorn/prefer-optional-catch-binding` only recognizes the binding names `e` and `error`, and decides whether the binding is "used" by regex-scanning the block's **raw source text** — including comments and string literals. The rule is inert on virtually all real code.

Verified independently against upstream source and by running the port.

## Root cause

`packages/lint/linthost/rules_unicorn_prefer_optional_catch_binding.go:28-31` (two-name allow-list), `:52-56` (`pattern.MatchString(nodeText(ctx.File, catch.Block))`).

Upstream (`rules/prefer-optional-catch-binding.md` + source, verified) reports **any** catch binding with no references, via real scope analysis (`context.sourceCode.getDeclaredVariables(node)` → `variable.references`). No name allow-list exists.

## Reproduction (rule at `error`, all upstream-invalid)

| source | ttsc | upstream |
|---|---|---|
| `try { f(); } catch (err) { g(); }` | **0** | invalid |
| `try { f(); } catch (exception) { g(); }` | **0** | invalid |
| `try { f(); } catch (error) { /* swallow the error */ g(); }` | **0** | invalid (comment ≠ use) |
| `try { f(); } catch (error) { g(); console.log("error"); }` | **0** | invalid (string ≠ use) |
| `try { f(); } catch (e) { g(); }` | 1 | invalid ✓ |
| `try { f(); } catch (err) { log(err); }` | 0 | valid ✓ (genuine use) |

Only the single canonical name `e` with no textual collision fires.

## Fix

Drop the allow-list and resolve references through the binder/`ctx.Checker` (or, minimally, walk the block's identifier nodes instead of its raw text).

## Regression shield

The four missed rows as invalid arms, plus the genuine-use negative twin.

samchon:
## Intent

Fixes #603. `unicorn/prefer-optional-catch-binding` was inert on real code: it only recognized the binding names `e`/`error`, and decided whether the binding was "used" by regex-scanning the catch block's **raw source text** (including comments and string literals). Any other binding name was ignored, and a comment or string that merely spelled the name suppressed the report.

## Change

- **Rule** (`rules_unicorn_prefer_optional_catch_binding.go`): drop the two-name allow-list and the raw-text scan. Resolve the binding's symbol through `ctx.Checker` and walk the block's identifier **nodes**, counting a use only when an identifier resolves to the binding's symbol. This mirrors upstream's `getDeclaredVariables(node).some(v => v.references.length > 0)`:
  - any identifier catch binding with no references reports, regardless of name;
  - a comment or string literal spelling the name is **not** a use;
  - a reference from a nested closure **is** a use;
  - a nested shadow that rebinds the name leaves the catch binding unused.
- The rule is now **type-aware** (`NeedsTypeChecker() == true`). Classification side effect: its behavioral-witness kind moves from `engine` to `checker`, and a config enabling only this rule now builds a Program/checker instead of staying on the AST-only fast path. Verified against the engine checker-gate and behavioral-witness audits.

## Scope

Identifier catch bindings only. Destructuring-pattern bindings (`catch ({ message }) {}`) keep the prior behavior of not reporting — that is a separate, pre-existing gap with different upstream messaging and is out of scope for this issue.

## Test plan

- Rewrote the Go corpus test and the `tests/test-lint` fixture (upstream-oracle) to cover the four previously-missed rows as invalid arms — bindings `err` and `exception`, a comment-only occurrence, and a string-literal occurrence — plus the canonical `e` case, a nested shadow, and the genuine-use negative twin `log(err)` (valid, no report).
- Before/after repro: against the old rule the new fixture yields 1 finding (only `e`); after the fix it yields all 6.
- `unicorn.mdx` rule section reworded to drop the `e`-only phrasing.
- Validated locally: the targeted Go test plus the engine classification and behavioral-witness suites are green. The full `node scripts/test-go-lint.cjs` and the `tests/test-lint` corpus are exercised in CI.

https://claude.ai/code/session_01BX9EHGJepvvBeNbUEDXXND

### DIFF
--- a/packages/lint/linthost/rules_unicorn_prefer_optional_catch_binding.go
+++ b/packages/lint/linthost/rules_unicorn_prefer_optional_catch_binding.go
@@ -5,39 +5,31 @@
 // and signals that the catch body still cares about the error when in
 // fact it doesn't.
 //
-// AST-only: visit each `CatchClause`, restrict to the common-case
-// binding names `e` and `error` (true scope analysis would require a
-// resolver), and confirm the body's source text does not contain the
-// binding name as a word-bounded identifier. The previous substring
-// scan was load-bearing-wrong: `strings.Contains(body, "e")` matches
-// nearly any English identifier, hiding the rule entirely; word
-// boundaries (`\bname\b`) correctly distinguish `e` from `error`,
-// `error` from `errorMessage`, etc. False positives on `name` inside
-// string literals or comments are acceptable because the rule is
-// conservative by design.
+// Report any identifier catch binding whose declared variable is never
+// referenced, matching upstream's `getDeclaredVariables(node).some(v =>
+// v.references.length > 0)` check. Binding identity comes from the
+// TypeScript checker: the block is walked for identifiers that resolve to
+// the binding's symbol, so a comment or string literal that merely spells
+// the name is not a use, a reference from a nested closure still is, and a
+// nested shadow that rebinds the name leaves the catch binding unused.
 // https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-optional-catch-binding.md
 package linthost
 
-import (
-  "regexp"
-
-  shimast "github.com/microsoft/typescript-go/shim/ast"
-)
-
-var unicornPreferOptionalCatchBindingNamePattern = map[string]*regexp.Regexp{
-  "e":     regexp.MustCompile(`\be\b`),
-  "error": regexp.MustCompile(`\berror\b`),
-}
+import shimast "github.com/microsoft/typescript-go/shim/ast"
 
 type unicornPreferOptionalCatchBinding struct{}
 
 func (unicornPreferOptionalCatchBinding) Name() string {
   return "unicorn/prefer-optional-catch-binding"
 }
+func (unicornPreferOptionalCatchBinding) NeedsTypeChecker() bool { return true }
 func (unicornPreferOptionalCatchBinding) Visits() []shimast.Kind {
   return []shimast.Kind{shimast.KindCatchClause}
 }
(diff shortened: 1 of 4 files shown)


samchon:
### CI triage — the red checks are pre-existing, not caused by this change

This change is exactly four files (the rule, its Go test, its `.ts` corpus fixture, and the `unicorn.mdx` section); it touches no LSP, corpus-harness, or `.tsx` fixture.

- **`go` / `windows-go`** — both fail on the same three `TestLSPExecuteCommandMaterializesSymlink*` tests (`ttsc.lint.fixAll returned no edit`). Those use `no-var`, `Skipf` on non-symlink platforms, and are untouched here. The full local `go test ./linthost` (Windows) is green including the behavioral-witness audit.
- **`lint native corpus` (partition 1) / `lint native corpus-3`** — fail on `.tsx` fixtures materialized as `src/main.ts` → JSX parse errors: partition 1 throws on `nextjs-google-font-preconnect.tsx`, partition 3 on `jsx-a11y-anchor-ambiguous-text.tsx`. Both are byte-identical on `master` and sort before this PR's `.ts` fixture, so each partition aborts before reaching it. `TestLint`/`assertLintCase` carry no JSX handling, so this is systemic and pre-existing.
- **`nestia`** — the standing expected-failure exemption.

This change's fixture (`unicorn-prefer-optional-catch-binding.ts`) lives in partition 1, which never reaches it. The equivalent end-to-end validation runs green through the Go corpus test `TestRuleCorpusUnicornPreferOptionalCatchBinding`, which drives the identical rule engine over a real Program + checker (six invalid catch bindings reported, the genuine-use twin clean, zero program diagnostics).

PR #644, opened 2026-07-15
fix(lint): honor no-else-return allowElseIf, restrict terminator to return

### LINKED ISSUE #598: fix(lint): no-else-return ignores allowElseIf (default true) and treats throw/break/continue as returns
## Summary

`no-else-return` ignores the `allowElseIf` option (upstream default `true`) and treats `throw`/`break`/`continue` as returns. Both diverge from ESLint; the first flags `return` + `else if` chains, one of the most common TypeScript control-flow shapes.

Verified independently against upstream source and by running the port.

## Root cause

`packages/lint/linthost/rules_no_else_return.go:32-41` reports on any `ElseStatement`; `noElseReturnTerminates:43-68` accepts `KindThrowStatement`, `KindBreakStatement`, `KindContinueStatement`.

Upstream (`eslint/lib/rules/no-else-return.js`, verified from source): `defaultOptions: [{allowElseIf: true}]`; with `allowElseIf` true it uses `checkIfWithoutElse`, which walks the `if/else if` chain and **bails** (`if (!currentNode.alternate) return;`) when the chain ends without a plain `else`. `checkForReturn` is exactly `node.type === "ReturnStatement"` — `throw`/`break`/`continue` do **not** count. Upstream also reports **once** on the chain's terminal `else`.

## Reproduction (default options, rule at `error`)

| source | ttsc | upstream |
|---|---|---|
| `if (a) { return 1; } else if (b) { return 2; }` | **1 finding** | valid (`allowElseIf:true`) |
| `if (a) { throw new Error("x"); } else { g(); }` | **1 finding** | valid (throw ≠ return) |
| `for(;;){ if (a) { break; } else { g(); } }` | **1 finding** | valid |
| `for(;;){ if (a) { continue; } else { g(); } }` | **1 finding** | valid |
| `if (a) { return 1; } else { return 2; }` | 1 finding | invalid ✓ |

Secondary: `if(a) return 1; else if(b) return 2; else return 3;` yields 2 findings vs upstream's 1.

## Fix

Implement `allowElseIf` (default `true`) with upstream's chain walk; restrict the terminator test to `ReturnStatement`; report once on the terminal `else`.

## Regression shield

The five rows plus the 3-branch chain (assert exactly one finding, on the final `else`).

samchon:
## Summary

Fixes #598. `no-else-return` ignored the `allowElseIf` option (upstream default
`true`) and treated `throw` / `break` / `continue` as returns, so it flagged the
common `return` + `else if` chain and other non-return terminators, and reported
twice on a three-branch `return` chain.

This ports the ESLint rule (`eslint/lib/rules/no-else-return.js`) faithfully:

- **Only `return` terminates a branch** (`checkForReturn`). `throw` / `break` /
  `continue` no longer count.
- **`allowElseIf` (default `true`)** walks the `if / else if` chain and bails
  when it ends without a plain `else`, so a `return` + `else if` early-return
  chain stays valid. The single report lands on the chain's **terminal `else`**.
- **`allowElseIf: false`** flags the `else if` via the `checkIfWithElse` path.
- Analysis starts only at a statement-list-level `if` (upstream's
  `STATEMENT_LIST_PARENTS` gate → `SourceFile` / `Block` /
  `ClassStaticBlockDeclaration` / `CaseClause` / `DefaultClause`), which is what
  makes a chain report exactly once.

## Before / after (default options, rule at `error`)

| source | before | after (= upstream) |
|---|---|---|
| `if (a) { return 1; } else if (b) { return 2; }` | 1 finding | valid |
| `if (a) { throw ...; } else { g(); }` | 1 finding | valid |
| `for(;;){ if (a) { break; } else { g(); } }` | 1 finding | valid |
| `for(;;){ if (a) { continue; } else { g(); } }` | 1 finding | valid |
| `if (a) { return 1; } else { return 2; }` | 1 finding | 1 finding ✓ |
| `if(a) return 1; else if(b) return 2; else return 3;` | 2 findings | 1 finding (terminal else) |

The five false-positive / double-report rows fail against the pre-fix rule and
pass after (verified with a targeted `go test` before/after run).

## Tests

- Six focused per-case Go tests (else-if / throw / break / continue valid; plain
  else-return invalid; three-branch chain reports exactly once on the terminal
  else), plus an `allowElseIf: false` positive covering the `checkIfWithElse`
  branch.
- The `tests/test-lint/src/cases/no-else-return.ts` corpus fixture and its Go
  mirror now pin the whole matrix; the website `core.mdx` section documents the
  corrected behavior and the `allowElseIf` option.
- Full `packages/lint` Go suite green locally (`go test ./linthost`, exit 0),
  including the end-of-run behavioral-witness aggregate.

## Scope note

The `allowElseIf` option is honored at runtime (decoded from the rule's options
blob). Surfacing it in the typed config autocomplete
(`packages/lint/src/structures/rules/ITtscLintCoreRules.ts`) is out of this
change's scope; that entry still types `no-else-return` as option-less. Flagging
as a possible follow-up.

https://claude.ai/code/session_01Giu15S7hyRxpupc1XwXe89


samchon:
Superseded by #645 (same fix for #598, cherry-picked clean onto current master). Closing as redundant.

Issue #651, opened 2026-07-15
fix(lint): reject an options payload for a genuinely optionless rule (needs an 'accepts-options' marker on every option-consuming rule)


samchon:
Follow-up to #626 (partially fixed by PR #648, which added the typed `targets` key for `unicorn/no-unnecessary-polyfills`). The engine-side half remains:

**Remaining problem.** The engine silently accepts and ignores an options payload for a rule that has no options schema (`engine.go ConfigError`), so a user's `["error", { …typo… }]` is inert with no error.

**Why it's blocked (evidence from PR #648).** A correct rejection needs a complete runtime "accepts options" signal on every option-consuming rule. Today only ~16 rules implement `ValidateOptions`; ~22 more consume real user options through `ctx.DecodeOptions` with NO runtime marker (`no-else-return` `allowElseIf`, `default-case` `commentPattern`, `prefer-const`, `no-typeof-undefined`, `prefer-number-properties`, `ban-ts-comment`, …). Rejecting off the incomplete `ValidateOptions` signal would regress those valid configs; a hardcoded rule list is disallowed.

**Fix direction.** Add a uniform `AcceptsOptions()`/schema marker to every option-consuming rule (or derive it from `DecodeOptions` usage), then reject payloads for rules without it. Also add the broad options-parity test (every option-consuming rule owns an options-typed TS key) — currently red for `no-typeof-undefined`, `prefer-number-properties`, `default-case`, `no-else-return`, `no-extend-native` (severity-only-keyed despite honoring options).

PR #654, opened 2026-07-15
fix(lint): reject options for optionless rules; type unicorn/no-unnecessary-polyfills targets

### LINKED ISSUE #626: fix(lint): rule options silently accepted-and-ignored for optionless rules; unicorn/no-unnecessary-polyfills targets option forbidden by published TS types
## Summary

Two related option-handling defects: (1) the engine **silently accepts and ignores** an options payload for a rule that has no options schema, so a user's config is inert; and (2) `unicorn/no-unnecessary-polyfills` fully supports a runtime `targets` option that the published TS types **forbid**.

Verified by code + running the port.

## Details

**1. Options silently ignored (`engine.go:596-603`).** `ConfigError` never rejects an options payload for an optionless rule, so `["error", { … }]` is accepted and dropped. This is option-dependent false-positive AND false-negative:

| config | ttsc | ESLint |
|---|---|---|
| `["error",{"allowLoop":true}]` on `no-labels` + labeled loop | fires | 0 (FP under config) |
| `["error","always"]` on `no-cond-assign` + parenthesized assign | 0 | fires (FN under config) |

Fix: reject an options payload for a rule with no options schema ("rule X does not accept options").

**2. `no-unnecessary-polyfills` option type mismatch.** The Go rule accepts and validates `["error", { targets: … }]` (`rules_unicorn_no_unnecessary_polyfills.go:55-62`, consumed at `:102`), mirroring upstream. But the published typed key is severity-only: `ITtscLintUnicornRules.ts:493` types it `TtscLintRuleSetting`, and it is **absent** from `ITtscLintRuleOptionsMap.ts`. So a type-checked `lint.config.ts` passing `{ targets: "node 18" }` is a TS error even though the runtime needs and honors it. It is the sole miss among the 34 option-consuming Go rules (33 have a typed tuple).

Fix: add `ITtscLintUnicornNoUnnecessaryPolyfillsRuleOptions` (with `targets`), wire it into `ITtscLintRuleOptionsMap.ts`, and change the key to `TtscLintRuleOptionsSetting<…>`.

## Regression shield

An options-parity test (mirror of the name-parity `typed_keys_match_registered_rules_test.go`): every Go rule with `ValidateOptions`/`DecodeOptions` must have a `TtscLintRuleOptionsSetting<…>` typed key; and the engine rejects an options payload for an optionless rule.

samchon:
Fixes #626.

## Intent

Two related option-handling defects in `@ttsc/lint`:

1. **Options silently ignored for optionless rules.** Engine construction accepted and dropped an options payload aimed at a rule with no options schema, so a config typo / unsupported option left the config inert with no diagnostic — option-dependent false positive AND false negative (`["error", {allowLoop: true}]` on `no-labels` still fired; `["error", "always"]` on `no-cond-assign` stayed silent).
2. **`unicorn/no-unnecessary-polyfills` `targets` forbidden by the published TS types.** The Go rule validates and honors a `targets` option, but the typed key was severity-only and absent from `ITtscLintRuleOptionsMap`, so a type-checked config passing `{ targets: "node 18" }` was a TS error the runtime nonetheless honors.

Verified against current `master` (both claims reproduce).

## Scope

**Defect 1 — general engine gate.** `ruleAcceptsOptions` classifies a rule as option-bearing via `ValidateOptions`, the format category, or a new `ConsumesOptions` marker; the binding loop rejects a real payload (`hasOptionsPayload`: non-nil, non-`null`) for any rule matching none, emitting `rule "X" does not accept options` and keeping it out of dispatch. This seals every optionless rule (~400), not just the two examples.
- Every built-in rule that reads `ctx.Options` / `ctx.DecodeOptions` without a validator carries the marker, declared in one auditable file (`rules_option_consumers.go`).
- Contributor rules keep accepting options unconditionally — the public `rule.Context` has no optionless marker, so the host cannot infer one (same conservative default as `NeedsTypeChecker`).

**Defect 2 — polyfills typing (this issue's named rule only).** New `ITtscLintUnicornNoUnnecessaryPolyfillsRuleOptions` (`targets`: query string | array of queries | targets object; optional, matching the decoder's config-discovery fallback), wired into `ITtscLintRuleOptionsMap`, key switched to `TtscLintRuleOptionsSetting<…>`. `TtscLintRuleOptionsSetting` is a superset of `TtscLintRuleSetting`, so no existing config breaks.

## Deferred / notes

- **Broader defect-2 class.** `no-unnecessary-polyfills` is *not* the sole option-consuming rule missing a typed key on current master; the same mismatch exists for `unicorn/no-typeof-undefined`, `unicorn/prefer-number-properties`, `default-case`, `no-else-return`, `no-extend-native`, `no-inner-declarations`, `no-return-assign`, and `unicorn/text-encoding-identifier-case`. Per the issue's stated scope this PR fixes only polyfills; the marker preserves the others' runtime behavior so nothing regresses. A full options-parity test (typed key ⇔ Go option support) is deferred until those are individually aligned, since a strict cross-check would red-CI on rules other branches own.
- **Overlap with `fix/lint-checker-pool-and-options-v2`.** That branch reworks `unicorn/prefer-number-properties` / `no-mixed-operators` options and does **not** touch the engine, the marker file, or `contrib_adapter.go`, so there is no overlap with this change. (`prefer-number-properties` uses `ValidateOptions`, so it needs no marker here.)

## Test plan

- `packages/lint/test/engine/`: an optionless rule's payload (object, scalar, array, `{}`) is rejected and dropped from dispatch; the negative twins — bare severity / empty / explicit `null` stays enabled, and a marker rule (`default-case`) and a validator rule (`unicorn/string-content`) keep their payloads.
- `test_lib_index_d_ts_rule_options_autocomplete_per_rule.ts`: the three valid `targets` forms plus omission compile; a typo'd key and a non-query value are `@ts-expect-error`.
- Full `packages/lint` Go suite green locally (engine, config, rules, plugin/contributor, format). TS type contract validated with a standalone `tsc --strict` probe over the structures.

### DIFF
--- a/packages/lint/linthost/contrib_adapter.go
+++ b/packages/lint/linthost/contrib_adapter.go
@@ -164,6 +164,17 @@ func (a contributorAdapter) VisitsDeclarationFiles() bool {
   return a.visitsDeclarationFiles
 }
 
+// ConsumesOptions keeps every contributor rule eligible for a configuration
+// options payload. The public `rule.Context` exposes `Options` /
+// `DecodeOptions` with no schema or optionless marker, so the host cannot infer
+// that a third-party rule rejects options. Treating them all as option-bearing
+// preserves the public contract — rejecting a payload a contributor honors
+// would break its config — at the cost of not surfacing an unsupported option
+// for contributor rules, the same conservative default as NeedsTypeChecker.
+func (a contributorAdapter) ConsumesOptions() bool {
+  return true
+}
+
 // formatContributorAdapter is the FormatRule-tagged variant of
 // contributorAdapter. Wrapping the lint-only adapter (rather than
 // duplicating its method set) keeps the marker addition trivial and
--- a/packages/lint/linthost/engine.go
+++ b/packages/lint/linthost/engine.go
@@ -18,6 +18,7 @@
 package linthost
 
 import (
+  "bytes"
   "encoding/json"
   "errors"
   "fmt"
@@ -81,6 +82,17 @@ type ruleOptionsValidator interface {
   ValidateOptions(json.RawMessage) error
 }
 
+// ruleOptionsConsumer marks a rule that reads a configuration options payload
+// but declares no ValidateOptions schema (it decodes leniently). It lets the
+// engine tell a genuinely optionless rule — where a supplied payload is a
+// configuration error rather than an ignored typo — from an options-bearing
+// rule that merely skips strict validation. ValidateOptions implementers and
+// format rules already advertise option support through their own interfaces
+// and need not implement this.
+type ruleOptionsConsumer interface {
+  ConsumesOptions() bool
+}
+
 // isFormatRule reports whether `r` opts into the format category.
 func isFormatRule(r Rule) bool {
   fr, ok := r.(FormatRule)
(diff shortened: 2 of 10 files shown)


samchon:
Superseded / conflicting. The `unicorn/no-unnecessary-polyfills` typed-options half of this PR already shipped in #648 (merged), which causes the conflict in `ITtscLintUnicornRuleOptions.ts`. The remaining engine-side "reject options for an optionless rule" work is tracked in the open follow-up #651. Closing this stale branch rather than resolving the conflict.

PR #690, opened 2026-07-15
fix(metro): preserve upstream transformer initialization failures

### LINKED ISSUE #685: fix(metro): preserve upstream transformer initialization failures
## Problem

`@ttsc/metro` loads an explicit or auto-detected upstream Babel transformer with `require`. Its production loader treats every exception as if the requested module did not exist. If an installed transformer throws during top-level initialization, has a missing peer or transitive dependency, or rejects the current runtime ABI, the adapter discards the original failure.

For an explicit path, the user receives only `Could not load the configured upstream transformer`. During auto-detection, the adapter may continue to a later candidate and select the wrong transformer, hiding a broken installed Expo/React Native configuration.

## Evidence

The owner is `tryRequire` in `packages/metro/src/core/upstream.ts`:

```ts
function tryRequire(modulePath: string): UpstreamTransformer | undefined {
  try {
    return nodeRequire(modulePath) as UpstreamTransformer;
  } catch {
    return undefined;
  }
}
```

`resolveUpstreamTransformer` interprets `undefined` as absence. For a configured path it throws a replacement error; for automatic discovery it continues through `UPSTREAM_CANDIDATES` in Expo, modern React Native, then legacy order.

The public resolver reproduces the information loss with an installed local module:

```js
// broken-upstream.cjs
throw new Error("upstream dependency ABI mismatch");
```

```ts
import { createRequire } from "node:module";
import { resolveUpstreamTransformer } from "./packages/metro/src/core/upstream";

const require = createRequire(import.meta.url);
resolveUpstreamTransformer(require.resolve("./broken-upstream.cjs"));
```

The current exception is `[@ttsc/metro] Could not load the configured upstream transformer: ...`; the original `upstream dependency ABI mismatch` message and stack are absent. The same catch also conflates `MODULE_NOT_FOUND` for the candidate itself with `MODULE_NOT_FOUND` thrown for one of that candidate's dependencies.

## Consequence surface

The transform path cannot distinguish “optional upstream is not installed” from “installed upstream is broken.” Users are directed to install or configure a package they already have, and auto-detection can run a fallback that does not match the active Expo/RN stack. This obscures version and peer-dependency faults at the point where their original stack is most useful.

True absence must remain non-fatal during automatic probing so optional peers continue to work. The transformer's separate defensive `getCacheKey` behavior is not automatically in scope; this issue owns module initialization and candidate selection, and broader cache-key error policy should change only with its own evidence.

## Approach

Keep `tryRequire` / `resolveUpstreamTransformer` as the ownership boundary and enforce a three-way invariant: a genuinely absent candidate may be skipped, a loaded candidate is returned, and an exception raised while initializing an existing candidate is preserved with its cause and stack.

Differentiate absence only when the module-resolution error proves that the requested candidate itself was not found. Do not infer absence merely from an error code that may belong to a transitive import. The precise wrapping format can be chosen with tests, but it must retain the original diagnostic.

- Automatic detection continues past a candidate that is genuinely not installed.
- An explicit configured module that throws during initialization fails with the original message/cause and useful stack context.
- An auto-detected candidate that exists but throws does not fall through to a later candidate.
- A candidate whose peer or transitive dependency is missing reports that dependency failure rather than claiming the candidate is absent.
- The existing “install one of these upstreams” guidance remains when all candidates are genuinely absent.
- Add coverage through the production loading path, not only the injectable `load` seam.
- Put the production-loader regressions under `tests/test-metro/src/features/upstream` and run them with `pnpm --filter @ttsc/test-metro start -- --include=upstream`.
- Then run the canonical repository checks `pnpm build` and `pnpm test`.

## Coordination

Closed issue #255 and merged PR #256 introduced `@ttsc/metro`, the optional upstream candidate order, and the current loader. Their scope did not separately track initialization-error preservation.

Open PR #657, `fix(test): audit every lint corpus source`, is confined to lint-corpus behavior and has no Metro files.

[Issue #672](https://github.com/samchon/ttsc/issues/672), **Evict failed unplugin transforms so long-lived adapters can recover**, also affects Metro workers, but it owns cached transform results after the upstream has loaded. This issue ends at upstream module selection and initialization.

samchon:
## Bundle b8 — Metro upstream transformer errors

This bundle hardens `@ttsc/metro`'s upstream Babel-transformer loader so a broken-but-installed transformer is no longer masked as absence.

### Issues

- Closes #685 (RA-15) — `fix(metro): preserve upstream transformer initialization failures`

### Problem

`resolveUpstreamTransformer` treated every exception from `require`ing an upstream candidate as "module not present." An installed transformer that throws during top-level initialization, has a missing peer/transitive dependency, or rejects the current runtime ABI was silently discarded:

- an **explicit** `upstreamTransformer` failure surfaced only the generic `Could not load the configured upstream transformer`, dropping the original message and stack;
- **auto-detection** could skip the broken candidate and select a later one that does not match the active Expo/RN stack.

### Fix

`tryRequire` now splits resolution from execution. `require.resolve` never runs third-party code, so a `MODULE_NOT_FOUND` / `ERR_MODULE_NOT_FOUND` there proves the requested candidate itself is genuinely absent — reported as `undefined` so automatic probing continues to the next optional peer. Any error thrown while executing the (resolvable) module body is a real initialization failure and is preserved with its original message and stack as the `cause` of a `@ttsc/metro` wrapper. A candidate's own missing transitive dependency surfaces that dependency, not a false "absent candidate."

### Tests

Production-loader regressions under `tests/test-metro/src/features/upstream`, run with `pnpm --filter @ttsc/test-metro start -- --include=upstream`:

- genuinely-absent explicit path stays the plain absence message, no `cause`;
- explicit init-throw preserves the original diagnostic + `cause` + stack;
- missing transitive dependency is named, not misreported as absence;
- auto-detect does not fall through to a later candidate when an earlier one is broken.

Docs: `packages/metro/README.md` documents the absence-vs-broken distinction.

### DIFF
--- a/packages/metro/README.md
+++ b/packages/metro/README.md
@@ -41,7 +41,7 @@ module.exports = withTtsc(getDefaultConfig(__dirname));
 
 `withTtsc` sets `transformer.babelTransformerPath` and leaves the rest of your config untouched. It auto-detects the upstream transformer to delegate to (`@expo/metro-config/babel-transformer` for Expo, then `@react-native/metro-babel-transformer`, then the legacy `metro-react-native-babel-transformer`).
 
-Auto-detection only skips a candidate that is genuinely **not installed**. A candidate that _is_ installed but fails while loading — a top-level throw, an incompatible runtime ABI, or a missing peer/transitive dependency — surfaces its original error (as the `cause` of a `@ttsc/metro` wrapper) instead of being treated as absent. This stops a broken Expo/React Native install from silently falling through to the wrong transformer, and stops an explicit `upstreamTransformer` failure from being reported as if the module did not exist.
+Auto-detection only skips a candidate whose entry point is genuinely **not available** — the package is not installed, or it is installed but the requested subpath is not exported (Expo/React Native version skew). A candidate that _does_ resolve but fails while loading — a top-level throw, an incompatible runtime ABI, or a missing peer/transitive dependency — surfaces its original error (as the `cause` of a `@ttsc/metro` wrapper) instead of being treated as absent. This stops a broken Expo/React Native install from silently falling through to the wrong transformer, and stops an explicit `upstreamTransformer` failure from being reported as if the module did not exist.
 
 ## Configuration
 
--- a/packages/metro/README.md
+++ b/packages/metro/README.md
@@ -41,6 +41,8 @@ module.exports = withTtsc(getDefaultConfig(__dirname));
 
 `withTtsc` sets `transformer.babelTransformerPath` and leaves the rest of your config untouched. It auto-detects the upstream transformer to delegate to (`@expo/metro-config/babel-transformer` for Expo, then `@react-native/metro-babel-transformer`, then the legacy `metro-react-native-babel-transformer`).
 
+Auto-detection only skips a candidate that is genuinely **not installed**. A candidate that _is_ installed but fails while loading — a top-level throw, an incompatible runtime ABI, or a missing peer/transitive dependency — surfaces its original error (as the `cause` of a `@ttsc/metro` wrapper) instead of being treated as absent. This stops a broken Expo/React Native install from silently falling through to the wrong transformer, and stops an explicit `upstreamTransformer` failure from being reported as if the module did not exist.
+
 ## Configuration
 
 By default `@ttsc/metro` finds the nearest `tsconfig.json` from the file being transformed and runs the plugins configured there: the standard `ttsc` model. If that is the config you want, `withTtsc(getDefaultConfig(__dirname))` is enough.
--- a/packages/metro/src/core/upstream.ts
+++ b/packages/metro/src/core/upstream.ts
@@ -118,27 +118,44 @@ function tryRequire(modulePath: string): UpstreamTransformer | undefined {
   try {
     nodeRequire.resolve(modulePath);
   } catch (error) {
-    if (isModuleNotFound(error)) {
+    if (isCandidateAbsent(error)) {
       return undefined;
     }
-    // A non-"not found" resolution error (e.g. an invalid specifier) is not
-    // evidence of a plain absence; surface it.
+    // A resolution error that is not one of the known "entry point absent"
+    // codes (e.g. an invalid specifier) is not evidence of a plain absence;
+    // surface it rather than silently skipping the candidate.
     throw error;
   }
   return nodeRequire(modulePath) as UpstreamTransformer;
 }
 
 /**
- * Whether a resolution error means the requested module could not be found.
+ * Whether a resolution error means the requested candidate's entry point is not
+ * available here — i.e. genuine absence, not a broken initialization.
  *
- * A CJS `require.resolve` failure carries `code === "MODULE_NOT_FOUND"`; the
- * ESM loader uses `ERR_MODULE_NOT_FOUND`. Because resolution never executes the
- * module, this error can only refer to the requested specifier, never a
- * transitive import of it.
+ * Resolution never executes the module body, so a `require.resolve` failure can
+ * only concern the requested specifier, never a transitive import of it. Each
+ * recognised code says the same thing about that specifier:
+ *
+ * - `MODULE_NOT_FOUND` / `ERR_MODULE_NOT_FOUND` — the package or file itself is
+ *   not installed (CJS and ESM loaders respectively).
+ * - `ERR_PACKAGE_PATH_NOT_EXPORTED` — the package is installed but the requested
+ *   subpath is not exported (or its export target is missing). This matters for
+ *   the `@expo/metro-config/babel-transformer` candidate, a package subpath:
+ *   under Expo/React Native version skew a present but non-exporting package
+ *   must stay non-fatal so auto-detection falls through to the next candidate,
+ *   exactly as a wholly absent package does.
+ *
+ * An error thrown later, while the resolved module executes, is a real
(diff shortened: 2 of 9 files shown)


samchon:
## Self-Review round 1

Verified the committed implementation against the RA-15 acceptance criteria and traced the consequence surface of splitting `require.resolve` from `require`.

**Finding (correctness regression introduced by the split).** `isModuleNotFound` classifies only `MODULE_NOT_FOUND`/`ERR_MODULE_NOT_FOUND` as absence. But the first candidate, `@expo/metro-config/babel-transformer`, is a package *subpath*. When `@expo/metro-config` is installed at a version whose `babel-transformer` subpath is not exported (version skew), Node's `require.resolve` throws `ERR_PACKAGE_PATH_NOT_EXPORTED`, not `MODULE_NOT_FOUND`. The committed code rethrows that and auto-detection **hard-fails** with 'installed but failed to initialize' instead of falling through to the React Native candidate.

This is a regression versus the old swallow-everything loader and violates the issue's criterion that *genuine absence stays non-fatal during automatic probing*. A resolution-time error can only concern the requested specifier (resolve never executes the module body), and `ERR_PACKAGE_PATH_NOT_EXPORTED` means the requested candidate entry point is not available here — that is absence, not a broken initialization. Empirically confirmed: absent package -> `MODULE_NOT_FOUND`; installed package, missing subpath export -> `ERR_PACKAGE_PATH_NOT_EXPORTED`.

**Remediation plan.**
1. Broaden the absence classifier (rename `isModuleNotFound` -> `isCandidateAbsent`) to also treat `ERR_PACKAGE_PATH_NOT_EXPORTED` as absence, with an updated doc explaining that every resolution error concerns the requested specifier and these codes all mean 'this entry point is not available here'. Execution-time failures (top-level throw, missing transitive dependency) are untouched — they still surface with cause.
2. Add a production-path regression under `tests/test-metro/src/features/upstream` that drives the real loader with an installed package whose requested subpath is not exported and asserts it is reported as absence (no `cause`), the boundary twin of the init-failure cases.
3. Reconcile the README wording so 'genuinely not installed' includes a missing/non-exported entry point.

Applying now.

samchon:
## Self-Review round 2 — clean

Re-inspected the full changed surface after the round-1 fix; no sound improvement or new issue found.

- **Classifier composition.** `isCandidateAbsent` now returns `undefined` for `MODULE_NOT_FOUND` / `ERR_MODULE_NOT_FOUND` / `ERR_PACKAGE_PATH_NOT_EXPORTED`, which composes cleanly with the existing auto-detect loop (absence -> continue) and the explicit-path branch (absence -> 'could not load'). Execution-time failures (top-level throw, missing transitive dependency, non-absence resolution errors) still surface with `cause` — the resolve/execute split is intact.
- **Explicit-path semantics.** A configured subpath whose export target is missing now reports the plain 'could not load' guidance rather than a wrapped init failure; that is correct — it is a resolution-time absence of the requested entry, and it never masks an execution-time transitive failure (those still surface).
- **Coverage.** Acceptance criteria 1-7 are each pinned: genuinely-absent explicit path, non-exported subpath (new), explicit init-throw with cause+stack, missing transitive dependency, auto-detect no-fall-through, priority-order fall-through, and terminal install guidance. All but the injected auto-detect-init case run through the production `require` loader.
- **Hygiene.** No stale `isModuleNotFound` references; `tsc --noEmit -p tests/test-metro/tsconfig.json` and `pnpm --filter @ttsc/metro build` (tsc emit) both pass; README reconciled to 'not available (not installed or subpath not exported)'.

Self-review complete at 2 rounds (1 finding remediated, 1 clean pass).

samchon:
## Final verification — green

RA-15 (#685) implemented and self-reviewed (2 rounds). Local, package-scoped verification on HEAD `e4d77b4a8`:

```
pnpm --filter @ttsc/metro build          # tsc emit + rollup, OK
pnpm --filter @ttsc/test-metro start -- --include=upstream   # 15/15 Success
npx tsc --noEmit -p tests/test-metro/tsconfig.json           # exit 0
```

The `upstream` suite covers every acceptance criterion, all but the injected auto-detect-init case running through the production `require` loader:

- `test_upstream_genuinely_absent_candidate_stays_non_fatal` — absent explicit path -> 'could not load', no cause
- `test_upstream_unexported_subpath_stays_non_fatal` — installed package, non-exported subpath (`ERR_PACKAGE_PATH_NOT_EXPORTED`) -> absence, keeps auto-detect fall-through under version skew
- `test_upstream_configured_init_failure_preserves_cause` — explicit init-throw preserves message + cause + stack
- `test_upstream_missing_transitive_dependency_reported` — missing transitive dep named, not misreported as absent
- `test_upstream_auto_detect_init_failure_does_not_fall_through` — broken earlier candidate surfaces, no fall-through to legacy
- plus the pre-existing priority-order, empty-path, and no-upstream-installed cases

Per the campaign HARD RULES I did not run whole-repo `pnpm test` or `pnpm format`; the canonical `pnpm build`/`pnpm test` belong to the lead's recheck. Marking ready for review. Not merging.

samchon:
## Squash-merged

Bundle b8 (RA-15, #685) squash-merged into `master` as a single `fix(metro)` commit; linked issues auto-close from the PR body.

- Merge commit: `fd33764b2446e0a632471662de2088fb23781b42`
- Merge-commit CI cancelled by exact SHA per campaign discipline (no wasted runners on the suspended-CI campaign).

PR #691, opened 2026-07-15
fix(paths): preserve JSON extensions when aliases are rewritten to emitted files

### LINKED ISSUE #680: fix(paths): preserve JSON extensions when aliases are rewritten to emitted files
## Problem

`@ttsc/paths` resolves an exact `compilerOptions.paths` alias to a JSON source when `resolveJsonModule` includes that file in the Program. Its output-path predictor treats every unrecognized source extension as JavaScript, so it rewrites the alias to `.js` even though TypeScript-Go copies the file as `.json`.

The build exits successfully and produces JavaScript that points at a nonexistent file. The compiler/plugin pipeline reports success; the failure is deferred until runtime.

## Evidence

The stable owners are:

- `packages/paths/driver/paths.go::lookupSource`, which accepts an exact Program source path regardless of extension.
- `packages/paths/driver/paths.go::outputPathForSource`.
- `packages/paths/driver/paths.go::emittedJavaScriptExtension`, whose default branch returns `".js"`.
- `packages/paths/driver/paths.go::replaceSourceExtension`, which strips the original extension before appending that predicted suffix.

A real fixture uses:

```jsonc
{
  "compilerOptions": {
    "module": "commonjs",
    "target": "ES2022",
    "resolveJsonModule": true,
    "esModuleInterop": true,
    "rootDir": "src",
    "outDir": "dist",
    "paths": { "@data": ["./src/data.json"] },
    "plugins": [{ "transform": "@ttsc/paths" }]
  },
  "include": ["src"]
}
```

`src/main.ts` imports `@data` and `src/data.json` contains valid JSON. Run:

```sh
npx ttsc --emit
node dist/main.js
```

The compile exits 0. The generated artifact set contains `dist/data.json` and no `dist/data.js`, but `dist/main.js` contains `require("./data.js")`. Node exits with `MODULE_NOT_FOUND`. TypeScript-Go copied the JSON asset under its correct name; only the paths output predictor invented the missing `.js` target.

This is repository-owned output prediction, not a Node or TypeScript-Go resolver failure. The plugin finds the correct Program source, then replaces its extension incorrectly.

## Consequence surface

- CommonJS JSON imports fail at runtime after a successful build.
- ESM JSON imports/attributes and declaration output can receive the same wrong predicted target when their specifier is rewritten.
- Exact `.json` path targets are affected; extensionless target lookup currently does not include JSON and should be evaluated separately rather than accidentally widened.
- TypeScript/JavaScript families must keep their existing `.js`, `.mjs`, `.cjs`, and JSX-preserve mappings.
- Other non-code Program sources must not be blindly classified as JavaScript; the output predictor should reflect what the compiler emits or copies.
- Projects without `outDir` remain outside the plugin's documented rewrite boundary and are not part of this issue.

## Approach

`outputPathForSource` owns the mapping from a resolved Program source to the artifact that consumers load. The invariant is that a rewritten specifier names the file TypeScript-Go actually emits or copies for that source under the active compiler options. Preserve copied asset extensions such as `.json` and retain the established JavaScript-family mappings.

Do not special-case a fixture path or only suppress the runtime error. Define the source-to-output classification at the output-path owner and cover the full supported extension matrix.

- A `resolveJsonModule` fixture with an exact JSON alias builds and runs successfully.
- The emitted JavaScript references `./data.json` and the artifact exists at that exact path.
- ESM and CommonJS JSON cases use the correct extension and retain required import-attribute syntax from the compiler.
- Declaration emit, when enabled, does not introduce a nonexistent `.js` JSON target.
- Existing `.ts`/`.tsx`/`.mts`/`.cts`/`.js`/`.jsx`/`.mjs`/`.cjs` mappings and JSX-preserve tests remain unchanged.
- Negative coverage proves unsupported/non-emitted assets are not rewritten to invented files.
- Add the source-to-output extension matrix beside `packages/paths/test/rewriter_helpers_cover_resolution_edges_test.go` and the real build/run fixture under `tests/test-paths/src/features/`.
- Run `go test ./...` from `packages/paths`, `pnpm --filter @ttsc/test-paths start`, and the broader `pnpm test`.

## Coordination

The output-path predictor originated in PR #91. PR #314 hardened root inference and path-pattern precedence but did not cover JSON emit. This issue is related to campaign issue #686, **Resolve npm tsconfig presets through package.json#tsconfig**, because both affect path-aware compilation, but preset discovery and emitted asset suffix prediction have separate owners.

The only current open pull request, #657 (**fix(test): audit every lint corpus source**), is confined to lint corpus and harness files and does not touch `packages/paths` or its feature tests.

samchon:
## Bundle b10 — `@ttsc/paths` JSON extension preservation

Part of the `repository-audit-2026-07` issue campaign.

### Scope

`@ttsc/paths` predicts the emitted output path for a resolved Program source so it can rewrite a `compilerOptions.paths` alias to the file the compiler actually produces. Its predictor treated **every** unrecognized source extension as JavaScript. When `resolveJsonModule` pulls a `.json` source into the Program and an exact alias targets it, the plugin rewrote the specifier to a nonexistent `./data.js` even though TypeScript-Go copies the asset as `.json`. The build exited 0 and Node failed at runtime with `MODULE_NOT_FOUND`.

### Fix

At the output-path owner (`packages/paths/driver/paths.go`):

- Rename `emittedJavaScriptExtension` → `emittedExtension` and change the default branch to **preserve** the source extension for assets the compiler copies verbatim (`.json`), instead of inventing a `.js` sibling.
- Keep every established TypeScript/JavaScript-family mapping (`.ts`/`.js`→`.js`, `.mts`/`.mjs`→`.mjs`, `.cts`/`.cjs`→`.cjs`, `.tsx`/`.jsx`→`.js` or `.jsx` under `jsx: preserve`).
- Extensionless target lookup (`sourceLookupExtensions`) is **not** widened to include `.json`; only exact `.json` aliases resolve, matching the issue's boundary.

### Tests

- `packages/paths/test/rewriter_maps_source_to_emitted_output_extension_test.go` — full source-to-output extension matrix (copied assets keep their extension incl. nested / double-extension / uppercase; JS/TS families transpile; negative twin at the helper proving unknown assets are never absorbed into `.js`).
- `tests/test-paths/src/features/test_paths_rewrites_commonjs_json_alias_to_copied_extension.ts` — real CommonJS build + Node run; asserts the emit requires `./data.json`, no `./data.js` is invented, the file exists, and the program runs.
- `tests/test-paths/src/features/test_paths_rewrites_esm_json_alias_with_import_attribute.ts` — NodeNext ESM + declarations; asserts the `with { type: "json" }` attribute survives, the specifier is `./data.json`, and neither the `.mjs` nor `.d.mts` output invents a `.js` JSON target.

Docs updated: `website/src/content/docs/plugins/paths.mdx`.

### Issues

Closes #680

### DIFF
--- a/packages/paths/driver/paths.go
+++ b/packages/paths/driver/paths.go
@@ -264,13 +264,21 @@ func (r *rewriter) outputPathForSource(source string) string {
   if err != nil || isOutsideRelativePath(rel) {
     return ""
   }
-  return normalizePath(filepath.Join(r.outDir, replaceSourceExtension(rel, emittedJavaScriptExtension(rel, r.jsxPreserve))))
+  return normalizePath(filepath.Join(r.outDir, replaceSourceExtension(rel, emittedExtension(rel, r.jsxPreserve))))
 }
 
-// emittedJavaScriptExtension returns the JavaScript file extension that TypeScript
-// emits for a given source path.
-func emittedJavaScriptExtension(source string, jsxPreserve bool) string {
-  switch strings.ToLower(filepath.Ext(source)) {
+// emittedExtension returns the output-file extension that TypeScript-Go writes
+// for a given Program source. TypeScript/JavaScript-family sources are
+// transpiled to the matching JavaScript runtime extension. Non-transpiled
+// assets that the compiler copies verbatim — such as `.json` sources pulled in
+// by resolveJsonModule — keep their original extension, so a rewritten
+// specifier names the file the compiler actually emits instead of an invented
+// `.js` sibling that never exists on disk.
+func emittedExtension(source string, jsxPreserve bool) string {
+  ext := filepath.Ext(source)
+  switch strings.ToLower(ext) {
+  case ".ts", ".js":
+    return ".js"
   case ".mts", ".mjs":
     return ".mjs"
   case ".cts", ".cjs":
@@ -281,7 +289,7 @@ func emittedJavaScriptExtension(source string, jsxPreserve bool) string {
     }
     return ".js"
   default:
-    return ".js"
+    return ext
   }
 }
 
--- a/packages/paths/test/linkname_helpers_test.go
+++ b/packages/paths/test/linkname_helpers_test.go
@@ -80,5 +80,5 @@ func pathsReplaceSourceExtension(value string, ext string) string
 //go:linkname pathsIsOutsideRelativePath github.com/samchon/ttsc/packages/paths/driver.isOutsideRelativePath
 func pathsIsOutsideRelativePath(rel string) bool
 
-//go:linkname pathsEmittedJavaScriptExtension github.com/samchon/ttsc/packages/paths/driver.emittedJavaScriptExtension
-func pathsEmittedJavaScriptExtension(source string, jsxPreserve bool) string
+//go:linkname pathsEmittedExtension github.com/samchon/ttsc/packages/paths/driver.emittedExtension
+func pathsEmittedExtension(source string, jsxPreserve bool) string
--- a/packages/paths/test/rewriter_helpers_cover_resolution_edges_test.go
+++ b/packages/paths/test/rewriter_helpers_cover_resolution_edges_test.go
@@ -180,14 +180,14 @@ func TestRewriterHelpersCoverResolutionEdges(t *testing.T) {
   if !pathsIsOutsideRelativePath("..") || !pathsIsOutsideRelativePath(filepath.Join("..", "x")) || pathsIsOutsideRelativePath("..x") {
     t.Fatal("outside relative path classification mismatch")
   }
-  if pathsEmittedJavaScriptExtension("x.mts", false) != ".mjs" ||
-    pathsEmittedJavaScriptExtension("x.mjs", false) != ".mjs" ||
-    pathsEmittedJavaScriptExtension("x.cts", false) != ".cjs" ||
-    pathsEmittedJavaScriptExtension("x.cjs", false) != ".cjs" ||
-    pathsEmittedJavaScriptExtension("x.tsx", true) != ".jsx" ||
-    pathsEmittedJavaScriptExtension("x.jsx", true) != ".jsx" ||
-    pathsEmittedJavaScriptExtension("x.jsx", false) != ".js" ||
-    pathsEmittedJavaScriptExtension("x.ts", false) != ".js" {
+  if pathsEmittedExtension("x.mts", false) != ".mjs" ||
+    pathsEmittedExtension("x.mjs", false) != ".mjs" ||
+    pathsEmittedExtension("x.cts", false) != ".cjs" ||
+    pathsEmittedExtension("x.cjs", false) != ".cjs" ||
+    pathsEmittedExtension("x.tsx", true) != ".jsx" ||
+    pathsEmittedExtension("x.jsx", true) != ".jsx" ||
+    pathsEmittedExtension("x.jsx", false) != ".js" ||
+    pathsEmittedExtension("x.ts", false) != ".js" {
     t.Fatal("emitted extension mismatch")
   }
 }
--- a/packages/paths/test/rewriter_maps_source_to_emitted_output_extension_test.go
+++ b/packages/paths/test/rewriter_maps_source_to_emitted_output_extension_test.go
@@ -1,0 +1,73 @@
+package paths_test
+
+import (
+  "testing"
+)
+
+// TestRewriterMapsSourceToEmittedOutputExtension verifies the source-to-output
+// extension matrix at outputPathForSource.
+//
+// The predictor must name the file TypeScript-Go actually emits or copies for a
+// resolved Program source. TypeScript/JavaScript-family sources transpile to
+// their runtime suffix, but copied assets (`.json` under resolveJsonModule) keep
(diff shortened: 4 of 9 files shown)


samchon:
### Self-Review round 1

Reviewed the whole declared surface (driver predictor, resolution helpers, Go matrix + edge tests, both feature fixtures, docs).

**Verified sound:**
- `emittedExtension` default branch preserves the source extension verbatim (case included) for copied assets while every TS/JS-family mapping is unchanged; confirmed `replaceSourceExtension`/`stripKnownSourceExtension` do not double-strip `.json` (incl. `data.config.json`, `data.ts.json`).
- Both feature fixtures pass end-to-end against a real `ttsc --emit` + Node run (CommonJS require, NodeNext ESM with `with { type: "json" }` attribute retained, `.d.mts` free of any invented `.js` JSON target).
- `sourceLookupExtensions` is untouched, so extensionless lookup is not widened.

**Finding (test-coverage gap):** RA-27 requires negative coverage that the extensionless target lookup is *not accidentally widened* to `.json`, and that an exact `.json` alias resolves to the copied source. The current matrix test only exercises `outputPathForSource` (post-resolution); nothing pins the resolution seam (`resolveSource`/`lookupSource`) for JSON. A future edit adding `.json` to `sourceLookupExtensions` would silently pass all current tests.

**Remediation:** add a dedicated resolution-seam test — an exact `@data` -> `src/data.json` alias rewrites to `./data.json`, while an extensionless `@bare` -> `src/config` alias with only `src/config.json` present stays unrewritten (negative twin). Then re-run the Go suite.

samchon:
### Self-Review round 2

Re-reviewed after adding the resolution-seam test. Two findings from tracing the rename/behavior-change consequence surface.

**Finding A (stale walkthrough doc) — FIXED in my surface.** `website/src/content/docs/development/walkthroughs/paths.mdx` reproduced the old `emittedJavaScriptExtension` implementation verbatim (default `.js`), which the rename + JSON-preservation change made inaccurate. Updated the code block and prose to `emittedExtension`, explained the `default` branch preserves copied-asset extensions (`.json` -> `./data.json`), and retitled the section to "emitted-output path". The `lookupSource` extension list in that walkthrough is unchanged and still accurate.

**Finding B (compiler's own emittedJavaScriptExtension) — investigated, no change needed; recorded for lead.** `packages/ttsc/src/compiler/internal/resolveEmittedJavaScript.ts` has a same-named helper that also maps non-`.mts`/`.cts` to `.js`. It is a *different owner* outside this bundle's surface, and it does NOT share the runtime-breaking bug: `resolveExactEmittedFile` guards its prediction with `fs.existsSync(exact)` and the fallback only scores `isJavaScriptOutput` files, so a `.json` source degrades to `null` (no match) rather than an invented specifier. Left untouched per the bundle boundary.

No code-behavior findings on the driver or tests. Next: commit the doc fix and re-run the Go suite for a clean round.

samchon:
### Self-Review round 3 — clean

Re-reviewed the full diff after the round-2 doc fix. The driver change is minimal and focused: `.ts`/`.js`/`.d.ts` (ext `.ts`) still map to `.js`, `.mts`/`.cts`/JSX-preserve mappings unchanged, and only the `default` branch changed — from inventing `.js` to preserving the source's own extension, which is exactly the copied-asset class (`.json`). No new findings. Ending the self-review loop at 3 rounds.

samchon:
### Final verification (all green)

Package-scoped, from the worktree on `fix/paths-json-ext`:

**Go unit + resolution + matrix tests** — `cd packages/paths && go test ./... -count=1`
```
ok  github.com/samchon/ttsc/packages/paths/test  14.023s
```
Includes `TestRewriterMapsSourceToEmittedOutputExtension` (source→output extension matrix), `TestRewriterResolvesExactJsonAliasWithoutWideningExtensionlessLookup` (resolution-seam positive + negative twin), and the unchanged `TestRewriterHelpersCoverResolutionEdges`.

**Real build-and-run feature fixtures** — `cd tests/test-paths && pnpm start`
```
- test_paths_rewrites_allow_js_extensionless_targets_to_emitted_extensions: 84,113 ms
- test_paths_rewrites_commonjs_json_alias_to_copied_extension: 2,198 ms
- test_paths_rewrites_esm_imports_and_re_exports: 2,081 ms
- test_paths_rewrites_esm_json_alias_with_import_attribute: 1,999 ms
- test_paths_uses_nodenext_output_extensions_from_tsconfig: 1,900 ms
Success
```
The CommonJS fixture asserts the emit requires `./data.json` (not `./data.js`), `dist/data.json` exists, no `dist/data.js` is invented, and Node runs the program. The ESM/NodeNext fixture asserts the `with { type: "json" }` attribute survives, the specifier is `./data.json`, `dist/data.json` exists, and `dist/main.d.mts` introduces no `.js` JSON target.

RA-27 / #680 acceptance criteria all satisfied. Marking ready for lead recheck — not merging.

samchon:
Squash-merged to master as `2799f0606e23654e9bac4dc61f2de19762185cc4` (one clean `fix` commit; linked issues auto-close from the PR body). Per campaign CI discipline, the merge-commit workflows were cancelled by exact SHA (gate STABLE: both runs cancelled, two consecutive idle passes).

PR #693, opened 2026-07-15
fix(wasm): keep boot and MemFS invariants consistent

### LINKED ISSUE #671: fix(wasm): preserve MemFS tree and file-data invariants for successful mutations
## Problem

`@ttsc/wasm` exposes a public in-memory filesystem as the `fs` host for Go/Wasm. Several mutation methods report success without establishing the state that Node/Go callers are entitled to observe:

- Renaming a directory moves only the directory node and leaves every descendant at the old path.
- Removing a non-empty directory deletes only its directory node and leaves orphaned descendants.
- `truncate` and `ftruncate` return success without changing file bytes or validating the path/file descriptor.

These are false-success results. Custom Wasm hosts and compiler code continue under an invalid filesystem model instead of receiving either the requested mutation or an error.

## Evidence

The owner is `createMemFS` in `packages/wasm/src/createMemFS.ts`, where one `nodes` map represents the complete filesystem tree.

- `fs.rename` deletes the normalized source key and inserts the same node at the destination key; it does not move keys below a directory prefix.
- `fs.rmdir` delegates directly to `fs.unlink`, which deletes the named node without checking directory type, children, or root constraints.
- `fs.truncate` and `fs.ftruncate` immediately invoke `callback(null)` and do not inspect or mutate any node or descriptor.

The public `createMemFS()` helpers and callback APIs produce this proof:

```ts
const host = createMemFS();
host.mkdirp("/old/child");
host.writeFile("/old/child/file.txt", "abcdef");

await new Promise<void>((resolve, reject) =>
  host.fs.rename("/old", "/new", (error) =>
    error ? reject(error) : resolve(),
  ),
);
console.log(host.exists("/old"), host.exists("/old/child/file.txt"));
console.log(host.exists("/new"), host.exists("/new/child/file.txt"));
// false, true, true, false

await new Promise<void>((resolve, reject) =>
  host.fs.truncate("/old/child/file.txt", 2, (error) =>
    error ? reject(error) : resolve(),
  ),
);
console.log(host.readFileText("/old/child/file.txt"));
// "abcdef"

await new Promise<void>((resolve, reject) =>
  host.fs.rmdir("/old/child", (error) =>
    error ? reject(error) : resolve(),
  ),
);
console.log(host.exists("/old/child"), host.exists("/old/child/file.txt"));
// false, true
```

The observed state shows each operation reports success while violating its named mutation.

## Consequence surface

The returned `IMemFSHost` is public and plugin-agnostic; it is not restricted to the currently observed tiny compiler fixture. Go's filesystem layer and custom browser hosts may call these methods and then rely on path existence, directory traversal, file length, descriptor contents, or subsequent cleanup. Orphaned nodes also make `exists`, `stat`, `readdir`, and later writes disagree about the tree.

Directory/file type checks, normalized paths, descriptor bookkeeping, overwrite behavior, open descriptor paths after rename, and truncation extension with zero bytes are part of the same mutation invariant. The issue is limited to this verified false-success family; unrelated intentionally unsupported operations need not be implemented.

## Approach

Keep the `nodes` map and descriptor table inside `createMemFS` as the state owners. A successful callback must mean the complete tree and file data satisfy the requested operation; otherwise return an appropriate `MemFSError` without partial mutation.

Define behavior against Node-style filesystem invariants before choosing the internal update algorithm. Directory moves must be atomic from the caller's perspective, removal must preserve a valid tree, and path- and descriptor-based truncation must update the same underlying file state.

- File rename moves the file and preserves its bytes; directory rename moves every descendant and leaves no old-prefix nodes.
- Rename collision, invalid-parent, self/descendant, root, and file-versus-directory cases either follow the selected Node-compatible contract or fail without partial state changes.
- `rmdir` succeeds for an empty directory, rejects a non-empty directory, and never leaves orphaned descendants.
- `truncate` shrinks bytes, extends with zero-filled bytes, rejects invalid lengths and non-files, and reports a missing path.
- `ftruncate` applies the same data rules through a valid open descriptor and reports invalid or inappropriate descriptors.
- Tests assert post-operation `exists`, `stat`, `readdir`, `readFile`, and descriptor reads, not only callback success.
- Add a TestExecutor-based `tests/test-wasm` package (none exists today) with focused `memfs` mutation cases, runnable with `pnpm --filter @ttsc/test-wasm start -- --include=memfs`.
- Run `pnpm --filter @ttsc/wasm build:ts`, the focused regression above, and then the canonical repository checks `pnpm build` and `pnpm test`.

## Coordination

Merged PRs #83, #87, #98, and #173 established the Wasm/browser host and its packaging evolution. None is an open tracker for these successful-but-incomplete mutations.

Open PR #657, `fix(test): audit every lint corpus source`, changes lint-corpus test discovery only and does not touch `createMemFS` or Wasm host behavior.

[Issue #688](https://github.com/samchon/ttsc/issues/688), **Reject bootTtsc when the Go host exits before readiness**, and [issue #673](https://github.com/samchon/ttsc/issues/673), **Keep the returned MemFS host aligned with the Go runtime across boot retries**, share the Wasm package but own boot lifecycle and host identity. Coordinate the shared `tests/test-wasm` package if worked concurrently; this issue owns filesystem mutation state only.

### LINKED ISSUE #673: fix(wasm): keep the returned MemFS host aligned with the Go runtime across boot retries
## Problem

`bootTtsc` creates a new default MemFS host for every retry, but `bootTtscOnce` installs `globalThis.fs` and `globalThis.process` only when those globals are absent. If an attempt fails after installing its host but before boot completes, the next attempt returns a newly created host while the Go runtime continues using the first attempt's global filesystem.

The returned `IBootResult.host` is the public handle callers mount and mutate. After this failure/retry sequence, those mutations are invisible to the compiler runtime. A successful boot must return the exact host backing that runtime.

## Evidence

The stable owner is `packages/wasm/src/bootTtsc.ts::bootTtscOnce`:

```ts
const host = options.host ?? createMemFS();
if (!globalAny.fs) globalAny.fs = host.fs;
if (!globalAny.process) globalAny.process = createProcessShim();
// ...
return { api, host };
```

The rejection handler correctly removes the failed Promise from `bootsInFlight`, so a later call retries. It does not restore or replace the globals installed before the failure.

The deterministic reproduction uses the default host and a fetch stub:

1. Start with no `globalThis.fs`.
2. Call `bootTtsc` and make the first WASM fetch return HTTP 503. Attempt one creates host 1, installs `host1.fs` globally, then rejects.
3. Call `bootTtsc` again for the same API and URL and let the fetch/boot succeed. Attempt two creates host 2, sees that `globalThis.fs` already exists, and leaves `host1.fs` installed.
4. Compare the successful result's `host.fs` with `globalThis.fs`.

The observed result was:

```json
{"fetches":2,"same":false}
```

The mismatch is established before any application-specific plugin behavior: it follows directly from retry eviction plus the conditional global assignment. A custom `options.host` on the retry has the same problem when an earlier attempt installed a different host.

## Consequence surface

- Playground and custom browser consumers can mount files into the returned host while the Go compiler reads a different filesystem.
- Diagnostics, build, transform, snapshot, and plugin operations all use the Go runtime's global fs and can report missing or stale files.
- Failures after global setup but before readiness are affected, including fetch, instantiate, runtime, and expose failures.
- Concurrent identical boots still need single-flight semantics, and separate workers remain the supported way to host separate WASM instances.
- Explicitly reusing the same host across attempts is safe today and must remain safe.

## Approach

`bootTtscOnce` owns the host/global/runtime relationship. The invariant is: on every successful return, `IBootResult.host.fs` must be the filesystem captured by that boot's Go runtime. Retry cleanup and serialization must maintain that identity without making two live runtimes share or overwrite incompatible bridge state.

The implementation should account for every failure point after global installation and for both default and caller-supplied hosts. Do not paper over the mismatch by returning an unrelated host or silently retaining stale files.

- A first attempt that fails after host installation can be retried successfully.
- On the successful retry, the returned host is the exact host used by the Go runtime.
- Files written through the returned host are visible to a real API operation.
- The invariant holds for default-created hosts and explicit `options.host` values.
- Concurrent calls for the same key remain single-flight; same-API different-URL boots remain serialized.
- A normal first-attempt boot and an explicit reuse of one host keep their current behavior.
- Add a focused boot-lifecycle regression under `tests/test-ttsc/src/features/wasm/` (or an equivalent package-owned Worker harness) that injects a failure before a successful retry and proves filesystem identity plus an API-visible file. The test must use the built `@ttsc/wasm` entry rather than a copied boot implementation.
- Prove it with `pnpm --filter @ttsc/wasm build`, `pnpm --filter @ttsc/test-ttsc start`, and the broader `pnpm test`.

## Coordination

This shares the `bootTtsc` lifecycle with campaign issue #688, **Reject bootTtsc when host.Expose panics before readiness**, but the failures are independent: #688 ensures a failed boot settles, while this issue ensures a later successful retry returns the runtime's real host. They should coordinate around one failure-cleanup invariant without being collapsed.

No corrective open or closed tracker was found in the WASM boot lineage. The only current open pull request, #657 (**fix(test): audit every lint corpus source**), changes lint corpus and harness files only and does not touch `packages/wasm`.

### LINKED ISSUE #688: fix(wasm): reject bootTtsc when the Go host exits before readiness
## Problem

`bootTtsc` waits for `globalThis[apiName + "Ready"]` or an explicit `Failed` callback from the Go host. If `host.Expose` panics before invoking either callback, the Go runtime exits but the public boot Promise remains pending forever. A compiler UI or custom Playground waits indefinitely with no rejectable cause even though Go already printed the panic and terminated.

Duplicate plugin names are one concrete public trigger. The generic lifecycle defect is any Go/Wasm exit before readiness that does not call the optional failure bridge.

## Evidence

The two owners expose the gap directly:

- `packages/wasm/host/host.go::Expose` reports a second `Expose` call through `<apiName>Failed`, but duplicate plugin registration still executes `panic(fmt.Sprintf("host.Expose: duplicate plugin name %q", name))` before the API or Ready signal is installed.
- `packages/wasm/src/bootTtsc.ts::bootTtscOnce` constructs a `ready` Promise with Ready/Failed callbacks, starts `void go.run(wasm.instance)`, and then awaits only `ready`. The `go.run` Promise is neither observed nor raced with readiness, so neither fulfillment nor rejection can settle the public boot.
- In the current Go `lib/wasm/wasm_exec.js`, `Go.exit` only warns for a non-zero code and resolves `_exitPromise`; `Go.run` awaits that Promise and fulfills without returning the exit code. A panic-driven exit therefore normally reaches the ignored fulfillment branch, not a rejection carrying the panic.

A minimal custom host is:

```go
package main

import "github.com/samchon/ttsc/packages/wasm/host"

type duplicate struct{}
func (duplicate) Name() string { return "duplicate" }
func (duplicate) Run(string, []string) int { return 0 }

func main() {
  host.Expose("ttsc", host.Config{
    Plugins: []host.Plugin{duplicate{}, duplicate{}},
  })
}
```

Build it with the normal custom-host target (`GOOS=js GOARCH=wasm go build -o duplicate.wasm`) and boot it through `bootTtsc({ wasmUrl: "duplicate.wasm" })` in the supported Worker environment. The runtime prints `panic: host.Expose: duplicate plugin name "duplicate"` and exits with code 2. Racing the returned Promise against a two-second timer yields the timer: the Promise is still pending rather than rejected.

## Consequence surface

This affects the public `@ttsc/wasm` boot API and every UI or worker client awaiting it. A malformed consumer-built host, an early initialization panic, or another pre-ready Go exit can strand the boot chain for that `apiName`; later boots serialized behind the pending Promise cannot begin either.

Normal long-lived hosts intentionally keep `go.run` pending after Ready, so the correct invariant is specifically about exit before readiness. Ready and Failed callback cleanup, in-flight cache eviction, per-api serialization, retry behavior, and unhandled Promise rejections are part of the lifecycle surface.

## Approach

Treat `bootTtscOnce` as the JavaScript lifecycle owner and `host.Expose` as the Go-side diagnostic owner. Enforce the invariant that boot settles exactly once: Ready resolves it, an explicit Failed signal rejects it, and any `go.run` settlement before Ready rejects it. Because the standard Go runner discards the exit code, a generic early fulfillment can only produce an actionable synthesized “runtime exited before readiness” error; known host validation failures such as duplicate plugin names must invoke Failed before returning so their original cause survives.

Host validation alone is not a complete boundary for unknown early exits. Likewise, racing `go.run` must preserve the expected post-Ready long-running runtime and attach handlers to the losing branch so later runtime completion cannot become an unhandled rejection. Do not replace lifecycle settlement with an arbitrary timeout.

- A custom host with duplicate plugin names rejects `bootTtsc` promptly and preserves the duplicate-name/Go exit cause.
- Any test Wasm that exits or panics before Ready without invoking Failed rejects with an actionable early-exit error rather than remaining pending.
- A normal host still resolves on Ready while its Go runtime continues running.
- Explicit `<apiName>Failed` behavior and the existing double-Expose refusal still reject exactly once.
- Failure removes the relevant in-flight entry and releases the per-api serialization chain so a later valid retry can run.
- No losing lifecycle branch produces an unhandled rejection, overwrites the winning result, or leaves stale Ready/Failed callbacks.
- Add the Worker lifecycle harness to a TestExecutor-based `tests/test-wasm` package (create it if issue #671 has not done so) and run `pnpm --filter @ttsc/test-wasm start -- --include=boot_ttsc_rejects_before_readiness`.
- Run `pnpm --filter @ttsc/wasm build`, the focused regression above, and then the canonical repository checks `pnpm build` and `pnpm test`.

## Coordination

The Wasm host and boot API evolved through merged PRs #83, #87, #98, and #173. Current code already documents and handles the explicit second-Expose failure signal, but no existing tracker covers a panic or runtime exit before readiness.

Open PR #657, `fix(test): audit every lint corpus source`, changes lint-corpus code and tests only and does not touch the Wasm host or boot lifecycle.

[Issue #673](https://github.com/samchon/ttsc/issues/673), **Keep the returned MemFS host aligned with the Go runtime across boot retries**, owns host identity after a failed boot and retry. It is directly adjacent but separate: this issue ensures the first failed boot settles and unblocks the retry. [Issue #671](https://github.com/samchon/ttsc/issues/671), **Preserve MemFS tree and file-data invariants for successful mutations**, owns filesystem operation semantics, not boot settlement.

samchon:
## Bundle b5 — wasm boot + MemFS invariants

Part of the `repository-audit-2026-07` issue campaign. This batch fixes three adjacent defects in `@ttsc/wasm`: two in the `bootTtsc` lifecycle and one in the `createMemFS` filesystem-mutation surface. They share the package and are validated by one new `tests/test-wasm` harness.

### Issues

- Closes #688 (RA-18) — **Reject `bootTtsc` when the Go host exits before readiness.** A pre-`Ready` Go exit (e.g. an early `host.Expose` panic) left the public boot Promise pending forever. Race `go.run` settlement against readiness so an early exit rejects with an actionable cause; route the duplicate-plugin-name host error through the `Failed` bridge (instead of `panic`) so its real cause survives.
- Closes #673 (RA-20) — **Keep the returned MemFS host aligned with the Go runtime across boot retries.** A failed attempt that had already installed `globalThis.fs`/`process` left them in place, so a later retry returned a different host than the runtime used. Restore only this attempt's own globals on failure, so a successful retry returns the exact host backing the runtime.
- Closes #671 (RA-13) — **Preserve MemFS tree and file-data invariants for successful mutations.** `rename` moved only a directory's own node (orphaning descendants); `rmdir` delegated to `unlink` with no type/emptiness/root checks; `truncate`/`ftruncate` were no-ops. Enforce Node/POSIX invariants: atomic subtree moves with descriptor follow, empty-directory `rmdir`, directory-refusing `unlink`, and validated path/descriptor truncation.

### Surface

- `packages/wasm/src/bootTtsc.ts` — lifecycle settlement + retry global restore.
- `packages/wasm/host/host.go` — duplicate-plugin-name via `Failed` bridge, no panic.
- `packages/wasm/src/createMemFS.ts` — rename/rmdir/unlink/truncate/ftruncate invariants.
- `packages/wasm/src/MemFSError.ts` — `EBUSY`/`ENOTEMPTY` errno mappings.
- `tests/test-wasm` — new TestExecutor package: `memfs` mutation cases + `boot` lifecycle harness.

### Verification

Package-scoped local verification (campaign CI is suspended per push):

- `pnpm --filter @ttsc/wasm build`
- `pnpm --filter @ttsc/test-wasm start`

Draft until implementation, tests, and solo Self-Review are complete.

### DIFF
--- a/packages/wasm/host/host.go
+++ b/packages/wasm/host/host.go
@@ -81,7 +81,19 @@ func Expose(apiName string, cfg Config) {
       continue
     }
     if _, dup := plugins[name]; dup {
-      panic(fmt.Sprintf("host.Expose: duplicate plugin name %q", name))
+      // A duplicate plugin name is a host-configuration error, but panicking
+      // here would terminate the Go runtime before any Ready/Failed signal is
+      // installed, leaving bootTtsc's `await ready` pending forever with no
+      // observable cause. Mirror the double-Expose path: surface the cause via
+      // console.error + the JS-visible Failed bridge, then return without
+      // starting the keepalive runtime so the boot rejects with the real
+      // reason instead of a generic "exited before readiness" error.
+      msg := fmt.Sprintf("host.Expose: duplicate plugin name %q", name)
+      fmt.Fprintln(os.Stderr, msg)
+      if failed := js.Global().Get(apiName + "Failed"); failed.Type() == js.TypeFunction {
+        failed.Invoke(js.Global().Get("Error").New(msg))
+      }
+      return
     }
     plugins[name] = p
     pluginNames = append(pluginNames, name)
--- a/packages/wasm/src/MemFSError.ts
+++ b/packages/wasm/src/MemFSError.ts
@@ -27,6 +27,8 @@ function errnoForCode(code: string): number {
       return -2;
     case "EBADF":
       return -9;
+    case "EBUSY":
+      return -16;
     case "EEXIST":
       return -17;
     case "ENOTDIR":
@@ -37,6 +39,8 @@ function errnoForCode(code: string): number {
       return -22;
     case "ESPIPE":
       return -29;
+    case "ENOTEMPTY":
+      return -39;
     default:
       return -1;
   }
--- a/packages/wasm/src/bootTtsc.ts
+++ b/packages/wasm/src/bootTtsc.ts
@@ -109,56 +109,124 @@ async function bootTtscOnce(
 
   const host = options.host ?? createMemFS();
   const globalAny = globalThis as Record<string, unknown>;
-  // Only install fs / process if they aren't already in place. A caller
-  // booting a second wasm over the same MemFS reuses the same shims.
-  if (!globalAny.fs) globalAny.fs = host.fs;
-  if (!globalAny.process) globalAny.process = createProcessShim();
-
-  // wasm_exec.js installs `globalThis.Go`. It also reads globalThis.fs at
-  // module-eval time, so this import must follow the assignment above.
-  importScripts(wasmExecUrl);
-
-  // Race the Ready resolver against a Failed signal so a wasm-side fault
-  // (e.g. `host.Expose` refusing a duplicate call) surfaces here instead of
-  // hanging on `await ready` forever. `go.run` is fire-and-forget so its
-  // own rejection cannot reach this promise without an explicit channel.
-  const ready = new Promise<void>((resolve, reject) => {
-    globalAny[apiName + "Ready"] = () => {
-      delete globalAny[apiName + "Failed"];
-      resolve();
-    };
-    globalAny[apiName + "Failed"] = (err: unknown) => {
-      delete globalAny[apiName + "Ready"];
-      reject(err instanceof Error ? err : new Error(String(err)));
-    };
-  });
-
-  const goCtor = (globalAny as { Go?: new () => IGoInstance }).Go;
-  if (typeof goCtor !== "function") {
-    throw new Error(
-      `bootTtsc: globalThis.Go was not installed by ${wasmExecUrl} — the file may not have loaded (CSP block, wrong content type, 404), or it is not the wasm_exec.js shipped with the Go toolchain.`,
-    );
+  // Install fs / process only if they aren't already in place, and remember
+  // whether THIS attempt installed them. The Go runtime this boot starts reads
+  // `globalThis.fs` when it runs, so the returned `host` must be the exact host
+  // backing those globals. A caller booting a second wasm over the same MemFS
+  // (or reusing one host across retries) reuses the same shims. When an earlier
+  // failed attempt already installed a different host's shims, those are torn
+  // down on failure below so this attempt can install its own.
+  const installedFs = !globalAny.fs;
+  const installedProcess = !globalAny.process;
+  let processShim: unknown;
+  if (installedFs) globalAny.fs = host.fs;
+  if (installedProcess) {
+    processShim = createProcessShim();
+    globalAny.process = processShim;
   }
-  const go = new goCtor();
 
-  const response = await fetch(wasmUrl);
-  if (!response.ok) {
-    throw new Error(`bootTtsc: failed to fetch ${wasmUrl}: ${response.status}`);
-  }
-  const wasm = await WebAssembly.instantiateStreaming(
-    response,
-    go.importObject,
-  );
-  // go.run never resolves until the wasm exits; we don't await it.
-  void go.run(wasm.instance);
-  await ready;
+  // Any failure after global installation must leave the globals as this
+  // attempt found them, so a retry installs its own host's fs (and the returned
+  // host keeps matching the runtime's filesystem). Only remove what we
+  // installed and only while it is still ours — never stomp a foreign fs or one
+  // a concurrently-booted runtime already claimed.
+  const restoreGlobals = (): void => {
+    if (installedFs && globalAny.fs === host.fs) delete globalAny.fs;
+    if (installedProcess && globalAny.process === processShim)
+      delete globalAny.process;
+  };
 
-  const api = (globalAny as Record<string, ITtscApi | undefined>)[apiName];
-  if (!api)
-    throw new Error(
-      `bootTtsc: ${apiName} global was not set — was the wasm built with host.Expose(${JSON.stringify(apiName)}, ...)?`,
+  try {
+    // wasm_exec.js installs `globalThis.Go`. It also reads globalThis.fs at
+    // module-eval time, so this import must follow the assignment above.
+    importScripts(wasmExecUrl);
+
+    // Race the Ready resolver against a Failed signal so a wasm-side fault
+    // (e.g. `host.Expose` refusing a duplicate call) surfaces here instead of
+    // hanging on `await ready` forever.
+    let readyCb!: () => void;
+    let failedCb!: (err: unknown) => void;
+    const ready = new Promise<void>((resolve, reject) => {
+      readyCb = () => {
+        delete globalAny[apiName + "Failed"];
+        resolve();
+      };
+      failedCb = (err: unknown) => {
+        delete globalAny[apiName + "Ready"];
+        reject(err instanceof Error ? err : new Error(String(err)));
+      };
+      globalAny[apiName + "Ready"] = readyCb;
+      globalAny[apiName + "Failed"] = failedCb;
+    });
+
+    const goCtor = (globalAny as { Go?: new () => IGoInstance }).Go;
+    if (typeof goCtor !== "function") {
+      throw new Error(
+        `bootTtsc: globalThis.Go was not installed by ${wasmExecUrl} — the file may not have loaded (CSP block, wrong content type, 404), or it is not the wasm_exec.js shipped with the Go toolchain.`,
+      );
+    }
+    const go = new goCtor();
+
+    const response = await fetch(wasmUrl);
+    if (!response.ok) {
+      throw new Error(
+        `bootTtsc: failed to fetch ${wasmUrl}: ${response.status}`,
+      );
+    }
+    const wasm = await WebAssembly.instantiateStreaming(
+      response,
+      go.importObject,
     );
-  return { api, host };
+
+    // A normal host keeps `go.run` pending forever after signaling Ready, so a
+    // settlement (fulfil OR reject) BEFORE Ready means the Go runtime exited or
+    // panicked before it could register — e.g. an early `host.Expose` panic
+    // that never reached the Failed bridge. Race that early exit against
+    // readiness so the boot rejects with an actionable cause instead of hanging.
+    // The standard Go runner discards the exit code, so an unknown early exit
+    // can only synthesize a generic message; known host validation failures
+    // reject through Failed above and keep their original cause.
+    const runPromise = Promise.resolve(go.run(wasm.instance));
+    const earlyExit = runPromise.then(
+      () => {
+        throw new Error(
+          `bootTtsc: the ${apiName} wasm runtime exited before signaling readiness (the host may have panicked; check the wasm stderr).`,
+        );
+      },
+      (err: unknown) => {
+        throw new Error(
+          `bootTtsc: the ${apiName} wasm runtime failed before signaling readiness: ${
+            err instanceof Error ? err.message : String(err)
+          }`,
+        );
+      },
+    );
+    // When Ready wins, the long-running runtime's eventual `go.run` settlement
+    // must not surface as an unhandled rejection. Attach a terminal handler to
+    // the losing branch.
+    earlyExit.catch(() => {});
+
+    try {
+      await Promise.race([ready, earlyExit]);
+    } finally {
+      // Drop this attempt's readiness bridge so a later boot for the same
+      // apiName installs a clean pair and no stale resolver survives.
+      if (globalAny[apiName + "Ready"] === readyCb)
+        delete globalAny[apiName + "Ready"];
+      if (globalAny[apiName + "Failed"] === failedCb)
+        delete globalAny[apiName + "Failed"];
+    }
+
+    const api = (globalAny as Record<string, ITtscApi | undefined>)[apiName];
+    if (!api)
+      throw new Error(
+        `bootTtsc: ${apiName} global was not set — was the wasm built with host.Expose(${JSON.stringify(apiName)}, ...)?`,
+      );
+    return { api, host };
+  } catch (err) {
+    restoreGlobals();
+    throw err;
+  }
 }
 
 /**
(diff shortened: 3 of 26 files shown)


samchon:
### Self-Review — Round 1

Fresh full-surface pass over the diff (bootTtsc.ts, createMemFS.ts, host.go, MemFSError.ts, the new tests, and the doc).

**Findings**

1. **Test coverage gap vs RA-13 acceptance (actionable).** RA-13 requires tests to "assert post-operation `exists`, `stat`, `readdir`, `readFile`, and descriptor reads, not only callback success." My memfs tests exercise `exists`, `readFile`/`readFileText`, and descriptor reads, but not `stat` or `readdir`. `readdir` is the strongest orphan detector (it scans the node map by prefix), so its absence leaves a real blind spot for the directory-rename and rmdir cases.

**Remediation plan**

- Add promisified `readdir`/`stat` helpers to `internal/callbackFs.ts`.
- Extend the directory-rename test to assert `readdir("/new/child")` lists the moved file and the old parent is gone, plus a `stat` on the moved file reports a file of the right size.
- Extend the rmdir test to assert `readdir` of the surviving non-empty directory still lists its child (orphan-free).

**Verified sound, no change needed**

- Boot race settles exactly once: `Promise.race([ready, earlyExit])` with `earlyExit.catch(()=>{})` and the `finally` bridge cleanup; the losing never-settling `go.run` branch produces no unhandled rejection (asserted by two tests). Confirmed each boot test hangs/fails on the pre-fix base and passes on the fix.
- `restoreGlobals` only removes globals this attempt installed and only while still its own; retry identity holds for default and supplied hosts (asserted).
- MemFS mutation guards (self/descendant, non-empty overwrite, file-vs-dir, root, missing parent) reject before any mutation — no partial state (asserted, tree-intact check).
- `unlink` EISDIR matches Go's `os.Remove` unlink→rmdir fallback; host temp-file cleanup only unlinks files, unaffected.
- Go host builds/vets clean for `js/wasm`.

samchon:
### Self-Review — Round 2 (clean)

Fresh full-surface pass after the Round 1 test strengthening. Re-derived the createMemFS mutation edge cases and the boot lifecycle from scratch:

- **rename subtree overwrite:** a non-empty destination directory is rejected with `ENOTEMPTY` *before* `moveSubtree` runs, so the only reachable overwrite is onto an absent or empty destination — no cross-subtree key can be clobbered.
- **`hasChildren`/`parentDir`:** prefix is `dir + "/"`, so a sibling like `/ab` is never a false child of `/a`; `parentDir` of a top-level path resolves to `/`.
- **boot race:** settles exactly once; both the `Promise.race` handler and the explicit `.catch(()=>{})` cover the losing `go.run` branch; the `finally` bridge cleanup keys off identity so it never deletes a later boot's resolver.
- **harness isolation:** every touched global (`fs`, `Go`, `importScripts`, `fetch`, `WebAssembly.instantiateStreaming`, the api/Ready/Failed slots) is snapshotted by presence and restored in `finally`; `process` (real in Node) is left untouched. Tests run sequentially (`simultaneous: 1`).

No sound improvement or new issue candidate found. Ending the Self-Review loop at 2 rounds (1 with a change, 1 clean).

samchon:
### Final local verification

Campaign CI is suspended per push; the following were run locally in the worktree, all green.

| Command | Result |
| --- | --- |
| `GOOS=js GOARCH=wasm go build ./host/...` (in `packages/wasm`) | exit 0 |
| `GOOS=js GOARCH=wasm go vet ./host/...` (in `packages/wasm`) | exit 0 |
| `pnpm --filter @ttsc/wasm build` | exit 0 — emitted `dist/ttsc.wasm` (32.40 MiB) + `tsc` lib |
| `tsc -p tests/test-wasm/tsconfig.json --noEmit` | exit 0 |
| `pnpm --filter @ttsc/test-wasm start` | Success — 15/15 cases pass |
| `pnpm --filter @ttsc/test-wasm start -- --include=memfs` | Success — 9/9 |
| `pnpm --filter @ttsc/test-wasm start -- --include=boot_ttsc_rejects_before_readiness` | Success |

**Regression proof (pre-fix base `a400d0ce1`):** with the two source files reverted and rebuilt, 7/9 memfs cases error and `test_boot_ttsc_rejects_before_readiness` / `test_boot_ttsc_retries_after_early_exit` hang forever (the RA-18 pending-boot bug), and `test_boot_ttsc_retry_returns_runtime_host` fails on host identity (RA-20). Restoring the fix turns them all green — the tests fail for the right reason.

**Issue outcomes:** #671 (RA-13), #688 (RA-18), #673 (RA-20) all implemented and covered. Ready for lead recheck; not merging.

samchon:
Squash-merged into `master` as `aabc3856f6ea6c8693aedb40d3c7a07df3e0fc57`. The merge-commit CI runs were cancelled by exact SHA per campaign CI-suspension discipline. Linked issues (#671, #688, #673) auto-close from the PR body.

PR #695, opened 2026-07-15
fix(unplugin): harden adapters, cache eviction, packaging, and tsconfig presets

### LINKED ISSUE #665: fix(unplugin): make the Bun adapter safe for plugin-reported dependencies
## Problem

`@ttsc/unplugin` supports transform plugins that report source dependencies so bundlers can invalidate generated code when type-only or plugin-consulted files change. The Bun adapter invokes the raw unplugin transform with an empty object as its transform context. When a valid dependency list is reported, the shared transform calls `this.addWatchFile`, which does not exist on that object, and the Bun loader throws.

The expected behavior is that a valid transform result, including a dependency list, never crashes either `Bun.build` or the `Bun.plugin` runtime loader. Any dependency capability that Bun cannot represent must be handled explicitly at the Bun adapter boundary rather than by passing an invalid unplugin context.

## Evidence

The stable owners are the `setup`/`onLoad` path in `packages/unplugin/src/bun.ts`, the unplugin factory's `transform` method in `packages/unplugin/src/core/index.ts`, and dependency notification in `packages/unplugin/src/core/transform.ts`.

- `bun.ts` creates `raw = unplugin.raw(options, ...)`, then invokes `raw.transform.call({} as never, source, args.path)`.
- The raw transform is the factory method from `core/index.ts`. It passes an `addWatchFile` hook to `transformTtsc` whose body is `this.addWatchFile(watched)`.
- `transformTtsc` calls that hook once for every dependency selected from a successful transform envelope. The empty receiver supplied by Bun has no such method.

The existing unplugin fixture can deterministically reproduce the failure:

1. create the standard `TestUnpluginProject` fixture with its `emit-dependencies` operation and a dependency such as `types/model.ts`;
2. instantiate the Bun adapter and capture the callback registered through `build.onLoad`;
3. invoke that callback for the fixture's TypeScript source; and
4. allow the transform to return its valid `dependencies` map.

The observed outcome is:

```text
TypeError: this.addWatchFile is not a function
```

The focused core test below proves the same dependency envelope is valid and reaches the watch hook when an adapter supplies one:

```bash
pnpm --filter @ttsc/test-unplugin start -- --include=transformttsc_forwards_plugin_dependencies_to_the_watch_hook
```

The defect is therefore in Bun context adaptation, not plugin output or dependency selection.

The adapter originated in merged PR [#47](https://github.com/samchon/ttsc/pull/47). Plugin dependency propagation was added by merged PR [#220](https://github.com/samchon/ttsc/pull/220), and merged PR [#327](https://github.com/samchon/ttsc/pull/327) added the Bun runtime preload path. None contains a correction for the invalid transform receiver.

## Consequence surface

| Bun path | Plugin reports no dependencies | Plugin reports dependencies |
| --- | --- | --- |
| `Bun.build({ plugins: [ttsc()] })` | transform can complete | loader throws before returning transformed source |
| `Bun.plugin(ttsc())` | runtime TypeScript can load | runtime loader throws while importing the module |
| `bunfig.toml` preload | runtime TypeScript can load | same runtime loader failure |
| unchanged/cached transform with dependencies | no visible problem without notification | dependency notification still executes and can throw |

Any plugin may report dependencies; type-driven generators are the primary case because their output can depend on declarations erased from Bun's runtime graph. Both relative and absolute dependency entries converge on the same hook and therefore the same crash.

## Approach

The architectural owner is the Bun adapter's mapping from the generic transform contract to Bun's loader contract. The adapter must provide only capabilities that are valid for Bun and must not expose a callable path that relies on a missing unplugin context method. Preserve transformation, loader selection, errors, and dependency semantics to the extent Bun's current API supports them.

This issue does not prescribe whether the correct boundary is a Bun-native dependency channel, an explicit no-op with documented invalidation limits, or another adapter-owned mechanism. Verify Bun's supported API before choosing the implementation; the non-negotiable invariant is that a valid dependency list cannot crash the loader.

- Add an adapter-level regression using the existing dependency-emitting fixture and the captured Bun `onLoad` callback.
- Cover `dependencies` absent, empty, one relative entry, absolute entries, duplicates, and cache hits.
- Assert that valid dependencies do not cause a `TypeError` and that transform output and `ts`/`tsx` loader selection remain correct.
- If Bun exposes a dependency-registration contract, assert the exact registered paths and invalidation behavior. If it does not, document and test the chosen explicit boundary rather than silently relying on `this.addWatchFile`.
- Preserve plugin errors as errors; do not convert a dependency crash fix into swallowed transform failures.
- Run the focused adapter and dependency tests:

  ```bash
  pnpm --filter @ttsc/test-unplugin start -- --include=bun_adapter,plugin_dependencies
  ```

- Then run package build and the broader gate:

  ```bash
  pnpm --filter @ttsc/unplugin build
  pnpm test
  ```

## Coordination

- Campaign tracker: [#665](https://github.com/samchon/ttsc/issues/665).
- #47, #220, and #327 are merged lineage, not active duplicates.
- The current open PR [#657](https://github.com/samchon/ttsc/pull/657), “fix(test): audit every lint corpus source,” was checked by full file diff and does not touch `packages/unplugin` or its tests.
- Campaign issue [#666](https://github.com/samchon/ttsc/issues/666), “Register plugin dependencies from the Turbopack loader,” owns the same dependency side-channel at a different adapter boundary. Shared transform-core changes must keep both adapters consistent, but the Bun crash and Turbopack invalidation gap require separate adapter regressions.
- Campaign issue [#676](https://github.com/samchon/ttsc/issues/676), “Prevent Bun auto-registration from shadowing explicit register options,” also affects Bun runtime setup, but it owns registration order and option precedence, not the transform context. It can proceed independently.

### LINKED ISSUE #666: fix(unplugin): register plugin dependencies from the Turbopack loader
## Problem

The standalone `@ttsc/unplugin/turbopack` loader transforms TypeScript but drops the transform plugin's reported dependency list. Type-only declarations and other files consulted by a plugin therefore do not enter Turbopack's invalidation graph. Editing one of those inputs can leave the transformed module and HMR output stale until another event happens to rerun the loader.

The expected behavior is that every file reported in the transform envelope for a module participates in Turbopack's loader dependency tracking, with the same path normalization and deduplication guarantees used by the other unplugin adapters.

## Evidence

The stable owner is `turbopack` in `packages/unplugin/src/turbopack.ts`, with the shared dependency-selection contract in `packages/unplugin/src/core/transform.ts`.

- `transformTtsc` accepts a `TtscTransformHooks` argument and calls `addWatchFile` for the selected absolute dependency paths on both fresh transforms and cache hits.
- The generic unplugin factory supplies that hook through its bundler context.
- The Turbopack loader calls `transformTtsc(file, source, options, undefined, transformCache)` with no hooks argument.
- `TtscTurbopackLoaderContext` exposes only `async`, `resourcePath`, and `getOptions`; it does not model or invoke webpack loader context's `addDependency(file)` method.

The current Turbopack webpack-loader bridge was checked against this path. It carries loader-reported file dependencies into Turbopack's `fileDependencies` set, so the relevant invalidation channel exists at the loader boundary; the ttsc loader simply never uses it.

A deterministic loader-context probe uses the existing `emit-dependencies` fixture:

1. configure the fixture transform to report a consulted type-only file outside the transformed module;
2. invoke the exported Turbopack loader with a context that records dependency-registration calls and captures the async callback;
3. observe that transformed code reaches the callback; and
4. observe zero dependency registrations because `transformTtsc` received `hooks === undefined`.

The core dependency tests already prove the envelope and path selection:

```bash
pnpm --filter @ttsc/test-unplugin start -- --include=transformttsc_forwards_plugin_dependencies_to_the_watch_hook,transformttsc_notifies_watch_files_on_cached_transforms
```

The missing link is specifically the Turbopack adapter.

Closed issue [#214](https://github.com/samchon/ttsc/issues/214) and merged PR [#220](https://github.com/samchon/ttsc/pull/220) established plugin-reported dependency lists for bundler HMR. Closed issue [#215](https://github.com/samchon/ttsc/issues/215) and merged PR [#221](https://github.com/samchon/ttsc/pull/221) added the standalone Turbopack loader. The two features were not connected, and there is no active corrective tracker.

## Consequence surface

| Dependency location/state | Current invalidation | Required invalidation |
| --- | --- | --- |
| transformed module itself | Turbopack already tracks it | unchanged |
| type-only import/declaration consulted by plugin | can be absent from runtime module graph and is dropped by loader | registered for the owning transformed module |
| sibling project source reported by plugin | project cache may notice only when loader reruns | edit must cause loader rerun |
| `node_modules` declaration or sibling workspace outside project walk | not covered by transform cache hashing and dropped by loader | reported dependency must enter Turbopack tracking |
| duplicate/relative/absolute paths | core normalizes and deduplicates, but adapter receives none | register normalized unique paths |
| cache hit | core would notify again if hook existed | registration must be restored for each build/request context |

The visible impact is stale generated code and HMR behavior rather than an immediate exception. A clean restart or unrelated rebuild can mask the issue, which makes adapter-level invalidation tests important.

## Approach

The architectural owner is the Turbopack loader's adaptation of the shared transform side-channel to the webpack-compatible loader context that Turbopack implements. Add `addDependency(file)` to the modeled `TtscTurbopackLoaderContext` and pass a `transformTtsc` hook that invokes it for each normalized dependency. Keep registrations build/request-scoped so cached transforms restore them.

Do not infer a private Turbopack API or duplicate dependency parsing in the adapter. Verify the loader contract and use the shared `transformTtsc` hook boundary as the source of normalized paths.

- Extend the Turbopack loader context and adapter-level tests to record dependencies from the existing `emit-dependencies` fixture.
- Cover fresh transforms and cache hits, relative and absolute paths, duplicates, self-dependencies, no dependency field, and transform failure.
- Add an integration-level invalidation proof where changing only a plugin-reported/type-only dependency reruns the owning transform and changes output without editing the module itself.
- Preserve declaration-file and `node_modules` passthrough behavior and rule-option forwarding.
- Run focused tests:

  ```bash
  pnpm --filter @ttsc/test-unplugin start -- --include=turbopack_loader,plugin_dependencies
  ```

- Then run package build and the broader gate:

  ```bash
  pnpm --filter @ttsc/unplugin build
  pnpm test
  ```

## Coordination

- Campaign tracker: [#666](https://github.com/samchon/ttsc/issues/666).
- #214/#220 own the shared dependency protocol; #215/#221 own the Turbopack entrypoint. All are closed or merged and none fixes this missing adapter connection.
- The current open PR [#657](https://github.com/samchon/ttsc/pull/657), “fix(test): audit every lint corpus source,” has no `packages/unplugin` or Turbopack-loader changes in its actual diff.
- Campaign issue [#665](https://github.com/samchon/ttsc/issues/665), “Make the Bun adapter safe for plugin-reported dependencies,” is the closest related work. It addresses an invalid context crash in Bun, while this issue addresses a missing dependency channel in Turbopack. Coordinate shared core assumptions, but keep adapter-specific acceptance separate.

### LINKED ISSUE #672: fix(unplugin): evict failed transforms so long-lived adapters can recover
## Problem

`@ttsc/unplugin` stores the Promise for a project-wide transform in its per-build cache before that transform has succeeded. The entry is removed only when a resolved entry later fails its input-hash validation. If the Promise rejects, or if it resolves to an `ITtscCompilerTransformation` whose type is `"exception"` and `selectTransformedSource` throws, the failed entry remains in the cache.

That makes a transient toolchain, plugin-build, filesystem, or host failure permanent for the lifetime of a Metro or Turbopack worker. Restoring the environment and retrying the unchanged module immediately replays the old failure instead of attempting a new transform. A failed computation must not become the authoritative cached generation.

## Evidence

The stable owner is `packages/unplugin/src/core/transform.ts::transformTtsc` and the `TtscTransformCache` value contract.

The current control flow is:

1. Compute the project cache key.
2. Reuse `cache.get(key)` and immediately `await` it.
3. On a miss, call `transformProject(...)` and store its Promise with `cache.set(key, transformed)` before awaiting it.
4. Delete the entry only after a successfully awaited cached result fails `matchesCachedSource`.

There is no rejection or host-exception eviction path around either await.

A read-only reproduction used one shared cache and an otherwise valid unplugin fixture:

1. Create the cache with `createTtscTransformCache()`.
2. Set `TTSC_GO_BINARY` to a nonexistent executable.
3. Call `transformTtsc` for the fixture's TypeScript module. The project transform resolves to a host-exception result, `selectTransformedSource` throws, and the cache size becomes one.
4. Replace `TTSC_GO_BINARY` with the valid Go executable without changing the module, options, aliases, or tsconfig.
5. Call `transformTtsc` again with the same cache.

The second call throws the first call's cached exception immediately; the transform is not re-run and the cache remains size one. Replacing the first result with a rejected Promise proves the same retention through the earlier `await transformed` branch.

The behavior is repository-owned: `@ttsc/unplugin` creates the cache, decides when entries become reusable, and keeps that cache alive in its core adapters. Metro and Turbopack intentionally retain their transform cache for a worker lifetime.

## Consequence surface

- Long-lived Metro and Turbopack workers cannot recover after a transient compiler or source-plugin failure.
- The ordinary unplugin adapters can also replay the failure until their next `buildStart` clears the cache; recovery depends on a bundler lifecycle event rather than the failed operation.
- Concurrent callers correctly share one in-flight transform, but all later callers inherit a failed generation indefinitely.
- Both rejected Promises and resolved `"exception"` envelopes are affected.
- Successful transform caching and input-hash invalidation are not the problem and must remain shared and efficient.

## Approach

The owner is the cache lifecycle in `transformTtsc`. The invariant is that a cache entry becomes reusable only after the project transform has produced a result that the adapter accepts as successful. Any rejected or rejected-by-validation generation must be removed only if the entry still refers to that same generation, so cleanup cannot delete a newer replacement created by another caller.

Keep the existing single-flight behavior for concurrent successful work and preserve dependency reporting and input-hash validation. Do not weaken failures into unchanged source.

- A rejected project-transform Promise rejects every caller sharing that in-flight generation, then leaves no reusable failed entry.
- A resolved `ITtscCompilerTransformation` with `type: "exception"` is surfaced and is not retained.
- After either failure, correcting the environment and retrying the same module/options/cache invokes a fresh transform and can succeed.
- Cleanup from an older failed generation cannot remove a newer cache entry for the same key.
- Concurrent successful calls still compile the project once.
- Existing cache hits, source-hash invalidation, adapter dependency notifications, and `buildStart` clearing continue to work.
- Add narrow regression coverage beside the existing cache scenarios in `tests/test-unplugin/src/internal/transform-project-cache.ts` for rejected and resolved-exception results.
- Prove it with `pnpm --filter @ttsc/unplugin build`, `pnpm --filter @ttsc/test-unplugin start`, and the broader `pnpm test`.

## Coordination

PR #253 introduced and hardened the project transform cache, but no earlier open or closed issue or pull request was found that owns failed-generation eviction. This is separate from campaign issue #665, **Make the Bun adapter safe for plugin-reported dependencies**, which concerns Bun's missing watch-file context, and it should preserve that issue's dependency-channel contract.

The only current open pull request, #657 (**fix(test): audit every lint corpus source**), is confined to the lint corpus, fixtures, and lint test utilities. It does not touch `packages/unplugin/src/core/transform.ts` or any adapter cache.

### LINKED ISSUE #676: fix(unplugin): prevent Bun auto-registration from shadowing explicit register options
## Problem

Importing `@ttsc/unplugin/bun-register` under Bun automatically calls `register()` with default options. The same module also documents and exports `register(options)` for explicit configuration. Because module evaluation happens before the importing module can call that function, an explicit call always follows the default registration.

Bun uses the first matching `onLoad` hook and does not continue to a later matching plugin. The auto-registered default plugin can therefore handle every TypeScript module and prevent the explicitly configured plugin from ever applying its `project`, `plugins`, or compiler options.

The preload convenience and the explicit API currently cannot coexist truthfully.

## Evidence

The stable owners are `packages/unplugin/src/bun-register.ts::register` and the module-level auto-registration block:

```ts
if (bunRuntime() !== undefined) {
  register();
}
```

`register(options)` always creates a new Bun adapter and calls `Bun.plugin`. There is no deduplication, replacement, or distinction between preload import and an import made to access the explicit function.

The reproduction requires only a Bun-like global:

```ts
const registrations = [];
globalThis.Bun = { plugin: (plugin) => registrations.push(plugin) };

const { register } = await import("@ttsc/unplugin/bun-register");
register({ project: "tsconfig.explicit.json" });

console.log(registrations.length); // 2
```

The first captured plugin was created with defaults during module evaluation; the second carries the explicit project. Both register the same TypeScript `onLoad` filter. Bun's first-match behavior means the first hook shadows the second. Existing tests import the entry off Bun first and install a Bun-like global only before the explicit call, so they prove each mode separately but miss the real same-runtime sequence.

This behavior arrived with the Bun runtime registration work in PR #327. It also matches the limitation tracked upstream in [Bun issue #20583](https://github.com/oven-sh/bun/issues/20583): later matching loaders do not compose after the first handles a module.

## Consequence surface

- `register({ project })` can silently use the nearest/default tsconfig instead of the requested one.
- Explicit `plugins` and compiler-option overlays can be ignored, producing untransformed or incorrectly transformed runtime modules.
- `bun run` and `bun test` are affected; the Bun bundler adapter itself is a separate entrypoint.
- A pure preload import with no explicit options must continue to auto-register once.
- Importing under Node must remain a harmless no-op, while explicitly calling `register` off Bun must keep its clear error.
- Repeated or mixed registration needs a defined ownership rule so behavior is deterministic rather than order-dependent.

## Approach

The owner is the public entrypoint contract for Bun runtime registration. The invariant is that the first effective TypeScript loader uses the caller's intended configuration: preload-only use gets defaults, and explicit use gets its supplied options without an earlier implicit loader shadowing it.

Preserve a one-line preload path and a callable explicit API, but make their activation modes unambiguous. Do not rely on Bun invoking a later overlapping hook, because upstream behavior explicitly does not guarantee that composition.

- Preloading the documented side-effect entry under Bun registers exactly one default plugin.
- Using the documented explicit API under Bun registers one effective plugin with the supplied `project`, `plugins`, and compiler options.
- Accessing the explicit API cannot first install a shadowing default plugin.
- A test exercises the actual same-runtime import/evaluation order with `globalThis.Bun` present before module import.
- Node import remains a no-op and an explicit off-Bun call retains its diagnostic.
- Real Bun runtime coverage proves the explicit project changes which tsconfig/plugin configuration transforms a fixture.
- Bun bundler adapter behavior is unchanged.
- Extend `tests/test-unplugin/src/internal/adapter-bun-register.ts` for the same-runtime import order, add a real-Bun fixture for option selection, and prove the package with `pnpm --filter @ttsc/unplugin build`, `pnpm --filter @ttsc/test-unplugin start`, the same `pnpm run experimental:unplugin` command used by `.github/workflows/bun.yml`, and the broader `pnpm test`.

## Coordination

PR #327 introduced this runtime entrypoint, and [Bun issue #20583](https://github.com/oven-sh/bun/issues/20583) documents the first-matching-loader constraint that makes duplicate registration observable. This issue is related to campaign issue #665, **Make the Bun adapter safe for plugin-reported dependencies**: both touch the Bun adapter family, but #665 owns watch-file dependency handling while this one owns registration order and configuration.

The only current open pull request, #657 (**fix(test): audit every lint corpus source**), contains only lint corpus and harness changes and does not overlap `packages/unplugin/src/bun-register.ts`.

### LINKED ISSUE #679: fix(unplugin): declare the external ttsc host in the published dependency contract
## Problem

The published `@ttsc/unplugin` runtime imports `ttsc`, and its Rollup build deliberately leaves that host external. The package manifest declares `ttsc` only as a development dependency, so a clean consumer can install `@ttsc/unplugin` successfully without installing the module every adapter needs.

The first import then fails with Node's generic `Cannot find module 'ttsc'` before the user can configure a bundler or receive an adapter-specific diagnostic. A required external runtime host must be represented in the published dependency contract.

## Evidence

The stable owners are:

- `packages/unplugin/src/core/transform.ts`, which imports `TtscCompiler` and ttsc types from `"ttsc"`.
- `packages/unplugin/src/core/options.ts`, whose public option types import `ITtscProjectPluginConfig` from `"ttsc"`.
- `packages/unplugin/rollup.config.mjs`, which explicitly includes `"ttsc"` in `externalPackages` and describes it as supplied by the consuming project.
- `packages/unplugin/package.json`, which lists only `unplugin` under dependencies and `ttsc` under devDependencies.

The production bundle therefore contains a live external `require("ttsc")` or ESM import, while npm metadata gives the package manager no runtime relationship to install, validate, or warn about.

The published `0.18.4` package reproduces in a clean directory:

```sh
npm init -y
npm install @ttsc/unplugin@0.18.4
node -e "require('@ttsc/unplugin')"
node -e "require('@ttsc/unplugin/vite')"
```

Installation succeeds with `unplugin` as the only directly declared runtime dependency. Both imports exit 1 with `Cannot find module 'ttsc'`. Workspace tests mask this because their fixture packages install workspace `ttsc` directly.

PR #47 established the external-host design. Commit `b84aaebead1657ecc61f3b19ff17de2d9e14ebe4` later made the externalization explicit while deriving Rollup externals from the manifest, but it did not add `ttsc` to published dependency metadata.

## Consequence surface

- Every entrypoint that reaches the core transform fails in a standalone install, including Vite, Rollup, Rolldown, webpack, Rspack, esbuild, Farm, Next, Metro through its unplugin dependency, Bun, and Turbopack.
- The failure occurs at module load, before project discovery or configuration.
- npm, pnpm, and Yarn may treat peer/dependency metadata differently; the package contract must be intentional across supported managers.
- `ttsc` should remain external to the adapter bundle so the consumer does not receive a hidden duplicate host.
- Type declarations also reference `ttsc` and need resolvable types under the same installation contract.

## Approach

`packages/unplugin/package.json` owns the published relationship, while the Rollup config owns bundling. The invariant is that a supported installation either provides a compatible `ttsc` host through declared package metadata or emits a clear package-manager/adapter diagnostic before loading core code. The runtime and type-level host must resolve to the same consumer-visible package.

Keep `ttsc` external and avoid bundling a private copy. Select dependency semantics that match whether the host is consumer-controlled, then verify them through packed-package installs rather than workspace resolution.

- A clean supported install of `@ttsc/unplugin` has a resolvable compatible `ttsc` host, or the package manager reports the missing required relationship explicitly.
- Requiring/importing the package root and representative CJS/ESM adapter subpaths no longer fails with an undeclared-module error.
- The built output continues to externalize `ttsc` and contains no second bundled compiler host.
- Published declarations resolve without consumers manually adding an undocumented dependency.
- Add a packed-host-contract regression at `tests/test-unplugin/src/internal/packaged-host-contract.ts` (or the neighboring package-contract suite if one is introduced first) covering npm and pnpm installation outside the workspace. Build and pack with `pnpm --filter @ttsc/unplugin build` and `pnpm --dir packages/unplugin pack --out <tmp>/ttsc-unplugin.tgz`, then install that tarball in clean npm and pnpm fixtures with no workspace links.
- Existing adapter, bundler, and externalization regression tests continue to pass.
- Prove the normal suite with `pnpm --filter @ttsc/test-unplugin start` and the broader `pnpm test`.

## Coordination

PR #47 and commit `b84aaebe` are design provenance, not corrective owners for the manifest gap. This issue is related to campaign issue #672, **Evict failed unplugin transforms so long-lived adapters can recover**, because both affect every adapter through the unplugin core, but dependency installation and cache recovery are independent.

The only current open pull request, #657 (**fix(test): audit every lint corpus source**), changes lint corpus fixtures and utilities only; it does not touch `packages/unplugin/package.json`, its Rollup config, or transform imports.

### LINKED ISSUE #686: fix(ttsc): resolve npm tsconfig presets through package.json#tsconfig
## Problem

TypeScript permits a bare `extends` specifier to name an npm tsconfig preset whose `package.json` selects a non-default config file through its `tsconfig` field. Both of ttsc's JavaScript-side config readers approximate bare resolution with Node's JavaScript entrypoint resolver and a `<specifier>.json` fallback. They do not honor `package.json#tsconfig`.

A TypeScript-valid project can therefore fail inside ttsc before native compilation. In the duplicated unplugin paths reader, the same preset is skipped best-effort and its inherited aliases silently disappear from the generated transform config.

## Evidence

There are two stable owners with the same missing case:

- `resolveExtendsConfig` in `packages/ttsc/src/compiler/internal/project/readProjectConfig.ts` calls `createRequire(tsconfig).resolve(specifier)`, then retries `resolver.resolve(`${specifier}.json`)`; failure propagates.
- `resolveExtendsConfig` in `packages/unplugin/src/core/tsconfigPaths.ts` performs the same two attempts and returns `null`; `findDeclaredPaths` then continues without the base config.

Create this local preset without a root JavaScript/JSON entrypoint:

```json
// node_modules/example-preset/package.json
{
  "name": "example-preset",
  "version": "1.0.0",
  "tsconfig": "base.json"
}
```

```jsonc
// node_modules/example-preset/base.json
{
  "compilerOptions": {
    "strict": true,
    "paths": { "#preset/*": ["./src/*"] }
  }
}
```

```json
// tsconfig.json
{ "extends": "example-preset", "compilerOptions": { "noEmit": true } }
```

`tsc --showConfig -p tsconfig.json` succeeds and includes the inherited options. A ttsc project load reaches the two Node-resolution attempts and fails with `MODULE_NOT_FOUND` for the package / appended `.json` candidate before the native compiler runs. The unplugin helper returns no inherited `paths`, so an alias such as `#preset/value` is absent from its overlay even though TypeScript accepts the project.

## Consequence surface

The core reader supplies effective compiler options, plugin inheritance, path-option anchoring, project identity consumers, CLI launch preparation, and in-memory compiler APIs. The unplugin reader independently preserves inherited `paths` when it builds a temporary transform tsconfig. Leaving either owner unfixed produces inconsistent behavior between CLI/runtime and bundler adapters.

Existing relative and absolute extends resolution, direct package JSON entrypoints, package subpaths, extends arrays and precedence, JSONC/BOM handling, symlink canonicalization, circular-chain detection, declaring-config path anchors, and best-effort unplugin diagnostics must remain intact.

## Approach

Treat TypeScript's accepted config-resolution contract as the invariant for both owners: a bare package root must resolve the preset selected by that package's manifest, including `package.json#tsconfig`, while explicit package subpaths and file paths retain their current meaning.

Determine the common resolution semantics first, then keep the core project reader and focused unplugin reader behaviorally aligned. This issue does not prescribe exporting an internal helper or adding a dependency; PR #217 deliberately kept the focused unplugin reader private, and that package boundary should change only with separate justification.

- A bare package with `package.json#tsconfig` pointing to a non-default JSON/JSONC file loads in ttsc and contributes inherited options/plugins according to existing merge rules.
- The unplugin reader preserves inherited `compilerOptions.paths` from the same preset and anchors relative targets at the preset config's directory.
- Cover a direct package JSON entrypoint, a package subpath, relative and absolute extends, multiple extends precedence, symlinked workspaces, and circular chains to prevent regressions.
- Cover missing or malformed manifest targets with diagnostics consistent with each owner's contract: core fails visibly, while unplugin remains best-effort without inventing aliases.
- Include an integration fixture where TypeScript accepts the preset and both ttsc and an unplugin transform resolve a type through its inherited path alias.
- Extend `test_readprojectconfig_resolves_package_tsconfig_extends` for the manifest-selected root and add a matching unplugin case named with `package_tsconfig`. Run them with `pnpm --filter @ttsc/test-ttsc start -- --include=readprojectconfig_resolves_package_tsconfig_extends` and `pnpm --filter @ttsc/test-unplugin start -- --include=package_tsconfig`.
- Then run `pnpm build` and `pnpm test`.

## Coordination

The core resolver comes from the project-config lineage. Merged PR #217 added the focused unplugin reader and explicitly mirrored ttsc's internal resolution behavior to fix silently lost aliases; it did not cover `package.json#tsconfig` presets.

Open PR #657, `fix(test): audit every lint corpus source`, changes lint-corpus fixtures and harness logic only and does not touch either config reader.

[Issue #680](https://github.com/samchon/ttsc/issues/680), **Preserve .json when @ttsc/paths rewrites aliases to emitted files**, also concerns path aliases, but it owns emitted-specifier rewriting after configuration has already resolved. This issue owns tsconfig preset discovery and inheritance.

samchon:
## Bundle b4 — unplugin adapters, cache, packaging + tsconfig preset

Part of the `repository-audit-2026-07` campaign (tracker #665). This batch fixes six independently-verified defects across the `@ttsc/unplugin` adapter family and the shared tsconfig-preset resolution used by both the core project reader and the unplugin paths reader.

### Issues

- Closes #665 (RA-04) — **Bun adapter safe for plugin-reported dependencies.** The Bun `onLoad` path invoked the shared transform with an empty receiver, so `this.addWatchFile` was `undefined` and any plugin reporting dependencies crashed the loader with `TypeError`. Bun exposes no per-module dependency channel, so the adapter now passes an explicit no-op `addWatchFile` context: a valid dependency list can never crash the loader, while transform output and `ts`/`tsx` loader selection are preserved.
- Closes #666 (RA-05) — **Register plugin dependencies from the Turbopack loader.** The Turbopack loader dropped the transform's reported dependency list. It now models the webpack-compatible `addDependency(file)` on the loader context and forwards each normalized dependency through the shared `transformTtsc` hook (fires on cache hits too), so type-only inputs enter Turbopack's `fileDependencies` invalidation set.
- Closes #686 (RA-16) — **Resolve npm tsconfig presets through `package.json#tsconfig`.** Both config readers (`readProjectConfig` core, `tsconfigPaths` unplugin) now honor a bare preset's manifest-selected config file, matching TypeScript's contract. Core fails visibly on a missing/malformed manifest target; the unplugin reader stays best-effort and never invents aliases.
- Closes #672 (RA-19) — **Evict failed transforms so long-lived adapters can recover.** A rejected transform Promise or a resolved `"exception"`/`"failure"` envelope no longer stays cached: the generation is evicted (identity-guarded so a newer replacement survives) so a corrected environment re-runs the transform. Single-flight success caching and input-hash invalidation are unchanged.
- Closes #676 (RA-23) — **Prevent Bun auto-registration from shadowing explicit register options.** `bun-register` now registers exactly one Bun loader whose effective options resolve lazily on first load, so an explicit `register(options)` after the import-time auto-register wins instead of being shadowed by a default loader (Bun uses the first matching `onLoad`; oven-sh/bun#20583).
- Closes #679 (RA-26) — **Declare the external `ttsc` host in the published dependency contract.** `@ttsc/unplugin` imports `ttsc` at runtime but declared it only as a dev dependency, so a clean install failed with `Cannot find module 'ttsc'`. `ttsc` is now a required `peerDependency`, keeping the host external and consumer-controlled while making the relationship visible to npm/pnpm/Yarn.

### Verification

Package-scoped builds and focused suites are listed in the PR comments as each increment lands.

### DIFF
--- a/packages/ttsc/src/compiler/internal/project/readProjectConfig.ts
+++ b/packages/ttsc/src/compiler/internal/project/readProjectConfig.ts
@@ -254,11 +254,67 @@ function resolveExtendsConfig(tsconfig: string, specifier: string): string {
     return resolveExistingExtendsPath(path.resolve(baseDir, specifier));
   }
   const resolver = createRequire(tsconfig);
+  // A bare package root selects its preset through `package.json#tsconfig`,
+  // matching TypeScript's config resolution. Presets shipped this way often
+  // have no JavaScript/JSON entrypoint at all, so Node's entrypoint resolver
+  // and the `<specifier>.json` fallback below both miss them.
+  const viaManifest = resolvePackageManifestTsconfig(resolver, specifier);
+  if (viaManifest !== undefined) {
+    return resolveExistingExtendsPath(viaManifest);
+  }
   try {
     return resolveRealPath(resolver.resolve(specifier));
   } catch {
     return resolveRealPath(resolver.resolve(`${specifier}.json`));
   }
+}
+
+/**
+ * When `specifier` names a bare package root, resolve the config file its
+ * `package.json#tsconfig` field selects (anchored at the package directory).
+ * Returns `undefined` when the specifier is a subpath, the manifest is
+ * unresolvable/unparsable, or it declares no `tsconfig` field, so the caller
+ * falls back to Node entrypoint resolution.
+ */
+function resolvePackageManifestTsconfig(
+  resolver: NodeRequire,
+  specifier: string,
+): string | undefined {
+  if (!isBarePackageRoot(specifier)) {
+    return undefined;
+  }
+  let manifestPath: string;
+  try {
+    manifestPath = resolver.resolve(`${specifier}/package.json`);
+  } catch {
+    return undefined;
+  }
+  let field: unknown;
+  try {
+    field = (
+      JSON.parse(stripLeadingBom(fs.readFileSync(manifestPath, "utf8"))) as {
+        tsconfig?: unknown;
+      }
+    ).tsconfig;
+  } catch {
+    return undefined;
+  }
+  if (typeof field !== "string" || field.length === 0) {
+    return undefined;
+  }
+  return path.resolve(path.dirname(manifestPath), field);
+}
+
+/**
+ * Return true when `specifier` is a bare package root (no subpath): a plain
+ * package name (`preset`) or a scoped name (`@scope/preset`). Subpaths such as
+ * `@scope/preset/base.json` resolve directly and keep their current meaning.
+ */
+function isBarePackageRoot(specifier: string): boolean {
+  if (specifier.startsWith("@")) {
+    return specifier.split("/").length === 2;
+  }
+  return !specifier.includes("/");
 }
 
 /**
--- a/packages/unplugin/package.json
+++ b/packages/unplugin/package.json
@@ -97,6 +97,9 @@
   "dependencies": {
     "unplugin": "^2.3.11"
   },
+  "peerDependencies": {
+    "ttsc": "workspace:*"
+  },
   "devDependencies": {
     "@rollup/plugin-commonjs": "^29.0.2",
     "@rollup/plugin-node-resolve": "^16.0.3",
--- a/packages/unplugin/src/bun-register.ts
+++ b/packages/unplugin/src/bun-register.ts
@@ -13,6 +13,21 @@ interface BunRuntimeGlobal {
 }
 
 /**
+ * Effective options for the single runtime loader.
+ *
+ * Bun uses the first matching `onLoad` hook and does not fall through to a
+ * later overlapping plugin (oven-sh/bun#20583). Registering twice — once
+ * implicitly on import, once explicitly — would let the default loader shadow
+ * the configured one. Instead exactly one Bun plugin is registered; it resolves
+ * these options lazily on first load (see {@link bun}'s provider form), so an
+ * explicit `register(options)` made right after importing this module overrides
+ * the preload defaults without a second shadowing registration.
+ */
+let activeOptions: TtscUnpluginOptions | undefined;
+/** Whether the single runtime loader has already been registered with Bun. */
+let registered = false;
+
+/**
  * Register the ttsc transform as a Bun **runtime** plugin.
  *
  * The other `@ttsc/unplugin/*` adapters cover bundlers (`Bun.build`, Vite,
@@ -23,6 +38,12 @@ interface BunRuntimeGlobal {
  * ["@ttsc/unplugin/bun-register"]` — or imperatively with `import
  * "@ttsc/unplugin/bun-register"`. Options are read from the nearest
  * `tsconfig.json`, identical to the bundler adapters.
+ *
+ * Registration is idempotent: the first call (implicit on import, or explicit)
+ * registers the one loader with Bun; later calls only update the effective
+ * options, so accessing the explicit API cannot install a second default loader
+ * that shadows the caller's configuration. Repeated explicit calls are
+ * last-write-wins for the effective options.
  *
  * @throws When called explicitly off the Bun runtime (`globalThis.Bun.plugin`
  *   is unavailable). The auto-registration below stays silent off Bun so the
@@ -37,7 +58,15 @@ export function register(options?: TtscUnpluginOptions): void {
         "@ttsc/unplugin/vite for non-Bun toolchains.",
     );
   }
-  runtime.plugin(bun(options));
+  activeOptions = options;
+  if (registered) {
+    return;
+  }
+  registered = true;
+  // Register a single loader whose options are read from `activeOptions` on
+  // first load, so an explicit call made after the import-time auto-register
+  // still wins.
+  runtime.plugin(bun(() => activeOptions));
 }
 
 function bunRuntime(): BunRuntimeGlobal | undefined {
--- a/packages/unplugin/src/bun.ts
+++ b/packages/unplugin/src/bun.ts
@@ -23,6 +23,44 @@ export interface BunLikePlugin {
 export type BunLoader = "ts" | "tsx";
 
 /**
+ * Options accepted by {@link bun}, either resolved eagerly or supplied through a
+ * provider evaluated lazily on the first `onLoad` call.
+ *
+ * The provider form exists for the runtime registration path
+ * (`bun-register`), where a single Bun plugin is registered on import but its
+ * effective options may be overridden by an explicit `register(options)` call
+ * made in the same synchronous tick. Resolving through the provider on first
+ * load, rather than at registration, lets that later call win without Bun ever
+ * seeing a second shadowing loader.
+ */
+export type TtscBunOptions =
+  | TtscUnpluginOptions
+  | (() => TtscUnpluginOptions | undefined);
+
+/**
+ * Transform context handed to the raw unplugin transform under Bun.
+ *
+ * The shared transform calls `addWatchFile` once per plugin-reported dependency
+ * so type-only inputs can enter a bundler's watch graph. Bun's bundler and
+ * runtime loaders expose no per-module dependency-registration channel, so the
+ * hook is an explicit no-op here: reported dependencies cannot participate in
+ * Bun invalidation, but a valid dependency list must never crash the loader by
+ * reaching a missing context method. Passing an empty object made
+ * `this.addWatchFile` `undefined`, so any plugin reporting dependencies threw
+ * `TypeError: this.addWatchFile is not a function`.
+ */
+const bunTransformContext = {
+  addWatchFile(): void {},
+};
+
+/** Resolve {@link TtscBunOptions} to a plain options object (or `undefined`). */
+function resolveBunOptions(
+  options?: TtscBunOptions,
+): TtscUnpluginOptions | undefined {
+  return typeof options === "function" ? options() : options;
+}
+
+/**
  * Minimal subset of the Bun `BuildConfig` plugin build object.
  *
  * Only the `onLoad` hook is used; other hooks are not needed for a
@@ -61,17 +99,31 @@ export interface BunLikeBuild {
  * transpiling the emitted TypeScript at runtime; `sourceFilePattern` only
  * matches TypeScript, so the loader is always `ts`/`tsx`.
  */
-export default function bun(options?: TtscUnpluginOptions): BunLikePlugin {
+export default function bun(options?: TtscBunOptions): BunLikePlugin {
   return {
     name: "ttsc-unplugin",
     setup(build) {
-      const raw = unplugin.raw(options, {} as UnpluginContextMeta);
+      // Build the raw transform lazily on first load rather than in `setup`.
+      // Bun runs `setup` synchronously when the plugin is registered, so a
+      // runtime registration (bun-register) that resolves its effective options
+      // through a provider must defer that resolution until after any explicit
+      // `register(options)` call in the same tick. Deferring also keeps a single
+      // transform (and its project cache) shared across every loaded module.
+      let raw: ReturnType<typeof unplugin.raw> | undefined;
       build.onLoad({ filter: sourceFilePattern }, async (args) => {
+        raw ??= unplugin.raw(
+          resolveBunOptions(options),
+          {} as UnpluginContextMeta,
+        );
         const loader = bunLoaderFor(args.path);
         const source = await fs.readFile(args.path, "utf8");
         const result =
           typeof raw.transform === "function"
-            ? await raw.transform.call({} as never, source, args.path)
+            ? await raw.transform.call(
+                bunTransformContext as never,
+                source,
+                args.path,
+              )
             : undefined;
         // Unpack both shorthand string and object result shapes.
         if (typeof result === "string") {
--- a/packages/unplugin/src/core/transform.ts
+++ b/packages/unplugin/src/core/transform.ts
@@ -147,10 +147,14 @@ export async function transformTtsc(
 
   let transformed = cache?.get(key);
   if (transformed !== undefined) {
-    const cached = await transformed;
+    // A rejected in-flight generation must not stay cached: evict it (only if
+    // it is still the current entry) so a later call re-runs the transform.
+    const cached = await awaitOrEvict(cache, key, transformed);
     if (matchesCachedSource(cached, file, source)) {
       reportSuccessDiagnostics(cached.result);
-      const code = selectTransformedSource({
+      // A resolved `"exception"` / `"failure"` envelope makes this throw; that
+      // is a failed generation too, so evict before surfacing it.
+      const code = selectOrEvict(cache, key, transformed, {
         file,
         projectRoot: cached.projectRoot,
         result: cached.result,
@@ -177,11 +181,79 @@ export async function transformTtsc(
     });
     cache?.set(key, transformed);
   }
-  const { projectRoot, result } = await transformed;
+  const generation = transformed;
+  const { projectRoot, result } = await awaitOrEvict(cache, key, generation);
   reportSuccessDiagnostics(result);
-  const code = selectTransformedSource({ file, projectRoot, result });
+  const code = selectOrEvict(cache, key, generation, {
+    file,
+    projectRoot,
+    result,
+  });
   notifyFileDependencies(hooks, { file, projectRoot, result });
   return createTransformResult(source, code);
+}
+
+/**
+ * Await a cached generation, evicting it on rejection.
+ *
+ * The cache stores the in-flight transform Promise before it settles so
+ * concurrent callers share one compilation. A rejected generation must not
+ * remain the authoritative cached result, or a transient toolchain/host
+ * failure becomes permanent for a long-lived worker. Eviction is
+ * identity-guarded so a newer generation another caller installed under the
+ * same key survives.
+ */
+async function awaitOrEvict(
+  cache: TtscTransformCache | undefined,
+  key: string,
+  generation: Promise<TtscCachedProjectTransform>,
+): Promise<TtscCachedProjectTransform> {
+  try {
+    return await generation;
+  } catch (error) {
+    evictGeneration(cache, key, generation);
+    throw error;
+  }
+}
+
+/**
+ * Extract the transformed source, evicting the generation when the result is a
+ * host `"exception"` or compiler `"failure"` (which makes
+ * {@link selectTransformedSource} throw). Such a failed generation must not be
+ * replayed to later callers of an unchanged module.
+ */
+function selectOrEvict(
+  cache: TtscTransformCache | undefined,
+  key: string,
+  generation: Promise<TtscCachedProjectTransform>,
+  props: {
+    file: string;
+    projectRoot: string;
+    result: ITtscCompilerTransformation;
+  },
+): string {
+  try {
+    return selectTransformedSource(props);
+  } catch (error) {
+    evictGeneration(cache, key, generation);
+    throw error;
+  }
+}
+
+/**
+ * Delete a failed generation from the cache only when it is still the entry
+ * stored under `key`. The identity check prevents an older failed generation's
+ * cleanup from removing a newer replacement created by another caller for the
+ * same key.
+ */
+function evictGeneration(
+  cache: TtscTransformCache | undefined,
+  key: string,
+  generation: Promise<TtscCachedProjectTransform>,
+): void {
+  if (cache?.get(key) === generation) {
+    cache.delete(key);
+  }
 }
 
 /**
--- a/packages/unplugin/src/core/tsconfigPaths.ts
+++ b/packages/unplugin/src/core/tsconfigPaths.ts
@@ -148,6 +148,15 @@ function resolveExtendsConfig(
     );
   }
   const resolver = createRequire(tsconfig);
+  // A bare package root selects its preset through `package.json#tsconfig`,
+  // matching TypeScript's config resolution and the core project reader. Such
+  // presets often ship no JS/JSON entrypoint, so Node's entrypoint resolver and
+  // the `<specifier>.json` fallback both miss them, silently dropping the
+  // preset's inherited `paths`.
+  const viaManifest = resolvePackageManifestTsconfig(resolver, specifier);
+  if (viaManifest !== null) {
+    return viaManifest;
+  }
   try {
     return resolveRealPath(resolver.resolve(specifier));
   } catch {
@@ -157,6 +166,58 @@ function resolveExtendsConfig(
       return null;
     }
   }
+}
+
+/**
+ * When `specifier` names a bare package root, resolve the config file its
+ * `package.json#tsconfig` field selects (anchored at the package directory).
+ * Best-effort: returns `null` for a subpath, an unresolvable/unparsable
+ * manifest, a missing `tsconfig` field, or a field target that does not exist —
+ * the compiler owns the real diagnostic, and this reader must not invent
+ * aliases.
+ */
+function resolvePackageManifestTsconfig(
+  resolver: NodeRequire,
+  specifier: string,
+): string | null {
+  if (!isBarePackageRoot(specifier)) {
+    return null;
+  }
+  let manifestPath: string;
+  try {
+    manifestPath = resolver.resolve(`${specifier}/package.json`);
+  } catch {
+    return null;
+  }
+  let field: unknown;
+  try {
+    const text = fs.readFileSync(manifestPath, "utf8");
+    field = (
+      JSON.parse(text.charCodeAt(0) === 0xfeff ? text.slice(1) : text) as {
+        tsconfig?: unknown;
+      }
+    ).tsconfig;
+  } catch {
+    return null;
+  }
+  if (typeof field !== "string" || field.length === 0) {
+    return null;
+  }
+  return resolveExistingExtendsPath(
+    path.resolve(path.dirname(manifestPath), field),
+  );
+}
+
+/**
+ * Return true when `specifier` is a bare package root (no subpath): a plain
+ * package name (`preset`) or a scoped name (`@scope/preset`). Subpaths such as
+ * `@scope/preset/base.json` resolve directly and keep their current meaning.
+ */
+function isBarePackageRoot(specifier: string): boolean {
+  if (specifier.startsWith("@")) {
+    return specifier.split("/").length === 2;
+  }
+  return !specifier.includes("/");
 }
 
 /**
--- a/packages/unplugin/src/turbopack.ts
+++ b/packages/unplugin/src/turbopack.ts
@@ -21,6 +21,14 @@ export interface TtscTurbopackLoaderContext {
   resourcePath: string;
   /** The rule's `options` object, when one was configured. */
   getOptions?(): TtscUnpluginOptions | undefined;
+  /**
+   * Register an additional file the transformed module depends on. Part of the
+   * webpack loader context contract Turbopack implements; a registered file
+   * enters Turbopack's `fileDependencies` set so editing it re-runs this loader
+   * for the owning module. Optional so a minimal stub context (or a Turbopack
+   * build that predates the method) still loads.
+   */
+  addDependency?(file: string): void;
 }
 
 /** Matches any path segment that is a `node_modules` directory (cross-platform). */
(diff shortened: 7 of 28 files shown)


samchon:
## Self-Review — Round 1

Reviewed all six fixes and their tests against the issue specs, the development skill (Forbidden, Testing contract, Consequence Analysis), and the shared transform seam.

### Findings & remediation

1. **RA-26 packed-manifest test: `tar` remote-host misparse.** The extraction passed an absolute Windows tarball path (`C:\...tgz`) to `tar -xzOf`, which GNU tar interpreted as a `host:path` remote spec (`Cannot connect to C:`). **Fixed** by running `tar` with `cwd` set to the pack destination and a relative tarball name, so a drive-letter colon never appears in the `-f` argument. Cross-platform (bsdtar/GNU) and offline.

2. **RA-26 lockfile consequence.** Adding `peerDependencies.ttsc` could have broken CI's `pnpm install --frozen-lockfile`. **Verified**: `pnpm install --lockfile-only` produces no content change (pnpm does not record a `workspace:*` peer satisfied by a workspace link in the importer), and `--frozen-lockfile` passes against the original lockfile. No lockfile edit committed.

3. **Whole-class consequence — Metro.** `packages/metro` also consumes the shared `transformTtsc`, but it calls it with no hooks argument, so `notifyFileDependencies` short-circuits and there is **no RA-04-style crash**; Metro recovers from failed transforms transitively through the RA-19 core cache fix. Metro's own out-of-walk dependency registration is a separate potential improvement outside this bundle's surface (`packages/metro`, not `packages/unplugin`) — flagged for the lead, not changed here.

4. **RA-05 real-Turbopack invalidation.** A live-Turbopack rerun proof is not deterministic in CI. The tests instead pin the loader's contribution deterministically: `addDependency` receives the normalized/deduped/self-excluded paths on both the fresh transform and the cache hit, plus a no-report negative twin and a missing-`addDependency`-context boundary. In-walk type-only invalidation is already covered by the existing project-cache tests.

### Verified
- `tsc --noEmit` for `tests/test-unplugin` and `tests/test-ttsc`: clean.
- Each new test is transformation-direction: it fails against the pre-fix code (Bun `TypeError`, dropped Turbopack deps, retained failed cache entry, double Bun registration, missing peer host, dropped preset paths → "Cannot find module" instead of "not assignable").
- No website/README doc change needed: the unplugin README already documents installing `ttsc` first (RA-26 makes the metadata enforce it); the other fixes restore intended behavior.

Round 2 will re-run the focused suites for a clean confirmation.

samchon:
## Self-Review — Round 2 (clean)

No new findings. Re-ran the focused suites; all green with no regressions to existing adapter/dependency/cache tests.

### Verification commands & results

Builds:
```
pnpm --filter ttsc build            # tsc → lib (RA-16 core reader)
pnpm --filter @ttsc/unplugin build  # tsc --emitDeclarationOnly + rollup
```
Both succeed.

Typecheck (CI `test:typecheck` equivalent):
```
tsc --noEmit -p tests/test-unplugin/tsconfig.json   # clean
tsc --noEmit -p tests/test-ttsc/tsconfig.json       # clean
```

Focused suites:
```
pnpm --filter @ttsc/test-ttsc start -- --include=package_tsconfig
  ✓ test_readprojectconfig_resolves_package_tsconfig_extends
  ✓ test_readprojectconfig_resolves_package_tsconfig_extends_via_manifest   (RA-16)
  ✓ test_readprojectconfig_rejects_missing_package_tsconfig_manifest_target (RA-16 negative twin)

pnpm --filter @ttsc/test-unplugin start -- --include=bun_adapter,bun_register,turbopack_loader,dependencies,transformttsc_caches,transformttsc_invalidates,evict,eviction,shares_one,survives,packaged_manifest,package_tsconfig_preset
  ✓ test_bun_adapter_survives_plugin_reported_dependencies                  (RA-04)
  ✓ test_turbopack_loader_registers_plugin_dependencies_as_file_dependencies (RA-05)
  ✓ test_turbopack_loader_registers_plugin_dependencies_on_cache_hit         (RA-05)
  ✓ test_turbopack_loader_registers_no_dependencies_without_a_report         (RA-05)
  ✓ test_turbopack_loader_transforms_without_an_add_dependency_context       (RA-05)
  ✓ test_transformttsc_evicts_a_rejected_transform_and_recovers              (RA-19)
  ✓ test_transformttsc_evicts_a_host_exception_transform_and_recovers        (RA-19)
  ✓ test_transformttsc_eviction_keeps_a_newer_generation_for_the_same_key    (RA-19)
  ✓ test_transformttsc_shares_one_compile_across_concurrent_callers          (RA-19)
  ✓ test_bun_register_explicit_options_are_not_shadowed_in_same_runtime_order (RA-23)
  ✓ test_bun_register_preload_only_registers_a_single_default_plugin          (RA-23)
  ✓ test_packaged_manifest_declares_the_external_ttsc_host                    (RA-26)
  ✓ test_transformttsc_alias_overlay_resolves_package_tsconfig_preset_paths   (RA-16 unplugin)
  + all pre-existing bun/turbopack/dependency/cache/invalidation tests still pass
```

All 26 selected unplugin cases and all 4 core `package_tsconfig` cases pass.

### Notes for the lead
- No lockfile change: `peerDependencies.ttsc: workspace:*` does not dirty `pnpm-lock.yaml`, and `--frozen-lockfile` passes.
- Two `alias_overlay` tests that depend on a built `@ttsc/banner` fail only when that package's `lib/` is absent in a worktree (environmental); they are unrelated to this bundle and pass once `@ttsc/banner` is built.
- `packages/metro` is out of this bundle's surface; it has no RA-04 crash and gains RA-19 recovery transitively. A Metro out-of-walk dependency-registration enhancement (parallel to RA-05) is a possible independent follow-up.
- Real-Bun runtime coverage for RA-23 (`bun.yml` / `experimental:unplugin`) and real-Turbopack invalidation for RA-05 run in dedicated CI jobs, not the deterministic package suite.

Marking ready.

samchon:
## Merged (squash)

Squash-merged into `master` as `1dbac266c327fe3765787560ff5e431cb7129d85`. Merge-commit CI cancelled per campaign policy (non-blocking; the six fixes were verified package-scoped in the Self-Review Rounds above).

All six linked issues implemented (none narrowed/false): RA-04 #665, RA-05 #666, RA-16 #686, RA-19 #672, RA-23 #676, RA-26 #679.

The squash auto-closes the first `Closes` reference (#665). Remaining issues to close manually: #666, #686, #672, #676, #679.

Final bundle of the `repository-audit-2026-07` campaign.

PR #696, opened 2026-07-15
fix(ttsx): harden launcher runtime cache, exports, Node range, and CLI flag adjacency

### LINKED ISSUE #662: fix(ttsx): fence dependency-cache generations and publish complete emits atomically
## Problem

`ttsx` reuses JavaScript emitted for raw-TypeScript dependencies through a cross-process cache. The cache currently has two independent publication races:

1. lock ownership is represented only by the existence of `<key>.lock`, so a waiter that timed out on an old generation and an old holder's `finally` block can both delete a newer owner's lock by pathname; and
2. cache completeness is inferred from an existing metadata file plus any JavaScript file in the emit directory, even while a replacement build is deleting and repopulating that same directory.

The expected invariant is that only the current lock generation may release or reclaim itself, and readers must observe either one complete old generation or one complete new generation, never metadata from one generation combined with a partial emit from another.

This affects users who run concurrent `ttsx` processes against the same raw-TypeScript dependency cache, especially test runners, workspace scripts, and process fan-out on a cold or stale cache.

## Evidence

The stable owners are `readDependencyCache`, `withBuildLock`, and `buildDependency` in `packages/ttsc/src/launcher/internal/runtimeHooks.ts`.

- `withBuildLock` waits for up to 600 seconds, then executes `rmSync(lockDir, { force: true, recursive: true })` without proving that the path is still the generation it observed. A successful holder also removes that same pathname unconditionally in `finally`.
- `buildDependency` removes and rebuilds `emitDir` in place, then overwrites `metaPath` only after JavaScript has been emitted.
- `readDependencyCache` accepts the cache whenever `metaPath` parses and `emittedAnything(emitDir)` finds at least one JavaScript file. It does not bind either artifact to a generation or verify that publication has completed.

Two deterministic barrier-driven probes against the built functions produced:

```json
{"successorDeletedByOldFinalizer":true,"contenderCAcquired":true}
{"observedBeforeBuildCompleted":{"moduleOption":"commonjs","rootDir":"OLD"}}
```

The first result follows this exact interleaving:

1. holder A owns lock generation L0;
2. waiter B times out while observing L0 and is paused before its pathname removal;
3. A releases L0, B removes it, and B acquires successor L1;
4. A's delayed `finally` removes the path now owned by L1; and
5. contender C creates the path and enters the build while B is still active.

The second result follows this exact cache interleaving:

1. a valid old metadata marker remains at `metaPath`;
2. a rebuilding holder removes `emitDir` and writes the first JavaScript file of the new generation;
3. a reader runs before the rest of the emit and before the new metadata write; and
4. `readDependencyCache` combines the old `moduleOption`/`rootDir` with the partial new directory and reports a reusable build.

This is repository-owned behavior introduced with the dependency-building runtime in merged PR [#195](https://github.com/samchon/ttsc/pull/195). The source-plugin cache had the analogous pathname-generation defect in closed issue [#452](https://github.com/samchon/ttsc/issues/452), fixed by merged PR [#460](https://github.com/samchon/ttsc/pull/460); that fix owns `buildSourcePlugin.ts`, not the separate ttsx dependency cache.

## Consequence surface

The same root cause covers all cache readers and builders sharing a dependency cache key:

| State | Current outcome | Required outcome |
| --- | --- | --- |
| stale observer acts after a successor acquires | successor lock can be deleted | stale observation cannot affect the successor |
| old holder finalizes after a successor acquires | successor lock can be deleted | former holder can release only its own generation |
| third process arrives after either deletion | concurrent build is admitted | exactly one live owner remains |
| previous metadata exists while replacement emit begins | partial new output is accepted with old metadata | readers see only a completed, matching generation |
| builder fails during emit | partial directory can coexist with old metadata | failed generation is never reusable |
| normal warm-cache hit | cache is reused | warm reuse remains fast and unchanged |

The failure can surface on Windows, Linux, and macOS because the unsafe operation is generation-blind directory removal, not a platform-specific lock primitive. Effects range from duplicate expensive dependency builds to resolving a module against missing or mismatched emitted files.

## Approach

The architectural owner is the dependency-cache publication protocol in `runtimeHooks.ts`. Establish a single generation invariant across acquisition, reclamation, release, emit publication, metadata publication, and reads. A stale observation or former owner must be unable to mutate a replacement generation, and completion must identify the exact immutable emit generation it describes.

The source-plugin v2 protocol from #460 is relevant proven precedent, but this issue does not prescribe copying it verbatim. The ttsx cache has different artifacts and read frequency; the chosen design should be validated against its own atomicity, cleanup, and cross-platform requirements.

- Add deterministic barrier-driven multi-process regressions for stale-observer reclamation, delayed old-owner release, and a third contender; assert that at most one successor executes the build and no current lock is removed.
- Add a publication barrier after the first emitted JavaScript file while old metadata exists; assert that no reader returns a cache hit until the matching generation is complete.
- Cover failed builders, abandoned-owner recovery, ordinary release, an uncontended cold build, and a warm cache hit.
- Verify that metadata and emitted files always belong to the same generation and that cleanup cannot remove a live successor.
- Preserve current dependency compilation semantics, source-map behavior, plugin-aware dependency builds, and warm-cache reuse.
- Run the focused ttsx dependency-cache tests, for example:

  ```bash
  pnpm --filter @ttsc/test-ttsc start -- --include=ttsx_dependency_cache
  ```

- Then run the package and repository gates:

  ```bash
  pnpm --filter ttsc build
  pnpm test
  ```

## Coordination

- Campaign tracker: [#662](https://github.com/samchon/ttsc/issues/662).
- #195 is the feature lineage; #452/#460 are a closed analogous fix, not an active duplicate.
- The current open PR [#657](https://github.com/samchon/ttsc/pull/657), “fix(test): audit every lint corpus source,” changes lint corpus fixtures and harness code only. Its actual diff does not touch `runtimeHooks.ts` or ttsx tests, so there is no overlap.
- Campaign issue [#678](https://github.com/samchon/ttsc/issues/678), “Align ttsx runtime APIs with the documented Node.js 18+ support range,” also affects ttsx, but it owns Node API/version compatibility rather than dependency-cache publication. The two issues are safe to implement and verify independently.

### LINKED ISSUE #663: fix(cli): preserve unknown tsgo flag/value adjacency in ttsc and ttsx
## Problem

`ttsc` and `ttsx` intentionally forward tsgo options that are not registered in ttsc's own flag schema. Space-valued unknown options are currently split into separate collections and later recombined out of order. A single unknown flag appears to work by coincidence, but multiple flags are reordered and `ttsx` can move the entry file into program-argument tail mode before finding it.

The expected invariant is that every forwarded token reaches tsgo in its original relative order, including adjacency between an option and its separate value, while ttsc-owned options and the ttsx entry/program tail remain correctly separated.

This affects any user who passes current or future tsgo flags in `--flag value` form, which is the compatibility reason unknown forwarding exists.

## Evidence

The stable owners are `parseFlags` in `packages/ttsc/src/flags/parser.ts`, `parseBuildArgs` in `packages/ttsc/src/launcher/internal/runTtsc.ts`, and `parseCLI` in `packages/ttsc/src/launcher/internal/runTtsx.ts`.

`parseFlags` appends an unknown dash-prefixed token to `passthrough`, but appends its following bare value to `positional`. It has no schema information with which to consume the value. The callers then reconstruct a token stream by concatenating the two buckets:

- `parseBuildArgs`: `[...result.passthrough, ...trailingValues]`;
- `parseCLI`: `[...result.passthrough, ...preEntryValues]`.

The built parser produced these exact results:

```json
{"subcommand":"ttsc","passthrough":["--target","--module"],"positional":["es2020","commonjs"]}
{"subcommand":"ttsx","passthrough":["--target"],"positional":["es2020"],"tail":["entry.ts"]}
```

An external project then exercised the public CLIs:

```bash
ttsc --target es2020 --module commonjs
ttsx --target es2020 entry.ts
```

The first command forwarded the relevant suffix as:

```text
--target --module es2020 commonjs
```

instead of:

```text
--target es2020 --module commonjs
```

The second command treated `es2020` as the first positional sentinel, routed `entry.ts` to `tail`, and failed with `ttsx: entry file is required` before invoking tsgo.

The behavior is owned by the flag-forwarding lineage in merged PR [#113](https://github.com/samchon/ttsc/pull/113) and the schema consolidation tracked in closed issue [#125](https://github.com/samchon/ttsc/issues/125) and merged PR [#124](https://github.com/samchon/ttsc/pull/124). Those changes did not preserve unknown flag/value grouping; there is no active corrective tracker.

## Consequence surface

| Input class | Current behavior | Required behavior |
| --- | --- | --- |
| one unknown `--flag value` in `ttsc` | often works after bucket concatenation | preserve the original pair |
| two or more unknown space-valued flags | all flags precede all values | preserve every flag/value adjacency |
| unknown `--flag=value` | remains one token | remain unchanged |
| known schema value flag | consumed structurally | remain unchanged |
| unknown boolean flag | forwarded as one token | remain unchanged without guessing a value |
| `ttsx` unknown flag/value before entry | value becomes first positional sentinel; entry moves to tail | forward the pair to tsgo and identify the real entry |
| `ttsx` tokens after entry or `--` | program argv | remain program argv and never reach tsgo |

The failure applies on every platform because it is deterministic array reconstruction. It blocks future tsgo option compatibility and can also silently change compiler meaning when reordered tokens still parse.

## Approach

The architectural owner is the shared launcher flag parser and its forwarded-token representation. Preserve ordered token identity through parsing rather than attempting to infer unknown option arity in a caller after tokens have been separated. The representation must still allow ttsc-owned values, source-file positionals, the ttsx entry, and post-entry arguments to be classified without changing their current contracts.

This issue intentionally does not prescribe a particular parser data structure. The proving invariant is the exact ordered token stream delivered to tsgo and to the ttsx program. Do not guess the arity of a truly unknown flag: a value that is itself indistinguishable from a TypeScript input/entry remains ambiguous and should use `--flag=value` until the flag is represented in the schema. The fix owns order preservation for tokens classified as forwarded, not a new private copy of tsgo's option table.

- Add CLI-level tests for at least two unknown space-valued tsgo flags and assert exact argv order at the tsgo boundary.
- Add ttsx coverage for one and multiple unknown space-valued flags before the entry, with and without `--flag=value` forms.
- Cover unknown booleans, known launcher flags, known tsgo-only flags, source-file inputs, `--`, and program arguments after the ttsx entry.
- Assert that program arguments never leak to tsgo and compiler flags never leak to the program.
- Preserve existing errors for missing known values and invalid ttsc-owned values.
- Run focused launcher tests:

  ```bash
  pnpm --filter @ttsc/test-ttsc start -- --include=forwards_unknown_flag,spaced_flag_value
  ```

- Then run the generated schema and repository gates:

  ```bash
  pnpm run check:flags
  pnpm test
  ```

## Coordination

- Campaign tracker: [#663](https://github.com/samchon/ttsc/issues/663).
- #113, #125, and #124 are closed feature/refactor lineage, not active duplicates.
- The current open PR [#657](https://github.com/samchon/ttsc/pull/657), “fix(test): audit every lint corpus source,” was checked by actual file diff and is confined to lint corpus fixtures and harness utilities. It does not touch the parser, either launcher, or launcher tests.
- Campaign issue [#678](https://github.com/samchon/ttsc/issues/678), “Align ttsx runtime APIs with the documented Node.js 18+ support range,” also affects `ttsx`, but owns child-runtime API compatibility after parsing. This argument-order issue is independently implementable and testable.

### LINKED ISSUE #668: fix(ttsx): ignore comments and string literals during CommonJS export-name discovery
## Problem

`ttsx` exposes names from nested CommonJS sources to ESM importers by statically discovering assignments such as `exports.answer = ...`. The current discovery scans raw emitted JavaScript with a regular expression. Text that merely resembles an assignment inside a comment or string is therefore published as an ESM export, even though the CommonJS module never defines that property.

The public module namespace is incorrect: a named ESM import of the false-positive name links successfully and produces `undefined` instead of failing because the export does not exist.

## Evidence

The owner is `collectStaticCommonJsExportNames` in `packages/ttsc/src/launcher/internal/runtimeHooks.ts`. Its global pattern matches `exports.<name> =` and `module.exports.<name> =` directly in source text, without separating executable tokens from comments, strings, or template-literal text. The collected names flow through `collectCommonJsExportNames` / `collectSourceCommonJsExportNames` into the reserved and exposed star-export sets.

The behavior is reproducible with the published `ttsc@0.18.4` runtime. Compile and import this nested CommonJS source through `ttsx`:

```ts
// exports.commentGhost = 123;
const decoy = "exports.stringGhost = 456;";
export const real = "real-ok";
export const keepDecoyAlive = decoy;
```

Then inspect the ESM namespace and named imports:

```ts
import * as mod from "./fixture";
import { commentGhost, stringGhost } from "./fixture";

console.log(Object.keys(mod));
console.log(commentGhost, stringGhost);
```

Run the consumer with `npx ttsx consumer.ts` in a fixture that depends on `ttsc@0.18.4` and compiles the imported source as CommonJS.

The namespace includes both `commentGhost` and `stringGhost`, and both named imports link with value `undefined`. Neither name is assigned by executable CommonJS code. The repository test `test_ttsx_exposes_nested_cjs_source_star_exports_to_esm_named_imports.ts` covers genuine nested CommonJS star exports, but its existing string decoy resembles an ESM declaration rather than the CommonJS-assignment pattern used by this collector, so it does not exercise this false-positive path.

## Consequence surface

This affects every `ttsx` path that uses the static CommonJS-name collector to synthesize ESM-visible names from nested source or emitted CommonJS. It changes linking semantics, namespace enumeration, and feature detection: consumers can import or observe names that are absent at runtime. Comments, ordinary strings, and template-literal text can all trigger the same class of false export.

The issue is not about removing CommonJS star-export support. Genuine top-level assignments must remain discoverable, including the nested-source case fixed previously.

## Approach

Keep `collectStaticCommonJsExportNames` as the ownership boundary, but enforce the invariant that a published name must come from executable CommonJS assignment syntax, not inert lexical text. The implementation may use tokenization, lexical masking, or another syntax-aware mechanism; this issue does not prescribe one before its cost and compatibility with emitted JavaScript are verified.

Apply the same invariant to every caller that consumes this collector so source and emitted-code discovery cannot diverge.

- Assignments in line comments, block comments, quoted strings, and template-literal text do not create namespace keys and cannot be imported as named ESM exports.
- Genuine `exports.name = ...` and `module.exports.name = ...` assignments remain linkable through the supported nested CommonJS-to-ESM path.
- Add regression coverage whose decoys exactly match the CommonJS-assignment forms recognized by the current pattern, alongside genuine assignments in the same fixture.
- Cover both namespace enumeration and named-import linking behavior; checking only the collector's returned strings is insufficient.
- Extend `tests/test-ttsc/src/features/ttsx-runtime/test_ttsx_exposes_nested_cjs_source_star_exports_to_esm_named_imports.ts` with the exact assignment-shaped decoys above and run it with `pnpm --filter @ttsc/test-ttsc start -- --include=ttsx_exposes_nested_cjs_source_star_exports_to_esm_named_imports`.
- Then run the canonical repository checks `pnpm build` and `pnpm test`.

## Coordination

Issue #200 and merged PR #222 introduced nested CommonJS star-export exposure. This issue is a lexical overmatch in that solution, not a duplicate of the missing-export problem they fixed.

Open PR #657, `fix(test): audit every lint corpus source`, is confined to lint-corpus fixture registration and harness coverage and does not touch `runtimeHooks.ts` or `ttsx` module linking.

[Issue #662](https://github.com/samchon/ttsc/issues/662), **Fence ttsx dependency-cache generations and publish complete emits atomically**, also touches `runtimeHooks.ts`, but it owns cache-generation state and atomic publication rather than export-name discovery. Coordinate edits if both changes are active so their tests and modifications in the shared file remain independent.

### LINKED ISSUE #678: fix(ttsx): align runtime APIs with the documented Node.js support range
## Problem

The public setup guide requires Node.js 18 or later and recommends `npx ttsx`, while the `ttsc` package declares no `engines.node` constraint. The current `ttsx` launcher and runtime unconditionally use Node APIs that are unavailable across the documented Node 18 and Node 20 lines.

As a result, users can satisfy every published prerequisite, install successfully, type-check a valid project, and still fail before their entrypoint executes. The admitted Node range and the runtime implementation need one truthful contract.

## Evidence

The stable owners are:

- `website/src/content/docs/development/index.mdx`, which states `Node.js 18+`.
- `packages/ttsc/package.json`, which has no Node engine floor.
- `packages/ttsc/src/launcher/internal/runTtsx.ts::runPreparedEntry`, which passes `--disable-warning=ExperimentalWarning` to the child process.
- `packages/ttsc/src/launcher/internal/runtimeHooks.ts::installRuntimeHooks`, which imports and calls `node:module.registerHooks`.
- `packages/ttsc/src/launcher/internal/runtimeHooks.ts::transformOrphanSource`, which calls `stripTypeScriptTypes`.
- GitHub workflows, which exercise Node 24 only.

The current runtime requirements arrived at different Node releases: `--disable-warning` in Node 20.11, `stripTypeScriptTypes` in Node 22.13, and synchronous `registerHooks` in Node 22.15. The last API sets the effective floor for the current code path.

A published-package fixture with a valid TypeScript entrypoint reproduced:

```text
Node 18.20.8: --disable-warning rejected, exit 9
Node 20.20.2: registerHooks undefined, TypeError, exit 1
Node 22.15.0: registerHooks and stripTypeScriptTypes are functions
```

Equivalent exact probes can be run after installing `ttsc@0.18.4` and `typescript`:

```sh
npx -y node@18.20.8 node_modules/ttsc/lib/launcher/ttsx.js src/index.ts
npx -y node@20.20.2 node_modules/ttsc/lib/launcher/ttsx.js src/index.ts
npx -y node@22.15.0 node_modules/ttsc/lib/launcher/ttsx.js src/index.ts
```

The failures occur in ttsx-owned child arguments and runtime hook installation, not in the user's program or TypeScript-Go.

## Consequence surface

- `ttsx` cannot run valid projects on Node 18 or the Node 20 LTS line despite the documented prerequisite.
- npm/pnpm do not warn or refuse installation because the manifest has no engine contract.
- Node 18 fails before the preload; Node 20 reaches the preload and then fails on the missing API.
- Direct `ttsc` compiler operations may have a different minimum; the correction must not make unsupported claims for unrelated CLIs.
- CommonJS and ESM runtime graphs both depend on hook installation, so this is not limited to one module format.
- CI cannot detect drift while all workflows use Node 24.

## Approach

The owners are the ttsx runtime boundary, package manifest, public requirements, and CI matrix. The invariant is: every Node version admitted by the published contract can execute the supported ttsx paths, and every version not supported is rejected or diagnosed before users reach an internal missing-API failure.

Choose and document one coherent floor or compatibility strategy based on the actual API surface. Keep the package, docs, launcher diagnostics, and CI in sync; do not merely hide the first failing flag while leaving later missing APIs.

- The documented minimum, `engines.node` policy, launcher behavior, and CI matrix agree.
- Every admitted minimum/LTS lane runs a real `ttsx` fixture through type-check, hook registration, and entry execution.
- The fixture covers ESM and CommonJS dependencies plus an orphan raw-TypeScript dependency that reaches `stripTypeScriptTypes` or its supported equivalent.
- Unsupported Node versions receive an actionable version diagnostic or an install-time engine signal, not an internal `TypeError`.
- Node 22.15+ and Node 24 behavior remains correct.
- Ordinary `ttsc` and `ttscserver` smoke tests continue to pass on their declared ranges.
- Add version-gated `ttsx` fixtures under `tests/test-ttsc/src/features/ttsx-runtime/` and run the built launcher with the selected minimum plus current Node 24. Prove the package with `pnpm --filter ttsc build`, `pnpm --filter @ttsc/test-ttsc start`, the explicit `npx -y node@<minimum> packages/ttsc/lib/launcher/ttsx.js <fixture>` smoke, and the broader `pnpm test`.

## Coordination

PR #195 introduced the current raw-TypeScript dependency runtime and its hook-based execution model. No earlier corrective tracker owns the documented Node floor. This issue is related to campaign issue #662, **Fence ttsx dependency-cache generations and publish complete emits atomically**, because both affect `ttsx` runtime reliability, but cache publication and Node API compatibility are independent changes.

The only current open pull request, #657 (**fix(test): audit every lint corpus source**), is confined to lint corpus coverage and has no overlap with the launcher, manifest, setup guide, or Node matrix.

samchon:
## Bundle b1 — ttsx / CLI launcher runtime

Repository-audit-2026-07 campaign batch covering the shared launcher flag
parser, the ttsx source runtime hooks, and the Node.js support contract. All
four issues touch the launcher/runtime surface and share verification.

### Issues

- Closes #663 (RA-02) — preserve unknown tsgo flag/value adjacency in `ttsc`
  and `ttsx`. The parser now classifies a bare token as a genuine positional or
  a forwarded flag value in place, so `--flag value` pairs reach tsgo in order
  and the ttsx entry is never mistaken for a flag value.
- Closes #668 (RA-10) — ignore comments and string/template-literal text during
  CommonJS export-name discovery, so a decoy `exports.x =` inside inert lexical
  text no longer leaks a phantom ESM export.
- Closes #678 (RA-25) — align the ttsx runtime with a single truthful Node.js
  floor (22.15.0): `engines.node`, an actionable launcher diagnostic, and the
  setup-guide requirement now agree.
- Closes #662 (RA-01) — fence ttsx dependency-cache generations and publish
  complete emits atomically. (In progress in this branch.)

### Verification

Package-scoped launcher + ttsx-runtime suites plus `pnpm run check:flags`;
exact commands and results posted on the thread. Campaign CI is cancelled per
the exact-SHA gate after every push.

https://claude.ai/code/session_01HYCpGEc14zxD99ZPPzAnpz

### DIFF
--- a/packages/ttsc/package.json
+++ b/packages/ttsc/package.json
@@ -40,6 +40,9 @@
   },
   "author": "Jeongho Nam",
   "license": "MIT",
+  "engines": {
+    "node": ">=22.15.0"
+  },
   "bugs": {
     "url": "https://github.com/samchon/ttsc/issues"
   },
--- a/packages/ttsc/src/flags/parser.ts
+++ b/packages/ttsc/src/flags/parser.ts
@@ -58,6 +58,24 @@ export interface ParseOptions {
    * `--` is appended to `passthrough` as-is (ttsx already does this).
    */
   readonly honorDoubleDashSeparator?: boolean;
+  /**
+   * Classifies a bare (non-dash) token as a genuine positional argument (a
+   * source file, the ttsx entry, a project path) rather than the space-separated
+   * value of a preceding forwarded flag.
+   *
+   * When omitted, every bare token is a positional — the historical behaviour
+   * for project-shaped subcommands that never forward `--flag value` pairs.
+   *
+   * When provided, a bare token that fails the predicate is appended to
+   * `passthrough` in its original position instead of `positional`, so an
+   * unknown `--flag value` pair reaches tsgo with its adjacency and relative
+   * order intact. The parser deliberately does not guess a forwarded flag's
+   * arity from the flag itself (it has no schema for a truly unknown flag);
+   * the predicate is the only signal that separates a forwarded value from a
+   * real input file, and both callers key it on the TypeScript source
+   * extension.
+   */
+  readonly isPositional?: (token: string) => boolean;
 }
 
 /**
@@ -153,7 +171,17 @@ export function parseFlags(opts: ParseOptions): ParseResult {
       continue;
     }
 
-    // Bare token: positional argument (file / entry / project path).
+    // Bare token: either a genuine positional argument (file / entry / project
+    // path) or the space-separated value of a preceding forwarded flag. When a
+    // caller forwards unknown flags it supplies `isPositional` to tell the two
+    // apart; a forwarded value is appended to `passthrough` in place so the
+    // `--flag value` pair reaches tsgo in its original order and adjacency,
+    // instead of being split into a separate bucket the caller later
+    // concatenates out of order.
+    if (opts.isPositional !== undefined && !opts.isPositional(current)) {
+      passthrough.push(current);
+      continue;
+    }
     positional.push(current);
     if (opts.forwardAfterFirstPositional === true && positional.length === 1) {
       forwardingTail = true;
--- a/packages/ttsc/src/launcher/internal/runTtsc.ts
+++ b/packages/ttsc/src/launcher/internal/runTtsc.ts
@@ -461,6 +461,12 @@ function parseBuildArgs(argv: readonly string[], checkOnly: boolean) {
   const result = parseFlags({
     argv,
     errorPrefix: "ttsc:",
+    // A bare token is a single-file input only when it carries a TypeScript
+    // source extension; any other bare token is the space-separated value of a
+    // preceding forwarded flag (e.g. the `es2020` in `--target es2020`). The
+    // parser routes those values into `passthrough` in place, so the forwarded
+    // flag/value pairs reach tsgo in their original order.
+    isPositional: looksLikeInputFile,
     subcommand: "build",
   });
   // Defaults: pinned by the previous hand-parser. `quiet` defaults true,
@@ -476,15 +482,13 @@ function parseBuildArgs(argv: readonly string[], checkOnly: boolean) {
   else if (explicitNoEmit === true) emit = false;
   if (emit === undefined && checkOnly) emit = false;
 
-  const files = result.positional.filter(looksLikeInputFile);
-  // Bare non-file positionals are flag values (e.g. the `es2020` in
-  // `--target es2020`). The engine left them in `positional` because the
-  // forwarded flag is unknown to the schema; route them back into
-  // passthrough so the user's pair reaches tsgo intact.
-  const trailingValues = result.positional.filter(
-    (token) => !looksLikeInputFile(token),
-  );
-  const passthrough = [...result.passthrough, ...trailingValues];
+  // `isPositional: looksLikeInputFile` guarantees every `result.positional`
+  // token is a TypeScript input file; forwarded flag values already live in
+  // `result.passthrough` in their original order, so no reconstruction is
+  // needed here (the previous `[...passthrough, ...trailingValues]` concat
+  // reordered every flag ahead of every value).
+  const files = [...result.positional];
+  const passthrough = [...result.passthrough];
 
   return {
     binary: getString(result, "--binary"),
--- a/packages/ttsc/src/launcher/internal/runTtsx.ts
+++ b/packages/ttsc/src/launcher/internal/runTtsx.ts
@@ -12,6 +12,7 @@ import {
 import { getCompilerVersionText } from "./getCompilerVersionText";
 import { prepareExecution } from "./prepareExecution";
 import { resolveCacheDir } from "./resolveCacheDir";
+import { checkNodeRuntimeSupport } from "./runtimeHooks";
 
 /**
  * CLI entry point for `ttsx`. Type-checks the owning project via tsgo, emits
@@ -43,6 +44,17 @@ function run(argv: readonly string[]): number {
       `${getCompilerVersionText().replace(/^ttsc\b/, "ttsx")}\n`,
     );
     return 0;
+  }
+
+  // Refuse an unsupported Node.js before type-checking and spawning the child:
+  // the child inherits this process's Node version, so an early diagnostic here
+  // pre-empts both the Node 18 `--disable-warning` rejection and the Node 20
+  // missing-`registerHooks` TypeError with one actionable message. `--help` and
+  // `--version` are handled above so they still print on any Node.
+  const nodeSupport = checkNodeRuntimeSupport(process.versions.node);
+  if (nodeSupport !== null) {
+    process.stderr.write(`ttsx: ${nodeSupport}\n`);
+    return 2;
   }
 
   const cwd = path.resolve(parsed.cwd ?? process.cwd());
@@ -109,6 +121,14 @@ function parseCLI(argv: readonly string[]) {
     errorPrefix: "ttsx:",
     forwardAfterFirstPositional: true,
     honorDoubleDashSeparator: true,
+    // Only a TypeScript-extensioned bare token is the entry; every other bare
+    // token before it (e.g. the `es2020` in `--target es2020 entry.ts`) is a
+    // forwarded flag value. Classifying values via the predicate keeps them in
+    // `passthrough` in order AND stops a pre-entry value from being mistaken for
+    // the first positional sentinel — which previously flipped the parser into
+    // tail mode and pushed the real entry into `tail`, failing with
+    // "entry file is required".
+    isPositional: looksLikeEntryFile,
     subcommand: "ttsx",
   });
 
@@ -116,18 +136,13 @@ function parseCLI(argv: readonly string[]) {
   if (entry === undefined) {
     throw new Error("ttsx: entry file is required");
   }
-  // With `forwardAfterFirstPositional: true` the parser reports
-  // `result.positional` as just the entry, `result.passthrough` as flags
-  // arriving BEFORE the entry (tsgo-forwarded), and `result.tail` as every
-  // token AFTER the entry — those are the user program's argv (e.g. the
-  // `generate --input src/input` tail of `ttsx typia.ts generate
-  // --input src/input`) and MUST NOT reach tsgo. Anything in positional
-  // that is not the entry is a pre-entry flag value (e.g. `--target es2020`)
-  // that the parser stored positionally; forward those to tsgo with the
-  // rest of `passthrough`.
-  const preEntryValues: string[] = result.positional.filter(
-    (token) => token !== entry && !looksLikeEntryFile(token),
-  );
+  // With `forwardAfterFirstPositional: true` and `isPositional:
+  // looksLikeEntryFile`, the parser reports `result.positional` as just the
+  // entry, `result.passthrough` as the tsgo-forwarded flags (and their
+  // in-order space values) arriving BEFORE the entry, and `result.tail` as
+  // every token AFTER the entry — the user program's argv (e.g. the `generate
+  // --input src/input` tail of `ttsx typia.ts generate --input src/input`),
+  // which MUST NOT reach tsgo.
   const postEntryArgs: string[] = [...result.tail];
 
   const preload: string[] = [];
@@ -170,7 +185,7 @@ function parseCLI(argv: readonly string[]) {
     preload,
     project: getString(result, "--tsconfig"),
     singleThreaded: getBoolean(result, "--singleThreaded") === true,
-    tsgoFlags: [...result.passthrough, ...preEntryValues],
+    tsgoFlags: [...result.passthrough],
   };
 }
 
--- a/packages/ttsc/src/launcher/internal/runtimeHooks.ts
+++ b/packages/ttsc/src/launcher/internal/runtimeHooks.ts
@@ -1187,8 +1187,19 @@ function serveBuiltDependency(
       };
 }
 
-/** On-disk completion marker for a built dependency, shared across processes. */
+/**
+ * On-disk completion marker for a built dependency, shared across processes.
+ *
+ * `generation` names the exact immutable emit directory this marker describes
+ * (`<cacheDir>/gen-<generation>`). Binding metadata to one generation is what
+ * makes publication atomic: a reader that parses this marker reads the emit of
+ * the SAME generation, never old metadata combined with a different, still
+ * partially-written directory. The marker is the last thing a build writes, and
+ * it is written by an atomic temp-and-rename, so a reader observes either one
+ * complete old generation or one complete new generation.
+ */
 interface DependencyCacheMeta {
+  generation: string;
   rootDir: string;
   moduleOption?: string;
 }
@@ -1210,25 +1221,28 @@ function ensureProjectBuilt(tsconfig: string): BuiltProject {
   if (cached !== undefined) {
     return cached;
   }
-  const { emitDir, lockDir, metaPath, root } = dependencyCachePaths(tsconfig);
+  const { cacheDir, lockDir, metaPath, root } = dependencyCachePaths(tsconfig);
 
-  const reuse = readDependencyCache(emitDir, metaPath);
+  const reuse = readDependencyCache(cacheDir, metaPath);
   if (reuse !== null) {
     builtProjects.set(tsconfig, reuse);
     return reuse;
   }
 
   fs.mkdirSync(root, { recursive: true });
-  const built = withBuildLock(lockDir, metaPath, emitDir, () =>
-    buildDependency(tsconfig, emitDir, metaPath),
+  const built = withBuildLock(cacheDir, metaPath, lockDir, () =>
+    buildDependency(tsconfig, cacheDir, metaPath),
   );
   builtProjects.set(tsconfig, built);
   return built;
 }
 
 interface DependencyCachePaths {
-  emitDir: string;
+  /** Container of this dependency's generation-stamped emit directories. */
+  cacheDir: string;
+  /** Fenced cross-process coordination directory (`<key>.lock`). */
   lockDir: string;
+  /** Atomic completion pointer (`<key>.json`) naming the live generation. */
   metaPath: string;
   root: string;
 }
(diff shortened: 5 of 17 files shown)


samchon:
### Self-Review — round 1

Reviewed the full bundle diff (parser adjacency, CJS export masking, Node floor, and the RA-01 fenced dependency-cache protocol) against the four issue specs.

**Findings**

1. **RA-01 fence correctness (verify).** The generation-fenced lock must guarantee a stale observer or former owner can never retire a successor. Plan: keep the invariant "the only way to free `current` is to create the retiring generation's tombstone, so a successor can only acquire after the predecessor's tombstone exists" and prove it with deterministic multi-process regressions (delayed old-owner release, two stale observers of a dead owner). Done — verified below.

2. **RA-01 publication atomicity (verify).** A reader must never combine old metadata with a partial new emit. Plan: bind the metadata marker to one immutable `gen-<generation>` directory and publish by atomic temp-and-rename; pin with a publication-barrier regression. Done.

3. **RA-10 mask coverage.** The string and template-literal decoys are guaranteed to survive tsgo emit as runtime values, so those mask branches are exercised (confirmed by a red run with the mask disabled → a decoy leaks and the test fails). Line/block-comment decoys rely on tsgo's default comment preservation; they are harmless if stripped. No change needed.

4. **RA-25 CI matrix — deliberate narrowing, flag to lead.** The floor is now consistent across `engines.node`, the launcher diagnostic, and the setup guide, and the boundary is pinned by exercising the exported `checkNodeRuntimeSupport` directly (the e2e suite runs under a single Node version). Adding a Node 22.15 CI lane touches shared `.github/workflows` — outside this bundle's launcher-runtime surface and high-conflict during the CI-suspended campaign. Recommending the lead add the lane as a follow-up rather than editing workflows here.

**Remediation plan for this round:** confirm no dead references to the removed cache helpers, re-run the RA-01 regressions and the dependency-build e2e suite, and confirm the flag schema is untouched (the local `check:flags` drift was a Windows CRLF artifact, reverted). Acting now.

samchon:
### Self-Review — round 2 (clean)

Re-read the assembled `runtimeHooks.ts` cache/lock region and the four test areas end to end. No new findings.

- **Fence invariant holds under every interleaving I traced**, including the 600s steal-timeout path where a waiter reclaims a still-live but slow holder: the reclaimer builds into its own `gen-<generation>` and republishes the marker atomically, the original holder's late release finds its tombstone occupied and fails, and both generation directories stay complete — duplicate work, never corruption.
- **Fast path intact**: a published marker is reused immediately even before the holder releases the lock.
- **No dead references** to the removed `waitForDependencyCache` / `writeDependencyMeta`; `tsc` (unused-locals on) passes; the flag schema is untouched (the local `check:flags` diff was a Windows CRLF-only artifact and was reverted).
- Working tree is clean; the diff is confined to the launcher/runtime surface plus its tests and the setup-guide requirement — no `.github/workflows` or generated `flags_gen` changes.

Two clean rounds total. Proceeding to final verification and marking ready.

samchon:
### Final local verification

All package-scoped, run on Windows (Node 22.15.0), green:

**Build**
```
pnpm --filter ttsc build        # tsc clean (unused-locals enforced)
```

**RA-01 fenced dependency cache + RA-02/10/25 launcher tests**
```
pnpm --filter @ttsc/test-ttsc start -- --include=dependency_cache,parseflags,\
preserves_two_spaced,runs_the_entry_after_spaced,diagnoses_unsupported_node,\
exposes_nested_cjs,forwards_an_unknown_flag,forwards_a_spaced_flag,\
post_entry_argv,forwards_argv_after,reports_frontdoor
# 16 passed:
#   test_ttsx_dependency_cache_old_finalizer_cannot_release_a_successor
#   test_ttsx_dependency_cache_fences_two_stale_observers_of_a_dead_owner
#   test_ttsx_dependency_cache_reader_never_mixes_metadata_with_a_partial_emit
#   test_ttsx_dependency_cache_recovers_from_a_dead_owner_and_reports_lock_states
#   test_parseflags_preserves_unknown_spaced_flag_value_order_for_ttsc
#   test_parseflags_keeps_the_ttsx_entry_after_a_spaced_flag_value
#   test_ttsc_preserves_two_spaced_tsgo_flag_pairs_in_order
#   test_ttsx_runs_the_entry_after_spaced_tsgo_flag_values
#   test_ttsx_exposes_nested_cjs_source_star_exports_to_esm_named_imports
#   test_ttsx_diagnoses_unsupported_node_versions_at_the_documented_floor
#   (+ 6 adjacent existing forwarding/entry/frontdoor tests)
```

**Dependency-build regression (RA-01 refactor did not disturb real builds)**
```
pnpm --filter @ttsc/test-ttsc start -- --include=built_dependency,\
builds_a_raw_ts_dependency,preserves_enum_runtime_object,\
preserves_module_not_found,runs_a_published,runs_a_dependency_with_a_circular,\
shares_one_dependency,dependency_own_type_diagnostics
# 15 passed
```

**RA-10 red→green proof**: with `maskCommentsAndStrings` disabled in the built lib, `test_ttsx_exposes_nested_cjs_source_star_exports_to_esm_named_imports` fails (a decoy leaks into the namespace, exit-code assertion `1 !== 0`); restored, it passes.

**Flag schema**: `check:flags` produced only Windows CRLF line-ending noise on the generated `flags_gen.go` / `flags.mdx`; the schema itself is untouched, so a Linux CI run is a no-op. Reverted the line-ending churn.

**Note for the lead — RA-25 CI matrix**: the floor is now consistent across `engines.node`, the launcher diagnostic, and the setup guide, with the boundary pinned via the exported `checkNodeRuntimeSupport`. Adding a Node 22.15 lane to `.github/workflows` is left as a follow-up for the lead (shared surface, outside this bundle, high-conflict during CI suspension).

Marking ready for review. Not merging — the lead merges after recheck.

samchon:
Squash-merged as `adafcc7e3cee82e5b4dfefccbdb31a206a20e77a`. Merge-commit CI (test + website) cancelled by exact SHA per the campaign gate.

**Note for the lead — multi-issue closure:** this PR links four issues but a squash merge only auto-closes the first `Closes` reference in the body. Please close the remaining ones manually:
- #663 (RA-02)
- #668 (RA-10)
- #678 (RA-25)
- #662 (RA-01)

(Whichever GitHub auto-closed from the squash body, close the other three.)

Issue #683, opened 2026-07-15
fix(flags): make generated drift checks independent of checkout line endings


samchon:
## Problem

The generated CLI-flag drift check compares raw working-tree text before and after regeneration. In a clean Windows checkout with `core.autocrlf=true`, committed generated files are materialized as CRLF, while the Node generator and Go formatter rewrite them as LF. The check reports all generated files as drifted even though Git's normalized content is identical.

The expected behavior is that the drift gate fails for semantic/generated-content differences, but passes for a clean checkout regardless of Git's configured working-tree line-ending conversion.

This blocks Windows contributors at the first step of the root `pnpm test` script and can obscure real schema drift behind four false-positive paths.

## Evidence

The stable owners are `snapshot` in `packages/ttsc/scripts/check-flags.cjs` and `writeFile` in `packages/ttsc/scripts/gen-flags.mts`.

- `check-flags.cjs` snapshots each target with `fs.readFileSync(file, "utf8")`, runs the generator and Go formatter, snapshots raw text again, and compares JavaScript strings.
- `gen-flags.mts` writes UTF-8 strings containing LF line terminators. Its Go outputs are then formatted by the repository's Go formatter, also yielding LF in the working tree.
- The check does not compare Git-normalized content or normalize line endings before deciding drift.
- The root `test` script begins with `pnpm run check:flags`. The only Windows CI lane runs Go tests and does not exercise this gate, so the contributor-specific failure is not currently covered.

Exact reproduction from a separate clean checkout:

```powershell
git clone -c core.autocrlf=true https://github.com/samchon/ttsc.git ttsc-autocrlf
Set-Location ttsc-autocrlf
git ls-files --eol packages/ttsc/cmd/ttsc/flags_gen.go packages/ttsc/utility/flags_gen.go packages/lint/linthost/flags_gen.go website/src/content/docs/ttsc/flags.mdx
node packages/ttsc/scripts/check-flags.cjs
$LASTEXITCODE
git diff --numstat
```

Before the check, `git ls-files --eol` reported `i/lf w/crlf` for the four targets. The check exited `1` with:

```text
ttsc flag schema: generated output drifted from src/flags/schema.ts:
  packages/ttsc/cmd/ttsc/flags_gen.go
  packages/ttsc/utility/flags_gen.go
  packages/lint/linthost/flags_gen.go
  website/src/content/docs/ttsc/flags.mdx
```

[...]

PR #697, opened 2026-07-15
fix(ttsc): reject malformed resident-transform replies and stop generated-flag drift

### LINKED ISSUE #674: fix(ttsc): reject malformed resident-transform replies as protocol failures
## Problem

The resident transform client advances its FIFO request queue before validating a reply and converts invalid JSON or a non-object JSON value into `{}`. The public `TtscService` then interprets that empty object as an ordinary domain result: `transformFile` returns `undefined` as if the file were absent, and `updateFile` returns `false` as if the edit merely failed to compile.

A malformed line is a protocol failure, not a valid negative response. Accepting it also loses the request/reply boundary: a later real reply can be assigned to the next request.

## Evidence

The stable owners are:

- `packages/ttsc/src/compiler/internal/residentTransformProcess.ts::ResidentTransformProcess.onLine`
- `packages/ttsc/src/compiler/internal/residentTransformProcess.ts::parseReply`
- `packages/ttsc/src/TtscService.ts::{transformFile,updateFile}`

`onLine` removes the first pending request and calls `request.resolve(parseReply(trimmed))`. `parseReply` catches JSON errors, rejects arrays and primitives, and returns an empty object for every invalid shape. It never rejects the dequeued request or fails the process.

An executable stub that reads one request line and writes either `not-json` or `[]` reproduces the behavior:

1. Construct `ResidentTransformProcess` against the stub.
2. Call `request({ file: "src/index.ts" })`.
3. For each malformed reply, the Promise resolves to `{}` rather than rejecting.
4. Through `TtscService.transformFile` that becomes `undefined`; through `updateFile` it becomes `false`.

If the host emits a malformed line and then the valid reply for that request, the malformed line consumes the FIFO slot. The valid reply is now unsolicited or can be paired with a later request, depending on timing. The process already treats an unsolicited valid line as fatal, so treating malformed replies as benign is internally inconsistent.

## Consequence surface

- Bundler, editor, watch, and code-generation consumers can mistake host corruption for a missing source or a normal rejected update.
- A single malformed line can desynchronize all following requests in the resident process.
- Invalid JSON, arrays, primitives, null, and object replies with the wrong operation shape need explicit treatment.
- A valid transform response and a valid update response have different required fields; operation-specific validation belongs either at the protocol boundary or before the public service converts it to a domain result.
- Host process exits and pipe failures already reject pending work and should retain their stderr-rich errors.

## Approach

The resident client owns framing and protocol integrity; `TtscService` owns the operation-level response contract. The invariant is that exactly one validated response completes each request. A line that cannot represent the expected reply must reject the affected request and put the resident process into a state that cannot mispair later lines.

Preserve FIFO performance and the existing behavior for valid `found: false` and `updated: false` replies. Do not convert transport/protocol errors into domain negatives.

- Invalid JSON, arrays, primitives, and null reject instead of resolving `{}`.
- Invalid transform and update object shapes reject with a protocol-specific error.
- After a malformed reply, no later response can be delivered to the wrong pending request.
- Valid `{ found: false }` still produces `undefined` and valid `{ updated: false }` still produces `false`.
- Valid transformed text and successful updates retain current behavior.
- Exit, stderr, EPIPE, disposal, blank-line, and unsolicited-reply tests remain correct.
- Extend `tests/test-ttsc/src/features/api/test_residenttransformprocess_lifecycle.ts` with stub-process cases for malformed JSON, wrong JSON types, wrong operation shapes, and two queued requests; add public `TtscService` assertions where operation-shape validation lives there.
- Prove it with `pnpm --filter ttsc build`, `pnpm --filter @ttsc/test-ttsc start`, and the broader `pnpm test`.

## Coordination

The resident service commits were developed from issue #255's Metro analysis, which identified a resident per-file transform API as the prerequisite for avoiding whole-project recompilation. This issue preserves that architecture and tightens only its protocol boundary. It is related to campaign issue #672, **Evict failed unplugin transforms so long-lived adapters can recover**, because both define recovery after a host failure, but resident framing and unplugin cache eviction are separate owners.

The only current open pull request, #657 (**fix(test): audit every lint corpus source**), changes lint corpus and test-harness paths only. It does not overlap `ResidentTransformProcess`, `TtscService`, or resident host tests.

samchon:
## Bundle b3 — resident transform protocol + generated-flag drift

Two independent hardening fixes from the `repository-audit-2026-07` campaign. Separate owners, one bundle.

### RA-21 (#674) — reject malformed resident-transform replies as protocol failures

`ResidentTransformProcess` used to funnel any unparseable or non-object reply line into `{}`, which `TtscService` then read as an ordinary domain negative: `transformFile` → `undefined` (as if the file were absent), `updateFile` → `false` (as if the edit failed to compile). A malformed line also stole a FIFO slot, so a later valid reply could pair with the wrong request.

- The client now validates each reply against the operation it answers (`request(payload, kind)`).
- A line that is not a JSON object is a **framing violation**: reject the affected request and fail the whole process, so a corrupt line can never mispair a following reply.
- A well-formed object of the **wrong operation shape** rejects only that request (FIFO framing is intact).
- Valid `{ found: false }` still resolves to `undefined`; valid `{ updated: false }` still resolves to `false`.
- Tests: framing violations (invalid JSON, array, primitive, `null`), wrong operation shape, malformed-reply-fails-queued-requests (desync guard), and the negative twin (valid negatives still resolve). `api.mdx` documents the domain-negative vs protocol-failure boundary.

### RA-09 (#683) — make generated flag-drift checks independent of checkout line endings

`check-flags.cjs` compared raw working-tree bytes before/after regeneration. On a clean `core.autocrlf=true` checkout the committed LF files materialize as CRLF while the generator/gofmt emit LF, so all four generated targets were reported as drifted even though git's normalized content is identical — blocking Windows contributors at the first step of `pnpm test`.

- Drift comparison now folds EOLs (CRLF and lone CR → LF) before comparing, so the gate is independent of git's configured checkout conversion while staying sensitive to every real content byte.
- Regression test proves CRLF-vs-LF identical content = no drift, genuine content change still drifts (named), and boundary cases (missing target, lone CR).

Closes #674, #683

---

https://claude.ai/code/session_01HYCpGEc14zxD99ZPPzAnpz


samchon:
### Self-Review — round 1

Inspected the whole declared surface of both fixes.

**RA-21 (resident protocol) — verified sound:**
- Confirmed the real Go host (`packages/ttsc/utility/serve.go`) always emits `{"typescript":<string>,"found":<bool>}` for transform and `{"updated":<bool>}` for update, both with their required typed fields — so `isValidReply` never rejects a legitimate production reply. Even the host's own malformed-request fallback encodes `serveResponse{}` = `{"typescript":"","found":false}`, a valid transform negative.
- Enumerated every `ResidentTransformProcess.request` caller: only `TtscService.{transformFile,updateFile}` (both pass `kind`) and the resident tests. The `ttscserver` `.request` is a separate LSP client.
- Framing-violation path rejects the dequeued request then `fail()`s the process; the trailing genuine reply lands on an empty queue with `failure` set, so it is discarded as benign (not mispaired). Wrong-operation-shape path rejects only that request, FIFO intact.
- Ran the real-host end-to-end suites: `service` (transform + reject-on-non-compile) and `service-incremental` (update) — all green.

**RA-09 (flag drift) — verified sound:**
- Ran the actual `check:flags` gate against this worktree's real `core.autocrlf=true`/CRLF working tree: exit 0 (the old raw compare flagged all four targets). Regeneration leaves git-normalized content clean (`git diff --numstat` empty).
- 7 deterministic regression cases pass, derived from the real committed targets.

**Finding to remediate:** one inaccurate doc-comment in `check-flags.test.cjs` — it says "flip one non-terminator byte" but the case appends bytes. Fixing wording for precision. No behavior change.

samchon:
### Self-Review — round 2 (clean)

Re-read the full diff after the round-1 fix. No further findings.

- `computeDrift` iterates the canonical `targets` list; a deleted/absent target normalizes to `""` and correctly drifts. `normalizeEol` folds only `\r\n`/`\r`, never tabs or interior spaces (asserted), so gofmt alignment drift is still caught.
- Guarded `require.main === module` means requiring `check-flags.cjs` from the test triggers no generation side effects.
- Resident client: exactly one settlement per request on every path (framing-fail, shape-reject, resolve, process-fail, write-error, disposal). No double-settle.
- Doc comment precision fixed in `check-flags.test.cjs`.

Rounds: 2 (round 1 found one doc-comment wording issue, remediated; round 2 clean).

samchon:
### Local verification (all green)

**RA-21 — resident protocol**
- `pnpm --filter ttsc build` → clean (tsc)
- `TTSC_TEST_DIR=features pnpm --filter @ttsc/test-ttsc start -- --include=residenttransformprocess` → 5/5 pass (lifecycle, framing-violations, wrong-operation-shape, malformed-fails-queued, valid-negative-replies)
- Real Go host end-to-end (`TTSC_GO_BINARY=$(go env GOROOT)/bin/go`, scope test-ttsc):
  - `TTSC_TEST_DIR=native-plugins/service` → 2/2 pass (transform through host; reject when project does not compile)
  - `TTSC_TEST_DIR=native-plugins/service-incremental` → 1/1 pass (reflects a file update through the host)

**RA-09 — flag drift**
- `node --test packages/ttsc/scripts/check-flags.test.cjs` → 7/7 pass
- `node packages/ttsc/scripts/check-flags.cjs` against this worktree's real `i/lf w/crlf` (autocrlf) checkout → exit 0; `git diff --numstat` on the four targets empty (git-normalized content clean, nothing staged). The pre-fix raw compare reported all four as drifted.

CI wiring: the regression runs on the `windows-go` lane (actual Windows checkout conversion) and beside `check:flags` on the ubuntu `typecheck` lane.

samchon:
Squash-merged as `25fed621e03832e53251b197afa53c2fcf422ea6`. Merge-commit CI was cancelled by exact SHA per campaign discipline (workflow runs are not left to execute during the suspended-CI campaign).

Issue #670, opened 2026-07-15
fix(playground): support valid CommonJS root package.json.exports shapes in the Execute sandbox


samchon:
## Problem

The Playground Execute sandbox installs CommonJS packages into an in-memory pack and resolves their entrypoints itself. Its root-package resolver accepts an `exports` object only when that object has a `"."` key. It rejects two other valid root export shapes: a root string target and a root conditional object.

As a result, a package can download successfully and be loadable by Node's CommonJS resolver, yet fail at Playground execution with `require("<package>") is not available in the playground sandbox`.

## Evidence

The owner is `resolvePackageEntry` inside `packages/playground/src/sandbox/createSandboxRequire.ts`. For a bare package name, it reads `package.json`, checks `exports` only as an object, indexes `exports["."]`, and passes that value to `pickConditionalExport`. It never passes a root string or a root condition map such as `{ "require": "...", "default": "..." }` to the existing conditional picker.

The public sandbox API reproduces the difference without network access:

```ts
import { createSandboxRequire } from "./packages/playground/src/sandbox/createSandboxRequire";

const execute = (exportsField: unknown) =>
  createSandboxRequire(
    {
      "fixture/package.json": JSON.stringify({ exports: exportsField }),
      "fixture/entry.cjs": "module.exports = { value: 'ok' };",
    },
    { console },
  )("fixture");

console.log(execute({ ".": { require: "./entry.cjs" } }));
console.log(execute("./entry.cjs"));
console.log(execute({ require: "./entry.cjs", default: "./entry.cjs" }));
```

The `"."` form returns `{ value: "ok" }`. The root string and root conditional forms both throw the sandbox-unavailable error, although both describe the same valid root entry for a CommonJS consumer.

## Consequence surface

This affects the browser Execute lane for downloaded packages whose `package.json#exports` uses either root form. Compilation and package installation can succeed before evaluation fails, making the diagnostic appear to be a missing package rather than an unsupported manifest shape.

Root `main` and `index` fallbacks, `exports["."]`, exact subpath exports, wildcard subpath exports, relative requires, JSON modules, and the module cache share this resolver and must remain stable. This issue does not ask the sandbox to execute ESM-only packages; it is limited to CommonJS-compatible targets expressed through valid root export shapes.

## Approach

[...]

Issue #669, opened 2026-07-15
fix(playground): discover npm dependencies from executable module syntax only


samchon:
## Problem

The Playground decides which external npm packages to download by scanning user source with a regular expression. The scan does not distinguish executable module syntax from comments or string contents, so inert text can request packages that the program never imports.

This is observable behavior, not only unnecessary work: the dependency loader can perform network requests for attacker-controlled or nonexistent package names and can block an otherwise valid compile when one of those phantom dependencies cannot be downloaded.

## Evidence

The owner is `collectModuleSpecifiers` in `packages/playground/src/npm/collectExternalPackageNames.ts`. `MODULE_SPECIFIER_REGEXP` is executed over the complete source string and recognizes textual forms of static imports/exports, dynamic imports, and `require(...)` without first excluding comments and string or template-literal contents.

This direct public-function proof demonstrates the false dependencies:

```ts
import { collectExternalPackageNames } from "./packages/playground/src/npm/collectExternalPackageNames";

console.log(
  collectExternalPackageNames(
    `
      // import "ghost-comment";
      const text = "require('ghost-string')";
      /* export { value } from "ghost-block"; */
      import "real-package";
    `,
    [],
  ),
);
```

The current result contains `ghost-block`, `ghost-comment`, `ghost-string`, and `real-package`. Only `real-package` is executable module syntax. `collectExternalPackageNames` then normalizes these specifiers with `packageNameFromSpecifier`, filters built-ins, and returns the false names to the Playground npm-loading path.

## Consequence surface

The failure covers every Playground lane that uses external-package discovery before installing a runtime pack. It can add registry traffic, latency, misleading download failures, and packages that should never enter the browser sandbox. A comment copied from documentation or a string used as test data is sufficient to change compilation behavior.

The supported discovery surface must remain intact for real static imports, `export ... from`, string-literal dynamic imports, and supported CommonJS `require` calls. Package-name normalization, built-in filtering, uniqueness, and stable sorting are separate existing contracts and must not regress.

## Approach

Keep `collectModuleSpecifiers` as the discovery owner and enforce the invariant that dependencies originate from executable module-loading syntax. Choose a lexical or syntax-aware implementation only after checking the Playground bundle-size and parser-availability constraints; a specific parser dependency is not assumed by this issue.

Discovery should remain conservative about expressions it cannot statically resolve. It must not turn arbitrary string data or comments into package installation requests.

## Acceptance and verification

[...]

PR #699, opened 2026-07-15
fix(playground): surface plugin failures and harden dependency and sandbox discovery

### LINKED ISSUE #664: fix(playground): surface plugin failures instead of clean lint or successful untransformed builds
## Problem

The Playground worker treats a resolved WASM plugin call as successful without checking its process-style `code`. A transform plugin can return a nonzero envelope and the worker will continue to compile the original, untransformed source; a lint plugin can return a nonzero envelope and the worker will return an empty diagnostic list. Promise rejection and malformed output are also deliberately swallowed in the same paths.

The expected behavior is that a plugin failure remains a user-visible failure. The Playground must not present a clean lint result or successful JavaScript when a configured transform/check plugin did not complete successfully.

This affects `@ttsc/playground` consumers, including the repository website and downstream browser playgrounds using `createWorkerCompiler` with typia or lint plugins.

## Evidence

The stable owners are `applyTypiaTransform` and `runLintPipeline` in `packages/playground/src/compiler/createWorkerCompiler.ts`.

- `applyTypiaTransform` awaits `api.plugin`, inspects only `stdout`, and catches every thrown or parsing error. It never checks `raw.code`. If no usable transform is found, `runBuildPipeline` still calls `api.build` against the original MemFS source and can return `{ type: "success" }`.
- `runLintPipeline` awaits the lint plugin and always parses `raw.stderr`; it never checks `raw.code`. Its catch path returns `{ diagnostics: [] }`.
- The build call already demonstrates the intended envelope boundary: `runBuildPipeline` explicitly handles `raw.code !== 0` before parsing the result. Plugin calls do not apply that rule.

A focused worker harness supplied this resolved plugin envelope:

```json
{"code":2,"stdout":"","stderr":"plugin failed"}
```

With the transform plugin enabled, compilation continued and returned successful untransformed JavaScript. With the lint plugin enabled, lint returned:

```json
{"diagnostics":[]}
```

The reproduction is deterministic because both plugin paths branch only on Promise rejection or output text, not on the nonzero code. It does not require a browser race or a real plugin crash; stubbing `api.plugin` to resolve the envelope above exercises the public worker service behavior.

This worker/compiler integration was introduced in merged PR [#173](https://github.com/samchon/ttsc/pull/173). No open or closed corrective issue owns the nonzero-envelope path.

## Consequence surface

| Plugin outcome | Current transform result | Current lint result | Required invariant |
| --- | --- | --- | --- |
| code 0 with valid payload | transformed source is built | diagnostics are parsed | preserve success |
| nonzero code with stderr | original source may build successfully | stderr may parse as zero diagnostics | surface failure; never report clean success |
| nonzero code with empty stderr | original source may build successfully | empty diagnostics | surface a generic plugin failure |
| rejected Promise | swallowed, then original source builds | empty diagnostics | surface host/plugin failure |
| code 0 with malformed transform JSON | swallowed, then original source builds | not applicable | surface unusable configured transform output |
| diagnostics reported by a completed plugin | existing mapping path | existing diagnostic parsing | preserve user diagnostics |

The transform consequence reaches both `compile` and `bundle`, because both call `runBuildPipeline`; bundle/Execute can therefore run code whose required transform never happened. The lint consequence reaches every UI that interprets an empty list as clean.

## Approach

The architectural owner is the worker's boundary between `@ttsc/wasm` process envelopes and `ICompilerService` results. Define one explicit success/failure interpretation for plugin verbs and map failures into the service's existing error/failure shapes without allowing the downstream build or empty-diagnostics fallback to erase them.

`createWorkerCompiler` currently imports and calls `bootTtsc` directly, so a deterministic service-level regression needs a narrow test seam. Keep `ICreateWorkerCompilerOptions` and the exported public API unchanged: factor the worker construction behind internal boot/result-parser dependencies (or an equivalently private seam), and let the public wrapper continue to supply the real `bootTtsc` and `parseResult`. Keep runtime `@ttsc/wasm` imports in that wrapper so the internal implementation can be exercised with a fake `IBootResult` without building or booting WASM and without exposing test-only options to consumers.

Do not assume every nonzero stderr line is a parseable lint diagnostic, and do not prescribe a UI-specific display mechanism. The invariant is that configured plugin failure remains observable and cannot be converted into a successful untransformed build or clean lint result.

- Add a focused `tests/test-playground` feature package following the existing `@ttsc/testing`/`TestExecutor` pattern, using the private dependency seam and a fake `IBootResult`. The existing `tests/*` workspace glob and root `test:features` filter will discover it; add an explicit Playground lane to `.github/workflows/test.yml` so pull requests run it without a browser or real WASM boot.
- Cover code 0 success, nonzero with structured diagnostics, nonzero with arbitrary stderr, nonzero with no stderr, rejected Promise, and malformed transform output.
- Assert that neither `compile` nor `bundle` invokes or reports a successful build of original source after a required transform fails.
- Assert that lint never returns an empty clean result solely because the plugin failed.
- Preserve explicit plugin disabling (`typiaPlugin: false`, `lintPlugin: false`) and successful diagnostics/output mapping.
- Run the focused Playground worker tests and package build, for example:

  ```bash
  pnpm --filter @ttsc/test-playground start -- --include=playground_plugin_failure
  pnpm --filter @ttsc/wasm build:ts
  pnpm --filter @ttsc/playground build
  ```

- Then run the broader gate:

  ```bash
  pnpm test
  ```

## Coordination

- Campaign tracker: [#664](https://github.com/samchon/ttsc/issues/664).
- #173 is the originating Playground/WASM feature lineage; it is merged and not an active duplicate.
- The current open PR [#657](https://github.com/samchon/ttsc/pull/657), “fix(test): audit every lint corpus source,” touches the native lint corpus harness and fixtures, not `@ttsc/playground`, `createWorkerCompiler`, or WASM plugin envelopes.
- Campaign issue [#669](https://github.com/samchon/ttsc/issues/669), “Discover npm dependencies from executable module syntax only,” affects pre-compilation npm discovery, while this issue owns post-boot plugin result handling. They are safe to address in parallel. Campaign issues [#670](https://github.com/samchon/ttsc/issues/670), “Support valid CommonJS root package.json.exports shapes in the Execute sandbox,” and [#671](https://github.com/samchon/ttsc/issues/671), “Preserve MemFS tree and file-data invariants for successful mutations,” are separate resolver/filesystem owners.

samchon:
## Bundle b6 — playground plugin failures, dependency discovery, sandbox

Part of the `repository-audit-2026-07` campaign. This bundle owns three independent `@ttsc/playground` correctness defects plus a new focused `tests/test-playground` package and a CI lane.

Closes #664, #669, #670

### RA-03 (#664) — surface plugin failures instead of clean lint / untransformed builds
The worker treated a *resolved* WASM plugin envelope as success without inspecting its process `code`. A transform plugin exiting nonzero (or rejecting, or emitting unparseable output) let the pipeline compile the original **untransformed** source and report `success`; a lint plugin exiting nonzero collapsed into an empty (false-clean) diagnostic list.

- Routed plugin envelopes through an explicit success/failure boundary that mirrors the existing `api.build` `code !== 0` handling.
- A failed required transform now returns `type: "error"` from **both** `compile` and `bundle`; `api.build` is never reached, so untransformed JS can never be emitted or executed.
- A lint plugin that fails to run surfaces a single error diagnostic (never an empty clean result). A *completed* linter that exits nonzero because it found violations still returns its parsed diagnostics unchanged.
- Factored the worker core behind an injectable boot/parse seam (`createWorkerCompilerService`) so the whole pipeline is testable against a fake `IBootResult` without building or booting WASM. The public `createWorkerCompiler` API and `ICreateWorkerCompilerOptions` are unchanged.

### RA-11 (#669) — discover npm dependencies from executable module syntax only
`collectModuleSpecifiers` scanned raw source with a regex, so `import`/`require` lookalikes in comments, strings, template contents, and regex literals became phantom package-download requests. Replaced the regex with a small lexical tokenizer that drops comments and treats string/template/regex/number bodies as opaque, so only real code-level specifiers (`import`/`export … from`, side-effect `import "x"`, dynamic `import("x")`, `require("x")`) are collected. `obj.require(...)` and `import.meta` are excluded; scoped names, subpaths, normalization, built-in filtering, uniqueness, and sorting are preserved.

### RA-12 (#670) — support valid CommonJS root `package.json#exports` shapes in the sandbox
`resolvePackageEntry` honored only `exports["."]`. It now interprets all three valid root shapes consistently via a `rootExportTarget` reducer — bare string (`"./index.cjs"`), `"."` subpath table, and bare condition map (`{ require, default }`) — before handing off to the existing conditional picker. `main`/`index` fallback, subpath/wildcard exports, scoped names, relative requires, and JSON loading are untouched; malformed / ESM-only targets still fail with the accurate sandbox error.

### Tests & CI
- New `tests/test-playground` TestExecutor package with focused cases:
  `playground_plugin_failure`, `collect_external_package_names_ignores_non_code`, `create_sandbox_require_supports_commonjs_root_exports`.
- Added an explicit `playground` lane to `.github/workflows/test.yml` (builds `@ttsc/wasm` TS types + `@ttsc/playground`, then runs the suite — no browser, no real WASM boot).

Draft while self-review and local verification complete. Do not merge — the campaign lead merges after recheck.


samchon:
### Self-Review — round 1

Reviewed the whole b6 surface (three source fixes, the new `tests/test-playground` package, the CI lane, and the `build-current` wiring).

**Logic review — no defects found:**
- RA-03: transform failure (nonzero code / rejection / unparseable) returns `type: "error"` from both `compile` and `bundle` with `api.build` never reached; a code-0 empty-stdout transform is a legitimate no-op that still builds; lint failure surfaces one error diagnostic while a completed linter's parsed findings (even on nonzero exit) pass through unchanged; disabled plugins never invoke the plugin verb. Boot rejection now also surfaces in the lint lane instead of a false-clean `[]`.
- RA-11: verified the tokenizer against `import.meta`, `obj.require(...)`, `import x = require("p")`, template/computed specifiers (skipped), division-vs-regex position, and newline-bounded regex scanning — no real specifier is lost and no inert lookalike leaks.
- RA-12: `rootExportTarget` distinguishes subpath tables (any `.`/`./`-prefixed key) from bare condition maps; empty/absent `exports` returns null; ESM-only roots still fail naming the package.

**Finding (style):** the changed source files and the new test files deviate from the repo Prettier config (comment prose-wrap, union/line wrapping, >80-col lines) — real diffs, not just local CRLF. There is no Prettier CI lane, but this would fail a Linux `prettier --check` and hurts reviewability.

**Remediation plan:** run targeted `prettier --write` on exactly the touched files (not the forbidden repo-wide `pnpm format`), rebuild `@ttsc/playground`, and re-run the suite + typecheck to confirm no behavior change.

samchon:
### Self-Review — round 2

Second full pass over the surface.

**Finding (test coverage gap):** RA-11 (#669) explicitly lists *escaped quotes* among the required boundary cases. The collector test already covers `obj.require(...)`, scoped names, subpaths, built-ins, duplicates, relative and template/computed specifiers, but not a string containing escaped quotes that embeds import-like text. The tokenizer decodes `\"`/`\'` escapes while scanning a string literal, so such text should stay inert — I will add an explicit assertion to lock that.

**Re-verified, no change needed:**
- `build-current` full scope now builds `@ttsc/wasm` (build:ts) + `@ttsc/playground`; scoped test-ttsc/test-lint lanes are untouched, and the new `playground` lane uses its own lean build override. This is what lets local `pnpm test` (typecheck + features) resolve the built playground lib.
- Tokenizer division-vs-regex, `import x = require("p")`, and newline-bounded regex/string scanning all behave correctly.

**Remediation plan:** extend `collect_external_package_names_ignores_non_code` with an escaped-quote boundary assertion, re-run the focused suite + typecheck.

samchon:
### Self-Review — round 3 (clean)

Third pass produced no new sound findings. Confirmed:
- No website/README doc becomes incorrect: `packages/playground/README.md` and `website/.../playground.mdx` describe these APIs at the contract level ("race-guarded compile/lint/bundle", "on-the-fly npm dependency installer", "console-capturing execute sandbox"), which stays accurate — the b6 changes are bug fixes that align behavior with the already-documented intent, not contract changes.
- Public API unchanged (`createWorkerCompiler`, `ICreateWorkerCompilerOptions`, `collectExternalPackageNames`, `createSandboxRequire`); the new seam (`createWorkerCompilerService`) is internal.
- Working tree clean; only tracked source/test/CI/build-wiring files changed, no build output committed.

Self-review complete after 3 rounds (round 1: prettier formatting of touched files; round 2: escaped-quote boundary case; round 3: clean).

samchon:
### Final local verification — all green

Run in the worktree at HEAD `f148baf09`:

```
pnpm --filter @ttsc/wasm build:ts              # OK (TS types only, no Go→WASM binary)
pnpm --filter @ttsc/playground build           # OK
pnpm --filter @ttsc/test-playground start      # Success (7 cases)
npx tsc --noEmit -p tests/test-playground/tsconfig.json   # OK
```

Focused per-issue runs (issue-named filters):

```
pnpm --filter @ttsc/test-playground start -- --include=playground_plugin_failure                            # Success (4 cases, RA-03)
pnpm --filter @ttsc/test-playground start -- --include=collect_external_package_names_ignores_non_code       # Success (RA-11)
pnpm --filter @ttsc/test-playground start -- --include=create_sandbox_require_supports_commonjs_root_exports # Success (RA-12)
```

Prettier `--check` clean on all touched source + test files.

CI was self-cancelled at every push per the campaign gate (claim, PR-open, tests+lane, prettier, escaped-quote); the lead re-runs CI on recheck before merge. Not merging.

samchon:
Squash-merged to `master` as `4078876d9251a5089acd151e0589b2f2720d026f` (one `fix(playground)` commit; linked issues auto-close from the PR body). The merge-commit CI was cancelled by exact SHA per campaign discipline.

PR #710, opened 2026-07-15
refactor(shim): replace dead compile guards with completeness probes

### LINKED ISSUE #706: refactor(shim): replace the dead Typia-specific compile guard with general producer closure and runtime probes
**Priority:** P2 prevention integrity. The current guard can remain green while another exposed shim operation is unusable or semantically broken, and it embeds a downstream consumer algorithm in production code that never executes.

### Problem

`packages/ttsc/shim/checker/signature_api_completeness.go` defines an anonymous function assigned to `_`. Its comments explicitly state that it lives in a normal non-test file because CI does not run tests for the nested shim module and that the function is never called.

The dead function copies Typia `plain.classify`'s exact construction-strategy algorithm into the production shim:

- resolve a class constructor type;
- enumerate `new C(seed)` construct signatures;
- inspect minimum/declared arity, return type, parameters, and rest element;
- look up the literal static member `from` and repeat the same checks for `C.from(seed)`.

This is consumer-specific hardcoding and test-only production logic. It proves only that one copied expression chain type-checks. It does not enforce a general invariant that exposed input types are obtainable through the shim, and it does not execute the APIs to prove their semantics.

### Why the current safeguards are insufficient

The mechanical audit currently ensures that reachable/escaping types can be named, but the file itself records the blind spot: it does not prove that a type consumed by an exposed operation can be produced through the shim. That is how `Checker_getMinArgumentCount(*Signature)` shipped while no shim operation could obtain a `*Signature`.

PR #233 introduced the exact dead guard and explicitly named a general producer-closure gate as the ideal follow-up. PR #249 later confronted the same pattern for base-type traversal. Its initial compile guard was removed in favor of `packages/lint/test/shim/base_chain_walk_crosses_generic_boundary_test.go`; the owner explanation states that a compile guard proves only linkage and cannot see the semantic dead-end that caused the bug. The current TypeScript-Go sync skill now requires runtime traversal-completeness probes for that class.

### Affected contract

- `packages/ttsc/shim/checker/signature_api_completeness.go`
- `packages/ttsc/tools/shim_audit` producer/consumer reachability model and baseline
- CI `shim-audit` and `pnpm test:go` coverage of nested shim modules
- All plugin authors, not only Typia, who depend on an obtainable and semantically usable exported checker surface

### Required fix

- Remove the never-called anonymous function from the production shim package.
- Extend `shim_audit` with a general producer/consumer reachability rule: a type accepted by an exposed shim operation must be obtainable from the shim's public roots/returns or be deliberately classified with an explicit rationale.
- Keep the rule independent of Typia, `plain.classify`, the member name `from`, and any consumer repository path.
- Add a ttsc-owned runtime signature-introspection fixture using only public shim operations. It must build a real checker, obtain construct and call signatures, and assert arity, parameter/rest, and return-type endpoints.
- Ensure CI executes the mechanical rule and runtime probe through canonical gates.
- Do not move the dead function to another production file, add another copied consumer chain, or merely rename it as an assertion.

### Acceptance

- A mutation that removes the only producer of an exposed consumed type fails the general audit without referring to `Signature` by name.
- The audit's own fixtures include reachable, unreachable, callback/input-only, and deliberately exempt type cases to prevent false positives.
- The real checker fixture distinguishes `()`, `(x?)`, `(x)`, `(...xs)`, and `(x, ...rest)` signatures and asserts the correct shim-observed endpoint for each.
- Removing or breaking a signature wrapper makes the runtime probe or general audit fail for the correct reason.
- `signature_api_completeness.go` no longer exists and no normal package file contains never-called consumer-algorithm code for CI.
- Run `pnpm --filter ttsc shim:audit:test`, `pnpm --filter ttsc shim:audit`, targeted shim runtime tests, and `pnpm test:go`.

### Provenance and overlap

PR #231 introduced the mechanical audit and documented its then-current closure boundary. PR #233 added the signature APIs and the dead function, while acknowledging that general producer closure remained unimplemented. PR #249 supplies the repository's runtime-probe precedent and intentionally does not add its initially proposed dead compile guard. Searches for `signature_api_completeness`, shim signature producer closure, and the exact filename found no issue or active pull request. Merged campaign PRs #689–#699 are unrelated.

### Non-goals

The issue does not promise that static analysis can predict every future unexported helper a consumer might want. It covers the mechanically knowable public producer/consumer graph and uses runtime probes for semantic traversal endpoints.

samchon:
## Intent

Replace the Typia-specific, never-executed checker-shim compile guard with a general producer-completeness invariant and ttsc-owned runtime probes.

## Implementation

- Remove `signature_api_completeness.go` from production code.
- Extend `shim_audit` from nameability closure to pointer-like compiler-object producer closure.
- Traverse direct and named callback/container contracts, inline interfaces, canonical aliases, source methods, exposed upstream method sets, and nested internal dependency packages.
- Compute reachability over operation inputs and receivers to a fixed point, rejecting rootless function/method cycles.
- Keep exceptional roots separate from the ordinary gap baseline, require non-empty rationales, and never infer them from `-write-baseline`.
- Add generic mutation fixtures for missing producers, callbacks, aliases, containers, methods, dependency types, exemptions, and cycles.
- Add a real Checker runtime fixture covering `()`, `(x?)`, `(x)`, `(...xs)`, `(x, ...rest)`, plus a static call signature through public shim operations only.

Closes #706

## Verification

- `pnpm --filter ttsc shim:audit:test`
- `pnpm --filter ttsc shim:audit`
- `go vet ./...` in `packages/ttsc/tools/shim_audit`
- targeted `TestSignatureIntrospectionReachesRuntimeEndpoints` through `scripts/test-go-lint.cjs`
- `pnpm test:go`
- `pnpm --filter ttsc build`
- `git diff --check`
- ten solo Self-Review rounds; final round clean

All commands above passed. Later audit-only refinements did not change the already revalidated runtime/build surface.

Campaign GitHub Actions remained enabled. Runs for every pushed or PR-mutated campaign SHA were cancelled individually and settled under the exact-SHA two-poll stability gate; local verification and solo Self-Review are the merge gates.


samchon:
Self-Review round 1 ? finding and remediation plan

Finding: the producer graph currently scans exported package functions from shim source, but exposed type aliases also publish upstream exported methods. Their receiver, parameter, callback, and return flows are therefore mechanically knowable and must participate in the same closure. The source spelling also treats aliases such as `ast.Expression = Node` as separate object identities, producing avoidable exemptions.

Plan:
1. Canonicalize parsed flow keys through the loaded upstream `go/types` graph.
2. Add exported method-set receiver/parameter/result flow for every exposed aliased type, including callback variance.
3. Re-run the audit, remove exemptions made redundant by real producers, extend the generic fixture, then repeat the full Self-Review from the complete diff.

samchon:
Self-Review round 2 ? findings and remediation plan

Findings:
1. The source-AST path reverses direction once for an inline interface and again for its method FuncType, so an input interface incorrectly treats method parameters as consumer inputs and method results as producers.
2. Pointer state propagates through slice/array/map/channel containers. A shape such as `*[]T` therefore misclassifies the value element `T` as pointer-like even though only the container is behind the pointer.

Plan:
1. Let FuncType perform the single variance reversal for interface methods.
2. Reset pointer state at every container boundary in both the AST and `go/types` walkers; nested `*T` still establishes its own pointer edge.
3. Extend the generic fixture with an inline-interface input and container boundary, rerun both audit gates, and restart the full review.

samchon:
Self-Review round 3 ? findings and remediation plan

Findings:
1. `scanProducerFile` rejects every declaration with a receiver. Hand-maintained exported shim methods are public operations too, but their explicit parameters and returns can currently escape producer closure.
2. The implementation and sync skill now cover package wrappers plus exposed methods, while several messages still describe the invariant as wrapper-only.

Plan:
1. Include exported source-defined receiver methods, excluding only the receiver itself from input flow for the same reason as upstream method sets.
2. Add a mutation case that removes the only producer used by a source-defined method.
3. Align invariant and failure wording on ?public shim operation?, rerun the audit gates, and restart the full review.

samchon:
Self-Review round 4 ? finding and remediation plan

Finding: producer exemptions now contain both constructible/nullable public roots and explicit host/compiler ownership boundaries (for example protocol payloads and emit-owned sourcemap state). The gate, baseline note, and skill still say only ?external/public root?, which understates the actual reviewed schema and makes future maintenance ambiguous.

Plan: align the three messages on ?reasoned root or ownership boundary?, state that automatic exposure alone is not an exemption, rerun the mechanical gates, then perform a fresh final round over the complete change.

samchon:
Self-Review round 5 ? finding and remediation plan

Finding (correctness / latent false negative): `addExposedMethodFlow` currently treats every result of an exposed alias method as an unconditional producer while deliberately omitting the receiver. A rootless method cycle such as `(*Owner).Clone() *Owner`, or an unobtainable `Owner` method returning a separately consumed `Token`, can therefore make the producer gate pass even though no caller can obtain the receiver needed to invoke the method. The same scan also keys method publication from the broad textual `reachable` set, which includes implementation-only selectors, rather than the aliases/signature types that actually publish callable methods.

Remediation plan:
1. Separate unconditional package/callback producers from receiver-dependent method producers.
2. Record the public receiver dependency for each exposed method and compute producer reachability to a fixed point, so rootless method cycles remain gaps.
3. Feed reasoned producer exemptions in as explicit roots, allowing legitimately caller/host-owned receivers to unlock their method results without exempting downstream objects.
4. Restrict automatic method-set scanning to genuinely public signature/alias types and add mutation tests for an unobtainable receiver, a rooted receiver, and a rootless self-cycle.
5. Re-run the audit fixtures, canonical audit, targeted runtime probe, full Go gate as warranted, then perform a fresh whole-diff Self-Review.

samchon:
Self-Review round 6 ? finding and remediation plan

Finding (correctness / traversal completeness): after making method production receiver-dependent, the whole-diff review found a second type-flow blind spot. Both scanners stop at named callback/container types: the AST path ignores a non-pointer selector or local named type, and the `go/types` path expands named interfaces but not named signatures, slices, maps, channels, arrays, or defined pointers. A public contract such as `type Factory func() *inner.Token; func Register(Factory)` or an upstream named callback used by an exposed method can therefore hide a consumed compiler object and let the zero-tolerance producer gate pass.

Remediation plan:
1. Resolve transparent local type definitions with recursion protection, preserving callback variance and container pointer semantics.
2. Resolve non-pointer internal named types through the loaded upstream type graph.
3. Traverse named callback/container underlyings in the method-set path while continuing to ignore struct/basic underlyings that do not imply required object construction.
4. Add direct-versus-named callback/container mutation cases and recursive-definition safety coverage.
5. Re-run audit fixtures, canonical audit, the real Checker runtime probe, and a fresh whole-diff Self-Review before readiness.

samchon:
Self-Review round 7 ? finding and remediation plan

Finding (correctness / dependency coverage): the named-contract remediation resolves selectors only through `inner`, but `loadInner` currently indexes only the explicitly registered shim packages returned as roots by `go/packages`. Named callback/container types can be declared in imported internal dependencies such as `lsp/lsproto` or `sourcemap`; direct pointers from those packages are seen textually, while non-pointer named contracts are still skipped because their defining package is absent from the lookup map. This leaves the exact traversal class from round 6 partially open.

Remediation plan:
1. Recursively index the loaded internal dependency graph, not only root packages.
2. Fail closed on any indexed internal dependency with load/type errors instead of silently losing its contract.
3. Add an upstream named-callback selector mutation test alongside the local named-type tests.
4. Re-run the audit suite, canonical audit, full Go gate as needed, then restart the whole-diff review.

samchon:
Self-Review round 8 ? finding and remediation plan

Finding (correctness / named-container parity): local defined types currently stop traversal whenever the public signature adds a pointer (`!definition.alias && pointerLike`). That makes `type Batch []*inner.Token; func Consume(*Batch)` invisible even though the equivalent direct form `func Consume(*[]*inner.Token)` is traversed after the container resets pointer state. The named-contract rule is therefore representation-dependent and a trivial alias-to-defined-type refactor can bypass it.

Remediation plan:
1. Traverse local defined underlyings with pointer state reset, matching the existing direct and `go/types` container/callback semantics; aliases continue to preserve pointer state.
2. Add a pointer-to-named-container mutation fixture with its own internal object type.
3. Re-run the audit suite and canonical audit, then restart the clean whole-diff review.

samchon:
Self-Review round 9 ? finding and remediation plan

Finding (correctness / rootless operation cycles): method returns are now receiver-dependent, but package-function outputs and method outputs remain unconditional with respect to their explicit compiler-object inputs. Two public operations such as `AFromB(*B) *A` and `BFromA(*A) *B` therefore mark both objects produced and can satisfy the gate despite having no initial root. A reachable receiver with two mutually dependent method arguments has the same hole.

Remediation plan:
1. Model every public operation as a flow node with receiver exposure, consumed-object requirements, and produced objects.
2. Expose package operations immediately and receiver methods only when the receiver is obtainable; count their inputs as demands, but unlock outputs only when every object input is obtainable.
3. Iterate operations to a fixed point so real producer chains work and rootless function/method cycles fail.
4. Add independent package-cycle and rooted-chain mutation tests, then rerun the canonical audit and whole-diff review.

samchon:
Self-Review round 10 ? clean

I re-reviewed the complete `origin/master...4e9a8b227` change and its consequences after the operation-graph remediation. No remaining correctness, regression, test-quality, documentation, forbidden-hardcoding, or CI-policy findings remain.

The final producer invariant now covers:
- direct, aliased, named, generic-container, callback, inline-interface, and dependency-declared contracts;
- package functions, source methods, and automatically published exposed method sets;
- canonical aliases and nested internal package identities;
- receiver/input-dependent fixed-point reachability, including rootless function and method cycles;
- explicit reasoned roots with empty/stale exemption hygiene;
- real Checker signature traversal at runtime without a copied downstream algorithm.

Verification evidence:
- `pnpm --filter ttsc shim:audit:test` ? PASS on final head
- `pnpm --filter ttsc shim:audit` ? PASS on final head
- `go vet ./...` in `packages/ttsc/tools/shim_audit` ? PASS on final head
- targeted real Checker runtime probe through `scripts/test-go-lint.cjs` ? PASS
- `pnpm test:go` ? PASS (all six runners; subsequent changes are isolated to the standalone audit tool, its fixtures, baseline rationale, and skill text)
- `pnpm --filter ttsc build` ? PASS
- `git diff --check` ? PASS

Campaign Actions remained enabled. Every run created for each pushed or mutated campaign SHA was cancelled individually and observed completed/cancelled under the exact-SHA stability gate.

PR #713, opened 2026-07-16
fix(lint): preserve project ownership through file dispatch

### LINKED ISSUE #712: fix(lint): carry live project state and JIT failures through file dispatch
## Summary

`@ttsc/lint@0.19.0` added project identity, project-rule execution, finalized project status/finding reads, and project-diagnostic transport through #406 / #437. The released contributor contract still cannot carry contributor-owned project state into later file rules or let a file-rule helper reject a just-in-time operation through the same named project result.

Complete that generic lifecycle contract. A project rule must be able to attach state owned only by the current loaded-Program cycle, later file rules must read that same state, and a file-rule helper must be able to change the named result from `passed` to `failed` and add one deterministic project finding before it performs the guarded operation.

This is a post-release successor to #406, not a consumer-specific filesystem feature. The concrete downstream blocker is [wrtnlabs/autobe-mcp#326](https://github.com/wrtnlabs/autobe-mcp/issues/326); [wrtnlabs/autobe-mcp#314](https://github.com/wrtnlabs/autobe-mcp/issues/314) transitively needs the same bound project owner.

## Verified 0.19.0 behavior

Audited against tag `v0.19.0` / release commit `0a53cdb6`.

- `packages/lint/rule/project.go` exposes `ProjectIdentity`, `ProjectRule`, `ProjectContext`, and `ProjectRuleResult`.
- `ProjectRuleResult` contains only `Status` and `Findings`; it has no contributor-owned state or live failure channel.
- `ProjectContext` receives identity and a private reporter only while `ProjectRule.Check` runs.
- file `rule.Context` receives only `ProjectResultReader`; it has no identity, bound state, or project reporter.
- `packages/lint/linthost/project_engine.go` finalizes the result immediately after the project check.
- `Engine.Run` and `program.runLintCycle` snapshot project findings before file-rule dispatch. Even if a reporter were exposed without changing finalization, a finding created during file dispatch would miss the returned CLI/API diagnostic list.

The existing integration coverage proves only that a project rule which fails during preflight is visible to later file rules. It does not cover a project rule that passes, shares a bound owner, and then fails when a file-rule helper detects a changed resource immediately before use.

## Why finalized status alone is insufficient

A generic project/config/I/O contributor needs one binding for the whole Program cycle:

1. the project rule uses `ProjectContext.Identity` to select and validate a root or other project-wide resource;
2. it constructs contributor-owned state containing that binding;
3. later file rules consume the exact same state instead of inferring identity again from a source file, process cwd, environment, or a package-global map; and
4. the shared helper revalidates a mutable external condition immediately before use, reports one owning project error, and makes later dependent helpers observe `failed`.

With 0.19.0, step 2 cannot reach step 3 and step 4 cannot update the cycle. A `LifecycleID -> state` global map is not an acceptable substitute: file contexts do not expose the lifecycle id, cleanup has no public owner, and concurrent public-API, watch, and LSP projects can leak or race. Re-inferring the project from `SourceFile.FileName()` loses the logical selected-project identity that #406 introduced.

## Required public contract

The exact API shape may follow repository conventions, but it must provide all of these semantics:

1. `ProjectRule.Check` can attach one contributor-owned state value to its named result.
2. `Context.ProjectResult(name)` exposes that same state to file rules in the same Program cycle without serialization or a process-global contributor cache.
3. An evaluated result exposes a cycle-scoped `Fail` / `Report` channel that remains live through file-rule dispatch.
4. `Report` atomically changes `passed` to `failed`, deduplicates equal messages, and makes subsequent `ProjectResult(name)` calls observe the failure.
5. `absent`, `off`, and `not_evaluated` results expose neither trusted state nor a live mutation channel.
6. Project findings are finalized after file dispatch but remain ordered before file findings in CLI and structured results.
7. Concurrent file dispatch is race-free. Status transitions and message deduplication are host-owned; a contributor documents and owns any synchronization inside its state value.
8. State and reporters are scoped to one loaded Program/check cycle. Watch/LSP rebuilds and two projects in one process never share them.
9. CLI, check/build/fix/watch, public compiler API, and LSP publish the same final project result. LSP still replaces and clears one project publication at the logical selected-config URI.

An illustrative contributor surface is:

```go
func (projectRule) Check(ctx *rule.ProjectContext) {
  ctx.SetState(newProjectBinding(ctx.Identity))
}

func (fileRule) Check(ctx *rule.Context, node *ast.Node) {
  result := ctx.ProjectResult(demo/project)
  if result.Status != rule.ProjectRulePassed {
    return
  }
  binding, ok := result.State.(*projectBinding)
  if !ok {
    return
  }
  if err := binding.Revalidate(); err != nil {
    result.Report(err.Error())
    return
  }
  // Consume the guarded resource.
}
```

An equivalent typed handle is acceptable. The host must not interpret or serialize contributor state.

## Acceptance criteria

- [ ] A project rule attaches state and a later file rule receives the same value in the same cycle.
- [ ] A file rule changes an initially passed project result to failed and emits one detached project finding.
- [ ] A later rule/helper observes failed and skips its dependent operation.
- [ ] Repeated and concurrent reports of the same message produce one deterministic finding with no data race.
- [ ] Multiple distinct messages are sorted deterministically.
- [ ] The final project finding precedes file findings even though it was created during file dispatch.
- [ ] Disabled, absent, and not-evaluated project rules expose no usable state/reporter.
- [ ] Panics and failed preflight retain the current deterministic behavior.
- [ ] Two projects in one process receive different state objects and reporters.
- [ ] Watch rebuilds replace the old cycle state instead of retaining it.
- [ ] CLI, public API, watch, and LSP surface the just-in-time project finding once and clear it when the next cycle is clean.
- [ ] Package-metadata contributor discovery works without `compilerOptions.plugins`.
- [ ] Public documentation explains state ownership, snapshot/live-result behavior, concurrency responsibility, and the reporter lifetime.

## Non-goals and forbidden shortcuts

- No AutoBE-specific path, rule name, workspace marker, or containment algorithm.
- No host dependency graph between project and file rules.
- No contributor process-global lifecycle map, `sync.Once`, environment lookup, process cwd lookup, or first-source-file identity inference.
- No serialization of arbitrary contributor state into CLI/LSP protocol payloads.
- No copying project findings onto every source document.
- No requirement for a TypeScript transformer `plugins` entry.

samchon:
## Intent

Closes #712.

Project contributors can now attach an opaque, Program-scoped binding during their project check. File rules receive the exact binding through `ProjectResult.State` and can report a just-in-time project failure through the live result without reconstructing ownership from filenames or keeping process-global lifecycle maps.

## Scope

- add `ProjectContext.SetState` and live `ProjectRuleResult.Report` / `Fail`
- keep project status, state, message deduplication, and finalization concurrency-safe
- finalize project diagnostics after file dispatch while preserving project-before-file output ordering
- isolate state and reporters between Program cycles and make retained reporters inert
- cover CLI, public API, watch, and LSP publication/clearing through package discovery
- document the snapshot and contributor-owned synchronization contract

## Deferred

- None.

## Local verification

- `pnpm format`
- `pnpm build:current`
- `pnpm test:typecheck`
- `node scripts/test-go-lint.cjs`
- `pnpm test:go`
- `TTSC_TEST_DIR=native-plugins/corpus-misc pnpm --filter @ttsc/test-ttsc start --include=project_rule_lifecycle_surfaces`

## Known gap

The Go race detector could not run locally because this Windows environment has CGO enabled only through an unavailable `gcc` toolchain. The concurrent live-report test and the repository's complete Go test runners pass locally; CI remains the authoritative race-capable environment if configured.


samchon:
Self-Review round 1 found no implementation defect, but it found one acceptance-coverage gap and one stale lifecycle phrase. Follow-up commit 7eff3ebad now pins deterministic project-preflight panic status/finding ordering and clarifies that opaque state contents are not copied. Revalidated with pnpm format and 
ode scripts/test-go-lint.cjs (pass, linthost 171.058s). The next complete review round starts from the updated diff.

samchon:
Self-Review rounds 2-4 completed on the full updated diff. Round 2 corrected the contributor-vs-host state ownership wording. Round 3 removed a pre-use value copy of the cycle's sync.Once by making evaluateProject return the cycle pointer directly. Round 4 rechecked the complete API, host, command, concurrency, test, E2E, and documentation surface and was clean. Follow-up commit: d72f46d62. Revalidated with pnpm format and 
ode scripts/test-go-lint.cjs (pass, linthost 227.504s).

samchon:
CI follow-up: the first nestia attempt failed because the hosted runner exhausted disk space while Go wrote its build cache (
o space left on device), not because of a compile or test assertion. I reran only the failed workflow on a fresh runner. Attempt 2 passed both Build nestia and Test nestia in 48m54s. Final latest-SHA rollup: 50/50 checks successful, 0 pending, 0 failed.

PR #718, opened 2026-07-16
feat(ttsc): host-owned reference graph for sound cache invalidation

### LINKED ISSUE #716: transform protocol: host-owned reference graph for sound cache invalidation (tsc --incremental semantics)
## Problem

Bundler caches invalidate a module only when its registered inputs change: webpack's filesystem cache (what Next.js stores under `.next/cache`) and watch-mode graphs both compare a per-module `fileDependencies` snapshot, and Turbopack's loader contract works the same way. A type-driven transform's output depends on files beyond the module itself — the files where the consulted types are declared — so those files must be registered as the module's inputs, or every cache layer serves stale generated code.

#220 (closing #214) shipped the consumer half: the transform envelope's optional `dependencies` field flows through `@ttsc/unplugin` → `addWatchFile` → (webpack) `addDependency` → `fileDependencies`, covering both watch mode and persistent caches. Production of the field, however, is opt-in per plugin — and no plugin produces it today. typia's transform host emits only `diagnostics` + `typescript`, so the channel carries nothing and the stale-cache bug class ships by default for every type-driven plugin. Field report: samchon/typia#2092 (Next.js on Vercel; the user deletes `.next/cache` before every build as a workaround).

## Evidence

Reproduced on npm artifacts (ttsc 0.19.1 + @ttsc/unplugin 0.19.1 + typia 13.1.1), webpack 5 with `cache: { type: "filesystem" }` + esbuild-loader; `src/mytype.ts` exports `interface MyType { id: string }`, `src/index.ts` runs `typia.validate<MyType>(...)`:

1. Build → generated validator works.
2. Add a required `age: number` to `MyType`, rebuild **with the cache kept** → the consumer module is served from cache; the bundle contains no `age` check; `{"id":"x"}` still validates as success.
3. Delete the cache, rebuild → `$input.age` is correctly rejected.

A one-line fixture loader that registers `src/mytype.ts` via `this.addDependency()` for `index.ts` makes step 2 rebuild correctly — the shipped consumer half works end-to-end; production is the only gap.

## Design: the invariant and who owns which knowledge

The only correctness invariant a cache has is **key ⊇ actual inputs of the transform**. Every scheme is an approximation of that input set, and the design question is who owns the knowledge:

| Input kind | Owner | Mechanism |
|---|---|---|
| Language-semantic inputs (types in other files) | compiler host (`driver`) | reference graph emitted always, plugin-agnostic |
| Extra-programmatic file inputs (plugin config, external schemas) | each plugin | self-reported `dependencies`, **union** — widening never breaks soundness |
| Non-file inputs (env, time, network) | each plugin | hermeticity declaration (`volatile`) → excluded from caching; no file-dependency scheme can represent these |
| Precision below the language bound | plugin semantics | **deferred**: a later opt-in with an explicit completeness contract |

Two grounding facts fix the shape of the host part:

- **Observed-dependency tracking is unavailable here.** ttsc plugins hold raw tsgo AST/checker pointers (no mediated query pipeline as in Roslyn incremental generators or salsa), so the host cannot trace what a plugin actually read. Under exactly this constraint, TypeScript's own incremental engine chose the language-semantic upper bound — per-file resolved references (`referencedMap`), files affecting the global scope, and the options hash — and that bound is sound by TS name resolution: any symbol a file can reference is reachable through its import/reference closure or is ambient. `tsc --incremental` semantics is therefore the domain's orthodoxy, not a compromise.
- **The protocol should carry the minimal sufficient statistic: direct edges, not per-file transitive closures.** `tsbuildinfo` stores `referencedMap` for the same reason. Flattening is a consumer-format concern (flat `fileDependencies` snapshots), so it belongs in the adapter — the same boundary at which `gcc -MD` depfiles flatten header graphs for `make`.

## Approach

1. **driver (Go):** expose the program's reference graph — per file, its direct resolved references (imports, exports, `/// <reference>`, `import type` included); the set of files contributing to the global scope (ambient declarations, script files, `typeRoots` entries); and the config chain (tsconfig plus its `extends` ancestry).
2. **driver (Go):** provide the envelope-encoding helper that stamps a `graph` section (`edges`, `globals`, `configs`) into every transform envelope. `cmd/ttsc api-transform` (the no-plugin path) uses it directly.
3. **Plugin path:** the envelope is encoded inside the plugin binary, so graph emission rides the same helper — automatic for every plugin that routes its envelope through the driver SDK encoder. Adoption per existing plugin is one mechanical change; typia's is tracked in samchon/typia#2098.
4. **ttsc (TS):** parse and forward `graph` on `ITtscCompilerTransformation`, with the same malformed-input tolerance as `dependencies`.
5. **@ttsc/unplugin:** per transformed file `F`, register `reach(edges, F) ∪ globals ∪ configs ∪ dependencies[F]` through the existing `addWatchFile` hook. Honor a plugin's `volatile` declaration by bypassing caching for the affected files.
6. **Out of scope, deliberately:** the precise-replacement contract (a plugin declaring its reported list complete so it can narrow the baseline). `dependencies` keeps pure union semantics until that contract is specified separately.

## Consequence surface

- All unplugin adapters (webpack, rspack, vite, rollup, rolldown, esbuild, farm, bun, turbopack, next) share the derivation, since it lives in `core/transform.ts`'s hook plumbing.
- The same graph data later allows replacing the whole-project rehash validation in the unplugin transform cache (`matchesCachedSource`) with per-file checks — a follow-up optimization, not part of this issue's acceptance.
- Envelope shape is backward compatible: `graph` is a new optional section; consumers that predate it ignore it; the shipped `dependencies` field is unchanged.

- `api-transform` envelope contains `graph` whose edges include type-only import edges and whose `globals` include ambient declaration files; `configs` lists the tsconfig `extends` chain.
- Fixture-plugin test: with a producer emitting `graph`, the webpack filesystem-cache reproduction above passes **without deleting the cache** (change the type file → consumer module rebuilds → fresh output).
- Watch-mode test: editing a file reachable only through a type-only edge re-runs the consumer's loader.
- Existing dependency-forwarding tests (`test_transformttsc_forwards_plugin_dependencies_to_the_watch_hook`, adapter survival tests) keep passing — union semantics preserved.

## Coordination

- samchon/typia#2098 — typia-side producer: precise per-file consulted-declaration lists now (standalone interim fix through the shipped consumer), SDK encoder adoption once this lands.
- History: #214 (request), #220 (consumer half, shipped v0.15.3).
- Field report: samchon/typia#2092.

samchon:
## Intent

Implement samchon/ttsc#716: make the compiler host own the language-semantic input set of a transform so bundler caches (webpack filesystem cache, watch graphs, Turbopack `fileDependencies`) stop serving stale generated code by default.

Closes #716.

## Scope (issue approach items)

1. **driver (Go):** expose the program's reference graph — per-file direct resolved references (imports, exports, `/// <reference>`, type-only included), files contributing to the global scope (ambient declarations, script files, `typeRoots` entries), and the tsconfig `extends` config chain.
2. **driver (Go):** envelope-encoding helper that stamps a `graph` section (`edges`, `globals`, `configs`) into transform envelopes; `cmd/ttsc api-transform` uses it directly.
3. **Plugin path:** the linked-plugin generic host (`utility.RunTransform`) rides the same helper, so first-party linked plugins emit the graph automatically; external sidecars adopt with one mechanical change through the driver SDK.
4. **ttsc (TS):** parse and forward `graph` on `ITtscCompilerTransformation` with the same malformed-input tolerance as `dependencies`.
5. **@ttsc/unplugin:** per transformed file `F`, register `reach(edges, F) ∪ globals ∪ configs ∪ dependencies[F]` through the existing `addWatchFile` hook plumbing (covers all adapters). Honor a plugin `volatile` declaration by bypassing caching for the affected files.

Deliberately **out of scope** (declared by the issue): the precise-replacement/narrowing contract. `dependencies` keeps pure union semantics.

## Verification plan

Local tests are the gate (campaign CI is cancelled per exact SHA and recorded in `.wiki/dependency-invalidation/ci-state.md`):

- Go unit tests in `packages/ttsc` (graph extraction, envelope stamping, api-transform acceptance: type-only edges, ambient globals, `extends` chain).
- `tests/test-unplugin` (graph-driven watch registration, webpack filesystem-cache scenario passing without cache deletion, watch-mode invalidation through a type-only edge, existing dependency-forwarding tests unchanged).

Solo Self-Review rounds will be recorded as comments on this thread before each remediation.

https://claude.ai/code/session_01U3kDYEkHutvSMcHPBXhd7P


samchon:
## Self-Review — Round 1 (findings before remediation)

Full fresh pass over the complete base→head diff (Go producer, shim linknames, TS envelope parsing, unplugin derivation, fixtures, tests, docs), plus the consequence surface beyond the diff (wasm host, Metro adapter, resident serve host, bun adapter, case-sensitivity of key joins).

**Findings to apply in this round:**

1. **Stale doc comment in `packages/unplugin/src/core/transform.ts`** — `matchesCachedSource`'s comment still says out-of-walk inputs invalidate "through the reported `dependencies` → `addWatchFile`" only; the graph now carries that duty too. Will update the comment to name both channels.
2. **Missing failure-path coverage for `graph` forwarding** — `IFailure.graph` is declared and the forwarding code is shared with the success path, but no test pins that a failed transform still carries the host graph (the field matters exactly then: a module that fails type-check must still invalidate through its edges once fixed). Will add `test_ttsccompiler_transform_failure_carries_reference_graph` (project with a type error, no plugins → native host envelope → `type: "failure"` with a populated `graph`).

**Verified as non-issues (no action):**

- **Pre-mutation ordering** — both envelope hosts compute the graph before linked plugin hooks mutate ASTs; pinned by `transform_graph_reflects_source_before_plugin_mutation_test.go`.
- **wasm host (`packages/wasm/host/api.go`)** — has its own transform envelope without `graph`; its consumers are in-browser playgrounds with no bundler cache, so stamping there is outside this issue's consumer set. Left untouched deliberately.
- **Metro (`packages/metro`)** — separate adapter with a per-file keyed cache and a documented `--reset-cache` caveat (#255 territory); the issue scopes registration to `@ttsc/unplugin`'s `addWatchFile` plumbing. Left untouched deliberately.
- **Resident serve host (`utility/serve.go`)** — replies are per-file text, not envelopes; no graph slot in that protocol today.
- **Bun adapter** — its no-op transform context lacks `getNativeBuildContext`; `markVolatile` uses optional chaining throughout, so volatile declarations degrade to no-ops there instead of throwing.
- **Windows drive-case joins for absolute edge keys** — same comparison semantics as the shipped `dependencies` channel (#220); both sides derive from OS-provided paths. No behavior change introduced.

**Local verification at this state:** full Go suite green (`scripts/test-go.cjs`, exit 0), full `tests/test-unplugin` suite green (exit 0, includes the webpack filesystem-cache scenario passing without cache deletion, the stale-control baseline, and the watch-mode type-only-edge rebuild), `tests/test-ttsc` targeted envelope tests green, test-package typechecks green for every suite except a pre-existing `tests/test-lint` failure unrelated to this diff (asserts against `@ttsc/lint`'s locally generated `lib/index.d.ts`; no lint source is touched here — deferring to CI's Linux run as arbiter).

samchon:
## Self-Review — Round 2 (findings before remediation)

Second complete from-scratch pass over the base→head diff after the Round 1 remediation, including the fixture-generated Go sources, the webpack scenario helpers, the watch-mode settle logic, and both docs pages as rendered diffs.

**Finding to apply in this round:**

1. **Broken example in `website/src/content/docs/development/reference/driver-api.mdx`** — the Reference graph snippet assigns `out.TypeScript[...] = text` with `text` never defined, so the illustration does not compile as shown. Will inline the printer call (`shimprinter.EmitSourceFile(printer, file)`) so the example is self-contained.

**Design note recorded for the thread (no action):** the issue mandates honoring "a plugin's `volatile` declaration" without fixing where the declaration lives. This PR places it on the transform envelope (`volatile: string[]`, keyed like `typescript`), because the envelope is the protocol channel every plugin path already shares, it keeps per-file granularity, and it inherits the same malformed-input tolerance as `dependencies`. Consumers: the adapter's transform cache is bypassed for declared files and the module is marked uncacheable where the bundler exposes that control (webpack/rspack `getNativeBuildContext` loader context, Turbopack loader `cacheable(false)`); adapters without such a control (bun, esbuild, rollup family) degrade to the cache-bypass half.

**Verified as non-issues in this round:**

- Watch-mode test settle logic cannot double-settle the promise into a failure (`finish` clears the timeout first; a second callback after resolution is a no-op on an already-settled promise).
- The stale-control webpack scenario shares its fixture builder with the positive scenario, so the two can only diverge on the single `emit-graph` plugin entry — the channel under test.
- The fixture's `emit-volatile` timestamp suffix keeps the `goUpper` rewrite observable (`"PLUGIN:<nanos>"`) without weakening any other operation's assertions.
- `graphFromConfig` / `stringListValue` / `boolValue` in the shared fixture tolerate missing or non-string config members, mirroring the host-side tolerance they exercise.

**CI status at time of writing:** bun/vscode/website/source-map green on the head SHA; test/build/typia/nestia/plugin-cache/experimental still running. The previous SHA's `typia` lane failed inside typia's own `tests/test-utils-automated` suite (`validateEquals` "Bug on superfluous"); typia master received four merges in the hour before that run and samchon/typia#2099 (fix/automated-runner-failures) is open against exactly that suite, while this diff carries no emit/validator semantics. A control dispatch of the same workflow on unmodified `master` (run 29499766568) is in flight to prove the failure is upstream typia churn; its result will be recorded here before merge.

samchon:
## Self-Review — Round 3 (findings and remediation)

Third complete pass over the base→head diff, this time including the CI/packaging gates the change interacts with (the `shim-audit` job, the `test` workflow's suite matrix, packaging surfaces, and the public API compatibility of `ttsc` and `@ttsc/unplugin`).

**Finding (remediation included in this round's push):**

1. **`shim-audit` gate fails on the new linkname target package.** `shim/compiler/incremental.go` blank-imports `internal/execute/incremental` (required so the linknamed symbols link into consumer binaries), but the audit only loads type information for packages reachable from `shimDirs`, so the import scan aborted with "no loaded type information". Remediation, applied and verified locally (`shim_audit: OK`):
   - `tools/shim_audit/main.go` gains a documented `auditOnlyInternal` list — internal packages referenced by shim sources (linkname targets) without a shim directory of their own — folded into the loader's expected set, so a future linkname to another engine package is a conscious one-line add instead of a silent skip.
   - Loading the package surfaces two ratcheted findings (`execute/incremental.ComputeHash` / `GetMTime`, buildinfo hashing helpers that are not plugin surface); accepted into `baseline.json` via the documented `-write-baseline` ratchet.

   This was applied ahead of this comment because the head SHA's `test` workflow was failing on exactly this gate; the thread records the reasoning chronologically all the same.

**Verified as non-issues in this round:**

- Public API compatibility: `ITtscCompilerTransformation` gains only optional fields; `TtscTransformHooks` / `TtscCachedProjectTransform` gain optional members; no existing envelope field changes shape. Consumers predating `graph`/`volatile` ignore them (protocol compatibility rules unchanged).
- No `go.mod`/`go.sum` changes were needed: the linkname target lives in the already-pinned typescript-go module.
- `tests/test-unplugin/package.json` adds only a `webpack` devDependency; no `version` fields change anywhere; `pnpm-lock.yaml` delta is the webpack subtree only.
- Test isolation: new scenarios build on `TestProject.tmpdir` roots with the shared plugin-binary cache; the webpack cache dir lives under the project-walk-ignored `.cache/`, output under `out/`.

**Local verification ledger (cumulative):** full Go suite green; full `tests/test-unplugin` suite green (63 scenarios incl. the webpack filesystem-cache acceptance, stale control, and watch-mode type-only-edge rebuild); `tests/test-ttsc` full suite in progress locally (targeted envelope/failure tests already green); `shim:audit` OK; per-package `tsc --noEmit` green for every touched package and suite.

samchon:
## Self-Review — Round 4 (findings before remediation)

Fourth complete pass, covering the Round 3 audit-tool commit and re-verifying the full surface, with the external evidence now in.

**Finding to apply in this round:**

1. **Inaccurate clause in the new `auditOnlyInternal` doc comment** (`tools/shim_audit/main.go`): it claims audit-only packages "join no completeness gate", but loading them does enter their exported symbols into the ratcheted reachable-gap scan — that is exactly why `ComputeHash`/`GetMTime` landed in `baseline.json`. Will correct the comment to say their surface is baseline-managed like any other loaded package, so the next reader is not misled about the gate's behavior.

**External evidence closed out:**

- **typia lane:** the head SHA's `typia` workflow now passes (run on `65fc3bbc8`), and the control dispatch of the same workflow on unmodified `master` (run 29499766568) also passes against current typia HEAD. The single earlier failure on `978fdb512` cloned typia mid-churn (four merges landed on typia master within the surrounding hour, including the tagged-union backtracking fix #2077 whose domain matches the failing `validateEquals` case). Conclusion: upstream-transient, not this diff.
- **CI on `65fc3bbc8`:** build, test (incl. shim-audit and the Linux typecheck of every suite), bun, vscode, website, source-map, experimental, plugin-cache, typia all green; nestia still running and will re-trigger on this round's push.
- **Local `tests/test-ttsc` full suite:** every transform/API scenario green including the new envelope tests. Three failures are the pre-existing Windows-only fake-consumer-binary class (`test_ttsc_no_plugin_build_honors_tsconfig_no_emit`, `test_ttsc_no_plugin_emit_invokes_tsgo_once`, `test_ttsx_executes_javascript_emitted_by_the_consumer_local_tsgo`): the fixture writes a Node script named `tsc.exe`, which Windows cannot spawn (`spawnSync UNKNOWN`). Untouched by this diff; the Linux `test` lane covering them is green on the head SHA.

No other finding survived this pass. After the comment fix below, the next full pass must come out clean for merge.

samchon:
## Self-Review — Round 5 (clean round)

Fifth complete pass over the final base→head diff (8 commits, 41 files): file inventory matches the intended surface exactly, no `version` field changes anywhere, `pnpm-lock.yaml` delta is the webpack devDependency subtree only, and the working tree is clean. No sound improvement or issue candidate survived this pass — the review terminates per the stop rule.

**Final gates, all green:**

- **CI on `bf4d4daaf` (10/10):** build, test (Go suites, shim-audit, per-suite typechecks, feature suites), bun, vscode, website, source-map, experimental, plugin-cache, typia, nestia — all success.
- **typia lane:** green here and on the unmodified-`master` control dispatch; the one historical failure was upstream typia mid-churn (Round 4 evidence).
- **Local:** full Go suite, full `tests/test-unplugin` (webpack filesystem-cache acceptance without cache deletion, stale-control baseline, watch-mode type-only-edge rebuild, volatile semantics, all existing dependency-forwarding and adapter survival tests), `tests/test-ttsc` full suite (three pre-existing Windows-only fake-`tsc.exe` spawn failures aside, covered green by the Linux `test` lane), `shim:audit` OK.

Acceptance mapping (issue #716):

| Acceptance criterion | Proof |
| --- | --- |
| `api-transform` envelope graph: type-only edges, ambient globals, `extends` chain | `TestAPITransformEnvelopeCarriesReferenceGraph`, `TestTransformGraphReportsTypeOnlyEdgesGlobalsAndConfigs` |
| Fixture-plugin webpack filesystem-cache scenario passes without cache deletion | `test_webpack_filesystem_cache_rebuilds_through_a_type_only_edge` (+ stale control pinning the reproduction) |
| Watch-mode invalidation through a type-only edge | `test_webpack_watch_rebuilds_through_a_type_only_edge` |
| Existing dependency-forwarding tests keep passing (union semantics) | full test-unplugin suite green incl. `test_transformttsc_forwards_plugin_dependencies_to_the_watch_hook` and adapter survival tests; `test_transformttsc_registers_graph_and_dependencies_as_a_union` pins the union |
| Out of scope honored | no precise-replacement/narrowing contract; `dependencies` keeps pure union semantics |

Proceeding to merge under the standing autonomous mandate.

Issue #720, opened 2026-07-16
plugin protocol: specify the completeness contract for dependency narrowing


samchon:
## Problem

Once #716 lands, the transform envelope carries two dependency sources with union semantics: the host-owned `graph` baseline (sound upper bound, `tsc --incremental` semantics) and plugin-reported `dependencies` (extra-programmatic inputs; widening only). Union can never narrow, so a plugin with precise consulted-file knowledge (typia's per-file consulted-declaration lists, samchon/typia#2100) cannot reduce over-invalidation below the language-semantic baseline: a change in any file reachable through a module's import closure re-runs its loader even when the transform never consulted that file.

This is correct but coarser than necessary. Reducing it requires an explicit contract by which a plugin declares its reported per-file list *complete*, shifting correctness responsibility for that file from the host baseline to the plugin — the Bazel declared-inputs model, where under-declaration is a plugin bug by contract rather than a silent platform failure.

## Design questions this issue owns

- **Marker shape:** per-file vs whole-envelope completeness declaration (e.g. `dependenciesComplete`), and how a mixed envelope (some files precise, some not) behaves.
- **What completeness replaces:** presumably `reach(edges, F)` for marked files only. The config chain must remain a universal input regardless (compiler options affect generated code independently of what the plugin consulted). Decide whether global-scope files stay universal or must appear in a complete list when consulted.
- **Multiple plugins:** semantics when several transform plugins contribute to one file — completeness is per (plugin, file); the consumer can only narrow when every contributing plugin declared complete for that file, otherwise union with the baseline.
- **Failure story:** an incomplete-but-declared-complete list is a plugin defect; document the responsibility transfer and diagnostic strategy explicitly.
- **Consumer semantics:** `@ttsc/unplugin` derivation rule update and cache-validation implications.

## Approach

Design first: specify the contract in the plugin-protocol documentation and the envelope types, then implement host parsing plus the unplugin derivation change, with fixture-plugin tests for complete, incomplete-marked (defect surface), and mixed envelopes. No producer is required to adopt it for correctness — adoption is a per-plugin optimization (typia's is the first candidate).

## Acceptance and verification

- Protocol documentation states the responsibility transfer and mixed/multi-plugin semantics unambiguously.
- Fixture tests: a marked-complete file is invalidated only by its declared inputs plus universal inputs; unmarked files keep baseline behavior; mixed envelopes compose correctly.
- Existing union-semantics tests keep passing for producers that never declare completeness.

## Coordination

[...]

PR #722, opened 2026-07-16
feat(metro): sound cache invalidation from the reference graph

### LINKED ISSUE #721: @ttsc/metro: sound cache invalidation from the reference graph
## Problem

Metro's babel-transformer contract has no per-file dependency registration — nothing like a webpack loader's `addDependency`. Its transform cache keys on each file's own content plus a static `getCacheKey`. `@ttsc/metro` (#255) therefore shipped with a documented `--reset-cache` caveat: when file A's generated code depends on a type declared in file B, editing only B changes neither A's content nor the static cache key, so Metro serves stale transformed A. This is the same bug class the dependency/graph channel closed for the web bundlers (samchon/typia#2092), acknowledged in #255 as its riskiest open question and left unsolved in v1.

## Evidence

- #255's design record, open question 1: "Metro's transform cache is keyed on per-file content + the static `getCacheKey` … so Metro can serve **stale transformed A**"; the reference implementation (`@elliots/metro-transformer-typical`) does not solve it either.
- #718 landed the host-owned reference graph (`driver.NewTransformGraph`: per-file direct resolved edges including type-only, global-scope files, config chain — `tsc --incremental` semantics), so the language-semantic input bound is now available to every driver-based host, including the resident compiler `@ttsc/metro` runs per worker.
- The web-bundler consumers register per-file watch inputs; Metro cannot, which is why this surface was consciously excluded from #718's scope (recorded on that PR's thread).

## Consequence surface

React Native / Expo users of every ttsc plugin (typia, nestia, …). Two distinct staleness layers must both be examined — fixing one without the other still serves stale output:

1. **Metro's persistent transform cache across runs** — the `getCacheKey` dimension.
2. **The resident compiler's loaded program within a session** — if the per-worker program does not reload changed files, even a fresh (cache-missing) transform of A answers from a stale program.

`getCacheKey` call timing (once per worker? per transform? per config change?) governs what a fingerprint can and cannot fix — verify against Metro's actual behavior, do not assume.

## Approach

The invariant, not a prescription: **Metro's cache key for a transformed file must incorporate every input that can influence that transform's output.** With no per-file hook in Metro's contract, the sound encoding available today is a project-level type-graph fingerprint folded into the cache key — derived from the reference graph's file set (#718 primitives). That is coarse: any type-relevant change invalidates every transformed file. The granularity loss is forced by Metro's contract, not chosen; it is still strictly better than a manual `--reset-cache`, and it is sound.

Investigate first, then implement:

- Metro's cache-key semantics and cache layers (in-memory vs disk store, dev-server watch vs cold `bundle` runs).
- The resident program's freshness model within a watch session (layer 2 above).
- Whether any Metro mechanism (including unstable/experimental APIs) offers per-file dependency registration; prefer it if real, and record the finding either way.

- Two-run reproduction: transform a project, change only the type's file, re-run **without** `--reset-cache` → the bundle carries the regenerated output.
- Watch-session behavior verified and documented (fixed, or its boundary stated precisely).
- Existing `@ttsc/metro` tests stay green; the metro guide's `--reset-cache` caveat is replaced or narrowed to whatever residue remains.

## Coordination

- #255 (shipped v1 and its open question 1), #718 / #716 (reference graph), samchon/typia#2092 (bug class).
- Independent of #720 (narrowing contract) and of the web-bundler adapters.

samchon:
Implements samchon/ttsc#721: sound cache invalidation for `@ttsc/metro`, derived from the host-owned reference graph (#718 primitives).

## Intent

`@ttsc/metro` v1 (#255) shipped with a documented `--reset-cache` caveat: Metro keys its transform cache on per-file content plus a static transformer key, so editing only a depended-upon type in another file leaves the dependent file's cache entry valid and Metro serves a stale transform. This PR makes Metro's cache key incorporate every input that can influence a transform's output, replacing the manual `--reset-cache` step for cross-run staleness, and closes the matching in-session gap in the shared transform core.

## Investigation record (mandated by the issue, verified against Metro 0.87 source)

**1. Metro's cache-key semantics and cache layers.**

- `getCacheKey` is called once per Metro instantiation (dev-server start or cold `metro bundle` run), on the main process, synchronously: `Server` -> `Bundler` -> `Transformer` constructor -> `getTransformCacheKey` (`metro/src/DeltaBundler/getTransformCacheKey.js`) -> `require(transformerPath).getCacheKey(transformerConfig, { projectRoot })`. For the default `transformerPath` (`metro-transform-worker`), its `getCacheKey` folds in `require(babelTransformerPath).getCacheKey({ projectRoot, enableBabelRCLookup })` — that is the `@ttsc/metro` hook. It is never re-called per file, per worker, or per rebuild within a session; `metro-babel-transformer`'s own docs state "called once by the main thread (not on worker instances)".
- The result becomes `Transformer._baseHash`; each file's cache key is `stableHash([baseHash, posix-relative path, transform options...]) + sha1(file content)` (`Transformer.transformFile`).
- Cache layers: the per-file key is looked up in `config.cacheStores` — default a single persistent `FileStore` at `os.tmpdir()/metro-cache`, shared across runs. Within a dev-server session, transformed modules additionally live in the in-memory DeltaBundler graph. `--reset-cache` clears the stores at `Server` construction. `unstable_autoSaveCache` persists only the file map (`metro-file-map`), not transform outputs, so a restart re-evaluates `getCacheKey` and re-keys every lookup.

**2. Resident-compiler freshness within a session.**

The per-worker "resident compiler" in `@ttsc/metro` is the module-level `createTtscTransformCache()` singleton. Its entries are revalidated on every transform call by re-hashing every file under the project root (`matchesCachedSource`), and a miss constructs a fresh `TtscCompiler` and recompiles — there is no long-lived stale program for in-project inputs. The verified residual gap: inputs outside the project-root walk (files outside the project root, or under walk-ignored directories such as `node_modules`) were not in the validation universe, so a worker could serve a project transform computed against a stale out-of-walk input for its whole lifetime. Web bundlers never hit this because the unplugin factory clears the cache on every `buildStart`; Metro (and the Turbopack loader, and Bun) hold the cache for the process lifetime. This PR closes that gap in the shared core (see below).

**3. Per-file dependency registration in Metro.**

None exists, including unstable APIs, as of metro 0.87 / metro-transform-worker 0.87 / metro-babel-transformer 0.87. The `BabelTransformer` contract is exactly `{ transform, getCacheKey? }`; the transform result (`{ ast, functionMap?, metadata? }`) has no dependency channel; the worker-level `dependencies` output is import extraction for module-graph resolution (adding entries would add modules to the bundle). The DeltaBundler re-transforms only watcher-reported changed files (`DeltaCalculator`), so no cache key can force a re-transform of an unchanged dependent file within a running session. The project-level fingerprint is therefore the sound encoding Metro's contract admits, exactly as the issue anticipated.

## Mechanism

**Cross-run soundness — a project fingerprint folded into `getCacheKey`** (`packages/metro`):

- `getCacheKey` now folds a fingerprint with two parts:
  1. a hash of every input file under the fingerprint roots (Metro's `projectRoot` from the cache-key options, plus the resolved tsconfig's directory when it lies outside), using the same directory-walk semantics as the transform core's own cache validation; and
  2. a re-hash, at key time, of every recorded out-of-walk reference-graph member (node_modules declarations, monorepo sibling sources, out-of-root tsconfig `extends` ancestry) from the snapshot the transformer maintains.
- The snapshot lives at `<projectRoot>/node_modules/.cache/ttsc-metro/`. Each worker owns a uniquely named snapshot file (no write races); `withTtsc`, which runs in the single config process before workers exist, compacts them into the main snapshot. Snapshot identity carries a random id preserved across merges, so a deleted-and-recreated snapshot can never alias an older key epoch. When no snapshot is readable at key time, a random nonce is folded instead: that run shares no cache with any other run, which is the sound degradation.
- The out-of-walk members are derived from the existing `addWatchFile` hook plumbing (#718): plugin-reported `dependencies` unioned with the reference graph's reach, globals, and configs.

**In-session soundness — out-of-walk revalidation in the shared core** (`packages/unplugin`):

- `TtscCachedProjectTransform` now records the hash of every out-of-walk reference-graph member at transform time, and cache validation re-hashes them alongside the project walk. A worker that lives across edits to `node_modules` declarations or monorepo sibling sources now recompiles instead of replaying the stale project transform. This also hardens the Turbopack loader and Bun runtime paths, which share the process-lifetime cache model.

**Soundness boundary (documented in the metro guide, README, and code):**

- Any change to a fingerprinted input invalidates every transformed file (project-level granularity — forced by Metro's single static key, not chosen).
- Within a running dev-server session, Metro re-transforms only files the watcher reports changed; editing only a type dependency updates the dependent file on the next transform of that file (save it, or restart the dev server — no `--reset-cache` needed). This is Metro's DeltaBundler contract; no transformer-side mechanism can reach it.
- If the snapshot cache directory is unwritable, cross-run caching is disabled (nonce) rather than made unsound.

## Verification

- Two-run reproduction (issue acceptance): new `tests/test-metro/src/features/cache` scenarios simulate Metro's exact key flow — `getCacheKey` once per run, per-file content key — and assert that editing only a type-dependency file changes the key between runs (regenerated output), that an unchanged project keeps the key stable (cache reuse preserved), and both the in-project and out-of-walk edit directions.
- `tests/test-unplugin` gains scenarios pinning the new out-of-walk revalidation (stale before, fresh after; stability twin included).
- Full `tests/test-metro` and `tests/test-unplugin` suites locally; broader suites per the repository's canonical commands. Exact commands and results recorded on this thread as they complete.

## Coordination

- Closes #721. Builds on #718 (`driver.NewTransformGraph` envelope `graph`) and #255 (v1 open question 1). Independent of #720.

https://claude.ai/code/session_01U3kDYEkHutvSMcHPBXhd7P

### DIFF
--- a/packages/metro/README.md
+++ b/packages/metro/README.md
@@ -74,10 +74,19 @@ For each TypeScript file Metro asks to transform:
 
 The plugin contract, `tsconfig` discovery, and per-build cache are identical to every other `ttsc` bundler integration.
 
+## Cache invalidation
+
+Metro keys its transform cache on each file's own content plus one static transformer key, and its babel-transformer contract has no per-file dependency registration. A `ttsc` transform can depend on a _type_ in another file, so `@ttsc/metro` folds a project fingerprint into that static key: every input file under the project root, plus the reference-graph inputs outside it (`node_modules` declarations, monorepo sibling sources, out-of-root tsconfig `extends` ancestry) recorded under `node_modules/.cache/ttsc-metro`. Editing any of them re-keys the next run, so `metro bundle` and dev-server starts pick up cross-file type changes without `--reset-cache`.
+
+The granularity is project-level by necessity: Metro evaluates the transformer key once per run, so any fingerprinted change re-transforms every file on the next run. What remains outside the mechanism's reach:
+
+- **Within a running dev server**, Metro re-transforms only files its watcher reports changed. Editing a type in file B updates a dependent file A on A's next transform — save A, or restart the dev server (no `--reset-cache` needed).
+- **Files a plugin declares `volatile`** depend on non-file inputs that no fingerprint can represent; while a volatile declaration is recorded, cross-run cache reuse is disabled entirely.
+- **If `node_modules/.cache` is unwritable**, the recorded input set is unknown, and cross-run cache reuse is disabled rather than made unsound.
+
 ## Caveats (v1)
 
 - **Cost model.** This release reuses `@ttsc/unplugin`'s transform core, which type-checks the whole `tsconfig` project and caches the result per process. Metro runs transforms in a multi-process worker pool, so the project is compiled once per worker (on that worker's first file). A resident, incremental, per-file compiler shared across workers is the planned optimization, tracked in [samchon/ttsc#255](https://github.com/samchon/ttsc/issues/255).
-- **Cache invalidation.** Metro keys its transform cache on per-file content plus a static transformer key. A `ttsc` transform can depend on a _type_ in another file; editing that type does not change the dependent file's content, so Metro may serve a stale transform. After changing `tsconfig`/plugin configuration or a depended-upon type, restart Metro with `--reset-cache`.
 - **Type errors fail the build.** The `ttsc` pass type-checks; a project type error surfaces as a Metro build error, matching the other `ttsc` bundler integrations.
 
 ## Sponsors
--- a/packages/metro/README.md
+++ b/packages/metro/README.md
@@ -80,7 +80,7 @@ Metro keys its transform cache on each file's own content plus one static transf
 
 The granularity is project-level by necessity: Metro evaluates the transformer key once per run, so any fingerprinted change re-transforms every file on the next run. What remains outside the mechanism's reach:
 
-- **Within a running dev server**, Metro re-transforms only files its watcher reports changed. Editing a type in file B updates a dependent file A on A's next transform — save A, or restart the dev server (no `--reset-cache` needed).
+- **Within a running dev server**, Metro re-transforms only files its watcher reports changed. Editing a type in file B updates a dependent file A on A's next transform: save A, or restart the dev server (no `--reset-cache` needed).
 - **Files a plugin declares `volatile`** depend on non-file inputs that no fingerprint can represent; while a volatile declaration is recorded, cross-run cache reuse is disabled entirely.
 - **If `node_modules/.cache` is unwritable**, the recorded input set is unknown, and cross-run cache reuse is disabled rather than made unsound.
 
--- a/packages/metro/src/core/fingerprint.ts
+++ b/packages/metro/src/core/fingerprint.ts
@@ -1,0 +1,468 @@
+/**
+ * Project fingerprint and reference-graph snapshot for `@ttsc/metro`.
+ *
+ * Metro's transform cache keys each file on its own content plus one static
+ * transformer key computed once per run (`getCacheKey`, called on the main
+ * process at `Transformer` construction). A ttsc transform's output can depend
+ * on inputs Metro never keys: other project sources reached through type-only
+ * edges, `node_modules` declarations, monorepo sibling sources, and the
+ * tsconfig `extends` ancestry. This module folds all of them into the static
+ * key so the cache key incorporates every input that can influence a
+ * transform's output:
+ *
+ * - **Project walk.** Every input file under the fingerprint roots (Metro's
+ *   `projectRoot` plus the resolved tsconfig's directory when it lies outside),
+ *   hashed with the exact walk universe the `@ttsc/unplugin` transform core
+ *   validates its own cache against.
+ * - **Recorded out-of-walk inputs.** The transform core cannot walk files outside
+ *   the roots or under ignored directories, but the host-owned reference graph
+ *   (samchon/ttsc#718) reports them per transform. Workers record them into a
+ *   snapshot under `node_modules/.cache/ttsc-metro`; the next run's
+ *   `getCacheKey` re-hashes the recorded set.
+ *
+ * Snapshot layout: one main file carrying a random epoch id plus per-worker
+ * files with unique names, so concurrent workers never race a shared write.
+ * `withTtsc` (the single config process, before workers exist) compacts worker
+ * files into the main file. Readers take the union of every file, reading the
+ * worker files strictly before the main file: the compactor renames the merged
+ * main into place strictly before deleting a worker file, so a worker file that
+ * disappears mid-read is always already merged into the main the reader loads
+ * afterwards.
+ *
+ * Sound degradations, by design:
+ *
+ * - No readable snapshot (first run, wiped cache dir, unwritable filesystem)
+ *   folds a random nonce: that run shares no cache entries with any other run.
+ * - A recreated snapshot carries a fresh epoch id, so it can never alias a key
+ *   from an older epoch whose recorded set is unknown.
+ * - A plugin-declared volatile output (non-file inputs; unrepresentable in any
+ *   file fingerprint) marks the snapshot volatile, which also folds a nonce
+ *   until a later run records the volatile declaration gone.
+ * - A recorded input that disappears hashes as a stable `missing` marker, so
+ *   deletion and reappearance both move the key.
+ */
+import {
+  collectExternalInputHashes,
+  collectProjectInputHashes,
+  isProjectWalkPath,
+} from "@ttsc/unplugin/api";
+import { createHash, randomBytes } from "node:crypto";
+import fs from "node:fs";
+import path from "node:path";
+
+/** Bumped when the snapshot JSON shape changes; mismatches read as corrupt. */
+const SNAPSHOT_VERSION = 1;
+
+/** Snapshot directory segments under the fingerprint base directory. */
+const SNAPSHOT_DIRECTORY = ["node_modules", ".cache", "ttsc-metro"];
+
+/** Main snapshot file name (epoch id + compacted recorded inputs). */
+const MAIN_SNAPSHOT = "graph-inputs.json";
+
+/** Worker snapshot file prefix; each worker appends a unique suffix. */
+const WORKER_SNAPSHOT_PREFIX = "graph-inputs.worker-";
+
+/** Union of the snapshot state readable on disk. */
+interface SnapshotState {
+  /** Random epoch id minted when the main snapshot was created. */
+  id: string;
+  /** Absolute paths of every recorded out-of-walk input. */
+  files: string[];
+  /** Whether any recorded transform declared volatile output. */
+  volatile: boolean;
+}
+
+/** Serialized shape of the main and worker snapshot files. */
+interface SnapshotDocument {
+  files: string[];
+  id?: string;
+  version: number;
+  volatile: boolean;
+}
+
+/**
+ * Resolve the base directory both fingerprint sides agree on: Metro's
+ * `projectRoot` when known (`withTtsc` reads it from the config, `getCacheKey`
+ * from Metro's cache-key options, the transformer from each file's transform
+ * options — all the same value in a real Metro run), else the working directory
+ * Metro was launched from.
+ */
+export function resolveFingerprintBase(
+  projectRoot: string | undefined,
+): string {
+  return path.resolve(
+    typeof projectRoot === "string" && projectRoot.length !== 0
+      ? projectRoot
+      : process.cwd(),
+  );
+}
+
+/**
+ * The directories whose walk universes the fingerprint hashes: the base
+ * directory, plus the resolved tsconfig's directory when the tsconfig is not
+ * already inside the base walk (an explicit out-of-root `project`, or a
+ * monorepo-root tsconfig discovered above the app). Matching the transform
+ * core's own validation universe keeps the invariant simple: everything the
+ * core treats as an input is fingerprinted, either by a walk here or by the
+ * recorded out-of-walk snapshot.
+ */
+export function fingerprintRoots(
+  base: string,
+  explicitProject: string | undefined,
+): string[] {
+  const tsconfig = resolveProjectTsconfig(base, explicitProject);
+  if (isProjectWalkPath(base, tsconfig)) {
+    return [base];
+  }
+  return [base, path.dirname(tsconfig)];
+}
+
+/**
+ * Locate the tsconfig governing the project, mirroring the transform core's
+ * discovery: an explicit `project` resolves against the working directory;
+ * otherwise ancestor directories starting at `base` are searched for a
+ * `tsconfig.json`, falling back to `<base>/tsconfig.json`.
+ */
+function resolveProjectTsconfig(
+  base: string,
+  explicitProject: string | undefined,
+): string {
+  if (explicitProject !== undefined && explicitProject.length !== 0) {
+    return path.isAbsolute(explicitProject)
+      ? explicitProject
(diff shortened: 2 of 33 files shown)


samchon:
## Implementation pushed — local verification ledger

Two logical commits on top of the claim commit:

1. `feat(unplugin): validate out-of-walk transform inputs in the project cache` — mechanism 2 (in-session), shared-core.
2. `feat(metro): fold a reference-graph project fingerprint into the cache key` — mechanism 1 (cross-run), `@ttsc/metro`.

### Failing-before / passing-after reproduction (mechanism 2)

With only the unplugin core change stashed (tests kept), rebuilt `@ttsc/unplugin` and ran the new scenario — it fails exactly on the stale output the issue describes:

```
test_transformttsc_invalidates_project_cache_when_an_external_input_changes
  actual:   'export const value = "PLUGIN:FIRST";\nconsole.log(value);\n'
  expected: /PLUGIN:SECOND/
```

With the fix restored, the same scenario passes.

### Two-run acceptance (mechanism 1)

`tests/test-metro/src/features/cache/` simulates Metro's exact key flow (one `getCacheKey` per run on the main process, per-file content keys — verified against metro 0.87 source): fresh transformer module per run, on-disk edits between runs, no `--reset-cache` anywhere.

- `test_cache_key_rekeys_a_dependent_transform_when_its_input_file_changes` — edit only the file a transform's output depends on: run 2's key differs and the transform carries `PLUGIN:SECOND`.
- `test_cache_key_changes_when_a_recorded_external_input_changes` — same, for an input outside the project root, recorded by run 1's worker snapshot and compacted by `withTtsc`.
- `test_cache_key_is_stable_across_runs_for_an_unchanged_project` — the negative twin: unchanged project keeps the key, so Metro's cache stays reusable.
- Epoch, nonce, volatile, compaction, and external-only-recording boundaries each pinned in their own case.

### Suites

- `tests/test-metro`: full suite green (49 cases, includes the real native plugin pass; local Go toolchain).
- `tests/test-unplugin`: full suite green (79 cases, includes the #718 webpack filesystem-cache/watch scenarios — no regression from the added external validation).
- Typecheck: `tsc --noEmit` green for `packages/unplugin`, `packages/metro`, and every test suite except the pre-existing `tests/test-lint` failure already documented on #718 (asserts against `@ttsc/lint`'s locally generated `lib/index.d.ts`; no lint source touched here — Linux CI is the arbiter).
- `pnpm` prettier flow run over every changed file before commit.

### Design note for the thread

The volatile channel (#718) is unrepresentable in Metro's contract (no per-file uncacheable control), so a recorded volatile declaration degrades the whole key to a per-run nonce — coarse, but the only sound encoding; documented in the guide. The remaining watch-session boundary (Metro's DeltaBundler re-transforms only watcher-reported files; no cache key can reach it) is documented in the metro guide and README exactly as the issue's acceptance requires.

samchon:
## Self-Review — Round 1 (findings before remediation)

Full fresh pass over the complete base→head diff (unplugin core validation, metro fingerprint/snapshot/recorder, both test suites, docs), plus the consequence surface beyond it: every `transformTtsc` caller (web adapters, Turbopack loader, Bun, Metro), the #718 watch/volatile channels, Windows rename/case semantics, concurrent Metro instances, crash/corruption paths, and the CI lanes that consume the built packages (typia, nestia, experimental, source-map).

**Findings to apply in this round:**

1. **Corrupt worker snapshot files are never healed — permanent nonce.** `readSnapshotState` treats an unparseable worker file as corrupt and degrades to a nonce (correct: its recorded set is unknowable), but `prepareSnapshot` skips unparseable files without deleting them, so the nonce becomes permanent and cross-run caching never recovers after a torn write from a crashed process or a snapshot-version bump. Remediation: compaction deletes unparseable worker files and mints a fresh epoch id, which soundly orphans every key that might have depended on the lost recordings while letting subsequent runs stabilize.
2. **Crash-leftover `.tmp` files accumulate forever.** The atomic writer's temp files are ignored by readers (correct) but nothing ever removes ones left by a crash between write and rename. Remediation: compaction sweeps `*.tmp` in the snapshot directory.
3. **Missing coverage for the in-root ignored-directory external class.** The external-validation tests cover out-of-project-root inputs, but the most common real-world member of the class — a `node_modules/**` declaration inside the project root, invisible to the walk because the segment is ignored — has no direct case. Remediation: add `test_transformttsc_invalidates_project_cache_when_a_node_modules_declaration_changes` (graph edge to an in-root `node_modules` declaration file, edit, same cache, expect a fresh generation).

**Verified as non-issues (no action):**

- **Mid-run edit + exact revert race.** A file edited after the run's key was computed and reverted byte-identically before the next run can alias the earlier key. Metro's own transformer keying (e.g. `babel.config.js` folded once per run) shares exactly this boundary; it is inherent to a once-per-run key and recorded in the design notes, not user docs.
- **Volatile flag stickiness.** A recorded volatile declaration nonces every later run until a run's workers record it gone; a project that removes all ttsc involvement entirely would stay nonced until `node_modules/.cache/ttsc-metro` is cleared. Self-heals on any ttsc transform; acceptable corner.
- **Windows semantics.** `fs.renameSync` over an existing target uses `MoveFileExW(MOVEFILE_REPLACE_EXISTING)`; libuv opens files with full share flags, so reader/writer interleavings degrade to a caught error (nonce) at worst. `path.relative` drive-letter/case behavior matches the shipped `dependencies` channel precedent (#718).
- **Out-of-project transform files** (`isFileOutsideProject` swallow path) record nothing: `notifyWatchInputs` runs only after output selection succeeds, so the recorder never sees them.
- **Synthetic callers without `projectRoot`** fall back to `process.cwd()` on both key and record sides symmetrically; the existing metro transform tests that reach the ttsc pass all pass `projectRoot`, and pass-through files never touch the recorder, so no repository-tree snapshot writes occur under the test runner.
- **Envelope malformed-shape tolerance** is upstream of this diff (#718's parse layer); `selectExternalInputPaths` still guards arrays/strings defensively, and a directory or unreadable member hashes as the stable `missing` marker (state, not error).
- **Web bundler behavior** is unchanged in practice: their caches clear on `buildStart`, and within one build the external re-hash compares equal.

Applying 1–3 now; verification with the affected suites follows in this round's push.

samchon:
## Self-Review — Round 2 (findings before remediation)

Second complete from-scratch pass over the base→head diff after the Round 1 remediation: final `fingerprint.ts` (heal/sweep/flush paths included), the transformer and `withTtsc` wiring, the unplugin core validation, every test file in final form, and both docs pages as rendered prose.

**Finding to apply in this round:**

1. **Voice violation in the metro README's new "Cache invalidation" section** — one added sentence uses an em-dash, which the repository documentation voice forbids ("Editing a type in file B updates a dependent file A on A's next transform — save A, or ..."). The guide page already phrases the same fact without one. Will rewrite the README sentence to match.

**Verified as non-issues in this round:**

- **Snapshot-version bumps self-heal.** A future `SNAPSHOT_VERSION` change makes old main/worker files unparseable: readers nonce (sound), and the next compaction re-mints the epoch and sweeps old worker files, so the key stabilizes again.
- **Worker recorder state is per-base and per-process.** The unique per-process file name cannot collide across runs or projects (directory is keyed by base; suffix is pid+random per module load), and compaction only ever widens the recorded set, so a worker observing fewer files in a later run cannot shrink coverage.
- **Plugin code edits inside `node_modules` are outside the claimed coverage.** The docs claim cross-file type changes, tsconfig edits, and plugin configuration changes; hand-editing an installed plugin's Go source is plugin code, not configuration, is equally uncovered by the web-bundler channel, and arrives in practice through a version bump (package.json/lockfile are in the walk). Wording stands.
- **`withTtsc` base vs. key base mismatch** (config without `projectRoot` where Metro later resolves a different default): the key side finds no snapshot at its base and degrades to a nonce — cache-less but sound; the documented Expo/RN default-config path always carries `projectRoot`.
- **Doc anchors** (`driver-api#reference-graph`, `protocol#transform`) resolve to existing headings already linked by the unplugin pages.
- Early-return paths (`shouldTransform` false, outside-project swallow) never touch the recorder; `getCacheKey`'s multiple-call behavior only diverges in the nonce state, which Metro's once-per-run contract never observes.

samchon:
## Self-Review — Round 3 (findings before remediation)

Third complete pass, focused on packaging, public API compatibility, platform behavior, and side effects observable outside the repository — including re-running the suite's filesystem footprint.

**Findings to apply in this round:**

1. **`prepareSnapshot` creates directory trees at arbitrary nonexistent roots.** `withTtsc({ projectRoot: "/workspace/app" })` (the existing config passthrough test) materialized `D:\workspace\app\node_modules\.cache\ttsc-metro` on the test machine: `mkdirSync(..., { recursive: true })` happily builds the whole chain under a root that does not exist. Metro itself requires the project root to exist, so a nonexistent base can never be a working setup — the snapshot cannot help it, and creating paths outside any real project is a side effect no config wrapper should have. Remediation: `prepareSnapshot` returns without touching the filesystem when the base directory does not exist; a new case pins that a nonexistent root creates nothing.
2. **Config passthrough tests are no longer hermetic.** The `withTtsc` tests that pass no `projectRoot` now write the snapshot under the suite's working directory (`tests/test-metro/node_modules/.cache/ttsc-metro`). Remediation: those tests get real temp-dir `projectRoot`s (their assertions are about config field passthrough and are unaffected), and the polluted directories are removed from the machine.

**Verified as non-issues in this round:**

- **No manifest or lockfile changes anywhere in the diff**; no `version` fields touched; `pnpm-lock.yaml` untouched (no dependency changes).
- **Public API compatibility:** `@ttsc/unplugin/api` gains three additive exports; `TtscCachedProjectTransform` gains one optional field; no existing signature changes. `@ttsc/metro`'s new `core/fingerprint` module is internal (package exports map unchanged; tests load it by file URL).
- **Built surface sanity-checked** against `lib/api.js`: the walk predicate classifies in-root `node_modules` as external, in-root sources as walk-visible, out-of-root as external.
- **POSIX/Windows:** all path handling goes through `path.*`; the snapshot writer's rename-over-existing works on both platforms; no shell or case assumptions beyond the shipped `dependencies`-channel precedent.
- **CI lanes:** metro/unplugin run commands unchanged; no new devDependencies, so installs and the packaged-manifest/externals tests are unaffected (both suites already green locally on the final code).

samchon:
## Self-Review — Round 4 (clean round)

Fourth complete pass over the final base→head diff (5 commits, 31 files), including every file touched by the Round 1–3 remediations in its final formatted form, a repository-wide sweep for stale `--reset-cache` claims (only the intentionally documented dev-server/no-reset-cache statements remain), packaging and public-API compatibility (no manifest, lockfile, or `version` changes; three additive `@ttsc/unplugin/api` exports; one optional cached-transform field; `@ttsc/metro`'s fingerprint module stays internal), the experimental perf harness (`pluginRuns === 1` is unaffected: the external re-hash is vacuous or stable within a build), and filesystem hermeticity re-verified after the suite runs (no suite-dir snapshot, no fake roots).

**Verification ledger (final):**

- `tests/test-metro`: full suite green (57 cases).
- `tests/test-unplugin`: full suite green (exit 0, includes the five new external-validation cases and the #718 webpack filesystem-cache/watch scenarios).
- Typecheck green for both packages and every suite except the pre-existing `tests/test-lint` local-generation failure documented on #718; the Linux `test` lane covering it is green on this head.
- Prettier clean across the entire diff.
- CI on head `97d62e1b4`: all 50 checks pass, including the metro and unplugin lanes, shim-audit, the native corpora, typia, and nestia.

No sound improvement or issue candidate survived this pass — the review terminates per the stop rule. Marking ready for review and merging under the standing autonomous mandate.

PR #724, opened 2026-07-16
feat(ttsc): dependency-completeness contract for narrowing invalidation

### LINKED ISSUE #720: plugin protocol: specify the completeness contract for dependency narrowing

samchon:
## Intent

Implement samchon/ttsc#720: give the plugin protocol an explicit contract by which a transform declares its reported per-file dependency list **complete**, so a consumer can invalidate a file *below* the host-owned language-semantic bound that #716 established.

Union semantics (#716, #722) can only ever widen: a precise producer such as typia (samchon/typia#2100) already knows exactly which declarations it consulted, yet a change to any file reachable through a module's import closure still re-runs its loader. This is the Bazel declared-inputs model. Under-declaration becomes a plugin bug by contract instead of a silent platform failure, and no producer needs the field for correctness.

Closes #720.

## Design decisions (the issue owns these; here is what was chosen and why)

- **Marker shape:** `dependenciesComplete`, a list of transformed-file keys (same convention as `typescript` and `volatile`), not a whole-envelope boolean. One `transform` invocation writes one envelope even when several plugin entries share the host, so a producer that is precise only for the files it generates into must still be able to leave every other file on the sound baseline. A file list makes a mixed envelope compose per file and reuses the existing `parseFileList` tolerance.
- **What completeness replaces:** for a listed file `F` the derivation is `dependencies[F] ∪ graph.configs` instead of `dependencies[F] ∪ reach(graph.edges, F) ∪ graph.globals ∪ graph.configs`.
  - **Configs stay universal.** Compiler options reach generated code through the host, not through any file a plugin can consult, so no consulted-declaration list can represent them.
  - **Globals do not stay universal.** Global-scope files are ordinary declaration files that a Checker-driven producer lists like any other, so a complete list must name every ambient file it consulted. Keeping them universal would leave `@types/*` churn re-running every listed file, which is most of what the narrowing is for.
- **Multiple plugins:** completeness is per (plugin, file), but the envelope is per host process, so the declaration is the envelope author's aggregate claim. The rule is stated in the protocol page: when several entries contribute to a file, list it only if every contributing entry declared its own list complete for it. The consumer cannot attribute `dependencies[F]` entries back to an entry and honors the claim as written.
- **Volatile wins.** A file declared both complete and volatile makes contradictory claims; the consumer keeps the baseline union for it and still bypasses caching.
- **Failure story:** the responsibility transfer is documented explicitly. There is deliberately no host-side audit of a complete list, because verifying it would require computing exactly the reachability the declaration exists to avoid. The diagnostic is removal: drop the field and the sound baseline returns. Malformed members drop to the baseline, which is the only safe drop direction.

## Scope

1. **Protocol documentation first** (`concepts/protocol.mdx`): new `#### Dependency completeness` section stating the responsibility transfer, the universal/non-universal split, mixed and multi-plugin semantics, the volatile interaction, and the diagnostic strategy. Consumer-facing notes added to `setup/unplugin.mdx`, `setup/metro.mdx`, and the `@ttsc/unplugin` README.
2. **Envelope types** (`ITtscCompilerTransformation`): `dependenciesComplete?: string[]` on `ISuccess` and `IFailure`.
3. **Host parsing** (`transformProjectInMemory`, `TtscCompiler`): parse and forward with the same drop-malformed-keep-valid tolerance as `dependencies` / `graph` / `volatile`.
4. **`@ttsc/unplugin` derivation** (`selectWatchInputs`): narrow per file. Every adapter inherits it through the existing `addWatchFile` plumbing, and `@ttsc/metro` inherits it through its snapshot recorder, which consumes the same hook. No `@ttsc/metro` code change was needed; its recorded out-of-walk set narrows automatically.

### Deliberately out of scope

- **The project transform cache's out-of-walk validation (`selectExternalInputPaths`) stays un-narrowed.** It replays one whole envelope, so its validity condition is the union over every file the envelope carries rather than one file's inputs; a miss there costs a re-transform, never a stale output; and it is the layer that re-runs the plugin's analysis, which is the only way a widened declaration is ever learned. Narrowing it too would make a stale declaration self-perpetuating in a long-lived host. Pinned by a test rather than left implicit.
- **Go driver SDK:** no change. The built-in and linked-plugin hosts never report `dependencies`, so they have nothing to declare complete. Adoption is an executable sidecar's one added envelope field.
- **No producer adoption.** typia (samchon/typia#2100) is the first candidate and is not part of this change.

## Tests

- `tests/test-unplugin/src/features/transform/`: complete narrows to declared + universal inputs; the empty-declaration boundary; a mixed envelope composing per file (the negative twin: the unmarked file keeps the baseline union out of the same envelope); volatile overriding completeness; completeness not narrowing external cache validation.
- `tests/test-ttsc/src/features/api/`: the field surfacing end to end through the programmatic API, and malformed members dropping while valid ones survive.
- The existing union-semantics tests are untouched and must keep passing for producers that never declare completeness.

## Verification

Run locally:

- `npx tsc --noEmit` on `packages/ttsc` and `packages/unplugin`: clean.
- Both Go fixture plugins (`TestUnpluginProject.writeGoPlugin`, `test-ttsc`'s `writeCompilerPluginBackend`) extracted and `go build`-ed: clean. The malformed-advisory envelope was JSON-parsed to confirm it carries the intended mixed members.
- `pnpm format`: run, only this change's files staged.

**Not run locally, stated honestly:**

- `tests/test-*` e2e suites: they cannot run on this Windows checkout (`file://` URL issue). CI is the gate for every scenario above.
- `pnpm build`: the local Node is 22 and the rollup build needs Node 24 (`RegExp.escape`). CI is the gate.
- The test-package typechecks resolve `ttsc` from its built `lib/`, which is stale in this checkout (it already fails on `graph`/`volatile` from #718), so they prove nothing here either way.

https://claude.ai/code/session_015XsPmVseWirNXRSm6HgGAj


samchon:
## Self-review rounds (solo)

Applied, each after a complete round over the base-to-head diff and its consequence surface:

1. **`dependenciesComplete` without `graph` was an undocumented trap.** The rule said the config chain stays universal without saying the chain comes from the same envelope's `graph` section. A sidecar that declared completeness but never adopted `graph` would keep *no* universal input, so a tsconfig edit would stop invalidating that file. Stated in the protocol page and the envelope type.
2. **Reused the sibling scenario module's `cacheEntry`** instead of duplicating it in `transform-complete.ts`.
3. **Restated the default derivation** inside the completeness section: it sits at the end of `transform`, so "the union above" pointed several paragraphs back.
4. **Added the under-declared defect-surface test** (`test_webpack_filesystem_cache_serves_stale_for_an_under_declared_complete_file`). The issue names an incomplete-marked envelope as its own test category, and the derivation tests only exercised the correct-producer side. It mirrors the existing no-graph control on the same type-only-edge project: the producer emits the graph edge but reports no dependencies while declaring the file complete, so webpack's kept filesystem cache serves stale output. It makes the responsibility transfer observable and locks in that the host does not audit the declaration (auditing would cost the exact reachability walk the declaration avoids).

Rounds since the last change produced no further findings. Consequence surface confirmed along the way: `packages/unplugin/src/core/transform.ts` is the single derivation site, so `core/index.ts`, `bun.ts`, `turbopack.ts`, and `@ttsc/metro`'s snapshot recorder all inherit the narrowing through the existing `addWatchFile` hook with no code change; `packages/wasm` carries no advisory envelope fields (consistent with #716); and there is no TS API baseline or programmatic-api page enumerating these fields.

samchon:
## CI

All 56 checks pass on `f4dcc8f14`.

`nestia` failed on the first run and was re-run. It was not an exemption case and not a code fault: the runner ran out of disk while `go build`-ing typia's native plugin.

```
compile: writing output: write $WORK/b221/_pkg_.a: no space left on device
github.com/samchon/typia/.../transform/features/http: mkdir /tmp/go-build.../b227/: no space left on device
```

The `src/decorators/TypedException.ts(84,32): error TS6196` line in that log is a red herring worth noting for the next reader: once the plugin build fails, `ttsc` runs its independent `tsgo --noEmit` fallback pass and appends the diagnostics the plugin never reported (documented under [Exit and Output](https://ttsc.dev/docs/development/concepts/protocol#exit-and-output)). The TS error is that pass reporting nestia's own source, not the cause. The re-run is green.

PR #728, opened 2026-07-17
feat(lint): let project rules opt out of the type checker

### LINKED ISSUE #727: feat(lint): let a project rule decline the type checker instead of serializing every rule in the run
## Problem

A project rule has no way to decline the type checker, and declaring one silently serializes the whole run.

`Engine.needsTypeChecker` is a single per-engine field. `runsSerial()` returns true when it is set, and `Engine.NeedsTypeChecker()` gates creation of the standalone checker for `compile`, `fix`, `format`, and every LSP verb. When the engine registers project rules, it sets that field for **any** project rule that is declared and not `off`, without consulting the rule at all.

File rules can opt out: a contributor implements `TypeAwareRule` with `NeedsTypeChecker() bool { return false }` and the adapter records it. Project rules cannot — `ProjectRule` is `Name()` plus `Check(ctx *ProjectContext)`, and nothing in the registration path asks a project rule anything about the checker.

**Expected:** a project rule that never reads `ctx.Checker` should not create the standalone checker, and should not force the serial walk.

**Impact.** The cost is not confined to the project rule that caused it. Because the flag is engine-wide, one declared project rule puts **every file rule in the run** on the serial walk — including rules that carefully declared `NeedsTypeChecker() bool { return false }` to stay parallel. Their opt-out silently buys nothing while any project contributor is installed, and nothing reports this.

**Affected users.** Every contributor shipping a project rule that inspects only `ctx.Sources` or the filesystem. Two exist today that I can point at:

| Package | Project rules | Reads `ctx.Checker`? |
| --- | --- | --- |
| `@autobe/lint` (`wrtnlabs/autobe-mcp`) | `project-boundary` | No |
| `@samchon/evidence` | `evidence/index`, `evidence/coverage` | No |

`@samchon/evidence` is the clearest case: both its project rules read markdown from disk and walk `ctx.Sources` syntactically, and its file rules already declare `NeedsTypeChecker() bool { return false }`. That declaration is dead the moment `evidence/index` is enabled, which is always — every other rule in the plugin gates on the index it publishes.

## Evidence

**Root cause, verified.** In `NewEngineWithResolver`, the loop over `allProjectRuleNames()` sets `eng.needsTypeChecker = true` whenever `setting.Declared && setting.Severity != SeverityOff`. There is no consultation of the adapter, and `projectRuleAdapter` carries no field that could answer. Compare the file-rule path in the same constructor, which routes through `ruleNeedsTypeChecker(rule)` and therefore honors the marker.

**`ProjectContext.Checker` is populated unconditionally** in the project engine — the checker is passed straight into `NewProjectContext` with no marker consultation, so even a rule that wanted to decline would still be handed one.

**Reproduction.** Register any project rule whose `Check` ignores `ctx.Checker`, configure it at `error`, and construct an engine:

```go
engine := NewEngineWithResolver(InlineRuleResolver{
  Rules: RuleConfig{"project-test/ast-only": SeverityError},
})
engine.NeedsTypeChecker() // true — no way to make this false
```

**Why the marker cannot simply be reused as-is.** `TypeAwareRule` embeds `Rule`, and `Rule` requires `Check(ctx *Context, node *shimast.Node)` while `ProjectRule` requires `Check(ctx *ProjectContext)`. Go permits one `Check` method per type, so **no `ProjectRule` can ever satisfy `TypeAwareRule` in its current shape**. This is not an oversight to work around in the host; the interface itself blocks it.

**Provenance.** Found while building `@samchon/evidence`, an external `@ttsc/lint` contributor, against 0.19.1.

## The default is correct and must not change

Worth stating plainly, because the obvious framing is wrong. `ProjectContext` carries a `Checker`, and the host cannot infer a third-party rule's shape, so treating an unmarked project rule as type-aware is the right conservative default — the same policy the contributor file-rule adapter applies. **This is not a request to flip a default.** The gap is narrower: project rules have no override at all.

## Consequence surface

| Case | Today | After |
| --- | --- | --- |
| Project rule, no marker | checker + serial | unchanged |
| Project rule, marker `true` | checker + serial | unchanged |
| Project rule, marker `false` | checker + serial | no checker, parallel walk preserved |
| No project rule declared | file rules decide | unchanged |
| Project rule declared but `off` | no effect | unchanged |
| File rule marker `false`, project rule declared | serial anyway | parallel, if the project rule also declines |

Backward compatibility is total: any project rule that does not implement the marker behaves byte-identically. `@autobe/lint` is affected today (it forces serial) and stays unaffected until it opts in.

Reaches `compile`, `fix`, `format`, and the LSP verbs, since all of them gate checker creation on the same `Engine.NeedsTypeChecker()`.

## Approach

The invariant: **the engine's checker decision must be the union of what every declared rule asked for, and a project rule must be able to answer.** The architectural owner is `projectRuleAdapter` and the project-rule registration path, mirroring how `contributorMetadata` already caches `needsTypeChecker` for file rules.

Two constraints worth surfacing before implementation, both verified:

1. **Sharing one marker requires dropping the `Rule` embed from `TypeAwareRule`.** `OptionsRule` already sets this precedent for exactly this reason — it deliberately does not embed `Rule`, relies on a domain-specific method name to prevent accidental opt-in, and its doc comment already says *"ProjectRule implementations may use the same marker."* Dropping the embed is backward compatible: every existing implementer still satisfies the interface, and the only consuming assertion starts from a value that is already a `Rule`. The alternative is a second marker interface for one method.

2. **The adapter field's zero value must be the safe direction.** Several test helpers construct `projectRuleAdapter` literals directly rather than going through `inspectProjectContributor`. A positive `needsTypeChecker bool` therefore makes those zero values deny a checker to rules that never declined one — `TestProjectRuleRunsForEmptySourceSet` catches this immediately. Spelling the field as an explicit opt-out (`declinesTypeChecker`) makes the zero value conservative and seals every construction site, present and future, rather than repairing the one that happens to be covered.

A rule that declines should also receive a nil `ctx.Checker`, which is what `TypeAwareRule`'s documentation already promises ("the host is free to leave it nil"). Handing it one anyway lets a declining rule read the checker and keep working while its own declaration says otherwise — until the engine acts on that declaration and the rule faults far from the marker that caused it.

| Outcome | Assertion |
| --- | --- |
| Positive | A project contributor whose marker returns `false`, declared at `error`, leaves `engine.NeedsTypeChecker()` false |
| Negative twin | A project contributor with **no** marker, declared at `error`, keeps `engine.NeedsTypeChecker()` true |
| Boundary | A declining project rule receives a nil `ctx.Checker` |
| Regression | Every existing test in `packages/lint/test/` passes **unmodified** — especially `TestProjectRuleRunsForEmptySourceSet`, which asserts the unmarked default and is the tripwire for the zero-value trap above |

Two new cases belong in `packages/lint/test/plugin/`, one assertion each per the development skill, mirroring the registration/teardown shape of the existing `project_contributor_*` tests.

Narrow proving command:

```bash
node scripts/test-go-lint.cjs
```

Broader, since this touches the shared walk decision:

```bash
pnpm test
```

**Change Integrity:** this alters a core algorithm — the serial/parallel walk decision — and that must be called out in the final report. No existing test should need to change; if one does, that is a signal the zero-value direction is wrong rather than a reason to edit the test.

**Docs:** the walkthrough at `website/src/content/docs/development/walkthroughs/lint.mdx` documents the options-compatibility boundary for project contributors under *Project-scoped contributor rules*. The checker boundary belongs beside it, in the same change.

## Coordination

No dependencies. Nothing else touches `needsTypeChecker` for project rules.

I have a working implementation of the above on a local branch, verified green with no existing test modified, including the website update. Happy to open it as a PR against this issue, or to leave it to whoever picks this up if you would rather the approach be settled here first.

**Note for whoever implements:** this repository formats Go with `.vscode/gofmt-2spaces.sh`, not standard `gofmt`. Running `gofmt -w` reformats the entire tree and buries the change in an unreviewable diff. Use `pnpm format`, or the script directly on the touched files.

samchon:
Closes #727.

## What changed

A project rule can now decline the type checker with the same `NeedsTypeChecker() bool { return false }` marker file rules already use, and the engine stops forcing the standalone checker and the serial walk on its behalf.

Nothing changes for a project rule that says nothing. The default was never the defect — `ProjectContext` carries a `Checker` and the host cannot infer a third-party rule's shape, so treating an unmarked project rule as type-aware stays correct. The gap was that there was no way to say otherwise.

## Why it is shaped this way

**`TypeAwareRule` drops its `Rule` embed.** This is not a convenience. `Rule` requires `Check(ctx *Context, node *shimast.Node)` while `ProjectRule` requires `Check(ctx *ProjectContext)`, and Go permits one `Check` per type — so no `ProjectRule` could ever satisfy the interface in its old shape. `OptionsRule` already set this precedent for exactly this reason: it deliberately does not embed `Rule`, leans on a domain-specific method name to prevent accidental opt-in, and its doc comment already promises that "ProjectRule implementations may use the same marker." Every existing implementer still satisfies the interface, and the one consuming assertion starts from a value that is already a `Rule`.

**The adapter field is `declinesTypeChecker`, not `needsTypeChecker`.** The negative spelling is load-bearing rather than stylistic. Several test helpers build `projectRuleAdapter` literals directly instead of going through `inspectProjectContributor`, so a positive field makes those zero values deny a checker to rules that never declined one — the unsafe direction, reached by forgetting rather than by deciding. `TestProjectRuleRunsForEmptySourceSet` caught this on the first attempt. Inverting the field seals every construction site, present and future, instead of repairing the one that happened to be covered.

**A declining rule receives a nil `ctx.Checker`.** `TypeAwareRule`'s own documentation already promised this ("the host is free to leave it nil"); the project path now honors it. Passing a checker anyway would let such a rule read one and keep working while its declaration says it does not, until the day the engine acts on that declaration and the rule faults far from the marker that caused it.

## Verification

`node scripts/test-go-lint.cjs` green.

**No existing test changed.** That is the load-bearing claim here: the one existing test that failed during development was pinning the correct behavior, and it passes untouched once the zero value is safe.

Two new cases in `packages/lint/test/plugin/`, one assertion each:

| Case | Asserts |
| --- | --- |
| `project_contributor_ast_only_marker_keeps_parallel_walk_test.go` | marker `false` ⇒ `NeedsTypeChecker()` false |
| `project_contributor_without_marker_keeps_type_checker_test.go` | no marker ⇒ `NeedsTypeChecker()` true |

The second is the negative twin: it fails if the default ever drifts, which is the change most likely to be made by accident here.

## Change Integrity

**This alters a core algorithm — the serial/parallel walk decision.** Calling it out per the development skill. Behavior for every rule that does not implement the marker is byte-identical, and the walk decision remains the union of what every declared rule asked for; project rules simply gain the ability to answer.

The website walkthrough is updated in the same change, under *Project-scoped contributor rules*, beside the existing options-compatibility paragraph it mirrors.

## Provenance

Found while building [`@samchon/evidence`](https://github.com/samchon/evidence), an external contributor whose two project rules read markdown from disk and never touch `ctx.Checker`, and whose file rules already declare `NeedsTypeChecker() bool { return false }` — a declaration that is dead today, because the plugin's index rule is always enabled and every other rule gates on it.


samchon:
## CI: the one red check is pre-existing on `master`, not from this PR

`ttsc fast` fails on `test_release_workflow_runs_preflight_before_publication`:

```
AssertionError: preflight (index 1906) must run before build (index 186)
```

Two things establish it is not mine:

- This branch touches neither `release.yml` nor `tests/test-ttsc` — the diff is `packages/lint` Go plus the website walkthrough. `git diff master..HEAD --stat` shows no overlap with either.
- Checking out clean `master` and running that single case reproduces the identical assertion.

So it is a step-ordering assertion against the release workflow that `master` currently violates, surfaced here only because the PR runs the suite. Flagging rather than touching it: the workflow is Change Integrity territory, and fixing an unrelated red in a PR about the walk decision would bury this change's diff in someone else's.

Every other check is green (25 pass at time of writing, remainder still running).

samchon:
Part of #733 — the umbrella surveying the contributor-facing editor-assistance gap. This PR is its first child and has no dependency on the rest.

PR #735, opened 2026-07-17
fix(graph): resolve details.literals from the checker's type

### LINKED ISSUE #732: `@ttsc/graph`: `details.literals` under-reports union/enum members (scraped from a truncated signature, not the checker type)
Hi — first, thank you for `@ttsc/graph`. The one-tool-with-a-typed-CoT design and the checker-resolved graph are excellent, and the "answer from names/edges/spans, never bodies" contract holds up in practice.

One small thing I ran into: For a `details` request on a string-literal **union** (or enum) whose members are written one per line, the returned `literals` array and `signature` are **truncated**, so the member list under-reports — and it does so silently. The completeness turns out to depend on source *formatting* rather than on the type.

```ts
// many-line union (one member per line)
export type Kind =
  | 'a'  // ...
  | 'b'
  | 'c'
  | 'd'
  | 'e';

// single-line union
export type Flat = 'a' | 'b' | 'c' | 'd' | 'e';
```

`details` with `handles: ["Kind"]` returns `literals: ["a","b","c"]` (first ~3), and a `signature` cut to the first few lines. `details` on `Flat` returns all five. So the same five-member type yields a complete list or a partial one depending only on line-wrapping.

(In a real codebase this bit harder: a 25-member union written one-per-line came back as 3 of 25.)

In `packages/graph/src/server/runDetails.ts`, `literals` is derived from the *signature snippet*, not from the type:

- `signatureOf(...)` reads the declaration's first `MAX_SIGNATURE_LINES = 4` **lines of source text**
  (it stops at a `{`, a trailing `;`, or the line cap).
- `literals` is then `literalSummaries(signature).slice(0, 6)`.

So `literals` reflects whatever string literals happen to fall in the first ~4 source lines, capped at 6 — it isn't read from the checker's resolved union/enum members.

The result `audit` tells the caller these facts are compiler-resolved. `literals` is the one field that's an exception: a source-text scrape.

Possible approaches:

1. Enumerate union/enum members from the resolved **type** (the checker has them) rather than from the signature snippet — ideally the highest-fidelity fix.
2. Or, if the snippet approach stays, mark truncation on this field (a `truncated`-style flag, as elsewhere) and/or document that `literals`/`signature` reflect only the captured signature prefix.

Thanks again for the tool.

**Environment:** `@ttsc/graph@0.19.1`, `ttsc@0.18.4`, `typescript@7.0.2`, linux-x64.

samchon:
Fixes #732.

## Intent

`details.literals` claimed to be a compiler-resolved fact and was a regex over source text. `signatureOf()` reads at most four lines from the declaration start (stopping at `{` or a trailing `;`), and `literals` was `literalSummaries(signature).slice(0, 6)` over that snippet, so what a union reported depended on how it was wrapped rather than on the type. All of it silently, under a `RESULT_AUDIT` telling the caller that facts here are resolved and need no re-verification.

The reporter's witness was one of four faces of that root cause. Verified against the shipped code before changing anything:

| case | before | now |
|---|---|---|
| 5-member union, one per line | `["a","b","c"]` | all 5 |
| 10-member union, single line | 6 of 10 (the cap bites whatever the layout) | all 10 |
| multi-line `enum` | **no `literals` at all**, and `members` empty too | all values |
| `type Indirect = Kind \| 'f'` | `["f"]` — a wrong answer of the right shape | all 6 |

The enum case was the worst and was not merely truncation: `signatureOf` stops at the line holding `{`, and enum members are not graph nodes (the builder records member nodes for classes and interfaces only), so a multi-line enum's `details` carried no member information whatsoever.

## Scope

Issue option 1, the checker route. The Go builder resolves each type-alias/enum symbol's declared type (`getDeclaredTypeOfSymbol`) and enumerates the union's literal constituents. That is complete by construction, indifferent to line wrapping, and follows indirection; the checker has also already flattened, deduped, and subtype-reduced, so the constituents are the members exactly once each. Values ride the dump uncapped — the dump is the whole graph by contract — and the MCP layer applies the response cap.

- `Node.Literals` + a `collectLiterals` checker pass (`packages/ttsc/internal/graph/literals.go`), rendered by the checker's own `ValueToString`, newly re-exported through `shim/checker` (numeric formatting and string escaping are not things a consumer should re-derive).
- `literals` on `ITtscGraphNode`; the scrape helpers are deleted rather than left reachable.
- `literalsTruncated` on `ITtscGraphDetails.INode`, which `details` had no field to express before.

Two decisions worth flagging, both a change of specification:

1. **All-or-nothing.** `type T = Kind | string` admits values no list can name, so it reports none rather than the four that can — a partial list is indistinguishable from a complete one, and the audit's promise is that a returned fact cannot be wrong. `signature` still carries the shape.
2. **Source form.** Values are rendered as written (`"a"`, `1`, `true`, `null`), not as bare contents (`a`). The old field was string-only, so `type N = 1 | 2` reported nothing; source form covers every literal kind and is the only rendering that can tell `1` from `"1"`. The field's sole consumer is the model reading the payload.

`ttsc` becomes a peer dependency of `@ttsc/graph` pinned to the same version (the `@ttsc/unplugin` pattern; `workspace:*` publishes as an exact pin — verified against the published `@ttsc/unplugin@0.19.2`, which carries `ttsc: '0.19.2'`). This fix is resolved in the Go builder and read in TypeScript, so a newer `@ttsc/graph` against an older `ttscgraph` would find `literals` absent and answer from a graph missing facts it believes are there. The reporter is running graph 0.19.1 against ttsc 0.18.4, which is exactly that skew.

## Deferred

`details` still caps `calls`/`types`/`implementedBy` at 2 by default, `members` at 6, and `signature` at 4 lines, none of them marked — while `RESULT_AUDIT` and the `IOutput.audit` doc both say a `details` result is "bounded only where `truncated` says". `literalsTruncated` closes that gap for the field this issue is about; the rest is the same class of silent bound and wants its own issue, not a drive-by here.

Enum members as graph nodes is likewise left alone. `literals` now carries their values, which is what #732 asked for, but their names still reach no consumer.

## Verification

Local, on this branch:

- `go test ./internal/graph/` — full package green, including four new cases. Each was confirmed to **fail** with the `collectLiterals` call commented out, so they pin the bug rather than the code.
  - `literals_do_not_depend_on_declaration_line_wrapping_test.go` — eight members wrapped vs flat, past the old six-cap so a reintroduced cap fails on the flat twin.
  - `literals_resolve_through_alias_indirection_test.go` — one hop, two hops, and a member named twice reported once.
  - `literals_cover_enum_members_the_graph_has_no_nodes_for_test.go` — multi-line, single-line, and implicitly numbered members (the values exist only in the checker), plus an assertion that no enum member node exists, so `literals` is genuinely the only carrier.
  - `literals_are_absent_when_the_type_admits_more_than_it_can_name_test.go` — the negative twins (`| string`, type parameter, primitive alias, computed enum member) against a four-member control.
- `pnpm --filter ttsc shim:audit` — OK, closed after the `ValueToString` re-export.
- `npx tsc --noEmit -p packages/graph/tsconfig.json` — clean.
- Real `ttscgraph dump` over a fixture, then the real `runDetails` driven over that dump: `Kind` 3→5, `Colors` (multi-line enum) none→3, `Indirect` `["f"]`→6, `Widened` absent, escaping intact (`'he said "hi"'`). An 80-member template-literal union arrives whole in the dump and comes back as 60 + `literalsTruncated`, with the marker staying off for every uncapped type.

Not run locally, left to CI:

- `tests/test-graph` e2e (`test_ttscgraph_details_literals_enumerate_the_resolved_type.ts`), which drives the real binary over MCP. The suite cannot start on this Windows checkout (`ERR_UNSUPPORTED_ESM_URL_SCHEME`), and `pnpm --filter @ttsc/graph build` could not complete here either — the D: volume is at 100%, so the typia plugin's Go build fails on disk, unrelated to this change.

### DIFF
--- a/packages/graph/package.json
+++ b/packages/graph/package.json
@@ -22,6 +22,9 @@
     "@typia/mcp": "catalog:samchon",
     "typia": "catalog:samchon"
   },
+  "peerDependencies": {
+    "ttsc": "workspace:*"
+  },
   "keywords": [
     "ttsc",
     "mcp",
--- a/packages/graph/src/model/TtscGraphSession.ts
+++ b/packages/graph/src/model/TtscGraphSession.ts
@@ -5,6 +5,7 @@ import typia from "typia";
 import { ensureExecutable } from "../nativeExecutable";
 import { resolveGraphBinary } from "../resolveGraphBinary";
 import { ITtscGraphSnapshot } from "../structures/ITtscGraphSnapshot";
+import { DUMP_SCHEMA_VERSION } from "./loadGraph";
 import { TtscGraphMemory } from "./TtscGraphMemory";
 
 /**
@@ -226,6 +227,32 @@ export class TtscGraphSession {
       );
       return;
     }
+    // The envelope's version is not the body's, and only the envelope has been
+    // held to one so far. A producer can speak this protocol and still carry a
+    // dump from another schema — the two move apart the moment a node field is
+    // added without the frame around it changing — and then the facts that field
+    // holds are silently absent rather than refused. `literals` is exactly that
+    // shape: an older producer resolves no value set, so a union comes back
+    // looking like a type with no members. Hold the body to its own number too,
+    // once the frame is understood.
+    if (
+      response.dump !== undefined &&
+      response.dump.provenance.schemaVersion !== DUMP_SCHEMA_VERSION
+    ) {
+      // Session-wide, for the same reason the protocol mismatch above is: it is
+      // the wrong binary, not one bad frame.
+      this.failPending(
+        new Error(
+          `@ttsc/graph: ttscgraph sends dump schema v${String(
+            response.dump.provenance.schemaVersion,
+          )}, this client reads v${String(DUMP_SCHEMA_VERSION)}. ` +
+            "Install a matching `ttsc` (the binary resolves from the target " +
+            "project, or from TTSC_GRAPH_BINARY).",
+        ),
+      );
+      return;
+    }
+
     const pending = this.pending.get(response.id);
     if (pending === undefined) return;
     this.pending.delete(response.id);
--- a/packages/graph/src/model/loadGraph.ts
+++ b/packages/graph/src/model/loadGraph.ts
@@ -18,8 +18,13 @@ const MAX_DUMP_BYTES = 1024 * 1024 * 1024;
  * `packages/ttsc/internal/graph/provenance.go`. The two are hand-synchronized,
  * and `dump_schema_version_matches_the_typescript_client_test.go` reads this
  * constant out of this file and fails if the pair drifts.
+ *
+ * Exported because the resident session reads the same dump body over the serve
+ * protocol and has to hold it to the same number: the envelope's version and the
+ * body's are independent, so a producer can speak this protocol and still send a
+ * body from another schema.
  */
-const DUMP_SCHEMA_VERSION = 1;
+export const DUMP_SCHEMA_VERSION = 2;
 
 /**
  * Build the resident {@link TtscGraphMemory} for a project by running `ttscgraph
--- a/packages/graph/src/resolveGraphBinary.ts
+++ b/packages/graph/src/resolveGraphBinary.ts
@@ -5,9 +5,15 @@ import path from "node:path";
  * Resolve the per-platform `ttscgraph` binary, or `null` when it cannot be
  * located.
  *
- * `ttsc` is a peer the user installs alongside `@ttsc/graph` (not a dependency
- * of this launcher), so resolution starts from the user's project, not from
- * this package's own tree.
+ * `ttsc` is a peer dependency pinned to this package's own version, installed
+ * alongside `@ttsc/graph` in the user's project rather than nested under this
+ * launcher, so resolution starts from the user's project, not from this
+ * package's own tree.
+ *
+ * The pin is what keeps the two halves of the graph in step. `literals` and the
+ * rest of the node facts are resolved by the Go builder and ride the dump, so a
+ * newer `@ttsc/graph` reading an older `ttscgraph` would find those fields
+ * simply absent and answer from a graph missing facts it believes are there.
  *
  * Resolution order:
  *
--- a/packages/graph/src/server/runDetails.ts
+++ b/packages/graph/src/server/runDetails.ts
@@ -13,6 +13,11 @@ import { IRunnerOutput, resultNext } from "./resultNext";
 
 // A signature is the declaration head up to the body brace: a handful of lines.
 const MAX_SIGNATURE_LINES = 4;
+// A value set is the answer to "what may this be", so it is listed rather than
+// sampled: the cap is high enough that a real union or enum arrives whole, and
+// exists only so a type expanded from a template literal cannot flood a
+// response. Past it, literalsTruncated says so and the declaration has the rest.
+const MAX_LITERALS = 60;
 // A doc summary is one sentence; the rest of the comment is the file's to keep.
 const MAX_DOC_CHARS = 200;
 // Neighbor lists are a map, not a dump; keep them scannable.
@@ -101,7 +106,6 @@ export function runDetails(
     if (sig !== undefined) detail.signature = sig;
     const doc = docOf(graph.project, node);
     if (doc !== undefined) detail.doc = doc;
-    const signatureLiterals = literalSummaries(sig);
     const decorators = decoratorsOf(node);
     if (decorators !== undefined) detail.decorators = decorators;
     const implementation = evidenceCoordinatesOf(node.implementation);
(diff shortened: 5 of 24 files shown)


samchon:
Rebased onto current `master`. The branch was opened from a stale local `master` that predated #725 and carried two unrelated in-progress lint commits; both are gone from the branch, and the PR is now this fix alone on top of `c93435207`.

Rebasing onto #725 surfaced something the original base could not show, now in `73a63be61`:

**The dump schema had to move to v2.** `ITtscGraphDump.IProvenance.schemaVersion` is documented as moving "when a field is added, removed, or redefined", and `literals` is an added field. Left at v1, a stale `ttscgraph` and a current client agree on every number the wire checks while the field `details` now answers from is simply absent — and an absent value set is indistinguishable from a type that has none, which is the same silent under-report #732 is about, arriving by a different road.

**The resident session was the actual hole.** `TtscGraphSession` held the envelope to `protocolVersion` and never held the dump it carries to `schemaVersion`. The two are independent by design, and #725 shipped protocol v1 and schema v1 together — so a binary from that release passes the frame check and hands back nodes with no `literals`, on the MCP server's own path, the one `details` answers from. Bumping the protocol instead would have papered over it by conflating two versions the design deliberately separates, so the session now refuses the body the way it already refuses the frame. `DUMP_SCHEMA_VERSION` is exported rather than duplicated, and `dump_schema_version_matches_the_typescript_client_test.go` follows it across the `export` — the regex change is the gate keeping up with the constant, not loosening: it still requires the declaration and still reads the number.

Verified on the new base:

- `go test ./internal/graph/ -count=1` and `go test ./cmd/ttscgraph/ -count=1` — both green, including the four new literal cases and #725's serve/provenance suites.
- The schema gate proven in both directions: forcing `Go=2 / TS=3` fails it (`dump schema drifted: Go produces v2, ... reads v3`), and it passes once restored, so the `export` did not blind it.
- `pnpm --filter ttsc shim:audit` — OK.
- Typecheck of `packages/graph` — clean.
- Real `ttscgraph dump` on the new base: `provenance.schemaVersion: 2`, and the values unchanged from the original verification — `Kind` all five, `Colors` (multi-line enum) all three, `Indirect` all six through the alias, `Widened` absent, escaping intact.

samchon:
## Self-review

Went back over the whole diff rather than the parts I had just written. Two real gaps came out of it, both now closed, plus one risk that measured clean.

**The truncation marker had only its negative** (`3a10e7615`). The suite asserted `literalsTruncated` stays off for value sets that fit, and never that it comes on for one that does not — the assertion the field exists for. A `literals` that claims completeness while silently returning a prefix is the shape of #732 itself, and it would have shipped green. The new case crosses two unions in a template literal so the checker expands 72 members, then pins that the dump keeps all 72 while the response cuts to 60 and says so. Verified against a real dump and the real `runDetails` before writing the assertions, not after: `Big count=60 truncated=true first="a0"`, `Small count=2 truncated=undefined`.

**The value rendering was verified by hand and by nothing that runs** (`d2ddac9f1`). Every literal kind the pass claims to handle now has a case — numbers and bigints (`1n`), which the old string-only scrape could not report at all; `null` and `undefined`, which name a value without being literals; and a string carrying a quote and a tab, which is the fixture that justifies using the checker's own `ValueToString` over a local formatter (`"he said \"hi\""`, `"tab\there"` — Go's own quoting is not TypeScript's).

**Dump cost measured, not assumed.** `collectLiterals` asks the checker for the declared type of every type alias and enum, including ones nothing references, so I built a binary from `master` and one from this branch and dumped `packages/lint` three times each. No regression — the spread is noise (base 660–1843ms, branch 483–1981ms) and the dump grows 0.1% (933,999 → 935,042 bytes). The checker resolves lazily and caches, and `collectTypeRefs` already walks these types, so the pass mostly rides work the build was doing anyway. No syntactic pre-filter needed, which is good: one would have put a heuristic back on the path this fix exists to take off it.

Also checked and clean: `literals` has exactly one consumer (`runDetails`), so no other runner needed updating; the viewer reads no node fields beyond its own, and an added optional field cannot reach it; `dump_never_serializes_a_list_as_null_test.go` is satisfied because absence here is `omitempty` + optional, the same shape `modifiers` and `decorators` already use, and absence is unambiguous now that the schema gate rules out "an older producer never resolved it"; determinism holds (`dump_is_a_function_of_the_source_test.go` passes — constituent order is the checker's, node order is the dump's sort); declaration-merged enums resolve to the same complete list from either statement; intersections, template literals that do not expand, type parameters, and `unique symbol` all fall to the non-enumerable branch rather than to `Types()`, which panics off-union and is why the flag test guards it.

samchon:
## CI: `ttsc fast` was red, and not from this branch

`test_release_workflow_runs_preflight_before_publication` failed with `preflight (index 1906) must run before build (index 186)`. Pre-existing on `master` — this branch does not touch `release.yml` — and a false alarm: the workflow's order is right (preflight at line 51, `pnpm run build` at line 53). What sat at index 186 was #726's disk-reclaim comment, prose that names `` `pnpm run build` `` a dozen lines above the step that runs it. The gate scanned raw text, so it read that sentence as the build and concluded the build ran first.

Fixed in `5447944f5` by dropping comment lines before locating the steps. Order is a property of what executes, so prose about a command should never have counted as the command; this also closes the class rather than the instance, since the next comment to mention a publish step would have turned it red the same way. Verified both directions — the real workflow passes (`preflight=1083, build=1161`), and moving the build ahead of the preflight still fails it, so the gate is more accurate rather than looser.

Included here because CI has to be green to mean anything on this PR. Happy to split it out if you would rather it landed on its own.

PR #736, opened 2026-07-17
feat(lint): let a project rule publish editor completion hints

samchon:
Part of microsoft/typescript-go#733. Addresses the hint-channel row of that table.

## What this does

A project rule that indexed something an editor could complete against can now publish it, and ttscserver serves it:

```go
func (indexRule) Hints(ctx *rule.HintContext) []rule.Hint {
  index, ok := ctx.State.(*evidenceIndex)
  if !ok { return nil }
  return []rule.Hint{{
    Insert:  "pricing",
    Detail:  "Pricing",
    Trigger: rule.HintTrigger{Scope: rule.HintScopeJSDoc, After: "@evidence docs/spec.md#"},
  }}
}
```

Nothing changes for a rule that does not implement the marker.

## The constraint that shaped everything

**The corpus travels; the rule does not.** The lint engine is a separate process that reloads the Program on every invocation, so nothing can ask a rule a question per keystroke. `HintTrigger` is therefore declarative data rather than the Go predicate the API obviously wants — a closure does not survive the process boundary.

This is not a preference. Surveying the field: every system where a rule contributes completion (Roslyn, IntelliJ) does it **in-process on a shared AST**; every system where analysis is a subprocess (ESLint, clippy, Semgrep, SonarLint, gopls, Ruff, PHPStan) feeds **diagnostics only**. rust-analyzer proves it structurally — `ide-completion` has no dependency on clippy at all. [typescript-go#2824](https://github.com/microsoft/TypeScript/issues/63800) hit the identical wall in the identical language and reached the identical answer: *"having callbacks into client code … is something we want to avoid,"* replaced by a static mapping API.

So the corpus-snapshot shape is the only one available, and there is no reference implementation to copy.

## Design notes worth reviewing

**Pull, not push.** `Report` is push because a finding is discovered mid-walk and belongs to the node under it. A corpus is a projection of *finished* state — a rule pushing hints while building that state would publish the entries it had found so far. Pull also buys three things free: the passed-and-published gate, zero cost during `ttsc check` (nothing asks, so no corpus is allocated), and option-correctness, since the corpus is a projection of the state `Check` built under its options. A rule that is `off` publishes nothing with no code of its own.

**Longest `After` wins.** That single rule is what lets a corpus be layered without a query language. A plugin must publish `"@"`, `"@evidence "`, and `"@evidence docs/spec.md#"` together — it cannot be asked per keystroke — and the most specific match is the one the user is inside.

**A literal, not a regex.** Go's `regexp` is RE2 so ReDoS is not the objection; the objection is that a pattern cannot be validated at discovery time. The one shipped example of plugin-supplied regex driving completion is Tailwind's `experimental.classRegex`, which [OOMs the editor](https://github.com/tailwindlabs/tailwindcss-intellisense/issues/893), [fires on any string](https://github.com/tailwindlabs/tailwindcss-intellisense/issues/1086), and is not in their README. Their *supported* siblings are deliberately coarser and structural.

**Scope is refused when unknown, not ignored.** A line prefix cannot tell `@evidence` in a doc comment from `@Injectable` above a class — both end in `@`. A hint from a newer plugin than the host must contribute nothing rather than fire in every decorator and string literal.

**The fetch is async and answers nil until it lands.** `lsp-hints` is unlike `lsp-command-ids` and `lsp-code-action-kinds`, which ignore their arguments and never build a Program; a corpus must load one. Paying that on initialize would delay every editor session for a feature most projects do not use; paying it on first completion would freeze the popup. "No hints yet" degrades honestly.

**Trigger characters merge; they never replace.** tsgo's own list is what wakes its completion. Substituting a plugin's would silence the compiler to make room for a rule.

**Every response shape is handled.** LSP allows a bare array, a `CompletionList`, or null, and tsgo picks per request. Guessing wrong does not error — it silently drops either side's items. null is the common case here rather than an edge one: tsgo returns it for JSDoc prose, which is exactly where these hints live.

## Verification

`go test ./internal/lspserver/` — seven cases:

| Case | Pins |
| --- | --- |
| `LongestTriggerWins` | the layering rule, and that the filter is the text after the trigger |
| `RefuseOutsideScope` | the negative twin — decorator position, unknown scope, degenerate hints |
| `CursorInJSDocTracksTheBlock` | the scope scan, including after-close and line-comment |
| `MergePreservesUpstreamItems` | all three response shapes, both sides surviving, `isIncomplete` preserved |
| `MergeLeavesUpstreamErrorsAlone` | an error response is not rewritten into a half-answer |
| `OffsetForPositionCountsUTF16` | LSP counts UTF-16 units; bytes or runes would land the cursor wrong |

`node scripts/test-go-lint.cjs` green, no existing test modified.

**Change Integrity**, both called out per the development skill:

1. This widens the public contributor API (`rule.HintRule`, `Hint`, `HintTrigger`, `HintContext`) and adds a sixth LSP verb. Additive — no existing signature changes.
2. It adds `./internal/lspserver` to `test-go-ttsc.cjs`. **That package had zero Go tests and no runner targeted it**, so the tests here would never have run — and a test that never runs is worse than none, because it reads as coverage. The tests live beside the package rather than under `./test` because they cover unexported logic; exporting the matcher to test it would put an internal decision on the public surface.

Website walkthrough updated in the same change.

## What I could not verify

- **`completionItem/resolve`.** If tsgo advertises `resolveProvider: true`, a client may send resolve for a plugin item to tsgo, which never produced it. Items here are fully resolved on arrival so there is nothing to fetch, but whether tsgo tolerates an unknown item is **unverified**. Worth a reviewer's eye.
- **Prettier on the `.mdx`.** My checkout has no `node_modules`, so `prettier.config.js` could not load its plugins. Every prose line I added is single-line, which is what `proseWrap: "never"` wants, and the only short lines are inside a fenced Go block that `embeddedLanguageFormatting: "off"` leaves alone — but the formatter has not actually run on it.
- **End to end in a real editor.** The pieces are unit-tested; nothing has driven a live VS Code session against a plugin publishing a corpus.

## Provenance

Found while building [`@samchon/evidence`](https://github.com/samchon/evidence), whose project rule indexes every markdown section and TypeScript symbol per Program and, until now, could only complain about a citation — never offer one. Its `HintRule` implementation is written and green against this branch, and parked on a branch it cannot merge, for the reason recorded in microsoft/typescript-go#733: a contributor cannot reference an unreleased host API without failing its own build outright.


samchon:
Part of microsoft/typescript-go#733 — the umbrella carrying the shared context for this area: the verified subset tables, the four structural walls, the prior-art survey, and the traps.

samchon:
CI caught a real defect in `aa85607c5`, worth naming since it is the kind this channel invites.

The hint fetch logged every plugin that rejected `lsp-hints` as an unknown command — which is **every plugin built before this channel existed, permanently, on every editor session**. `TestLSPNativePluginSourceRoutesSidecarProtocol` asserts the bridge writes nothing to stderr, and it was right to fail.

Absence is an answer. Not knowing an optional verb is the common case, not the exceptional one — the same reasoning that makes `CompletionHints` an optional interface rather than a `PluginSource` method, applied to the wire instead of the type. Invalid JSON stays loud: a plugin that answered and answered wrongly implemented the verb and got it wrong.

Now pinned by its own test rather than left to the integration case that caught it — that one asserts on stderr as a proxy for "nothing went wrong" and would never say *why* a log line is forbidden here.

Also, for the record on the earlier red: I could not reproduce this locally. `test-go-lint.cjs` needs node_modules and my checkout has none, so the lane that caught it was only ever going to be CI. I should have said that in the description rather than implying the Go lanes were covered.

PR #739, opened 2026-07-17
fix(graph): name the fields details cut

### LINKED ISSUE #737: `@ttsc/graph`: `details` promises a `truncated` marker it has no field for
Found while fixing #732, verified, and deliberately left out of #735 because deciding it needs a benchmark measurement rather than a patch.

## The contract

`RESULT_AUDIT` covers `details`, and tells the caller:

> This is the structure the graph holds for the handles you named... Trust every fact it gives and re-verify none: a returned span is the citation, not a cue to open the file. **Where the walk was bounded, `truncated` marks it.**

`ITtscGraphApplication.IOutput.audit` says the same: for `trace`, `details`, `overview` the result is "the structure the graph holds for the handles named, **bounded only where `truncated` says**".

## The gap

`ITtscGraphDetails` has no `truncated` field. It exists on `ITtscGraphTour`, `ITtscGraphEntrypoints`, and `ITtscGraphTrace`, but not here. So `details` cannot express the marker its own audit promises, while it bounds five things:

| field | cap | marked? |
|---|---|---|
| `calls` / `types` / `implementedBy` | `dependencyLimit`, default **2**, max 4 | no |
| `dependsOn` / `dependedOnBy` | `neighborLimit`, default 2, max 3 | no |
| `members` | `memberLimit`, default 6, max 8 | no |
| `signature` | `MAX_SIGNATURE_LINES = 4` | no |
| `literals` | `MAX_LITERALS = 60` | **yes**, via `literalsTruncated` (#735) |

#735 closed the `literals` row, because there completeness is the field's whole claim. The rest are still unmarked.

## What is and is not silent

Worth being precise, because it changes the severity. The cap *value* is not hidden: `dependencyLimit`, `memberLimit`, and `neighborLimit` are request knobs whose defaults the tool schema publishes, so a model that reads the schema knows an omitted `dependencyLimit` yields 2 per group. (Two of those published defaults were themselves stale and are fixed in #735, with a gate.)

What the caller cannot know is whether the cap **bit**. `calls: [2 refs]` is the same payload whether the symbol calls exactly two things or forty, and the audit tells the model not to open the file and check. That is the same shape as #732 — a bounded answer that reads as a complete one — arriving through a different field.

## Why this is not a drive-by fix

Every default here was set by a benchmark result, deliberately: `3db685223` bounded graph evidence to keep the benchmarks token-positive, `77dc14d04` raised `DEFAULT_DEPENDENCIES` 1 → 2 to stop agents revalidating facts, `c5243ad1c` raised `tour.limit` 4 → 5 to restore benchmark discipline.

A truncation marker is not neutral against that tuning. Telling a model "there are more" is an invitation to ask again, and the whole point of these caps is that it should not. The marker could plausibly pay for itself (a model that knows the list is partial stops treating it as exhaustive) or cost real tokens (an extra `details` round per node). That is a measurement, not an opinion, and it belongs to whoever owns the benchmark axis.

## Options, if it is worth doing

1. **Per-field flags**, following `literalsTruncated`: explicit, but five more fields on a node.
2. **One `truncated?: string[]`** naming the cut fields (`["calls", "members"]`): one field, extensible, reads well; would mean folding `literalsTruncated` into it before it ships widely.
3. **Narrow the audit instead**: say plainly that `details` returns a bounded slice per the request's limits, and stop promising a marker. Cheapest, honest, and costs no tokens — but leaves the caller unable to tell 2-of-2 from 2-of-40.

Option 3 is the only one with no benchmark risk. Options 1 and 2 need a run before they can be justified.

samchon:
Closes #737.

## Intent

`details` caps every list it returns and the payload never said so. `calls: [2 refs]` is the same bytes whether the symbol calls two things or forty, while `RESULT_AUDIT` tells the caller the result is "bounded only where `truncated` says" and to re-verify nothing. `ITtscGraphDetails` had no `truncated` field at all — it exists on tour, entrypoints and trace, but not here — so the audit promised a marker the type could not express, and a slice read as the whole. That is the defect #732 was about, arriving through the neighbor lists instead of the value set.

Each node now names what it lost:

```json
{ "name": "many", "calls": [ ... 2 refs ... ], "truncated": ["calls"] }
```

It points, it does not dump. The graph still holds all of it and the request's own limits still say how much to send; the marker only tells the caller whether looking further is worth it, which is what an index is for. `literalsTruncated` from #735 folds into this field before it ever shipped in a release.

## The marker measures loss, not the cap

Reaching a limit is not the same fact as there being more. A function calling exactly `dependencyLimit` things is whole, and marking it would lie in the direction the marker exists to prevent — so dedup and counting run past the limit and compare. Members are still built one page at a time, because a member's signature is a file read; only the count runs long, off edges already in memory.

## Also here

A signature no longer runs past its own declaration. `signatureOf` stopped at a brace, a trailing semicolon, or the four-line cap, so a declaration ending in none of those read its neighbors instead. The span was already on the node. This is independent of the marker and is a plain correctness fix.

## Scope

Nothing else moves. `truncated` is a `details` field; tour, lookup and trace do not carry it and do not read `members`. No node or edge is added, so no ranking, flow, seed or payload outside `details` changes — this needs no re-measure of the graph arm.

`ITtscGraphApplication` and `RESULT_AUDIT` are untouched.

## Verification

Left to CI, per instruction. Before the split I drove the real `runDetails` over a real `ttscgraph dump` of a fixture with a function calling twenty others, one calling exactly two, one calling one, and a class of ten methods:

```
many        calls=2 members=undefined truncated=["calls"]
exactlyTwo  calls=2 members=undefined truncated=undefined
justOne     calls=1 members=undefined truncated=undefined
Wide        calls=undefined members=6 truncated=["members"]
```

`exactlyTwo` is the assertion that matters: at the cap and unmarked.

New case `test_ttscgraph_details_names_each_field_it_cut.ts` pins that end to end through the real binary, including the boundary. The two literal cases from #735 are updated to the new field.

### DIFF
--- a/packages/graph/src/server/runDetails.ts
+++ b/packages/graph/src/server/runDetails.ts
@@ -16,7 +16,8 @@ const MAX_SIGNATURE_LINES = 4;
 // A value set is the answer to "what may this be", so it is listed rather than
 // sampled: the cap is high enough that a real union or enum arrives whole, and
 // exists only so a type expanded from a template literal cannot flood a
-// response. Past it, literalsTruncated says so and the declaration has the rest.
+// response. Past it, `truncated` names the field and the declaration has the
+// rest.
 const MAX_LITERALS = 60;
 // A doc summary is one sentence; the rest of the comment is the file's to keep.
 const MAX_DOC_CHARS = 200;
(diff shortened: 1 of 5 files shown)


samchon:
## Self-review

Went after the two claims in the PR body that would be embarrassing if wrong, rather than re-reading what I had just written.

**"Tour is unaffected" — checked, and it was not obvious.** `runTour` calls `runDetails` directly (`runTour.ts:166`), with its own limits, so a new `details` field could ride into the tour payload and this PR's no-re-measure claim would be false. It does not: `nearbyAnchorsOf` reads only the reference lists off each node (`dependedOnBy`, `calls`, `implementedBy`, `dependsOn`, `types`) and maps them to `IAnchor{id,name,kind,file,line}`. `truncated` never reaches a tour. The claim holds, and it holds by the shape of the projection rather than by luck.

**Every cut path now goes through `cut()` or reports its own count.** Swept the file for `slice(0, `, `>= limit`, and `break`. What is left:

- `members` counts owned nodes off the edges and builds only the page — a member's signature is a file read, so building the dropped ones to count them would pay a class's whole member list to answer with six of it.
- `objectLiteralMembers` scans the literal's whole span instead of stopping at the cap. Bounded by `MAX_OBJECT_MEMBER_LINES` (300) and it reads no file the page did not already.
- `docOf` cuts at `MAX_DOC_CHARS` and appends `…` — it marks its own cut inline, and has all along.
- `signature` is the one bound not named in `truncated`. Left deliberately: it is *defined* as the declaration head, and a signature cut mid-parameter ends visibly mid-expression, so it does not read as complete the way `calls: [2 refs]` does. That is the whole reason this field exists — the invisible cuts, not the visible ones.

**`members` can be pushed twice** — a container node and an object-literal variable both feed the same field. Deduped through a `Set` before it goes on the node, so `truncated` can never say `["members","members"]`.

**The boundary is the assertion that matters**, and it is pinned in the e2e rather than argued: a function calling exactly `dependencyLimit` things comes back unmarked. A check that asked "did we reach the cap" instead of "was anything lost" passes every other case in this PR and fails only that one.

Nothing else in the diff touches a node, an edge, or a ranking. `ITtscGraphApplication` and `RESULT_AUDIT` are untouched.

samchon:
Superseded by #741. The experiment this PR's `truncated` marker implied — cap the lists and tell the caller they were cut — turned out to solve the wrong problem. Measured on typeorm, an uncapped `details(DataSource)` is 390 KB (~100k tokens), 380 KB of it the 1060 inbound type-references: a symbol's identity (members, values) is bounded by its declaration and should not be cut at all, while its fan-out is a trace/impact question that does not belong in a `details` answer whole. So the fix is not to mark the cut but to not cut the identity, and to keep fan-out an explicit small slice. Continued on `fix/graph-details-whole-identity`.

PR #740, opened 2026-07-17
feat(graph): answer an enum with its member names

### LINKED ISSUE #738: `@ttsc/graph`: `details` on an enum never answers with its member names
Found while fixing #732 and scoped out of #735, which fixed the half the issue asked for.

## An enum's members are not in the graph

The build pass records member nodes for classes and interfaces only — `collectMembers` is called from the `KindClassDeclaration` and `KindInterfaceDeclaration` arms of `collectStatements`, and nothing handles `KindEnumMember`. `contains` edges are then derived in the TypeScript loader from node ownership, so an enum owns nothing either.

Verified against a real dump of:

```ts
export enum Colors {
  Red = 'red',
  Green = 'green',
}
```

The graph holds exactly one node for it, `src/app.ts#Colors:enum`, no `Colors.Red`, and no edge out of `Colors`.

## What #735 did and did not fix

Before #735, `details` on a multi-line enum returned no member information at all: `literals` was scraped from the signature snippet, which stops at the line holding `{`, and `members` was empty because there are no member nodes. #735 makes `literals` a checker fact, so the **values** now come back (`["red", "green"]`).

The **names** still reach no consumer. `Colors.Red` is not a handle `details` or `trace` can resolve, it cannot be the endpoint of an edge, and it appears in no member outline. A caller that asks what to write gets `"red"` — the value — while the code writes `Colors.Red`.

## Why it matters beyond naming

An enum member is a real reference target. `paint(Colors.Red)` is an access the checker resolves, and the graph drops it: the edge has nowhere to land, so a trace of who uses `Colors.Red` specifically has no answer, and impact analysis on one member is indistinguishable from impact on the whole enum.

## Not obviously worth doing

Enum members are numerous and mostly leaves, so nodes for them would grow the graph for facts a tour never ranks — the same argument that keeps `closure` nodes out of tour seeds while leaving them addressable by name. The measured cost matters here, and the `closure` precedent (3% node growth on Vue, 5% on VS Code, tour payload unchanged) suggests the shape a decision could take: record them, keep them out of ranking, answer when asked by name.

Options:

1. **Member nodes**, like class/interface members, keyed `path#Colors.Red:property`. Makes them handles, edge endpoints, and members. Costs graph size.
2. **Names alongside values on the enum node** — `literals` gains a parallel member list, or becomes objects. Cheap, no new nodes, but they stay unaddressable and it complicates a field whose contract is now clean.
3. **Leave it.** `literals` answers "what values does this admit", which is the common question, and `Colors` as a whole is already a handle.

samchon:
Closes #738.

## Intent

`Colors` is a top-level exported declaration and has been a node in this graph all along. Ask `details` about it and you get `signature: "export enum Colors {"`, an empty member outline (its members are not nodes, so nothing hangs off `contains`), and — since #732 — its values. You never get the names. The code writes `Colors.Red`, never `"red"`, so the one kind whose entire content *is* its member list was the kind `details` could not describe, and a caller that had already named the enum opened the file anyway. That is the grep this index exists to remove.

Now:

```json
{
  "name": "Colors",
  "kind": "enum",
  "members": [
    { "name": "Colors.Red", "kind": "property", "signature": "Red = \"red\"" },
    { "name": "Colors.Green", "kind": "property", "signature": "Green = \"green\"" }
  ],
  "literals": ["\"red\"", "\"green\""]
}
```

`literals` still answers what values the enum admits — the question a serializer asks. `members` answers what to write.

## Why not member nodes

I built that version first, because #738 as I filed it said "members are not nodes" — a fix stated as a problem. It is the wrong instinct: this graph indexes declarations so an agent greps and reads less; it is not an AST mirror, and completeness is not the goal. `Colors.Red` is a literal, unambiguous string that `grep -rn "Colors.Red"` answers exactly and cheaply. The index earns its keep on what grep cannot do.

Measured rather than argued. Member nodes added 194 nodes (1.0%) and 223 edges on typeorm, and the tour moved in 3 of 8 offline cells — `getObservableInputType`'s rxjs flow went from six steps to eight, the two additions being `-[accesses]-> ObservableInputType.Own` and `.InteropObservable`: leaves that end the walk, an arbitrary two of seven members, sitting in the field that is supposed to say how the code runs. Seed order shifted in `rxjs dedicated`. That is the closure lesson again, and the fact that it would then need a tour gate is the tell that it should not be let in.

**The line: filling in a node the graph already holds is indexing. Minting nodes to carry detail is over-information.**

## Scope

- The names and values come out of the same constituents #732 already resolves, so the pairing is the checker's, not a zip of two lists that can drift. A computed member the checker cannot fold keeps its name and drops its value.
- Dump schema moves to **v3** — a field was added, which is what moves it, per `IProvenance.schemaVersion`.
- No node, no edge, no ranking, no flow, no seed. `details` differs only when the handle is an enum. **The graph arm needs no re-measure.**
- `ITtscGraphApplication` and `RESULT_AUDIT` untouched.

Stacked on `master`, independent of #739.

## Verification

Left to CI, per instruction. Before pushing I drove the real binary and the real `runDetails` over a fixture:

```
dump  schemaVersion: 3
      Colors   [{"name":"Red","value":"\"red\""}, ...]
      Implicit [{"name":"First","value":"0"},{"name":"Second","value":"1"}]
      member became a node: no

details Colors    members=["Colors.Red | Red = \"red\"", ...]  literals=["\"red\"", ...]
        Implicit  members=["Implicit.First | First = 0", ...]
        Cls       members=["Cls.run | public run(): void {}"]
```

`Implicit` is the case that pins the source of the facts: 0 and 1 appear nowhere in the file. `Cls` is the negative twin — a class's outline still comes from its member nodes and is untouched.

New cases: `enum_members_carry_their_names_on_the_enum_node_test.go` (pairs, implicit numbering, no member node minted, and a class carrying none of it) and `test_ttscgraph_details_answers_an_enum_with_its_member_names.ts` end to end.

### DIFF
--- a/packages/graph/src/model/loadGraph.ts
+++ b/packages/graph/src/model/loadGraph.ts
@@ -24,7 +24,7 @@ const MAX_DUMP_BYTES = 1024 * 1024 * 1024;
  * body's are independent, so a producer can speak this protocol and still send a
  * body from another schema.
  */
-export const DUMP_SCHEMA_VERSION = 2;
+export const DUMP_SCHEMA_VERSION = 3;
 
 /**
  * Build the resident {@link TtscGraphMemory} for a project by running `ttscgraph
--- a/packages/graph/src/server/runDetails.ts
+++ b/packages/graph/src/server/runDetails.ts
@@ -152,6 +152,15 @@ export function runDetails(
       );
       if (list.length > 0) detail.members = list;
     }
+    // An enum's members ride on its own node rather than on `contains` edges,
+    // because they are not nodes: the outline above finds nothing for an enum
+    // and always did. Its signature stops at the `{`, so without this the one
+    // kind whose entire content is its member list answered with none of it.
+    // Uncapped like every other identity list — the members are the enum.
+    if (node.kind === "enum") {
+      const list = enumMembers(node, memberLimit);
+      if (list.length > 0) detail.members = list;
+    }
     if (node.literals !== undefined && node.literals.length > 0) {
       detail.literals = node.literals;
     }
@@ -223,6 +232,28 @@ function members(
     if (out.length >= limit) break;
   }
   return out;
+}
+
+/**
+ * An enum's members, owner-qualified so the name reads the way the code writes
+ * it, with the value each carries as its signature.
+ *
+ * The name is why this exists. `literals` answers what values the enum admits,
+ * but a caller writes `Colors.Red` and never `"red"`, so an enum the graph
+ * already held sent a caller that had named it to the file for the one fact it
+ * came for (#738).
+ */
+function enumMembers(
+  node: ITtscGraphNode,
+  limit: number,
+): ITtscGraphDetails.IMember[] {
+  return (node.enumMembers ?? []).slice(0, limit).map((member) => ({
+    name: `${node.qualifiedName ?? node.name}.${member.name}`,
+    kind: "property",
+    ...(member.value !== undefined
+      ? { signature: `${member.name} = ${member.value}` }
+      : {}),
+  }));
 }
 
 function objectLiteralMembers(
--- a/packages/graph/src/structures/ITtscGraphNode.ts
+++ b/packages/graph/src/structures/ITtscGraphNode.ts
@@ -77,6 +77,19 @@ export interface ITtscGraphNode {
   literals?: string[];
 
   /**
+   * What an enum declares, in checker order: the name a caller writes and the
+   * value it carries. Absent on every other kind.
+   *
+   * `literals` says which values the enum admits, which is what a serializer
+   * asks. The code says `Colors.Red`, so the names are the other half, and
+   * without them a caller that had already named the enum still had to open the
+   * file to learn what to type. The members are not nodes — `Colors.Red` is a
+   * string a grep finds exactly — so this fills in the node the graph already
+   * holds instead of minting one per member.
+   */
+  enumMembers?: ITtscGraphNode.IEnumMember[];
+
+  /**
    * Decorators written on this declaration, in source order: raw facts
    * (`@Controller`, `@Get`) a consumer interprets without re-parsing source.
    */
@@ -90,4 +103,18 @@ export interface ITtscGraphNode {
    * function assignment separate from its declaration.
    */
   implementation?: ITtscGraphEvidence;
+}
+export namespace ITtscGraphNode {
+  /** One member of an enum: the name a caller writes and the value it carries. */
+  export interface IEnumMember {
+    /** The member's own name, unqualified (`Red` on `Colors.Red`). */
+    name: string;
+
+    /**
+     * The value it carries, in TypeScript source form (`"red"`, `1`). Absent
+     * for a computed member the checker could not fold to a constant; the name
+     * still stands.
+     */
+    value?: string;
+  }
 }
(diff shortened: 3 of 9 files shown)


samchon:
## Self-review

Found a real bug in my own patch, pushed as `e0983e557`.

I read the member list off the declared type's constituents, reusing what #732 already resolves — which felt economical and was wrong. **A type is a set.** The checker folds two members carrying one value into a single constituent, so:

```ts
export enum Dup { A = 'x', B = 'x', C = 'y' }
```

came back as `[A, C]`. `B` vanished, silently, with no marker and no way for a caller to tell. That is #732's defect exactly, reproduced by me, out of #732's own instinct: taking a declaration fact from a type.

The member list is a fact about the *declaration*, and nothing folds a declaration. It now walks the enum's own members; each member's value still comes from the checker per member, so implicit numbering still resolves with nothing reading source.

`literals` needed no change and is right as it stands: the value set really does hold `"x"` once, because a set is what it is. Two members, two names, one value — three true statements, and the two fields now each carry their own instead of one of them lying. That the same fixture pins both is the point.

Pinned in both suites (`Dup` in the Go case and the e2e), and the Go case fails on the old shape:

```
FAIL: TestEnumMembersCarryTheirNamesOnTheEnumNode
    a member sharing another's value was dropped: [A C]
```

Verified against the real binary after the fix: `members: ["A","B"]`, `literals: ["\"x\""]`.

### Also checked, clean

- **No member node minted.** Asserted in the Go case, not just intended — a stray `Colors.Red` or `Red` node fails it. The diff greps clean for `KindEnumMember`, `SymbolFlagsEnum` and `collectMembers`.
- **Enums only.** `enumMembers` is read under `node.kind === "enum"`, and the Go field is written under `kind == NodeEnum`. The class in both fixtures is the negative twin: its outline still comes from its member nodes.
- **Ordering** is the declaration's now, which is also what makes it stable — the checker's union order sorts by type id, and the earlier version rode on that.
- **Merged enum declarations** each write the same list to the same node; idempotent.
- **Empty enum** yields no members and no key.
- **Schema v3** is the rule applying (a field was added), and the Go/TS gate holds the pair.

### One wart, flagged rather than fixed

`details` caps an enum's members at `memberLimit` (6, max 8) like any container, while `literals` runs to 60. So a 20-member enum answers with 20 values and 8 names. The names are the half you write, so that is the wrong way round.

I left it because the fix is a judgement call, not an oversight: either enums get a cap of their own (the member list *is* an enum's content, not an outline of it), or `literals` becomes redundant for enums now that members carry name and value together and should be dropped there. Both change a contract, and #739 is what makes the cut visible either way. Worth deciding, not worth me guessing inside this PR.

samchon:
Restacked on #742 (details identity policy). The two overlapped in `runDetails`' member handling, so this now sits on top of it rather than beside it.

The alignment is exact: an enum's members are identity, so they follow #742's rule and are **uncapped** — a 20-member enum answers with all 20 (verified against the real binary), not the old 6/8 page. `enumMembers` renders through the same `memberLimit`, which #742 defaults to unlimited. `literalsTruncated` from the pre-rework #739 is gone; enum members and values both come whole.

Nothing else changed in this branch — the Go `EnumMembers` field, the dump schema bump to v3, and the checker-per-member value resolution are as reviewed. Base is `fix/graph-details-whole-identity`; it merges after #742.

PR #742, opened 2026-07-17
fix(graph): details returns a symbol's whole identity, a slice of its fan-out

### LINKED ISSUE #737: `@ttsc/graph`: `details` promises a `truncated` marker it has no field for

samchon:
Closes #737. Supersedes #739.

## The question, answered with a measurement

`details` capped every list it returned. #739 tried to *mark* the cuts with a `truncated` field. Building it and measuring on real repos showed that was the wrong problem.

An uncapped `details` on typeorm:

| handle | capped | uncapped | what blew up |
|---|---|---|---|
| `DataSource` | 3 KB | **390 KB** (~100k tokens) | `dependedOnBy` **1060** |
| `EntityMetadata` | 3 KB | 92 KB | members 120, `dependedOnBy` 190 |
| `transformColumnsRecursive` | 3 KB | 102 KB | `calls` 121 |
| `SimpleColumnType` | 3 KB | 3 KB | — (105 literals) |

The 1060 on `DataSource` are all 1-hop `type_ref`s — every place that writes `ds: DataSource`. Not a deep walk; a popular type. That splits the fields cleanly:

- **Identity** (`members`, `literals`) is bounded by the *declaration*: `EntityMetadata` has 120 members, `SimpleColumnType` 105 values. That is the symbol itself, and the caller named the handle to see it. Capping it just sends the model to read the 1134-line source — the read the index exists to remove.
- **Fan-out** (`calls`, `types`, `dependedOnBy`) is bounded by *popularity*, not the symbol. Returning it whole is a hundred thousand tokens of "who uses this" in a "what is this" call — a `trace`/impact answer nobody asks `details` for.

## The change

- `members` and `literals`: **uncapped**. `DataSource` is 8 KB now with all 41 members; `EntityMetadata` 18 KB with all 120.
- `calls` / `types` / neighbor lists: a small orientation slice by default (`dependencyLimit`/`neighborLimit`, default 2), documented as pointing to `trace` for the whole fan-out.
- `truncated` (unreleased, from #739) is **removed** — the identity is not a slice to mark, and the fan-out slice is a documented request default, not a silent bound.
- `signature` still stops at its own declaration span instead of reading a neighbor's lines — an independent correctness fix that rode along from #739.

The tour is unaffected: it calls `runDetails` with its own explicit small limits, so its payload does not move.

## One seam, left for measurement

`RESULT_AUDIT` still says a result is "bounded only where `truncated` says." That sentence covers `trace`'s multi-hop walk (which keeps `truncated`); details' fan-out is now a 1-hop documented slice. Whether the audit needs its own word for that is a benchmark question — the audit text is a measured surface, so I did not reword it blind. `ITtscGraphApplication` is untouched.

## Verification

Left to CI. Measured offline on typeorm before pushing: `DataSource` 390 KB → 8 KB, identity intact, fan-out sliced. New e2e `test_ttscgraph_details_gives_whole_identity_and_caps_fan_out.ts` pins a 20-member class returning all 20 while its 20 inbound references come back sliced. The two `truncated`-specific tests from #739 are removed; the literals-enumeration test drops its truncated assertion.


samchon:
## Self-review

The load-bearing claim of this PR is "the tour is unaffected, so the graph arm needs no re-measure." `runTour` calls `runDetails` directly (`runTour.ts:166`), so that claim is not free — I measured it rather than asserting it.

Built this branch's `runTour` and master's `runTour` against the same typeorm dump, same benchmark question, and diffed the payloads:

```
EntityMetadata members  master: 6   this branch: 120   (the intended change is present)
tour payload            byte-identical
```

The tour passes explicit limits (`memberLimit:4, dependencyLimit:2, neighborLimit:2`), and under the new policy those resolve to the same numbers the old `bound()` gave — `limitOf(4)=4`, `capOf(2,2,4)=2`, `capOf(2,2,3)=2` — so the tour's embedded `details` slice does not move a byte, while a bare `details` on the same class goes 6→120 members. The change lands exactly where intended and nowhere else.

### Also checked

- **The fan-out cap is what bounds the blowup, not the ref shape.** `DataSource`'s 1060 inbound refs are `dependedOnBy`, which only appears under `neighbors:true`; `types`/`calls` are capped by `dependencyLimit`. All fan-out paths route through `capOf`, none through `limitOf`.
- **`limitOf` vs `capOf` split is total.** Only `memberLimit` (identity) uses `limitOf` (unbounded default). `dependencyLimit`/`neighborLimit` (fan-out) use `capOf` (default 2, ceiling). Grepped to confirm no fan-out list reads `memberLimit`.
- **`objectLiteralMembers`** (a `variable`'s literal outline) also rides `memberLimit`, so an object literal's members are uncapped too — consistent, since that is also identity, and it stays bounded by `MAX_OBJECT_MEMBER_LINES` against a runaway source scan.
- **`signature` span fix** is independent and unchanged from its prior review.

### The one seam, restated

`RESULT_AUDIT` still says "bounded where truncated says." Details' fan-out is now bounded (default 2) and carries no `truncated`. I read "the walk was bounded" as `trace`'s multi-hop walk — details' 1-hop fan-out is a documented request default (`dependencyLimit`), not a walk — so the runtime audit is arguably still true. But it is a judgement call on measured text, so I left it for the benchmark rather than reword it blind. If a bench cell shows a model trusting `calls:[2]` as complete, that is the signal to add a word.

samchon:
Added the audit reconcile the PR flagged as an open seam, and cleaned the branch history.

**Audit.** `details` now carries its own `RESULT_AUDIT_DETAILS` instead of the shared `RESULT_AUDIT`. The shared one says "trust every fact, do not open the file, bounded only where truncated says" — which fit details only by contradiction once identity went uncapped (trust a member list, but open the file to complete it). The details audit states the split it actually returns: identity is complete (trust it, do not read the file), fan-out is a short orientation slice (`trace` for the whole). This is the sentence that makes the identity-uncap legible to the model — and it removes the "truncated marks bounds" claim that details no longer honors. `trace` and `overview` keep `RESULT_AUDIT`. The two-audits essay and `IOutput.audit` doc go to three; the audit-ordering e2e now exercises details.

So the whole change is now internally consistent: nothing about details claims a bound it does not carry.

**History.** Rebuilt the branch as two clean commits (whole-identity, audit) on top of current master. The earlier history had a commit that introduced `truncated` and a later one that removed it — churn from the #739 approach this supersedes. Net tree is identical; only the commits are reorganized.

The RESULT_AUDIT text change is a benchmark-measured surface, so this is exactly the state to measure: audit and behavior now agree, which is what the graph arm should be re-run against.

samchon:
## Benchmark: old details vs new, dedicated arm

Light A/B before the full sweep. `claude-code-sonnet`, `ttsc-graph` arm, `--prompt-family=dedicated`, one run per cell, old `@ttsc/graph` (master, capped details) vs this branch, same fixtures.

| repo | old (tokens / reads / turns) | new (tokens / reads / turns) | tokens |
|---|---|---|---|
| rxjs | 2.52M / 5 / **25** | 1.54M / 7 / **11** | **−39%** |
| zod | 1.98M / 8 / **18** | 1.30M / 5 / **10** | **−35%** |
| excalidraw | 2.60M / 6 / **27** | 1.12M / **0** / **8** | **−57%** |

All three improve, tokens down 35–57%, turns roughly halved. The clearest signal is turns (variance-robust): the agent stops thrashing. excalidraw is the shape the change is for — new details answered the dedicated question in 8 turns reading **zero** files, where old spent 27 turns and 6 reads. rxjs old called `details` 0 times and toured/read instead; new called it 6 times and trusted it — which is what the new audit ("identity complete, do not open the file") is meant to induce.

Caveat: one run per cell, so absolute numbers carry agent variance. But three cells all move the same way with turns cut in half, so the direction is not noise. The full arm re-measure is the owner's to run.

Issue #733, opened 2026-07-17
epic(lint): a rule contributor can complain, and nothing else — the editor-assistance gap, surveyed


samchon:
## What this is

An umbrella for the contributor-facing gaps found while building [`@samchon/evidence`](https://github.com/samchon/evidence), an external `@ttsc/lint` rule contributor, against 0.19.1. It exists so a campaign can pick any child up cold: each finding below is verified against source, each child issue is self-contained, and the traps are named so nobody spends a day rediscovering them.

**The one-sentence finding:** a rule contributor can complain, and nothing else. Every other channel the host has — a choice of fixes, a data answer, an editor hint — is either internal-only or absent, so a contributor holding useful knowledge either drops it or reimplements its analysis in TypeScript and ships a second extension.

## Children

| # | Title | State | Depends on |
| --- | --- | --- | --- |
| microsoft/typescript-go#727 | project rules cannot decline the type checker | PR **#728** open, green, no existing test changed | — |
| microsoft/typescript-go#730 | contributors cannot offer a choice of fixes | open, unimplemented | — |
| microsoft/typescript-go#729 | contributors have no data channel | open, **deferred pending microsoft/typescript-go#730** — see below | — |
| microsoft/typescript-go#731 | four `boundaries/*` rules discard the allowed set | open, unimplemented | — |
| microsoft/typescript-go#734 | `LSPDiagnostic` is missing four LSP fields | open, unimplemented | — |
| *(this)* | editor hints from a rule corpus | PR **#736** open — rule API, verb, proxy, 7 tests, docs | microsoft/typescript-go#727 helps, not required |

**#729 is deliberately deprioritized by its own author.** I filed it asking for a general data channel, then found microsoft/typescript-go#730 does most of the same work far more cheaply. Read microsoft/typescript-go#730 first; treat microsoft/typescript-go#729 as the fallback only if microsoft/typescript-go#730 proves insufficient.

## The verified picture

**The public contributor surface is a strict subset of the internal one.**

| `rule.Context` (contributors) | `linthost.Context` (built-ins) |
| --- | --- |
| `Report`, `ReportFix`, `ReportRange`, `ReportRangeFix` | those four, **plus** `ReportSuggestion`, `ReportFixSuggestions`, `ReportRangeSuggestion` |

[...]

PR #743, opened 2026-07-17
fix(lint): four boundaries rules name the allowed set instead of hiding it

### LINKED ISSUE #731: fix(lint): four boundaries rules compute the allowed set, then report only that you got it wrong
## Problem

Four `boundaries/*` rules compute the set of things that *would* be acceptable, use it to decide, then throw it away and report only that the code is wrong. The reader is told "not allowed" and has to go read `lint.config.ts` to learn what is.

This is out of step with the rest of the corpus. Three other rules with the same shape all name the set in the message — so this is a handful of outliers against an established convention, not a missing feature.

The sharpest case is `boundaries/entry-point` (`linthost/rules_boundaries.go:112`):

```go
reportBoundaryDependency(ctx, dep, fmt.Sprintf("Import %q through an allowed boundary entry point.", target.RelativePath))
```

The message uses the word **"allowed"** and names not one allowed thing. `target.Entry` — the entry patterns it means — was read two lines earlier at `:110` to make the decision.

## Evidence

All four sites, with the set that is in scope at the moment of reporting:

| Rule | Report | The set, in hand | Message today |
| --- | --- | --- | --- |
| `boundaries/element-types` | `:60` | `rule.Allow`, via `evaluateBoundaryPolicy` returning the matched rule (`:426-430`) | `Import from boundary element %q is not allowed in %q.` |
| `boundaries/external` | `:80` | `opts.Allow`, read **on the line above** at `:76` | `External dependency %q is not allowed.` |
| `boundaries/entry-point` | `:112` | `target.Entry`, read at `:103` and `:110` | `Import %q through an allowed boundary entry point.` |
| `boundaries/no-unknown` | `:160` | `opts.Elements` and their patterns | `Imported file %q does not match any configured boundary element.` |

`boundaries/external` is the tidiest to read:

```go
allowed := len(opts.Allow) == 0 || matchAnyBoundaryPattern(opts.Allow, dep.specifier, name)
if !disallowed && allowed {
    continue
}
message := fmt.Sprintf("External dependency %q is not allowed.", dep.specifier)
```

`opts.Allow` decides the outcome and then goes unmentioned.

`boundaries/no-unknown` says a file "does not match any configured boundary element" while holding `opts.Elements`, each with the `Pattern` that would have matched. The one thing the reader needs is the one thing withheld.

## This repository already decided the other way

Three rules with the same shape name the set. This is the convention the four above miss:

- **`unicorn/import-style`** (`rules_unicorn_import_style.go:756-760`) renders it with a dedicated helper:
  ```go
  ctx.Report(node, fmt.Sprintf(
      "Use %s import for module `%s`.",
      unicornImportStyleDisjunction(module.allowed),
      moduleName,
  ))
  ```
  Its own header comment (`:219`) states the intent outright: *"the diagnostic message lists allowed styles in declaration order."* There is even a helper for turning a set into prose.

- **`unicorn/prevent-abbreviations`** — `"Please rename the %s \`%s\`. Suggested names are: %s."`

- **`unicorn/string-content`** (`rules_unicorn_string_content.go:33`) — `"Prefer \`{{suggest}}\` over \`{{match}}\`."`

So the question is not whether ttsc wants diagnostics to name the acceptable set. It already does, in three places, one with tooling for it. Four rules in one file simply do not.

## Consequence surface

Affects `ttsc lint` CLI output, CI logs, editor diagnostics, and LSP — every consumer, because it is the message text.

Not affected: rule decisions, autofix, code actions. No behavior changes; only what the diagnostic says.

Boundary worth flagging: three of the four honor a user-supplied `Message` override (`:61-62`, `:81-82`). Appending to a message the user wrote would be presumptuous, so the addition belongs on the default path only.

Also worth stating plainly: **`boundaries/no-private` (`:140`) is deliberately excluded.** Its set is what is *forbidden* (`target.Private`), and the acceptable set is its complement — every other file. There is nothing to name. Deny lists cannot be helped this way, and I would rather scope this to the four where the set genuinely exists than pad it.

## Approach

The invariant: **a diagnostic that says "not allowed" should say what is.** The rule already computed it; the only question is whether the message carries it.

Owner is the four report sites in `linthost/rules_boundaries.go`. `unicornImportStyleDisjunction` sets the precedent for rendering; something equivalent for boundary patterns would keep the four consistent with each other and with `import-style`.

Shape, roughly, without prescribing:

- `boundaries/element-types` → name the element types allowed from this source.
- `boundaries/external` → name the allowed dependency patterns; when `opts.Allow` is empty the constraint is deny-only and there is nothing to add, so say nothing.
- `boundaries/entry-point` → name the entry patterns. The word "allowed" already in the message is a promise the message should keep.
- `boundaries/no-unknown` → name the configured element types.

Truncate long sets rather than printing a wall — a message that scrolls is its own failure. `@samchon/evidence` caps such lists at 5-8 with an ellipsis, and that seems about right.

| Outcome | Assertion |
| --- | --- |
| Positive | Each of the four default messages names the acceptable set derived from the configured options |
| Negative twin | A user-supplied `Message` override is emitted verbatim, unchanged |
| Boundary | An empty allow set (deny-only config) adds nothing rather than an empty list |
| Boundary | A large set truncates rather than emitting an unbounded message |
| Regression | `boundaries/no-private` is untouched; rule decisions and fix behavior unchanged |

Narrow: `node scripts/test-go-lint.cjs`, plus whichever `packages/lint/test/rules/` cases cover boundaries. Broader: `pnpm test`.

**Change Integrity:** diagnostic text is specification — existing tests asserting these messages will need updating, and per the development skill that must be called out explicitly by whoever implements rather than done quietly. If a test asserts the current text, the update is the *point* of the change, not an obstacle to route around.

## Coordination

Independent of #727 / #728 / #730.

Related to #730 by theme — that issue is about contributors being unable to *offer* choices, this one about built-ins declining to *name* them — but they share no code and can land in either order.

Note this is deliberately not a completion or architecture request. I went looking for rules where editor completion would beat a better error message across the corpus and found essentially none: the rules that hold a real valid set already surface it, the deny-list rules have an infinite complement with nothing to complete, and config authoring is already covered by the typed `ITtscLintRules` interfaces. This is what survived that search — a message gap in four sites, fixable with a `strings.Join` and no new architecture.

Happy to send a PR if the shape above is what you want.

samchon:
Closes #731. Part of #733. Independent of the other children.

## The problem

Four `boundaries/*` rules compute the set of things that *would* be acceptable, use it to decide, then discard it and report only that the code is wrong. The reader is told "not allowed" and has to open `lint.config` to learn what is.

The sharpest case is `boundaries/entry-point` (`rules_boundaries.go:112`): the message uses the word **"allowed"** and names not one allowed thing, though `target.Entry` — the entry patterns it means — was in hand two lines earlier.

This is out of step with the corpus, not a new opinion. `unicorn/import-style`, `unicorn/prevent-abbreviations`, and `unicorn/string-content` all name the set in the message; `import-style` has a dedicated helper for rendering it. Four rules in one file just do not.

## What changed

| Rule | Names |
| --- | --- |
| `element-types` | `rule.Allow`, when an allow-list denied the import |
| `external` | `opts.Allow`, read one line above to decide |
| `entry-point` | the element's entry patterns |
| `no-unknown` | the configured element types |

Real output:

```
External dependency "@legacy/sdk/client" is not allowed. Allowed here: react, @app/*.
Import "src/domain/internal.ts" through an allowed boundary entry point. Allowed here: index.ts.
```

## What is deliberately left alone

- **A user-supplied `Message` override.** Appending to prose someone wrote would be presumptuous, so the clause is added only on the default path.
- **A deny-only policy.** Its allowed set is the infinite complement — there is nothing to name — so `describeBoundaryAllowed` returns `""` rather than an empty `Allowed here:`. Proven by a negative twin per rule.
- **`no-private`.** Same reason: its set is what is *forbidden*.

Long sets truncate at six; a message that scrolls is its own failure.

## Change Integrity

Diagnostic text is specification. The one existing test asserting a boundary message (`no-unknown`) is **extended** to also assert the named set, not replaced. The other three rules had **no test asserting their message text at all** — a real gap — so this adds a positive case for each with a deny-only negative twin.

## Verification

The six boundary cases pass locally. The lint Go lane needs the runner's `go.work` overlay (it materializes in-tree shims) which needs a node_modules this isolated checkout lacks; I reproduced that overlay by hand to run these, but CI is the authoritative check.


samchon:
CI: the one failure is the `typia` workflow's Ubuntu job, and the evidence says it is inherited from this PR's base, not from this change.

The failure is a typia `TypeGuardError` in `test_http_migrate_v30.ts` — `CreateCompletionRequest.properties.seed.maximum` expected `number & Type<"int64">`. That is OpenAPI schema type inference, and this PR touches only four boundaries-rule **message strings** plus their tests. There is no path from a lint diagnostic's text to typia's migrate output.

What points at the base rather than at me:

- This branch forks from the current master tip `8b5934cbe`, whose HEAD is #735 (`fix(graph): resolve details.literals from the checker's type`) — a change to checker-driven type resolution, which is exactly the area this failure lives in.
- #728 and #736, which pass typia, fork from **older** master (`1c087781e` / `c93435207`, both before #735).
- I can find no typia run on `8b5934cbe` itself, so the base was never independently shown green here.

I did not touch typia, migrate, or any schema code — `git diff origin/master --name-only` is four `rules_boundaries` files. Flagging rather than chasing it: if #735 regressed a typia fixture, that is its own fix on master, and rebasing this PR onto a corrected master will clear it. Happy to rebase once master is green, or to be corrected if the link to this change is real and I am missing it.

samchon:
Correction to my previous comment — I was wrong about the cause, and the corrected evidence is stronger.

I guessed the typia failure came from #735 (this PR's base). It does not. The same `test_http_migrate_v30` failure appears on PRs forked from **different** bases — #742 (base `8b5934c`) and #740 (base `0a8ff86`) both fail typia identically. A regression from one commit cannot fail on branches that do not share it.

So this is a repository-wide pre-existing failure, not from #735 and not from this change. The likely cause is a typia fixture drift (the expected `CreateCompletionRequest.properties.seed.maximum` type), independent of ttsc's own code. Also, #735's diff is entirely `packages/graph` — there is no path from it to typia's OpenAPI migrate output either, so my original suspicion was doubly wrong.

Net: every other check on this PR is green, and the one red is inherited and shared across the repo's open PRs. This change is four boundaries-rule message strings and their tests.

Issue #768, opened 2026-07-18
epic(lint): the rule corpus can say more than 'this is wrong' — a 743-rule survey


samchon:
## What this is

An umbrella for making the built-in rule corpus use the editor channels ttsc now has. It is grounded in an **exhaustive survey of all 743 built-in rules** — every rule read at its `Check` body, never inferred from its name — run as six parallel audits partitioned by family.

**The one-sentence finding:** the corpus does not merely under-use the new channels; a third of it emits nothing but a bare squiggle, and across the rest the dominant pattern is that a rule *computes the answer and then throws it away*.

## The channels this is about

Four landed in the #733 series and are now on master: `codeDescription` (#745), a choice of fixes via `rule.Suggestion` (#744), diagnostic `tags` (#746), and — for project rules — a completion corpus via `rule.HintRule` (#736). `relatedInformation` and `data` are in flight (#750, #749).

## The survey

| Family group | Rules | Something beyond a plain fix | Nothing |
| --- | --- | --- | --- |
| core (no prefix) | 149 | 59 | 90 |
| unicorn | 146 | 17 | 125 |
| typescript + functional + regexp | 140 | ~65 | ~75 |
| jsx-a11y + react + react-perf + solid + nextjs | 112 | 47 | 64 |
| playwright + testing-library + jest + vitest + cypress | 105 | 61 (~30 confirmed) | 44 |
| format + promise + storybook + security + jsdoc + tanstack-query + boundaries | 91 | 31 (~20 confirmed) | 60 |

### Finding 1 — a third of the corpus emits no fix at all

`grep` for any `ReportFix`/`ReportSuggestion` returns **nothing** across the five test-framework families (105 rules) and the five React-adjacent families (112 rules). The whole `regexp` family (21) and the whole `functional` family (20) likewise emit zero edits. That is ~258 rules whose entire output is a squiggle — the premise "we already show a fix preview" does not hold for them.

### Finding 2 — the dominant pattern is *computed, then discarded*

Not "the rule could compute this." The rule **already did**, and drops it:

- `nextjs/no-typos` runs a full Levenshtein (`rules_nextjs.go` ~`:744-762`), binds the correctly-spelled name in `target` (~`:439`), and reports `"Possible typo…"` without it.
- `solid/imports` computes the exact canonical module in `correct` (`rules_solid.go` ~`:258`, table ~`:1023-1033`) and emits a generic message.
[...]

PR #773, opened 2026-07-18
feat(lint): storybook, security and solid rules say what they already know

samchon:
Part of #768.

Nine rules across `storybook/*`, `security/*`, and `solid/*` already knew more than they said. Each one had computed the canonical module, the preferred package list, or the exact replacement name, then dropped it and emitted a generic sentence — and several reported a construct whose whole meaning ("this is dead", "this is deprecated") the editor could have rendered but was never told.

This routes what they already know through the two channels that landed in #746 (`rule.DiagnosticTags`) and #744 (`ctx.ReportSuggestion` / `rule.Suggestion`).

### Diagnostic tags

| Rule | Tag | Why the range makes it sound |
| --- | --- | --- |
| `solid/no-react-deps` | `Unnecessary` | Reports the dependency-array literal alone. Solid tracks dependencies automatically, so the array is inert and deleting exactly the faded range is the resolution. |
| `storybook/no-redundant-story-name` | `Unnecessary` | Both arms — the `name` / `storyName` property and a `Story.storyName = ...` assignment — span exactly the annotation to remove. |
| `storybook/no-title-property-in-meta` | `Unnecessary` | Reports the `title` property itself; CSF3 derives the title from the file path. |
| `storybook/no-stories-of` | `Deprecated` | Reports the `storiesOf` import specifier. The builder still runs, so the instruction is migration, not deletion. |
| `storybook/hierarchy-separator` | `Deprecated` | Reports the `title` property whose separator is superseded. Storybook still reads it and the story still renders. |
| `security/detect-new-buffer` | `Deprecated` | Reports the whole `new Buffer(...)` expression, which is exactly the DEP0005 construct. |
| `security/detect-pseudoRandomBytes` | `Deprecated` | Reports the `crypto.pseudoRandomBytes` member access, which is exactly the DEP0115 alias. It still resolves and still returns bytes. |

### Fixes and suggestions

- **`solid/imports`** — the rule computed the canonical entry point in `solidPreferredSource` and threw it away. The message now names the symbol and the module (``Import `render` from `solid-js/web`.``), and an automatic fix rewrites the module specifier **only when the misplaced specifier is the declaration's sole binding**. A declaration with a second specifier or a default binding keeps its diagnostic: rewriting the specifier would drag the other symbols along, and separating them means splitting the import block rather than editing one token. Only the text between the quotes is replaced, so quote style survives.
- **`solid/no-react-specific-props`** — `className` → `class` and `htmlFor` → `for` are autofixed by rewriting the name token alone, so the value (string, expression container, or shorthand) is untouched. The `key` arm stays diagnostic-only.
- **`storybook/no-renderer-packages`** — `storybookRendererPackages` was a banned → **preferred list** map looked up with `_, ok :=`. The value is now both named in the message and offered as one titled suggestion per framework package (`@storybook/react` → three). None is automatic: which framework is right depends on the project's bundler, which the import does not state.
- **`security/detect-new-buffer`** — three titled suggestions (`Buffer.from` / `Buffer.alloc` / `Buffer.allocUnsafe`), each replacing the `new Buffer` span and leaving the argument list intact. Deliberately not automatic: the rule fires precisely because the argument is not a literal, so it cannot tell which successor applies, and imposing `Buffer.allocUnsafe` on what was really a string recreates the constructor's original uninitialized-memory hazard.
- **`security/detect-pseudoRandomBytes`** — one automatic fix renaming the member to `randomBytes`. It is imposed rather than suggested because there is one successor and it is the same function: the alias is `crypto.randomBytes` on every release that still exposes it, so the edit cannot change behavior, and on the older releases where the two differed it can only strengthen the result. Only the member name is replaced, so the API is repaired whether it is called or passed as a value.

## Rejected candidates

- **`security/detect-buffer-noassert` — `Unnecessary` rejected on range.** `bufferNoAssertIndex` does return the exact argument index, as the audit says, but `Check` calls `ctx.Report(node, …)` with `node` being the whole `CallExpression`. Greying that range would fade a live buffer read, not the dead `noAssert` argument. Narrowing the range to the argument would diverge from the upstream `eslint-plugin-security` rule's reporting location, which this port mirrors, so the tag is withheld rather than the range silently changed. `TestSecurityDeprecatedNodeApiFindingsTagDeprecated` asserts both the absent tag and the whole-call range, so the rejection is checked rather than merely claimed.
- **`solid/no-react-specific-props` — `Unnecessary` rejected on grain.** The audit is right that the `key` arm reports exactly the attribute whose deletion resolves it. But `rule.TaggedRule` is rule-level by construction (`engine.go` reads it once per (file, rule) and stamps every finding), and the same rule's `className` and `htmlFor` arms mean "rename this". One tag cannot say both, and greying a prop that only needs renaming is the exact failure mode the marker's own doc comment warns about. The rule gets the two renames as autofixes instead, and stays untagged.

## Message changes (specification)

Three messages changed. All three widen what the diagnostic states; none narrows it.

- `solid/imports`: `"Import Solid API from its canonical module."` → ``"Import `<symbol>` from `<module>`."``
- `storybook/no-renderer-packages`: `"… Use a framework package instead."` → `"… Use a framework package instead: <disjunction of framework packages>."` (rendered the way upstream's `Intl.ListFormat` disjunction does, so a single-successor renderer such as `@storybook/server` gets no "or")
- `security/detect-pseudoRandomBytes`: appends ``" Use `crypto.randomBytes` instead."``

The rule-corpus fixtures under `tests/test-lint/src/cases` assert rule name and severity through `// expect:` annotations, not message text, so none of them move.

## Tests

Nine new cases, one per file.

Tag cases (`packages/lint/test/rules/**`) each pin the tag **and** the exact reported byte range, since a tag is only sound if the range is, and each carries a negative twin one property away:

- `solid_no_react_deps_tags_dependency_array_unnecessary_test.go` — twin: `solid/no-react-specific-props` reports `className` and `key` untagged.
- `storybook_deletion_findings_tag_unnecessary_test.go` — twin: `storybook/csf-component`, whose finding means "not done yet", reports untagged.
- `storybook_legacy_api_findings_tag_deprecated_test.go` — twin: `storybook/no-renderer-packages`, also an import finding but the wrong layer rather than a deprecated API, reports untagged.
- `security_deprecated_node_api_findings_tag_deprecated_test.go` — twin: `security/detect-buffer-noassert`, asserting both the absent tag and the over-wide range that withholds it.

Fix and suggestion cases (`packages/lint/test/fix/**`) assert the transformation direction — mangled input to canonical output — plus negative twins and boundaries:

- `fix_solid_imports_rewrites_sole_specifier_module_test.go` — rewrites `"solid-js"` → `"solid-js/web"`; a single-quoted specifier keeps its quotes; multi-specifier and default-binding declarations report without a fix; an already-canonical import is silent.
- `fix_solid_no_react_specific_props_renames_react_props_test.go` — renames both props while keeping their values; the `key` arm applies nothing; `class`/`for` and a component-cased tag are silent.
- `fix_storybook_no_renderer_packages_offers_framework_packages_test.go` — three suggestions each rewriting the specifier, zero automatic edits, disjunction rendering for the three- and one-successor cases, and `@storybook/react-vite` silent.
- `fix_security_detect_new_buffer_offers_three_successors_test.go` — three suggestions, zero automatic edits, literal argument and a different constructor silent.
- `fix_security_detect_pseudo_random_bytes_renames_to_random_bytes_test.go` — rewrites both the call and the value-position read; `crypto.randomBytes` and a non-`crypto` object silent.

Expectations come from the upstream rule semantics and the Node deprecation records (DEP0005, DEP0115), not from current output.

## Verification

- `go build ./linthost/` under a temporary `go.work` (`.`, `../ttsc`, every `../ttsc/shim/**` module): clean. The `go.work` was deleted and is not committed.
- The full lint Go lane, replicating `scripts/test-go-lint.cjs` (scratch overlay, flattened test files, workspace `go.work`, `go test -count=1 ./linthost`): **PASS**. This worktree has no `node_modules`, so `ttsx`/`tsgo` were resolved from the main checkout via `TTSC_TTSX_BINARY` / `TTSC_TSGO_BINARY`, exactly the two variables the runner itself honours.
- Formatting: `.vscode/gofmt-2spaces.sh -w` over the touched Go files only — no repository-wide `pnpm format`, per the active issue campaign.
- Not run locally: the TypeScript e2e suites (`pnpm test:features`), which need a built workspace this worktree lacks. CI is authoritative for those.

## Change integrity

`rule.TaggedRule` and the suggestion surface are used, not modified; no public signature changes. Three diagnostic messages change, listed above. The rule-typing doc comments (`ITtscLint{Solid,Storybook,Security}Rules.ts`), the three family pages, the autofix-coverage list on the rules index, and the tag paragraph in the lint walkthrough are updated in the same change.

https://claude.ai/code/session_01PbiF9VH8TAGwdgFBnEQYfz


samchon:
Self-review found two Deprecated-tag ranges wider than the deprecated token: an aliased \\storiesOf\\ import included its live local alias, and a piped title included the live property/value. I will narrow both ranges and add alias/escaped-separator coverage before repeating the full review.

samchon:
Second self-review refinement: the new exact-range helper deliberately handles active \\\\\\|\\, \\\\\\x7C\\, fixed Unicode, and braced Unicode spellings. The first correction tested only fixed Unicode, so I will cover every supported escape form before the final review.

samchon:
Final self-review found that a 
ew /* comment */ Buffer(...) suggestion replaced the entire 
ew-through-callee span and discarded the comment. I will preserve intervening comments, remove only the 
ew token when needed, and add that regression before restarting review and validation.

samchon:
Final review also found two untested positive branches in the new editor fixes: a sole aliased Solid import resolves through ImportSpecifier.PropertyName, and a bare JSX boolean prop has no value token to preserve. I will add both before the last validation pass.

samchon:
Solo self-review of the local lint run caught a test expectation defect: each 
ew Buffer suggestion intentionally carries two non-overlapping edits (remove 
ew, rename Buffer), so the uncommented assertion must expect two applied edits just like the comment-preservation case. The emitted source was already correct. I have corrected that test expectation and will rebase onto the current master before rerunning the full lint suite.

samchon:
Solo Self-Review finding: the newly added website prose uses em dashes, which the repository documentation rules prohibit. I will replace those sentence boundaries with direct punctuation only, then commit, re-run the full lint Go suite, and restart the complete review on the new snapshot.

samchon:
Second solo-review finding: the updated walkthrough line still carries pre-existing em dashes because the whole paragraph is one Markdown source line. To make the final diff comply with the documentation rule, I will rewrite that touched paragraph with colons and sentences, then create a final snapshot and restart the review once more.

samchon:
Third solo-review validation finding: targeted Prettier --check reports formatting drift in the five MDX files changed by this PR. I will run Prettier only on those five declared documentation files, never the repository-wide pnpm format command, then commit the mechanical result and restart the full review on that snapshot.

samchon:
Final solo Self-Review on 6175309e is clean. I re-read the complete 20-file base-to-head surface after the documentation corrections, including diagnostic range/tag semantics, automatic versus suggestion-only edits, escaped Storybook separators, negative and boundary tests, public rule docs, packaging, and the Windows/POSIX-independent byte-edit paths. Local validation passed: `node scripts/test-go-lint.cjs` (`packages/lint/linthost`, 480.399s); targeted Prettier --check passes for all five changed MDX files. `git diff --check` is clean. Ordinary PR CI is the remaining merge gate.

samchon:
After rebasing onto master 53901bc4, the final head is a6cbe0c5f412e00f95f425e7febfb278a0fbefd5. A fresh complete solo Self-Review of the rebased 20-file surface is clean, `git diff --check` is clean, the protected graph file hash/tree guards match, and targeted Prettier --check passes on all five changed MDX files. Rebased local validation passed: `node scripts/test-go-lint.cjs` (`packages/lint/linthost`, 313.936s). The ordinary CI for this final SHA is the remaining merge gate.

samchon:
Fresh solo Self-Review completed on c1046b705b06a9aa94bde6b2cc144a736ca62b78 after rebasing onto efeebd2e22440b13a1338bf5398ba12e0d46d36 (clean review round, no further findings). Re-read the whole 20-file diff: linthost behavior, public rule structures, all nine regression tests, and five MDX pages. 
ode scripts/test-go-lint.cjs passed (linthost 299.202s); targeted Prettier check and git diff --check passed. The protected Graph SHA-256 and structures-tree guards are unchanged.

samchon:
Fresh solo Self-Review of the rebased head found two semantic-safety gaps before the final CI push.

1. The `Unnecessary` Storybook property findings currently stop before a trailing comma. Deleting exactly the faded `title` / redundant `name` range from a multi-property object therefore leaves a leading empty element and invalid TypeScript. I will make the reported deletion grain include the property's trailing comma when present, and add tests that actually remove each tagged range and parse/lint the remaining source, including shorthand and last-property boundaries.
2. `security/detect-pseudoRandomBytes` currently imposes `randomBytes` on every object whose identifier text is `crypto`, including a local application object. I will use the existing security binding table: an import/require proven to be Node `crypto` or `node:crypto` may receive the automatic rename; an unproven/shadowed `crypto` keeps the diagnostic and gets the opt-in suggestion only. Tests will cover proven ESM/CJS bindings, the existing unbound corpus shape, and a local-object negative safety case.

After these corrections I will rerun the full lint Go lane, targeted formatting checks, diff guards, and a fresh complete solo review before updating the branch.

samchon:
The same rule-level-tag audit extends the second finding to two adjacent cases before any source edit:

- `security/detect-new-buffer` is also purely name-based, so a user-defined `Buffer` constructor cannot soundly inherit Node's DEP0005 tag. Both security rules will keep their existing diagnostics and editor replacements, but the rule-level `Deprecated` tags will be removed rather than stamped onto false-positive-compatible findings.
- `solid/no-react-deps` currently accepts a raw local function named `createEffect` whenever the file imports any Solid symbol, and `storybook/no-stories-of` accepts that imported name from any module. I will bind both tag-bearing checks to their actual library provenance (named/namespace `solid-js` imports and the established Storybook renderer module list), with shadow/custom-module negative tests. This preserves intended findings while making every tagged finding support the classification.

samchon:
The fresh full review on the rebased snapshot found one integration regression in my provenance tightening: replacing the existing `solid-js` prefix test with the three canonical entry points would stop every Solid rule for legitimate package subpaths such as `solid-js/universal`. I will restore the complete package boundary as `source == "solid-js" || strings.HasPrefix(source, "solid-js/")`, use that boundary for state/provenance, and keep the `solid-js-testing` negative twin so similarly named packages remain excluded. The scoped full lint run that started on the prior snapshot will be treated as superseded, then rerun after this correction.

samchon:
Final Self-Review found one mechanical cleanup item before the last validation pass: targeted Prettier check currently reports all eight changed TypeScript/MDX files as nonconforming. I will format only those changed files (not run repository-wide formatting), inspect the resulting diff, and restart the final whole-change review.

samchon:
Resumption Self-Review found one more rule-grain boundary in the live worktree before committing the next snapshot: `storybook/no-redundant-story-name` carries a rule-wide `Unnecessary` tag, but an assignment nested inside another expression (for example, `const derived = Primary.storyName = Primary`) still contributes its value and cannot truthfully be marked as wholly removable. I will keep the standalone expression-statement arm, skip the nested assignment, add the negative regression, align the public/docs wording, then format only these touched files and restart the final whole-change review after the pending website prerequisite lands.

samchon:
Self-Review follow-up before correction: the automatic `crypto.pseudoRandomBytes` rename currently proves only that this file has a Node `crypto` binding with the same text. The shared security binding table is file-wide, so `import * as crypto from node:crypto; function use(crypto: LocalCrypto) { return crypto.pseudoRandomBytes(); }` is still auto-fixed even though the property access resolves to the shadowing parameter. That violates this PR's promise that only a proven Node binding is imposed. I am correcting the automatic-fix gate to compare TypeScript checker binding identity, adding the shadowed-parameter regression, and keeping the same rename as an opt-in suggestion when exact provenance is absent. The final validation/review snapshot will be restarted after this change and the #900-gated latest-master rebase.

samchon:
Fresh whole-diff Self-Review follow-up before correction on the latest-master snapshot: `solid/no-react-deps` still proves a Solid call through file-wide text maps. If a file imports `createEffect` and a function parameter shadows that name, the parameter call is still reported with the rule-level `Unnecessary` tag even though its dependency array belongs to an application helper and may be live. The namespace form has the same risk. I am making only this tagged rule checker-aware, recording the exact symbols of named and namespace Solid imports, and requiring each call-site symbol to match before reporting. Shadowed named and namespace regressions will stay silent. Because this changes the reviewed snapshot, final full validation and the solo review restart after the correction.

samchon:
Self-Review correction before edit: a raw gofmt invocation used during the resumed focused validation normalized the two touched Go files to tabs, while this repository's canonical formatter is .vscode/gofmt-2spaces.sh. I will immediately re-run the repository formatter on those exact files and amend the local, unpushed correction commit so the final diff retains only the intended binding-identity changes and regression cases.

samchon:
Self-Review correction before edit: after applying the rule-wide `Unnecessary` tag, the standalone `Story.storyName = ...` arm still accepted any matching local identifier. That could fade a private `Local.storyName = Local` assignment as safe to delete even though `Local` is not a named story export. I will gate that arm on the file's actual named-export set and add a silent private-local regression case before restarting validation and whole-diff review on the new head.

samchon:
Self-Review follow-up before edit: the named-export name gate must also require the legacy `Story.storyName = ...` annotation to be a top-level expression statement. Otherwise a same-named parameter inside a function can shadow an exported story and still inherit the rule-wide `Unnecessary` tag. I will constrain this legacy annotation path to a direct SourceFile statement and add the shadowed nested-scope case to the regression coverage.

samchon:
## 작업 중단 체크포인트 인수인계

사용자 지시에 따라 추가 개발·전체 검증·CI 대기·병합을 즉시 중단하고, 현재 완료 변경을 체크포인트로 정리해 푸시했습니다.

### 체크포인트

- PR HEAD: `a39693771ee98a5148ae7deb16031a323bd98b47`
- base / 최신 확인 `origin/master`: `43539ab39325c41a3504acc6244131aea7d3ce1e`
- `merge-base(HEAD, origin/master)`도 `43539ab39325c41a3504acc6244131aea7d3ce1e`로, 체크포인트는 확인 시점 최신 master에 재베이스되어 있습니다.
- 이전 원격 topic HEAD `3ad7fb757ef1be293b5fed4e3d66a405c13e3a66`를 먼저 재확인한 뒤 다음 explicit lease로 갱신했습니다.
  - `git push --force-with-lease=refs/heads/feat/lint-storybook-security-solid-editor-channels:3ad7fb757ef1be293b5fed4e3d66a405c13e3a66 origin HEAD:refs/heads/feat/lint-storybook-security-solid-editor-channels`
- 작업 트리는 clean입니다.

### 완료된 변경

- Security
  - `new Buffer(nonLiteral)`에 `Buffer.from` / `Buffer.alloc` / `Buffer.allocUnsafe` 3개를 suggestion-only로 제공하고 주석을 보존합니다.
  - `crypto.pseudoRandomBytes`의 member rename을 제공합니다. 정확한 Node `crypto` / `node:crypto` 바인딩일 때만 자동 수정하고, 로컬 객체나 shadowed binding은 suggestion-only입니다.
  - 이름 기반 Node API 규칙은 사용자 정의 객체를 오인할 수 있으므로 Deprecated tag를 붙이지 않습니다.
- Solid
  - `solid/imports`가 정확한 canonical module을 메시지에 쓰고, misplaced specifier가 단독 binding일 때만 module specifier를 자동 수정합니다.
  - `className`→`class`, `htmlFor`→`for`를 token-only autofix로 제공합니다. `key`는 진단-only입니다.
  - `solid/no-react-deps`는 정확한 Solid import binding의 dependency array만 `Unnecessary`로 표시하며, named/namespace shadow를 제외합니다.
- Storybook
  - `storiesOf`와 hierarchy separator를 정확한 token/escape 범위에 `Deprecated`로 표시합니다.
  - renderer package 대체 후보를 suggestion-only로 제공합니다.
  - `title` 및 redundant story annotation의 완전 삭제 가능 범위를 `Unnecessary`로 표시합니다.
  - 마지막 Self-Review 수정으로 `Story.storyName = ...`는 실제 named export의 top-level standalone statement일 때만 삭제 가능으로 표시합니다. private local, value-producing assignment, nested-scope shadow는 제외합니다.
- 위 동작의 Go 회귀 테스트와 TypeScript/MDX 설명을 포함합니다.

### 통과한 검증

최신 master 재베이스 이후의 체크포인트 계열에서 다음이 통과했습니다.

- 변경된 Go 파일: 저장소 정식 `./.vscode/gofmt-2spaces.sh` 적용
- 변경된 TS/MDX 파일: targeted Prettier 적용, 추가 diff 없음
- `git diff --check`
- `pnpm --filter @ttsc/lint build`
- Crypto + Solid binding 집중 회귀: PASS (`3.977s`, 당시 HEAD `c808ffde2`)
- 마지막 Storybook 삭제 범위/비공개·shadow 제외 집중 회귀: PASS (`2.944s`, 최종 HEAD `a39693771`)

### 미해결 검증과 정확한 실패

최종 전체 Go 검증은 통과하지 않았습니다. `c808ffde2`에서 실행한 전체 lint Go 테스트가 다음 1건으로 실패했습니다.

```text
--- FAIL: TestSolidImportProxyAndDependencyRules
solid_import_proxy_and_dependency_rules_test.go:28:
want [... {solid/no-react-deps error 7} ...],
got  [... 해당 finding 없음 ...]
```

정황상 `solid/no-react-deps`가 정확한 binding 판별을 위해 `NeedsTypeChecker()`를 요청하게 된 반면, 기존 `packages/lint/test/rules/solid/solid_helpers_test.go`의 `assertSolidFindings`는 `NewEngine(...).Run(files, nil)`로 checker 없이 실행하는 것이 원인 후보입니다. 실제 checker-backed 신규 집중 테스트는 통과했습니다. 안전성 때문에 구현을 이름 기반 fallback으로 되돌리면 shadow 오탐을 재도입하므로, 재개 시 테스트 harness를 engine 요구사항에 따라 checker-backed 경로로 올리는 방향부터 검토해야 합니다.

또한 최종 `a39693771` 전체 Go 재실행과 이 immutable SHA 기준 20파일 전체 Self-Review 완료는 중단 지시로 수행하지 않았습니다.

### CI 상태

- 과거 PR red는 이전 원격 SHA `3ad7fb757...`의 결과라 현재 체크포인트의 판정이 아닙니다.
- 새 HEAD `a39693771...` 푸시로 새 ordinary CI가 생성되었습니다.
- 댓글 작성 시점: Socket 검사 2개 PASS, 나머지는 pending입니다.
- CI를 취소하거나 기다리지 않았고, 병합도 시도하지 않았습니다.

### 재개 순서

1. 아래 집중 명령으로 기존 Solid 복합 테스트 실패를 재현합니다.
2. checker 요구사항을 보존한 채 `assertSolidFindings` 테스트 harness를 principled하게 보완하고, 사전 PR 댓글→수정→집중 검증→커밋 순서를 지킵니다.
3. 변경 20파일에 저장소 formatter/Prettier, lint build, 전체 lint Go를 실행합니다.
4. 최종 immutable SHA에서 20파일 whole-diff solo Self-Review를 처음부터 다시 수행합니다.
5. 그 사이 master가 전진했으면 재베이스하고 위 검증/리뷰를 새 SHA에서 반복합니다.
6. explicit force-with-lease로 푸시하고 ordinary CI 전부를 terminal까지 확인합니다. red는 해당 job log 기준으로 수정합니다.
7. PR 본문의 오래된 Security tag 설명도 실제 동작에 맞게 갱신합니다.
8. 전부 green 후 PR을 병합하고 umbrella issue #768은 계속 OPEN인지 확인합니다.
9. 병합 완료 후에만 이 worktree와 아래 전용 Go/cache/log 자산을 제거하고 `git worktree prune`을 수행합니다.

### 재개 경로와 명령

```powershell
Set-Location 'D:\github\samchon\ttsc@pr-773'
git status --short
git fetch origin --prune
git rev-parse HEAD
git rev-parse origin/master

$env:GOCACHE='D:\github\samchon\ttsc@pr-773\.campaign-tmp\go-cache'
$env:GOTMPDIR='D:\github\samchon\ttsc@pr-773\.campaign-tmp\go-tmp'
$env:GOFLAGS='-run=^TestSolidImportProxyAndDependencyRules$'
node scripts/test-go-lint.cjs
Remove-Item Env:GOFLAGS -ErrorAction SilentlyContinue

pnpm --filter @ttsc/lint build
node scripts/test-go-lint.cjs
gh pr checks 773
```

### worktree / temp 상태

- 보존 worktree: `D:\github\samchon\ttsc@pr-773`
- branch: `feat/lint-storybook-security-solid-editor-channels`
- 활성 test/Go 프로세스: 없음
- 보존 전용 자산: `D:\github\samchon\ttsc@pr-773\.campaign-tmp`
  - `go-cache`, `go-tmp`, 전체 검증 stdout/stderr 로그
  - 현재 약 5,263,748,708 bytes / 4,455 files
- 미완료 체크포인트이므로 재개 효율을 위해 보존했습니다. 병합 완료 시 worktree 제거와 함께 반드시 정리해야 합니다.
- 다른 worktree와 전역 stash는 건드리지 않았습니다.

samchon:
AutoBE 벤치마크 캠페인의 통합 인수인계는 https://github.com/wrtnlabs/autobe-mcp/issues/609 에 정리했습니다. 재개 시 이 PR의 상세 인수인계 댓글과 #609의 순서를 함께 따르십시오. 현재 head는 a39693771ee98a5148ae7deb16031a323bd98b47이며 병합하지 않았습니다.

samchon:
## Superseded by #914

The maintainer directed that the leftover open pull requests be consolidated into one, so this branch's commits were merged into `fix/campaign-2026-07-21` with their history and authorship intact and are validated there as part of a single CI run.

Nothing from this branch is dropped. The branch itself is preserved for now; it will be deleted only after #914 merges.

Closing here to stop this pull request from consuming a parallel CI run.

Issue #759, opened 2026-07-18
fix(graph): rank all ambiguous symbol candidates before applying the candidate limit


ailectra:
## Summary

Ambiguous symbol candidates are truncated before they are ranked.

The resolver first applies `.slice(0, candidateLimit)` in several resolution branches. The later `rank()` function can therefore rank only the arbitrary prefix that survived the first slice. A highly relevant candidate outside that prefix can never appear in the result.

Relevant code:

- [`resolveGraphName` slices exact and suffix matches](https://github.com/samchon/ttsc/blob/master/packages/graph/src/server/resolveHandle.ts#L93-L105)
- [`resolveFileQualified` slices matches](https://github.com/samchon/ttsc/blob/master/packages/graph/src/server/resolveHandle.ts#L127-L132)
- [`rank` sorts only the already-truncated list](https://github.com/samchon/ttsc/blob/master/packages/graph/src/server/resolveHandle.ts#L150-L159)

## Minimal reproduction

Construct a graph with 13 nodes that all resolve from the same name:

- candidates 1-12 have low `candidateScore` values;
- candidate 13 is exported or otherwise has the highest score;
- the candidate limit is 12.

Resolve the shared name.

A resolver-level regression test can use a synthetic `TtscGraphMemory`/fixture where the highest-scoring node is deliberately placed after the first 12 entries.

## Actual behavior

The first 12 nodes are sliced from the graph's traversal/index order. Those 12 are then ranked.

The 13th, highest-scoring node is absent from the final candidates because it was removed before ranking.

## Expected behavior

All matching candidates should be ranked first, and only the ranked result should be limited:

```ts
const ranked = [...allMatches]
  .sort((a, b) => candidateScore(graph, b) - candidateScore(graph, a))
  .slice(0, candidateLimit);
```

The highest-scoring candidate must be returned regardless of its original graph order.

## Why this matters

The resolver documentation says candidates are ordered by how likely the caller means them. Pre-ranking truncation makes the result depend on declaration visitation/index order instead.

This can direct an agent to the wrong declaration even when the scoring logic correctly identifies a better candidate.

## Suggested fix

Remove the early `.slice(0, candidateLimit)` calls from:

- exact-name matches;
- qualified-name suffix matches;
- file-qualified matches;
- any other branch that later passes candidates through `rank()`.

Apply the limit once, after scoring and sorting.

If sorting a very large candidate set is a concern, use a bounded top-k selection algorithm, but it must consider every match.

## Regression tests

[...]

PR #778, opened 2026-07-19
fix(graph): preserve filenames in legacy absolute reroots

### LINKED ISSUE #761: fix(graph viewer): preserve the filename when rerooting a single-file legacy absolute-path dump
## Summary

The viewer's legacy absolute-path rerooting can remove the entire filename when every project node belongs to one source file.

`commonRoot()` compares complete file paths. With a single distinct path, the common root becomes the full path including the filename. `relativize()` then removes that complete path and returns an empty string.

Relevant code:

- [`commonRoot` and `relativize`](https://github.com/samchon/ttsc/blob/master/packages/graph/src/reduce.ts#L63-L93)
- [Legacy absolute-path root selection](https://github.com/samchon/ttsc/blob/master/packages/graph/src/reduce.ts#L138-L172)

## Minimal reproduction

Reduce a legacy viewer payload with two connected nodes from the same absolute file:

```ts
const raw = {
  project: "/work",
  nodes: [
    {
      id: "/work/src/only.ts#a:function",
      kind: "function",
      name: "a",
      file: "/work/src/only.ts",
      external: false,
    },
    {
      id: "/work/src/only.ts#b:function",
      kind: "function",
      name: "b",
      file: "/work/src/only.ts",
      external: false,
    },
  ],
  edges: [
    {
      from: "/work/src/only.ts#a:function",
      to: "/work/src/only.ts#b:function",
      kind: "calls",
    },
  ],
};
```

Run the normal viewer reduction function.

## Actual behavior

The root is calculated as:

```text
/work/src/only.ts
```

Relativizing the same path against that root produces:

```text
""
```

Node display IDs can consequently degrade from:

```text
/work/src/only.ts#a:function
```

to:

```text
#a:function
```

and the displayed `file` field is empty.

## Expected behavior

The filename must remain visible. Acceptable results include:

```text
only.ts
```

or:

```text
src/only.ts
```

depending on the intended legacy rerooting policy.

## Root cause

`commonRoot()` is operating on file paths rather than their parent directories. A path is always its own full common prefix when it is the only unique file.

The same issue can occur when multiple nodes share one source file, because the input array contains repeated copies of the same complete path.

## Suggested fix

Calculate the common root from directories:

```ts
const projectDirectories = projectFiles.map(dirname);
const root = commonRoot(projectDirectories);
```

Alternatively, ensure the computed root can never consume the basename of every input file.

The implementation should handle both POSIX and Windows drive paths.

## Regression tests

Add tests for:

1. One node in one absolute file.
2. Multiple nodes in one absolute file.
3. Multiple files in one directory.
4. Multiple files in nested directories.
5. Windows paths such as `C:\work\src\only.ts`.
6. Current project-relative dumps, which should remain unchanged.

## Acceptance criteria

- A single-file absolute-path dump never produces an empty `file`.
- The source basename remains present in node IDs/display fields.
- Multi-file rerooting behavior is preserved.
- Relative current-format dumps are not rerooted.

samchon:
## Intent

Preserve source filenames when graph viewers reroot legacy absolute-path dumps, without changing current project-relative dump behavior.

Closes #761.

## Scope

- Correct legacy absolute-path projection in the package, website, and benchmark-fixture reducers.
- Cover single, repeated, nested, POSIX, Windows drive, and UNC paths while preserving each reducer's ignored-node policy.
- Keep current relative dumps unchanged.
- Keep ITtscGraphApplication and TtscGraphApplication untouched.

## Verification

- Full Graph end-to-end suite: 37/37.
- @ttsc/graph build and viewer bundle.
- graph, test-graph, and website TypeScript checks.
- git diff check and protected-file diff zero.

No paid AI/model benchmark was run or used as a merge gate, by owner direction.

### DIFF
--- a/experimental/benchmark/graph/viewer.mjs
+++ b/experimental/benchmark/graph/viewer.mjs
@@ -29,14 +29,22 @@ const PUBLIC_GRAPH_DIR = path.join(REPO_ROOT, "website", "public", "graph");
 // Pure transform
 // ---------------------------------------------------------------------------
 
-/** Longest shared directory prefix of POSIX-normalized paths. */
-function commonRoot(files) {
-  if (files.length === 0) return "";
-  let parts = posix(files[0]).split("/");
-  for (const file of files.slice(1)) {
-    const other = posix(file).split("/");
+/** Longest shared prefix of POSIX-normalized directories. */
+function commonRoot(directories) {
+  if (directories.length === 0) return "";
+  let parts = posix(directories[0]).split("/");
+  const caseInsensitive = directories.every(isWindowsPath);
+  for (const directory of directories.slice(1)) {
+    const other = posix(directory).split("/");
     let i = 0;
-    while (i < parts.length && i < other.length && parts[i] === other[i]) i++;
+    while (
+      i < parts.length &&
+      i < other.length &&
+      (caseInsensitive
+        ? parts[i].toLowerCase() === other[i].toLowerCase()
+        : parts[i] === other[i])
+    )
+      i++;
     parts = parts.slice(0, i);
     if (parts.length === 0) break;
   }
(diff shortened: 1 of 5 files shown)


samchon:
## Self-Review Round 1

Reviewed the complete `865f982d..284bcd1d` change from scratch across all five changed files, the three reducer consumers, the graph feature harness, viewer documentation, Windows and POSIX path semantics, current relative compatibility, packaging, and the protected MCP API surface.

### Findings

1. The implementation handles UNC paths, but the three nearby absolute-path comments still name only POSIX and Windows drive paths. The comments should state the actual accepted contract.
2. The regression covers a source file below a POSIX root and mixed-case Windows directories, but does not pin the two directory-root branches introduced by `directoryOf`: `/only.ts` and `C:\\only.ts`.
3. Prettier wrapped the test's verification headline across two JSDoc lines. The development contract requires a one-line `Verifies` headline, so it should be shortened.

### Remediation plan

- Name UNC support in all three mirrored reducer comments.
- Add root-level POSIX and drive-root cases to the existing path matrix.
- Shorten the test headline, then rerun the direct regression, graph/test/website typechecks, explicit Prettier check, protected-file diff, and `git diff --check`.

samchon:
## Self-Review Round 2

Fresh complete review of `865f982d..28972ee1` found no sound improvement.

I reread every changed hunk and traced the package, website, and benchmark reducer copies through root selection, file/id rewriting, edge endpoint stability, orphan pruning, cap behavior, external and ignored policies, and already-reduced payload handling. The matrix now covers one self-connected source, repeated nodes in one file, same-directory and nested files, POSIX root and case sensitivity, Windows drive root and mixed casing, UNC casing, and current relative dumps. Documentation matches the behavior, explicit Prettier check is clean, the worktree has no tracked residue, and `ITtscGraphApplication` plus `TtscGraphApplication` remain byte-unchanged from the base.

No remediation is planned for this round. The remaining gate is the deterministic offline 16-cell comparison after issue #759 releases the exclusive VS Code lane.

PR #779, opened 2026-07-19
fix(graph): keep detail members on the compiler snapshot

### LINKED ISSUE #757: fix(graph): object literal member extraction is corrupted by braces in block comments
## Summary

The `details` implementation reconstructs members of object-literal variables by scanning source lines with regular expressions and a brace-depth counter.

The scanner strips strings and `//` comments, but it does not strip block comments. A `{` or `}` inside a block comment therefore changes the calculated nesting depth and can cause valid top-level object members to be omitted.

Relevant code:

- [`objectLiteralMembers`](https://github.com/samchon/ttsc/blob/master/packages/graph/src/server/runDetails.ts#L259-L301)
- [`objectMemberOf` and `stripStrings`](https://github.com/samchon/ttsc/blob/master/packages/graph/src/server/runDetails.ts#L303-L337)

## Minimal reproduction

Create a TypeScript source file containing:

```ts
export const control = {
  real: 1,
};

export const broken = {
  /* { */
  real: 1,
};
```

Build/synchronize the graph and request `details` for both variables.

## Actual behavior

`control` reports `real` as an object member.

`broken` can omit `real` or return no `members` field because the `{` in `/* { */` increments the scanner's depth. The `real` line is then treated as nested rather than as a top-level object member.

A closing brace in a block comment can cause the inverse depth error.

## Expected behavior

Comments must not affect the object-literal structure. Both variables should report the same member outline:

```json
{
  "members": [
    {
      "name": "real",
      "kind": "property"
    }
  ]
}
```

## Why this matters

The package presents graph results as compiler-resolved facts. In this path, however, members are inferred from a line-oriented source heuristic. Valid TypeScript syntax can therefore silently remove members from an otherwise successful `details` result.

The same approach is also vulnerable to other lexical constructs that can contain braces or span lines, including multiline comments and some regular-expression/template-literal cases.

## Suggested fix

Preferred approach:

1. Emit object-literal members from the compiler-backed graph builder, or
2. Parse the declaration with the TypeScript AST and enumerate the `ObjectLiteralExpression` properties.

If the source scanner remains, it should use a proper lexical state machine that handles:

- block comments across lines;
- line comments;
- quoted strings;
- template literals and substitutions;
- regex literals where applicable.

The result audit should also distinguish compiler-produced facts from source-derived heuristics.

## Regression tests

Please add tests covering at least:

```ts
const a = { /* { */ real: 1 };
const b = { /* } */ real: 1 };
const c = {
  /*
   * {
   */
  real: 1,
};
const d = {
  text: "{",
  real: 1,
};
```

## Acceptance criteria

- Braces inside comments or strings do not alter object-member extraction.
- All top-level members in the examples above are returned.
- Nested object members are not incorrectly promoted to the outer object.
- The implementation no longer relies on regex-only parsing for compiler-claimed facts, or the audit wording clearly reflects the remaining limitation.

### LINKED ISSUE #764: perf(graph): cache source file lines during details processing
## Summary

The helper comment says it reads a file's lines "once," but `fileLines()` performs a synchronous `readFileSync` on every invocation.

Relevant code:

- [`fileLines`](https://github.com/samchon/ttsc/blob/master/packages/graph/src/server/runDetails.ts#L519-L527)
- [`details` calls signature/doc/member helpers repeatedly](https://github.com/samchon/ttsc/blob/master/packages/graph/src/server/runDetails.ts#L178-L258)

A single `details` request can read the same source file multiple times while producing:

- the node signature;
- the node documentation;
- object-literal members;
- each contained member's signature;
- details for multiple requested nodes in the same file.

## Actual behavior

Each helper invocation calls:

```ts
fs.readFileSync(path.join(project, file), "utf8").split(/\r?\n/);
```

There is no request-level or snapshot-level cache.

For a class with many members, or a request containing several symbols from one file, this can amplify synchronous filesystem I/O on the MCP server's main thread.

## Expected behavior

Within one synchronized graph/details operation, each source file should normally be read at most once.

Repeated consumers should reuse the same immutable line array.

## Suggested fix

The safest option is a request-scoped cache:

```ts
const fileCache = new Map<string, string[] | undefined>();
```

Pass a source reader/cache object to:

- `signatureOf`;
- `docOf`;
- `objectLiteralMembers`;
- `members`.

Cache read failures as `undefined` as well, so a missing/unreadable file is not retried for every member.

A snapshot-scoped cache could provide more reuse, but it must be invalidated whenever the graph synchronizes to a new on-disk snapshot. Request-scoped caching avoids stale-data complexity.

## Suggested regression/performance test

Spy on or wrap `fs.readFileSync`, then request details for:

- a class and several members in one file;
- multiple handles from the same file;
- an object-literal variable in the same file.

Assert that the file is read once per details request.

Also verify that a later request after a file edit sees the new contents.

## Benefits

- less blocking filesystem I/O;
- predictable behavior for large classes and multi-handle requests;
- implementation matches the helper's "once" comment;
- a single source snapshot is used consistently throughout one response.

## Acceptance criteria

- The same file is not synchronously read repeatedly within one `details` request.
- Cache lifetime cannot return source text from an older synchronized snapshot.
- Read failures are handled consistently.
- Signatures, docs, and members remain unchanged.

samchon:
## Intent

Keep object-literal detail members on the immutable compiler snapshot and remove request-time source reconstruction.

Closes #757.
Closes #764.

## Implemented

- Native graph schema v5 carries direct, statically named object-literal member identity, bounded signatures, evidence, and documentation.
- Spread and dynamic computed members are omitted instead of guessed.
- Details reuse snapshot-owned source text and fail closed when snapshot identity or digest no longer matches.
- Legal duplicate variable declarations cannot mix the first node with a later initializer.
- Signature compaction preserves whitespace inside strings, regular expressions, and multiline templates.
- The lifecycle test fixture emits schema v5.
- ITtscGraphApplication and TtscGraphApplication remain untouched.

## Verification

- Native graph and ttscgraph Go suites: pass.
- Focused duplicate-declaration, literal-whitespace/template, and snapshot-member regressions: pass.
- Go vet: pass.
- Full Graph end-to-end suite: 39/39.
- @ttsc/graph build and viewer bundle: pass.
- graph and test-graph TypeScript checks: pass.
- Changed-file Prettier, git diff check, and protected-file diff zero: pass.

Website-wide typecheck was unavailable in this isolated worktree because its Nextra/React dependencies are not linked; the only website change is formatted MDX prose. No paid AI/model benchmark was run or used as a merge gate, by owner direction.

### DIFF
--- a/packages/graph/src/model/TtscGraphMemory.ts
+++ b/packages/graph/src/model/TtscGraphMemory.ts
@@ -4,16 +4,18 @@ import { ITtscGraphEvidence } from "../structures/ITtscGraphEvidence";
 import { ITtscGraphNode } from "../structures/ITtscGraphNode";
 import { ITtscGraphSpan } from "../structures/ITtscGraphSpan";
 import { TtscGraphEdgeKind } from "../structures/TtscGraphEdgeKind";
+import { TtscGraphSourceReader } from "./TtscGraphSourceReader";
 
 /**
  * The in-memory resident graph the MCP tools answer from.
  *
  * It loads one `ttscgraph dump` — the checker-resolved fact graph — then
  * synthesizes the structural relationships the dump deliberately leaves to this
- * layer: `file` container nodes, the `contains` ownership tree, and `exports`
- * edges, plus class/interface member implementation edges and the refinement of
- * a class-member `variable` to a `property`. Every tool call is then a lookup
- * or a traversal over the indexes built here; nothing recompiles.
+ * layer: `file` container nodes and the `contains` ownership tree, plus the
+ * refinement of a class-member `variable` to a `property`. Export and member
+ * implementation relationships are checker facts already present in the dump.
+ * Every tool call is then a lookup or traversal over the indexes built here;
+ * nothing recompiles.
  */
 export class TtscGraphMemory {
   private readonly byId: Map<string, ITtscGraphNode>;
@@ -26,17 +28,21 @@ export class TtscGraphMemory {
   readonly project: string;
   /** Every node, raw plus synthesized (file containers). */
   readonly nodes: readonly ITtscGraphNode[];
-  /** Every edge, raw plus synthesized (contains, exports). */
+  /** Every edge, raw plus synthesized containment. */
   readonly edges: readonly ITtscGraphEdge[];
+  /** Provenance-gated source display facts cached for this exact snapshot. */
+  readonly source: TtscGraphSourceReader;
 
   private constructor(
     project: string,
     nodes: ITtscGraphNode[],
     edges: ITtscGraphEdge[],
+    provenance: ITtscGraphDump.IProvenance,
   ) {
     this.project = project;
     this.nodes = nodes;
     this.edges = edges;
+    this.source = new TtscGraphSourceReader(project, provenance);
 
     this.byId = new Map(nodes.map((n) => [n.id, n]));
     this.byNameIndex = new Map();
@@ -63,7 +69,7 @@ export class TtscGraphMemory {
   /** Build a model from a parsed dump, synthesizing structural relationships. */
   static from(dump: ITtscGraphDump): TtscGraphMemory {
     const { nodes, edges } = synthesize(dump);
-    return new TtscGraphMemory(dump.project, nodes, edges);
+    return new TtscGraphMemory(dump.project, nodes, edges, dump.provenance);
   }
 
   /** The node with this id, or undefined. */
@@ -149,8 +155,8 @@ function basename(file: string): string {
 
 /**
  * Derive the structural layer from a dump's faithful facts: refine class-member
- * variables to properties, add a `file` node per workspace source, and connect
- * the `contains` ownership tree and `exports` surface.
+ * variables to properties, add a `file` node per workspace source, connect the
+ * `contains` ownership tree, and re-anchor compiler-owned `exports` edges.
  */
 function synthesize(dump: ITtscGraphDump): {
   nodes: ITtscGraphNode[];
--- a/packages/graph/src/model/TtscGraphSourceReader.ts
+++ b/packages/graph/src/model/TtscGraphSourceReader.ts
@@ -1,0 +1,83 @@
+import { createHash } from "node:crypto";
+import fs from "node:fs";
+import path from "node:path";
+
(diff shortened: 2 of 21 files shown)


samchon:
## Implementation and Self-Review status

The combined #757 + #764 implementation is now present locally; this pull request remains draft and the issues themselves remain untouched.

The first complete solo Self-Review found two real object-member boundary defects before the implementation push:

1. The empty static key `[""]` collided with the builder's unsupported-member sentinel and was omitted. The identity helper now returns an explicit validity bit, so an empty static name is preserved while spreads and genuinely dynamic computed names remain absent.
2. A class-valued property could carry nested class member source into its compact signature. Native AST signature boundaries now stop method/accessor/function bodies at `{`, arrows at `=>`, class values at `class`, object values at `{`, and array values at `[`. Compact signatures are additionally bounded to 160 Unicode code points.
3. One regression-test comment called a signature both “source-free” and rendered from source. The implementation was already snapshot-owned and body-bounded; the comment now says that precisely instead of conflating “no source bodies” with “not source-derived.”

Regression coverage first reproduced the old body leak and the empty-key omission, then passed after remediation. It asserts that method, getter, setter, arrow, function, class, nested-object, and array body/content sentinels never appear on the wire. The real MCP path with a branch-built `ttscgraph.exe` also passes the snapshot identity and provenance-gated source-reader cases.

Local evidence completed before this comment:

- branch-built Graph MCP feature tests: targeted cases green;
- full Graph E2E suite: 21/21 green;
- native graph and `ttscgraph` command suites: green;
- `go vet`, Graph/test-graph TypeScript typechecks, and `@ttsc/graph` build: green;
- `packages/graph/src/structures/ITtscGraphApplication.ts` and `packages/graph/src/TtscGraphApplication.ts`: zero diff.

The implementation uses dump schema v5. Object-member identity and its bounded outline ride on the compiler snapshot. Remaining source-derived signatures/docs use one immutable per-snapshot reader and are returned only when the current bytes match the provenance `checkerDigest`; success and failure are both cached.

Next gates are a fresh full verification and Self-Review Round 2 from the complete remediated diff, followed by the deterministic offline tour comparison. The non-VS Code 14 cells will run only from the final implementation commit/native binary. The two VS Code cells and the model arm remain deferred until the campaign lead releases that exclusive lane.

Issue #818, opened 2026-07-20
fix(graph): perform the post-assembly verification every non-escape result claims, or refuse to return the result


samchon:
## Problem

Every non-escape `@ttsc/graph` MCP result opens with an audit that states, as its own first sentence, that the server assembled the result and then **checked it**: that every name, span, edge, signature, member, value and step in it was taken back to the type-checked program for the snapshot the call synced to and resolved there, that "the server verified them again on the way out", and therefore that "for this snapshot a fact here cannot be wrong". It then instructs the caller to trust every fact, to re-verify none, and to treat a returned span as a citation rather than a cue to open the file.

No such pass exists. The layer that assembles the result and stamps the audit holds no `Program` and no checker. `TtscGraphApplication.inspect_typescript_graph` awaits its graph source once, runs exactly one pure projection over an in-memory `TtscGraphMemory`, and returns one of four unconditional string constants selected purely by request type. Nothing in that path resolves a projected fact against a compiler.

The distinction is not cosmetic, and it is not "the facts came from the checker". The facts really do originate in checker work, and that provenance is genuine. What the audit asserts on top of it is a **second party at a later time**: a check performed by the server, after assembly, on the assembled result. PR #400 chose exactly that subject and timing on purpose, and measured the consumer behavior it produced. That pass is absent, so the claim that a returned fact "cannot be wrong" is unbacked, and the accompanying instruction suppresses the one action that could expose a projection defect, a schema-valid but wrong producer fact, or a reconstruction bug.

Expected behavior: the audit's claim holds for every result that carries it. Because the public contract that publishes this promise is immutable (see Coordination), the only available direction is to make the claim true.

Affected: every agent consuming the MCP tool, on all six non-escape request branches (`trace`, `overview`, `details`, `lookup`, `entrypoints`, `tour`). The same assertion is reflected into the generated MCP tool instructions and into the typed output documentation, so it reaches the model twice.

## Evidence

All anchors below are unchanged between the audited commit `2b724664eee71c57cd688c55122cac137774bf62` and `master` at `612abc2d20d3f99429319081c1f6520bbe34ac7a`.

### The claim

`packages/graph/src/server/resultAudit.ts`, `RESULT_AUDIT` (verbatim, first paragraph):

```text
The server assembled this `result`, then checked it: every fact in it — each name, span,
edge, signature, and step — was taken back to the type-checked program for the snapshot this
call synced to, and resolved there. The compiler resolved these facts and the server
[...]

PR #860, opened 2026-07-20
fix(factory): keep printed text faithful to the AST across parentheses, list commas and JSX whitespace

### LINKED ISSUE #836: fix(factory): keep JSX text meaning independent of `printWidth`
## Problem

`@ttsc/factory`'s `TsPrinter` is a width-aware pretty-printer: `printWidth` is supposed to choose a layout, not change the program. For JSX it changes the program.

`TsPrinter` renders `JsxElement` and `JsxFragment` children by joining them inside a `group` with a `softline` before every child, and it emits `JsxText` verbatim. When the group fits, the children stay on one line and the text keeps its bytes. When the group breaks, two mechanisms destroy significant whitespace:

1. Each `softline` becomes a real newline plus indentation. JSX's own whitespace rule then deletes any `JsxText` child that is whitespace-only and contains a newline, and trims leading and trailing whitespace-with-newline off every text child.
2. `printDocToString`'s `newlineTo` strips trailing spaces and tabs from the emitted buffer before writing a newline. That trimming is correct for ordinary code and wrong for `JsxText`, the only node emitted as raw unquoted text.

The result is that the same AST, printed at two widths, produces two different rendered UIs. The default `printWidth` is 80, so ordinary component markup crosses the threshold routinely, and the failure appears and disappears as unrelated content changes length.

Affected users are external consumers of the published `@ttsc/factory` package that generate JSX or TSX. Nothing in this workspace depends on `@ttsc/factory`, and `tests/test-factory` contains no JSX case at all, so nothing in the repository would notice.

### Lead reproduction, with the transpiled `children` as the observable

The probe builds one `JsxElement` for `<div>Hello there, {nameOfTheCurrentlySignedInVisitor}!</div>`, prints it at two widths with the real `TsPrinter` loaded from the audited sources, and transpiles each printed text with `ts-legacy` (`typescript@6.0.3`) under `jsx: react-jsx` so the `children` argument is directly observable.

```text
ttsc jsx width=200 {"children":"[\"Hello there, \", nameOfTheCurrentlySignedInVisitor, \"!\"]"}
ttsc jsx width=40  {"children":"[\"Hello there,\", nameOfTheCurrentlySignedInVisitor, \"!\"]"}
```

`"Hello there, "` becomes `"Hello there,"`. The generated component renders `Hello there,NAME!`.

### Independent reproduction of both whitespace classes

A reviewer built two ASTs and printed and transpiled each at two widths.

`<div>{alpha} {bravo}</div>`, two expression children separated by one significant space:

```text
=== printWidth 200 ===
<div>{alphaAlphaAlphaAlpha} {bravoBravoBravoBravo}</div>;
--- transpiled ---
_jsxs("div", { children: [alphaAlphaAlphaAlpha, " ", bravoBravoBravoBravo] });

=== printWidth 40 ===
<div>
  {alphaAlphaAlphaAlpha}

  {bravoBravoBravoBravo}
</div>;
--- transpiled ---
_jsxs("div", { children: [alphaAlphaAlphaAlpha, bravoBravoBravoBravo] });
```

The whitespace-only child is deleted outright. `<div>Hello there, {userName}!</div>`, the far more common shape:

```text
=== printWidth 200 ===
<div>Hello there, {userNameuserNameuserName}!</div>;
--- transpiled ---
_jsxs("div", { children: ["Hello there, ", userNameuserNameuserName, "!"] });

=== printWidth 40 ===
<div>
  Hello there,
  {userNameuserNameuserName}
  !
</div>;
--- transpiled ---
_jsxs("div", { children: ["Hello there,", userNameuserNameuserName, "!"] });
```

### Root cause in the printer

At the audited baseline, `TsPrinter.emitNode`'s JSX cases:

```ts
case "JsxElement":
  return group(
    concat([
      this.emit(node.openingElement),
      indent(
        concat(node.children.map((c) => concat([softline, this.emit(c)]))),
      ),
      softline,
      this.emit(node.closingElement),
    ]),
  );
```

`JsxFragment` is the same shape with the opening and closing fragment tokens, and `JsxText` is `return node.text;`. In `packages/factory/src/internal/doc.ts`, `newlineTo` is the trailing-trim half:

```ts
const newlineTo = (ind: number): void => {
  if (out.length)
    out[out.length - 1] = out[out.length - 1]!.replace(/[ \t]+$/, "");
  out.push(newLine + tab.repeat(ind));
  pos = tab.length * ind;
};
```

### Coverage and ownership

- `tests/test-factory/src/features/` has twelve directories and none is `jsx`. There is no JSX test file anywhere in the suite, despite thirteen `createJsx*` builders and thirteen printer cases. The JSX surface has never had a behavioral witness.
- Closed `@ttsc/factory` issues #238, #354, #355, #356, #357 and #383 never mention JSX children or whitespace.
- No open issue and no open pull request touches `packages/factory`. The diff between the audited baseline and current `master` over `packages/factory` is a version-field bump only, so this is live on `master` and in the published `@ttsc/factory@0.19.3`.

## Consequence surface

| Whitespace class | Trigger | Effect once the group breaks |
| --- | --- | --- |
| Whitespace-only text child between two expressions | `{a} {b}` | The child is deleted; the rendered output loses the separator |
| Leading or trailing space of a text child adjacent to an expression | `Hello there, {n}!` | The edge space is trimmed; words run together |
| A `JsxText` whose own text already contains newlines | multi-line literal text | Emitted verbatim inside an `indent`, so continuation lines are neither re-indented nor width-accounted |

Both element and fragment cases are affected identically, because they share the code shape. The failure nests: one long attribute at the top of a tree forces the outer group to break, which puts every descendant text child on its own line, so whitespace is stripped deep inside a subtree that is itself short.

The blast radius is external consumers of the published package on every platform. There is no configuration that avoids it short of raising `printWidth` beyond every element's width. The corruption is silent: the printed text is valid JSX, it compiles, and only the rendered string differs.

Compare `TsPrinter`'s comment handling, which already recognizes that raw text needs different treatment: `commentBody` explicitly splits on newlines and rejoins with `hardline` so a multi-line comment stays aligned. `JsxText` received no equivalent.

## Approach

**Invariant.** `printWidth` selects layout only; it must never change the value of the printed program. For JSX that means a line break introduced between children must not delete a whitespace-only child and must not trim a text child's significant edge whitespace.

**Architectural owner.** `TsPrinter.emitNode`'s `JsxElement`, `JsxFragment` and `JsxText` cases, together with `printDocToString`'s trailing-whitespace trim in `packages/factory/src/internal/doc.ts`, which must become inert for raw JSX text.

Two directions are known to work, and the choice is the implementer's: materialize a lost separator as an explicit `{" "}` child, which is what Prettier does, or refuse to break where the break would not be whitespace-preserving. Whichever is chosen, both whitespace classes must be sealed in the same change. The deleted whitespace-only child and the trimmed text edge are the same root cause, and the trimmed edge is the more common of the two, so fixing one direction alone leaves the reachable half of the class open.

**Positive.** For each of `<div>a {b}</div>`, `<div>{a} {b}</div>`, `<>{a} {b}</>` and `<div>Hello there, {n}!</div>`, the printed text transpiles to the same `children` array at `printWidth` 200, 80, 40 and 10.

**Negative twins.** An element whose children carry no significant whitespace, such as `<div>{a}{b}</div>`, must still break freely and must not gain a `{" "}`. Non-JSX code must keep today's trailing-whitespace trimming: a broken argument list, object literal or binary expression must not start emitting trailing spaces before a newline.

**Boundary.** A whitespace-only element (`<div> </div>`); a text child that already contains a newline; nested fragments; a self-closing element; an element whose opening tag alone exceeds `printWidth`; an outer element that breaks while an inner one fits.

**Oracle.** Expectations must come from the JSX whitespace specification, exercised through `ts-legacy` transpilation of both layouts and compared as `children` arrays, not from whatever `TsPrinter` currently emits. An expectation copied from the code's own output cannot fail when the behavior is wrong: it asserts only that the printer still does what it did the day the test was written, and it turns the defect into a documented requirement that the next reader will defend. This surface has no existing tests, so every expectation added here is a fresh chance to get that wrong.

**Coverage owner.** A new `tests/test-factory/src/features/jsx/` directory. It does not exist today, and its absence is why thirteen builders and thirteen printer cases shipped with no behavioral witness.

**Proving commands.**

| Scope | Command | What it executes |
| --- | --- | --- |
| Narrow | `pnpm --filter @ttsc/test-factory start` | Every `test_*` export under `tests/test-factory/src/features/`; this is verbatim the `factory` lane in `.github/workflows/test.yml`. No build step is needed: in-repo, `@ttsc/factory` resolves to `src/index.ts` |
| Narrowed rerun | `pnpm --filter @ttsc/test-factory start --include=jsx` | Substring filter over **exported case names**, not directory paths. Creating `features/jsx/` is not by itself enough for this filter to match: name the new cases `test_jsx_*` so it does |
| Broad | `pnpm run test:features` | Every `tests/test-*` package, one at a time |

The runner cannot pass vacuously. `TestExecutor.main` in `tests/utils/src/TestExecutor.ts` exits 1 with `No tests were discovered under …`, or `No tests matched --include=…`, whenever the execution list is empty. That matters more here than elsewhere: `--include=jsx` matches nothing today, and it must fail rather than report success until the new cases land.

## Coordination

- **Same-file siblings in this batch.** The width-break trailing-comma issue, the optional-chain and heritage-clause parenthesizer issue, and the signature-and-escaping issue all edit `packages/factory/src/TsPrinter.ts`. This issue additionally edits `packages/factory/src/internal/doc.ts`, which the trailing-comma issue's `delim` work reads but does not change. Sequence the four under one owner, or land them in a fixed order, rather than running them in parallel worktrees.
- **No competing work.** No open issue and no open pull request touches `packages/factory`, and no closed factory issue covers JSX.
- **No existing expectations change.** Because there is no JSX test today, this fix breaks nothing in the suite; the risk is entirely in the new expectations, which is why the oracle requirement above is not optional.

### LINKED ISSUE #837: fix(factory): a comma added only because a list broke changes whether the printed text parses and what it means
## Problem

`TsPrinter` appends a synthetic trailing comma to every delimited list it breaks across lines. That comma is meant to be cosmetic. For three constructs it is not: it decides whether the text parses at all, and it decides how many elements the printed array has.

The decision is made in `TsPrinter.delim`, which emits `opts.trailingComma ? ifBreak(",") : ""` without inspecting the last element. `TsPrinter.listTrailingComma` exists to compute that flag safely, and it is wired into exactly three call sites. Three other call sites hardcode `true`:

| Call site | Trailing-comma decision |
| --- | --- |
| `TsPrinter.params` (parameter lists) | `listTrailingComma(params)` |
| `ObjectBindingPattern` | `listTrailingComma(...)` |
| `ArrayBindingPattern` | `listTrailingComma(...)` |
| `ArrayLiteralExpression` | hardcoded `{ trailingComma: true, forceBreak: node.multiLine === true }` |
| `ObjectLiteralExpression` | hardcoded `{ space: true, trailingComma: true, forceBreak: node.multiLine === true }` |
| `TsPrinter.args` (call and `new` arguments) | hardcoded `{ trailingComma: true }` |

The three unguarded sites are exactly the ones where the comma is load-bearing:

- An array or object literal is not only an rvalue. It is also how a **destructuring assignment target** is represented, and ECMAScript forbids a comma after an assignment rest element. `[a, b, ...rest] = source` parses flat and becomes `SyntaxError: Rest element must be last element` once it breaks, at the **default `printWidth: 80`** with ordinary identifiers.
- An array literal may end in an `OmittedExpression`. There the comma is the token that materializes the hole, so the flat layout has one element fewer than the broken layout. Both parse; they are different programs.
- A call-argument list may end in an `OmittedExpression`. There the broken layout is a hard syntax error.

`multiLine: true` on an array or object literal sets `forceBreak`, so the array case triggers at any width, not only past the threshold.

Affected users are external consumers of the published `@ttsc/factory` package. Nothing in this workspace depends on it, so no in-repo suite would ever see this.

### Lead reproductions, observed output

Destructuring assignment target with a rest element, ordinary identifiers, at the default width. The printed text is fed to V8 through `new Function`:

```text
rest-target width=200 {"text":"[firstDestructuredBinding, secondDestructuredBinding, ...remainingDestructuredBindings] = sourceCollectionValue;","parsed":"parses"}
rest-target width=80  {"text":"[\n  firstDestructuredBinding,\n  secondDestructuredBinding,\n  ...remainingDestructuredBindings,\n] = sourceCollectionValue;","parsed":"SyntaxError: Rest element must be last element"}
```

Array literal ending in an elision, printed at two widths and evaluated, against the same node printed by the pinned `ts-legacy` (`typescript@6.0.3`) oracle:

```text
ttsc array width=200 {"printed":"[\"alpha\", \"bravo\", ]","runtimeLength":2}
ttsc array width=20  {"printed":"[\n  \"alpha\",\n  \"bravo\",\n  ,\n]","runtimeLength":3}
ts-legacy array      {"printed":"[\"alpha\", \"bravo\", ,]","runtimeLength":3}
```

Call-argument list ending in an elision:

```text
ttsc call width=200 {"printed":"handlerFunctionName(\"alphaAlphaAlphaAlphaAlpha\", )","parsed":"parses"}
ttsc call width=20  {"printed":"handlerFunctionName(\n  \"alphaAlphaAlphaAlphaAlpha\",\n  ,\n)","parsed":"SyntaxError: Unexpected token ','"}
```

A reviewer reproduced the same call-argument case through the repository's own parity oracle rather than V8, and confirmed the `multiLine` trigger:

```text
call flat        "f(aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa, )"                ok
call broken      "f(\n  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa,\n  ,\n)"      PARSE-ERR: Argument expression expected.
array multiLine  "[\n  1,\n  ,\n]"                                    ok  (arity 2, vs arity 1 flat)
```

### Negative twins, verified correct today

A reviewer ran 19 shapes at 6 widths, comparing the structural signature of the re-parsed AST across widths, plus a dedicated tuple probe. These must not regress:

| Shape | Flat | Broken |
| --- | --- | --- |
| rvalue array literal ending in a spread, `const t = [a, ...b]` | parses | parses; the comma is legal |
| call arguments ending in a spread, `callee(a, ...b)` | parses | parses; the comma is legal |
| tuple type ending in a rest, `type T = [string, ...number[]]` | 0 diagnostics | 0 diagnostics; the comma is legal |
| parameter list ending in a rest | correct | correct, guarded |
| array and object binding patterns ending in a rest | correct | correct, guarded |
| array binding pattern ending in an elision | correct | correct, guarded |
| interior elision, `["a", , "b"]` | round-trips | round-trips |

The interior-elision row is why this survives casual inspection: only a **trailing** elision is affected.

### The premise that sealed the class, and where it is false

`listTrailingComma`'s doc comment at the audited baseline:

> A trailing comma after a rest element (`...rest`) is a syntax error (TS1013 / V8 `SyntaxError`), and one after a trailing elision (`OmittedExpression`) is not cosmetic: `[a, ,]` parses to one more hole than `[a, ]`, so the flat and broken layouts of the same node would disagree. Call arguments and array / object literals are unaffected. A trailing comma after a spread is legal there.

The final sentence is true for a **spread in rvalue position** and false for the two cases it is used to exclude. In destructuring assignment-target position the same `ArrayLiteralExpression` and `ObjectLiteralExpression` node kinds carry an assignment rest element, where the comma is a syntax error; and a trailing elision is an arity question, not a legality question, so the spread argument does not answer it at all.

The same premise appears in the commit message of `17d1927cd`, "fix(factory): drop trailing comma after rest and trailing elision in broken lists", which closed **#356** through PR **#374**:

> Other delim consumers are unaffected: a trailing comma after a spread is legal in call arguments and array / object literals, and legal after a rest element in tuple types.

The tuple half is true and was verified. The array and object literal half is false in assignment-target position.

### The repository already states the correct rule, in its other implementation

`packages/lint/linthost/rules_format_trailing_comma.go` implements the same rule for the Go formatter. Its helper's doc comment:

> `isRestAssignmentTargetLiteral` reports whether node is an object- or array-literal used as a destructuring ASSIGNMENT TARGET whose last element is a rest (`...x`). ECMAScript forbids a trailing comma after an `AssignmentRestElement` / `AssignmentRestProperty`, so neither this rule nor the print-width printer (which shares this helper) may add one there.

The file header states the same distinction at length, and explains why the discrimination has two prongs rather than one:

> Destructuring assignment TARGETS are the one array/object-literal shape where the same rest restriction bites: `({ a, ...rest } = obj)` and `[a, ...rest] = arr` parse as ObjectLiteralExpression / ArrayLiteralExpression (not binding patterns) and are therefore visited, yet a trailing comma after their `AssignmentRestProperty` / `AssignmentRestElement` is a syntax error.

`isRestAssignmentTargetLiteral` is consulted by `rules_format_trailing_comma.go` for both literal kinds and by the Go print-width printer in `print_nodes_array.go` and `print_nodes_object.go`. That printer therefore does not have this defect. `@ttsc/factory`'s `TsPrinter` is a separate TypeScript implementation of the same rule and shares nothing with it.

### Why this is one repair short of a sealed class, not a newly discovered case

- **Purpose.** PR #374's stated purpose was to stop a synthetic width-break comma from producing invalid or arity-changing output. That is a property of the comma, not of one node kind.
- **Control flow.** The guard was written as a general helper, `listTrailingComma`, and wired into three of the six `delim` call sites that can receive a rest- or elision-terminated list. The other three kept a literal `true`.
- **Consequence.** The identical failure still reproduces through the untouched sites, including the exact arity divergence the guard's own comment describes with the example `[a, ,]` against `[a, ]`, and it reproduces at the **default** width with ordinary identifiers rather than at a contrived width.
- **History.** The surviving cases were not overlooked in silence. They were examined and declared unaffected, in both the doc comment and the commit message, on a premise that only holds in rvalue position. The class was declared closed while a live member of it remained, and the correct invariant was already written down elsewhere in this repository.

### Ownership and provenance

`packages/factory` was vendored in PR #234; the trailing-comma logic is owned entirely in-repo and imports no `typescript`. #356 is closed and scoped to the two node kinds that are already correct. No open issue and no open pull request touches `packages/factory`. The diff between the audited baseline and current `master` over `packages/factory` is a version-field bump only, so this is live on `master` and in the published `@ttsc/factory@0.19.3`.

## Consequence surface

Complete case matrix for the one cause, a `trailingComma: true` that is not derived from the list's last element:

| Construct | Position | Last element | Flat | Broken, or `multiLine: true` |
| `ArrayLiteralExpression` | rvalue | ordinary | ok | ok |
| `ArrayLiteralExpression` | rvalue | `SpreadElement` | ok | ok |
| `ArrayLiteralExpression` | rvalue | `OmittedExpression` | arity *n-1* | arity *n*, a silent value change |
| `ArrayLiteralExpression` | assignment target: `=`, `createAssignment`, `for…of`, `for…in`, nested target | `SpreadElement` | ok | **SyntaxError: Rest element must be last element** |
| `ObjectLiteralExpression` | rvalue | `SpreadAssignment` | ok | ok |
| `ObjectLiteralExpression` | assignment target, all forms above | `SpreadAssignment` | ok | **SyntaxError: Rest element must be last element** |
| call and `new` arguments | any | `OmittedExpression` | parses | **SyntaxError: Unexpected token ','** |
| call and `new` arguments | any | `SpreadElement` | ok | ok |
| `TupleTypeNode`, `NamedImports`, `NamedExports` | any | rest or ordinary | ok | ok; no elision is admissible and a tuple rest comma is legal |

Nested targets are hit as well: `[outer, [inner, ...innerRest]] = source` is valid flat and a syntax error broken, because the corruption is on the **inner** literal, which can break on its own.

The behavior is platform-independent, being pure string layout with no path or line-ending involvement. The blast radius is external consumers of the published package: the same generator emits correct output while its content is short and corrupt output once the content grows past `printWidth`, so a downstream project can pass its own tests and ship a broken generated file.

## Approach

**Invariant.** A comma the printer adds only because a group broke must never change whether the text parses, nor what it parses to. `listTrailingComma` already encodes exactly that invariant for two node kinds; the correction is to make the remaining `delim` consumers obey the same rule rather than to add a third special case.

**Architectural owner.** `packages/factory/src/TsPrinter.ts`: the trailing-comma decision in `delim` and every call site that can receive a rest- or elision-terminated list, which today means `ArrayLiteralExpression`, `ObjectLiteralExpression` and `args` in addition to the three already guarded.

**The rest half.** The printer has no parent links, so discriminating assignment-target position would have to be threaded from the emitting sites. Suppressing the synthetic comma after a spread in an array or object literal unconditionally is valid in every position and needs no context. Both are sound; choose one rather than being told which. The Go helper quoted above is a specification reference for the two-pronged discrimination if the positional route is taken; it operates on a different AST and cannot be shared as code.

**The elision half runs the other way.** The lead's oracle run shows the legacy printer emitting the comma in **both** layouts for a factory-built array (`["alpha", "bravo", ,]`, length 3), so ttsc's broken layout matches the oracle and its flat layout does not. Two reviewers disagreed about which direction is oracle-correct here, one reading it as parity with `emitList`'s original-`NodeArray` behavior. Re-derive the direction from the pinned oracle for each construct rather than assuming suppression, and make the flat and broken layouts agree with each other and with whatever the oracle prints.

**Preserve the deliberate asymmetry.** For `ArrayBindingPattern` the legacy oracle drops a trailing hole, because a trailing hole binds nothing there. That existing behavior is oracle-correct and must not be changed alongside the literal cases.

**Correct the premise as part of the fix.** `listTrailingComma`'s doc comment currently states the exclusion that produced this defect. Leaving it in place lets the next reader derive the same wrong conclusion from the code's own documentation.

**Positive.**

- Broken array and object destructuring assignment targets ending in a rest element parse cleanly, through `=`, `createAssignment`, `for…of`, `for…in`, and a nested inner target.
- A broken rvalue array literal ending in an `OmittedExpression` re-parses to the same arity as its flat layout, and both match the pinned oracle's arity. `tests/test-factory/src/features/width/test_array_binding_break_trailing_elision.ts` is the exemplar to mirror: it already parses both layouts back and compares.
- A call-argument list ending in an `OmittedExpression` parses in both layouts.
- All of the above hold under `multiLine: true` as well as under width pressure.

**Negative twins.** The rvalue array literal ending in a spread, call arguments ending in a spread, and a tuple type ending in a rest keep their trailing comma byte for byte. An interior elision round-trips unchanged. A plain `[a, b]` does not gain a flat trailing comma.

**Boundary.** A lone `OmittedExpression`; single-element and empty lists; a spread that is not last; `f(a, <hole>)` broken, which must not emit two consecutive commas on separate lines; `multiLine: true` forced breaks at every width.

**Regression.** These must keep passing unchanged, because they pin deliberately different, oracle-correct behavior: `tests/test-factory/src/features/width/test_array_binding_break_trailing_elision.ts`, `test_array_binding_break_rest_no_trailing_comma.ts`, `test_object_binding_break_rest_no_trailing_comma.ts`, `test_params_break_rest_no_trailing_comma.ts`, `test_params_rest_short_stays_flat.ts`, `test_array_object_breaks.ts`, and `tests/test-factory/src/features/expressions/test_array_literal.ts`.

**Oracle.** Every new expected string must come from printing the same tree with `ts-legacy`'s own `createPrinter`, or from parsing the printed text back and asserting the structural property (arity, or "parses"). An expectation copied from `TsPrinter`'s current output is not evidence: it asserts only that the printer still does what it did the day the test was written, so it cannot fail when the behavior is wrong, and it converts the defect into a documented requirement. Five width tests were added with PR #374 and every one of them covers a parameter list or a binding pattern, which is why the literal and argument cases stayed invisible.

**Coverage owner.** New cases under `tests/test-factory/src/features/width/`, one per row of the case matrix above, including its negative twins.

**Proving commands.**

| Scope | Command | What it executes |
| --- | --- | --- |
| Narrow | `pnpm --filter @ttsc/test-factory start` | Every `test_*` export under `tests/test-factory/src/features/`; this is verbatim the `factory` lane in `.github/workflows/test.yml`. No build step is needed: in-repo, `@ttsc/factory` resolves to `src/index.ts` |
| Narrowed rerun | `pnpm --filter @ttsc/test-factory start --include=break` | Substring filter over **exported case names**, not directory paths; matches all nine existing width-break cases, plus the unrelated `test_break_continue`, so name the new cases to match too |
| Broad | `pnpm run test:features` | Every `tests/test-*` package, one at a time |

The runner cannot pass vacuously. `TestExecutor.main` in `tests/utils/src/TestExecutor.ts` exits 1 with `No tests were discovered under …`, or `No tests matched --include=…`, whenever the execution list is empty, so a mistyped filter is a failure rather than a silent success.

All seven regression files named above exist today under `tests/test-factory/src/features/width/` and `.../expressions/`, and every one of them runs in the narrow command.

## Coordination

- **Same-file siblings in this batch.** The JSX-whitespace issue, the optional-chain and heritage-clause parenthesizer issue, and the signature-and-escaping issue all edit `packages/factory/src/TsPrinter.ts`. The JSX issue also edits `packages/factory/src/internal/doc.ts`, which this fix reads through `delim` but does not need to change. Sequence the four under one owner, or land them in a fixed order, rather than running them in parallel worktrees.
- **Do not reopen or re-scope #356.** It is closed and its scope is the two node kinds that are already correct. This issue owns the remaining `delim` consumers and the doc comment that excluded them.
- **No competing work.** No open issue and no open pull request touches `packages/factory`.
- **Cross-package note.** `packages/lint/linthost/rules_format_trailing_comma.go` and the Go print-width printer already implement the correct rule and are not changed by this issue. Do not attempt to share the helper across the language boundary; it is quoted here as the specification, not as a dependency.

### LINKED ISSUE #851: fix(factory): parenthesize optional chains and heritage expressions so the printed program still compiles
## Problem

`@ttsc/factory` re-implements the legacy `ts.factory` parenthesizer so that a generated AST prints as source text that means what the AST says. Two operand positions are missing from that implementation, and both produce output the compiler rejects or silently reinterprets.

**Optional chains under a non-optional parent.** The upstream rule is `parenthesizeLeftSideOfAccess(expression, optionalChain)`: a `LeftHandSideExpression` is emitted bare only when the consuming node is itself part of the same chain. `TsPrinter.leftSideExpression` takes no `optionalChain` argument at all:

```ts
private leftSideExpression(expression: Expression): Doc {
  return this.isLeftHandSideExpression(expression)
    ? this.emit(expression)
    : this.parenthesizedExpression(expression);
}
```

`isLeftHandSideExpression` returns true for `PropertyAccessChain`, `ElementAccessChain`, `CallChain` and `NonNullChain`, so a chain nested under a non-optional parent is emitted bare and re-parses as part of the chain. Six of the ten call sites are non-optional consumers (`Decorator`, `PropertyAccessExpression`, `ElementAccessExpression`, `CallExpression`, `NonNullExpression`, `TaggedTemplateExpression`); the other four are the chain cases themselves, where bare emission is correct. `newExpressionTargetNeedsParentheses`, which owns the `new` target, likewise never asks whether its target is a chain.

**Heritage clauses.** `ExpressionWithTypeArguments` emits its expression with a raw `this.emit`:

```ts
case "ExpressionWithTypeArguments":
  return concat([
    this.emit(node.expression),
    this.typeArguments(node.typeArguments),
  ]);
```

Upstream applies `parenthesizeLeftSideOfAccess` inside `createExpressionWithTypeArguments`. This is the only expression-operand position in the printer still using a raw `this.emit`; every other `this.emit(node.<expr>)` site sits inside a delimiter or after a statement keyword, where any expression is grammatically legal. `class` and `interface` `extends` clauses, `implements` clauses, and class expressions all funnel through this one branch.

The failures split two ways. Some emit source that does not compile. Others compile and quietly change what the program does: a chain that should throw on a nullish head instead evaluates to `undefined`.

Affected users are external consumers of the published `@ttsc/factory` package, on every platform, with no configuration that avoids it.

### Lead reproduction: the same AST printed both ways, then compiled

The probe builds each node with `@ttsc/factory` and with the pinned `ts-legacy` (`typescript@6.0.3`) oracle, prints each with its own printer, and runs both texts through `ts.transpileModule` with `reportDiagnostics: true`.

```text
new (a?.b)()
  ttsc:               "new a?.b();"
  ttscDiagnostics:    TS1209: Invalid optional chain from new expression. Did you mean to call 'a()'?
  legacy:             "new (a?.b)();"
  legacyDiagnostics:  compiles

class A extends (X || Y) {}
  ttsc:               "class A extends X || Y {}"
  ttscDiagnostics:    TS1005: ',' expected. | TS1005: '{' expected.
  legacy:             "class A extends (X || Y) {\n}"
  legacyDiagnostics:  compiles
```

### Enumerated chain positions, against the published artifact

A reviewer ran the same differential against the **published** `@ttsc/factory@0.19.3`:

| AST built | `@ttsc/factory` prints | `ts.factory` prints | Result of the ttsc text |
| --- | --- | --- | --- |
| `createCallExpression(createPropertyAccessChain(a, ?., b))` | `a?.b()` | `(a?.b)()` | parses, different program |
| `createNewExpression(createPropertyAccessChain(a, ?., b))` | `new a?.b()` | `new (a?.b)()` | **TS1209** Invalid optional chain from new expression |
| `createPropertyAccessExpression(createPropertyAccessChain(a, ?., b), c)` | `a?.b.c` | `(a?.b).c` | parses, different program |
| `createElementAccessExpression(createPropertyAccessChain(a, ?., b), 0)` | `a?.b[0]` | `(a?.b)[0]` | parses, different program |
| `createPropertyAccessExpression(createCallChain(a, ?.), c)` | `a?.().c` | `(a?.()).c` | parses, different program |
| `createTaggedTemplateExpression(createPropertyAccessChain(a, ?., b), …)` | ``a?.b`x` `` | ``(a?.b) `x` `` | **TS1358** Tagged template expressions are not permitted in an optional chain |
| `createNonNullExpression(createPropertyAccessChain(a, ?., b))` | `a?.b!` | `(a?.b)!` | indistinguishable from `createNonNullChain` |
| `createDecorator(createPropertyAccessChain(a, ?., b))` | `@a?.b` | `@(a?.b)` | **TS1146** Declaration expected |

The four "parses, different program" rows are the dangerous half. Running both emitted texts with `const a = null`:

```text
ttsc emitted  a?.b()    => undefined
ts   emitted  (a?.b)()  => THROWS TypeError: (intermediate value) is not a function
ttsc emitted  a?.b.c    => undefined
ts   emitted  (a?.b).c  => THROWS TypeError: Cannot read properties of undefined (reading 'c')
ttsc emitted  a?.b[0]   => undefined
ts   emitted  (a?.b)[0] => THROWS TypeError: Cannot read properties of undefined (reading '0')
ttsc emitted  a?.().c   => undefined
ts   emitted  (a?.()).c => THROWS TypeError: Cannot read properties of undefined (reading 'c')
```

A generator that faithfully models "call this method, and it must exist" emits a program that swallows the error instead.

### Enumerated heritage shapes

The printed class declaration is fed back to the parser:

| Heritage expression | `@ttsc/factory` prints | `ts.factory` prints | Parse of the ttsc text |
| --- | --- | --- | --- |
| `X \|\| Y` | `class A extends X \|\| Y {}` | `class A extends (X \|\| Y) { }` | **TS1005** twice |
| `c ? X : Y` | `class A extends c ? X : Y {}` | `class A extends (c ? X : Y) { }` | **TS1005** twice |
| `Base as any` | `class A extends Base as any {}` | `class A extends (Base as any) { }` | **TS1005** twice |
| `await Base` | `class A extends await Base {}` | `class A extends (await Base) { }` | **TS1005** |
| `() => x` | `class A extends () => x {}` | `class A extends (() => x) { }` | **TS1109**, **TS1005** |
| `X = Y` | `class A extends X = Y {}` | `class A extends (X = Y) { }` | **TS1005** |
| `(a, B)` comma sequence | `class A extends a, B {}` | `class A extends (a, B) { }` | parses, as **two** heritage entries |
| `m?.Base` | `class A extends m?.Base {}` | `class A extends (m?.Base) { }` | parses; the chain-guard case |
| `mixin(Base)` | `class A extends mixin(Base) {}` | identical | clean; negative twin |

The comma-sequence row is the silent one: a single parenthesized base expression becomes two base classes.

### Upstream rule, for reference

The pinned oracle's implementation, which the package mirrors everywhere else:

```js
function parenthesizeLeftSideOfAccess(expression, optionalChain) {
  const emittedExpression = skipPartiallyEmittedExpressions(expression);
  if (isLeftHandSideExpression(emittedExpression)
      && (emittedExpression.kind !== 215 /* NewExpression */ || emittedExpression.arguments)
      && (optionalChain || !isOptionalChain(emittedExpression))) {
    return expression;
  }
  return setTextRange(factory2.createParenthesizedExpression(expression), expression);
}
```

### Why eight rounds of auditing missed it

Three separate audit rounds ran parenthesizer fuzzers over this printer and all three reported it clean: a 1,312-case sweep, a 5,000-case sweep, and a 100,000-case differential. A fourth reviewer ran a 2,282-case expression differential covering every binary-operator pair in both nesting positions and nineteen wrapper shapes, and concluded the parenthesizer was sound.

None of them found this, because **no generator emitted an optional chain or a heritage clause**. The corpora were large and their grammar was narrow. Adding cases to a grammar that cannot produce the shape does not increase coverage of it, which is why the acceptance below constrains the grammar rather than the count.

The in-repo tests have the same shape of gap. `tests/test-factory/src/features/expressions/test_optional_chaining.ts` prints each chain form standalone and never nests one under a non-optional parent. `test_expression_context_parentheses.ts` exercises every tight context with a binary operand only. The three heritage-clause tests, `test_class_declaration.ts`, `test_interface_declaration.ts` and `test_full_module.ts`, all use a plain identifier as the base.

### Ownership and provenance

- The parenthesizer arrived whole in `06d30fe76`, "fix(factory): preserve printer operand precedence" (PR #243), which ported the legacy rules and dropped the `optionalChain` argument. Later printer work (#369, #374, #381) touched arrow precedence, rest commas and the `new` target spine; none revisited chains or `ExpressionWithTypeArguments`.
- Closed #238's parenthesizer table does not enumerate chains or heritage clauses, and #241, #354, #355, #356 and #357 are all closed and about other printer classes. This was not a knowingly skipped case; it is a gap in a faithful port.
- No open issue and no open pull request touches `packages/factory`. The diff between the audited baseline and current `master` over `packages/factory` is a version-field bump only, so this is live on `master` and in the published `@ttsc/factory@0.19.3`.

## Consequence surface

| Class | Reachable through | Failure |
| --- | --- | --- |
| Optional chain under a non-optional access, call or element access | `PropertyAccessExpression`, `ElementAccessExpression`, `CallExpression` | Compiles; the nullish-head case evaluates to `undefined` instead of throwing |
| Optional chain as a `new` target | `NewExpression` | TS1209, does not compile |
| Optional chain as a tagged-template tag | `TaggedTemplateExpression` | TS1358, does not compile |
| Optional chain as a decorator expression | `Decorator` | TS1146, does not compile |
| Optional chain under a non-null assertion | `NonNullExpression` | Output is indistinguishable from `NonNullChain`, so a print round trip cannot preserve the distinction |
| Non-left-hand-side heritage expression | `class extends`, `class implements`, class expressions, `interface extends` | Six shapes do not compile |
| Comma-sequence heritage expression | same | Compiles as two base classes instead of one |

`class X extends (Base as any) {}` and `class X extends (cond ? A : B) {}` are ordinary code-generation shapes: mixins, conditional bases, and casts around a generic base. Every one of them emits a file that does not compile, so the failure is loud for heritage clauses and silent for chains.

## Approach

**Invariant.** Every operand position the grammar requires to be a `LeftHandSideExpression` or `MemberExpression` must be emitted through one parenthesizer helper, and that helper must take the `optionalChain` flag of the **consuming** node, exactly as `parenthesizeLeftSideOfAccess(expression, optionalChain)` does upstream.

**Architectural owner.** `TsPrinter.leftSideExpression` and its ten call sites, `newExpressionTargetNeedsParentheses`, and the `ExpressionWithTypeArguments` case, which must route through the same helper instead of a raw `this.emit`.

Do not add per-node ad hoc parentheses. Commit `06d30fe76` owns the rule that parenthesizer decisions stay centralized and are checked against the precedence table rather than patched at the node that surfaced a problem; adding a local wrap at `ExpressionWithTypeArguments` alone would leave the chain half of the same helper unfixed.

**Out of scope.** The `ExpressionStatement`, concise-body and export-default predicates were verified at parity with the oracle and must not change. Type-position parenthesization is a separate surface.

**Positive.** Each of the eight chain rows and the seven failing heritage rows prints the parenthesized form the oracle prints, and the emitted text type-checks clean.

**Negative twins, which must not gain parentheses.** `createPropertyAccessChain(createCallChain(a, ?.), ?., b)` stays `a?.()?.b`; `createElementAccessChain` and `createCallChain` over a chain stay bare; `class A extends mixin(Base) {}` and `interface I extends a.B {}` stay unparenthesized; `createNonNullChain(a?.b)` stays `a?.b!`.

**Extend the differential corpus's grammar, not its case count.** Three fuzzers of 1,312, 5,000 and 100,000 cases all passed on this defect because their generators could not produce the shape. Any harness this issue adds or extends must be able to generate, in every operand position it already covers:

- each optional-chain node kind (`PropertyAccessChain`, `ElementAccessChain`, `CallChain`, `NonNullChain`), including a chain whose head is itself a chain;
- heritage-clause expressions drawn from the full expression grammar, not identifiers only, for `class extends`, `class implements`, class expressions and `interface extends`;
- consuming positions that are easy to omit: decorators, tagged-template tags, non-null assertions, and `new` targets.

A case-count increase without those productions is not acceptance evidence.

**Re-derive the one existing expectation this changes.** `tests/test-factory/src/features/expressions/test_new_expression_target_spine_operator_parentheses.ts` asserts `new (f?.().bar)()` for `NewExpression(PropertyAccessExpression(CallChain(f), bar))`. That expectation is itself a witness of this defect and must be re-derived by building the same tree with `ts-legacy`'s factory and printing it with `ts.createPrinter()`, then taking the oracle's string. Do not carry the current value forward, and do not adopt a reviewer's predicted value without re-running the oracle.

An expectation taken from the code's own output is not evidence. It asserts only that the printer still does what it did the day the test was written, so it cannot fail when the behavior is wrong, and it converts the defect into a documented requirement that the next reader will defend. This suite already carries at least two such expectations, so re-derivation is a requirement of this change, not a style preference.

**No other existing expectation should move.** `test_optional_chaining.ts` prints standalone chains only, and the three heritage tests use plain identifiers. If any other assertion changes, that is a signal the fix over-parenthesized and must be investigated rather than accommodated.

**Boundary.** A chain whose head is a parenthesized expression; `new` with and without an argument list on the target spine; an `ExpressionWithTypeArguments` that carries type arguments as well as a non-trivial expression; `implements` with several entries; a class expression in expression-statement position.

**Proving commands.**

| Scope | Command | What it executes |
| --- | --- | --- |
| Narrow | `pnpm --filter @ttsc/test-factory start` | Every `test_*` export under `tests/test-factory/src/features/`; this is verbatim the `factory` lane in `.github/workflows/test.yml`. No build step is needed: in-repo, `@ttsc/factory` resolves to `src/index.ts` |
| Narrowed rerun | `pnpm --filter @ttsc/test-factory start --include=new_expression_target` (or `--include=optional_chaining`) | Substring filter over **exported case names**, not directory paths; both of these match existing cases today |
| Broad | `pnpm run test:features` | Every `tests/test-*` package, one at a time |

The runner cannot pass vacuously. `TestExecutor.main` in `tests/utils/src/TestExecutor.ts` exits 1 with `No tests were discovered under …`, or `No tests matched --include=…`, whenever the execution list is empty, so a mistyped filter is a failure rather than a silent success. A differential harness added under this issue must fail the same way when its generator produces no case for a declared production, so that a grammar gap reports as a failure instead of an empty pass.

The existing files this issue touches or must leave unchanged, all present today: `tests/test-factory/src/features/expressions/test_new_expression_target_spine_operator_parentheses.ts` (the one expectation that changes), `test_optional_chaining.ts`, `test_expression_context_parentheses.ts`, `tests/test-factory/src/features/declarations/test_class_declaration.ts`, `test_interface_declaration.ts`, and `tests/test-factory/src/features/integration/test_full_module.ts`.

## Coordination

- **Same-file siblings in this batch.** The JSX-whitespace issue, the width-break trailing-comma issue and the signature-and-escaping issue all edit `packages/factory/src/TsPrinter.ts`. This issue's region (the parenthesizer helpers and the `ExpressionWithTypeArguments` case) is disjoint from theirs, but they are the same file. Sequence the four under one owner, or land them in a fixed order, rather than running them in parallel worktrees.
- **Shared harness.** The signature-and-escaping issue also asks for an oracle-driven gate against the pinned `ts-legacy`. If a differential harness is built here, build it once and let both issues use it rather than standing up a second one.
- **No competing work.** No open issue and no open pull request touches `packages/factory`. Closed #238, #241, #354, #355, #356 and #357 own other printer classes and must not be reopened by this fix.

samchon:
Campaign batch **B3** of the `repository-audit-2026-07-19` campaign. One owner, one branch, one pull request for the three `@ttsc/factory` printer issues whose common cause is the same: **the printed text does not mean what the AST says.**

Closes #851
Closes #837
Closes #836

## Scope

| Issue | Defect | Owner in the code |
| --- | --- | --- |
| #851 | Optional chains under a non-optional parent, and heritage-clause expressions, are emitted without the parentheses the grammar requires. `new (a?.b)()` prints as `new a?.b();` (TS1209); `class A extends (X \|\| Y) {}` prints without the parentheses (TS1005). Four further chain rows compile and silently change the program. | `TsPrinter.leftSideExpression` and its call sites, `newExpressionTargetNeedsParentheses`, the `ExpressionWithTypeArguments` case |
| #837 | A trailing comma added only because a list broke changes whether the text parses and what arity it has, in destructuring assignment targets and in argument lists ending in an elision — at the **default** `printWidth: 80`. | `TsPrinter.delim` and the three call sites that hardcode `trailingComma: true` |
| #836 | JSX children are joined with a line-break-capable separator, so `printWidth` deletes whitespace-only children and trims significant edge whitespace off text children. | `TsPrinter`'s `JsxElement`/`JsxFragment`/`JsxText` cases and `printDocToString`'s trailing-space trim in `internal/doc.ts` |

They are batched together because all three edit `packages/factory/src/TsPrinter.ts`; the issues themselves require the same-file siblings to be sequenced under one owner rather than run in parallel worktrees.

#818 (signature and escaping parity) is **not** in this batch and this branch does not touch its surface.

## Method commitment

Three prior audit rounds fuzzed this parenthesizer with 1,312, 5,000 and 100,000 cases and all reported it clean, because no generator could emit an optional chain or a heritage clause. This batch therefore **extends the corpus grammar, not the case count**.

Several existing `tests/test-factory` expectations were taken from this printer's own output rather than from the pinned `ts-legacy` oracle, so they pin the defect. Every expectation this branch touches is re-derived from the oracle and the change is stated explicitly in a later comment on this thread.

## Campaign CI

Per `.agents/skills/issue-campaign/development.md`, automatic CI is suspended for this branch: every push passes an exact-SHA cancellation gate and the results are recorded in `.wiki/repository-audit-2026-07-19/ci-state.md`. Repository Actions permissions and workflow states are unchanged. `pnpm format` is not run on this branch; Post-Campaign Cleanup owns the repository-wide formatter result.

Local verification is the gate. Commands and results are posted to this thread as the work lands.

## Status

Draft, claim only. No implementation in this commit.

https://claude.ai/code/session_016HZ6QbzkgY5nKd14RKEqdN

### DIFF
--- a/packages/factory/README.md
+++ b/packages/factory/README.md
@@ -89,6 +89,8 @@ factory.createCallExpression(id("foo"), undefined, [a, b]); // foo(a, b)
 // )
 ```
 
+`printWidth` picks a layout, never a meaning. The break-time trailing comma is dropped where it would change the program (after the rest element of a destructuring assignment target, or after an argument-list hole), and it is written in both layouts where it is a value (the hole in `["a", "b", ,]`). JSX children stay on one line whenever a break would delete a whitespace-only child or trim the significant edge space off a text child, so the same tree renders the same text at every width.
+
 ### Comments
 
 Attach leading / trailing comments with the legacy `ts.addSyntheticLeadingComment` family. The printer renders them in place — multi-line bodies re-indent with their node, so JSDoc on a nested member stays aligned.
--- a/packages/factory/src/TsPrinter.ts
+++ b/packages/factory/src/TsPrinter.ts
@@ -22,6 +22,7 @@ import {
   join,
   line,
   printDocToString,
+  raw,
   softline,
 } from "./internal/doc";
 import { NodeFlags, SyntaxKind } from "./syntax";
@@ -116,7 +117,7 @@ export class TsPrinter {
     close: string,
     opts: {
       space?: boolean;
-      trailingComma?: boolean;
+      trailingComma?: TrailingComma;
       forceBreak?: boolean;
     } = {},
   ): Doc {
@@ -126,7 +127,11 @@ export class TsPrinter {
       concat([
         open,
         indent(concat([ln, join(concat([",", line]), items)])),
-        opts.trailingComma ? ifBreak(",") : "",
+        opts.trailingComma === "always"
+          ? ","
+          : opts.trailingComma === "onBreak"
+            ? ifBreak(",")
+            : "",
         ln,
         close,
       ]),
@@ -191,28 +196,78 @@ export class TsPrinter {
           args.map((a) => this.emit(a)),
           ">",
           {
-            trailingComma: false,
+            trailingComma: "never",
           },
         )
       : "";
   }
 
   /**
-   * Whether a broken parameter list / binding pattern may append a synthetic
-   * trailing comma after its last element.
+   * Trailing-comma policy for a parameter list or binding pattern.
    *
-   * A trailing comma after a rest element (`...rest`) is a syntax error (TS1013
-   * / V8 `SyntaxError`), and one after a trailing elision (`OmittedExpression`)
-   * is not cosmetic: `[a, ,]` parses to one more hole than `[a, ]`, so the flat
-   * and broken layouts of the same node would disagree. Call arguments and
-   * array / object literals are unaffected — a trailing comma after a spread is
-   * legal there.
+   * A comma the printer adds only because a group broke must never change
+   * whether the text parses, nor what it parses to. After a rest element
+   * (`...rest`) it changes the first: a trailing comma there is a syntax error
+   * (TS1013 / V8 `SyntaxError`). After a trailing elision it changes the
+   * second: `[a, ,]` has one more hole than `[a, ]`, so the flat and broken
+   * layouts of the same node would disagree. A binding pattern is the one place
+   * where dropping that hole is lossless, since a trailing hole binds nothing;
+   * {@link literalTrailingComma} materializes it instead, because in an array
+   * literal the hole is a value.
    */
-  private listTrailingComma(nodes: readonly Node[]): boolean {
+  private listTrailingComma(nodes: readonly Node[]): TrailingComma {
     const last: Node | undefined = nodes[nodes.length - 1];
-    if (last === undefined) return true;
-    if (last.kind === "OmittedExpression") return false;
-    return !("dotDotDotToken" in last && last.dotDotDotToken !== undefined);
+    if (last === undefined) return "onBreak";
+    if (last.kind === "OmittedExpression") return "never";
+    return "dotDotDotToken" in last && last.dotDotDotToken !== undefined
+      ? "never"
+      : "onBreak";
+  }
+
+  /**
+   * Trailing-comma policy for a call or `new` argument list.
+   *
+   * A trailing `OmittedExpression` prints as nothing, so the list already ends
+   * in the separator comma of its last real argument: `f(a, )`, which is what
+   * the legacy printer emits too and parses as one argument. Adding the break
+   * comma on top produces `f(a, ,)`, which is a syntax error. A trailing spread
+   * is unaffected — a comma after it is legal in an argument list.
+   */
+  private argsTrailingComma(args: readonly Expression[]): TrailingComma {
+    const last: Expression | undefined = args[args.length - 1];
+    return last !== undefined && last.kind === "OmittedExpression"
+      ? "never"
+      : "onBreak";
+  }
+
+  /**
+   * Trailing-comma policy for an array or object literal.
+   *
+   * Two positions make the comma load-bearing rather than cosmetic.
+   *
+   * A trailing elision is a **value**: the comma is the token that materializes
+   * the hole, so `["a", ]` has one element and `["a", ,]` has two. The legacy
+   * printer emits it in every layout, so this printer emits it in every layout
+   * too; leaving it to the break would make the same node mean different things
+   * at different widths.
+   *
+   * A destructuring **assignment target** is the same node kind as an rvalue
+   * literal, but ECMAScript forbids a comma after its `AssignmentRestElement` /
+   * `AssignmentRestProperty`: `[a, ...rest,] = source` is a syntax error, while
+   * the identical rvalue `[a, ...rest,]` is legal. Only the target position
+   * suppresses it, so the rvalue twin keeps its break comma.
+   */
+  private literalTrailingComma(
+    elements: readonly Node[],
+    assignmentTarget: boolean,
+  ): TrailingComma {
+    const last: Node | undefined = elements[elements.length - 1];
+    if (last === undefined) return "onBreak";
+    if (last.kind === "OmittedExpression") return "always";
+    return assignmentTarget &&
+      (last.kind === "SpreadElement" || last.kind === "SpreadAssignment")
+      ? "never"
+      : "onBreak";
   }
 
   private params(params: readonly Node[]): Doc {
@@ -232,7 +287,7 @@ export class TsPrinter {
       args.map((a) => this.expressionForDisallowedComma(a)),
       ")",
       {
-        trailingComma: true,
+        trailingComma: this.argsTrailingComma(args),
       },
     );
   }
@@ -277,6 +332,51 @@ export class TsPrinter {
       : "";
   }
 
+  /**
+   * Lay out a JSX element's or fragment's children.
+   *
+   * A line break between JSX children is not cosmetic. JSX deletes a
+   * whitespace-only text child that contains a newline and trims
+   * whitespace-carrying-a-newline off both edges of every other text child, so
+   * a break introduced only because the group did not fit changes what the
+   * component renders: `<div>Hello there, {name}!</div>` becomes
+   * `Hello there,NAME!`, and the separator in `<div>{a} {b}</div>` disappears
+   * outright.
+   *
+   * Children are therefore laid out across lines only when the break survives
+   * that transformation unchanged: every text child must carry non-whitespace
+   * content, must not begin or end with whitespace, and must not sit next to
+   * another text child, since inserting a newline between two of them would
+   * merge into one text with a space in the middle. Otherwise the children are
+   * emitted verbatim on one line, whatever `printWidth` says — width may choose
+   * a layout, never a meaning.
+   */
+  private jsxChildren(
+    open: Doc,
+    children: readonly Node[],
+    close: Doc,
+  ): Doc {
+    if (!this.jsxChildrenMayBreak(children))
+      return concat([open, concat(children.map((c) => this.emit(c))), close]);
+    return group(
+      concat([
+        open,
+        indent(concat(children.map((c) => concat([softline, this.emit(c)])))),
+        softline,
+        close,
+      ]),
+    );
+  }
+
+  private jsxChildrenMayBreak(children: readonly Node[]): boolean {
+    return children.every(
+      (child, index) =>
+        child.kind !== "JsxText" ||
+        (isBreakSafeJsxText(child.text) &&
+          children[index + 1]?.kind !== "JsxText"),
+    );
+  }
+
   private optType(type: Node | undefined): Doc {
     return type ? concat([": ", this.emit(type)]) : "";
   }
@@ -285,8 +385,16 @@ export class TsPrinter {
     return body ? concat([" ", this.emit(body)]) : ";";
   }
 
-  private emit(node: Node): Doc {
-    const body: Doc = this.emitNode(node);
+  /**
+   * @param assignmentTarget Whether `node` occupies destructuring
+   *   assignment-target position, where an array or object literal is a pattern
+   *   rather than a value. The flag is set by the assignment and `for…in` /
+   *   `for…of` cases, forwarded by every node that is transparent to it (a
+   *   spread, a property's initializer, a parenthesis, an `=` default), and
+   *   dropped by every other node.
+   */
+  private emit(node: Node, assignmentTarget: boolean = false): Doc {
+    const body: Doc = this.emitNode(node, assignmentTarget);
     const leading: SynthesizedComment[] | undefined =
       getSyntheticLeadingComments(node);
     const trailing: SynthesizedComment[] | undefined =
@@ -338,7 +446,7 @@ export class TsPrinter {
     ]);
   }
 
-  private emitNode(node: Node): Doc {
+  private emitNode(node: Node, assignmentTarget: boolean): Doc {
     switch (node.kind) {
       /* names & tokens */
       case "Identifier":
@@ -350,7 +458,7 @@ export class TsPrinter {
       case "Token":
         return node.token;
       case "Decorator":
-        return concat(["@", this.leftSideExpression(node.expression)]);
+        return concat(["@", this.leftSideExpression(node.expression, false)]);
 
       /* literals */
       case "StringLiteral":
@@ -364,18 +472,29 @@ export class TsPrinter {
       case "ArrayLiteralExpression":
         return this.delim(
           "[",
-          node.elements.map((e) => this.expressionForDisallowedComma(e)),
+          node.elements.map((e) =>
+            this.expressionForDisallowedComma(e, assignmentTarget),
+          ),
           "]",
-          { trailingComma: true, forceBreak: node.multiLine === true },
+          {
+            trailingComma: this.literalTrailingComma(
+              node.elements,
+              assignmentTarget,
+            ),
+            forceBreak: node.multiLine === true,
+          },
         );
       case "ObjectLiteralExpression":
         return this.delim(
           "{",
-          node.properties.map((p) => this.emit(p)),
+          node.properties.map((p) => this.emit(p, assignmentTarget)),
           "}",
           {
             space: true,
-            trailingComma: true,
+            trailingComma: this.literalTrailingComma(
+              node.properties,
+              assignmentTarget,
+            ),
             forceBreak: node.multiLine === true,
           },
         );
@@ -383,7 +502,7 @@ export class TsPrinter {
         return concat([
           this.emit(node.name),
           ": ",
-          this.expressionForDisallowedComma(node.initializer),
+          this.expressionForDisallowedComma(node.initializer, assignmentTarget),
         ]);
       case "ShorthandPropertyAssignment":
         return concat([
@@ -400,24 +519,24 @@ export class TsPrinter {
       case "SpreadAssignment":
         return concat([
           "...",
-          this.expressionForDisallowedComma(node.expression),
+          this.expressionForDisallowedComma(node.expression, assignmentTarget),
         ]);
       case "PropertyAccessExpression":
         return concat([
-          this.leftSideExpression(node.expression),
+          this.leftSideExpression(node.expression, false),
           ".",
           this.emit(node.name),
         ]);
       case "ElementAccessExpression":
         return concat([
-          this.leftSideExpression(node.expression),
+          this.leftSideExpression(node.expression, false),
           "[",
           this.expressionForDisallowedComma(node.argumentExpression),
           "]",
         ]);
       case "CallExpression":
         return concat([
-          this.leftSideExpression(node.expression),
+          this.leftSideExpression(node.expression, false),
           this.typeArguments(node.typeArguments),
           this.args(node.arguments),
         ]);
@@ -429,11 +548,19 @@ export class TsPrinter {
           this.args(node.arguments ?? []),
         ]);
       case "ParenthesizedExpression":
-        return concat(["(", this.emit(node.expression), ")"]);
+        return concat(["(", this.emit(node.expression, assignmentTarget), ")"]);
       case "BinaryExpression":
+        // the left side of `=` is a destructuring assignment target, both for a
+        // top-level assignment and for a `[a = init]` default inside one
         return group(
           concat([
-            this.binaryOperand(node.operator, node.left, true),
+            this.binaryOperand(
+              node.operator,
+              node.left,
+              true,
+              undefined,
+              node.operator === SyntaxKind.EqualsToken,
+            ),
             " ",
             node.operator,
             indent(
@@ -501,11 +628,11 @@ export class TsPrinter {
           this.emit(node.type),
         ]);
       case "NonNullExpression":
-        return concat([this.leftSideExpression(node.expression), "!"]);
+        return concat([this.leftSideExpression(node.expression, false), "!"]);
       case "SpreadElement":
         return concat([
           "...",
-          this.expressionForDisallowedComma(node.expression),
+          this.expressionForDisallowedComma(node.expression, assignmentTarget),
         ]);
       case "AwaitExpression":
         return concat(["await ", this.prefixUnaryOperand(node.expression)]);
@@ -543,7 +670,7 @@ export class TsPrinter {
           node.elements.map((e) => this.emit(e)),
           "]",
           {
-            trailingComma: true,
+            trailingComma: "onBreak",
           },
         );
       case "ParenthesizedTypeNode":
@@ -564,8 +691,11 @@ export class TsPrinter {
       case "TypeQueryNode":
         return concat(["typeof ", this.emit(node.exprName)]);
       case "ExpressionWithTypeArguments":
+        // heritage clauses take a LeftHandSideExpression: `class A extends
+        // (X || Y) {}` does not parse without the parentheses, and a bare comma
+        // sequence silently becomes two base classes
         return concat([
-          this.emit(node.expression),
+          this.leftSideExpression(node.expression, false),
           this.typeArguments(node.typeArguments),
         ]);
       case "PropertySignature":
@@ -840,7 +970,7 @@ export class TsPrinter {
           "{",
           node.elements.map((e) => this.emit(e)),
           "}",
-          { space: true, trailingComma: true },
+          { space: true, trailingComma: "onBreak" },
         );
       case "ImportSpecifier":
         return concat([
@@ -868,7 +998,7 @@ export class TsPrinter {
           "{",
           node.elements.map((e) => this.emit(e)),
           "}",
-          { space: true, trailingComma: true },
+          { space: true, trailingComma: "onBreak" },
         );
       case "ExportSpecifier":
         return concat([
@@ -914,7 +1044,7 @@ export class TsPrinter {
       case "ForInStatement":
         return concat([
           "for (",
-          this.emit(node.initializer),
+          this.emit(node.initializer, true),
           " in ",
           this.emit(node.expression),
           ") ",
@@ -925,7 +1055,7 @@ export class TsPrinter {
           "for ",
           node.awaitModifier ? "await " : "",
           "(",
-          this.emit(node.initializer),
+          this.emit(node.initializer, true),
           " of ",
           this.emit(node.expression),
           ") ",
@@ -1195,7 +1325,7 @@ export class TsPrinter {
         ]);
       case "TaggedTemplateExpression":
         return concat([
-          this.leftSideExpression(node.tag),
+          this.leftSideExpression(node.tag, false),
           this.typeArguments(node.typeArguments),
           this.emit(node.template),
         ]);
@@ -1276,13 +1406,13 @@ export class TsPrinter {
         ]);
       case "PropertyAccessChain":
         return concat([
-          this.leftSideExpression(node.expression),
+          this.leftSideExpression(node.expression, true),
           node.questionDotToken ? "?." : ".",
           this.emit(node.name),
         ]);
       case "ElementAccessChain":
         return concat([
-          this.leftSideExpression(node.expression),
+          this.leftSideExpression(node.expression, true),
           node.questionDotToken ? "?." : "",
           "[",
           this.expressionForDisallowedComma(node.argumentExpression),
@@ -1290,27 +1420,20 @@ export class TsPrinter {
         ]);
       case "CallChain":
         return concat([
-          this.leftSideExpression(node.expression),
+          this.leftSideExpression(node.expression, true),
           node.questionDotToken ? "?." : "",
           this.typeArguments(node.typeArguments),
           this.args(node.arguments),
         ]);
       case "NonNullChain":
-        return concat([this.leftSideExpression(node.expression), "!"]);
+        return concat([this.leftSideExpression(node.expression, true), "!"]);
 
       /* jsx */
       case "JsxElement":
-        return group(
-          concat([
-            this.emit(node.openingElement),
-            indent(
-              concat(
-                node.children.map((c) => concat([softline, this.emit(c)])),
-              ),
-            ),
-            softline,
-            this.emit(node.closingElement),
-          ]),
+        return this.jsxChildren(
+          this.emit(node.openingElement),
+          node.children,
+          this.emit(node.closingElement),
         );
       case "JsxSelfClosingElement":
         return concat([
@@ -1331,24 +1454,19 @@ export class TsPrinter {
       case "JsxClosingElement":
         return concat(["</", this.emit(node.tagName), ">"]);
       case "JsxFragment":
-        return group(
-          concat([
-            this.emit(node.openingFragment),
-            indent(
-              concat(
-                node.children.map((c) => concat([softline, this.emit(c)])),
-              ),
-            ),
-            softline,
-            this.emit(node.closingFragment),
-          ]),
+        return this.jsxChildren(
+          this.emit(node.openingFragment),
+          node.children,
+          this.emit(node.closingFragment),
         );
       case "JsxOpeningFragment":
         return "<>";
       case "JsxClosingFragment":
         return "</>";
       case "JsxText":
-        return node.text;
+        // the one node emitted as unquoted source text: its trailing spaces are
+        // rendered content, so they must survive the layout engine's line trim
+        return raw(node.text);
       case "JsxAttribute":
         return node.initializer === undefined
           ? this.emit(node.name)
@@ -1626,16 +1744,63 @@ export class TsPrinter {
       : concat(["(", this.emit(expression), ")"]);
   }
 
-  private expressionForDisallowedComma(expression: Expression): Doc {
+  private expressionForDisallowedComma(
+    expression: Expression,
+    assignmentTarget: boolean = false,
+  ): Doc {
     return this.expressionPrecedence(expression) > ExpressionPrecedence.Comma
-      ? this.emit(expression)
+      ? this.emit(expression, assignmentTarget)
       : this.parenthesizedExpression(expression);
   }
 
-  private leftSideExpression(expression: Expression): Doc {
-    return this.isLeftHandSideExpression(expression)
-      ? this.emit(expression)
-      : this.parenthesizedExpression(expression);
+  /**
+   * Emit an operand the grammar requires to be a `LeftHandSideExpression`,
+   * mirroring the legacy parenthesizer's
+   * `parenthesizeLeftSideOfAccess(expression, optionalChain)`.
+   *
+   * `optionalChain` is the **consuming** node's own chain-ness, not the
+   * operand's. An optional chain may be emitted bare only when the node
+   * consuming it continues the same chain: `a?.b?.()` is one chain, while
+   * `(a?.b)()` is a plain call on the chain's value. Emitting the second as
+   * `a?.b()` re-parses as the first, which stops throwing on a nullish head,
+   * and in `new`, tagged-template and decorator position it does not compile at
+   * all.
+   */
+  private leftSideExpression(
+    expression: Expression,
+    optionalChain: boolean,
+  ): Doc {
+    return this.leftSideNeedsParentheses(expression, optionalChain)
+      ? this.parenthesizedExpression(expression)
+      : this.emit(expression);
+  }
+
+  /**
+   * Whether {@link leftSideExpression} wraps this operand.
+   *
+   * The legacy rule also parenthesizes an argument-less `new` here, because it
+   * prints `new X` bare and `new X.y` would re-parse with `y` on the target.
+   * This printer always emits the argument list, so `new X().y` already says
+   * what the tree says and needs no wrapper.
+   */
+  private leftSideNeedsParentheses(
+    expression: Expression,
+    optionalChain: boolean,
+  ): boolean {
+    if (!this.isLeftHandSideExpression(expression)) return true;
+    return !optionalChain && this.isOptionalChain(expression);
+  }
+
+  private isOptionalChain(expression: Expression): boolean {
+    switch (expression.kind) {
+      case "CallChain":
+      case "ElementAccessChain":
+      case "NonNullChain":
+      case "PropertyAccessChain":
+        return true;
+      default:
+        return false;
+    }
   }
 
   private newExpressionTarget(expression: Expression): Doc {
@@ -1648,21 +1813,76 @@ export class TsPrinter {
    * Whether a `new` target must be parenthesized to keep its call arguments
    * from re-binding to the `new` — mirroring the legacy printer's
    * `parenthesizeExpressionOfNew`. A `new` target is grammatically a
-   * `MemberExpression`, so a call anywhere on the target's left spine (not just
-   * a direct one: `new (f().bar)()`, `new (a.b().c)()`) would otherwise
-   * re-parse with the call's arguments consumed by the `new` — a different
-   * program. Argument-less `new` on the spine is kept parenthesized for
-   * continuity with the direct case, though this printer always prints an
-   * argument list, which already disambiguates it.
+   * `MemberExpression`, so a call anywhere on the target's printed left spine
+   * (not just a direct one: `new (f().bar)()`, `new (a.b().c)()`) would
+   * otherwise re-parse with the call's arguments consumed by the `new` — a
+   * different program. Argument-less `new` on the spine is kept parenthesized
+   * for continuity with the direct case, though this printer always prints an
+   * argument list, which already disambiguates it. Anything else falls back to
+   * the shared left-side rule, which is what parenthesizes an optional-chain
+   * target (`new (a?.b)()`, TS1209 without it).
    */
   private newExpressionTargetNeedsParentheses(expression: Expression): boolean {
-    if (!this.isLeftHandSideExpression(expression)) return true;
-    const leftmost: Expression = this.leftmostExpression(expression, true);
-    return (
-      leftmost.kind === "CallExpression" ||
-      leftmost.kind === "CallChain" ||
-      (leftmost.kind === "NewExpression" && leftmost.arguments === undefined)
-    );
+    const leftmost: Expression | undefined =
+      this.leftmostPrintedExpression(expression);
+    if (leftmost !== undefined) {
+      if (leftmost.kind === "CallExpression" || leftmost.kind === "CallChain")
+        return true;
+      if (leftmost.kind === "NewExpression")
+        return leftmost.arguments === undefined;
+    }
+    return this.leftSideNeedsParentheses(expression, false);
+  }
+
+  /**
+   * The node whose own text opens `expression`'s printed form, or `undefined`
+   * when that text opens with a printer-inserted `(`.
+   *
+   * The legacy factory parenthesizes each operand as it builds the node, so its
+   * `getLeftmostExpression` walk halts on the resulting
+   * `ParenthesizedExpression`. This printer decides the same parentheses at
+   * emit time instead, so the walk has to ask {@link leftSideNeedsParentheses}
+   * the same question directly; otherwise `new` re-wraps a target whose call is
+   * already behind parentheses, and `new (f?.()).bar()` comes out as
+   * `new ((f?.()).bar)()`. Calls halt the walk, matching the legacy
+   * `stopAtCallExpressions` mode this predicate is the only user of.
+   */
+  private leftmostPrintedExpression(
+    expression: Expression,
+  ): Expression | undefined {
+    switch (expression.kind) {
+      case "CallExpression":
+      case "CallChain":
+        return expression;
+      case "ElementAccessExpression":
+      case "NonNullExpression":
+      case "PropertyAccessExpression":
+        return this.leftmostPrintedLeftSide(expression.expression, false);
+      case "ElementAccessChain":
+      case "NonNullChain":
+      case "PropertyAccessChain":
+        return this.leftmostPrintedLeftSide(expression.expression, true);
+      case "TaggedTemplateExpression":
+        return this.leftmostPrintedLeftSide(expression.tag, false);
+      case "AsExpression":
+      case "SatisfiesExpression":
+        return this.leftmostPrintedExpression(expression.expression);
+      case "BinaryExpression":
+        return this.leftmostPrintedExpression(expression.left);
+      case "ConditionalExpression":
+        return this.leftmostPrintedExpression(expression.condition);
+      default:
+        return expression;
+    }
+  }
+
+  private leftmostPrintedLeftSide(
+    operand: Expression,
+    optionalChain: boolean,
+  ): Expression | undefined {
+    return this.leftSideNeedsParentheses(operand, optionalChain)
+      ? undefined
+      : this.leftmostPrintedExpression(operand);
   }
 
   private prefixUnaryOperand(operand: Expression, operator?: SyntaxKind): Doc {
@@ -1728,6 +1948,7 @@ export class TsPrinter {
     operand: Expression,
     isLeftSide: boolean,
     leftOperand?: Expression,
+    assignmentTarget: boolean = false,
   ): Doc {
     return this.binaryOperandNeedsParentheses(
       operator,
@@ -1736,7 +1957,7 @@ export class TsPrinter {
       leftOperand,
     )
       ? this.parenthesizedExpression(operand)
-      : this.emit(operand);
+      : this.emit(operand, assignmentTarget);
   }
 
   private binaryOperandNeedsParentheses(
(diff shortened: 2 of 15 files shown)


samchon:
## Self-Review, round 1

Complete pass over the `origin/master...HEAD` diff (15 files) plus the consequence surface: the printer's other operand positions, the layout engine, the assignment-target thread, the new test oracle, and the package README. Three findings survive verification; the remediation plan follows each.

### 1. `PartiallyEmittedExpression` bypasses the whole left-side parenthesizer, including everything this branch just fixed

The legacy rule opens with `skipPartiallyEmittedExpressions(expression)` before it decides anything. This printer's `leftSideNeedsParentheses` and both leftmost walks look at the node they are handed, and `isLeftHandSideExpression` reports a `PartiallyEmittedExpression` as left-hand-side whatever it wraps. Every class this branch closes is therefore still reachable through one wrapper. Measured against the pinned oracle on this branch's current head:

| tree | `@ttsc/factory` | `ts.factory` |
| --- | --- | --- |
| `PropertyAccess(PEE(PropertyAccessChain(a, ?., b)), c)` | `a?.b.c` | `(a?.b).c` |
| `PropertyAccess(PEE(x \|\| y), c)` | `x \|\| y.c` | `(x \|\| y).c` |
| `New(PropertyAccess(PEE(f()), bar))` | `new f().bar()` | `new (f().bar)()` |
| `class A extends PEE(X \|\| Y)` | `class A extends X \|\| Y {}` | `class A extends (X \|\| Y) {}` |

Row 2 is a plain precedence inversion (`x || y.c` means `x || (y.c)`), row 3 re-binds the argument list to the `new`, and row 4 is issue #851's TS1005 heritage case surviving intact. `createPartiallyEmittedExpression` is public API and the printer has always had a case for it, so this is live, not theoretical.

**Remediation.** Route the predicate through a `skipPartiallyEmittedExpressions` helper, exactly where the legacy rule puts it, and let both leftmost walks descend through the wrapper as `getLeftmostExpression` does. The parentheses still wrap the original node, wrapper included, which is also what the legacy factory does.

**Scoped out, reported not fixed.** The same wrapper also defeats `binaryOperandNeedsParentheses` and the assertion-operand predicate. Those are different helpers, they are not part of #851's declared owner set, and the gap predates this branch, so I am recording it here rather than widening the batch. It is worth its own issue.

### 2. The JSX oracle helper can report "no children" instead of failing

`jsxChildren` transpiles the printed text and returns `"<no children>"` when it cannot find a `children` property. Only one of the three JSX cases asserts against that sentinel, so a print that stopped producing children at all would compare equal across widths and pass. The helper also requests transpile diagnostics and never reads them.

**Remediation.** Make the helper throw on a transpile diagnostic and on a missing `children` property, so a broken print fails the case instead of collapsing into a constant.

### 3. `isBreakSafeJsxText`'s positive branch has no witness

Every committed JSX case exercises the *refuse to break* side. An element whose single text child carries no edge whitespace (`<div>Hello</div>`) is the one shape the predicate is supposed to let break, and nothing asserts it does. Without it, replacing the predicate with a constant `false` would leave the suite green.

**Remediation.** Add that case to the boundary set, asserting both that the layout breaks and that the rendered children are unchanged.

### Verified, no change needed

- The assignment-target flag reaches every route in issue #837's matrix (`=`, `createAssignment`, `for…of`, `for…in`, nested inner target, a target behind a spread, a target behind a property assignment, and an `=` default) and is dropped by `ShorthandPropertyAssignment`'s `objectAssignmentInitializer`, which is a default value rather than a target.
- No path can reach an array or object literal in target position through `parenthesizedExpression`, which does not forward the flag: both literals have primary precedence, so neither `expressionForDisallowedComma` nor `binaryOperandNeedsParentheses` ever parenthesizes them.
- The `raw` doc's trailing-trim exemption cannot leak into generated code: only `JsxText` produces it, and the negative twin over a broken call, object literal and binary expression is asserted.
- `delim`'s `"always"` policy is unreachable for an empty list, which short-circuits before the comma is emitted.
- Tuple types, named imports and named exports keep `"onBreak"`; their trailing commas are unchanged byte for byte.

### Out of scope, recorded for the lead

- `factory.createComma(a, b)` prints `a , b` where the legacy printer prints `a, b`. `tests/test-factory/src/features/expressions/test_all_operator_aliases.ts` pins the current spacing. It is a binary-operator spacing question, not a parenthesization, a width-break comma or JSX, so none of #851, #837 or #836 owns it.
- `@ttsc/factory`'s `Expression` union does not include the JSX node kinds, so `createExpressionStatement` rejects a `JsxElement` even though JSX is a primary expression upstream. The new JSX cases print the element directly instead; widening a published union is a public-API change that needs its own decision.

samchon:
## Self-Review, round 2

Fresh complete pass over the current tree. It opens with a correction to round 1, then one new finding, then what round 2 verified clean.

### Correction: round 1's finding 1 was over-stated, and its fix is reverted

I claimed the `PartiallyEmittedExpression` parenthesizer gap was "live, not theoretical". It is not reachable through the typed public surface, and I should have checked that before proposing the fix.

`PartiallyEmittedExpression` is a member of `Node` but **not** of the exported `Expression` union, so `createPropertyAccessExpression`, `createNewExpression`, `createExpressionWithTypeArguments` and every other operand-taking builder reject it at compile time. It can only be printed standalone, where no parenthesizer applies. The remediation I wrote does not even compile against the package's own types: `expression.kind === "PartiallyEmittedExpression"` on an `Expression` is an unreachable comparison (TS2367), which is the type system stating the same thing.

The four rows I measured were built through untyped probe code. The runtime behavior they show is genuine, but reaching it requires a cast, and closing it properly means deciding whether `Expression` should list `PartiallyEmittedExpression` at all. That is a published-type decision, it is the same omission that keeps the JSX node kinds out of `Expression`, and it belongs to the typing-parity surface rather than to #851, #837 or #836.

So the printer changes for it are reverted, along with the three corpus operands that needed them. Recorded below for the lead instead.

One piece of that round survives on its own merit: the corpus gained a `typeof` operand consumer, which exercises the prefix-unary parenthesizer that nothing else in the differential covered. It passes.

### New finding: the trailing-comma fix introduced a line of pure indentation

An `OmittedExpression` prints as the empty string. Once `argsTrailingComma` and `listTrailingComma` suppress the break comma after one, the broken layout ends with a separator, a newline, an indent, and then nothing:

```
handlerFunctionName(
  "alphaAlphaAlphaAlphaAlpha",
  ⟵ two spaces, then the closing line
);
```

The layout engine has a trim for exactly this, but it only inspects the last chunk it wrote, and the empty item is that chunk, so the real end of the line hides one chunk behind it. The array binding pattern with a trailing elision has the same shape and has had this blemish since PR #374; the arity assertions in `test_array_binding_break_trailing_elision.ts` never looked at the whitespace.

**Applied.** `printDocToString` no longer pushes empty strings at all. They contribute nothing to the output or the column, and skipping them restores the trim's invariant that the last chunk is the end of the line. Both layouts now end cleanly, and `test_call_args_break_trailing_elision` asserts no line ends in a space or tab. Reverting the skip fails that assertion.

### Verified clean in round 2

- The three trailing-comma policies are decided from the last element only, and every `delim` consumer now names its policy explicitly; type arguments keep `"never"`, tuple types, named imports and named exports keep `"onBreak"`.
- `"always"` is unreachable for an empty list, which short-circuits before the comma.
- The `raw` doc cannot leak: `JsxText` is its only producer, and skipping empty strings does not reset the raw tail, which is the more accurate reading of where the line ends.
- The JSX oracle helper now throws on a transpile diagnostic and on a missing `children` property, so a print that stopped producing children fails instead of comparing equal to itself across widths.
- `isBreakSafeJsxText`'s positive branch has a witness: a text child with no edge whitespace breaks at width 10 and renders identically at 200, 80, 40 and 10.
- Mutation-checked, each independently: reverting the chain guard fails 28 differential cells plus both focused cases; reverting either trailing-comma policy fails all four width cases; reverting the JSX break guard fails both positive JSX cases while the negative twin stays green; reverting the raw-text exemption fails only the negative-twin case that owns it; reverting the empty-string skip fails the new whitespace assertion.

### Standing out-of-scope findings for the lead

1. **`Expression` omits `PartiallyEmittedExpression` and the JSX node kinds.** Upstream both are expressions. The consequences are that a JSX element cannot be placed in any expression position through the typed API, and that the parenthesizer's `skipPartiallyEmittedExpressions` step has no typed way to matter. Worth its own issue against the typing surface; it overlaps #818's territory.
2. **`factory.createComma(a, b)` prints `a , b`; the legacy printer prints `a, b`.** `test_all_operator_aliases.ts` pins the current spacing. Binary-operator spacing, owned by none of the three issues in this batch.

samchon:
## Self-Review, round 3

Fresh complete pass over the current tree, including the surface outside the diff: the package's exports, its dependencies and consumers, and the printer's statelessness.

### Finding: the JSX cases could not cover an element with no children

`jsxChildren` was hardened in round 2 to throw when the transpiled output has no `children` property. That is right for every shape the suite covers, but it also made the one boundary that legitimately has none unreachable: `<div></div>` emits `_jsx("div", {})`, with no `children` argument at all. The boundary set was silently missing it, and adding it would have thrown rather than asserted.

**Applied.** `jsxChildren` and a new `jsxEmit` now share one `transpileJsx` step; `jsxEmit` returns the structural signature of the whole emitted module, which exists for every shape. The childless element is asserted through it at widths 200 and 5 and emits the identical call. Two case doc comments were corrected to match the cases they actually run.

### Verified clean in round 3

- `packages/factory/src/internal/doc.ts` is not re-exported from the package index, so the new `raw` doc node and the widened `Doc` union are internal. `packages/factory/package.json` and its rollup config are untouched, and no dependency was added.
- `tests/test-factory` is still the only workspace consumer of `@ttsc/factory`, so the blast radius of the behavior change is external consumers of the published package and this suite.
- `TsPrinter` remains stateless. The assignment-target flag is a parameter and the raw-tail flag is local to `printDocToString`, so a printer instance is still safe to share, which the tests now do.
- Line endings are unaffected: the `newLine` option still owns every break the printer writes, and raw JSX text is passed through byte for byte as before.
- The assignment-target flag cannot leak into an rvalue: `obj.prop = [1, ...rest]` drops it at the property access on the left and never sets it on the right, so the rvalue array keeps its break comma. `+=` and `??=` do not set it at all.
- A `for…of` / `for…in` initializer that is a `VariableDeclarationList` drops the flag at that case, so binding patterns keep `listTrailingComma` and the deliberate trailing-hole asymmetry the issue asked to preserve.

Round 4 follows as a clean confirming pass.

samchon:
## Self-Review, round 4 — clean

Complete pass over the current tree. Nothing survives verification, so the review stops here.

Re-read in full: the printer diff region by region, the layout engine, all nine new cases, the shared oracle, the one changed expectation, and the README paragraph. Re-checked the consequence surface: exports, dependencies, workspace consumers, printer statelessness, line endings, and the assignment-target flag's reachability in both directions. No further improvement found.

### Final local verification

Campaign CI is deliberately suspended, so this is the evidence in place of checks.

| command | result |
| --- | --- |
| `pnpm --filter @ttsc/test-factory start` | `Success`, 123 cases. Verbatim the `factory` lane of `.github/workflows/test.yml` |
| `npx tsc --noEmit -p packages/factory/tsconfig.json` | clean |
| `npx tsc --noEmit -p tests/test-factory/tsconfig.json` | clean |
| `pnpm --filter @ttsc/test-factory start --include=optional_chain` | 2 cases, both run |
| `pnpm --filter @ttsc/test-factory start --include=heritage` | 1 case |
| `pnpm --filter @ttsc/test-factory start --include=parenthesizer_oracle` | 1 case |
| `pnpm --filter @ttsc/test-factory start --include=jsx` | 3 cases; this filter matched nothing before this branch |
| `pnpm --filter @ttsc/test-factory start --include=break` | 15 cases: the nine that existed plus the four added here plus the two unrelated `break` matches |

Each filter's output was read for the names it actually executed, not just its exit code, because a filter that matches only pre-existing cases runs green without touching anything new.

**Not run, and why.** `pnpm run test:features` and `node scripts/test-typecheck.cjs` both fail in this environment for reasons unrelated to the change: the feature lanes need `pnpm build:current` and a POSIX `file://` resolution this Windows checkout does not provide, and the typecheck script shells out to a `tsc` that is not on PATH here, failing on `tests/lint-contributor-demo` before it reaches anything in this batch. The typecheck was run directly against both affected projects instead. `tests/test-factory` is the only workspace package that depends on `@ttsc/factory`, verified by grep over every `package.json`, so no other lane can observe this change.

### Which pinned expectations were re-derived

One, exactly as issue #851 required.

`tests/test-factory/src/features/expressions/test_new_expression_target_spine_operator_parentheses.ts`, the "optional call at chain head" row, `New(PropertyAccess(CallChain(f), bar))`:

- previously `new (f?.().bar)()`, which is this printer's own former output;
- now `new (f?.()).bar()`, re-derived by building the tree with `ts-legacy`'s factory and printing it with `ts.createPrinter()`.

The old value is not merely a different formatting of the same program. Inside those parentheses, `f?.().bar` re-parses with `.bar` as part of the optional chain, while the tree says a non-optional property access over a chain. The case now asserts that too, on the re-parsed node. The other four rows in that file are unchanged and each matches the oracle's parenthesization.

No other expectation in the suite moved, which is the signal issue #851 asked for.

### Ready to merge

Implementation, the package-scoped verification above, and this Self-Review all pass. Merging under the campaign's standing mandate.

Issue #806, opened 2026-07-20
fix(test): give every committed executable test a CI owner, proven mechanically instead of by hand-kept lists


samchon:
## Problem

`ttsc` decides what CI runs from two hand-written lists: `scripts/test-go.cjs` names six runner scripts, each of which names its own Go packages, and `.github/workflows/test.yml` names one matrix lane per suite. Nothing binds either list to the test packages that exist on disk. A committed, executable, passing test suite that no list names simply never runs, with no signal of any kind.

Four such owners exist today. All four are green, so this is lost signal rather than a hidden red suite, and every regression they were written to prevent can land with a fully green CI.

This is the same defect closed issue #622 was filed for, and its remediation was a longer list. Within one day of that fix landing, three new orphans appeared and one older one had already survived it. The finite-list design is the cause, so the fix must be an invariant that fails on the next unclaimed suite, not four more names.

## Evidence

Verified on `master` at `612abc2d20d3f99429319081c1f6520bbe34ac7a` as well as on the audited commit `2b724664eee71c57cd688c55122cac137774bf62`. Between the two, the only changes anywhere in `.github/workflows`, `scripts/`, `tests/test-wasm`, `packages/wasm/test`, `website/test`, and `packages/ttsc/shim/ast/test` are `.github/workflows/typia.yml` (an unrelated downstream-fixture override change) and two `package.json` version bumps. All four orphans are still orphaned on `master`.

### The two executors and their lists

`scripts/test-go.cjs`:

```js
const runners = [
  "test-go-driver.cjs",
  "test-go-ttsc.cjs",
  "test-go-transformer.cjs",
  "test-go-utility-plugins.cjs",
  "test-go-lint.cjs",
  "test-go-graph.cjs",
];
```

Each runner hard-codes its own package list. `scripts/test-go-ttsc.cjs` lists `./internal/lspserver`, `./test/cli`, `./test/ttscserver`, `./test/platform`, `./test/utility`, `./cmd/ttsc`, `./cmd/ttscserver`, run with `cwd: packages/ttsc`. `scripts/test-go-utility-plugins.cjs` lists `banner`, `paths`, `strip`. `scripts/test-go-driver.cjs` runs `./test/driver` only. `scripts/test-go-graph.cjs` runs `./internal/graph/...`, `./cmd/ttscgraph/...`, `./cmd/graphbench/...`.

[...]

PR #861, opened 2026-07-20
fix(wasm): keep MemFS a tree, honor descriptor semantics, and correct the pipe2 contract

### LINKED ISSUE #808: fix(wasm): stop advertising the MemFS `pipe2` shim as Go wasm `os.Pipe()` support
## Problem

The public `@ttsc/wasm` filesystem contract tells plugin and host authors that Go's wasm `os.Pipe()` routes through the MemFS `pipe2` shim. It does not, and cannot, under the pinned Go toolchain. `packages/wasm/src/structures/IWasmExecFS.ts` says:

```ts
  /**
   * `pipe2` is what Go's wasm `os.Pipe()` calls. Returns two fds: a read end
   * and a write end. The host's stdout/stderr capture currently uses MemFS temp
   * files, but this keeps direct pipe callers compatible with the wasm fs
   * surface.
   */
```

A reader of that interface reasonably concludes that installing `createMemFS()` makes `os.Pipe()` work inside a custom Go/wasm host. It returns `ENOSYS` instead, because Go rejects the call before it ever reaches JavaScript. The promised caller cannot reach the implementation, so the "keeps direct pipe callers compatible" claim is a compatibility contract the code cannot fulfil.

Affected: anyone building a custom wasm binary against `@ttsc/wasm`'s host package or writing against the published `IWasmExecFS` type. The first-party compiler path is not broken by this, because stdout/stderr capture uses MemFS temp files.

### Go never calls it

`packages/wasm/go.mod` declares `go 1.26`. The pinned toolchain's `src/os/pipe_wasm.go` is the whole implementation for both wasm operating systems:

```go
//go:build wasm

package os

import "syscall"

// Pipe returns a connected pair of Files; reads from r return bytes written to w.
// It returns the files and an error, if any.
func Pipe() (r *File, w *File, err error) {
	// Neither GOOS=js nor GOOS=wasip1 have pipes.
	return nil, nil, NewSyscallError("pipe", syscall.ENOSYS)
}
```

`src/syscall/fs_js.go`, the file that owns the `globalThis.fs` bridge for `GOOS=js`, agrees:

```go
func Pipe(fd []int) error {
	return ENOSYS
}
```

Neither function performs an `fsCall`, so no JavaScript method is consulted. Checked against a local `go1.26.4` toolchain; the published `@ttsc/wasm` binary reports `go1.26.5` through `api.version()`, and both are 1.26 releases with this same implementation.

### History records the boundary the comment contradicts

`cbc56cfe91` (PR #87, "make typia transform actually run in the wasm playground") introduced the pipe state and `pipe2`, and its own commit message states the boundary directly:

```text
1. WASM Go binary's runWithCapturedIO uses os.Pipe to capture plugin
   stdout/stderr. Go's js/wasm syscall returns "pipe: not implemented
   on js" for syscall.Pipe; os.Pipe returns nil files, every plugin
   write silently drops, [...] Fix: capture via MemFS-backed temp files
   (/tmp/ttsc-host-capture-*.{stdout,stderr}) instead — MemFS file I/O
   works where pipes don't.
```

So the working fix deliberately avoided pipes, and the pipe emulation was kept alongside it. The interface wording that claims active Go compatibility arrived later, with the structures split in `c82156437` ("fountain snapshot/AST/checker verbs + structures split"). Later MemFS work (PR #693 boot and tree invariants, PR #708 invocation-output ownership) did not change the syscall boundary and did not make Go route `os.Pipe` here.

### The direct JavaScript path is also incomplete

The shim is reachable from JavaScript through `createMemFS().fs`. A reviewer called it through the compiled artifact after confirming the compiled and audited sources were byte-identical, running `pipe2`, closing the read end, then writing one byte:

```json
{"readFd":100,"writeFd":101,"writeAfterReadCloseError":null,"written":1}
```

A pipe write after the last read end closes must fail; here it reports success and retains the byte. Minimal reproduction:

```js
const { fs } = createMemFS();
fs.pipe2(0, (_pipeError, [readFd, writeFd]) => {
  fs.close(readFd, () => {
    const byte = new Uint8Array([120]);
    fs.write(writeFd, byte, 0, 1, null, (error, written) => {
      console.log(error?.code ?? null, written); // actual: null 1
    });
  });
});
```

The cause is visible in `packages/wasm/src/createMemFS.ts`. `IPipeState` tracks `writeClosed` but has no read-end state. `close` marks `writeClosed` only when the closed descriptor is the write end, and it never settles a reader that was blocked on the descriptor being closed. The `write` branch queues every accepted buffer without asking whether a reader still exists, so repeated writes grow memory with no possible consumer. `packages/wasm/src/MemFSError.ts` has no `EPIPE` mapping, so even a correctly named error would carry the default numeric errno.

Verified against the audited commit `2b724664eee71c57cd688c55122cac137774bf62`; `packages/wasm/src` is byte-identical on current `master` (`612abc2d2`), where only `packages/wasm/package.json` differs by a version bump.

## Scope: what this issue accepts, and what it deliberately does not decide

This issue accepts one thing: **the false Go compatibility contract must go**. `IWasmExecFS`'s `pipe2` documentation must stop asserting that Go's wasm `os.Pipe()` calls it, and must state the real boundary (Go returns `ENOSYS` under the pinned toolchain; host output capture uses MemFS temp files).

Deleting the `pipe2` member and the pipe state is **deferred, not decided**. `@ttsc/wasm` is a published package and `IWasmExecFS` is part of its public type surface, so removal requires first enumerating external JavaScript consumers of `fs.pipe2`. Inside the repository there is no caller: the only occurrences of `pipe2` are the declaration in `IWasmExecFS.ts` and the implementation plus its state in `createMemFS.ts`, and nothing under `tests/`, `website/`, `experimental/`, or `scripts/` mentions it. (The `os.Pipe` calls that do exist, in `packages/ttsc/test` and `packages/lint/test` helpers, are native Go test plumbing and never run under `GOOS=js`.) That is evidence for removal, not proof, and the external check owns the decision.

## Consequence surface

| Surface | Current state | Required |
| --- | --- | --- |
| `IWasmExecFS.pipe2` JSDoc | Claims Go wasm `os.Pipe()` calls it | States the real `ENOSYS` boundary |
| Go wasm `os.Pipe()` in a custom host | Returns `ENOSYS` | Unchanged; documented, and guarded by a regression so a future upstream change is noticed deliberately |
| First-party stdout/stderr capture | MemFS temp files | Unchanged |
| `fs.pipe2` from JavaScript, correct endpoints | Queues, drains, wakes blocked readers | Preserved if the method stays supported |
| `fs.pipe2` write after the read end closes | Reports success and buffers unreachable bytes | Fails, if the method stays supported |
| Pending read when its own read descriptor closes | Never settled | Settled exactly once, if the method stays supported |
| `close` on either end | Deletes only the passed fd's map entry; shared state stays reachable through the other fd | Both endpoint mappings and buffers released when the lifecycle permits |
| `MemFSError` errno table | No `EPIPE` | `EPIPE` mapped, if the method stays supported |
| External JavaScript consumers of `fs.pipe2` | Unknown | Enumerated before any removal |

## Approach

The invariant: a published type's documentation must describe what the implementation can actually do for the caller it names. The architectural owner of the contract is `packages/wasm/src/structures/IWasmExecFS.ts`; the owner of the behavior is `packages/wasm/src/createMemFS.ts`.

Correct the contract first, in this issue. Then choose one coherent end state rather than keeping the current middle:

1. If no supported consumer exists after the external check, plan removal of the `pipe2` member and the pipe state in a compatible window, and keep the documented statement that Go wasm `os.Pipe` is unsupported under the pinned toolchain.
2. If direct JavaScript callers are supported, document `pipe2` as a JavaScript-only API and make it correct: read-end closure state, a mapped `EPIPE` for writes after the reader is gone, pending reads settled or cancelled exactly once, both endpoints and their buffers released, and flag validation.

Do not keep the present state, which pays the maintenance cost of pipe emulation with neither a real Go caller nor correct direct-call semantics.

A small Go/wasm regression asserting that `os.Pipe` returns `ENOSYS` belongs in either direction, so that a future upstream toolchain that gains pipes is detected deliberately instead of guessed at.

- The published `IWasmExecFS.pipe2` documentation no longer claims Go's wasm `os.Pipe()` calls it, and names the temp-file capture path as the real mechanism.
- A Go/wasm test asserts `os.Pipe()` returns an `ENOSYS`-bearing error under the pinned toolchain.
- The external-consumer question is answered in writing before any public member is removed; if it is removed, the removal lands in a release window that treats it as a public type change.
- If `pipe2` remains supported: writing after the last read end closes fails with `EPIPE`, a read blocked on a closed read end is settled exactly once, closing either end releases both mappings, repeated closes are safe, EOF behavior after the writer closes is preserved, and the currently working queue/drain/wake path stays green.
- Existing MemFS behavior that is already correct stays correct: `rmdir` on a non-empty directory returns `ENOTEMPTY`, `unlink` on a directory returns `EISDIR`, directory rename moves the whole subtree, `truncate` growth zero-fills, `symlink`/`link` return `EPERM`.

Narrow proving commands: `pnpm --filter @ttsc/wasm build:ts` and the `@ttsc/test-wasm` MemFS cases. Broader: `pnpm test:go` and `pnpm test`.

## Coordination

- Two sibling issues also edit `packages/wasm/src/createMemFS.ts`: one for MemFS tree invariants across `writeFile`, `O_CREAT`, and `O_TRUNC`, and one for descriptor open/write semantics. All three must be sequenced by one owner rather than run in parallel.
- New regressions belong in `tests/test-wasm`, which currently has no CI lane. A separate issue owns giving that suite a canonical CI path; until it lands, a green local run is not CI signal.
- Closed #671 and merged PR #693 own MemFS mutation invariants; PR #708 owns invocation-output ownership; PR #87 is the capture fix that chose MemFS temp files. None of them owns the pipe contract, and searches across open and closed issues and pull requests for `pipe2`, wasm broken pipe, MemFS pipe close, and `EPIPE` return nothing.

### LINKED ISSUE #811: fix(wasm): keep MemFS a tree across `writeFile`, `O_CREAT`, and `O_TRUNC`
## Problem

`@ttsc/wasm`'s MemFS reports success for file-creating and file-replacing operations that leave a state which cannot be a filesystem tree:

- a file can be created below an ancestor that is already a file;
- an existing directory, including a non-empty one, can be replaced by a file while every descendant stays in the node map;
- the root directory can be replaced by a file while root descendants remain;
- `open(..., O_CREAT)` silently creates the entire missing ancestor chain, which the low-level `open` contract does not promise;
- `open(..., O_TRUNC)` reproduces the directory replacement.

After each of these, `exists` and `stat` report both the file ancestor and its descendants, while `readdir` on the ancestor fails with `ENOTDIR`. The caller received neither the requested coherent mutation nor an error, so it cannot roll back or retry.

This contradicts the published contract. `website/src/content/docs/wasm/index.mdx` states: "Mutations enforce POSIX-style invariants: renaming a directory moves its whole subtree (and follows open descriptors), `rmdir` requires an empty directory, `unlink` refuses directories, and `truncate`/`ftruncate` resize file bytes, zero-filling any growth, or reject an invalid path, descriptor, or length. A mutation that cannot satisfy its contract returns an error instead of partially changing the tree."

Affected: every browser or custom Go/wasm host that seeds a project through `host.writeFile` or lets the compiler create output through `fs.open`. A conflicting or malformed file map poisons a long-lived host, after which compiler resolution, later writes, removals, and open descriptors disagree about who owns a path prefix. The implementation is platform-independent TypeScript.

### Reproduction matrix, public API

Reviewer probe against the compiled `@ttsc/wasm` module, with the audited TypeScript source confirmed byte-identical first:

| Operation | Callback result | Observable state afterward |
| --- | --- | --- |
| `writeFile("/parent")`, then `writeFile("/parent/child.ts")` | Second write succeeds | `/parent` is a file and `/parent/child.ts` is a file; `readdir("/parent")` returns `ENOTDIR` |
| `mkdirp("/dir/child")`, then `writeFile("/dir")` | Write succeeds | `/dir` is a file while `/dir/child` remains a directory; `readdir("/dir")` returns `ENOTDIR` |
| File `/base`, then `open("/base/generated.js", O_CREAT \| O_WRONLY)` | Returns fd 100 | Both file nodes exist; `readdir("/base")` returns `ENOTDIR` |
| Directory tree `/tree/leaf`, then `open("/tree", O_TRUNC \| O_WRONLY)` | Returns fd 100 | `/tree` becomes a file while `/tree/leaf` remains; `readdir("/tree")` returns `ENOTDIR` |
| File `/keep.ts`, then `writeFile("/")` | Write succeeds | `/` becomes a file while `/keep.ts` remains; `readdir("/")` returns `ENOTDIR` |

Lead reproduction of the child-below-a-file case, after `writeFile('/parent', 'x')` and `fs.open('/parent/child', O_CREAT | O_WRONLY, ...)`:

```json
{"error":null,"fd":100,"childExists":true,"parentText":"x"}
```

Lead reproduction of the `open` cases, starting from `/dir/child/file.ts`:

```text
F open-directory-trunc {"error":null,"fd":100,"dirIsDirectory":false,"readdirError":"ENOTDIR","orphanStillExists":true}
F open-creates-missing-parents {"fd":103,"parentCreated":true}
```

### Root cause

The state owner is `packages/wasm/src/createMemFS.ts`, where one `Map<string, INode>` represents the whole tree.

- `ensureParentDirs` walks the ancestor chain and creates any missing segment, but its only test on an existing segment is `nodes.has(cursor)`. It never checks that the existing node is a directory.
- `writeFile` calls `ensureParentDirs` and then unconditionally `nodes.set(norm, file)`. It does not reject a directory target, inspect descendants, or protect `/`.
- `open` repeats `ensureParentDirs` for the `O_CREAT` branch, and its `O_TRUNC` branch unconditionally replaces whatever node is at the path with an empty file. It allocates the descriptor before applying `O_TRUNC`, and performs no node-kind or ancestor validation at any point.

The same file already contains the guard these paths lack. `mkdirp`, further down in `createMemFS.ts`, walks the identical ancestor chain and rejects a non-directory segment:

```ts
      } else if (existing.kind !== "dir") {
        throw new MemFSError("ENOTDIR", "mkdir", cursor);
      }
```

`rename`, `rmdir`, `unlink`, `truncate`, and `ftruncate` likewise carry kind, root, and children guards (`ENOTDIR`, `EISDIR`, `ENOTEMPTY`). The map is clearly intended to remain a tree; the writer and creation paths bypass the invariant every sibling enforces.

### Provenance, and why this is a whack-a-mole recurrence

Closed issue #671 defined exactly this public invariant: a successful callback must leave a complete tree, and it named orphaned descendants and mutually disagreeing `exists`, `stat`, and `readdir` as the failure to eliminate. Merged PR #693 added central helpers for child detection, parent validation, root protection, and atomic mutation, added the documentation sentence quoted above, and added nine MemFS regressions.

It applied those helpers to `rename`, `rmdir`, `unlink`, `truncate`, and `ftruncate` only. The primary public writer and the `open` create and truncate branches kept their original unconditional `nodes.set` flow, so the exact invalid state #671 set out to prohibit stays reachable through the most-used entry point in the package. The classification is proven from purpose (the closed issue's own statement of the invariant), control flow (helpers exist and are not called on these paths), consequence (the prohibited state reproduces), and history (blame shows `ensureParentDirs` and `writeFile` still originate in the initial MemFS implementation and were untouched by the repair).

`tests/test-wasm/src/features/memfs` contains nine cases covering rename, rmdir, unlink, truncate, and ftruncate. None covers a file-valued ancestor, a directory target, the root, or `open`'s create and truncate branches.

Verified against the audited commit `2b724664eee71c57cd688c55122cac137774bf62`; `packages/wasm/src` is byte-identical on current `master` (`612abc2d2`), where only `packages/wasm/package.json` differs by a version bump.

## Consequence surface

| Path state before the mutation | Required writer/create behavior |
| --- | --- |
| Missing ordinary ancestor, `host.writeFile` or `host.mkdirp` | Preserve the documented convenience: create the missing directories, then the file |
| Missing ancestor, low-level `fs.open` | Fail; parent creation is documented as belonging to `writeFile`/`mkdirp`, not `open` |
| Existing directory ancestor | Create or update the descendant normally |
| Existing file ancestor | Fail before any mutation; never create a descendant below a file |
| Existing regular-file target | `writeFile` overwrite and a valid `O_TRUNC` may replace bytes, preserving file identity |
| Existing empty directory target | Reject file replacement with a type error and keep the directory |
| Existing non-empty directory target | Reject without deleting or orphaning any descendant |
| Root target `/` | Reject; the root stays a directory |
| Normalized aliases with repeated separators, `.`, or `..` | Apply the same checks to the canonical path |
| Any rejected operation | No node, no byte, and no descriptor changed |

Downstream consumers of the same host include documented custom browser hosts, Go's `wasm_exec.js` filesystem surface, playground project rewrites, dependency mounting, source-pack mounting, and transformed compiler output.

## Approach

The invariant: a successful MemFS mutation leaves a valid tree, and an operation that cannot produce one returns an error without changing anything.

Concretely, these four properties must hold after every successful call:

1. `/` is always a directory.
2. Every proper ancestor of a node exists and is a directory.
3. A file has no descendants.
4. A rejected operation makes no node, byte, or descriptor change.

The architectural owner is `packages/wasm/src/createMemFS.ts`. Route every file-creating or file-replacing entry point through one tree-aware preflight that runs before ancestors are auto-created and before a descriptor is allocated, so a late type failure cannot leave partial directories or a leaked fd. Reuse the existing `parentDir`, `hasChildren`, normalization, and `MemFSError` conventions rather than layering consumer-specific checks in the playground, which should surface the host error rather than duplicate filesystem rules.

Do not broaden this into full Node `fs` compatibility. Symlinks, hardlinks, ownership, and permission modeling are intentionally unsupported and stay unsupported.

Positive, unchanged behavior:

- ordinary nested `host.writeFile` still creates missing directories and the file;
- overwriting an existing regular file still replaces its bytes;
- `open` with `O_CREAT` inside an existing directory still creates the file;
- `O_TRUNC` on a regular file still empties it;
- the existing rename, rmdir, unlink, truncate, and ftruncate regressions stay green.

Negative:

- `writeFile` below a file ancestor fails with `ENOTDIR` and creates nothing;
- `writeFile` onto an empty directory, a non-empty directory, and `/` all fail and change nothing;
- `open` with `O_CREAT` below a file ancestor fails, allocates no descriptor, and creates no ancestor;
- `open` with `O_TRUNC` on a directory fails and preserves the directory and every descendant;
- `open` with `O_CREAT` on a path whose parent directory is missing fails without creating the chain.

Boundary and regression: normalized aliases reach the same decisions as their canonical paths; after every rejected call, `exists`, `stat`, `readdir`, `readFile`, and descriptor behavior all agree with the pre-call state; a conflicting dependency or source-pack map surfaced through a playground mounting caller reports the error instead of poisoning the resident host.

Narrow proving commands: `pnpm --filter @ttsc/wasm build:ts` and `pnpm --filter @ttsc/test-wasm start`, with one case per file under `tests/test-wasm/src/features/memfs` following the existing naming convention. Broader: `pnpm test`.

## Coordination

- **This issue and the MemFS descriptor-semantics issue were split on purpose and must be sequenced, not parallelized.** They edit the same file and the same `open` function: this one owns tree and path ownership (ancestors, directory targets, root protection, orphan prevention, implicit parent creation), and the sibling owns descriptor semantics (`O_EXCL`, `O_DIRECTORY`, access and append modes, offsets, positioned writes, bad-descriptor errors). Whichever lands first must not make the other's cases harder to add, and the second must not re-open the first's `O_TRUNC`-on-a-directory hole.
- A third issue corrects the `pipe2` compatibility contract in the same file. Include it in the same sequencing.
- New regressions belong in `tests/test-wasm`, which has no CI lane today. A separate issue owns giving that suite a canonical CI path; until it lands, a green local run is not CI signal.
- Closed #671 and merged PR #693 are the prior work. This is their unsealed remainder, not a duplicate: their operation lists, code changes, and regressions omit `writeFile`, `O_CREAT`, and `O_TRUNC`. Prefer a follow-up over reopening the merged operation-specific patch.

### LINKED ISSUE #813: fix(wasm): make MemFS descriptors honor open modes, offsets, and bad-descriptor errors
## Problem

`@ttsc/wasm` publishes `IWasmExecFS` as a Node-style filesystem for Go's wasm runtime, and `createMemFS` implements it. Its descriptor layer reports success while establishing a different state from the one requested:

- a positioned write at offset 0 appends instead of overwriting;
- every other explicit position is rejected with `ESPIPE`, including in-bounds ones;
- a write with `position: null` ignores the descriptor's tracked cursor and appends;
- access mode is not retained, so a read-only descriptor accepts writes and a write-only descriptor accepts reads;
- `O_APPEND` is not retained, so append is only ever a coincidence of the append-always implementation;
- `O_EXCL` on an existing path returns a descriptor instead of `EEXIST`;
- `O_DIRECTORY` on a regular file returns a descriptor;
- an unknown or closed descriptor is routed to the captured stderr buffer and reported as a successful write.

Callers therefore continue after a `null` error with file contents that differ from what they wrote, or with a descriptor whose mode means nothing. The last case is worse than a wrong write: it silently contaminates the host's captured stderr, which is a diagnostic channel consumers read.

Affected: every custom Go/wasm host and plugin that uses `os.File.Write`, `WriteAt`, access flags, multiple descriptors for one file, or exclusive creation. The default compiler path hides it, because compiler output is normally created or truncated first and then written sequentially, so appending and a zero cursor coincide.

### Positioned and cursor writes

Reviewer probe against the compiled `@ttsc/wasm` module, opening `/f.txt` containing `abcdef` with `O_RDWR` and writing `Z` at position 0, then `Z` at position 1, then in a second host reading two bytes and writing `Q` at `position: null`, then writing three bytes to descriptor 999:

```json
{"n0":1,"after0":"abcdefZ","position1":"ESPIPE: write","afterCursor":"abcdefQ","invalidN":3,"stderr":"BAD","logged":"[wasm] writeSync to unknown fd 999 (3 bytes); routed to stderr buffer"}
```

Expected: `Zbcdef` for the first, success overwriting byte one for the second, `abQdef` for the third, and `EBADF` with zero bytes and no stderr mutation for the fourth.

Lead reproduction of the same class on a three-byte file `abc`, writing `X` at explicit position 0 through a writable descriptor:

```json
{"writeError":null,"n":1,"text":"abcX"}
```

### Open flags and access modes

Reviewer probe opening an existing file with `O_CREAT | O_EXCL`, opening that file with `O_DIRECTORY`, opening `/dir` with `O_TRUNC | O_RDWR`, writing through a read-only descriptor, and reading through an `O_WRONLY` descriptor:

```json
{"exclusive":{"fd":100},"directoryFlag":{"fd":101},"truncateDirectory":{"fd":102,"dirIsNowFile":"","childStillExists":true},"writeReadOnly":{"n":1,"text":"abcZ"},"readWriteOnly":{"n":1,"text":"a"}}
```

Every operation succeeded. Lead reproduction of the same two flag cases against the audited sources:

```text
F open-excl-existing {"fd":101}
F open-directory-flag-on-file {"fd":102}
```

These flags are not foreign constants that MemFS never claimed to support. `createMemFS` publishes them itself through the `fs.constants` object it hands to the Go runtime: `O_WRONLY`, `O_RDWR`, `O_CREAT`, `O_TRUNC`, `O_APPEND`, `O_EXCL`, `O_DIRECTORY`.

### Root cause

The owner is `packages/wasm/src/createMemFS.ts`.

- The descriptor table stores `{ path, position, isStdout?, isStderr? }`. Nothing records the access mode or the append flag, so no code below it can enforce either. The table lacks the state its own public contract needs.
- `writeSync` for any regular-file descriptor builds a new buffer of `existing + incoming` and sets `entry.position` to the new end. It is append-only by construction and can never overwrite.
- The callback `write` rejects `position !== null && position !== 0` with `ESPIPE`, and forwards both `null` and `0` to that same append implementation. The tracked cursor is never consulted.
- When `fdTable` has no entry, `writeSync` falls through to appending the bytes to the captured stderr buffer, emits a `console.error`, and returns `buf.length`, so the caller sees a successful write.
- `open` decides only two things: create the file when it is missing and `O_CREAT` is set, and replace the node with an empty file when `O_TRUNC` is set. The descriptor is allocated between those two steps, unconditionally, and `O_EXCL`, `O_DIRECTORY`, node kind, and access mode are never consulted at all.

`read` on a regular file has the mirror problem: it validates the descriptor's existence and the node kind but never the access mode.

The intentional special case is the stdout and stderr routing for descriptors 1 and 2, which is correct and must be preserved. The current unknown-descriptor fallback is much broader than that special case.

### Provenance, and why this is a whack-a-mole recurrence

Closed issue #671 defined the public false-success mutation family for MemFS and explicitly named descriptor bookkeeping and overwrite behavior among its consequences. Merged PR #693 repaired `rename`, `rmdir`, `unlink`, `truncate`, and `ftruncate` and added nine regressions, and PR #708 only updated a comment on the write branch to say it serves ordinary virtual files and explicit outputs. The open, write, and descriptor branches kept their original behavior, so the same false-success cause remains in the sibling operations of the ones that were fixed, in the same file, with no negative or boundary coverage. The classification is proven from purpose (the closed issue's own consequence list), control flow (the repaired helpers are not reachable from these branches), consequence (the false successes reproduce), and history (blame ties the write and open skeleton to the original MemFS implementation, untouched by the repair).

`tests/test-wasm/src/features/memfs` has no case mentioning `O_EXCL`, `O_DIRECTORY`, `O_APPEND`, a positioned write, an access mode, or an invalid descriptor.

Verified against the audited commit `2b724664eee71c57cd688c55122cac137774bf62`; `packages/wasm/src` is byte-identical on current `master` (`612abc2d2`), where only `packages/wasm/package.json` differs by a version bump.

## Consequence surface

| Operation or boundary | Current behavior | Required invariant |
| --- | --- | --- |
| New empty `O_TRUNC` output, sequential writes | Works because append and a zero cursor coincide | Preserve this common compiler-output path |
| Existing file, `position: null` | Always appends | Write at the descriptor cursor and advance it |
| Existing file after a read or seek | Ignores the tracked cursor and appends | Overwrite at the current cursor |
| Explicit position 0 | Appends and moves the cursor to the end | Overwrite at byte 0 without moving the cursor |
| Explicit in-bounds non-zero position | `ESPIPE` | Overwrite at that position without moving the cursor |
| Explicit position beyond EOF | `ESPIPE` | Extend and zero-fill the gap before the written bytes |
| `O_APPEND` | Appends only by coincidence; the flag is discarded | Append because the flag was retained |
| Multiple descriptors on one file | Cursors are tracked but never used by writes | Independent cursors over shared file data |
| After `ftruncate` or a descriptor-preserving rename | Path and bytes update, the next write still appends | The surviving cursor and flags apply to the updated node |
| Unknown or closed descriptor | Reports success and appends to captured stderr | `EBADF`, zero bytes, no state, stdout, or stderr change |
| Read-only descriptor write, write-only descriptor read | Both succeed | Enforce the access mode captured at `open` |
| `O_CREAT \| O_EXCL` on an existing path | Returns a descriptor | `EEXIST`, no descriptor |
| `O_DIRECTORY` on a regular file | Returns a descriptor | Rejected |
| Descriptors 1 and 2 | Append to the capture buffers | Unchanged |
| Pipe descriptors | Handled by the pipe branch | Unchanged by this issue |

The implementation is platform-independent TypeScript, so no operating system protects any consumer from it.

## Approach

The invariant: a MemFS callback that reports success must mean the requested descriptor operation was actually established, with the bytes, offset, and access rules the caller asked for. An unsupported combination must fail explicitly rather than succeed differently.

The architectural owner is `packages/wasm/src/createMemFS.ts`. The descriptor record must own access mode, append state, and the current offset, because no correct write or read decision can be derived without them. `open` must validate node kind, the supported advertised flags, and access mode before allocating a descriptor or mutating a node. Writes must select between an explicit position and the descriptor cursor correctly, zero-fill extensions, and update the cursor only where the contract requires it. Unknown, closed, or inappropriate descriptors must fail without touching file data or the captured stdio.

Preserve the two behaviors that are deliberate: descriptors 1 and 2 routing to the capture buffers, and the working sequential compiler-output path.

Define explicitly, in the issue's own acceptance, which Node and Go semantics are supported, so "unsupported" is a documented answer rather than an accidental one. Do not expand this into full Node `fs` compatibility: links, ownership, and permission modeling stay unsupported.

Positive:

- write at explicit position 0 on `abc` yields `Xbc`, and the descriptor cursor is unchanged;
- write at an in-bounds non-zero position overwrites there without moving the cursor;
- write with `position: null` after reading two bytes writes at the cursor and advances it;
- a write beyond EOF zero-fills the gap;
- `O_APPEND` writes at the end regardless of the cursor;
- two descriptors on one file keep independent cursors over shared data;
- a descriptor that survives `ftruncate` or a rename applies its cursor and flags to the updated node.

Negative:

- a read-only descriptor rejects writes; a write-only descriptor rejects reads;
- `O_CREAT | O_EXCL` on an existing path returns `EEXIST` and allocates no descriptor;
- `O_DIRECTORY` on a regular file is rejected;
- an unknown or closed descriptor returns `EBADF` with zero bytes, and neither file data nor the captured stdout and stderr change.

Boundary and regression:

- descriptors 1 and 2 still capture into `host.stdout.buffer` and `host.stderr.buffer`;
- a fresh, empty, sequentially written compiler output is byte-identical to today;
- every open rejection is atomic: no node change and no leaked descriptor;
- the existing rename, rmdir, unlink, truncate, and ftruncate regressions stay green.

Narrow proving commands: `pnpm --filter @ttsc/wasm build:ts` and `pnpm --filter @ttsc/test-wasm start`, with one case per file under `tests/test-wasm/src/features/memfs`. A Go-side host test that drives the same operations through the real wasm bridge is the stronger form, since the probes above exercise the JavaScript callback surface directly. Broader: `pnpm test:go` and `pnpm test`.

## Coordination

- **This issue and the MemFS tree-invariant issue were split on purpose and must be sequenced, not parallelized.** They edit the same file and the same `open` function. The tree issue owns ancestors, directory targets, root protection, orphan prevention, and implicit parent creation, including `O_TRUNC` applied to a directory. This issue owns descriptor semantics: `O_EXCL`, `O_DIRECTORY`, access and append modes, offsets, positioned writes, and bad-descriptor errors. Whichever lands second must not re-open the other's cases, and in particular the descriptor validation added here must not bypass the tree preflight added there.
- A third issue corrects the `pipe2` compatibility contract in the same file. Include it in the same sequencing, and leave the pipe branch's current queue, drain, and wake behavior alone here.
- New regressions belong in `tests/test-wasm`, which has no CI lane today. A separate issue owns giving that suite a canonical CI path; until it lands, a green local run is not CI signal.
- Closed #671 and merged PR #693 are the prior work. This is their unsealed remainder, not a duplicate: they named descriptor bookkeeping and overwrite behavior but changed neither.

samchon:
Campaign batch **B4 — the WASM MemFS host**. One branch, one pull request, three sequenced issues that all edit `packages/wasm/src/createMemFS.ts` and its `open` function.

## Scope

| Issue | Owns | Landing order |
| --- | --- | --- |
| #811 | The **tree**: `writeFile`, `O_CREAT`, `O_TRUNC` must never leave the node map describing something that is not a tree — file-valued ancestors, directory targets, root protection, orphan prevention, implicit parent creation. | first |
| #813 | The **descriptor**: open modes, `O_EXCL`, `O_DIRECTORY`, append state, offsets, positioned writes, and bad-descriptor errors. | second, on top of #811 |
| #808 | The **contract**: `IWasmExecFS.pipe2`'s JSDoc claims Go wasm `os.Pipe()` calls it, which is false under the pinned Go toolchain. | with the above |

#811 and #813 were split on purpose and are sequenced rather than parallelized: they edit the same function, and the second must not re-open the first's `O_TRUNC`-on-a-directory hole.

For #808 only the **accepted half** is in scope — correcting the false Go-compatibility contract. Deleting the `pipe2` member and its pipe state is explicitly deferred pending an enumeration of external JavaScript consumers of `fs.pipe2`; this pull request does not decide it.

## Campaign conditions

- `pnpm format` is **not** run on this branch; Post-Campaign Cleanup owns the repository-wide formatter result.
- Repository Actions and workflow settings are unchanged. Every push to this branch passes an exact-SHA cancellation gate, and the run states are recorded in the campaign knowledge base.
- `tests/test-wasm` has no CI lane today (a separate published issue owns that gap), so a green run of this suite is local verification, not CI signal.

## Verification

Pending. Narrow: `pnpm --filter @ttsc/wasm build:ts` and `pnpm --filter @ttsc/test-wasm start`. Broader as the batch requires.

Closes #811
Closes #813
Closes #808

### DIFF
--- a/packages/wasm/src/createMemFS.ts
+++ b/packages/wasm/src/createMemFS.ts
@@ -72,19 +72,52 @@ export function createMemFS(): IMemFSHost {
   const stdout = { buffer: "" };
   const stderr = { buffer: "" };
 
-  const fdTable = new Map<
-    number,
-    { path: string; position: number; isStdout?: boolean; isStderr?: boolean }
-  >();
+  /**
+   * One open descriptor.
+   *
+   * `readable`, `writable`, and `append` are the access mode and `O_APPEND`
+   * flag captured at `open`; without them no read or write below can tell an
+   * allowed operation from a forbidden one. `position` is the cursor a
+   * `position: null` read or write uses and advances, exactly as Node's `fs`
+   * defines it.
+   */
+  interface IDescriptor {
+    path: string;
+    position: number;
+    readable: boolean;
+    writable: boolean;
+    append: boolean;
+    isStdout?: boolean;
+    isStderr?: boolean;
+  }
+  const fdTable = new Map<number, IDescriptor>();
   let nextFd = 100;
   // Reserve 1/2 for stdout/stderr writeSync routing.
-  fdTable.set(1, { path: "/dev/stdout", position: 0, isStdout: true });
-  fdTable.set(2, { path: "/dev/stderr", position: 0, isStderr: true });
+  fdTable.set(1, {
+    path: "/dev/stdout",
+    position: 0,
+    readable: false,
+    writable: true,
+    append: true,
+    isStdout: true,
+  });
+  fdTable.set(2, {
+    path: "/dev/stderr",
+    position: 0,
+    readable: false,
+    writable: true,
+    append: true,
+    isStderr: true,
+  });
 
   // Pipe state. fs.pipe2 mints a pair of fds backed by a shared queue;
   // writes append, reads consume. The state is keyed by fd so a single Map
   // lookup in read/write/close can detect "this is a pipe end" without
   // changing the existing fdTable entries.
+  //
+  // These fds only ever come from a direct JavaScript `fs.pipe2` call. Go's
+  // wasm `os.Pipe` returns ENOSYS without crossing the `globalThis.fs` bridge,
+  // so the Go runtime never reaches this state — see IWasmExecFS.pipe2.
   interface IPipeState {
     buffers: Uint8Array[];
     pendingReaders: Array<{
@@ -143,21 +176,55 @@ export function createMemFS(): IMemFSHost {
     }
   }
 
-  /** Silently create any missing ancestor directories for path `p`. */
-  function ensureParentDirs(p: string): void {
-    const segments = normalize(p).split("/").filter(Boolean);
+  /**
+   * Validate the ancestor chain of normalized path `norm` and report the
+   * ancestors that do not exist yet, root-first.
+   *
+   * An ancestor that exists as a file makes `norm` impossible: a file has no
+   * descendants, so creating one below it would leave a node map that is not a
+   * tree. Validation is deliberately separate from creation so a caller can
+   * reject before touching anything, and so a rejected operation never leaves
+   * half a directory chain behind.
+   */
+  function missingAncestors(norm: string, syscall: string): string[] {
+    const segments = norm.split("/").filter(Boolean);
     segments.pop();
+    const missing: string[] = [];
     let cursor = "";
     for (const seg of segments) {
       cursor += "/" + seg;
-      if (!nodes.has(cursor)) {
-        nodes.set(cursor, {
-          kind: "dir",
-          data: new Uint8Array(),
-          mtimeMs: Date.now(),
-        });
-      }
+      const existing = nodes.get(cursor);
+      if (!existing) missing.push(cursor);
+      else if (existing.kind !== "dir")
+        throw new MemFSError("ENOTDIR", syscall, cursor);
     }
+    return missing;
+  }
+
+  /** Materialize the ancestor directories `missingAncestors` reported. */
+  function createDirs(paths: string[]): void {
+    for (const path of paths)
+      nodes.set(path, {
+        kind: "dir",
+        data: new Uint8Array(),
+        mtimeMs: Date.now(),
+      });
+  }
+
+  /**
+   * Validate that normalized path `norm` may hold a regular file, returning the
+   * node already there when one exists.
+   *
+   * A directory is never silently replaced by a file — including the root,
+   * which is a directory node like any other. POSIX answers `EISDIR`, and
+   * overwriting the node here would strand every descendant in the map behind a
+   * path `readdir` can no longer walk.
+   */
+  function assertFileTarget(norm: string, syscall: string): INode | undefined {
+    const existing = nodes.get(norm);
+    if (existing && existing.kind !== "file")
+      throw new MemFSError("EISDIR", syscall, norm);
+    return existing;
   }
 
   /** Absolute parent directory of a normalized path (`/` for a top-level path). */
@@ -183,6 +250,39 @@ export function createMemFS(): IMemFSHost {
     const next = new Uint8Array(length);
     next.set(data.subarray(0, Math.min(length, data.byteLength)));
     return next;
+  }
+
+  /**
+   * Write `view` through descriptor `entry` and return the bytes stored.
+   *
+   * Three offsets are possible and only one of them is right per call: an
+   * `O_APPEND` descriptor always writes at end-of-file, an explicit `position`
+   * writes exactly there without disturbing the cursor (POSIX `pwrite`), and
+   * `position: null` writes at the cursor and advances it. A write that starts
+   * past end-of-file zero-fills the gap rather than silently relocating.
+   */
+  function writeThroughDescriptor(
+    entry: IDescriptor,
+    view: Uint8Array,
+    position: number | null,
+    syscall: string,
+  ): number {
+    if (!entry.writable) throw new MemFSError("EBADF", syscall, entry.path);
+    const node = nodes.get(entry.path);
+    if (!node) throw new MemFSError("ENOENT", syscall, entry.path);
+    if (node.kind !== "file")
+      throw new MemFSError("EISDIR", syscall, entry.path);
+    const start = entry.append
+      ? node.data.byteLength
+      : (position ?? entry.position);
+    if (!Number.isInteger(start) || start < 0)
+      throw new MemFSError("EINVAL", syscall, entry.path);
+    const end = start + view.byteLength;
+    if (end > node.data.byteLength) node.data = resizeFileData(node.data, end);
+    node.data.set(view, start);
+    node.mtimeMs = Date.now();
+    if (position === null || entry.append) entry.position = end;
+    return view.byteLength;
   }
 
   /**
@@ -233,10 +333,19 @@ export function createMemFS(): IMemFSHost {
 
   function writeFile(p: string, data: string | Uint8Array): void {
     const norm = normalize(p);
-    ensureParentDirs(norm);
+    // Resolve left to right the way POSIX does: an impossible ancestor is
+    // reported before the target, and both are checked before any mutation.
+    const missing = missingAncestors(norm, "open");
+    const existing = assertFileTarget(norm, "open");
     const bytes =
       typeof data === "string" ? encoder.encode(data) : new Uint8Array(data);
-    nodes.set(norm, { kind: "file", data: bytes, mtimeMs: Date.now() });
+    createDirs(missing);
+    // Overwriting keeps the same node so descriptors already pointing at this
+    // file observe the replacement instead of a detached predecessor.
+    if (existing) {
+      existing.data = bytes;
+      existing.mtimeMs = Date.now();
+    } else nodes.set(norm, { kind: "file", data: bytes, mtimeMs: Date.now() });
   }
 
   function readFile(p: string): Uint8Array | null {
@@ -334,40 +443,19 @@ export function createMemFS(): IMemFSHost {
           console.error("[wasm]", line);
         return buf.length;
       }
-      // Open-file fds (>= 100): append to the underlying file. Browser-hosted
-      // tools use this path for ordinary virtual files and explicit outputs.
+      // Open-file fds (>= 100): write at the descriptor cursor, the way Node's
+      // own `writeSync` without a position does. Browser-hosted tools use this
+      // path for ordinary virtual files and explicit outputs.
+      //
+      // An unknown or already-closed fd throws instead of being diverted into
+      // the captured stderr: reporting the byte count of a write nobody
+      // performed made the caller continue on a false success and quietly
+      // contaminated a diagnostic channel consumers read.
       const entry = fdTable.get(fd);
-      if (entry) {
-        const node = nodes.get(entry.path);
-        if (node && node.kind === "file") {
-          // subarray(0) is a zero-copy view over the full incoming buffer; the
-          // bytes are copied into `next` before writeSync returns.
-          const incoming = buf.subarray(0);
-          const existing = node.data;
-          const next = new Uint8Array(
-            existing.byteLength + incoming.byteLength,
-          );
-          next.set(existing, 0);
-          next.set(incoming, existing.byteLength);
-          node.data = next;
-          node.mtimeMs = Date.now();
-          entry.position = next.byteLength;
-          return incoming.byteLength;
-        }
-      }
-      // Unknown fd. Fall back to the stderr buffer so the bytes aren't lost
-      // entirely (and surface as a console.error so the regression is
-      // visible to whoever's looking).
-      stderr.buffer += decoder.decode(buf);
-      // eslint-disable-next-line no-console
-      console.error(
-        "[wasm] writeSync to unknown fd " +
-          fd +
-          " (" +
-          buf.byteLength +
-          " bytes); routed to stderr buffer",
-      );
-      return buf.length;
+      if (!entry) throw new MemFSError("EBADF", "write");
+      // subarray(0) is a zero-copy view over the full incoming buffer; the
+      // bytes are copied into the node before writeSync returns.
+      return writeThroughDescriptor(entry, buf.subarray(0), null, "write");
     },
 
     write(fd, buf, offset, length, position, callback) {
@@ -390,44 +478,82 @@ export function createMemFS(): IMemFSHost {
           callback(null, length);
           return;
         }
-        if (position !== null && position !== 0) {
-          callback(new MemFSError("ESPIPE", "write"), 0);
+        const view = buf.subarray(offset, offset + length);
+        // The stdout/stderr capture buffers are streams, not files: they have
+        // no seekable offset, so an explicit position is meaningless on them.
+        if (fd === 1 || fd === 2) {
+          if (position !== null && position !== 0)
+            throw new MemFSError("ESPIPE", "write");
+          callback(null, this.writeSync(fd, view));
           return;
         }
-        const view = buf.subarray(offset, offset + length);
-        const written = this.writeSync(fd, view);
-        callback(null, written);
+        const entry = fdTable.get(fd);
+        if (!entry) throw new MemFSError("EBADF", "write");
+        callback(null, writeThroughDescriptor(entry, view, position, "write"));
       } catch (err) {
         callback(err as NodeJS.ErrnoException, 0);
       }
     },
 
+    // Every rejection happens before the first mutation and before a
+    // descriptor is minted, so a refused open leaves neither a half-created
+    // node nor a leaked fd behind.
     open(p, flags, _mode, callback) {
-      const norm = normalize(p);
-      const node = nodes.get(norm);
-      const creating = (flags & (this.constants.O_CREAT ?? 0)) !== 0;
-      if (!node) {
-        if (!creating) {
-          callback(new MemFSError("ENOENT", "open", norm), -1);
-          return;
+      try {
+        const norm = normalize(p);
+        const creating = (flags & (this.constants.O_CREAT ?? 0)) !== 0;
+        const exclusive = (flags & (this.constants.O_EXCL ?? 0)) !== 0;
+        const truncating = (flags & (this.constants.O_TRUNC ?? 0)) !== 0;
+        const appending = (flags & (this.constants.O_APPEND ?? 0)) !== 0;
+        const directoryOnly = (flags & (this.constants.O_DIRECTORY ?? 0)) !== 0;
+        // The access mode is the low bits of `flags`: absent both means
+        // read-only, and `O_RDWR` alongside `O_WRONLY` still grants both.
+        const writeOnly = (flags & (this.constants.O_WRONLY ?? 0)) !== 0;
+        const readWrite = (flags & (this.constants.O_RDWR ?? 0)) !== 0;
+        const writable = writeOnly || readWrite;
+        const readable = !writeOnly || readWrite;
+
+        let node = nodes.get(norm);
+        if (!node) {
+          if (!creating) throw new MemFSError("ENOENT", "open", norm);
+          // `open` can only ever create a regular file, so a caller demanding a
+          // directory cannot be satisfied by creating one.
+          if (directoryOnly) throw new MemFSError("ENOTDIR", "open", norm);
+          // Creating the missing ancestor chain is the documented job of
+          // `writeFile` and `mkdirp`; the low-level `open` never promised it.
+          const missing = missingAncestors(norm, "open");
+          if (missing.length > 0)
+            throw new MemFSError("ENOENT", "open", missing[0]!);
+          node = { kind: "file", data: new Uint8Array(), mtimeMs: Date.now() };
+          nodes.set(norm, node);
+        } else {
+          if (creating && exclusive)
+            throw new MemFSError("EEXIST", "open", norm);
+          if (node.kind === "dir") {
+            // A directory opens read-only — Go stats the fd and lists the path
+            // that way. Writing to it or truncating it would turn it into a
+            // file and orphan every descendant.
+            if (writable || truncating)
+              throw new MemFSError("EISDIR", "open", norm);
+          } else if (directoryOnly)
+            throw new MemFSError("ENOTDIR", "open", norm);
         }
-        ensureParentDirs(norm);
-        nodes.set(norm, {
-          kind: "file",
-          data: new Uint8Array(),
-          mtimeMs: Date.now(),
+        if (truncating) {
+          node.data = new Uint8Array();
+          node.mtimeMs = Date.now();
+        }
+        const fd = nextFd++;
+        fdTable.set(fd, {
+          path: norm,
+          position: 0,
+          readable,
+          writable,
+          append: appending,
         });
+        callback(null, fd);
+      } catch (err) {
+        callback(err as NodeJS.ErrnoException, -1);
       }
-      const fd = nextFd++;
-      fdTable.set(fd, { path: norm, position: 0 });
-      if ((flags & (this.constants.O_TRUNC ?? 0)) !== 0) {
-        nodes.set(norm, {
-          kind: "file",
-          data: new Uint8Array(),
-          mtimeMs: Date.now(),
-        });
-      }
-      callback(null, fd);
     },
 
     close(fd, callback) {
@@ -475,16 +601,29 @@ export function createMemFS(): IMemFSHost {
         callback(new MemFSError("EBADF", "read"), 0);
         return;
       }
+      // A write-only descriptor is not a readable one. POSIX answers EBADF for
+      // an operation the descriptor's access mode never granted.
+      if (!entry.readable) {
+        callback(new MemFSError("EBADF", "read", entry.path), 0);
+        return;
+      }
       const node = nodes.get(entry.path);
       if (!node || node.kind !== "file") {
         callback(new MemFSError("ENOENT", "read", entry.path), 0);
         return;
       }
       const start = position ?? entry.position;
+      if (!Number.isInteger(start) || start < 0) {
+        callback(new MemFSError("EINVAL", "read", entry.path), 0);
+        return;
+      }
       const end = Math.min(start + length, node.data.byteLength);
       const slice = node.data.subarray(start, end);
       buffer.set(slice, offset);
-      if (position === null) entry.position = end;
+      // The cursor advances by the bytes actually read. Assigning `end` would
+      // rewind a cursor already past end-of-file (after `ftruncate`, say) back
+      // onto live bytes and make the next sequential write overwrite them.
+      if (position === null) entry.position = start + slice.byteLength;
       callback(null, slice.byteLength);
     },
 
@@ -521,9 +660,9 @@ export function createMemFS(): IMemFSHost {
     },
 
     fstat(fd, callback) {
-      // fstat against a pipe end returns a synthetic file-stat. Go's
-      // os.Pipe-backed File uses fstat at construction time to populate
-      // Stat_t; without this it errors and falls back to invalid fds.
+      // fstat against a pipe end returns a synthetic file-stat so a direct
+      // JavaScript caller that wraps the fd in a file-like abstraction can
+      // populate its stat fields. A pipe end has no node in the tree.
       if (pipes.has(fd)) {
         callback(
           null,
--- a/packages/wasm/src/createMemFS.ts
+++ b/packages/wasm/src/createMemFS.ts
@@ -64,6 +64,12 @@ function normalize(p: string): string {
  * `globalThis.fs` before loading `wasm_exec.js`) and convenience helpers
  * (`writeFile`, `readFile`, `mkdirp`, …) for seeding source files and reading
  * compiler output without touching the real filesystem.
+ *
+ * Every successful mutation leaves a valid tree: `/` stays a directory, each
+ * proper ancestor of a node exists and is a directory, and a file has no
+ * descendants. An operation that cannot satisfy that throws (`writeFile`,
+ * `mkdirp`) or reports a POSIX error through its callback, having changed no
+ * node, byte, or descriptor.
  */
 export function createMemFS(): IMemFSHost {
   const nodes = new Map<string, INode>();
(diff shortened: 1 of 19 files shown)


samchon:
## Self-Review, round 1 — findings and remediation plan

Complete round over the base-to-head diff (`master...HEAD`), the whole of `createMemFS.ts` including untouched methods, both consumers of `IMemFSHost.writeFile` in `packages/playground`, the Go `syscall/fs_js.go` bridge in the pinned toolchain, typescript-go's `osvfs` write path, and the website page. Three findings; two are changes, one is a recorded non-change.

### 1. A zero-length write past end-of-file extends the file — apply

`writeThroughDescriptor` computes `end = start + view.byteLength` and resizes whenever `end > node.data.byteLength`. With a cursor already past EOF (reachable after `ftruncate`, which is exactly the state the new `test_memfs_descriptor_state_survives_ftruncate_and_rename` sets up) a zero-byte write resizes the file to the cursor and zero-fills the gap. POSIX `write()` with a count of 0 returns 0 and changes nothing. Go reaches this whenever `os.File.Write` is handed an empty slice.

Remediation: return 0 immediately once the descriptor and node are validated, before any resize or cursor update, so a write of nothing changes nothing. Pin it with a boundary assertion in the ftruncate/rename case, which already owns a cursor past EOF.

### 2. `mkdirp` still mutates partially on a rejected call — apply

This one is the same cause #811 names, in the sibling entry point the issue's reproduction matrix does not list. `mkdirp` creates each segment as it walks and throws `ENOTDIR` when it meets a file, so `mkdirp("/a/b/c")` with `/a/b` a file leaves `/a` behind. #811's fourth stated property is "a rejected operation makes no node, byte, or descriptor change", and leaving one entry point able to violate it is the whack-a-mole the development skill forbids — especially now that `writeFile` and `open` route through a validate-then-create helper that `mkdirp` could share.

Remediation: hoist the ancestor walk into one `pathChain` + `missingDirs` pair and have all three callers use it, `mkdirp` over the full chain and the two file writers over the chain minus the leaf. This removes a helper rather than adding one, and it makes `mkdirp` strictly less destructive than before. Add a negative case pinning that a rejected `mkdirp` creates no prefix.

Note for the record: GNU `mkdir -p` does leave the successfully created prefix behind, so this is deliberately stricter than the shell. MemFS's published contract is the stricter one — "a mutation that cannot satisfy its contract returns an error instead of partially changing the tree" — and no caller depends on a partial chain.

### 3. `open` allocating fd -1 on rejection, and every other atomicity path — verified, no change

Traced each rejection in `open`: the only node mutation before the descriptor is minted is the `O_CREAT` creation, and nothing between that point and `fdTable.set` can throw, so there is no state in which a node exists and the caller got an error. `nextFd` is only incremented on the success path, so a rejected open cannot burn a descriptor number either. The `O_TRUNC` block is unreachable for a directory because the directory branch throws first.

Also confirmed unchanged and correct: descriptors 1 and 2 route to the capture buffers; a directory still opens read-only, which is what Go's `syscall.Open` needs (it `fstat`s the fd and then `readdir`s the path); and typescript-go's `writeFileEnsuringDir` retries through `MkdirAll` after any failed open, so removing implicit parent creation from `open` cannot break compiler output.

samchon:
## Self-Review, round 2 — findings and remediation plan

Fresh complete round over the current head, re-reading `createMemFS.ts` end to end rather than only the round-1 delta, plus every acceptance bullet of all three issues checked against an actual case rather than against my memory of writing one. One correction to round 1, one coverage gap, one non-change.

### 1. Correction to round 1's finding 2 — the `mkdirp` partial mutation was not reachable

I described `mkdirp` leaving a created prefix behind as a live partial-mutation bug. On re-derivation it is not reachable once #811 lands. For `mkdirp` to create a prefix before failing, some segment must be missing while a *deeper* segment exists as a file — and a file whose ancestors do not exist is precisely the state the tree invariant now forbids. In a valid tree the first non-directory segment always has all its ancestors present, so nothing new is ever created before the throw.

The change stands, but as a simplification and a structural guarantee, not a bug fix: `mkdirp` now shares one validate-then-create walk with `writeFile` and `open` instead of duplicating it, so the property holds by construction rather than by depending on a global invariant. The commit message and the new case's doc comment both say this; recording it here so the thread does not carry the overstated round-1 claim. The zero-byte-write finding was real and is unaffected.

### 2. `link`, `symlink`, and `readlink` are named in #808's acceptance and have no case — apply

#808's "existing MemFS behavior that is already correct stays correct" list names five behaviors. Four are covered by pre-existing regressions (`rmdir` `ENOTEMPTY`, `unlink` `EISDIR`, directory rename moving the subtree, `truncate` growth zero-filling). The fifth, `symlink`/`link` returning `EPERM`, has no coverage anywhere in `tests/test-wasm` — grepping the suite for `EPERM`, `symlink`, or `readlink` returns nothing.

These sit next to the paths this batch tightened and are exactly the operations that would produce an incoherent tree if one ever started succeeding, so the gap matters more after this change than before it. Remediation: add a case pinning `link` (on a file and on a directory), `symlink`, and `readlink`, and that no node appears as a result.

### 3. Full acceptance-matrix sweep — verified, no further change

Walked every positive, negative, and boundary bullet of #811 and #813 to a specific assertion, including the ones easy to assume: normalized aliases reaching the same decision as their canonical path (three cases), a rejected operation leaving `exists`/`stat`/`readdir`/`readFile` agreeing with the pre-call state, a fresh sequentially written compiler output staying byte-identical, and both reserved stdio descriptors still capturing. Every bullet has a case.

Also re-derived the remaining `open` orderings and confirmed each answer is deliberate: `open("/missing/x", O_DIRECTORY)` without `O_CREAT` answers `ENOENT` rather than the ancestor's `ENOTDIR`, because the path genuinely does not exist and the ancestor is not consulted before the existence check; `O_CREAT | O_DIRECTORY` on a free path answers `ENOTDIR` without creating anything, rather than reproducing Linux's create-then-fail quirk, which would violate the atomicity property #811 requires; and an access mode of `O_WRONLY | O_RDWR` is read as granting both rather than rejected, since the intent is unambiguous and no in-tree caller produces it.

`fs.pipe2` and its state remain untouched, including the write-after-read-close case, which still reports success by design — that half of #808 is the lead's to decide.

samchon:
## Self-Review, round 3 — clean round, with end-to-end evidence

Fresh complete round over the current head. No finding survived verification, so this is the terminating round. What it covered beyond re-reading the diff:

**The real wasm, driven through the new MemFS.** The batch's largest untested risk was the Go bridge, since `open` no longer creates parent directories and now enforces an access mode — and `tests/test-wasm` only exercises the JavaScript callback surface. So I built the actual binary (`pnpm --filter @ttsc/wasm build:wasm`, ~8 min, `go1.26.4` → `js/wasm`, 34 MB) and drove it from Node: install `createMemFS().fs` as `globalThis.fs`, load the staged `wasm_exec.js`, instantiate, seed `/work` with a `tsconfig.json` and three sources, and call `api.build`, `api.check`, then rewrite a source in place on the resident host and rebuild — the playground's exact pattern.

Then I did the same against `master`'s `createMemFS` compiled into a scratch directory, same wasm binary, same fixture. Both runs:

```
build code: 2 outputs: [ 'dist/bad.js', 'dist/index.js' ]
diagnostics: [ "2322 Type 'string' is not assignable to type 'number'." ]
check code: 2
tree intact: {"src":true,"work":true,"stderr":""}
rebuild code: 0 diagnostics: 0
```

`diff` of the two logs is empty apart from JS object key enumeration order in the Go→JS marshalling, which is not stable between runs. The compiler resolves directories, reads every source, reports the real type error, emits, and recompiles cleanly after an in-place rewrite — identically before and after. That closes the Go-bridge question with evidence rather than with source reading.

**Downstream consumer.** `@ttsc/playground` is the only in-repo consumer of `IMemFSHost.writeFile`, through `installDependenciesIntoMemFS`, `installTypiaSourcePack`, and `createWorkerCompilerService`. It typechecks clean and `@ttsc/test-playground` is 7/7 green. Its `runBuildPipeline` already wraps the seeding call in a try/catch that surfaces the message, which is the behavior #811 asks for: the host reports a conflicting file map instead of the playground duplicating filesystem rules.

**Removed-behavior sweep.** `git grep` for `routed to stderr`, `writeSync to unknown`, and `ensureParentDirs` finds no consumer that depended on the deleted unknown-fd fallback or the removed helper — the only hits are two of the new doc comments naming the pre-fix symbol as history.

**Two things observed and deliberately left alone.** `open` now calls its success `callback` inside the `try`, so a throwing callback would be invoked twice; that is the established shape of `readdir`, `mkdir`, `stat`, and `write` in this file already, and changing only `open` would make it inconsistent while changing all five is outside the batch. And `close` on one pipe end still leaves the other end's mapping, which #808 lists under the deferred half.

### Verification recorded for the merge gate

| Command | Result |
| --- | --- |
| `pnpm --filter @ttsc/wasm build:ts` | clean |
| `pnpm --filter @ttsc/test-wasm start` | 29/29 pass (15 pre-existing, 14 new) |
| `pnpm --filter @ttsc/playground build` | clean |
| `pnpm --filter @ttsc/test-playground start` | 7/7 pass |
| `tsc --noEmit -p tests/test-wasm/tsconfig.json` | clean |
| `go build ./...` and `go vet ./...` in `packages/wasm` (native) | clean |
| `GOOS=js GOARCH=wasm go vet ./test/host/` | clean |
| `GOOS=js GOARCH=wasm go test -exec="node $(go env GOROOT)/lib/wasm/wasm_exec_node.js" ./test/host/` | 2/2 pass |
| `pnpm --filter @ttsc/wasm build:wasm` + real-wasm A/B against `master`'s MemFS | identical observable results |

**State of the suite as found, before any change:** all 15 pre-existing `tests/test-wasm` cases passed on Node 22.21.0 / Windows. That is worth stating explicitly because this suite has no CI lane (#806), so nothing had ever proven it green — it was, and it still is.

`pnpm test` and `pnpm test:go` were not run. They are out of the batch's package scope, cannot complete on this Windows checkout, and no non-test Go source changed: the only Go file added is `//go:build js && wasm`, which every native lane excludes, as the clean native `go build ./...` and `go vet ./...` above confirm.

Issue #784, opened 2026-07-19
fix(graph): report dispatch-hub omissions as trace truncation


samchon:
## Problem

An open trace suppresses implementation dispatch when a bodyless declaration has at least the dispatch-hub threshold of eligible implementations. The helper returns an empty list, so the caller cannot distinguish “no dispatch fact” from “eligible hops omitted by the hub bound” and returns `truncated: false`.

The trace contract says truncation is true whenever an eligible node or hop is omitted by a bound. Path mode already preserves target-aware dispatch and can report omitted continuations; open mode loses that signal.

## Acceptance

- Dispatch selection returns both selected and omitted candidates.
- Open trace sets `truncated: true` when at least one in-scope implementation hop is suppressed.
- A fanout consisting only of excluded external endpoints remains `truncated: false` unless `includeExternal` is enabled.
- Preserve target-aware path behavior.
- Add positive/negative regressions at and around the hub threshold and keep graph consumer parity synchronized (samchon/graph#105).

samchon:
A repository-wide audit just published several issues that touch `dispatchEdges` alongside this one, so here is the landing order they were written against, and why.

**Order: #828 first, then this issue, then #829.**

#828 shows that `hasExecutionBody` decides whether a declaration has a body from its outgoing-edge degree — one `calls`/`instantiates`/`accesses`/`renders` edge — rather than from declaration facts. That predicate selects the very candidate set this issue counts against `DISPATCH_HUB`, in both directions: an implementation whose body names no dependency is excluded from the set, and a concrete base with no dependency edge is admitted to it. Building the truncation signal on top of that selection would pin a count taken over the wrong candidates, and the regressions added here would then encode it.

#829 adds the reverse and impact direction, which has no dispatch synthesis at all today. It should inherit whatever truncation contract this issue settles, rather than growing a second one.

If this issue is already claimed and in progress, nothing here is wasted — #828 still lands first and this rebases onto it, because the fix here is the reporting of an omission and the fix there is which candidates exist to omit.

Related from the same audit: #809 covers the tour filter that erases a whole high-fan-in class, which shares this issue's degree-based notion of a hub (`DISPATCH_HUB`'s own comment cites `isSharedUtility` as "the graph's existing definition of a hub"), so the two are best sequenced under one owner.

Issue #786, opened 2026-07-19
fix(graph): make details completeness honor member caps


samchon:
## Problem

Details accepts an explicit `memberLimit` and truncates declaration, object, or enum members, but the unconditional result audit still says the symbol's members were returned whole. A two-member node requested with `memberLimit: 1` returns one member under a false completeness claim.

## Acceptance

- The audit never claims whole-member coverage after a caller cap.
- Prefer per-node typed truncation/available-count metadata when extending the canonical structure; otherwise make the audit request/result-aware.
- Cover declaration members, object members, enum members, complete controls, and multi-handle Details.
- Synchronize the generalized consumer tracked by samchon/graph#106.

samchon:
A finding from campaign batch B1 (#862) that belongs to this issue's budget accounting rather than to a new issue, recorded so whoever implements the cap knows it exists.

**`details.implementedBy` can list one implementation twice, and each copy consumes the `dependencyLimit` budget.** For `class Impl extends Base implements Base`, the checker yields two member relations for the same pair — one from the heritage `extends` clause and one from `implements` — and the details projection carries both through. The duplicate is arguably informative at the relation level, but it spends a slot of the very budget this issue is about, so a caller with `dependencyLimit: 10` can receive nine distinct implementations and one repeat while the audit reports a complete set.

That interaction is why it is filed here rather than separately: deduplication by implementation identity and honest cap accounting are the same decision made once. If the implementer concludes the duplicate should be preserved as two distinct relations, then the cap must count distinct implementations rather than relations, and say so.

Verified on `master` at `818b181ac`, after #828 merged. The B1 agent found and fixed the same duplication in the dispatch path during its own Self-Review round 2, where the symptom was a repeated hop (`['OnlyOnce.emit', 'OnlyOnce.emit']`); that fix is limited to dispatch edges and does not reach this projection.

PR #862, opened 2026-07-20
fix(graph): decide dispatch from declaration facts, cross it in reverse, and stop calling a bounded path a disconnection

### LINKED ISSUE #827: fix(graph): path mode reports a depth-bounded miss as a proven global disconnection
## Problem

`trace` in path mode (`to` supplied) answers with a statement about the whole graph whenever its bounded search fails, including when the only reason it failed is the depth bound the caller asked for:

> No call path runs from the start to the target and they touch nothing in common, so the graph holds no connection between them.

The response carries `hops: 0`, `truncated: false`, and `next.action: "outside"`, which directs the caller away from the graph entirely. On the same graph, raising the bound returns the full path and `next.action: "answer"`.

Expected: a negative existence claim is made only when the traversal actually exhausted the eligible graph. A traversal stopped by a bound reports a bound and how to continue.

This is the worst failure mode for an index. The caller is told authoritatively that two subsystems are unconnected, so it stops asking and either reads files or concludes that a real dependency does not exist. When the two ends happen to share a symbol, the `junctions` branch fires instead and redirects the caller to a "seam" that is not the seam, because a direct call path exists.

Path mode clamps `maxDepth` to the range 1 to 12 and defaults to 12 when the field is absent. A shortest path of 13 hops therefore cannot be answered by any request: there is no larger value to retry with. `ITtscGraphTrace.IRequest.maxDepth` is documented with `@default 3`, so a caller that carries a value over from an open trace gets the false verdict for any chain of four or more hops.

Affected: every `trace` call with `to` set, in all three focus modes, over the MCP stdio server and in resident and one-shot sessions. `tour` embeds path traces, so it inherits the verdict.

### Reproduction

The probe constructs a straight call chain in memory and calls the shipped `runTrace`, so it needs no compiler and no fixture project. Build the package once, then save the script as `probe.cjs` **in the repository root**, because `require` resolves relative to the script file:

```bash
pnpm --filter @ttsc/graph build
node probe.cjs
```

```js
const { TtscGraphMemory } = require("./packages/graph/lib/model/TtscGraphMemory.js");
const { runTrace } = require("./packages/graph/lib/server/runTrace.js");

function chain(length) {
  const file = "src/app.ts";
  const nodes = Array.from({ length }, (_u, i) => ({
    id: `${file}#n${i}:function`,
    kind: "function",
    name: `n${i}`,
    file,
    external: false,
    evidence: { startLine: i + 1 },
  }));
  return TtscGraphMemory.from({
    project: process.cwd(),
    tsconfig: "tsconfig.json",
    provenance: {
      schemaVersion: 5,
      capabilities: [],
      producer: { tool: "probe", version: "", typescript: "" },
      universe: { configs: [], roots: [] },
      sources: [],
    },
    diagnostics: [],
    nodes,
    edges: Array.from({ length: length - 1 }, (_u, i) => ({
      from: nodes[i].id,
      to: nodes[i + 1].id,
      kind: "calls",
    })),
  });
}

function probe(label, length, maxDepth) {
  const out = runTrace(chain(length), {
    type: "trace",
    from: "n0",
    to: `n${length - 1}`,
    focus: "execution",
    maxDepth,
  });
  console.log(label, JSON.stringify({
    hops: out.result.hops.length,
    truncated: out.result.truncated,
    next: out.next,
  }));
}

probe("A path-13-hops", 14, 12);
probe("B path-12-hops", 13, 12);
probe("C small maxDepth=2", 4, 2);
probe("D small maxDepth=3", 4, 3);
```

Observed output:

```text
A path-13-hops {"hops":0,"truncated":false,"next":{"action":"outside","reason":"No call path runs from the start to the target and they touch nothing in common, so the graph holds no connection between them."}}
B path-12-hops {"hops":12,"truncated":false,"next":{"action":"answer","reason":"The path result is the structural flow: its path nodes and evidence ranges are what the graph holds between the two ends."}}
C small maxDepth=2 {"hops":0,"truncated":false,"next":{"action":"outside","reason":"No call path runs from the start to the target and they touch nothing in common, so the graph holds no connection between them."}}
D small maxDepth=3 {"hops":3,"truncated":false,"next":{"action":"answer","reason":"The path result is the structural flow: its path nodes and evidence ranges are what the graph holds between the two ends."}}
```

A and B are the hard case: at the maximum bound the tool admits, a 13-hop path is unanswerable and is reported as absent, while the 12-hop control succeeds. C and D are the small witness: the identical four-node graph claims global disconnection at `maxDepth: 2` and returns the full path at `maxDepth: 3`.

### Root cause

The owner is `packages/graph/src/server/runTrace.ts`.

`findPath` is a breadth-first walk. It skips expansion with `if (depth >= maxDepth) continue;` and, when the queue empties, executes `return null`. The same `null` therefore encodes two different facts: the eligible search space was exhausted and no path exists under the selected focus and external policy, or the walk was cut off before it could decide. Nothing records whether an eligible frontier remained.

The path branch reads that sentinel as complete knowledge:

```ts
const found = findPath(graph, start.node.id, target.node.id, bound(props.maxDepth, 12, 1, 12), focus, includeExternal);
const hasPath = found !== null;
```

`base` carries `truncated: false` unconditionally, `junctionsBetween` is then computed as though absence were proven, and the `next` verdict is either the junction seam or the global no-connection claim quoted above.

The open trace in the same function already makes the distinction the path branch lacks. `eligibleTraceEndpoint` re-applies the traversal's own focus, external, file-node, and direction policy at the boundary, so `truncated` becomes true only when an eligible hop actually lay beyond it. Seeing any raw outgoing edge is not proof of an incomplete search; cycles and filtered external or file nodes must remain governed by that same eligibility policy.

### Provenance and retrospective classification

Path mode arrived in `5320cb7d57feb52453ceb719d7a6d497a6e5b069`. Two later repairs touched this exact call pair:

- Closed #758, "zero-hop self traces are reported as outside the graph", taught the caller to respect `findPath`'s null sentinel for the self-path case.
- Closed #762, "set `trace.truncated` only when eligible graph data was actually omitted", introduced `eligibleTraceEndpoint` for the open trace.

Both landed in `e92bf171a3fa1fe6cd35fdac8d6f7621ea119a4f`, titled "Make graph trace existence and completeness truthful", merged as PR #777 with the summary "Make trace results distinguish path existence from hop count and report truncation only when an eligible open-trace fact was actually omitted."

That record proves a whack-a-mole repair:

- Purpose. The change made a class-level claim about existence and completeness across the trace tool, not a claim about one witness.
- Control flow. Its diff touched path mode only to replace the `hops.length > 0` existence test with `found !== null`. The unconditional `truncated: false` in `base` and the unqualified verdict were left as they were, so the sentinel still conflates exhaustion with the boundary.
- Consequence. The identical untruth, a configured bound reported as a fact about the graph, survives in the sibling traversal mode of the same function.
- History. Two prior repairs at the same site, both aimed at the same invariant, and the correct rule is already written in that pull request's own commit trailers: "A configured trace boundary is not evidence of omission; apply focus, external, direction, and dispatch policy first."

The existence of a configured bound alone would not prove the classification; the two prior repairs plus the class-level claim do.

Two pieces of counter-evidence belong in the record, and both still bind this fix. PR #777 explicitly recorded "Rejected: Extend path-mode truncation semantics | outside issues #758 and #762; scope the public contract instead", and carries the directive "Preserve exact open-trace completeness semantics when extending traversal modes". The author therefore scoped that change deliberately and narrowed the public `truncated` prose to open traces rather than widening path-mode behavior. That is why the correction belongs inside path mode and must not touch the open trace or the protected field.

### Duplicate check

- #758 (closed) is zero-hop self traces. Different observable.
- #762 (closed) is open-trace `truncated`. Its fix does not touch path mode's `base.truncated` or its verdict.
- #784 (open), "report dispatch-hub omissions as trace truncation", is a different omission source in the open trace's hop set and says nothing about the depth bound or the existence claim.
- Searches for `trace maxDepth`, `path truncated graph`, and the reproduced message found no owning open or closed item.

### Anchors

Verified at commit `2b724664eee71c57cd688c55122cac137774bf62`. `packages/graph/src/server/runTrace.ts` is byte-identical on `master` at `612abc2d20d3f99429319081c1f6520bbe34ac7a`.

| Symbol | File |
| --- | --- |
| Path branch, `findPath` call, `hasPath`, `junctionsBetween`, `next` verdicts | `packages/graph/src/server/runTrace.ts` |
| `findPath`, its depth skip and its `null` return | `packages/graph/src/server/runTrace.ts` |
| `eligibleTraceEndpoint`, the open trace's boundary policy | `packages/graph/src/server/runTrace.ts` |
| `bound(props.maxDepth, 12, 1, 12)` | `packages/graph/src/server/runTrace.ts` |
| `truncated` and `maxDepth` contract prose (protected, do not edit) | `packages/graph/src/structures/ITtscGraphTrace.ts` |
| Open-trace omission regression that must stay green | `tests/test-graph/src/features/test_ttscgraph_trace_reports_only_actual_omissions.ts` |

## Consequence surface

| Situation | Current result | Correct result |
| --- | --- | --- |
| Path shorter than or equal to `maxDepth` | Path, `answer` | Unchanged |
| Path longer than `maxDepth`, no shared junction | `hops: 0`, `truncated: false`, `outside`, "the graph holds no connection between them" | A bound report with actionable continuation |
| Path longer than `maxDepth`, ends share a junction | `inspect` with the callback-seam narration, presented as the way across | Bound report; the seam narration is only true for an exhausted search |
| Genuinely disconnected ends, search exhausted, no junction | `outside` with the same message | Unchanged, and now actually earned |
| Genuinely disconnected ends, search exhausted, shared junction | `inspect` junction verdict | Unchanged |
| `from === to` | Zero hops, valid path, `answer` | Unchanged; this is what #758 fixed |
| Shortest path of 13 or more hops | Unanswerable at every legal `maxDepth`, and reported as absent | Must not be reported as absent; the continuation must be actionable despite the hard cap |
| `focus: "execution"`, `"types"`, `"all"` | All affected, all share `findPath` and `traversable` | All corrected together |
| `includeExternal` true or false, and file nodes | Filtered endpoints must not count as a remaining frontier | Same eligibility policy as the open trace |
| Cyclic graphs | A visited node is not an eligible frontier | Same |
| Paths crossing synthesized `dispatches` edges | Affected, since `findPath` expands them | Same policy |
| `tour` flows that embed a path trace | Inherit the verdict | Inherit the correction |
| Open trace (`to` absent) | Correct today | Must not change |

## Approach

The invariant: a negative existence claim is made only when the traversal exhausted the eligible graph. A traversal stopped by a caller-supplied bound reports the bound, not an absence.

The architectural owner is the path branch of `packages/graph/src/server/runTrace.ts` together with `findPath`. What `findPath` returns today, a nullable path, cannot express the needed answer; the internal result needs three states rather than two: found, exhausted, and bounded with an eligible frontier remaining. "Eligible" must mean the same thing it means for the open trace, so a frontier consisting only of filtered external nodes, file nodes, non-traversable kinds, or already-visited nodes is not a frontier.

Two constraints on the shape of the fix:

- Do not repurpose the public `truncated` field. Its JSDoc is protected and explicitly scoped: "In an open trace, true when a bound omitted an eligible node or hop." Redefining it for path mode would require editing protected prose.
- Do not change the open trace's completeness semantics, and do not change the `junctions` computation or its ranking. What changes is when a junction narration or an `outside` verdict may be presented as the answer.

Because path mode caps `maxDepth` at 12, the bounded response must be actionable on its own. A message that merely invites a larger `maxDepth` is not sufficient for a 13-hop path; the caller needs a continuation it can actually take. The mechanism is the implementer's to choose and is not prescribed here.

| Case | Expected |
| --- | --- |
| Positive: 3-hop path queried with `maxDepth: 2` | No claim that the graph holds no connection; the result says the walk stopped at the requested depth and gives a usable continuation |
| Positive: same query at `maxDepth: 3` | Path and `answer` verdict unchanged |
| Positive: 13-hop path at the maximum `maxDepth: 12` | Not reported as absent, and the continuation is actionable despite the cap |
| Negative: two genuinely unconnected nodes, no shared junction | `outside` and the existing message, unchanged |
| Negative: two unconnected nodes that share a junction | `inspect` junction verdict, unchanged |
| Negative: the only frontier beyond the bound is an excluded external or file node with `includeExternal: false` | Treated as exhausted, not as bounded |
| Boundary: `from === to`; path exactly `maxDepth` hops long; `maxDepth: 1` | All unchanged |
| Boundary: cyclic graph where the frontier is entirely visited nodes | Treated as exhausted |
| Coverage: `focus` of `execution`, `types`, and `all`; `includeExternal` both ways; a path crossing a synthesized `dispatches` edge | Same policy in every combination |
| Regression: open-trace `truncated` behavior pinned by `test_ttscgraph_trace_reports_only_actual_omissions.ts` | Unchanged |
| Regression: `tour` results | Unchanged except where a flow embedded an affected path trace |
| Protected contract | `packages/graph/src/structures/` byte-identical |

Add cases under `tests/test-graph/src/features/` following the repository test contract: one `test_<snake_case>` export per file, named for what it asserts, opened with the three-part doc comment the development skill requires.

Narrow proving commands, in order:

```bash
pnpm run build:current
pnpm --filter @ttsc/graph build
```

Broader: `pnpm test`.

## Coordination

- **Protected Graph contract.** `packages/graph/src/structures/ITtscGraphApplication.ts`, every detailed and public subordinate type under `packages/graph/src/structures/`, and all of their JSDoc are immutable by owner instruction. This fix is implementation and tests only. Before submitting, confirm `git rev-parse HEAD:packages/graph/src/structures` is still `3c7399e3ce9f2404e64d652c7f1259ebd0db2613` and that `ITtscGraphApplication.ts` still hashes to SHA-256 `6AB7451FCD17E8A8C1F0F7BAEAA6A96AA4E6DEDCB379F70F030ECEE64197BE6C` as checked out on Windows (`872a838f01106eafd777de2c8d202d900a21396fce70206d1736fbf77ab6f8d5` for the LF blob content). The file is not Prettier-clean, verified with `npx prettier --check`, so running the repository `format` script silently breaks the constraint. Format only the files you touched.
- Open #784 owns the dispatch-hub truncation signal in the open trace. Its acceptance requires dispatch selection to return selected and omitted candidates and to set `truncated: true` for suppressed in-scope hops, and it states that path mode must keep its target-aware dispatch behavior. Do not extend into that boundary; coordinate if both land close together.
- Two sibling issues change dispatch in the same file: one replaces the outgoing-edge proxy that decides whether a declaration has a body, and one adds the reverse direction of virtual dispatch. Both change which edges `findPath` expands. Their own fixed order is dispatch body detection, then #784, then reverse-direction dispatch. This issue does not touch `dispatchEdges`, so it can be sequenced anywhere in that chain by the same owner; landing it after the dispatch-fact fix is preferable, because the frontier this issue must judge as eligible then already includes the corrected dispatch edges.
- A fourth sibling issue changes `runTour.ts` flow filtering. `runTour` calls `runTrace`, so re-run the graph suite after both land.
- Open #783 (cross-project wire identity) and #786 (details member caps) are separate Graph surfaces with no overlap.

### LINKED ISSUE #828: fix(graph): dispatch decides whether a declaration has a body from outgoing-edge degree
## Problem

A forward `trace` continues a call that landed on a bodyless declaration into the implementations that actually run, emitting a `dispatches` hop. The decision "does this declaration have a body" is made by counting outgoing edges: a node counts as having a body when it has at least one outgoing `calls`, `instantiates`, `accesses`, or `renders` edge.

Having a body and naming a modeled dependency are different facts, and the substitution fails in both directions:

- **A real implementation is dropped.** A concrete implementation of an interface member or an abstract method whose body names no modeled symbol, an empty body, a returned literal, local arithmetic or control flow, or a thrown literal, is judged bodyless and refused as a dispatch target. The trace dead-ends at the declaration and reports `truncated: false`, so the caller is told the flow ends there.
- **A concrete method is promoted through its override.** A concrete base method with a real body is judged bodyless whenever that body names no modeled symbol, so a call to it is continued into an override and the trace reports a `dispatches` hop that the feature was never meant to emit. The same graph answers two ways: two polymorphic setups identical in every declaration fact, the same base class, the same `overrides` relation, the same call site, differ only in one statement inside the base body, and that statement decides whether the override is reached at all.

Expected: whether a declaration is bodyless is read from what the declaration is, using the compiler facts the graph already carries, and the answer does not change because a body's contents happen to reference a symbol the graph models.

The feature's own contract already states the intended rule: "a called declaration with no body continues in the implementations that have one". The protected `TtscGraphEdgeKind` prose states it too: "the checker resolves a call to the declaration it names, and where that declaration is abstract or an interface member, the code that runs is its implementation." Both speak of declaration facts. The helper's own comment, "Whether the declaration has a body: something it calls, reads, or renders", describes the proxy that was implemented instead.

The rule this issue adopts is that contract, stated once so it does not have to be rediscovered during implementation: **dispatch continues only from a genuinely bodyless declaration**, which means an interface member, an abstract method, an ambient or `declare`d declaration, and an overload signature resolved to its implementation. **A concrete declaration that has a body is a real destination and is never promoted through an override**, whatever its dependency degree. Dispatch exists to rescue a call that lands on a declaration with no body, and nothing else.

Whether a concrete base should *additionally* offer its overrides is a separate product question about virtual dispatch semantics, and it is explicitly out of scope here. The silence is deliberate, not an oversight: this issue restores the declared contract, and widening it would be a new feature with its own answer-shape and hub-policy consequences.

Affected: forward open traces and path mode, under `focus: "execution"` and `focus: "all"`, over the MCP stdio server and in resident and one-shot sessions. `focus: "types"` deliberately synthesizes no dispatch. `tour` embeds forward traces and inherits the result. Reverse and impact traversal do not synthesize dispatch at all, which is a separate defect with its own issue.

### Reproduction

Each case builds a graph in memory and calls the shipped `runTrace`, so no compiler and no fixture project are needed. Build the package once, then save the script as `probe.cjs` **in the repository root**, because `require` resolves relative to the script file:

```bash
pnpm --filter @ttsc/graph build
node probe.cjs
```

```js
const { TtscGraphMemory } = require("./packages/graph/lib/model/TtscGraphMemory.js");
const { runTrace } = require("./packages/graph/lib/server/runTrace.js");

const decl = (file, id, kind, name, qualifiedName, line) => ({
  id,
  kind,
  name,
  ...(qualifiedName === undefined ? {} : { qualifiedName }),
  file,
  external: false,
  evidence: { startLine: line },
});

const memory = (nodes, edges) =>
  TtscGraphMemory.from({
    project: process.cwd(),
    tsconfig: "tsconfig.json",
    provenance: {
      schemaVersion: 5,
      capabilities: [],
      producer: { tool: "probe", version: "", typescript: "" },
      universe: { configs: [], roots: [] },
      sources: [],
    },
    diagnostics: [],
    nodes,
    edges,
  });

const show = (label, out) =>
  console.log(label, JSON.stringify({
    reached: out.result.reached.map((n) => n.name),
    hops: out.result.hops.map((h) => h.kind),
    truncated: out.result.truncated,
  }));

// C / C-control: interface member, implementation with and without a modeled dependency.
for (const implementationCallsHelper of [false, true]) {
  const file = "src/app.ts";
  const contract = `${file}#Pipeline.execute:method`;
  const impl = `${file}#Impl.execute:method`;
  const caller = `${file}#run:function`;
  const helper = `${file}#accepted:function`;
  const nodes = [
    decl(file, `${file}#Pipeline:interface`, "interface", "Pipeline", undefined, 1),
    decl(file, contract, "method", "execute", "Pipeline.execute", 2),
    decl(file, `${file}#Impl:class`, "class", "Impl", undefined, 4),
    decl(file, impl, "method", "execute", "Impl.execute", 5),
    decl(file, caller, "function", "run", undefined, 7),
    decl(file, helper, "function", "accepted", undefined, 8),
  ];
  const edges = [
    { from: caller, to: contract, kind: "calls" },
    { from: impl, to: contract, kind: "implements" },
  ];
  if (implementationCallsHelper) edges.push({ from: impl, to: helper, kind: "calls" });
  show(
    implementationCallsHelper ? "C-control implementation calls helper" : "C implementation body empty",
    runTrace(memory(nodes, edges), {
      type: "trace", from: "run", focus: "execution", maxDepth: 4, maxNodes: 8,
    }),
  );
}

// D / D-control: concrete base with an override, base body with and without a modeled dependency.
for (const baseCallsHelper of [false, true]) {
  const file = "src/base.ts";
  const base = `${file}#Base.run:method`;
  const derived = `${file}#Derived.run:method`;
  const caller = `${file}#callBase:function`;
  const derivedHelper = `${file}#derivedOnly:function`;
  const baseHelper = `${file}#baseOnly:function`;
  const nodes = [
    decl(file, `${file}#Base:class`, "class", "Base", undefined, 1),
    decl(file, base, "method", "run", "Base.run", 2),
    decl(file, `${file}#Derived:class`, "class", "Derived", undefined, 4),
    decl(file, derived, "method", "run", "Derived.run", 5),
    decl(file, caller, "function", "callBase", undefined, 7),
    decl(file, derivedHelper, "function", "derivedOnly", undefined, 8),
    decl(file, baseHelper, "function", "baseOnly", undefined, 9),
  ];
  const edges = [
    { from: caller, to: base, kind: "calls" },
    { from: derived, to: base, kind: "overrides" },
    { from: derived, to: derivedHelper, kind: "calls" },
  ];
  if (baseCallsHelper) edges.push({ from: base, to: baseHelper, kind: "calls" });
  show(
    baseCallsHelper ? "D-control base body calls baseOnly" : "D base body empty",
    runTrace(memory(nodes, edges), {
      type: "trace", from: "callBase", focus: "execution", maxDepth: 4, maxNodes: 8,
    }),
  );
}
```

Observed output:

```text
C implementation body empty {"reached":["Pipeline.execute"],"hops":["calls"],"truncated":false}
C-control implementation calls helper {"reached":["Pipeline.execute","Impl.execute","accepted"],"hops":["calls","dispatches","calls"],"truncated":false}
D base body empty {"reached":["Base.run","Derived.run","derivedOnly"],"hops":["calls","dispatches","calls"],"truncated":false}
D-control base body calls baseOnly {"reached":["Base.run","baseOnly"],"hops":["calls","calls"],"truncated":false}
```

C is the dropped implementation. The `implements` relation is present and checker-valid; the trace stops at the interface member, reports `truncated: false`, and never names `Impl.execute`.

D and D-control are the inconsistency. Every declaration fact is identical between them: same base class, same derived class, same `overrides` relation, same call site. The only difference is one statement inside the base's body, and it flips whether the override is reachable at all. Under the rule this issue adopts, both must answer the same way and both must answer "no dispatch": `Base.run` is a concrete declaration with a body, so it is the destination, and D's synthesized `dispatches` hop is the defect rather than D-control's silence.

### Root cause

The owner is `packages/graph/src/server/runTrace.ts`.

`dispatchEdges(graph, id, focus)` returns nothing when `focus === "types"` or when `hasExecutionBody(graph, id)` is true. Otherwise it walks incoming `overrides` and `implements` edges and admits an implementation only when `hasExecutionBody(graph, edge.from)` is true. Its complete definition is:

```ts
function hasExecutionBody(graph: TtscGraphMemory, id: string): boolean {
  for (const edge of graph.outgoing(id))
    if (EXECUTION_KINDS.has(edge.kind)) return true;
  return false;
}
```

`EXECUTION_KINDS` is `calls`, `instantiates`, `accesses`, `renders`. The predicate therefore measures dependency degree, and degree is zero for any body that names nothing the graph models. The native builder derives those edges by walking runtime value uses, while declaration kind, modifiers, and ownership are recorded separately, so the two facts are independently available and simply are not the same fact.

Both traversal sites call `dispatchEdges`: the forward branch of `traceEdges`, and `findPath`'s expansion. The defect therefore reaches open forward traces and path mode alike.

### Why the tests do not catch it

Every existing dispatch regression gives each implementation a helper call, so degree and body coincide in the fixtures:

- `tests/test-graph/src/features/test_ttscgraph_trace_dispatches_to_the_implementation.ts` builds `TransformPipeline.execute` calling `transform()` and `PersistPipeline.execute` calling `persist()`.
- `tests/test-graph/src/features/test_ttscgraph_member_relations_follow_checker_dispatch.ts` builds `Good.execute` calling `accepted()` and `Bad.execute` calling `rejected()`.

No case has an implementation that names nothing, and no case has a concrete base leaf. The suite proves downstream reachability, not body identity.

### Provenance

`hasExecutionBody`, `dispatchEdges`, and the dispatch contract arrived together in `e59e5ec0b0650fd8f1ddccff6918eb311b2915ea`, merged PR #440, "fix(graph): follow the call into the implementation it dispatches to". `865f982debab7235336b4bc670271cfa64a65a36` later made `implements` and `overrides` checker-validated; it did not change body detection, and its accepted implementation also calls a helper.

No strict classification under the development skill's Forbidden section is claimed. This is a semantic proxy that arrived with its original implementation; there is no consumer or fixture special case and no repeated symptom history. The history is still relevant evidence, because the behavior contradicts the contract the same commit wrote down.

### Duplicate check

No open or closed issue or pull request owns this. Searches for `hasExecutionBody`, `leaf implementation graph trace`, and related dispatch and body terms returned nothing. Open #784 owns only the dispatch-hub cap, where twelve or more eligible implementations suppress the whole set; here a single implementation is dropped or a single override is synthesized, independently of any threshold.

### Anchors

Verified at commit `2b724664eee71c57cd688c55122cac137774bf62`. `packages/graph/src/server/runTrace.ts` is byte-identical on `master` at `612abc2d20d3f99429319081c1f6520bbe34ac7a`.

| Symbol | File |
| --- | --- |
| `dispatchEdges`, `hasExecutionBody`, `EXECUTION_KINDS`, `DISPATCH_KINDS`, `DISPATCH_HUB` | `packages/graph/src/server/runTrace.ts` |
| `traceEdges` forward branch and `findPath` expansion, the two dispatch call sites | `packages/graph/src/server/runTrace.ts` |
| Member relation production and modifier emission | `packages/ttsc/internal/graph/edges.go`, `packages/ttsc/internal/graph/build.go` |
| `contains` ownership synthesis | `packages/graph/src/model/TtscGraphMemory.ts` |
| `dispatches` contract prose (protected, do not edit) | `packages/graph/src/structures/TtscGraphEdgeKind.ts` |
| Existing dispatch regressions | `tests/test-graph/src/features/test_ttscgraph_trace_dispatches_to_the_implementation.ts`, `test_ttscgraph_member_relations_follow_checker_dispatch.ts` |

## Consequence surface

| Called declaration | Implementation body | Current result | Correct result |
| --- | --- | --- | --- |
| Interface member | Calls a helper | Dispatches to it, reaches its callees | Unchanged |
| Interface member | Empty | No dispatch, trace dead-ends at the declaration, `truncated: false` | Dispatches to the implementation |
| Interface member | Returns a literal, assigns a local, throws, or is pure control flow | Same dead end | Dispatches |
| Abstract method | Empty or dependency-free | Same dead end | Dispatches |
| Ambient or `declare` member | Dependency-free by construction | Same dead end | Dispatches; bodilessness is decided from the declaration, not degree |
| Concrete base method with an override | Empty | Synthesizes a `dispatches` hop into the override | No dispatch: the base has a body and is itself the destination |
| Concrete base method with an override | Calls one helper | No dispatch; the override is never reached | No dispatch, unchanged, and now for the stated reason rather than by accident |
| Overload signatures with one implementation | Signature has no outgoing edges | Signature is treated as bodyless and the implementation as the target only when it has degree | Resolved through the implementation without duplicate or false dispatch |
| Any of the above, with 12 or more eligible implementations | Any | The hub cap suppresses the whole set; owned by open #784 | Cap behavior is #784's, but it applies to the candidate set this issue corrects |
| Any declaration under `focus: "types"` | Any | No dispatch | Unchanged |
| Any declaration in the reverse or impact direction | Any | No dispatch synthesis at all | Owned by a separate issue |

Every row propagates past the hop itself: whatever the implementation reaches is added to or missing from `reached`, `steps`, path results, and any `tour` flow built on that trace.

## Approach

The invariant: whether a called declaration is bodyless is a compiler fact about the declaration, not a count of its outgoing edges, and two graphs identical in their declaration facts must produce identical dispatch answers.

The architectural owner is `dispatchEdges` and its body predicate in `packages/graph/src/server/runTrace.ts`. Both the gate on the called declaration and the admission test on the candidate implementation read the same predicate, and both are affected.

The facts needed are already on the wire and need no schema change:

- `ITtscGraphNode.kind` distinguishes `interface`, `class`, `method`, `property`, and `function`.
- `ITtscGraphNode.modifiers` carries `abstract` and `declare`, emitted by the native producer from the declaration's modifier flags. An interface member carries no modifier keyword, so its bodilessness comes from ownership rather than from a modifier.
- The `contains` ownership tree, synthesized in `TtscGraphMemory`, names the container of every member, which is how an interface member is recognized.
- `implements` and `overrides` remain the checker-owned candidate set. Do not reintroduce name matching to find implementations.

The rule is settled and is not an implementation choice. Dispatch continues only from a genuinely bodyless declaration: an interface member, an abstract method, an ambient or `declare`d declaration, or an overload signature resolved to its implementation. A concrete declaration with a body is a real destination and is never promoted through an override, so both concrete-base rows in the table above must come out as "no dispatch". Extending dispatch to concrete bases is out of scope, as stated under Problem.

The same rule governs both sides of the decision. The gate on the called declaration asks whether that declaration is bodyless; the admission test on a candidate implementation asks whether that implementation is a concrete declaration with a body. Neither side may consult dependency degree.

Preserve the existing focus, external-node, depth and node bound, ordering, and hub policies. This issue changes which candidates are eligible, not how many survive the bounds.

| Case | Expected |
| --- | --- |
| Positive: interface member whose sole implementation has an empty body | The trace dispatches to the implementation |
| Positive: implementation that returns a constant, uses only locals, or throws | Dispatches |
| Positive: abstract method with a dependency-free implementation | Dispatches |
| Positive: ambient or `declare`d member with a concrete implementation | Dispatches |
| Positive: overload signatures resolved to one implementation | Dispatches to the implementation |
| Positive: the existing implementation-calls-a-helper case | Still reaches the implementation and its callees |
| Negative: concrete base method with an empty body and an override | No `dispatches` hop; the base is the destination |
| Negative: concrete base method whose body calls a helper, with an override | No `dispatches` hop, as today |
| Negative: checker-rejected member relation, the `Bad.execute` shape with an incompatible signature | Still excluded |
| Negative: unrelated same-named member with no checker relation | Still excluded |
| Consistency: two graphs identical except for one statement inside the base body | Identical dispatch answer, and in both cases no dispatch |
| Boundary: overload signatures resolving through one implementation | No duplicate and no false dispatch |
| Boundary: hub threshold behavior at 11, 12, and 13 eligible implementations | Unchanged from whatever open #784 lands |
| Coverage: open forward trace and path mode | Agree with each other |
| Coverage: `focus: "types"` | Unchanged |
| Regression: existing dispatch and member-relation suites | Pass |
| Protected contract | `packages/graph/src/structures/` byte-identical |

Write the new cases end to end through the real dump and MCP path under `tests/test-graph/src/features/`, following the repository test contract: one `test_<snake_case>` export per file, named for what it asserts, opened with the three-part doc comment the development skill requires. An in-memory server-level fixture proves the branch but not that the native producer emits the modifiers and relations the fix reads.

Narrow proving commands, in order:

```bash
pnpm run build:current
pnpm --filter @ttsc/graph build
```

Broader: `pnpm test`. If any native producer change is needed, also run `pnpm test:go`.

## Coordination

- **Protected Graph contract.** `packages/graph/src/structures/ITtscGraphApplication.ts`, every detailed and public subordinate type under `packages/graph/src/structures/`, and all of their JSDoc are immutable by owner instruction. The whole structures tree must stay byte-identical, so no node field, edge kind, or dump shape may be added; the facts listed under Approach are already present. If an implementation seems to need a new field, that is a signal to re-examine the approach. Before submitting, confirm `git rev-parse HEAD:packages/graph/src/structures` is still `3c7399e3ce9f2404e64d652c7f1259ebd0db2613` and that `ITtscGraphApplication.ts` still hashes to SHA-256 `6AB7451FCD17E8A8C1F0F7BAEAA6A96AA4E6DEDCB379F70F030ECEE64197BE6C` as checked out on Windows (`872a838f01106eafd777de2c8d202d900a21396fce70206d1736fbf77ab6f8d5` for the LF blob content). The file is not Prettier-clean, verified with `npx prettier --check`, so running the repository `format` script silently breaks the constraint. Format only the files you touched.
- **Landing order: this issue first, then #784, then the reverse-dispatch issue.** All three change `dispatchEdges` in `runTrace.ts` and must be sequenced by one owner rather than landed in parallel. This issue fixes the fact the other two build on: #784 reports how the bound on the candidate set is communicated, and reverse dispatch inherits both the body decision and the truncation signal. Synthesizing reverse dispatch on top of the current degree proxy would duplicate this defect into a second direction, and the hub truncation signal should exist before reverse traversal starts inheriting it.
- Open #784, "report dispatch-hub omissions as trace truncation", changes the same function to return selected and omitted candidates and to set `truncated: true` for suppressed in-scope hops. If #784 is already claimed by an implementer, this issue still lands first and #784 rebases onto it, because #784's candidate set is exactly what this issue corrects.
- A sibling issue fixes path mode reporting a depth-bounded miss as a global disconnection, in the same file. A fourth changes `runTour.ts` flow filtering and calls `runTrace` for every flow. One owner should sequence all four.
- Open #783 (cross-project wire identity) and #786 (details member caps) are separate Graph surfaces with no overlap.

### LINKED ISSUE #829: fix(graph): reverse and impact traces never cross verified virtual dispatch
## Problem

A forward `trace` that lands on an interface member or an abstract method continues into the implementation that runs, emitting a `dispatches` hop. The reverse direction has no counterpart. `reverse` and `impact` traces starting at a concrete implementation reach neither the declaration it implements nor any caller resolved to that declaration.

For the ordinary shape where `Runner.run` calls `Pipeline.execute` and `Good.execute` implements it, an impact query on `Good.execute` returns `reached: []`, `hops: []`, `truncated: false`. The graph holds every fact needed to answer: a forward trace from `Runner.run` on the same graph reaches `Pipeline.execute`, `Good.execute`, and beyond.

Expected: a change to an implementation reaches, in reverse, the declaration it implements and everything that calls that declaration, under the same eligibility and bound policy the forward direction uses.

`ITtscGraphTrace.IRequest.direction` documents `impact` as a "reverse trace prioritizing public API and test nodes a change reaches". An impact answer that omits the caller side of every polymorphic seam understates the blast radius while reporting `truncated: false`, so the caller has no signal that anything was left out. This is the query an agent runs before changing a method.

Affected: `direction: "reverse"` and `direction: "impact"` in every focus mode, over the MCP stdio server and in resident and one-shot sessions, for every project with an interface, an abstract base, a strategy, an adapter, or a visitor.

### Reproduction

The probe constructs the graph in memory and calls the shipped `runTrace`, so no compiler and no fixture project are needed. Build the package once, then save the script as `probe.cjs` **in the repository root**, because `require` resolves relative to the script file:

```bash
pnpm --filter @ttsc/graph build
node probe.cjs
```

```js
const { TtscGraphMemory } = require("./packages/graph/lib/model/TtscGraphMemory.js");
const { runTrace } = require("./packages/graph/lib/server/runTrace.js");

const file = "src/app.ts";
const decl = (id, kind, name, qualifiedName, line) => ({
  id,
  kind,
  name,
  ...(qualifiedName === undefined ? {} : { qualifiedName }),
  file,
  external: false,
  exported: true,
  evidence: { startLine: line },
});

const contract = `${file}#Pipeline.execute:method`;
const impl = `${file}#Good.execute:method`;
const runner = `${file}#Runner.run:method`;
const helper = `${file}#accepted:function`;
const entry = `${file}#main:function`;

const graph = TtscGraphMemory.from({
  project: process.cwd(),
  tsconfig: "tsconfig.json",
  provenance: {
    schemaVersion: 5,
    capabilities: [],
    producer: { tool: "probe", version: "", typescript: "" },
    universe: { configs: [], roots: [] },
    sources: [],
  },
  diagnostics: [],
  nodes: [
    decl(`${file}#Pipeline:interface`, "interface", "Pipeline", undefined, 1),
    decl(contract, "method", "execute", "Pipeline.execute", 2),
    decl(`${file}#Good:class`, "class", "Good", undefined, 4),
    decl(impl, "method", "execute", "Good.execute", 5),
    decl(`${file}#Runner:class`, "class", "Runner", undefined, 8),
    decl(runner, "method", "run", "Runner.run", 9),
    decl(helper, "function", "accepted", undefined, 12),
    decl(entry, "function", "main", undefined, 14),
  ],
  edges: [
    { from: runner, to: contract, kind: "calls" },
    { from: impl, to: contract, kind: "implements" },
    { from: impl, to: helper, kind: "calls" },
    { from: entry, to: runner, kind: "calls" },
  ],
});

const show = (label, out) =>
  console.log(label, JSON.stringify({
    reached: out.result.reached.map((n) => n.name),
    hops: out.result.hops.map((h) => h.kind),
    truncated: out.result.truncated,
  }));

for (const [direction, focus] of [
  ["reverse", "execution"], ["reverse", "all"],
  ["impact", "execution"], ["impact", "all"],
]) {
  show(`${direction} focus=${focus}`, runTrace(graph, {
    type: "trace", from: "Good.execute", direction, focus, maxDepth: 4, maxNodes: 12,
  }));
}

show("control reverse-from-base", runTrace(graph, {
  type: "trace", from: "Pipeline.execute", direction: "reverse", focus: "execution", maxDepth: 4, maxNodes: 12,
}));
show("control forward-from-caller", runTrace(graph, {
  type: "trace", from: "Runner.run", direction: "forward", focus: "execution", maxDepth: 4, maxNodes: 12,
}));
```

Observed output:

```text
reverse focus=execution {"reached":[],"hops":[],"truncated":false}
reverse focus=all {"reached":[],"hops":[],"truncated":false}
impact focus=execution {"reached":[],"hops":[],"truncated":false}
impact focus=all {"reached":[],"hops":[],"truncated":false}
control reverse-from-base {"reached":["Runner.run","main"],"hops":["calls","calls"]}
control forward-from-caller {"reached":["Pipeline.execute","Good.execute","accepted"],"hops":["calls","dispatches","calls"]}
```

The two controls establish that nothing is missing from the graph. Reverse from the base reaches `Runner.run` and `main`. Forward from the caller crosses the dispatch and reaches the implementation. Only the reverse crossing of the seam is absent, and it is absent for both `reverse` and `impact` under both `execution` and `all` focus.

### Root cause

The owner is `packages/graph/src/server/runTrace.ts`.

`traceEdges` selects candidate edges by direction:

```ts
return reverse
  ? graph.incoming(id)
  : [...graph.outgoing(id), ...dispatchEdges(graph, id, focus)];
```

Dispatch synthesis appears only on the forward side, here and in `findPath`'s expansion. `dispatchEdges` itself works by reading incoming `overrides` and `implements` edges on a bodyless declaration and emitting a synthetic `base -> implementation` edge of kind `dispatches`.

The raw checker relation is oriented implementation to base. From `Good.execute`, the `implements` edge is outgoing, so reverse traversal never considers it. The callers are incoming to `Pipeline.execute`, which reverse traversal cannot arrive at. The two halves of the path are each one step away in a direction the traversal does not take.

`focus: "all"` does not help. Reverse only ever looks at `graph.incoming(id)`, and the implementation has no incoming relation to follow.

### Provenance

Forward dispatch arrived in `e59e5ec0b0650fd8f1ddccff6918eb311b2915ea`, merged PR #440, "fix(graph): follow the call into the implementation it dispatches to". Its stated motivation is that runtime calls land in concrete bodies, and it implemented the forward consequence of that. The reverse consequence was not implemented. `865f982debab7235336b4bc670271cfa64a65a36` later made the member relations checker-validated without changing direction handling.

Existing impact and reverse regressions cover direct callers only. `tests/test-graph/src/features/test_ttscgraph_trace_dispatches_to_the_implementation.ts` and `test_ttscgraph_member_relations_follow_checker_dispatch.ts` both assert forward behavior, including the rejection of a checker-incompatible same-named member, and neither issues a reverse or impact request from an implementation.

No strict classification under the development skill's Forbidden section is claimed. This is a one-directional feature that arrived with its original implementation, with no consumer or fixture special case and no repeated symptom history.

### Duplicate check

No open or closed issue or pull request owns this. Searches for `graph reverse dispatch impact implementation`, `graph impact interface callers`, and `dispatch implementation reverse` returned nothing. Open #784 concerns the forward dispatch-hub cap and its truncation signal; it is coordination, not a duplicate. Open #783 (cross-project wire identity) and #786 (details member caps) are separate Graph surfaces.

### Anchors

Verified at commit `2b724664eee71c57cd688c55122cac137774bf62`. `packages/graph/src/server/runTrace.ts` is byte-identical on `master` at `612abc2d20d3f99429319081c1f6520bbe34ac7a`.

| Symbol | File |
| --- | --- |
| `traceEdges` and its `reverse` branch | `packages/graph/src/server/runTrace.ts` |
| `dispatchEdges`, `hasExecutionBody`, `DISPATCH_KINDS`, `DISPATCH_HUB` | `packages/graph/src/server/runTrace.ts` |
| `eligibleTraceEndpoint`, `orderedEdges`, `traversable` | `packages/graph/src/server/runTrace.ts` |
| `MAX_IMPACT_DEPTH`, `MAX_IMPACT_NODES`, role tagging for impact | `packages/graph/src/server/runTrace.ts` |
| Member relation production | `packages/ttsc/internal/graph/edges.go` |
| `direction` and `dispatches` contract prose (protected, do not edit) | `packages/graph/src/structures/ITtscGraphTrace.ts`, `packages/graph/src/structures/TtscGraphEdgeKind.ts` |

## Consequence surface

| Query | Current result | Correct result |
| --- | --- | --- |
| `reverse` from an implementation of an interface member | Does not reach the interface member or its callers | Reaches both |
| `impact` from an implementation of an interface member | Same omission, with `truncated: false` | Reaches both, and reports omission honestly when a bound applies |
| `reverse` or `impact` from an override of an abstract method | Same omission | Same correction |
| `reverse` or `impact` from a method with no `implements` or `overrides` relation | Correct today | Unchanged |
| `reverse` from the base declaration itself | Correct today, reaches its callers | Unchanged |
| Forward trace and path mode | Cross the seam today | Unchanged |
| `focus: "types"` | Follows `implements` and `overrides` as type edges and synthesizes no dispatch | Unchanged |
| `focus: "execution"` and `"all"` in reverse | Empty across the seam | Corrected |
| Impact role tagging of exported and test nodes | Cannot tag what it never reaches | Tags the newly reached callers |
| External endpoints with `includeExternal: false` | Must stay filtered | Same policy as forward |
| 12 or more implementations of one declaration | Forward suppresses the set; reverse has no dispatch at all | Reverse must apply the same hub policy and report it the same way, coordinated with open #784 |
| Cycles, depth and node bounds, hop ordering | Governed by the shared traversal policy | Same policy, applied to the new edges |

The omission compounds. Everything the base's callers reach in reverse, their own callers, the exported surface, and the tests that exercise them, is also missing, so the impact answer is short by an entire subtree rather than by one hop.

## Approach

The invariant: virtual dispatch is one fact, and both traversal directions must see it. If a forward walk crosses from a declaration to the implementation that runs, a reverse walk from that implementation must cross back to the declaration and continue through its callers.

The architectural owner is `traceEdges` in `packages/graph/src/server/runTrace.ts`, together with `dispatchEdges`. The reverse branch needs the same checker-verified synthetic edge the forward branch already builds, derived from the implementation's own outgoing `overrides` and `implements` relation rather than from a name match.

Constraints on the fix:

- Reuse the existing `dispatches` edge kind and the relation's evidence. Do not invent a new kind, and do not add a field to any structure.
- Apply the same focus, external-node, file-node, depth and node bound, ordering, and hub policies the forward direction applies. The reverse direction must not become more permissive than the forward one.
- Keep checker validity as the gate. A same-named member with an incompatible signature must stay disconnected in reverse exactly as it is in forward.
- Mirror the forward rule for what is dispatchable, do not restate it. Dispatch crosses the seam only where the declaration is genuinely bodyless: an interface member, an abstract method, an ambient or `declare`d declaration, or an overload signature resolved to its implementation. A concrete declaration with a body is a real destination and is never promoted through an override, in either direction. That rule is settled by the dispatch-fact issue that lands first, and this fix must read it from the same place rather than reimplementing it.
- `focus: "types"` already follows `implements` and `overrides` directly and must keep behaving as it does.

| Case | Expected |
| --- | --- |
| Positive: `reverse` from a valid concrete implementation | Reaches the interface member or abstract base, then its callers, then their callers within the depth bound |
| Positive: `impact` from the same node | Same reachability, with exported and test roles tagged on the newly reached nodes |
| Positive: abstract-method override, not only an interface implementation | Same |
| Negative: checker-rejected same-named member | Stays disconnected in reverse, as it is in forward |
| Negative: method with no member relation | Unchanged |
| Negative: external endpoints with `includeExternal: false` | Still filtered |
| Boundary: `maxDepth` and `maxNodes` at and beyond the impact caps | Bounds honored, and omission reported by the same rule the forward direction uses |
| Boundary: 12 or more implementations of one declaration, in reverse | Same hub policy and the same truncation signal as forward, coordinated with open #784 |
| Boundary: cycle through an implementation and its base | Terminates, no duplicate hops |
| Regression: forward trace and path mode | Unchanged |
| Regression: `focus: "types"` reverse | Unchanged |
| Regression: existing direct-caller impact tests | Pass |
| Protected contract | `packages/graph/src/structures/` byte-identical |

Extend the existing `Pipeline` / `Good` / `Bad` / `Runner` fixture with reverse and impact requests from the valid concrete method, asserting the base, the caller, the exported and test roles, and the downstream caller chain. Write the cases end to end under `tests/test-graph/src/features/` following the repository test contract: one `test_<snake_case>` export per file, named for what it asserts, opened with the three-part doc comment the development skill requires.

Narrow proving commands, in order:

```bash
pnpm run build:current
pnpm --filter @ttsc/graph build
```

Broader: `pnpm test`.

## Coordination

- **Protected Graph contract.** `packages/graph/src/structures/ITtscGraphApplication.ts`, every detailed and public subordinate type under `packages/graph/src/structures/`, and all of their JSDoc are immutable by owner instruction. The whole structures tree must stay byte-identical, so the fix uses the existing `dispatches` kind and the existing public `IHop` and `INode` fields. Before submitting, confirm `git rev-parse HEAD:packages/graph/src/structures` is still `3c7399e3ce9f2404e64d652c7f1259ebd0db2613` and that `ITtscGraphApplication.ts` still hashes to SHA-256 `6AB7451FCD17E8A8C1F0F7BAEAA6A96AA4E6DEDCB379F70F030ECEE64197BE6C` as checked out on Windows (`872a838f01106eafd777de2c8d202d900a21396fce70206d1736fbf77ab6f8d5` for the LF blob content). The file is not Prettier-clean, verified with `npx prettier --check`, so running the repository `format` script silently breaks the constraint. Format only the files you touched.
- **Landing order: the dispatch-fact issue first, then #784, then this one.** All three change `dispatchEdges` in `runTrace.ts` and must be sequenced by one owner rather than landed in parallel. The dispatch-fact issue replaces the outgoing-edge proxy that decides whether a declaration has a body, and it settles the rule this fix must mirror: dispatch continues only from a genuinely bodyless declaration, so a concrete declaration with a body is never promoted through an override in either direction. Synthesizing reverse dispatch on top of the current degree proxy would duplicate that defect into a second direction.
- Open #784, "report dispatch-hub omissions as trace truncation", changes `dispatchEdges` to return selected and omitted candidates and to set `truncated: true` for suppressed in-scope hops. It lands second so the truncation signal exists before reverse traversal inherits it; the reverse direction must report a suppressed hub the same way rather than returning a silently complete result. If #784 is already claimed by an implementer, the dispatch-fact issue still lands first and #784 rebases onto it.
- Two further sibling issues change the same file, path-mode depth verdicts, and `runTour.ts` flow filtering, which calls `runTrace`. One owner should sequence all of them in `runTrace.ts`.
- Open #783 (cross-project wire identity) and #786 (details member caps) are separate Graph surfaces with no overlap.

samchon:
Campaign batch **B1 — Graph traversal**. One branch, one pull request, three issues, all in `packages/graph/src/server/runTrace.ts`.

- Closes #828 — dispatch decides "has a body" from outgoing-edge degree instead of declaration facts.
- Closes #829 — reverse and impact traces never cross verified virtual dispatch.
- Closes #827 — path mode reports a depth-bounded miss as a proven global disconnection.

## Landing order inside the batch

1. **#828 first.** It settles the rule the rest of the batch reads: dispatch continues only from a genuinely bodyless declaration (an interface member, an abstract member, an ambient or `declare`d declaration, a declaration-file declaration, an external leaf), and a concrete declaration with a body is a real destination that is never promoted through an override — whatever its dependency degree.
2. **#829 second.** Reverse dispatch mirrors #828's settled rule by reusing the same forward candidate builder rather than reimplementing it, so eligibility, checker validity and the hub cap cannot diverge between directions.
3. **#827 last.** It is independent of dispatch, but it judges the path-mode frontier, and landing it after #828 means the frontier it judges already contains corrected dispatch edges.

## Not in this batch

Published issue #784 (dispatch-hub omissions reported as trace truncation) is **not** implemented here. The `DISPATCH_HUB` cap keeps its current silent-suppression behaviour in both directions.

## Verification

Campaign CI is deliberately suspended; every push passes the exact-SHA cancellation gate and the tables are recorded in the campaign knowledge base. Local, package-scoped verification is recorded in the thread.

## Protected contract

`packages/graph/src/structures/` stays byte-identical. No structure, field, edge kind or JSDoc is touched, and `pnpm format` is not run on this branch.

### DIFF
--- a/packages/graph/src/server/pathPolicy.ts
+++ b/packages/graph/src/server/pathPolicy.ts
@@ -9,6 +9,15 @@ export function isExternalNode(node: ITtscGraphNode): boolean {
   );
 }
 
+/**
+ * True for a `.d.ts` declaration file. Every declaration such a file holds is
+ * ambient by construction, so none of them carries a body the graph can walk
+ * into, whether or not the `declare` keyword is written out.
+ */
+export function isDeclarationFile(file: string): boolean {
+  return /\.d\.[cm]?ts$/.test(file);
+}
+
 /** True for tests, examples, fixtures, generated output, and build artifacts. */
 export function isSupportPath(file: string): boolean {
   return (
@@ -20,7 +29,7 @@ export function isSupportPath(file: string): boolean {
     ) ||
     /\.(test|spec)\.[cm]?tsx?$/.test(file) ||
     /(^|\/)typings\.[cm]?ts$/.test(file) ||
-    /\.d\.[cm]?ts$/.test(file) ||
+    isDeclarationFile(file) ||
     /(^|\/)(dist|build|coverage|generated|__generated__)\//.test(file)
   );
 }
--- a/packages/graph/src/server/runTrace.ts
+++ b/packages/graph/src/server/runTrace.ts
@@ -753,13 +753,22 @@ function dispatchEdges(
   if (declaration === undefined || hasDeclarationBody(graph, declaration))
     return [];
   const out: ITtscGraphEdge[] = [];
+  // Per implementation, not per relation. A class may name one base in two
+  // heritage clauses — `class Impl extends Base implements Base` is legal — and
+  // the producer records the member pair once per clause, as `overrides` and as
+  // `implements`. That is one implementation and one crossing: emitting it
+  // twice would repeat the hop and count the same class twice against the hub
+  // cut.
+  const dispatched = new Set<string>();
   for (const edge of relations) {
+    if (dispatched.has(edge.from)) continue;
     const implementation = graph.node(edge.from);
     if (
       implementation === undefined ||
       !hasDeclarationBody(graph, implementation)
     )
       continue;
+    dispatched.add(edge.from);
     out.push({
       from: id,
       to: edge.from,
--- a/packages/graph/src/server/runTrace.ts
+++ b/packages/graph/src/server/runTrace.ts
@@ -741,12 +741,19 @@ function dispatchEdges(
   focus: ITtscGraphTrace.IRequest["focus"],
 ): ITtscGraphEdge[] {
   if (focus === "types") return [];
+  // The checker relations first: almost no node has one, and reading a
+  // declaration's own facts walks its ownership chain, which is work worth
+  // doing only where there is something to dispatch to.
+  let relations: ITtscGraphEdge[] | undefined;
+  for (const edge of graph.incoming(id)) {
+    if (DISPATCH_KINDS.has(edge.kind)) (relations ??= []).push(edge);
+  }
+  if (relations === undefined) return [];
   const declaration = graph.node(id);
   if (declaration === undefined || hasDeclarationBody(graph, declaration))
     return [];
   const out: ITtscGraphEdge[] = [];
-  for (const edge of graph.incoming(id)) {
-    if (!DISPATCH_KINDS.has(edge.kind)) continue;
+  for (const edge of relations) {
     const implementation = graph.node(edge.from);
     if (
       implementation === undefined ||
--- a/packages/graph/src/server/runTrace.ts
+++ b/packages/graph/src/server/runTrace.ts
@@ -3,7 +3,7 @@ import { ITtscGraphEdge } from "../structures/ITtscGraphEdge";
 import { ITtscGraphEvidence } from "../structures/ITtscGraphEvidence";
 import { ITtscGraphNode } from "../structures/ITtscGraphNode";
 import { ITtscGraphTrace } from "../structures/ITtscGraphTrace";
-import { isExternalNode, isTestPath } from "./pathPolicy";
+import { isDeclarationFile, isExternalNode, isTestPath } from "./pathPolicy";
 import { resolveGraphHandle } from "./resolveHandle";
 import { IRunnerOutput, resultNext } from "./resultNext";
 import { edgeEvidenceOf, signatureOf } from "./runDetails";
@@ -20,15 +20,23 @@ const MAX_OPEN_DEPTH = 8;
 const MAX_OPEN_NODES = 32;
 const MAX_IMPACT_DEPTH = 4;
 const MAX_IMPACT_NODES = 16;
+// Path mode walks further than an open trace because it is looking for one
+// named end, not building a picture. The cap is the one the request contract
+// publishes for path mode.
+const MAX_PATH_DEPTH = 12;
 const MAX_HOPS_PER_NODE = 2;
 const MAX_STEPS = 12;
-const EXECUTION_KINDS = new Set<string>([
-  "calls",
-  "instantiates",
-  "accesses",
-  "renders",
-]);
 const DISPATCH_KINDS = new Set<string>(["overrides", "implements"]);
+// A declaration whose kind is a type surface never carries a body, and an
+// external leaf carries one the graph deliberately does not hold.
+const BODYLESS_KINDS = new Set<string>([
+  "interface",
+  "type",
+  "external_symbol",
+]);
+// `abstract` and `declare` are the two keywords that take the body away from a
+// declaration that would otherwise have to have one.
+const BODYLESS_MODIFIERS = new Set<string>(["abstract", "declare"]);
 // An interface the codebase implements everywhere — a disposable, a listener, a
 // lifecycle hook — is not a step in one flow, and naming its implementors is a
 // dump of the codebase rather than an answer. Past this many, the declaration
@@ -143,20 +151,26 @@ export function runTrace(
         ),
       };
     }
-    const found = findPath(
+    const pathDepth = bound(props.maxDepth, MAX_PATH_DEPTH, 1, MAX_PATH_DEPTH);
+    const search = findPath(
       graph,
       start.node.id,
       target.node.id,
-      bound(props.maxDepth, 12, 1, 12),
+      pathDepth,
       focus,
       includeExternal,
     );
-    const hasPath = found !== null;
-    const path = found?.path ?? [];
-    const hops = found?.hops ?? [];
-    const junctions = hasPath
-      ? []
-      : junctionsBetween(graph, start.node.id, target.node.id, focus);
+    const hasPath = search.found !== undefined;
+    const path = search.found?.path ?? [];
+    const hops = search.found?.hops ?? [];
+    // Junctions explain an absence, so they are only computed once absence is
+    // established: a walk the depth bound stopped has not established one, and
+    // presenting a shared symbol as "the seam" when a direct call path may run
+    // past the bound sends the caller to a seam that is not there.
+    const junctions =
+      hasPath || search.bounded
+        ? []
+        : junctionsBetween(graph, start.node.id, target.node.id, focus);
     return {
       result: {
         ...base,
@@ -169,27 +183,31 @@ export function runTrace(
       },
       // A missing path is a fact, not an answer, and the old message called it
       // one: "its path nodes and evidence ranges are what the graph holds
-      // between the two ends" — of a result that held nothing. `findPath` uses
-      // null for that state; zero hops is instead the valid path
-      // from a node to itself. Distinct nodes without a path do not call each
-      // other, which in an event-driven codebase is the common
-      // case: a pointer handler emits, an emitter's `emit()` runs listeners a
+      // between the two ends" — of a result that held nothing. `findPath` says
+      // which fact it is; zero hops is instead the valid path from a node to
+      // itself. A walk the depth bound stopped establishes nothing at all, so
+      // it reports the bound. A walk that ran out of eligible graph did
+      // establish an absence, and distinct nodes without a path do not call
+      // each other, which in an event-driven codebase is the common case: a
+      // pointer handler emits, an emitter's `emit()` runs listeners a
       // registration put in an array, and no call edge crosses that array. The
       // callers of the target are the way across, and the graph has them, so
       // say which call to make instead of handing back an empty result dressed
       // as the answer. Excalidraw's tour spent eleven calls finding this out.
       next: hasPath
         ? pathNext
-        : junctions.length > 0
-          ? resultNext(
-              "inspect",
-              "No call path runs between the two ends — a callback stands between them (an event emitter, a subscription, a lifecycle hook), and no call edge crosses one. `junctions` names the symbols both ends touch, which is the seam: trace the junction to see who registers on it and who fires it.",
-              "trace",
-            )
-          : resultNext(
-              "outside",
-              "No call path runs from the start to the target and they touch nothing in common, so the graph holds no connection between them.",
-            ),
+        : search.bounded
+          ? resultNext("inspect", boundedPathReason(pathDepth), "trace")
+          : junctions.length > 0
+            ? resultNext(
+                "inspect",
+                "No call path runs between the two ends — a callback stands between them (an event emitter, a subscription, a lifecycle hook), and no call edge crosses one. `junctions` names the symbols both ends touch, which is the seam: trace the junction to see who registers on it and who fires it.",
+                "trace",
+              )
+            : resultNext(
+                "outside",
+                "No call path runs from the start to the target and they touch nothing in common, so the graph holds no connection between them.",
+              ),
     };
   }
 
@@ -299,7 +317,7 @@ function traceEdges(
   focus: ITtscGraphTrace.IRequest["focus"],
 ): readonly ITtscGraphEdge[] {
   return reverse
-    ? graph.incoming(id)
+    ? [...graph.incoming(id), ...reverseDispatchEdges(graph, id, focus)]
     : [...graph.outgoing(id), ...dispatchEdges(graph, id, focus)];
 }
 
@@ -340,11 +358,6 @@ function traceSteps(
   });
 }
 
-/**
- * The shortest dependency path from `startId` to `targetId` over real (non-
- * structural) forward edges, breadth-first, or null when `targetId` is not
- * reachable within maxDepth. Returns the nodes in order and the hops between.
- */
 /**
  * The symbols both ends of an unreachable path touch.
  *
@@ -433,6 +446,57 @@ function touchedBy(
   return touched;
 }
 
+/**
+ * What to say when the depth bound, not the graph, ended a path search.
+ *
+ * The caller asked a bounded question and the old answer returned a claim about
+ * the whole graph — "they touch nothing in common, so the graph holds no
+ * connection between them" — which is the worst thing an index can say wrongly:
+ * the caller stops asking and either reads files or concludes the dependency is
+ * not there. So report the boundary, and make the continuation one the caller
+ * can actually take. At the 12-hop ceiling there is no larger `maxDepth` to
+ * retry with, and a message that only invites one would be a dead end of its
+ * own; two bounded walks from opposite ends cover twice the distance and are
+ * requests the tool already answers.
+ */
+function boundedPathReason(depth: number): string {
+  return (
+    `No path was found within the requested depth of ${depth}, but the walk stopped on that bound with eligible graph still ahead of it. ` +
+    `This is a boundary, not an absence: the two ends may be connected further out, and nothing here says they are not. ` +
+    (depth < MAX_PATH_DEPTH
+      ? `Re-run the same path request with a larger \`maxDepth\` (up to ${MAX_PATH_DEPTH}).`
+      : `\`maxDepth\` is already at its ${MAX_PATH_DEPTH}-hop maximum, so close the gap from both ends: trace forward from the start, trace the target with \`direction: "reverse"\`, then request the path between a symbol both results name.`)
+  );
+}
+
+/**
+ * What a bounded shortest-path walk learned: the path when it found one, and
+ * otherwise whether the walk was stopped by the caller's depth bound or ran the
+ * eligible graph out. The two are not the same answer and the caller must not
+ * be told the second when only the first happened.
+ */
+interface IPathSearch {
+  /** The shortest eligible path and its hops, when one was found. */
+  found?: { path: ITtscGraphNode[]; hops: ITtscGraphTrace.IHop[] };
+
+  /**
+   * True when the walk stopped at `maxDepth` with an eligible, unvisited node
+   * still ahead of it, so nothing was proven about the two ends.
+   */
+  bounded: boolean;
+}
+
+/**
+ * The shortest dependency path from `startId` to `targetId` over real (non-
+ * structural) forward edges, breadth-first, within `maxDepth` hops.
+ *
+ * When no path is found, the walk reports whether the bound stopped it. A
+ * boundary is a fact about the request; absence is a fact about the graph, and
+ * a search that never reached the far side of its own bound has not established
+ * one. Eligibility is the open trace's, so a frontier made only of nodes the
+ * selected focus, the external-node policy, a file node, or an earlier visit
+ * already excluded is not a frontier and the walk is exhausted.
+ */
 function findPath(
   graph: TtscGraphMemory,
   startId: string,
@@ -440,10 +504,11 @@ function findPath(
   maxDepth: number,
   focus: ITtscGraphTrace.IRequest["focus"],
   includeExternal: boolean,
-): { path: ITtscGraphNode[]; hops: ITtscGraphTrace.IHop[] } | null {
+): IPathSearch {
   const startNode = graph.node(startId);
-  if (startNode === undefined) return null;
-  if (startId === targetId) return { path: [startNode], hops: [] };
+  if (startNode === undefined) return { bounded: false };
+  if (startId === targetId)
+    return { found: { path: [startNode], hops: [] }, bounded: false };
   const parent = new Map<
     string,
     {
@@ -453,21 +518,36 @@ function findPath(
     }
   >();
   const visited = new Set<string>([startId]);
+  let bounded = false;
   let queue: Array<{ id: string; depth: number }> = [{ id: startId, depth: 0 }];
   while (queue.length > 0) {
     const next: Array<{ id: string; depth: number }> = [];
     for (const { id, depth } of queue) {
-      if (depth >= maxDepth) continue;
-      for (const edge of [
+      const candidates = [
         ...graph.outgoing(id),
         ...dispatchEdges(graph, id, focus),
-      ]) {
-        if (!traversable(edge.kind, focus)) continue;
-        const otherId = edge.to;
-        if (visited.has(otherId)) continue;
-        const other = graph.node(otherId);
-        if (other === undefined || other.kind === "file") continue;
-        if (!includeExternal && isExternalNode(other)) continue;
+      ];
+      if (depth >= maxDepth) {
+        if (
+          candidates.some(
+            (edge) =>
+              pathEndpoint(graph, edge, focus, includeExternal, visited) !==
+              undefined,
+          )
+        )
+          bounded = true;
+        continue;
+      }
+      for (const edge of candidates) {
+        const endpoint = pathEndpoint(
+          graph,
+          edge,
+          focus,
+          includeExternal,
+          visited,
+        );
+        if (endpoint === undefined) continue;
+        const otherId = endpoint.otherId;
         visited.add(otherId);
         const evidence = edgeEvidenceOf(edge);
         parent.set(otherId, {
@@ -503,14 +583,42 @@ function findPath(
               hop.evidence = parentEdge.evidence;
             hops.push(hop);
           }
-          return { path, hops };
+          return { found: { path, hops }, bounded };
         }
         next.push({ id: otherId, depth: depth + 1 });
       }
     }
     queue = next;
   }
-  return null;
+  return { bounded };
+}
+
+/**
+ * The node a path expansion would represent, or undefined when the selected
+ * policy or an earlier visit excludes it. The expansion and the boundary probe
+ * share this decision, so what the walk would have followed and what counts as
+ * unexplored graph beyond the bound cannot disagree.
+ *
+ * A path walk is always forward and, unlike the open trace, a node it already
+ * reached is not a continuation: the shortest path to it is already known, so a
+ * second arrival adds nothing to explore.
+ */
+function pathEndpoint(
+  graph: TtscGraphMemory,
+  edge: ITtscGraphEdge,
+  focus: ITtscGraphTrace.IRequest["focus"],
+  includeExternal: boolean,
+  visited: ReadonlySet<string>,
+): ITraceEndpoint | undefined {
+  const endpoint = eligibleTraceEndpoint(
+    graph,
+    edge,
+    false,
+    focus,
+    includeExternal,
+  );
+  if (endpoint === undefined || visited.has(endpoint.otherId)) return undefined;
+  return endpoint;
 }
 
 function orderedEdges(
@@ -632,12 +740,18 @@ function dispatchEdges(
   id: string,
   focus: ITtscGraphTrace.IRequest["focus"],
 ): ITtscGraphEdge[] {
-  if (focus === "types" || hasExecutionBody(graph, id)) return [];
+  if (focus === "types") return [];
+  const declaration = graph.node(id);
+  if (declaration === undefined || hasDeclarationBody(graph, declaration))
+    return [];
   const out: ITtscGraphEdge[] = [];
   for (const edge of graph.incoming(id)) {
     if (!DISPATCH_KINDS.has(edge.kind)) continue;
     const implementation = graph.node(edge.from);
-    if (implementation === undefined || !hasExecutionBody(graph, edge.from))
+    if (
+      implementation === undefined ||
+      !hasDeclarationBody(graph, implementation)
+    )
       continue;
     out.push({
       from: id,
@@ -649,11 +763,106 @@ function dispatchEdges(
   return out.length >= DISPATCH_HUB ? [] : out;
 }
 
-/** Whether the declaration has a body: something it calls, reads, or renders. */
-function hasExecutionBody(graph: TtscGraphMemory, id: string): boolean {
-  for (const edge of graph.outgoing(id))
-    if (EXECUTION_KINDS.has(edge.kind)) return true;
+/**
+ * The declaration a change to this implementation is a change to.
+ *
+ * `dispatches` is one fact and both directions have to see it. Forward, a call
+ * that lands on a bodyless declaration continues in the implementation that
+ * runs; reverse, a change to that implementation is a change every caller of
+ * the declaration feels — and `impact` is the query an agent runs _before_
+ * editing a method. The checker relation is oriented implementation-to-base, so
+ * from the implementation it is an outgoing edge and a reverse walk, which only
+ * reads incoming edges, never sees it. Both halves of the path are one step
+ * away in a direction the traversal does not take.
+ *
+ * The synthetic edge is the forward one, unchanged, so eligibility, checker
+ * validity, and the hub cut cannot drift apart between the two directions: ask
+ * the base what it dispatches to, and keep the edges that land here.
+ */
+function reverseDispatchEdges(
+  graph: TtscGraphMemory,
+  id: string,
+  focus: ITtscGraphTrace.IRequest["focus"],
+): ITtscGraphEdge[] {
+  if (focus === "types") return [];
+  const out: ITtscGraphEdge[] = [];
+  const bases = new Set<string>();
+  for (const edge of graph.outgoing(id)) {
+    if (!DISPATCH_KINDS.has(edge.kind) || bases.has(edge.to)) continue;
+    bases.add(edge.to);
+    for (const dispatch of dispatchEdges(graph, edge.to, focus))
+      if (dispatch.to === id) out.push(dispatch);
+  }
+  return out;
+}
+
+/**
+ * Whether the declaration this node stands for has a body of its own.
+ *
+ * Having a body and naming a modeled dependency are different facts. Counting
+ * outgoing `calls`/`accesses`/`instantiates`/`renders` edges measures the
+ * second and answers as though it were the first, and it is wrong in both
+ * directions: an implementation whose body returns a literal, throws, or only
+ * moves locals around has degree zero and was refused as a dispatch target,
+ * while a concrete base method was promoted through its own override the moment
+ * its body stopped naming anything the graph models. Two graphs identical in
+ * every declaration fact then answered differently because of one statement
+ * inside a body.
+ *
+ * So read the declaration instead. A type surface has no body; `abstract` and
+ * `declare` take it away; an interface member and anything inside an ambient
+ * container never had one; a `.d.ts` declaration is ambient whether or not the
+ * keyword is written; and an external leaf has a body the graph deliberately
+ * does not hold. Everything else is a concrete declaration, which is a real
+ * destination and is never promoted through an override, whatever it calls.
+ */
+function hasDeclarationBody(
+  graph: TtscGraphMemory,
+  node: ITtscGraphNode,
+): boolean {
+  if (BODYLESS_KINDS.has(node.kind)) return false;
+  if (isExternalNode(node)) return false;
+  if (isDeclarationFile(node.file)) return false;
+  if (hasBodylessModifier(node)) return false;
+  return !inBodylessContainer(graph, node);
+}
+
+function hasBodylessModifier(node: ITtscGraphNode): boolean {
+  return node.modifiers?.some((m) => BODYLESS_MODIFIERS.has(m)) === true;
+}
+
+/**
+ * Whether an owner up the `contains` tree makes this declaration bodyless: an
+ * interface, or an ambient container. A member writes no keyword of its own —
+ * `declare` on a class or namespace is not repeated on what it holds — so the
+ * fact lives on the owner and the walk has to go get it.
+ */
+function inBodylessContainer(
+  graph: TtscGraphMemory,
+  node: ITtscGraphNode,
+): boolean {
+  const seen = new Set<string>([node.id]);
+  let current: ITtscGraphNode | undefined = node;
+  while (current !== undefined) {
+    const container: ITtscGraphNode | undefined = containerOf(graph, current);
+    if (container === undefined || container.kind === "file") return false;
+    if (seen.has(container.id)) return false;
+    seen.add(container.id);
+    if (container.kind === "interface" || hasBodylessModifier(container))
+      return true;
+    current = container;
+  }
   return false;
+}
+
+/** The declaration that owns this one, through the synthesized ownership tree. */
+function containerOf(
+  graph: TtscGraphMemory,
+  node: ITtscGraphNode,
+): ITtscGraphNode | undefined {
+  for (const edge of graph.incoming(node.id))
+    if (edge.kind === "contains") return graph.node(edge.from);
+  return undefined;
 }
 
 /** An edge the trace should follow: a real dependency, not a structural edge. */
--- a/packages/graph/src/server/runTrace.ts
+++ b/packages/graph/src/server/runTrace.ts
@@ -316,9 +316,13 @@ function traceEdges(
   reverse: boolean,
   focus: ITtscGraphTrace.IRequest["focus"],
 ): readonly ITtscGraphEdge[] {
-  return reverse
-    ? [...graph.incoming(id), ...reverseDispatchEdges(graph, id, focus)]
-    : [...graph.outgoing(id), ...dispatchEdges(graph, id, focus)];
+  const edges = reverse ? graph.incoming(id) : graph.outgoing(id);
+  const dispatched = reverse
+    ? reverseDispatchEdges(graph, id, focus)
+    : dispatchEdges(graph, id, focus);
+  // Nothing to add is the common case, and a walk visits the nodes with the
+  // largest edge lists, so hand back the stored list rather than a copy of it.
+  return dispatched.length === 0 ? edges : [...edges, ...dispatched];
 }
 
 /**
--- a/packages/graph/src/server/runTrace.ts
+++ b/packages/graph/src/server/runTrace.ts
@@ -527,10 +527,9 @@ function findPath(
   while (queue.length > 0) {
     const next: Array<{ id: string; depth: number }> = [];
     for (const { id, depth } of queue) {
-      const candidates = [
-        ...graph.outgoing(id),
-        ...dispatchEdges(graph, id, focus),
-      ];
+      // The forward step the open trace would take, built in one place so the
+      // two walks cannot disagree about what a step follows.
+      const candidates = traceEdges(graph, id, false, focus);
       if (depth >= maxDepth) {
         if (
           candidates.some(
--- a/tests/test-graph/src/features/test_ttscgraph_dispatch_reads_declaration_facts_not_dependency_degree.ts
+++ b/tests/test-graph/src/features/test_ttscgraph_dispatch_reads_declaration_facts_not_dependency_degree.ts
@@ -35,12 +35,15 @@ interface TraceResult {
  * 1. Materialize implementations that name nothing (empty, literal, local
  *    arithmetic, thrown literal) behind an interface member, an abstract
  *    method, and an ambient `declare class` member, plus two concrete bases
- *    with overrides that differ only in one statement, plus an overload set.
+ *    with overrides that differ only in one statement, an abstract member
+ *    standing between an interface and its concrete class, and an overload
+ *    set.
  * 2. Trace forward from each caller through the real MCP launcher, and request the
  *    same continuation in path mode.
  * 3. Assert every genuinely bodyless declaration dispatches to its dependency-
  *    free implementation, that neither concrete base is promoted through its
- *    override, and that `focus: "types"` still synthesizes nothing.
+ *    override, that no bodyless candidate is admitted as an implementation, and
+ *    that `focus: "types"` still synthesizes nothing.
  */
 export const test_ttscgraph_dispatch_reads_declaration_facts_not_dependency_degree =
   async () => {
@@ -113,6 +116,24 @@ export const test_ttscgraph_dispatch_reads_declaration_facts_not_dependency_degr
         "",
         "export class QuietTask extends Task {",
         "  public perform(): void {}",
+        "}",
+        "",
+        "export interface Shape {",
+        "  area(): number;",
+        "}",
+        "",
+        "export abstract class BaseShape implements Shape {",
+        "  public abstract area(): number;",
+        "}",
+        "",
+        "export class Square extends BaseShape {",
+        "  public area(): number {",
+        "    return 4;",
+        "  }",
+        "}",
+        "",
+        "export function callShape(shape: Shape): number {",
+        "  return shape.area();",
         "}",
         "",
         "declare class Native {",
@@ -239,6 +260,21 @@ export const test_ttscgraph_dispatch_reads_declaration_facts_not_dependency_degr
       assert.ok(
         dispatchedIn(abstractTask).includes("QuietTask.perform"),
         `an abstract method dispatches to a dependency-free override: ${JSON.stringify(abstractTask.hops)}`,
+      );
+
+      // The other half of the same predicate: a candidate that is itself
+      // bodyless is not an implementation. An abstract member standing between
+      // an interface and the concrete class is that candidate, and it must
+      // never be the destination of a hop that claims to name what runs.
+      const abstractCandidate = await forward("callShape");
+      assert.ok(
+        !dispatchedIn(abstractCandidate).includes("BaseShape.area"),
+        `an abstract member is never a dispatch destination: ${JSON.stringify(abstractCandidate.hops)}`,
+      );
+      const abstractIntermediate = await forward("BaseShape.area");
+      assert.ok(
+        dispatchedIn(abstractIntermediate).includes("Square.area"),
+        `and the concrete override below it still is: ${JSON.stringify(abstractIntermediate.hops)}`,
       );
 
       const ambient = await forward("callNative");
--- a/tests/test-graph/src/features/test_ttscgraph_dispatch_reads_declaration_facts_not_dependency_degree.ts
+++ b/tests/test-graph/src/features/test_ttscgraph_dispatch_reads_declaration_facts_not_dependency_degree.ts
@@ -36,8 +36,8 @@ interface TraceResult {
  *    arithmetic, thrown literal) behind an interface member, an abstract
  *    method, and an ambient `declare class` member, plus two concrete bases
  *    with overrides that differ only in one statement, an abstract member
- *    standing between an interface and its concrete class, and an overload
- *    set.
+ *    standing between an interface and its concrete class, one implementation
+ *    named in two heritage clauses, and an overload set.
  * 2. Trace forward from each caller through the real MCP launcher, and request the
  *    same continuation in path mode.
  * 3. Assert every genuinely bodyless declaration dispatches to its dependency-
@@ -134,6 +134,18 @@ export const test_ttscgraph_dispatch_reads_declaration_facts_not_dependency_degr
         "",
         "export function callShape(shape: Shape): number {",
         "  return shape.area();",
+        "}",
+        "",
+        "export abstract class Twice {",
+        "  public abstract emit(): void;",
+        "}",
+        "",
+        "export class OnlyOnce extends Twice implements Twice {",
+        "  public emit(): void {}",
+        "}",
+        "",
+        "export function callTwice(twice: Twice): void {",
+        "  twice.emit();",
         "}",
         "",
         "declare class Native {",
@@ -268,6 +280,10 @@ export const test_ttscgraph_dispatch_reads_declaration_facts_not_dependency_degr
       // never be the destination of a hop that claims to name what runs.
       const abstractCandidate = await forward("callShape");
       assert.ok(
+        abstractCandidate.reached.some((node) => node.name === "Shape.area"),
+        `the walk stands on the interface member: ${abstractCandidate.reached.map((n) => n.name).join(", ")}`,
+      );
+      assert.ok(
         !dispatchedIn(abstractCandidate).includes("BaseShape.area"),
         `an abstract member is never a dispatch destination: ${JSON.stringify(abstractCandidate.hops)}`,
       );
@@ -275,6 +291,17 @@ export const test_ttscgraph_dispatch_reads_declaration_facts_not_dependency_degr
       assert.ok(
         dispatchedIn(abstractIntermediate).includes("Square.area"),
         `and the concrete override below it still is: ${JSON.stringify(abstractIntermediate.hops)}`,
+      );
+
+      // One implementation named in two heritage clauses is one implementation.
+      // The producer records the member pair once per clause, as `overrides`
+      // and as `implements`, so a walk over relations would cross to the same
+      // body twice and count one class twice against the hub cut.
+      const twoRelations = await forward("callTwice");
+      assert.deepEqual(
+        dispatchedIn(twoRelations),
+        ["OnlyOnce.emit"],
+        `a doubly related implementation is dispatched to once: ${JSON.stringify(twoRelations.hops)}`,
       );
 
       const ambient = await forward("callNative");
--- a/tests/test-graph/src/features/test_ttscgraph_dispatch_reads_declaration_facts_not_dependency_degree.ts
+++ b/tests/test-graph/src/features/test_ttscgraph_dispatch_reads_declaration_facts_not_dependency_degree.ts
@@ -1,0 +1,320 @@
+import { TestProject } from "@ttsc/testing";
+
+import { TtsgraphClient, assert } from "../internal/ttsgraph";
+
+interface ToolResult {
+  structuredContent?: {
+    result?: TraceResult;
+  };
+}
+
+interface TraceResult {
+  type: "trace";
+  hops: { from: string; to: string; kind: string }[];
+  reached: { id: string; name: string }[];
+  path?: { id: string; name: string }[];
+}
+
+/**
+ * Verifies a dispatch hop is decided by what a declaration is, not by how many
+ * modeled symbols its body happens to name.
+ *
+ * Body detection used to count outgoing `calls`/`accesses`/`instantiates`/
+ * `renders` edges, which measures dependency degree instead. That is wrong in
+ * both directions: an implementation whose body is empty, returns a literal,
+ * throws, or only moves locals around scored zero and was refused as a dispatch
+ * target, so the trace dead-ended on the declaration and reported nothing was
+ * omitted; and a concrete base method was promoted through its own override the
+ * moment its body stopped naming anything the graph models, so two graphs
+ * identical in every declaration fact answered differently because of one
+ * statement inside a body. The declaration facts the fix reads — kind,
+ * `abstract`/`declare` modifiers, and interface or ambient ownership — are only
+ * proven end to end through the real producer, which is why this runs the
+ * shipped binary rather than a synthetic in-memory graph.
+ *
+ * 1. Materialize implementations that name nothing (empty, literal, local
+ *    arithmetic, thrown literal) behind an interface member, an abstract
+ *    method, and an ambient `declare class` member, plus two concrete bases
+ *    with overrides that differ only in one statement, plus an overload set.
+ * 2. Trace forward from each caller through the real MCP launcher, and request the
+ *    same continuation in path mode.
+ * 3. Assert every genuinely bodyless declaration dispatches to its dependency-
+ *    free implementation, that neither concrete base is promoted through its
+ *    override, and that `focus: "types"` still synthesizes nothing.
+ */
+export const test_ttscgraph_dispatch_reads_declaration_facts_not_dependency_degree =
+  async () => {
+    const root = TestProject.createProject({
+      "tsconfig.json": JSON.stringify(
+        {
+          compilerOptions: {
+            target: "ES2022",
+            module: "commonjs",
+            strict: true,
+            rootDir: "src",
+            outDir: "dist",
+          },
+          include: ["src"],
+        },
+        null,
+        2,
+      ),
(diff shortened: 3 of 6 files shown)


samchon:
## Implementation landed: `4b3739e`

### What each issue got

**#828, dispatch reads declaration facts.** `hasExecutionBody` (outgoing-edge degree) is replaced by `hasDeclarationBody`, which reads the declaration: `interface`/`type`/`external_symbol` kinds, `isExternalNode`, a `.d.ts` declaring file, an `abstract` or `declare` modifier, and an interface or ambient owner found by walking the synthesized `contains` tree. Both sides of the decision, the gate on the called declaration and the admission test on a candidate implementation, use the same predicate, so dispatch continues only from a genuinely bodyless declaration and a concrete declaration with a body is never promoted through an override.

The ownership walk is not one level. `GetCombinedModifierFlags` does not propagate `declare` from a container to its members (`getCombinedFlags` only walks the variable-declaration chain), so a member of `declare class Native` and a class inside `declare namespace N` carry no keyword of their own. The fact lives on the owner, so `inBodylessContainer` walks up until a `file` node.

`.d.ts` is included as a declaration fact, not a path heuristic: `export class Foo { m(): void }` inside a declaration file is ambient without writing `declare`, and every declaration such a file holds is bodyless. `isDeclarationFile` was factored out of `isSupportPath` in `pathPolicy.ts` so the two cannot drift.

**#829, reverse and impact cross the seam.** `reverseDispatchEdges` does not rebuild the synthetic edge. For each outgoing `overrides`/`implements` relation it asks `dispatchEdges` what that base dispatches to and keeps the edges that land on the start node. Eligibility, checker validity, the `focus: "types"` exclusion, and the `DISPATCH_HUB` cut are therefore literally the forward ones, and #828's rule is read from where it is defined rather than mirrored by hand. The edge orientation is unchanged (`from` is the declaration, `to` is the implementation), which is exactly what the reverse branch of `eligibleTraceEndpoint` expects.

**#827, a bound is not an absence.** `findPath` returns `IPathSearch` with three states: found, exhausted, and bounded. The depth-limit branch no longer just `continue`s; it probes whether any candidate edge has an eligible, unvisited endpoint, through `pathEndpoint`, which delegates to the open trace's own `eligibleTraceEndpoint` and adds the visited rule (a node already reached is not a continuation, because its shortest path is already known). A bounded result reports `inspect` with `request: "trace"`, and it does not compute `junctions`, because the field's protected prose scopes them to "when no call path runs between them" and a bounded walk has established nothing. `truncated` is untouched, as the issue requires.

The bounded continuation adapts: below the ceiling it says to re-run with a larger `maxDepth`; at the 12-hop ceiling, where no larger value exists, it says to trace forward from the start, trace the target with `direction: "reverse"`, and request the path between a symbol both results name. Two bounded walks from opposite ends cover twice the distance, using requests the tool already answers.

### Effect on the #784 hub bound, stated explicitly

This batch does not implement #784 and does not touch `DISPATCH_HUB`, but it **does change which candidates the bound counts**, and that is worth recording before #784 lands on top.

`dispatchEdges` still suppresses the whole set at `out.length >= DISPATCH_HUB`. `out` is now the set of concrete implementations, where it used to be the set of implementations with non-zero dependency degree. The candidate set can therefore only grow, never shrink, so:

- a declaration whose implementations all had degree behaves identically;
- a declaration with, say, 14 implementations of which 8 named a modeled symbol used to emit 8 hops and now emits none, because 14 crosses the cut;
- conversely, nothing that was suppressed before becomes unsuppressed.

That is the corrected candidate set #784's coordination section asks for ("the cap applies to the candidate set this issue corrects"), and #784's own selected/omitted reporting will apply to it. The same set is now counted in the reverse direction too, since `reverseDispatchEdges` consults `dispatchEdges`.

### Local verification, campaign CI suspended

Narrow, then broad, package-scoped. Exact commands and their real output:

| command | result |
| --- | --- |
| `pnpm install` | `Done in 6m 14.5s using pnpm v10.6.4` |
| `pnpm run build:current` | exit 0 |
| `pnpm --filter @ttsc/graph build` | `bundled viewer: 654 KB`, exit 0 |
| `pnpm --filter @ttsc/test-graph start -- --include=dispatch_reads_declaration_facts` | `Success`, 9,556 ms |
| `pnpm --filter @ttsc/test-graph start -- --include=reverse_and_impact_cross` | `Success`, 10,569 ms |
| `pnpm --filter @ttsc/test-graph start -- --include=path_reports_its_depth_bound` | `Success`, 6,063 ms |
| `pnpm --filter @ttsc/test-graph start` (42 cases) | `Success`, 151,689 ms / 92,578 ms / 240,253 ms / 139,111 ms across four runs |
| `npx tsc --noEmit -p tests/test-graph/tsconfig.json` | clean |
| `npx tsc --noEmit -p packages/graph/tsconfig.json` | clean |

**Fails before the fix.** `packages/graph/src/server/{runTrace,pathPolicy}.ts` were reverted to `HEAD`, `@ttsc/graph` rebuilt, and the three new cases re-run. All three failed, each on the symptom its issue reports:

```
AssertionError: an implementation with an empty body is still the code that runs: [{"from":"src/app.ts#callPipeline:function","to":"src/app.ts#Pipeline.execute:method","kind":"calls","depth":1,...}]
AssertionError: nor as a callback seam, when a direct path runs past the bound
  actual: 'No call path runs between the two ends — a callback stands between them ...'
AssertionError: reverse/execution crosses the seam to Pipeline.execute:   (reached nothing)
```

The sources were then restored and `@ttsc/graph` rebuilt before the committed suite runs above.

**One transient failure, disclosed.** Of five full-suite runs, the second reported `Failed` after 264,703 ms (the slowest run by far). Its grep output carried no `AssertionError` line, so it was a thrown error rather than a failed assertion, the shape the suite's own 120 s per-request and 30 s exit timeouts produce under machine load. I did not capture the case name before the output scrolled, which is my mistake. Three subsequent full-suite runs (92 s, 240 s, 139 s) were clean, and the three new cases pass individually. I am recording it rather than dismissing it; if it recurs on any later run in this branch it will be reported and diagnosed rather than re-run away.

`pnpm test:go` was not run: no Go source changed. `pnpm format` was not run, per the campaign's no-format rule. Only the six files this branch touches were formatted, with `npx prettier --write` on those exact paths.

### Protected contract

- `git rev-parse HEAD:packages/graph/src/structures` → `3c7399e3ce9f2404e64d652c7f1259ebd0db2613`
- SHA-256 of `ITtscGraphApplication.ts` as checked out on Windows → `6AB7451FCD17E8A8C1F0F7BAEAA6A96AA4E6DEDCB379F70F030ECEE64197BE6C`

Both match the required values, checked before the first commit and again after the implementation commit. No structure, field, edge kind, or JSDoc was changed.

### Documentation

`website/src/content/docs/graph/index.mdx` gained two paragraphs next to the existing completeness paragraph: what path mode says when it finds nothing, and how dispatch is decided and crossed in both directions.

samchon:
## Self-Review, round 1

Solo, per `.agents/skills/review/SKILL.md`. Surface: the full `origin/master...HEAD` diff (6 files, 1396 insertions, 57 deletions) plus the working tree, read whole, and traced into `runTour.ts`, `TtscGraphMemory.ts`, `pathPolicy.ts` consumers, the Go producer's modifier emission, and the existing graph suite. Findings and the plan first; changes follow this comment.

### Sound improvements to apply

**1. `dispatchEdges` runs the expensive predicate before it knows there is anything to dispatch to.**

`hasDeclarationBody` on the called declaration now walks the `contains` chain, and `containerOf` scans a node's whole incoming edge list. `TtscGraphMemory` appends synthesized `contains` edges last, so that scan is always a full pass. The old degree predicate scanned outgoing edges and returned early for any node that called something, which is most nodes a walk visits; the new gate has no such early exit, and it is evaluated once per traversal step in both `traceEdges` and `findPath`, including the new boundary probe in `findPath`.

The gate is only meaningful when the declaration has at least one incoming `overrides`/`implements` edge, which almost no node has. Collect those relations first and return early when there are none, then run the gate. Same result on every input, one linear scan instead of two on the hot path.

**2. The admission side has no negative twin in the tests.**

The development skill requires a counter-example wherever a predicate acts. `test_..._dispatch_reads_declaration_facts_...` pins that a bodyless declaration dispatches and that a concrete base is not promoted, but nothing pins the other half of the same predicate: a candidate that is itself bodyless must not be admitted as an implementation. An abstract member sitting between an interface and the concrete class is exactly that case, and it is silently correct today in both the old and the new code, which is precisely how an over-match stays invisible.

Add `Shape` / `abstract BaseShape.area` / `Square.area` and assert that no `dispatches` hop targets the abstract `BaseShape.area`, while a trace from `BaseShape.area` itself does dispatch to `Square.area`. The assertion is written on the rule ("an abstract member is never a dispatch destination"), not on the resulting dead end, so it stays correct if transitive dispatch through an abstract intermediate is ever added as its own feature. That chain stopping at `Shape.area` is unchanged from before this batch and is out of scope here.

### Verified, deliberately unchanged

**An `interface` node itself is a dispatch source, and now unconditionally.** This is behaviour worth stating plainly rather than leaving to be discovered. A class carries a heritage `implements` edge to the interface it implements, so `dispatchEdges` on an *interface* node (not just on its members) yields `dispatches` to the implementing classes. That already happened before this batch, gated by whether the class had a non-zero execution degree; #828's rule removes that gate on both sides, so it now happens for every implementing class, subject to the unchanged `DISPATCH_HUB` cut.

`focus: "execution"` cannot reach an interface node (no `calls`/`accesses`/`instantiates`/`renders` edge targets one), so this is visible only under `focus: "all"`, which is the default, arriving via `type_ref`. `reverseDispatchEdges` inherits it symmetrically: a reverse walk from a class now crosses to the interface it implements. Under the settled rule this is right, an interface is genuinely bodyless and a class is a concrete declaration with a body, and narrowing dispatch to member kinds would *remove* an answer the tool gives today. Keeping it, and disclosing it.

**`reverseDispatchEdges` recomputes the base's whole forward candidate set per reverse step.** That is deliberate and not reducible: the hub cut is a property of the whole set, so the set has to be built to know whether it is suppressed. Reusing `dispatchEdges` is what makes eligibility, checker validity, the `focus: "types"` exclusion and the cut provably identical between directions, which is worth more than the scan.

**`IPathSearch` can in principle carry `bounded: true` next to `found`.** It cannot in practice: BFS processes one depth level at a time, so a level either expands (and can find) or probes (and can set `bounded`), never both. The caller reads `found` first, so the field is inert there. Leaving the honest value rather than forcing it to `false`.

**`focus: "types"` is excluded twice**, once in each dispatch builder and once in `traversable`. Redundant, matches the pre-existing shape, and each guard is locally correct. No change.

**Windows and POSIX.** `isDeclarationFile` matches on the dump's project-relative, forward-slashed path, the same input the identical regex already had inside `isSupportPath`. No separator assumption is added.

**Downstream.** `runTour`'s flow filter (`isTourHop`) rejects only structural kinds, so `dispatches` hops pass exactly as before; its impact anchors now see the reverse crossing, which is the intended widening. No new edge kind, node field, or dump shape, so the viewer, `reduce.ts`, and every structure consumer are untouched.

### Plan

Apply 1 and 2, rebuild `@ttsc/graph`, re-run the three new cases and then the whole `@ttsc/test-graph` suite plus both typechecks, push, pass the exact-SHA cancellation gate, and start a fresh full round.

samchon:
## Self-Review, round 2

Fresh full round over the current head (`22b583a`), not a recheck of round 1's fix. Same surface: whole base-to-head diff, plus the producer's edge output, which is where this round's finding came from.

### Sound improvements to apply

**1. One implementation can be counted, and dispatched to, twice.**

`memberRelationEdges` runs once per heritage clause, and a class may name the same base in two clauses. `class Impl extends Base implements Base` is legal TypeScript, and the producer has a dedicated Go regression for keeping both container edges (`heritage_edges_keep_extends_and_implements_to_same_base_test.go`). Dumping that shape shows the member pair recorded twice, once per origin:

```
src/app.ts#Impl.run:method -[implements]-> src/app.ts#Base.run:method
src/app.ts#Impl.run:method -[overrides]-> src/app.ts#Base.run:method
```

`dispatchEdges` walks relations, not implementations, so it emits two identical `dispatches` edges to `Impl.run`. The traversal records the first as a hop and the second as a back-edge hop to an already-visited node, so the answer carries the same crossing twice, and the pair counts as two candidates against `DISPATCH_HUB`.

This is pre-existing in the forward direction, and this batch mirrors it into reverse, where #829's acceptance explicitly asks for no duplicate hops. Fixing it in `reverseDispatchEdges` alone would make the two directions disagree, which is the exact failure #829 asks to avoid, so the dedupe belongs in `dispatchEdges` by implementation id. Two relations to one implementation are one implementation: the hop is emitted once, and the hub cut counts implementations rather than heritage clauses, which is also the set #784 will report as selected and omitted.

**2. The new negative twin can pass vacuously.**

`callShape` asserts that no `dispatches` hop targets the abstract `BaseShape.area`. As written that assertion also holds if the trace never got near the hierarchy at all. Assert that the trace reaches `Shape.area` first, so the absence of the hop is an absence at a place the walk actually stood.

### Verified, no change

**An excluded external target is still answered as "no connection".** When `to` resolves to an external node and `includeExternal` is false, `findPath` can never represent it, so the result is the no-connection verdict rather than a policy report. That is entirely pre-existing, and #827's acceptance deliberately requires an externals-only frontier to count as exhausted rather than bounded, so the issue endorses the current treatment. Recording it as a separate observation, not smuggling a fix for it into this batch.

**`relations ??= []` and the reordered gate.** Output is identical for every input: the relation filter is the same predicate the old loop applied, the gate and admission tests are unchanged, and the hub comparison still runs on the emitted set. Confirmed by the full suite, including the pre-existing dispatch regressions.

**Everything re-verified from round 1** still holds: interface-node dispatch (disclosed, deliberate), the reverse builder's reuse of the forward set, `IPathSearch`'s inert `bounded` beside `found`, the doubled `focus: "types"` guard, path-separator handling, and `runTour`'s kind-agnostic hop filter.

### Plan

Dedupe candidates by implementation id in `dispatchEdges` with a case in the dispatch test built on the `extends`-and-`implements`-same-base shape, strengthen the `callShape` assertion, rebuild, re-run the three cases then the whole suite and both typechecks, push, pass the gate, and start round 3.

samchon:
## Self-Review, round 3

Fresh full round over `23cf76e5`. This round followed the cause outward instead of re-reading the same hunks: every other place in `packages/graph` that consumes `overrides`/`implements` or decides "does this have a body".

### Sound improvement to apply

**1. `traceEdges` now copies the reverse edge list on every step, even when nothing is added.**

Before this batch the reverse branch returned `graph.incoming(id)` directly. It now returns `[...graph.incoming(id), ...reverseDispatchEdges(...)]`, which allocates and copies once per traversal node, and a reverse walk visits exactly the nodes with the largest in-degree. Almost none of them have a dispatch relation, so almost every copy is for nothing. Return the list as-is when the synthesized set is empty, and do the same on the forward side, where the copy predates this batch.

### Out of scope, reported rather than fixed

Two verified findings sit outside all three issues in this batch. I am not touching them; the lead owns whether they become issues.

**`runTour`'s centrality reach carries an independent copy of #828's proxy.** `computeCentrality` in `runTour.ts` has its own `DISPATCH_KINDS` and its own body test:

```ts
let hasBody = false;
for (const edge of graph.outgoing(id)) {
  if (!INVOKE_KINDS.has(edge.kind)) continue;
  hasBody = true;
  out.push(edge.to);
}
if (!hasBody) {
  for (const edge of graph.incoming(id)) {
    if (DISPATCH_KINDS.has(edge.kind)) out.push(edge.from);
  }
}
```

That is the same substitution #828 removes from `runTrace.ts`, in a different feature: a concrete base whose body names nothing modeled has its overrides folded into its reach, inflating its centrality, and a bodyless implementation is admitted with no admission test at all. It changes tour seed ranking, not trace hops. #828 names `dispatchEdges` as the architectural owner and accounts for `tour` only through the traces it embeds, and the sibling `runTour.ts` issue is about flow filtering, so no published issue owns this. Fixing it here would widen the batch's scope and change tour ranking without an issue's acceptance criteria to check it against.

**`details.implementedBy` lists one implementation twice for the doubly related shape.** `incomingDependencyRefs` ranks per edge, so `class Impl extends Base implements Base` yields two references to `Impl`, one with `relation: "implements"` and one with `"overrides"`, and both consume the `dependencyLimit` budget. It is arguably informative rather than wrong, unlike a duplicated `dispatches` hop, and `details` is not in any of these three issues. Recorded for the lead; open #786 already owns details member caps and may be the natural home.

### Verified, no change

**The reverse hub assertion is not vacuous in context.** `dispatchHops(hubReverse) === 0` would pass if reverse dispatch did nothing at all, but the same test first proves reverse dispatch crosses for a non-hub declaration in four direction/focus combinations, on the same graph and the same client. The positive is the control for the negative.

**`runEntrypoints` and `runDetails` kind ranking, the viewer, and `reduce.ts`** consume edge kinds but not the body decision, and no kind, field, or dump shape changed.

**Everything from rounds 1 and 2** re-verified against the current head: the relation-first gate, the per-implementation dedupe, the ownership walk, interface-node dispatch (disclosed), `IPathSearch`, the boundary probe's shared eligibility, path separators, and the test fixtures' isolation.

### Plan

Apply 1, rebuild, re-run the three cases then the whole suite and both typechecks, push, pass the gate, and start round 4.

samchon:
## Self-Review, round 4

Fresh full round over `8057610`.

### Sound improvement to apply

**1. `findPath` rebuilds the forward candidate list by hand.**

It spells out `[...graph.outgoing(id), ...dispatchEdges(graph, id, focus)]`, which is exactly what `traceEdges(graph, id, false, focus)` returns. Two consequences: the round-3 short-circuit does not reach path mode, so its BFS still copies the outgoing list on every step and on every boundary probe, and the open walk and the path walk keep two independent spellings of "the edges a forward step follows". #827's whole point is that path mode's eligibility must be the open trace's; `pathEndpoint` already delegates to `eligibleTraceEndpoint` for the endpoint half, and the candidate half should come from the same place for the same reason.

Call `traceEdges` with `reverse: false`. Identical edges, one construction, and path mode inherits the short-circuit.

### Verified, no change

**Handing back the stored array is safe.** Round 3 made `traceEdges` return `graph.incoming(id)` / `graph.outgoing(id)` itself when nothing is synthesized. No caller mutates it: `orderedEdges` sorts a copy (`[...edges].sort`), the depth probe only calls `.some`, and the traversal loop only reads. The return type is `readonly ITtscGraphEdge[]`, so a future mutation would not compile either.

**The tests hold up under a fresh read.** Every positive has its negative twin, and every negative is anchored: the abstract candidate is refused at a node the walk is asserted to stand on, the doubly related implementation is asserted to be dispatched to exactly once (two hops with identical evidence on the pre-dedupe build), the concrete-base pair differs by one statement and must answer alike, the hub is silent in both directions while a non-hub declaration crosses in four direction/focus combinations on the same graph, and the bounded verdicts are compared against the two exhausted verdicts from the same run rather than against a literal message.

**Everything from rounds 1 through 3** re-verified against this head.

### Plan

Apply 1, rebuild, re-run the three cases then the whole suite and both typechecks, push, pass the gate, and start round 5.

samchon:
## Self-Review, round 5: clean

Fresh full round over `2af2862`, the whole base-to-head diff read again from the top. No sound improvement and no missing case survived verification, so this is the closing round.

Re-verified this round, each from the current state rather than from the earlier round's notes: the relation-first gate and its early return; the per-implementation dedupe and the hub count it feeds; the ownership walk and its cycle guard; the `.d.ts` factoring in `pathPolicy` and the unchanged `isSupportPath` result; the reverse builder's reuse of the forward set; `traceEdges` returning the stored array unmutated by any caller; `findPath` taking its forward step from `traceEdges`; the boundary probe's shared eligibility and the visited rule; `boundedPathReason` at, below, and above the clamp (a request of 50 clamps to 12 and is narrated as 12, which is what was applied); the junction suppression against the protected prose that scopes junctions to an established absence; `truncated` left untouched; and the two documentation paragraphs.

### The flaky case is identified

The single unattributed full-suite failure disclosed in the verification comment recurred once and named itself: `test_ttscgraph_native_request_response_clears_watchdog`, failing with `@ttsc/graph: native snapshot request timed out after 500 ms`.

It is unrelated to this batch by construction. It drives the fake native session (`internal/nativeSessionFake`), never loads a compiler, never calls `runTrace`, and asserts that a response clears a deliberately tiny 500 ms watchdog. It fails when the fake child needs more than half a second to answer on a loaded machine, and both failures happened on the two slowest suite runs of the session. On a re-run in isolation it passed, taking 24,325 ms for a case that takes well under a second when the machine is idle, which is the load it was competing with.

That accounts for both failures. Every full-suite run since is green.

### Final verification on the merge candidate

| command | result |
| --- | --- |
| `pnpm --filter @ttsc/graph build` | `bundled viewer: 654 KB`, exit 0 |
| `npx tsc --noEmit -p packages/graph/tsconfig.json` | clean |
| `npx tsc --noEmit -p tests/test-graph/tsconfig.json` | clean |
| `pnpm --filter @ttsc/test-graph start -- --include=native_request_response_clears_watchdog` | `Success`, 24,325 ms |
| `pnpm --filter @ttsc/test-graph start` (42 cases) | `Success`, 159,589 ms |

Protected contract re-checked on the merge candidate: `git rev-parse HEAD:packages/graph/src/structures` is `3c7399e3ce9f2404e64d652c7f1259ebd0db2613` and `ITtscGraphApplication.ts` hashes to `6AB7451FCD17E8A8C1F0F7BAEAA6A96AA4E6DEDCB379F70F030ECEE64197BE6C`.

Every push in this branch passed the exact-SHA cancellation gate, and each table is recorded under `## Batch B1` in the campaign's `ci-state.md`. Marking ready and merging.

Issue #822, opened 2026-07-20
fix(graph): make the viewer's path projection injective for every path mix a dump can contain


samchon:
## Problem

The Graph viewer reduction projects each node's file onto a display path and rewrites node ids with it. That projection is not injective, so two distinct declarations in two distinct files can end up with one identical viewer id.

Two reachable shapes produce it:

- **Disjoint absolute roots.** A legacy absolute dump containing `/a/foo.ts#same:function` and `/b/foo.ts#same:function` has no non-empty common directory, so `relativize` falls back to the basename and both become `foo.ts#same:function`.
- **A mixed dump the current producer can emit.** The native producer deliberately emits different path forms from one valid project: in-project sources are project-relative, package sources become a `node_modules/...` tail, and other out-of-project sources stay absolute. The reducer chooses one global mode from the **first** project file. When that first file is absolute, it computes a common root across a mix of absolute and relative directory strings, that root comes out empty, and the basename fallback then collapses `src/main.ts` and `src/nested/main.ts` onto one `main.ts`.

Expected behavior: distinct source files receive distinct display paths, distinct nodes receive distinct viewer ids, and every edge endpoint resolves back to the node it was drawn from.

Affected: all three copies of the reduction. `ttsc-graph view` (Node, `packages/graph/src/reduce.ts`), the website's uploaded-dump viewer (`website/src/components/graph/TtscWebsiteGraphReduce.ts`), and the benchmark viewer (`experimental/benchmark/graph/viewer.mjs`).

## Evidence

All anchors below are unchanged between the audited commit `2b724664eee71c57cd688c55122cac137774bf62` and `master` at `612abc2d20d3f99429319081c1f6520bbe34ac7a`.

### Root cause

`packages/graph/src/reduce.ts::reduce` picks one global mode:

```ts
const projectFiles = raw.nodes.filter((n) => !n.external).map((n) => n.file);
const root =
  projectFiles.length > 0 && isAbsolute(projectFiles[0]!)
    ? commonRoot(projectFiles.map(directoryOf))
    : null;
```

`commonRoot` intersects directory prefixes across the whole list, so a single absolute out-of-root entry mixed with relative entries reduces the shared prefix to the empty string. `relativize` then ends with an unconditional basename fallback:

```ts
const nm = a.lastIndexOf("node_modules/");
if (nm >= 0) return a.slice(nm);
const slash = a.lastIndexOf("/");
return slash >= 0 ? a.slice(slash + 1) : a;
```

[...]

Issue #783, opened 2026-07-19
fix(graph): give cross-project sources portable collision-free wire identities


samchon:
## Problem

`ttscgraph` uses one `dumpContext.rel()` string vocabulary for every identity-bearing file field:

- project source: project-relative
- `node_modules`: last `node_modules/` tail
- compiler virtual lib: `bundled:///...`
- any other source outside `project`: host absolute path

This makes node IDs and facts checkout-dependent for sibling/project-reference sources, and the tail-only package rule can collapse two resolved package versions/peer contexts to the same `node_modules/pkg/...` identity. The same mapping reaches node files/IDs, module IDs, implementation/evidence, edge endpoints, diagnostics, source manifests, and universe roots.

The now-complete manifest from #781 makes these cross-root identities more visible; it does not make them portable or injective.

## Acceptance

Define a producer-owned schema-v6 wire vocabulary that is stable and collision-free without using raw absolute filesystem paths as identity:

- in-project source remains project-relative;
- compiler virtual source remains `bundled:///...`;
- project-reference/sibling source receives a stable workspace coordinate;
- package source receives a coordinate that distinguishes name/version/resolution context and subpath;
- an arbitrary cross-volume source without a configured logical root fails precisely instead of being disguised by a hash or basename.

One cached mapper must own every identity-bearing path and reject a collision before serialization.

## Required tests

1. Build the same sibling `.ts` (`external:false`) and sibling `.d.ts` (`external:true`) fixture under two checkout roots; excluding the local `project` locator, IDs/files/endpoints/evidence/diagnostics/manifest/universe are byte-identical.
2. Cover POSIX, Windows drive, and UNC layouts; configured cross-drive roots are stable or rejected precisely.
3. Two package versions or peer-resolution contexts with the same subpath remain distinct.
4. Bundled identities remain stable.
5. Every field listed above uses the same mapping.
6. Distinct compiler sources can never serialize to one wire identity.
7. Update Go/TypeScript schema drift tests and consumer mismatch diagnostics.

## Downstream

samchon/graph#104 tracks adapter/viewer support and collision-free reduction. Related history: #723/#725 (schema/provenance), #761/#778 (legacy absolute viewer paths), #259/#260 (external package coordinates).

PR #865, opened 2026-07-20
fix(graph): verify source display against reproducible evidence and recover identity from structured facts

### LINKED ISSUE #810: fix(graph): recover Graph identity from structured facts, not from splitting a display id on `#` or `.`
## Problem

A `@ttsc/graph` node id is built by raw string concatenation and then taken apart again by six independent parsers, each guessing where one component ends and the next begins. Two ordinary, valid inputs break that guess in opposite directions.

A `#` inside a **file path** breaks every consumer that splits at the first `#`. For a project rooted at `/work/a#b`, the node id `/work/a#b/src/main.ts#main:function` is read as path `/work/a` plus symbol `b/src/main.ts#main:function`. The dump then fails to make the id project-relative (so ids keep an absolute machine path and are not portable between checkouts), the edge-evidence lookup misses the source map and drops spans that exist, and graph-backed `textDocument/references` compares `/work/a` against real source paths and loses every usage from that file.

A `#` inside a **symbol name** breaks the one consumer that splits at the last `#`. The producer deliberately restores authored private-member names, so `Counter.#count` is a real, intended node name. `resolveHandle`'s stale-handle fallback takes the text after the last `#`, so the handle `src/main.ts#Counter.#count:variable` resolves to a symbol called `count`, which does not exist. The promised recovery for a stale id silently cannot find any private member.

A `.` inside a **quoted member name** breaks ownership. The producer sends both `name: "a.b"` and `qualifiedName: "Box.a.b"` precisely so the boundary does not have to be guessed. The TypeScript memory layer discards `name` and recomputes the owner by cutting the qualified name at its last dot, so it looks for a container called `Box.a`, finds nothing, and attaches the member to the **file** instead of to `Box`. The same failed lookup also skips the variable-to-property refinement, and `details(Box)` returns the class with no members at all.

Expected behavior: file identity, owner identity, simple name, qualified name and kind are already structured facts at serialization time, and no consumer should have to recover them by scanning a display string for a delimiter that can legally appear inside a component.

Affected: every `@ttsc/graph` consumer. The MCP tools (`details`, `trace`, `tour`, `lookup`, `entrypoints`, `overview`), the graph-backed LSP symbol/reference provider, the `ttsc-graph view` viewer, the website viewer, the benchmark viewer, and anything that stores or compares node ids across runs.

## Evidence

All anchors below are unchanged between the audited commit `2b724664eee71c57cd688c55122cac137774bf62` and `master` at `612abc2d20d3f99429319081c1f6520bbe34ac7a`.

### The ambiguous join

`packages/ttsc/internal/graph/graph.go::nodeID` is the only producer of a node id:

```go
func nodeID(path string, name string, kind NodeKind) string {
	return path + "#" + name + ":" + string(kind)
}
```

There is no escaping and no length framing for either component.

### The six parsers that disagree

| Consumer | Symbol | Split | Cannot represent |
| --- | --- | --- | --- |
| `packages/ttsc/internal/graph/dump.go` | `dumpContext.relID` | first `#` | a path containing `#` |
| `packages/ttsc/internal/graph/dump.go` | `nodeFile` (used by `edgeEvidence`) | first `#` | a path containing `#` |
| `packages/ttsc/internal/graphsymbols/provider.go` | `nodeFile` | first `#` | a path containing `#` |
| `packages/graph/src/model/TtscGraphMemory.ts` | `fileOfNodeId` | first `#` | a path containing `#` |
| `packages/graph/src/reduce.ts` (mirrored in `website/src/components/graph/TtscWebsiteGraphReduce.ts` and `experimental/benchmark/graph/viewer.mjs`) | `rewriteId` | first `#` | a path containing `#` |
| `packages/graph/src/server/resolveHandle.ts` | `symbolPartOf` | last `#` | a symbol containing `#` |

Both inputs are real. `#` is a legal character in POSIX and Windows file and directory names. Authored private members are deliberately preserved as `#field`: `packages/ttsc/internal/graph/build.go::stripPrivateMangling` exists to remove the checker's per-run counter from `__#41@#field` and restore the authored `#field`, and its doc comment explains that the number "identifies nothing a reader can use ... and `#field` is what the source calls it". The Go dump test `dump_is_a_function_of_the_source_test.go` already builds classes with `#count` and `#step`.

### The discarded simple name

`packages/ttsc/internal/graph/graph.go` documents the invariant on the `Node.Simple` field itself:

> Simple is the unqualified declared name (`create`, `OrderService`), taken straight from the declaration's symbol. Name may join an owner chain to a member with a single dot, and a quoted member name can itself contain a dot (`"a.b"` -> Name `C.a.b`), so the simple/qualified boundary cannot be recovered from Name by splitting on a dot. Recording it here keeps the dump split exact instead of guessing.

`packages/graph/src/model/TtscGraphMemory.ts` then guesses anyway, in `synthesize`:

```ts
function keyOf(node: ITtscGraphNode): string {
  return node.qualifiedName ?? node.name;
}

/** The owner key of a dotted key (`A.B.c` -> `A.B`), or "" for a top-level key. */
function ownerKey(key: string): string {
  const dot = key.lastIndexOf(".");
  return dot >= 0 ? key.slice(0, dot) : "";
}
```

`owner(node)` looks that prefix up in `byFileKey`, and the result drives both property refinement (`variable` becomes `property` only when the owner resolves to a class or interface) and the `contains` ownership tree (`const container = parent ? parent.id : fileNodeId(node.file)`). `packages/graph/src/server/runDetails.ts` enumerates a container's members exclusively from outgoing `contains` edges, so a misrouted owner becomes a user-visible empty member list.

**Quoted dotted members.** Build a project containing:

```ts
export class Box {
  "a.b": number = 1;
  "m.n"(): number { return this["a.b"]; }
}
```

Load it through `packages/graph/lib/model/loadGraph.js` with the platform `ttscgraph` binary, then inspect the incoming `contains` edges of both members and call `runDetails(graph, { type: "details", handles: [boxId] })`. Observed:

```json
{
  "box": "src/index.ts#Box:class",
  "property": {
    "id": "src/index.ts#Box.a.b:variable",
    "owner": ["index.ts"]
  },
  "method": {
    "id": "src/index.ts#Box.m.n:method",
    "owner": ["index.ts"]
  }
}
```

`details` for `Box` returned **no** `members`. Expected: both members owned by `Box`, the field refined to `property`, and both present in the class details.

**A `#` in the project path.** Build the same kind of project (a `main()` calling a `dep()`) under a directory whose name contains `#`, for example `project#tag`, and dump it. Observed: node ids retain the absolute root (`C:/.../project#tag/src/index.ts#main:function`) instead of being project-relative, and the surviving `main -> dep` call edge carries **no** `evidence` even though the call expression has a source span.

The path half is reproducible without building the graph, by applying the shipped formulas directly:

```bash
node -e "const p='/work/a#b',f=p+'/src/main.ts',id=f+'#main:function'; const h=id.indexOf('#'); const rel=x=>x.startsWith(p+'/')?x.slice(p.length+1):x; const wire=rel(id.slice(0,h))+'#'+id.slice(h+1); console.log({file:rel(f),wire,recovered:wire.slice(0,wire.indexOf('#'))})"
```

Observed:

```text
{ file: 'src/main.ts', wire: '/work/a#b/src/main.ts#main:function', recovered: '/work/a' }
```

The same node's separately serialized `file` field is correctly `src/main.ts`, so the dump gives one source two identities.

**Private-member handle.** `symbolPartOf("src/main.ts#Counter.#count:variable")` takes the text after the **last** `#`, giving `count:variable`, then strips at the last `:`, giving `count`. The stale-handle fallback therefore searches for a symbol that does not exist. Exact current ids still resolve because the id map is consulted before the fallback, which is why the existing id-stability test does not detect this.

### Repair history

This is not the first pass over this cause, and the earlier pass wrote the correct rule down.

Commit `21f84362e` ("feat(graph): emit node modifiers; fix qualified-name split and path root") added `Node.Simple` **specifically because** a quoted dotted name cannot be recovered by splitting a qualified name, and put that reasoning in the struct comment quoted above. The producer was corrected and now serializes `name` and `qualifiedName` exactly. The consumer that had to act on that fact, the ownership synthesis in `TtscGraphMemory`, kept the same last-dot split it had carried since the initial graph-memory engine (`520f3093b`, one day earlier), and the tests added with the fix asserted the serialized `name` and `qualifiedName` values rather than the ownership, refinement and `details` consequences that depend on them. The repair sealed the witness that surfaced (the serializer) and left the live member of the same class (the projection) with the invariant already written into the producer's own comment beside it.

The delimiter half has no repair history at all: raw node ids originated in `25ccfcc9ae04` and the first-`#` dump parsing in `46d5b03e44`, viewer parsing in `bd0fba6dd4`, graph-backed LSP parsing in `a24f457665`, and the TypeScript memory's `fileOfNodeId` in `74af5c587`. No commit in that lineage added escaping, and no test uses a `#` path.

## Consequence surface

| Input | Layer that misreads it | Observable failure |
| --- | --- | --- |
| Project root, directory, or filename containing `#` | `relID` | Node ids keep the absolute host path; dumps and stored ids are machine-specific and not comparable across checkouts |
| Same | `edgeEvidence` via `nodeFile` | Edge `evidence` spans are omitted even though the span exists |
| Same | `graphsymbols/provider.go::nodeFile` | Graph-backed `textDocument/references` and edge cursor targeting lose every usage in that file |
| Same | `TtscGraphMemory.fileOfNodeId` | Edge spans are re-anchored on a truncated file, so the reconstructed `evidence.file` is wrong |
| Same | `reduce.ts::rewriteId` and its two mirrors | Viewer rerooting rewrites the wrong prefix |
| Authored `#private` member in a stale handle | `resolveHandle.symbolPartOf` | The documented stale-id fallback resolves to a nonexistent symbol |
| Quoted member name containing `.` (`"a.b"`, `"m.n"`) | `TtscGraphMemory.ownerKey` | Member is contained by the file, not the class; `details(container)` omits it; a quoted property stays `variable` instead of `property`; tour and entrypoint containment classify it as a top-level declaration |

Cross-cutting properties of the whole class:

- Silent. Nothing errors, and every affected result is schema-valid, so a resident refresh or a one-shot reload reproduces the same wrong facts rather than invalidating them.
- Platform-independent. `#` in a path is legal on Windows and POSIX; quoted dotted members are legal TypeScript everywhere.
- Persistent. Wrong ids and missing evidence cross the native JSON boundary into every TypeScript projection, and no later layer holds enough information to repair them.

## Approach

The invariant: **a component of a Graph identity is recovered from the structured fact that produced it, never by scanning a joined display string for a delimiter that can legally occur inside a component.**

Two architectural owners:

1. **Id component grammar.** One canonical codec owns the join and the split. Escape the reserved characters (at least `#` and the escape character itself) in the path and name components before joining, and decode only where a consumer needs the raw path or name. Ids that contain no reserved character must stay byte-identical, so the documented ordinary form `path#Qualified.Name:kind` and every published example remain valid. Every parser in the table above must move to that codec together: repairing `dump.go` alone leaves the LSP provider, the memory loader, the handle resolver and the three viewers divergent.
2. **Owner derivation.** Ownership synthesis derives the owner from the producer's already-exact simple name (or an equivalent explicit relation), validates the expected `owner + "." + simple` relationship, and fails closed on malformed synthetic input rather than silently attaching a member to its file.

Do not add a second heuristic on the consumer side, and do not change the separately serialized `file`, `name`, `simpleName` and `qualifiedName` values, which are already unambiguous and must stay unescaped.

Positive:

- A project whose root, an intermediate directory, and a filename each contain `#` produces project-relative ids, preserves node evidence, preserves edge evidence, resolves graph-backed LSP references, and reduces in the viewer without rewriting the wrong prefix.
- A stale handle naming an authored private member (`file.ts#Counter.#count:variable`) resolves through the stale-id fallback to `Counter.#count`.
- A quoted dotted field and a quoted dotted method are contained by their declaring class, appear in `details(class)`, and the field is refined to `property`.

Negative twins:

- Genuine nested owners keep working: `namespace A.B { export class C { m() {} } }` still resolves `A.B.C.m` to `A.B.C`.
- A member whose owner truly does not exist in the file does not get silently attached to a wrong container.
- Ids containing no reserved character are byte-identical to today's output, proven against a stored fixture.

Boundaries: a path containing the escape character itself; a path containing both the escape character and `#`; a module id whose name is itself a path; a symbol containing `:`; an id with no path component; external and `node_modules` paths; POSIX, Windows drive and UNC spellings.

Regression: the existing id-stability test (`packages/ttsc/internal/graph/dump_is_a_function_of_the_source_test.go`), the graph-backed LSP cases under `packages/ttsc/test/driver` (`run_lsp_answers_references_for_a_symbol_test.go`, `run_lsp_answers_document_symbol_for_declarations_test.go`), and the resolver ranking suite under `tests/test-graph/src/features/test_ttscgraph_resolver_*.ts` must pass unchanged. Extend the resolver's stale-id fallback case rather than replacing it.

Proving commands:

```bash
# narrow, native producer and dump parsers

# narrow, graph-backed LSP references (the graphsymbols tests live here, not beside the source)
cd packages/ttsc && go test ./test/driver -count=1

# narrow, TypeScript consumers

# broader
pnpm test
```

New test cases follow the repository contract: one case per file, one `Test*` per Go test file, Go cases under `packages/ttsc/internal/graph` for the producer and `packages/ttsc/test/driver` for the editor path, and one exported `test_<snake_case>` function per matching filename under `tests/test-graph/src/features/`. Assert the downstream membership and evidence, not only the serialized `name` fields, so one layer cannot regress independently.

## Coordination

- **Protected Graph contract.** `packages/graph/src/structures/ITtscGraphApplication.ts` (SHA-256 `6AB7451FCD17E8A8C1F0F7BAEAA6A96AA4E6DEDCB379F70F030ECEE64197BE6C`), every detailed and public subordinate type in that application contract, and all of their JSDoc must remain **byte-for-byte unchanged**. This fix lives entirely in internal codec and parser helpers, the memory synthesis, the native producer, and tests. Note that the protected file is not currently Prettier-clean, so a blanket `pnpm format` would rewrite it: exempt it explicitly from any formatting step.
- **#783** owns the cross-project file-coordinate vocabulary (a schema-v6 wire vocabulary for sibling, project-reference and package sources). This issue is adjacent, not a duplicate: #783's acceptance can be met in full while an in-project `src/a#b.ts` still tokenizes wrongly and the private-name handle fallback still fails, because neither is a cross-root coordinate question. Land them in a compatible order and make the coordinate mapper use this issue's codec rather than re-implementing a split.
- **Sibling issue in this batch, #822** (viewer path projection) edits `packages/graph/src/reduce.ts`, which also holds `rewriteId`. Whichever lands second must adopt the other's mapping instead of adding a parallel one. Sequence them under one owner.
- **Sibling issue in this batch, #812** (ECMAScript line terminators) edits `dump.go` and `graphsymbols/provider.go` as well, but in their line and comment helpers rather than their id parsers. No functional overlap; expect a merge-order dependency in those two files.
- Closed #259 and #275 established the `path#qualifiedName:kind` identity but never specified escaping or a structured boundary. #761 and PR #778 repaired common-root calculation for legacy absolute paths, not id tokenization. Targeted searches for graph node-id delimiters, hash-bearing paths, quoted dotted members and private-field handles found no open or closed duplicate.

### LINKED ISSUE #812: fix(graph): count every ECMAScript line terminator so Graph coordinates match the compiler's
## Problem

Graph derives source coordinates twice, in two layers, and neither uses the line-terminator set the compiler uses. The native dump counts only `\n` when it turns a byte offset into a line and column. The TypeScript source reader splits the verified source text on `/\r?\n/` when it builds the line array that signatures and JSDoc are sliced from. TypeScript-Go counts LF, CR, CRLF, `U+2028` (LS) and `U+2029` (PS).

For a file that uses lone CR as its terminator, the native dump reports every declaration on line 1 with a column equal to its byte offset, a leading `//` comment never terminates so evidence trimming runs to end of file, and `details` returns the entire file as the `signature` of several unrelated symbols.

For a file that uses LS or PS, the native side is correct (the compiler computed the evidence) but the reader disagrees: it returns one line where the compiler counted several, so `signatureOf` and `docOf` index past the end of their array and return `undefined`. The symbol keeps its correct line number and loses its signature and documentation entirely.

Expected behavior: a project written with LF, CRLF, lone CR, LS or PS produces semantically identical graph evidence, signatures and documentation. Nothing about the choice of line terminator is a supported reason for a Graph answer to change.

Affected: every consumer of Graph source display. `details`, `lookup`, `entrypoints`, `trace` (path mode summaries) and `tour` all read signatures through the same helper, and tours additionally read declaration docs. The same line model is reimplemented a third time in the graph-backed LSP provider that answers `textDocument/documentSymbol` and `textDocument/references`, where the two directions of its own offset-to-position conversion already disagree with each other for a CR-terminated file.

The failure is silent by construction. The reader's provenance gate hashes the file's bytes against the snapshot's `checkerDigest`, and a CR-only or LS-bearing file passes that gate, because the bytes really are the bytes the checker read. Provenance therefore positively authorizes the wrong slicing rather than blocking it, and a resident refresh reproduces it.

## Evidence

All anchors below are unchanged between the audited commit `2b724664eee71c57cd688c55122cac137774bf62` and `master` at `612abc2d20d3f99429319081c1f6520bbe34ac7a`.

### What the compiler counts

The pinned `github.com/microsoft/typescript-go@v0.0.0-20260429010842-56ab4af42157` computes line starts in `internal/core/core.go::ComputeECMALineStartsSeq`. It advances on `\r` (consuming a following `\n` as one terminator), on `\n`, and on any rune for which `internal/stringutil/util.go::IsLineBreak` is true. `IsLineBreak` returns true for `\n`, `\r`, `U+2028` and `U+2029`, and its comment cites ES5 7.3 Table 3 as the authority. This is upstream-defined, not a convention inferred from JavaScript practice.

### What Graph counts

`packages/ttsc/internal/graph/dump.go::newLineStarts`:

```go
func newLineStarts(text string) lineStarts {
	starts := make(lineStarts, 1, 1+strings.Count(text, "\n"))
	starts[0] = 0
	for i := 0; i < len(text); i++ {
		if text[i] == '\n' {
			starts = append(starts, i+1)
		}
	}
	return starts
}
```

`dump.go::firstCodeOffset` skips a `//` comment with an inner loop bounded by `text[i] != '\n'`, and `dump.go::lineCommentEnd`, which compacts signatures, uses `strings.IndexByte(text[start:], '\n')`. Both run past a lone CR to the end of the file. The neighbouring `regularExpressionEnd` and whitespace handling in the same file already treat `\r` as significant, so CR is not an intentionally unsupported byte here.

`packages/graph/src/model/TtscGraphSourceReader.ts::lines`:

```ts
const lines: readonly string[] = Object.freeze(text.split(/\r?\n/));
```

`packages/graph/src/server/runDetails.ts::signatureOf` and `::docOf` index that array with the compiler's `evidence.startLine` and `evidence.endLine`, so a disagreement between the two line models is an out-of-range index, not a visible error.

### A third site, and a proof that CR is not deliberately unsupported

`packages/ttsc/internal/graphsymbols/provider.go`, which answers `textDocument/documentSymbol` and `textDocument/references` from the graph, carries its own copy of `firstCodeOffset` with the same `text[i] != '\n'` line-comment loop, plus its own offset-to-position conversion:

- `offsetToPosition` increments `line` only on `'\n'`.
- `lspPositionToOffset`, in the same file, advances the line counter on `'\r'` (consuming a following `'\n'`), **and** on `'\n'`.

The two directions of one conversion therefore disagree for a CR-terminated file: an editor position maps to an offset as if CR were a line break, and that offset maps back to a position as if it were not. Neither handles LS or PS. That asymmetry, and the fact that `dump.go`'s own `firstCodeOffset` already treats `\r` as skippable whitespace, show CR was never a deliberately unsupported terminator here. It was simply not counted.

**Native half.** Write a source file with literal `\r` separators and no `\n`, plus a minimal tsconfig:

```text
// lead comment<CR>export function alpha(): number { return beta(); }<CR>export function beta(): number { return 1; }<CR>
```

Run `ttscgraph dump --cwd <project> --tsconfig tsconfig.json` with the platform binary, then load the same graph with `loadGraph({ cwd, binary })` and call `runDetails(graph, { type: "details", handles: ["alpha", "beta"] })`.

Observed: the dump reported `alpha` at line 1, column 113 and `beta` at line 1, column 68. The call evidence also claimed line 1. Both `details` entries carried the same whole-file signature:

```json
"// lead comment\rexport function alpha(): number { return beta(); }\rexport function beta(): number { return 1; }"
```

`alpha` was additionally placed at end of file, because the leading line comment never terminated.

Expected: evidence on the actual declaration and call lines, and a signature restricted to each declaration.

**Reader half, without building the graph.** Construct `TtscGraphSourceReader` with a matching checker digest and an injected read function returning four logical lines separated by lone CR, by LS, or by PS, and read `lines()`. Observed: an array of length **1** for each of the three; the identical text with LF or CRLF returns length **4**. With the compiler reporting a declaration on line 4, `signatureOf` computes a start index of 3 against a one-element array and returns `undefined`, and `docOf` cannot find the JSDoc on line 3.

### Provenance

`newLineStarts` arrived with the initial full graph dump in `46d5b03e4` and kept its LF-only behavior through later evidence revisions. `firstCodeOffset` was added to move evidence past leading trivia and was later rewritten, but its line-comment loop still tests only `\n`. The `/\r?\n/` split predates the audited commit: `runDetails` used the same assumption inline before `2b724664e` centralized display extraction behind `TtscGraphSourceReader`. That commit's own regression, `tests/test-graph/src/features/test_ttscgraph_source_reader_enforces_snapshot_identity_once.ts`, uses `\n` only, so it pins snapshot identity and immutability but never the line-coordinate contract.

Searches across open and closed issues for `TtscGraphSourceReader`, graph line separators, carriage returns, and `U+2028` found no owner. Unrelated `U+2028` results concern factory printing and lint escaping.

## Consequence surface

| Terminator | Native `newLineStarts` | Reader `split(/\r?\n/)` | `graphsymbols::offsetToPosition` | Result |
| LF | correct | correct | correct | correct |
| CRLF | correct (the `\n` is counted once) | correct | correct | correct |
| Lone CR | **wrong**: the whole file is line 1 | **wrong**: one line | **wrong**: one line, and it contradicts `lspPositionToOffset` in the same file | Wrong evidence lines and columns, whole-file signatures shared by unrelated symbols, `//` comment runs to EOF, editor document symbols and references land on the wrong line |
| `U+2028` | correct (the byte scan never sees it, and the compiler's evidence is authoritative) | **wrong**: one line | **wrong**: one line | Signature and doc silently absent for declarations after the separator; editor positions wrong |
| `U+2029` | same as LS | **wrong**: one line | **wrong**: one line | same as LS |

The `//` comment termination bug is duplicated in both `firstCodeOffset` copies (`dump.go` and `graphsymbols/provider.go`), so it fires in the dump and in the editor path independently.

Additional properties:

- Every layer must be fixed. Repairing only the native side leaves the reader slicing the wrong text; repairing only the reader leaves the dump's coordinates wrong; repairing neither of the `graphsymbols` conversions leaves the editor path inconsistent with itself.
- The failure crosses the wire. Dumps persist the wrong coordinates, so every consumer that ranks or navigates by evidence line inherits them, and no later layer can repair them.
- Not platform-specific. CR-only files are uncommon but valid, and LS and PS appear in generated and minified sources on any platform.
- Not a caching or resident-session symptom. A fresh one-shot dump reproduces it.

## Approach

The invariant: **Graph's line model is the compiler's line model.** One definition of a line boundary (LF, CR, CRLF as a single terminator, LS, PS) governs both the native coordinate producer and the TypeScript display reader, so a coordinate produced by one is always valid in the other.

Architectural owners:

- `packages/ttsc/internal/graph/dump.go`: `newLineStarts` and the comment-termination helpers `firstCodeOffset` and `lineCommentEnd`.
- `packages/ttsc/internal/graphsymbols/provider.go`: its duplicate `firstCodeOffset`, and the `offsetToPosition` / `lspPositionToOffset` pair, which must agree with each other and with the compiler.
- `packages/graph/src/model/TtscGraphSourceReader.ts`: `lines`, which must produce the same logical line sequence the compiler's line starts describe.

The upstream helper is the authority for the rule. Prefer reusing the pinned compiler's line computation over hand-transcribing a second list of terminators, and prefer one shared helper over the two Go copies that already drifted.

Preserve the reader's digest gating, its immutable both-outcomes caching, and its absence behavior for unreadable or non-matching files.

Positive: four projects containing the same declarations, separated by LF, CRLF, lone CR, LS and PS respectively, produce semantically identical dumps (same declaration lines and columns, same edge evidence) and identical `details` output (same `signature`, same `doc`).

Positive: the reader's line array for each of the five spellings has the same length and content as the compiler's line starts for the same text.

Negative twins: CRLF is counted once, not twice; a `\r` inside a string literal or a template is still just a character in a line, not an extra structural boundary for the purposes of the comment scanners; a file whose bytes do not match the snapshot digest still returns no lines.

Boundaries: a file with no trailing terminator; an empty file; a file that is a single terminator; a leading `//` comment terminated by CR, LS and PS; a block comment; a regular-expression literal on a CR-terminated line; a declaration on the last line.

Graph-backed LSP positions, which must be proven on their own and not inferred from the display fixes:

- Positive: `textDocument/documentSymbol` over a CR-terminated file reports each symbol's range on the declaration's real line, for both the range start and the range end, and the same holds for LS and PS.
- Positive: `textDocument/references` returns each usage on its real line for all five spellings, and a cursor position sent by the editor selects the symbol actually under it.
- Round trip: `offsetToPosition` and `lspPositionToOffset` are inverse for every offset in a file using each of the five spellings. The current pair is not inverse for lone CR, so this case fails before the fix.
- Leading trivia: a declaration preceded by a `//` comment in a CR-terminated file gets a range at the declaration, not at end of file, which pins the second `firstCodeOffset` copy independently of the dump's copy.
- Negative twin: an LF and a CRLF file produce byte-identical LSP ranges to today's output.

Regression: the existing LF-based cases must pass unchanged. That includes `tests/test-graph/src/features/test_ttscgraph_source_reader_enforces_snapshot_identity_once.ts`, the dump evidence tests under `packages/ttsc/internal/graph`, and the editor-path cases under `packages/ttsc/test/driver`: `run_lsp_answers_document_symbol_for_declarations_test.go`, `run_lsp_answers_references_for_a_symbol_test.go`, and `lsp_symbol_provider_helpers_test.go`.

Assert all four surfaces separately: raw dump coordinates, `details` display output, the LSP position round trip, and the LSP ranges themselves. Each of the three implementations can regress on its own, so a fixture that only checks `details` would leave the editor path unpinned.

Proving commands:

```bash
# narrow, native dump coordinates

# narrow, graph-backed LSP positions (the graphsymbols tests live here, not beside the source)
cd packages/ttsc && go test ./test/driver -count=1

# narrow, TypeScript display

# broader
pnpm test
```

New cases follow the repository contract: one case per file, one `Test*` per Go test file, Go cases under `packages/ttsc/test/driver` for the editor path and `packages/ttsc/internal/graph` for the dump, and one exported `test_<snake_case>` per matching filename under `tests/test-graph/src/features/` for the display path.

## Coordination

- **Protected Graph contract.** `packages/graph/src/structures/ITtscGraphApplication.ts` (SHA-256 `6AB7451FCD17E8A8C1F0F7BAEAA6A96AA4E6DEDCB379F70F030ECEE64197BE6C`), every detailed and public subordinate type, and all of their JSDoc must remain **byte-for-byte unchanged**. The fix lives in `dump.go`, `graphsymbols/provider.go`, `TtscGraphSourceReader.ts` and tests. The protected file is not currently Prettier-clean, so exempt it explicitly from any formatting step rather than running a blanket `pnpm format`.
- **Sibling issue in this batch, #833** (verify source display against a digest the consumer can reproduce) edits the same method, `TtscGraphSourceReader.lines`. The two are independent causes (this one is the line model, that one is the encoding of the bytes being hashed), but they conflict textually. Sequence them under one owner.
- **Sibling issue in this batch, #810** (structured Graph identity) also touches `dump.go` and `graphsymbols/provider.go`, but in the id-parsing helpers (`relID`, `nodeFile`, `edgeEvidence`), not in the line and comment helpers. No functional overlap, but the two share files, so expect a merge-order dependency.

### LINKED ISSUE #833: fix(graph): verify Graph source display against a digest the consumer can actually reproduce
## Problem

`@ttsc/graph` gates every source-derived display fact behind a digest comparison, and it compares against the one digest the snapshot's own contract says a consumer cannot reproduce.

`TtscGraphSourceReader.lines` reads the file's raw bytes, decodes them with Node's `Buffer.toString("utf8")`, hashes the result, and requires it to equal the snapshot's `checkerDigest`. `checkerDigest` is the hash of the **decoded** text the checker resolved against, and the compiler decodes before hashing: it strips a UTF-8 BOM and transcodes UTF-16 to UTF-8. Node's `toString("utf8")` does neither. For a BOM-prefixed or UTF-16-encoded source, the two sides therefore hash different encodings of the same unmodified file, and the comparison can never succeed. The failure is cached, so it is permanent for the life of the graph.

The visible result is that every `signature` and every declaration `doc` disappears for that file, from `details`, `lookup`, `entrypoints`, `trace` and `tour`. The answer is well formed and simply has no signature, which is exactly the state that sends a model back to reading the file, the read the index exists to remove.

The producer already publishes the digest a consumer can reproduce. `SourceDigest.Disk` (`diskDigest` on the wire) hashes the file's on-disk bytes, and its doc comment states the purpose in as many words. The reader never reads that field.

Expected behavior: a project whose sources carry a UTF-8 BOM or are stored as UTF-16 returns the same signatures and docs as the same project without them, and a snapshot whose text genuinely differs from disk still refuses.

Affected: anyone whose TypeScript sources carry a byte order mark, which is the ordinary output of several Windows editors and of some code generators, and anyone with UTF-16 sources.

## Evidence

`packages/graph/src/model/TtscGraphSourceReader.ts` and `packages/ttsc/internal/graph/provenance.go` are byte-identical between the audited commit `2b724664eee71c57cd688c55122cac137774bf62` and `master` at `612abc2d20d3f99429319081c1f6520bbe34ac7a`.

### Root cause

`packages/graph/src/model/TtscGraphSourceReader.ts::lines`:

```ts
const expected = this.checkerDigests.get(key);
...
text = this.read(path.resolve(this.project, file)).toString("utf8");
...
if (sha256(text) !== expected) {
  this.cache.set(key, undefined);
  return undefined;
}
```

`checkerDigests` is populated from `source.checkerDigest`, and both outcomes are cached, so one adjudication is permanent for the snapshot.

`checkerDigest` is produced by `packages/ttsc/internal/graph/provenance.go::NewProvenance` as the SHA-256 of `SourceFile.Text()`, which is the decoded text. The decode happens upstream, before the parser sees the file: `packages/ttsc/driver/host.go::DefaultFS` builds `bundled.WrapFS(cachedvfs.From(osvfs.FS()))`, and in the pinned `github.com/microsoft/typescript-go@v0.0.0-20260429010842-56ab4af42157`, `internal/vfs/osvfs/os.go` reaches `internal/vfs/internal/internal.go::decodeBytes`, which returns UTF-16LE or UTF-16BE decoded to UTF-8 when it sees the corresponding byte order mark, and strips a three-byte UTF-8 BOM otherwise.

`provenance.go`'s own `SourceDigest` doc comment states the intended split, and names the field the consumer should use:

> Publishing both lets a consumer verify its own read against Disk and still know, from Checker != Disk, that the checker saw augmented text and byte-identity with the facts is not available for that file.

`TtscGraphSourceReader` reads only `checkerDigest`, and `diskDigest` is not referenced anywhere in the class.

Run the baseline `ttscgraph` binary on a project with one BOM-prefixed source and one plain control, then hash the same files three ways (the manifest's `checkerDigest`, the consumer's own `readFileSync(...).toString("utf8")`, and the BOM-stripped text):

```text
src/bom.ts   checker e9ed18f3009a / consumerRead 6aaf7afe844c / bomStripped e9ed18f3009a
src/plain.ts matches on all three
```

The checker's digest equals the hash of the BOM-stripped text. The consumer's read does not, so the gate can never pass for that file.

An independent probe drove the reader class directly, with the snapshot's own published digest pair for one unmodified file:

```text
checkerDigest: 2f2d8c8a1fb7189a2c1db43974a1e55d785a240b35dad8094598ed05ed8a9954
diskDigest   : cffbf79a577c968ab243fb6e448e5e5473282c7b159b55d176ac74b543776af5
BOM   file lines(): undefined
plain file lines(): [ '/** Doc. */', 'export const x = 1;', '' ]
```

Resident behavior, measured on the same two projects with no edit between requests:

```text
BOM project   : initial -> incremental (changed: true) -> unchanged
plain control : initial -> unchanged -> unchanged
```

The BOM project reports one spurious `incremental` generation with no source edit and then converges on the third request. The resident consequence is therefore bounded: one extra generation, not permanent churn.

### Consumers of the missing lines

`packages/graph/src/server/runDetails.ts::signatureOf` and `::docOf` both return `undefined` when `graph.source.lines(file)` is `undefined`, so `details` omits `signature` and `doc` for every symbol declared in the affected file **and** for every `member` of a container declared there. `runTrace.ts::summary(..., withSignature = true)` loses `signature` for the same nodes in path mode, and the tour inherits both losses. Every entry point that builds a `TtscGraphMemory` is affected: the one-shot `loadGraph` path and the resident `TtscGraphSession` path alike. The viewer reducers do not read source lines, so `ttsc-graph view` is unaffected.

### Provenance

The gate arrived with the audited commit `2b724664e` ("Keep graph details faithful to compiler snapshots"), which centralized display extraction behind `TtscGraphSourceReader` and its immutable digest adjudication. Before that commit `runDetails` read lines from disk ungated, so BOM-prefixed files displayed their signatures. This is therefore a regression introduced by the centralization, not a longstanding gap.

The same encoding class has been fixed three times in three unrelated readers: closed #216 (JS tsconfig readers rejecting a BOM-prefixed `tsconfig.json`), closed #505 (the lint formatter settings reader), and closed #621 (the LSP proxy's dirty-buffer classification). None of them touched graph provenance, and no issue mentions `checkerDigest`. Closed #764 is the performance line cache this reader replaced. Merged #781 completed the source manifest, which increases the number of files whose display depends on this gate but does not change the comparison. #780, #782, #783 and #786 concern manifest completeness, release, wire identity, and details caps.

## Consequence surface

| Source encoding | `checkerDigest` is the hash of | Consumer's `toString("utf8")` yields | `lines()` | Display facts |
| UTF-8, no BOM | the file's bytes | the same bytes | lines | present (correct today) |
| UTF-8 with BOM | the BOM-stripped text | text with a leading `U+FEFF` | `undefined`, cached | `signature` and `doc` gone everywhere |
| UTF-16LE with BOM | the UTF-8 transcoding | mojibake | `undefined`, cached | same |
| UTF-16BE with BOM | the UTF-8 transcoding | mojibake | `undefined`, cached | same |
| Any encoding, file edited after the snapshot | the snapshot's text | current text | `undefined`, cached | correct behavior, must be preserved |
| `SourcePreamble` plugin injected text (`checkerDigest != diskDigest`) | the augmented text | the disk text | `undefined`, cached | correct behavior, must be preserved |
| Snapshot without the `diskDigests` capability | the checker text | the disk text | must not become newly permissive | must stay conservative |

The BOM case is not merely a lost convenience: a BOM shifts only the first column of line 1, and a UTF-16 file has identical line structure once decoded, so both are cases where the consumer could have verified its read and is instead refused.

## Approach

The invariant: **a consumer verifies its own read against evidence it can reproduce, and only then decides whether the bytes it holds are the bytes the facts were computed from.**

The architectural owner is `packages/graph/src/model/TtscGraphSourceReader.ts`. Everything needed is already on the wire: `diskDigest` is published per source, and `CapabilityDiskDigests` already distinguishes "the producer did not hash the disk" from "the file could not be read". If the producer must publish more for a complete fix, `packages/ttsc/internal/graph/provenance.go` is the second owner.

Two properties the fix must have:

- The comparison must be made against a digest of the same domain as the consumer's read, so an encoding difference the compiler resolves cannot be mistaken for a byte difference.
- The text the reader slices must be the text the compiler's coordinates describe. Passing the gate is not enough; the line array must correspond to the checker's line model for that file.

Exclusion: do not weaken the preamble case. A snapshot whose `checkerDigest` differs from its `diskDigest` because a plugin injected text must still refuse, because its line numbers do not describe the file on disk. Preserve the reader's immutable both-outcomes caching and its absence behavior for unreadable files.

Positive: a project whose source carries a UTF-8 BOM returns `signature` and `doc` in `details` identical to the same project without the BOM. A UTF-16LE source behaves the same. A UTF-16BE source behaves the same.

Negative twins: a file whose disk bytes were edited after the snapshot still returns no `signature` and no `doc`. A `SourcePreamble` project whose checker text differs from disk by injected lines still returns none.

Boundary: an empty file; a file whose entire content is a BOM; a snapshot that does not declare `diskDigests`; a file present in the manifest with an empty `diskDigest` (it vanished mid-load, or it is a virtual source such as `bundled:///`); a file missing from the manifest entirely.

Regression: `tests/test-graph/src/features/test_ttscgraph_source_reader_enforces_snapshot_identity_once.ts` continues to pass, including the once-only adjudication and the immutability of the returned array.

Resident check: the two-request sequence on a BOM project must not report a generation change with no edit.

Proving commands:

```bash
# narrow

# if provenance.go participates

# broader
pnpm test
```

New cases follow the repository contract: one case per file, one exported `test_<snake_case>` per matching filename under `tests/test-graph/src/features/`, placed alongside the existing source-reader case.

## Coordination

- **Protected Graph contract.** `packages/graph/src/structures/ITtscGraphApplication.ts` (SHA-256 `6AB7451FCD17E8A8C1F0F7BAEAA6A96AA4E6DEDCB379F70F030ECEE64197BE6C`), every detailed and public subordinate type, and all of their JSDoc must remain **byte-for-byte unchanged**. `TtscGraphSourceReader` appears in no application-contract type, and no wire field needs to be added or renamed for the minimal fix because `diskDigest` and the digest capabilities already exist. The protected file is not currently Prettier-clean, so exempt it explicitly from any formatting step.
- **Sibling issue in this batch, #812** (count every ECMAScript line terminator) edits the same method, `lines`. The two causes are independent: this issue is the encoding of the bytes being compared, #812 is the line model applied to the text that passes. They conflict textually and together define what a correct `lines()` returns, so sequence them under one owner.
- Merged **#781** completed the source manifest, so more files (outside `.d.ts` sources, `bundled:///` virtual libraries) now carry manifest entries. Virtual sources carry an empty `diskDigest` by design; the fix must treat that as "cannot verify" rather than as a match.

samchon:
Campaign batch **B5 — Graph source display and structural identity**.

Three issues share the Graph model and native dump layers, so they land under one owner in one pull request.

- Closes #833 — `TtscGraphSourceReader` gates every source-derived display fact on a digest a consumer cannot reproduce. The compiler hashes the *decoded* text (UTF-8 BOM stripped, UTF-16 transcoded), while the reader hashes `Buffer.toString("utf8")` of the raw bytes, so `signature` and `doc` vanish permanently for every BOM-prefixed or UTF-16 source. The producer already publishes `diskDigest` for exactly this purpose.
- Closes #812 — Graph's line model is not the compiler's. `dump.go::newLineStarts` counts only `\n`; `TtscGraphSourceReader.lines` splits on `/\r?\n/`; `graphsymbols/provider.go` counts `\r` in one direction of its offset/position conversion and not in the other. TypeScript-Go counts LF, CR, CRLF, `U+2028` and `U+2029`.
- Closes #810 — Graph identity is joined by raw concatenation and recovered by six independent parsers splitting on `#` or `.`. A `#` in a path, a `#` in an authored private member name, and a `.` in a quoted member name each break a different consumer.

## Landing order

`#833` first (the digest gate is what decides whether `lines()` returns anything at all), then `#812` (the line model applied to text that passes the gate), then `#810` (identity, independent of both but sharing `dump.go`, `graphsymbols/provider.go` and the model layer).

## Constraints honored

- `packages/graph/src/structures/ITtscGraphApplication.ts` stays byte-for-byte unchanged; its SHA-256 is verified as `6AB7451FCD17E8A8C1F0F7BAEAA6A96AA4E6DEDCB379F70F030ECEE64197BE6C` before the first commit and again immediately before merge.
- No `pnpm format` on this branch; Post-Campaign Cleanup owns the repository-wide formatter result.
- Repository Actions and every workflow stay untouched. Campaign CI is suspended by exact-SHA cancellation only; local verification is recorded in the thread and in `.wiki/repository-audit-2026-07-19/ci-state.md`.

## Verification

Pending. Package-scoped local commands and their real output will be recorded here as the batch lands.

## Not in scope

#783 (cross-project file-coordinate wire vocabulary) and #822 (viewer path projection) are separate owners. #810 is deliberately scoped to the component grammar and the simple-versus-qualified name boundary that #783's acceptance leaves untouched.


samchon:
The reservation is stale and is being released under the explicit #872
handover rule.

The branch still contains only its empty claim commit
`77a64b670226157e2484c54ff4ff3a47ab760069` from 2026-07-20T03:55:50Z:
GitHub reports no changed files, no implementation commits followed, and the
repository has no worktree for this branch. The exact SHA has no Actions runs.

Closing this draft does not reject #833, #812, or #810. It returns them to the
lead-admission queue so their current evidence, dependency ordering, and
overlap can be checked before a fresh cohesive batch claims them. The resumed
campaign ledger records this disposition in
`.wiki/repository-audit-2026-07-19/candidates.md`.

Issue #771, opened 2026-07-18
fix(test): the LSP test client ignores server→client requests, so tsgo-dependent paths hang


samchon:
## Problem

The shared LSP test client in `tests/test-ttsc/src/native-plugins/server/` **ignores server→client requests**. tsgo sends `client/registerCapability` during startup and never receives a response, so any test path that needs an answer from tsgo hangs until its deadline.

Observed while building the whole-chain LSP test (#769), by probing the live wire rather than reading code:

| Path | Result |
| --- | --- |
| ttsc-owned verbs (`codeAction` with an `only` filter, `workspace/executeCommand`) | answer in ~1s |
| `textDocument/codeAction` **without** `only` | times out |
| pull `textDocument/diagnostic` | times out |

The ttsc-owned paths work because ttsc answers them itself; anything that must round-trip through tsgo stalls behind the unanswered `client/registerCapability`.

## Why it matters

It silently bounds what the LSP suite can test. Today every test in that directory is written against ttsc-owned paths, so the limitation is invisible — but it means the suite **cannot** currently assert anything about tsgo's own responses, including the merge behavior that is the proxy's entire reason to exist. A future test that reaches for one of those paths will look like a product bug rather than a harness gap.

## Related finding worth recording

While probing this, tsgo 7.0.2 was confirmed to advertise `diagnosticProvider` (**pull**) and to **never push-publish** `textDocument/publishDiagnostics`. Any test — or any reasoning — that assumes ttsc lint diagnostics arrive merged into an upstream push is asserting something that does not happen with this tsgo. The real contract the proxy must uphold is that it *merges into* tsgo's advertised capabilities (`codeActionKinds`) rather than replacing them, and does not drop `diagnosticProvider`. That is what #769 asserts.

## Approach

The invariant: **a test client that speaks LSP must answer server→client requests, or it is not speaking LSP.**

Minimally, `client/registerCapability` (and its `unregisterCapability` sibling) should get a well-formed empty result. The general shape is to route inbound frames by whether they carry an `id`: a request gets a response, a notification does not.

Owner is the shared client helper in `tests/test-ttsc/src/native-plugins/server/`. Deliberately **not** folded into #769: the change touches the helper that all five tests in that directory share, and a harness change that silently alters four existing tests deserves its own diff and its own review.

## Acceptance and verification

| Outcome | Assertion |
| --- | --- |
| Positive | A `codeAction` request with no `only` filter returns tsgo's actions plus ttsc's, within the deadline |
[...]

PR #866, opened 2026-07-20
fix(ttscserver): refresh compiler-backed LSP state on every disk signal and honor the negotiated position encoding

### LINKED ISSUE #799: fix(ttscserver): invalidate compiler-backed LSP state on clean `didOpen` and watched workspace changes
## Problem

`ttscserver` keeps two compiler-backed caches warm for the whole editor session, and exactly one editor notification refreshes either of them.

- `residentProgramCache` in `packages/lint/linthost/serve.go` holds one loaded Program per `(cwd, tsconfig, checker)` key for the daemon's lifetime. `acquire` returns the cached Program untouched on a hit; there is no TTL and no disk re-validation. Its only refresh paths are `invalidate()` (drop everything) and `applyChanges(paths)` (per-file `UpdateProgram`, with a full reload when a changed path is not a known source).
- `graphsymbols.Provider` in `packages/ttsc/internal/graphsymbols/provider.go` caches the symbol graph until `Invalidate()` is called.

In `Proxy.handleEditorEnvelope` (`packages/ttsc/internal/lspserver/lsp_proxy.go`) only the `textDocument/didSave` arm calls `invalidateSymbolProvider()` and `invalidateResidentPlugins(env)`. The `textDocument/didOpen` arm caches the buffer text and publishes plugin diagnostics without invalidating anything, and there is no `workspace/didChangeWatchedFiles` arm at all, so a watched-file notification is forwarded to tsgo and otherwise ignored.

The result is that any on-disk change the editor does not report as a save leaves both caches answering from the pre-change project for the rest of the session. A user sees lint diagnostics at wrong ranges or for code that no longer exists, quick fixes that silently do nothing, and `tsconfig.json` edits that never take effect, while tsgo's own diagnostics in the same editor are correct. The expected behavior is that a warm compiler-backed snapshot answers a request only when its text for the queried project is the text on disk right now.

Affected: every `ttscserver` and `@ttsc/vscode` user of a project with a `capabilities.lsp` sidecar, which today means every project using `@ttsc/lint` in the editor. The staleness window is the editor session, which for VS Code is days.

### The `didOpen` transition, reproduced

A Go test overlaid into `packages/ttsc/internal/lspserver` at the audited baseline drives one clean `didOpen` (the notification text equals current disk) through a proxy whose plugin source and symbol provider both count invalidations:

```go
proxy := NewProxy(ProxyOptions{
  EditorOut:      io.Discard,
  UpstreamIn:     io.Discard,
  Source:         plugins,          // counts InvalidateResidentPrograms
  SymbolProvider: symbols,          // counts Invalidate
})
handled, err := proxy.handleEditorEnvelope(Envelope{
  Method: methodDidOpen,
  Params: params,                   // textDocument.text == the bytes on disk
}, nil)
```

Observed failure:

```text
clean didOpen invalidations: resident=0 symbol=0; want 1 each
```

`didOpen` is still correctly forwarded upstream (`handled == false`); the defect is only the missing invalidation.

### The control flow behind that result

`handleEditorEnvelope` treats the three notifications differently (anchors as of `612abc2d2`, for navigation):

| Arm | Calls | Line |
| --- | --- | --- |
| `case methodDidOpen` | `cacheDidOpenText`, `publishPluginDiagnosticsForDidOpen` | `lsp_proxy.go:287` |
| `case methodDidSave` | `invalidateSymbolProvider`, `invalidateResidentPlugins`, publish | `lsp_proxy.go:294` |
| `case methodDidChange` | `invalidateSymbolProvider`, cache text, mark dirty (resident deliberately untouched) | `lsp_proxy.go:300` |

`publishPluginDiagnosticsForDidOpen` calls `documentTextMatchesDisk`, marks the URI clean, and asynchronously calls `p.source.Diagnostics`. Equality with today's disk is not evidence that the cached Program was built from today's disk, so the clean branch is exactly the branch that needs the invalidation.

### The missing watched-file transport

`git grep -n didChangeWatchedFiles -- packages/` returns nothing on `612abc2d2`. `invalidateResidentPlugins` decodes only `params.textDocument.uri`, which is the `didSave` shape, and falls back to a full invalidation otherwise, so no caller can currently deliver a watched-file batch.

The repository's own VS Code client registers a watcher for precisely the files that can never arrive as a `didSave`:

```ts
synchronize: {
  fileEvents: workspace.createFileSystemWatcher(
    new RelativePattern(spec.cwd, "**/{tsconfig,jsconfig}*.json"),
  ),
```

`packages/vscode/src/extension.ts` sets its `documentSelector` to typescript / typescriptreact / javascript / javascriptreact, so `tsconfig.json` is outside it and is delivered only as `workspace/didChangeWatchedFiles`. tsgo receives the forwarded event; the lint sidecar never does.

### The code's own contract contradicts the wiring

`residentProgramCache.invalidate`'s doc comment in `packages/lint/linthost/serve.go`:

```go
// invalidate releases every cached Program. Called when the editor signals that
// a document in the project changed on disk (didSave) or opened, so the next
// verb rebuilds against current source.
```

The proxy never calls it on an open. `applyChanges`'s comment documents the intended full-reload fallback for "a config edit, or a new or removed file, which tsgo's per-file `UpdateProgram` cannot express", but that fallback is unreachable because the events that would trigger it have no transport.

### Root cause

The set of signals wired into the two compiler-backed caches is `{didSave}`, while the set of ways a project's sources can change under a live server is strictly larger. `handleEditorEnvelope` treats `didOpen` as a diagnostics trigger rather than as a possible disk-snapshot boundary, and never classifies `workspace/didChangeWatchedFiles` as an editor envelope at all.

### Provenance

- The resident daemon arrived in commit `21560d99f` (PR #755), which invalidated on `didChange` and `didSave` and never on `didOpen`.
- Commit `8bfec2a48` (PR #765, closing issue #754) replaced save-time full invalidation with incremental changed-URI updates and correctly removed dirty-buffer invalidation. It did not add the open transition or a watched-file transport.
- Closed issue #754 explicitly required a full reload on `tsconfig` and root-file-set changes. The sidecar implements that fallback; the host never feeds it.
- Before #755 every LSP verb spawned a fresh sidecar that ran `loadProgram` against current disk, so none of this was reachable. This is a regression introduced with residency, not an original gap.
- `website/src/content/docs/development/concepts/protocol.mdx` documents the current `didSave`-only contract in the `lsp-serve` paragraph and must be updated with the fix.

Existing coverage opens, changes, and saves inside one editor session (`tests/test-ttsc/src/native-plugins/server/test_ttscserver_lsp_editor_session_merges_suppresses_and_fixes.ts`); nothing warms a Program, closes the document, mutates disk externally, and reopens.

## Consequence surface

| Event | Reaches tsgo | Reaches resident lint Program | Reaches graph symbol provider | User-visible result |
| `didSave` of a known source | yes | yes, incremental | yes | correct |
| `didChange` on a dirty buffer | yes | deliberately no | yes | correct, plugin output suppressed while dirty |
| Clean `didOpen` after an external edit to that file | yes | no | no | diagnostics and code actions computed over the pre-change AST, published against the new buffer |
| External edit to a closed dependency, then `didSave` of another file | yes | only the saved file is updated | no | project-wide findings computed over a stale universe |
| `tsconfig.json` / `jsconfig.json` edit (watched-file event) | yes | no | no | compiler options and root set frozen for the session |
| Source file created or deleted with no `didSave` | yes | no | no | the "unknown path means full reload" branch is never reached |
| Branch switch, `git pull`, generator, second editor, terminal `ttsc fix` | yes | no | no | wrong-position, phantom, or missing findings |

Two further consequences of the same cause:

- Quick fixes silently do nothing. `computeLSPCodeActions` returns actions carrying a `Command` rather than an embedded edit; executing it goes through `lsp-execute-command`, which is spawn-per-verb and therefore loads current disk, then compares the fresh source hash against the `SourceHash` recorded from the stale text and returns no edit. The guard prevents corruption, so the user sees a fix that does nothing.
- The playground and CLI paths are unaffected because they do not use the resident daemon. `lsp-command-ids`, `lsp-code-action-kinds`, and `lsp-execute-command` are spawn-per-verb, and a sidecar that does not implement `lsp-serve` falls back to spawn-per-verb with no staleness.

Platform-neutral: this is protocol wiring, not path or filesystem behavior.

## Approach

Invariant: a warm compiler-backed snapshot may answer a verb only when its text for the queried project equals the text on disk at that moment. Every host-visible signal that a project file changed on disk must reach both compiler-backed caches, and a change the host cannot localize must drop the entry rather than update it.

Architectural owner: the notification switch in `Proxy.handleEditorEnvelope` together with `invalidateResidentPlugins` in `packages/ttsc/internal/lspserver/lsp_proxy.go`. Downstream state owners are `NativePluginSource.InvalidateResidentPrograms` (`lsp_native_resident.go`), `residentProgramCache` (`packages/lint/linthost/serve.go`), and `graphsymbols.Provider`.

Two transports are missing and both belong to this issue, because they are one cause: the proxy decides what counts as a disk-snapshot boundary.

1. Recognize `workspace/didChangeWatchedFiles`, decode its `changes` array (URI plus change type), and deliver it to the resident invalidator and the symbol provider before any subsequent read.
2. Treat a clean `didOpen` as a snapshot boundary, before the asynchronous diagnostics request and before any local symbol or reference answer can run.

Boundaries the fix must preserve, stated as constraints and not suggestions:

- Dirty-buffer suppression stays exactly as it is. A `didChange` on a dirty buffer must continue to send nothing to the resident daemon, and a dirty `didOpen` must continue to publish an empty diagnostic set rather than invalidate.
- The incremental `didSave` path stays incremental. A save of a known, stable source must keep sending `{ "changed": ["<uri>"] }` and must not degrade into a full reload.
- Spawn-per-verb sidecars and the `lsp-execute-command` path keep their current behavior.
- Prefer localizable incremental invalidation where the host can name the changed paths; reserve full reload for changes to the root set or the configuration topology, which a per-file `UpdateProgram` cannot express.

`protocol.mdx`'s `lsp-serve` paragraph is part of the change: it currently publishes the `didSave`-only contract as the design.

| Case | Expected |
| --- | --- |
| Positive: change a project file on disk while the editor has it closed, reopen it | published `@ttsc/lint` diagnostics equal a cold `lsp-diagnostics` run over the same disk state, ranges included |
| Positive: edit `tsconfig.json` so the included file set or a compiler option changes, with only a watched-file event | the next diagnostics reflect the new configuration |
| Positive: add or delete a root source file with no `didSave` | the next request reflects the new root set |
| Positive: same warm/close/external-change/reopen sequence against the local graph symbol fallback | document symbols and references equal a cold snapshot |
| Negative twin: `didChange` on a genuinely dirty buffer | nothing sent to the resident daemon, no Program rebuild, plugin output still suppressed |
| Negative twin: `didSave` of a known source | still the incremental changed-URI path, not a full reload |
| Boundary: file created on disk and only opened, never saved | fresh results |
| Boundary: file deleted on disk while open | no stale findings for the deleted file |
| Boundary: `didOpen` whose buffer differs from disk | existing dirty path, unchanged |
| Regression | `packages/lint/test/command/resident_apply_change_matches_cold_load_test.go` and `resident_whole_corpus_matches_cold_load_test.go` keep passing |
| Unit | a proxy test with fake invalidators proving both invalidators run before diagnostics are scheduled for a clean open, and that a watched-file batch reaches the resident invalidator |

Narrow proving commands:

```bash
cd packages/ttsc && go test -count=1 ./internal/lspserver
node scripts/test-go-lint.cjs
```

Broader:

```bash
node scripts/test-go-ttsc.cjs
pnpm --filter @ttsc/test-ttsc start
pnpm --filter @ttsc/test-lint start
```

## Coordination

- Open epic #752 is the resident/incremental umbrella and lists the editor signals as wiring; it does not track this missing transport as an actionable item. This issue is a correctness child of it, not a duplicate, and #752 should stay open.
- Closed #754 and its implementation PR #765 required full reload on config and root-set changes. That requirement is only half satisfied: the sidecar has the fallback, the host has no transport for the events. Do not reopen #754; fix the transport here.
- #788 (refresh completion-hint corpora) and #789 (JSDoc completion scope) concern the completion caches, which are different state with different refresh triggers. Do not fold them in.
- Merged PR #790 touched `lsp_proxy.go` for the completion bridge only and did not change any invalidation arm.
- #819 (negotiated LSP position encoding) edits the same `handleEditorEnvelope` switch and the same conversion helpers. Assign both to one agent and sequence them, or expect a conflict in `lsp_proxy.go`.
- #771 (the LSP test client ignores server-to-client requests) can block a full round-trip end-to-end test that needs tsgo to talk back. If it does, prove the transitions at the proxy and daemon level and say so in the pull request rather than skipping the case.
- Nothing in `packages/graph` is involved.

### LINKED ISSUE #819: fix(ttscserver): use the negotiated LSP position encoding in every local conversion instead of assuming UTF-16
## Problem

LSP 3.17 negotiates a `PositionEncodingKind`. The client offers `general.positionEncodings`, the server picks one and advertises it in `capabilities.positionEncoding`, and from then on `Position.character` is counted in that encoding: bytes for `utf-8`, UTF-16 code units for `utf-16`.

The pinned typescript-go server selects `utf-8` whenever the client offers it. `ttscserver` proxies that initialize response to the client unchanged, so the client is told `utf-8` and starts sending byte columns. Every position conversion ttsc performs on its own, in the proxy and in the `@ttsc/lint` sidecar, counts UTF-16 code units unconditionally.

The two halves of a single editor response therefore disagree: tsgo's items and ranges use the negotiated encoding, and ttsc's locally added items and ranges use UTF-16. On any line containing a non-ASCII character before the cursor, incremental edits splice at the wrong byte, plugin completion matches the wrong prefix, and plugin diagnostics, related locations, code-action ranges, fixes, suggestions, and formatting edits are displayed or applied at the wrong columns.

Affected: any conforming LSP client that offers UTF-8, on any file with non-ASCII content before a position ttsc computes. ASCII-only files hide the defect completely, and clients that do not offer UTF-8 keep the correct UTF-16 default, which is why this has stayed invisible.

### The pinned compiler selects UTF-8

`github.com/microsoft/typescript-go v0.0.0-20260429010842-56ab4af42157` (`packages/ttsc/go.mod`, `packages/lint/go.mod`), `internal/lsp/server.go`:

```go
s.positionEncoding = lsproto.PositionEncodingKindUTF16
if slices.Contains(s.clientCapabilities.General.PositionEncodings, lsproto.PositionEncodingKindUTF8) {
	s.positionEncoding = lsproto.PositionEncodingKindUTF8
}
```

and the same function returns `Capabilities: &lsproto.ServerCapabilities{ PositionEncoding: new(s.positionEncoding), … }`. `internal/ls/lsconv/converters.go` then branches on that value, returning the raw byte column for UTF-8 and scanning UTF-16 code units otherwise.

### ttsc never reads it

`git grep -n -i positionEncoding -- packages/` returns nothing on `612abc2d2`.

`Proxy.augmentInitializeResult` (`packages/ttsc/internal/lspserver/lsp_proxy.go`) unmarshals the upstream `capabilities` map, mutates the entries it cares about, and re-emits the rest verbatim. The negotiated `positionEncoding` reaches the client through that passthrough while nothing in ttsc records it.

### Every local converter is UTF-16 only

| Owner | Symbol | What it converts | Consumers |
| --- | --- | --- | --- |
| `packages/ttsc/internal/lspserver/lsp_proxy.go` | `lspPositionToByteOffset` | ranged `didChange` edit positions to byte offsets in `Proxy.documentText` | incremental buffer maintenance, which feeds completion scope and live-buffer formatting |
| `packages/ttsc/internal/lspserver/lsp_completion_request.go` | `offsetForPosition` | completion request position to a byte offset | plugin completion matching |
| `packages/ttsc/internal/lspserver/lsp_completion_request.go` | `utf16Length` | filter length to the `textEdit` replace range | the replace range every compliant client applies |
| `packages/lint/linthost/lsp.go` | `byteOffsetToLSPPosition` | finding byte offsets to LSP positions | diagnostic and related ranges, fix and suggestion workspace edits, whole-document formatting edits |

Each states the assumption in its own comment, for example `offsetForPosition`: "LSP counts characters in UTF-16 code units by default, so a line holding an emoji or a CJK character past the BMP would land the cursor mid-token if counted as bytes."

The sidecar protocol has no encoding field at all, so even a proxy that learned the encoding could not tell the lint host which encoding to emit.

### Reproduction

A Go test overlaid into `packages/ttsc/internal/lspserver` at the audited baseline feeds the proxy's own converter a column a UTF-8 client would send:

```go
const text = "const 变量 = 1;"
const utf8ColumnAfterIdentifier = 12 // byte column just past 变量
const wantByteOffset = 12
got, ok := lspPositionToByteOffset(text, lspPositionWire{Line: 0, Character: utf8ColumnAfterIdentifier})
```

Observed failure:

```text
local position conversion = (16, true), want (12, true) for negotiated UTF-8
```

The arithmetic is checkable by hand. `const ` is 6 bytes and 6 UTF-16 units; `变` and `量` are 3 bytes each and 1 unit each. The cursor just past the identifier is byte 12 and UTF-16 column 8. Given the UTF-8 client's column 12, the UTF-16 algorithm spends its 12 units on `const 变量 = 1` and stops at byte 16, four bytes past the identifier. The same divergence runs in reverse for every range the lint sidecar emits.

### Root cause and provenance

There is no session-scoped owner of the negotiated encoding. Three independent changes each introduced a local UTF-16 assumption: proxy buffer conversion in `d3f261f8d` (editor format-on-save), lint position emission in `68016abd4` (the plugin bridge), and completion conversion in `8ea420492` (PR #736). Merged PR #790 added a fourth site, `utf16Length`, for the completion replace range. This is a cross-path protocol assumption that keeps being re-derived, not one stale helper.

`TestOffsetForPositionCountsUTF16` in `packages/ttsc/internal/lspserver/lsp_completion_merges_into_upstream_response_test.go` pins the default encoding and therefore cannot detect negotiation drift.

Closed #621 concerns BOM handling in the proxy's clean/dirty comparison, not position encoding. No issue or pull request mentions position encoding, UTF-8, or UTF-16 negotiation.

## Consequence surface

| Path | Local conversion | Symptom under negotiated UTF-8 |
| --- | --- | --- |
| Incremental `didChange` | `lspPositionToByteOffset` | the edit is spliced at the wrong byte, corrupting `Proxy.documentText`; every later feature that reads the cache inherits the corruption |
| Plugin completion request | `offsetForPosition` | the cursor lands mid-token, so hints disappear or match the wrong prefix |
| Plugin completion response | `utf16Length` replace range | a compliant client replaces the wrong range on accept |
| Lint diagnostics and related locations | `byteOffsetToLSPPosition` | squiggles at the wrong columns, and tsgo's own squiggles on the same line are at the right ones |
| Lint code-action ranges, fixes, suggestions | `byteOffsetToLSPPosition` | a workspace edit applied at the wrong columns, which is source corruption rather than a display problem |
| Whole-document formatting edits | `byteOffsetToLSPPosition` | formatting edits applied at the wrong columns |
| Clients that do not offer UTF-8 | unchanged | correct today and must stay correct |
| ASCII-only files | unchanged | correct in both encodings |

Combining marks, CJK, and astral characters all trigger it; astral characters are the boundary where the two counts differ by two rather than one. Platform-neutral.

## Approach

Invariant: one position encoding is negotiated per session, and every position ttsc reads or writes on that session uses it. A local converter must not carry its own encoding assumption.

Architectural owner: the initialize exchange in `packages/ttsc/internal/lspserver/lsp_proxy.go`. The proxy already brokers both halves of that exchange (`rememberInitializeRequest`, `augmentInitializeResult`), so it is the only component that sees the client's offer and the server's selection.

Two settlements are legitimate and the implementer should choose one deliberately and say which:

1. Capture the negotiated encoding at the initialize exchange, store it on the session, and thread it through every conversion listed above, including a new field on the sidecar command and resident protocols so the lint host emits ranges in the same encoding.
2. Constrain the negotiation so the session is always UTF-16, by removing UTF-8 from the offer the proxy forwards upstream, and document that ttscserver speaks UTF-16 only.

Option 2 is a smaller change but gives up the byte-oriented fast path tsgo prefers; option 1 keeps it. Either way the invariant is the same: exactly one encoding governs the session and no helper may assume a different one.

Do not fix only the completion helper. All four converters share the cause, and the sidecar boundary must carry the encoding or the lint half will silently keep emitting UTF-16.

| Case | Expected |
| --- | --- |
| Positive | initialize with `general.positionEncodings: ["utf-8", "utf-16"]`, open a document with BMP CJK and an astral character before the cursor, send an incremental edit at a byte column: the cached text matches the intended edit |
| Positive | request plugin completion after the same prefix: the matched filter and the emitted replace range land on the requested boundary |
| Positive | emit a plugin diagnostic and a fix after the same prefix: their ranges agree with tsgo's range for the same span |
| Positive | whole-document formatting edits apply at the intended columns |
| Negative twin | a client that omits UTF-8 keeps the UTF-16 default and every existing behavior is byte-identical |
| Boundary | astral character before the cursor (2 UTF-16 units, 4 bytes); combining mark; a line that is pure ASCII |
| Regression | `TestOffsetForPositionCountsUTF16` keeps passing as the UTF-16 twin, and the whole `internal/lspserver` suite passes |

Narrow proving commands:

```bash
cd packages/ttsc && go test -count=1 ./internal/lspserver
node scripts/test-go-lint.cjs
```

Broader:

```bash
node scripts/test-go-ttsc.cjs
pnpm --filter @ttsc/test-ttsc start
```

If the encoding is threaded through the sidecar protocol, `website/src/content/docs/development/concepts/protocol.mdx` documents that protocol and must be updated in the same change.

## Coordination

- #799 (invalidate compiler-backed LSP state on clean `didOpen` and watched workspace changes) edits the same `handleEditorEnvelope` switch and the same file. Assign both to one agent and sequence them.
- Merged PR #790 rewrote `lsp_completion_request.go` and added `utf16Length`. Base this work on master after that merge, and treat the new helper as one of the sites to convert rather than as settled.
- #734 (missing LSP diagnostic fields) and #751 (cross-file diagnostic URI vocabulary) touch the same diagnostic payloads but not their encoding. No conflict expected; check before editing `packages/lint/linthost/lsp.go`.
- #771 (the LSP test client ignores server-to-client requests) may limit how much of the negotiation can be driven end to end from the existing test client. Prove what can be proven at the proxy level and record the limitation.
- Nothing in `packages/graph` is involved.

samchon:
Campaign batch **B7** of the 2026-07-19 repository audit. Both issues live in `packages/ttsc/internal/lspserver/**` and its `@ttsc/lint` sidecar boundary, and both edit the same `Proxy.handleEditorEnvelope` switch and the same position-conversion helpers, so they are implemented together in one pull request.

Closes #799
Closes #819

### #799 — invalidate compiler-backed LSP state on clean `didOpen` and watched workspace changes

The resident lint Program and the graph symbol provider are refreshed by exactly one editor notification, `textDocument/didSave`. A clean `didOpen` publishes plugin diagnostics from a possibly stale Program, and `workspace/didChangeWatchedFiles` — which the repository's own VS Code client sends for `tsconfig`/`jsconfig` changes — has no handler at all, so config edits and root-set changes never reach the sidecar. The sidecar's own full-reload fallback exists but has no transport.

Constraints the fix preserves, per the issue:

- dirty-buffer suppression is unchanged (`didChange` on a dirty buffer still sends nothing to the resident daemon; a dirty `didOpen` still publishes an empty diagnostic set and does not invalidate);
- `didSave` of a known stable source stays on the incremental changed-URI path and does not degrade into a full reload.

### #819 — use the negotiated LSP position encoding in every local conversion

The pinned typescript-go selects `utf-8` whenever the client offers it, and `ttscserver` proxies that selection to the client unchanged, while every position conversion ttsc performs counts UTF-16 code units unconditionally. Merged PR #790 added a fourth such site, so this works against current `master` rather than the issue's baseline quotes.

## Verification

Campaign CI is deliberately suspended; every run for this branch's SHAs is cancelled by the exact-SHA gate and local verification is the only evidence. Commands and their real output are recorded in the pull-request thread and in the final comment.

## Status

Draft — reserving the batch before implementation.


samchon:
## Self-Review round 1

Full base-to-head re-read of every changed file and hunk (`packages/ttsc/internal/lspserver/lsp_proxy.go`, `lsp_completion_request.go`, `packages/ttsc/internal/graphsymbols/provider.go`, `packages/lint/linthost/lsp.go`, `website/src/content/docs/development/concepts/protocol.mdx`, and the five new test files), plus the consequence surface: notification ordering against the upstream pump, the resident daemon's `sc.changed`/`sc.invalidate` batching, Windows and POSIX URI handling, and the documentation the change makes stale.

### Findings to act on

**1. The prose I added to `protocol.mdx` uses em-dashes, which the documentation skill forbids.** `.agents/skills/documentation/SKILL.md` says plainly: "No em-dashes. Use a period, comma, colon, or parentheses." The surrounding legacy paragraphs use them, but the sentences I authored are mine to get right, and one of them is a rewrite of an existing sentence, so I own it now. Remediation: rewrite my added and rewritten sentences without em-dashes; leave untouched legacy prose alone rather than widening the diff.

**2. `website/src/content/docs/lint/editor.mdx` is now incomplete for the behavior a user observes.** The development skill requires updating the matching page when code behavior changes. That page tells the reader "Plugin diagnostics and code actions use the saved project state", which stays true, but it says nothing about the two transitions this change adds: reopening a file that changed on disk now reports against the current file, and a `tsconfig.json` edit now takes effect without restarting the editor. Before this, both were stale for the life of the session. Remediation: add one short paragraph to the "Save before relying on a new lint diagnostic" area stating what refreshes and what still requires a save.

### Findings recorded, deliberately not changed

**3. A very broad client watcher can now cost full reloads it did not cost before.** A `changed` event for a path the warm Program does not hold as a source reaches the daemon's documented "unknown path means full reload" branch. Under a client that registers `**/*`, saving an unrelated file (a README) would therefore drop the warm Program, where previously nothing happened at all because the notification had no handler.

I am keeping this. The issue's invariant is that a change the host cannot localize must drop the entry rather than update it, and the proxy has no knowledge of which paths the daemon's Program holds. Narrowing by file extension would be a guess (`.json` can be a source under `resolveJsonModule`, and a newly relevant file is exactly the case a reload is for), and guessing here trades correctness for a cost that is bounded to extra reloads. The realistic watchers are narrow: this repository's own VS Code client watches `**/{tsconfig,jsconfig}*.json`, and the compiler registers globs for TS/JS/config files, all of which land on the intended paths.

**4. A dirty `didOpen` still leaves both caches untouched, including the symbol provider.** This is the issue's explicit constraint and the pre-existing behavior, and it is correct on its own terms: the dirty branch publishes an empty diagnostic set and reports nothing until the buffer reaches disk, so there is no stale answer to prevent. Recording it so the boundary is visible rather than accidental.

### Verification limitation to state honestly

Open issue #771 records that the e2e test LSP client ignores server-to-client requests, so a full editor round trip that needs the compiler to talk back cannot be driven from it. I did not work around it. What is proved instead is the proxy and daemon boundary directly: the notification arms and their invalidation payloads in `packages/ttsc/internal/lspserver`, the position conversions on both sides of the sidecar wire, and the exact bytes of the `initialize` request the proxy forwards upstream, read back through the real editor-to-upstream pump rather than from the helper alone.

Applying findings 1 and 2 now, then restarting a fresh full round.

samchon:
## Self-Review round 2

Fresh full round over the current head, not a recheck of round 1's fixes. Re-read every changed file and hunk, then re-traced the consequence surface from the notification arms outward: what each new invalidation does to already-published diagnostics, what the daemon does with a URI it does not recognize, multi-root behavior, the initialize payload's degenerate JSON shapes, and test isolation.

### Findings to act on

**1. A file deleted on disk while it is open keeps its `@ttsc/lint` squiggles.** This is a boundary the issue names and the change does not yet satisfy. A deleted file arrives as a `workspace/didChangeWatchedFiles` entry of type 3, which now correctly drops the warm Program, so the sidecar stops reporting findings for it. But nothing clears what was already published: `pluginDiagnostics[uri]` still holds the last set, and an open document's content is owned by the client rather than by disk, so the compiler has no reason to republish for that URI and trigger the merge that would replace it. The user is left looking at findings for a file that no longer exists, which is the same phantom-finding class this issue is about.

Remediation: on a deletion entry, drop ttsc's cached findings for that URI, bump its diagnostic generation so an in-flight computation over the pre-deletion Program cannot win the race, and republish the merged set. Leave upstream diagnostics alone: the compiler owns its own answer for the same document. No-op when nothing was published for that URI, which is the common case. The watched-files arm then has to propagate a write error the way the `didOpen` arm already does.

**2. `constrainInitializePositionEncoding` rewrites two offers it promises to leave alone.** Its doc comment and the new documentation both claim a client that does not offer UTF-8 reaches the compiler byte for byte. The guard actually implemented is narrower than that claim: it returns early only for a list that is exactly `["utf-16"]`, so `"positionEncodings": null` and `"positionEncodings": []` are both rewritten to `["utf-16"]`. Semantically identical (both already mean the LSP default), but it contradicts the stated contract, and a contract that is only mostly true is the thing that makes the next reader re-derive the assumption.

Remediation: make the early return the actual predicate, "the offer names no encoding other than UTF-16", which covers the empty, null, single-element, and repeated-element spellings at once.

### Findings recorded, deliberately not changed

**3. A URI outside the sidecar's own project costs that project a full reload.** If a proxy is handed a document that its `PluginSource` does not own, the URI reaches the daemon, no Program holds it as a source, and the entry reloads for nothing. This is not new and not specific to the open transition: the `didSave` path has behaved this way since residency landed. It is also unreachable through this repository's own client, which keeps one server per non-overlapping root with a root-scoped `documentSelector`. Fixing it would need the proxy to learn the source's project boundary, which is a wider change than either issue asks for, and the failure mode is wasted work rather than a wrong answer.

**4. Diagnostics published for a document are not withdrawn when it closes.** `didClose` drops the caches but sends no empty set, so the editor keeps the last published findings for a closed file. Pre-existing, orthogonal to both issues, and not touched here.

Applying findings 1 and 2 now, then restarting a fresh full round.

samchon:
## Self-Review round 3

Fresh full round over the current head. Re-read every changed file and hunk from the top, including the two round-2 additions, and re-traced the consequence surface: the deletion withdrawal's interaction with project diagnostics and with the generation counters, the error propagation a notification arm now performs, documentation vocabulary against the rest of the page, and whether each doc comment still describes the code beneath it.

### Findings to act on

**1. `constrainInitializePositionEncoding`'s doc comment no longer matches its own predicate.** Round 2 widened the early return from "the list is exactly `["utf-16"]`" to "the list names no encoding other than UTF-16", but the comment above it still enumerates the old, narrower set: "for a client that offers no encoding list ... and for a client that offers UTF-16 alone". This is precisely the drift class round 2 flagged, reintroduced one commit later by fixing the code and not the prose. Remediation: state the predicate once, in the terms the code uses, and name the four spellings it covers.

**2. The new `protocol.mdx` paragraph calls the upstream server `the wrapped tsc --lsp`, which is not the vocabulary of the page it sits on.** Every other mention on that page is "TypeScript-Go" or "tsgo", including the paragraph directly below mine. A reader hitting a third name for the same process has to work out whether it is a third thing. Remediation: use the page's existing name.

### Findings recorded, deliberately not changed

**3. Deleting a `tsconfig.json` while the editor is open can leave its project diagnostics on screen.** The deletion drops the warm Program correctly, and the withdrawal path is a no-op for that URI because project findings live in `projectDiagnostics` rather than in `pluginDiagnostics`. The next diagnostics run then fails to load a project that no longer has a config, so nothing republishes and the previous project set stays. Deliberately out of scope: it needs the project-diagnostics lifecycle to gain its own withdrawal path, it is a strictly narrower case than the source-file deletion the issue names, and inventing a partial fix for it here would widen this pull request past both issues.

**4. `withdrawPluginDiagnosticsForDeletedDocument` does not bump `projectDiagnosticGeneration`.** Correct as written. The generation it does bump is the per-document one, which is what cancels an in-flight computation for that URI. Bumping the project generation would cancel an unrelated project publication that is still valid.

**5. The race detector could not be run locally.** `go test -race` needs cgo and there is no gcc on this machine (`cgo: C compiler "gcc" not found`). The one place this change lets a test fake be reached from a spawned goroutine is the clean-open path, and that fake is now mutex-guarded with its reads going through an accessor, so the guard does not depend on the detector to be correct. Recording the gap rather than claiming coverage I do not have.

Applying findings 1 and 2 now, then restarting a fresh full round.

samchon:
## Self-Review round 4

Fresh full round over the current head. This pass read the four comment-only hunks and both documentation hunks as carefully as the code, since after three rounds they carry as much of the change's meaning as the logic does, and re-walked the six new test files case by case against what each one would still catch if the corresponding branch were deleted.

### Findings to act on

**1. The `@ttsc/lint` sidecar's new comment names a private symbol of one particular host.** `byteOffsetToLSPPosition` in `packages/lint/linthost/lsp.go` now explains its encoding by pointing at `constrainInitializePositionEncoding in packages/ttsc/internal/lspserver/lsp_proxy.go`. `@ttsc/lint` is a separate Go module and a separate package that implements the documented sidecar protocol; it is a contributor to any host that speaks that protocol, not a component of `ttscserver`. Pointing its source at an unexported function in another module makes the sidecar read as though it depends on that host's internals, and the reference rots the moment the function is renamed.

Remediation: point at the protocol contract that actually binds both sides, which is the LSP verb section of `website/src/content/docs/development/concepts/protocol.mdx`, and state the invariant in protocol terms. The same reference inside `packages/ttsc/internal/graphsymbols/provider.go` is fine and stays: that is the same module and the same binary as the proxy.

**2. The paragraph added to `lint/editor.mdx` ends by restating the rule stated directly above it.** Its last sentence, "Only an unsaved buffer is excluded, which is why a save is still what publishes your own edits", says what the preceding paragraph's "Save before relying on a new lint diagnostic or invoking a lint action" already says. The documentation skill asks for the rule stated once at its owner and forbids a wrap-up sentence that restates the paragraph. Remediation: delete the sentence. The paragraph's job is to say what refreshes without a save, and it does that before the sentence begins.

### Confirmed sound this round, recorded so a later round does not re-litigate them

- The full-invalidation cases assert `[][]string{{}}`, which matches a call recorded with no URIs, so a regression that started passing URIs on a config edit would fail the assertion rather than pass it by accident.
- The deletion withdrawal's negative twin re-seeds the plugin diagnostics before sending a plain `changed` event, so it proves the withdrawal is keyed on the deletion type rather than on any watched-file traffic at all.
- `withdrawPluginDiagnosticsForDeletedDocument` deliberately leaves `upstreamDiagnostics` intact. The compiler owns its own answer for a document whose buffer the client still holds, and dropping it would blank diagnostics the compiler still considers current.
- Every new test names a distinct file basename, which matters because the lint runner flattens `packages/lint/test/**` into one package and fails loudly on a basename collision.

Applying findings 1 and 2 now, then restarting a fresh full round.

samchon:
## Self-Review round 5

Fresh full round over the current head. This pass re-read the whole diff and then went outside it, to the pinned compiler and to the VS Code client, to test one premise the earlier rounds had taken on trust: whether the watched-file transport this change adds actually receives the events the issue's consequence table depends on.

### Evidence gathered this round

The pinned `github.com/microsoft/typescript-go v0.0.0-20260429010842-56ab4af42157` registers its own file watchers with the client at runtime. `internal/lsp/server.go` implements `project.Client.WatchFiles` by sending a `client/registerCapability` request whose `RegisterOptions.WorkspaceDidChangeWatchedFiles` carries the project's `FileSystemWatcher` globs, and `UnwatchFiles` unregisters them. A conforming client honors those dynamic registrations and sends the resulting `workspace/didChangeWatchedFiles` notifications.

Two things follow, and both matter.

**The transport is real, not hypothetical.** External changes to project source files that the editor does not have open reach the proxy through the compiler's own registrations, not only through the repository extension's static `**/{tsconfig,jsconfig}*.json` watcher. The `git pull`, branch switch, and generator rows of the issue's consequence table are therefore covered by the arm this change adds, without widening any client-side watcher.

**Widening the extension's watcher would be duplication, so it stays out.** I had this queued as a candidate: the extension's `documentSelector` covers TS and JS documents while its watcher covers only config files, which looks like a gap for a closed source file changed on disk. The compiler already asks the client to watch exactly those files, so adding a second overlapping static watcher would produce duplicate events at a real cost in a large tree, and neither issue asks for a client change. Recording the evidence rather than acting on it.

**It also explains the shape of the #771 limitation precisely.** `client/registerCapability` is a server-to-client request, which is the exact message class open issue #771 records the e2e test client as ignoring. An end-to-end test that waited for the compiler to register a watcher and then drove a real file change through it would hang on that unanswered request. That is the concrete reason this change is proved at the proxy and daemon boundary instead, and the reason I did not attempt a workaround inside the test client.

### Findings

None. The full re-read produced no correctness, boundary, concurrency, platform, documentation, or test-isolation defect that survives verification.

Two things I re-derived rather than assumed, both confirmed sound:

- **Lock order.** `withdrawPluginDiagnosticsForDeletedDocument` takes `writeMu` then `diagnosticsMu`, matching every other write-under-state path in the proxy (`writePublishDiagnosticsIfCurrent`, `writeProjectDiagnosticsIfCurrent`, `writeLocalCodeActionsResultIfCurrent`), and it is called holding neither.
- **The deletion does not need to bump `documentGeneration`.** An in-flight code action for a deleted document already resolves to an empty action list because the reloaded Program holds no findings for it, and the `executeCommand` path reads current disk and fails closed when the file is gone. Bumping it would change no outcome, so adding one would be speculative rather than corrective.

This is a complete round with nothing to apply, so Self-Review is finished.

samchon:
## Local verification

Campaign CI is suspended for this wave, so every run for every SHA on this branch was cancelled by the exact-SHA gate and local verification is the only evidence. Commands and their real output follow.

### Narrow: the branches this change adds

```
cd packages/ttsc && go test -count=1 -v -run 'TestLSPCleanDidOpenRefreshesCompilerState|TestLSPWatchedFileChangesReachCompilerCaches|TestLSPDidSaveKeepsTheIncrementalChangedURIPath|TestLSPInitializeConstrainsPositionEncodingToUTF16|TestLSPPositionConversionsSpanAstralAndCombiningRunes|TestLSPDeletedWatchedFileWithdrawsPluginDiagnostics|TestOffsetForPositionCountsUTF16' ./internal/lspserver
```

```
--- PASS: TestLSPCleanDidOpenRefreshesCompilerState (0.02s)
--- PASS: TestOffsetForPositionCountsUTF16 (0.00s)
--- PASS: TestLSPDeletedWatchedFileWithdrawsPluginDiagnostics (0.00s)
--- PASS: TestLSPDidSaveKeepsTheIncrementalChangedURIPath (0.00s)
--- PASS: TestLSPInitializeConstrainsPositionEncodingToUTF16 (0.00s)
--- PASS: TestLSPPositionConversionsSpanAstralAndCombiningRunes (0.00s)
--- PASS: TestLSPWatchedFileChangesReachCompilerCaches (0.01s)
PASS
ok  	github.com/samchon/ttsc/packages/ttsc/internal/lspserver	1.945s
```

Every test named above appears in the output, including `TestOffsetForPositionCountsUTF16`, which is the pre-existing UTF-16 twin the issue requires to keep passing. Naming them explicitly matters because `go test -run` with a pattern that matches nothing exits 0 with `[no tests to run]`.

### Broad: every ttsc Go package

```
node scripts/test-go-ttsc.cjs
```

```
ok  	github.com/samchon/ttsc/packages/ttsc/internal/lspserver	1.855s
ok  	github.com/samchon/ttsc/packages/ttsc/test/cli	356.233s
ok  	github.com/samchon/ttsc/packages/ttsc/test/ttscserver	280.707s
ok  	github.com/samchon/ttsc/packages/ttsc/test/platform	47.687s
ok  	github.com/samchon/ttsc/packages/ttsc/test/utility	30.481s
ok  	github.com/samchon/ttsc/packages/ttsc/cmd/ttsc	8.428s
ok  	github.com/samchon/ttsc/packages/ttsc/cmd/ttscserver	0.710s
```

`go vet ./internal/lspserver ./internal/graphsymbols` is clean.

### Broad: the lint sidecar

`scripts/test-go-lint.cjs` cannot run in this worktree: it resolves `typescript` and the built `ttsx` launcher from the checkout root, and a `git worktree` has no `node_modules`. I did not skip the suite. Instead I replicated exactly what that runner materializes (copy `packages/lint` into a scratch module, flatten every `packages/lint/test/**/*.go` into `linthost` through the repository's own `scripts/ci/go-test-overlay.cjs`, write a `go.work` over the in-tree shims and `packages/ttsc`, then `go test -count=1 ./linthost`) with the real `tsgo` and `ttsx` binaries from the main checkout passed through `TTSC_TSGO_BINARY` and `TTSC_TTSX_BINARY`.

The new lint-side test:

```
--- PASS: TestLSPPositionColumnsCountUTF16AcrossPlanes (0.00s)
    --- PASS: .../the_start_of_the_document (0.00s)
    --- PASS: .../ascii_counts_one_unit_per_byte (0.00s)
    --- PASS: .../bmp_cjk_is_one_unit_and_three_bytes (0.00s)
    --- PASS: .../an_astral_rune_is_two_units_and_four_bytes (0.00s)
    --- PASS: .../a_combining_mark_costs_its_own_unit (0.00s)
    --- PASS: .../a_line_feed_starts_the_next_line_at_column_zero (0.00s)
    --- PASS: .../a_carriage_return_pair_is_one_line_break (0.00s)
    --- PASS: .../an_offset_past_the_end_clamps_to_the_end (0.00s)
    --- PASS: .../a_negative_offset_clamps_to_the_start (0.00s)
    --- PASS: .../an_offset_inside_a_rune_stops_before_it (0.00s)
ok  	github.com/samchon/ttsc/packages/lint/linthost	1.553s
```

The whole suite, in two parts that together cover it:

```
go test -count=1 -timeout 40m -skip 'TestLSPFormatBufferRealBinaryE2E' ./linthost
ok  	github.com/samchon/ttsc/packages/lint/linthost	380.310s

go test -count=1 -v -run 'TestLSPFormatBufferRealBinaryE2E' ./linthost
--- PASS: TestLSPFormatBufferRealBinaryE2E (369.08s)
    --- PASS: .../dirty_buffer_formats_from_stdin,_not_disk (1.25s)
    --- PASS: .../already-formatted_buffer_is_a_no-op_null_edit (0.09s)
ok  	github.com/samchon/ttsc/packages/lint/linthost	370.088s
```

Why it is split, stated rather than hidden: a first single-process run of the whole suite failed inside `TestLSPFormatBufferRealBinaryE2E`, in that test's own `copyTree` helper, on an `os.ReadFile` of a path inside the scratch tree it builds a real sidecar binary from. That test then passed on its own at the same commit, and the rest of the suite passed with it skipped. The failure is in my replica harness's environment, not in the change: the only functional delta this pull request makes to `packages/lint` is a doc comment on `byteOffsetToLSPPosition` plus one new self-contained unit test, neither of which is reachable from the tree-copy helper that failed.

### What could not be run

- **`go test -race`.** It needs cgo and there is no gcc on this machine (`cgo: C compiler "gcc" not found`). The one fake this change lets a spawned goroutine reach is mutex-guarded with reads behind an accessor, so its correctness does not depend on the detector.
- **A full editor round trip through the e2e LSP client.** Open issue #771 records that the test client ignores server-to-client requests. That is not incidental here: the pinned compiler registers its file watchers by sending `client/registerCapability`, so a test that waited for a watcher registration and then drove a real file change through it would hang on an unanswered request. Proved at the proxy and daemon boundary instead: the notification arms and their exact invalidation payloads, the position conversions on both sides of the sidecar wire, and the exact bytes of the `initialize` request the proxy forwards, read back through the real editor-to-upstream pump rather than from the helper alone.
- **`pnpm format`.** Deliberately not run: the campaign's Post-Campaign Cleanup owns the repository-wide formatter result.

PR #867, opened 2026-07-20
fix(ttscserver): decide JSDoc completion scope by lexical scan

### LINKED ISSUE #789: fix(ttscserver): do not treat JSDoc delimiters inside strings as completion scope
## Problem

`cursorInJSDoc` decides whether completion is inside JSDoc by comparing the last raw `/**` and `*/` before the cursor. It does not distinguish comment delimiters from the same bytes inside a JavaScript or TypeScript string/template.

That creates a false-positive completion scope. For example, a cursor after `@par` in this string is treated as JSDoc because the raw prefix contains an unmatched `/**`:

```ts
const example = "/** @par";
```

The proxy can then append `jsdoc/check-tag-names` items in ordinary source text even though the owning lint rule would never inspect that location as a JSDoc block.

## Evidence

`packages/ttsc/internal/lspserver/lsp_completion.go` intentionally uses `strings.LastIndex` over raw document text. The existing tests cover real, closed, and ordinary block comments, but not delimiter-shaped text inside quoted or template literals.

The first release documents “inside a real `/** ... */` block” as the supported scope; the matcher should enforce that statement.

## Required behavior

Determine the cursor's lexical comment state without putting a TypeScript-Go round trip on the completion hot path. A cached lightweight lexical-span index keyed by live-document generation is one possible shape; the implementation should choose the smallest sound boundary and measure its consequence on large dirty buffers.

## Verification

| Input at cursor | `cursorInJSDoc` |
| --- | --- |
| `/** @par` | true |
| `const x = "/** @par"` | false |
| `` const x = `/** @par` `` | false |
| `const x = '/** @par'` with escaped quotes nearby | false |
| `/* @par` | false |
| `/** closed */ @par` | false |
| multiline JSDoc with CRLF and non-ASCII text | true |

Add a completion-level regression test as well as the lexical helper table so a later matcher refactor cannot bypass the scope decision.

Related: #733, #736, #768.

samchon:
Fixes #789.

## Intent

`cursorInJSDoc` decided the completion scope by comparing the last raw `/**` and `*/` before the cursor, so any position after JSDoc-shaped bytes counted as a doc comment. `const example = "/** @par"` was JSDoc scope, and the proxy appended `jsdoc/check-tag-names` items in ordinary source text where the owning rule never inspects.

Comment state is not a property of the bytes before the cursor, it is a property of the token stream that produced them. The decision now comes from a forward scanner-level pass, `lexicalScopeAt`, which reports the token kind a byte offset sits in: code, line comment, block comment, JSDoc, string, template, or regex literal. `cursorInJSDoc` is that pass answering `jsdoc`.

## Scope

- `packages/ttsc/internal/lspserver/lsp_completion_scope.go` (new): the scanner, plus `cursorInJSDoc` moved out of `lsp_completion.go`.
- `packages/ttsc/internal/lspserver/lsp_completion.go`: the backward-search implementation removed. `matchCompletionHints` and `completionHintApplies` are untouched, and the refuse-unknown-scope stance stays as written.
- Tests: a lexical table, a completion-level regression through `completionItemsFor`, and a hot-path benchmark.
- `website/src/content/docs/lint/editor.mdx`: the supported scope sentence now names literal text, which is what the code enforces.

`lsp_proxy.go` and `lsp_native_plugin_source.go` are deliberately untouched; nothing in this fix needed them.

## Design decisions

**Scanner, not a parse.** The proxy holds text, not a tree, and completion runs on the live dirty buffer, so a tsgo round trip per keystroke is the cost this channel exists to avoid. The forward pass has the same complexity the backward search already had.

**No span cache.** `BenchmarkCursorInJSDoc` walks a ~1 MB buffer with the cursor at its end, the worst case for a forward scan: 4.3 to 5.6 ms per pass on a loaded dev machine, so well under a millisecond for an ordinary source file and small beside the upstream completion round trip the same request makes. A cross-request cache would buy that back at the price of invalidation on every `didChange`, which is a larger correctness surface than the one being fixed.

**Recovery is part of the contract.** The buffer is usually invalid mid-edit, so an unterminated string ends at the line break exactly as the TypeScript scanner recovers. Without that, one stray apostrophe would hide every doc comment below it.

**Regex literals are skipped.** A `/` that opens a literal rather than divides is decided from the previous significant byte plus a keyword check, and the literal scan is bounded to its line, so an unterminated one falls back to division. This keeps `const re = /["]/; /** @par` reading as a real doc comment instead of a string that swallowed it. `/**` never starts a regex, so a comment opener always wins that test and a misread cannot hide a doc comment.

## Deferred

JSX element text is scanned as code, so `/**` typed between JSX tags still reads as a comment opener. Separating JSX text from a comparison operator needs the grammar, which is the parse this pass exists to avoid, and the position is far rarer than the literals fixed here. The boundary is stated in the code comment.

## Verification

Cases covered by `TestLSPCompletionScopeIgnoresJSDocBytesInLiterals`: real block, multi-line block, CRLF block with non-ASCII prose, closed block, `/**/`, plain block comment, line comment holding `/**`, no delimiter, cursor before the opener, double quoted string, single quoted string, template literal, template text after an interpolation, string inside an interpolation, nested template inside an interpolation, real comment inside an interpolation, escaped quote inside a string, escaped backslash ending a string, block after a string holding the opener, block after a string holding the terminator, quote inside a block, unterminated string recovering at the line end, regex class holding the opener, block after a regex holding a quote, regex after a keyword, division that is not a regex, cursor inside a regex literal, and out-of-range offsets.

Before the fix, the new tests fail: the completion-level case offers `param` inside the string literal, and nine lexical cases report `jsdoc` (line comment, both quoted strings, template, template after interpolation, string inside interpolation, nested template, escaped quote, regex class). After it, all pass.

Local, from `packages/ttsc`:

```
go vet ./internal/lspserver
go test ./internal/lspserver -count=1
go test ./internal/... -count=1
go build ./...
go test ./internal/lspserver -run='^$' -bench='^BenchmarkCursorInJSDoc$'
```

Go files formatted with `.vscode/gofmt-2spaces.sh -w`, the MDX file with the repository Prettier config. The full `pnpm test` suite was not run locally; CI owns it.

### DIFF
--- a/packages/ttsc/internal/lspserver/lsp_completion.go
+++ b/packages/ttsc/internal/lspserver/lsp_completion.go
@@ -59,31 +59,6 @@ func completionHintApplies(
   return strings.Contains(linePrefix, hint.After)
 }
 
-// cursorInJSDoc reports whether an offset sits inside a `/** */` block.
-//
-// A plain backward scan for the nearest `/**` or `*/`, which is enough because
-// the question is only ever asked about the cursor. It deliberately does not
-// parse: the proxy holds text, not a tree, and reaching for tsgo's parser here
-// would put a round trip on the keystroke path — the exact cost this design
-// exists to avoid.
-//
-// Line comments are not considered: `// @evidence` is not a doc comment and a
-// tag there means nothing to the rule that published the hint.
-func cursorInJSDoc(text string, offset int) bool {
-  if offset < 0 || offset > len(text) {
-    return false
-  }
-  head := text[:offset]
-  open := strings.LastIndex(head, "/**")
-  if open == -1 {
-    return false
-  }
-  close := strings.LastIndex(head, "*/")
-  // `/**` is itself followed by no `*/` yet, or the last close precedes the
-  // last open — either way the cursor is still inside the block.
-  return close < open
-}
-
 // linePrefixAt returns the text from the start of the cursor's line up to the
 // cursor.
 //
--- a/packages/ttsc/internal/lspserver/lsp_completion_refuses_hints_inside_a_string_literal_test.go
+++ b/packages/ttsc/internal/lspserver/lsp_completion_refuses_hints_inside_a_string_literal_test.go
@@ -1,0 +1,69 @@
+package lspserver
+
+import (
+  "encoding/json"
+  "testing"
+)
+
+type jsdocTagCompletionHintSource struct{ NullPluginSource }
+
+func (jsdocTagCompletionHintSource) CompletionHints() []LSPCompletionHint {
+  return []LSPCompletionHint{{
+    Scope: "jsdoc",
+    After: "@",
+    Items: []LSPCompletionItem{{Insert: "param"}},
+  }}
+}
+
+// TestLSPCompletionRefusesHintsInsideAStringLiteral verifies the scope decision
+// where it is observable.
+//
+// The lexical table proves the scanner; this proves the request path still asks
+// it. A refactor that recomputed scope from the line prefix, or that passed the
+// wrong offset, would put `jsdoc/check-tag-names` tags into ordinary source
+// text — a position the owning rule never inspects — and no unit test of the
+// matcher alone would notice.
+//
+//  1. Ask for completion after `@par` inside a string literal that contains the
+//     JSDoc opener.
+//  2. Ask again at the same tag inside a real doc comment.
+//  3. Assert silence for the first and the published item for the second.
+func TestLSPCompletionRefusesHintsInsideAStringLiteral(t *testing.T) {
+  const literalURI = "file:///project/src/literal.ts"
+  const blockURI = "file:///project/src/block.ts"
+  proxy := &Proxy{
+    source: jsdocTagCompletionHintSource{},
+    documentText: map[string]string{
+      literalURI: "const example = \"/** @par\";\n",
+      blockURI:   "/**\n * @par\n */\nexport const value = 1;\n",
+    },
+  }
+
+  literalParams, err := json.Marshal(map[string]any{
+    "textDocument": map[string]any{"uri": literalURI},
+    "position":     map[string]any{"line": 0, "character": 25},
+  })
+  if err != nil {
+    t.Fatalf("encode completion params: %v", err)
+  }
+  if pending := proxy.completionItemsFor(Envelope{Params: literalParams}); len(pending.items) != 0 {
+    t.Errorf("a string literal was offered %v, want nothing", inserts(pending.items))
+  }
+
+  blockParams, err := json.Marshal(map[string]any{
+    "textDocument": map[string]any{"uri": blockURI},
+    "position":     map[string]any{"line": 1, "character": 7},
+  })
+  if err != nil {
+    t.Fatalf("encode completion params: %v", err)
+  }
+  pending := proxy.completionItemsFor(Envelope{Params: blockParams})
+  if got := inserts(pending.items); !equalStrings(got, []string{"param"}) {
+    t.Fatalf("a real doc comment was offered %v, want [param]", got)
+  }
+  // The refusal must not have come from a broken filter range either.
+  if pending.replaceRange.Start != (LSPPosition{Line: 1, Character: 4}) ||
+    pending.replaceRange.End != (LSPPosition{Line: 1, Character: 7}) {
+    t.Errorf("replacement range = %+v, want character 4..7", pending.replaceRange)
+  }
+}
--- a/packages/ttsc/internal/lspserver/lsp_completion_scope.go
+++ b/packages/ttsc/internal/lspserver/lsp_completion_scope.go
@@ -1,0 +1,320 @@
+package lspserver
+
+// lexicalScope names the kind of token a byte offset sits in.
+//
+// It is what a scanner can decide without a parse: comments, string and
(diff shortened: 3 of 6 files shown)


samchon:
Self-Review, round 1. Complete pass over the branch diff against `master` (the scanner, both test files, the benchmark, and the guide edit). Four findings, all in `lexicalScopeAt` and its helpers; each is a false decision the current round can produce, so I am applying all four with coverage.

1. **A keyword used as a property name enables the regex path.** `regexKeywordPrecedes` reads the identifier before the `/` without looking at what precedes that identifier, so `obj.in / 2` and `obj.of / 2` are read as expression positions. `regexLiteralEnd` then consumes to the next `/` on the line, which can be the opener of an inline doc comment: `const n = obj.in / 2; /** @par` loses its JSDoc scope. `in`, `of`, `new`, `delete`, `case`, `do`, `else`, `void`, and `typeof` are all legal property names, so this is reachable from ordinary code. Fix: refuse the keyword when a `.` precedes it.

2. **A lone `\r` does not end a line comment.** The line-comment branch only accepts `\n`, so a CR-only buffer, which the proxy takes verbatim from the editor, keeps every byte after `//` in comment scope for the rest of the file. TypeScript treats CR as a line terminator. Fix: end the line comment on `\r` as well; under CRLF the following `\n` is ordinary whitespace, so nothing changes there.

3. **A completed regex leaves `last` as `/`, which reads as an expression position.** After skipping `/re/` with no flags, the next `/` is offered the regex path again. A regex literal is a value, so what follows it divides. Fix: treat `/` in `regexAllowedAfter` like the other value-producing bytes.

4. **`templateBraces` is initialized with an empty composite literal.** A nil slice appends identically and states "nothing pending" without the literal. Style only, no behavior change.

Not changed, deliberately:

- JSX element text still scans as code. Telling it from a comparison operator needs the grammar, which is the parse this path exists to avoid. Stated in the code comment and in the pull request body.
- A cursor between the `*` and the `/` of a closing `*/` reports code. Both answers are defensible for a position inside the terminator token, and no trigger in the corpus can match a prefix that ends in `*`.

samchon:
Self-Review, round 2. Fresh complete pass over the branch at `2a001f71` (scanner, both test files, benchmark, guide edit). No correctness defect survived this round; three coverage and accuracy findings, all in the test file.

1. **The CRLF line continuation in `escapeWidth` is load-bearing and untested.** `"a\` followed by CRLF continues the string, so without the three-byte escape the `\n` would trip the unterminated-literal recovery and a `/**` on the next line would read as a real doc comment. Verified by hand at `lexicalScopeAt("const s = \"a\\r\nb /** @par", 24) == string`; adding it to the table so the rule cannot be dropped silently.

2. **The table's headline overstates what every row proves.** It reads as if each case fails under the old backward search, which is true of the impostors and not of the real blocks. The real blocks are there to stop a scanner that refuses everything from passing. Rewording to say that.

3. **The regex-cursor assertion carries a bare offset 15.** Nothing states that offset is the second `*` inside the character class, which is the whole point of the case.

Considered and left as is:

- `lexicalScopeAt` clamps a high offset but not a negative one. `cursorInJSDoc` is the only caller and rejects both, and the table asserts that refusal.
- No test discriminates the interpolation brace counter. Every text I could construct ends in the same scope whether or not the counter exists, because the trailing bytes stay inside the same template either way, so a test would assert nothing about the counter.

samchon:
Self-Review, round 3. Fresh complete pass over the branch at `a23725e4`: the scanner and its four helpers, the lexical table, the completion-level regression, the benchmark, and the guide sentence. Nothing survived verification, so this is the clean round that ends the review.

Consequence surface checked this round, beyond the diff itself:

- **No sibling scope decision exists to fix.** `packages/lint` publishes hints with `rule.HintScopeJSDoc` and never decides where the cursor is; the host owns that. The other `/**` literals in the workspace are a printer emit, a graph doc-slice, and `unicorn/prevent-abbreviations` reading a real comment range from the AST, none of them a guess over raw text.
- **`driver` consumers are unaffected.** `TestLSPPluginSourceEmbedderCanPublishCompletionHints` exercises the public hint aliases with no document text, and the public surface did not change: `cursorInJSDoc` is unexported and its signature is unchanged.
- **No TypeScript e2e suite drives completion**, so the Go tests beside the package remain the only level that can pin this.
- **Platform:** the scan is pure byte logic with no path or filesystem behavior, and the CR, CRLF, and non-ASCII rows cover what a Windows editor buffer differs by.

Verification run locally from `packages/ttsc`: `go vet ./internal/lspserver`, `go test ./internal/lspserver -count=1`, `go test ./internal/... -count=1`, `go test ./test/driver/... -count=1`, `go build ./...`, and `go test ./internal/lspserver -run='^$' -bench='^BenchmarkCursorInJSDoc$'`. Go files formatted with `.vscode/gofmt-2spaces.sh -w`, the MDX with the repository Prettier config. The full `pnpm test` suite is CI's.

samchon:
Benchmark re-measured after the review commits: 4.1 ms per 1 MB pass on a quiet run, 12.7 ms on a run competing with other builds on the same machine. The body quotes the 4.3 to 5.6 ms range measured before those commits; the scan itself did not change shape, and the spread is machine load, not the code. The conclusion is unchanged: well under a millisecond for an ordinary source file, and small beside the upstream completion round trip.

PR #859, opened 2026-07-20
fix(ttsc): make the CLI flag boundary agree with the compiler it wraps

samchon:
## Intent

Campaign batch **B2 — the CLI flag boundary**. One branch, one pull request, five issues that all change `packages/ttsc/src/flags/**` and the launchers that consume it. They are sequenced under one owner because they contend on the same three files (`schema.ts`, `parser.ts`, `runBuild.ts`) and on the generated artifacts derived from the schema.

## Scope

| Issue | Title |
| --- | --- |
| #844 | Resolve CLI flag names the way the compiler the launcher wraps resolves them |
| #845 | Declare every forwarded flag with the arity its compiler actually has, and stop the forwarding path from swallowing an input file |
| #846 | Answer project-independent terminal flags before resolving a project, so `ttsc --init` can create the tsconfig it currently demands |
| #840 | Keep every documented `ttsx --require` preload, including TypeScript values and the inline forms |
| #849 | Name the configuration file and position when a config fails to parse |

## Landing order

1. **#844** first. It establishes one flag identity — normalization applied everywhere the schema's derived sets are built, including the generated Go allow-lists — and the later issues key off that identity existing.
2. **#845** next, because its fix keys off the resolved flag's declared kind.
3. **#846** next, inheriting #844's identity rule for its derived project-free set rather than inventing a second rule. The project-free set is derived from `FLAG_SCHEMA`, never hand-listed.
4. **#840** and **#849** land in the same branch; they are independent of that ordering but share the launcher surface.

## Campaign constraints honored

- `pnpm format` is **not** run on this branch. Post-Campaign Cleanup owns the repository-wide formatter result.
- Repository Actions and workflow settings are unchanged. Every push on this branch passes the exact-SHA cancellation gate; run IDs and final states are recorded in the campaign knowledge base.
- Generated artifacts are regenerated through `pnpm run gen:flags`, never hand-edited.

## Verification

Pending. Local, package-scoped verification will be recorded here as a comment when implementation lands, together with the exact commands and their results.

Closes #844
Closes #845
Closes #846
Closes #840
Closes #849

PR #868, opened 2026-07-20
fix(tests): answer the requests TypeScript-Go sends its client

### LINKED ISSUE #771: fix(test): the LSP test client ignores server→client requests, so tsgo-dependent paths hang

### LINKED ISSUE #863: test(ttscserver): nothing TypeScript-Go must answer is covered, and nothing it must answer is answered
## What was observed

A new end-to-end session in `tests/test-ttsc/src/native-plugins/server/` drives a real `ttscserver` over stdio, and no request that TypeScript-Go must answer ever comes back.

Three CI runs on `test/ttscserver-completion-e2e` (#793) agree:

- `initialize` returns.
- `textDocument/didOpen` produces `publishDiagnostics` carrying a `@ttsc/lint` `no-var` finding, so the sidecar is built and answering.
- `textDocument/hover` at a plain identifier — **times out** (60s).
- `textDocument/completion` at a plain caret the proxy forwards untouched — **times out**, five attempts.

Hover is the decisive one. It has nothing to do with the lint corpus, nothing to do with completion capabilities, and nothing the proxy enriches. The last run declared full client capabilities for completion, hover, and synchronization, and sent the invoked trigger context, and the result did not change.

So the session receives exactly what `ttscserver` produces on its own, and nothing that TypeScript-Go produces.

## Why no existing test caught it

Every LSP request the server suite sends today is one `ttscserver` answers itself:

```
$ grep -oE '"(textDocument|workspace)/[a-zA-Z]+"' tests/test-ttsc/src/native-plugins/server/*.ts | sort | uniq -c
      4 "workspace/executeCommand"
      4 "textDocument/codeAction"
```

`textDocument/publishDiagnostics` is server-initiated and also ours. The `initialize` assertions in `test_ttscserver_lsp_editor_session_merges_suppresses_and_fixes` do read TypeScript-Go's advertised capabilities out of the merged result, so the upstream process answers *that* request — but no test has ever asked it for a language feature and waited.

The proxy's whole purpose is that TypeScript-Go language features and plugin features arrive on one stream. Half of that has no coverage.

## Settled (see the comment below)

**The cause is the harness, not the product.** The section that follows is the original framing, kept as filed; its "Product" branch is ruled out. `tsc --lsp --stdio` sends `client/registerCapability` from its `initialized` handler and blocks the dispatch loop that serves every later request until the client replies. The test client dropped server→client requests, so `initialize` (answered by the read loop) returned, ttsc's own publications kept arriving, and every forwarded request queued behind a loop that never advanced. Duplicate of #771. Fixed in #868.

## What this does not yet establish

Whether the fault is in the product or in the harness. Both remain open:

- **Product.** Requests that reach the forwarding path go unanswered under some condition this session creates.
- **Harness.** The launcher is started with `TTSCSERVER_BINARY` and a resolved `TTSC_TSGO_BINARY`; if the upstream process is not spawned or not reachable in this configuration, an editor would not be affected but every forwarded request in the test would hang exactly like this.

The distinction matters and should be settled before either is assumed. A first step that costs nothing: assert in the existing session test that a hover or document symbol request is answered, which turns a hang into a named failure in a lane that already runs.

## Reproduction

#793 carries the session and reports both probes in its failure message.

samchon:
Closes #863. Closes #771, which is the same root cause, found earlier from the other end and with the same proposed invariant.

## Diagnosis: the harness, not the proxy

Both hypotheses in #863 were open. The cause is the test client.

`tsc --lsp --stdio` sends `client/registerCapability` from inside `handleInitialized`, and that handler is a synchronous notification handler running on the server's `dispatchLoop`:

```go
// typescript-go@v0.0.0-20260429010842 internal/lsp/server.go:1183
_, err = sendClientRequest(ctx, s, lsproto.ClientRegisterCapabilityInfo, &lsproto.RegistrationParams{ ... })
```

`sendClientRequest` blocks on a response channel until the client replies. `initialize` is answered earlier and directly by `readLoop`, before `initializeParams` is set, so it returns normally; every later request goes onto `requestQueue`, which only `dispatchLoop` drains. A client that never answers the registration therefore sees `initialize` succeed, keeps receiving everything `ttscserver` publishes on its own, and gets nothing back from TypeScript-Go, forever.

`TtscserverClient.dispatch` did exactly that:

```ts
if (typeof message.id !== "undefined" && message.method) {
  // Server→client request. ttscserver currently sends no such
  // requests we need to answer for these tests; ignore.
  return;
}
```

A real editor is unaffected. `vscode-languageclient` answers `client/registerCapability` itself, so `@ttsc/vscode` never hit this.

What that stalls, precisely: **one loop**. `dispatchLoop` is the single consumer of `requestQueue`, and while it is parked inside `initialized` nothing else is dispatched. `textDocument/didOpen` never reaches tsgo's document store, so `textDocument/hover` and `textDocument/completion` have nothing to answer from and are never dequeued to try. The `@ttsc/lint` `no-var` publication keeps arriving through the same stream because it never goes near tsgo: the proxy computes it from the sidecar and writes it to the editor itself. That asymmetry is the whole observation in #863, and it is a property of the client, not of which verbs the proxy enriches.

### Evidence

Two probes against the real binaries, identical except for whether the client answers server→client requests.

Direct to `tsc --lsp --stdio`:

| Answers server requests | Result |
| --- | --- |
| no | `<= SERVER REQUEST client/registerCapability id="ts1"` (ignored), then `HOVER FAILED: timeout hover` |
| yes | `=> answered client/registerCapability`, then `HOVER ANSWERED: {"contents":{"kind":"markdown","value":"var legacy: number"}, ...}` |

Through the real `ttscserver` binary: same contrast, same outcome. That run also proves the proxy side is correct end to end, since it forwards the server-initiated request to the editor and routes the editor's response back upstream untouched, which is what `TestLSPProxyRoundTripsServerInitiatedApplyEdit` already pins at the Go level.

Finally, driving the actual `TtscserverClient` from this branch against a real `ttscserver`: hover, `documentSymbol`, and completion all answered, `serverRequestMethods()` is exactly `["client/registerCapability"]`, and `shutdownTtscserverClient` still asserts a clean exit. Re-running the same script with only the `dispatch` fix reverted reproduces `timed out waiting for textDocument/hover response`.

### Correcting the issue's framing

#863 leaves open that "requests that reach the forwarding path go unanswered". They are not unanswered by the forwarding path. They never reach a dispatch loop in a servable state, because that loop is parked inside `initialized`. #793's own note, that the unenriched path is the code path every editor uses and a defect there would have surfaced long ago, was right and pointed at the client.

## Changes

- `tests/test-ttsc/src/internal/ttscserver.ts` answers server→client requests. A small table gives the spec-shaped result for the two a bare `null` would not decode: `workspace/configuration` returns one settings object per requested item, and `workspace/applyEdit` declines, since these tests never apply one. Everything else, including a request no version of TypeScript-Go has sent yet, is answered `null`, so an unknown request can never stall the server again. A responder that throws still replies, with a JSON-RPC error. `serverRequestMethods()` exposes what was answered.
- `tests/test-ttsc/src/native-plugins/server/test_ttscserver_answers_typescript_go_language_requests.ts` covers the class #863 names: a request only TypeScript-Go can answer is answered through `ttscserver`. It asserts TypeScript-Go's own `hoverProvider`, `documentSymbolProvider`, and `completionProvider` survive ttsc's rewrite of the initialize result, waits for the `@ttsc/lint` `no-var` finding so the ttsc half is known live first, asserts `client/registerCapability` was answered so a regression names the cause instead of reporting three timeouts, and then asserts TypeScript-Go's own answers: `legacy: number` in the hover, `greet` in the symbols, and `legacy` among the non-ttsc completion items.
- `website/src/content/docs/development/reference/driver-api.mdx` states the obligation in the `Hosting ttscserver` section, whose declared audience is exactly the reader who would write a hand-rolled client and lose three CI runs to this.

I chose a dedicated test file over the "assert a hover in the existing session test" first step #863 suggests. Same signal, at the granularity the repository's one-case-per-file rule asks for, without overloading a session test that already pins five other things.

## #771 is the same defect

#771, *the LSP test client ignores server→client requests, so tsgo-dependent paths hang*, filed the same root cause from the other end: it probed `textDocument/codeAction` without an `only` filter and pull `textDocument/diagnostic`, saw both time out while ttsc-owned verbs answered in about a second, and named the invariant this branch implements, that a test client which speaks LSP must answer server→client requests. #863 rediscovered it through hover and completion. Its two other observations hold as well: tsgo 7.0.2 advertises `diagnosticProvider` and never push-publishes, and the change to the shared helper does alter what all five tests in that directory do on the wire, which is why the shutdown defect below surfaced.

## A second defect this unmasks, and fixes

Answering the registration let the lane complete a real LSP shutdown for the first time, and the first CI run went red on it: the new test and `test_ttscserver_merges_project_plugin_diagnostics`, which this branch does not touch, both failed with `ttscserver: tsgo --lsp --stdio: exit status 1`, while the other four passed.

```go
// typescript-go internal/lsp/server.go
func (s *Server) handleExit(ctx context.Context, _ lsproto.NoParams) error { return io.EOF }
```

`dispatchLoop` maps that `io.EOF` to `lspExit(nil)`, after which its own `select` returns `context.Canceled`, and `Run`'s `if err := g.Wait(); err != nil && !errors.Is(err, io.EOF) && ctx.Err() != nil` returns it, so `tsc --lsp --stdio` exits 1. The alternative is that `readLoop` sees the closed stdin first and delivers `io.EOF`, and the process exits 0. It is a coin flip between the dispatch loop draining `exit` from `requestQueue` and the closed stdin reaching `readLoop`: twelve out of twelve local runs came up 0, and two of six CI runs came up 1. `defaultUpstreamRunner` turned any nonzero status into an error.

Answering the registration did not cause that. It revealed it. Before this branch no test in the repository ever completed a real LSP shutdown, because the dispatch loop was parked in `initialized` and `exit` was never processed at all, while a real editor has always answered the registration. `vscode-languageclient.stop()` sends `shutdown` then `exit`, so a VS Code user quitting today already sees `ttscserver` report a failure on roughly every other quit.

The proxy now records the editor's `exit` notification and `RunLSPServer` drops the runner error once it has arrived. After `exit` the upstream is required to terminate and its status says nothing about the session. Proxy errors still surface, and `TestLSPServerPropagatesRunnerError` remains the negative twin for an upstream failure with no exit requested. `TestLSPServerFoldsUpstreamErrorAfterEditorExit` fails before the fold and passes after.

## Consequence surface

- `TtscserverClient` is shared by all five `native-plugins/server` tests, so this changes what every one of them does on the wire. After answering, `tsc` still emits only `window/logMessage` notifications and no `textDocument/publishDiagnostics`, since it advertises `diagnosticProvider` and uses the pull channel, so no existing diagnostics assertion gains an unexpected contributor. `handleExit` returns `io.EOF`, so the now genuinely processed shutdown sequence still exits 0, confirmed by the local end-to-end run.
- The repository has exactly one hand-written LSP client: `git grep "Content-Length" -- "*.ts"` returns only this harness. `@ttsc/vscode` goes through `vscode-languageclient`.
- Product code changed only for the shutdown fold above. The Go-level contract for server-initiated requests was already locked by `packages/ttsc/test/driver/lsp_proxy_round_trips_server_initiated_apply_edit_test.go`.

## Relationship to #793

#793 is the reproduction, not a defect of its own. Its `test_ttscserver_lsp_completion_serves_the_live_jsdoc_corpus` uses this same harness, so it should be able to run for real once rebased onto master with this merged. That branch is untouched here.

## Verification

- `npx tsc --noEmit -p tests/test-ttsc/tsconfig.json`, clean.
- `npx prettier --check` on the three changed files, clean.
- Local end-to-end script driving this branch's `TtscserverClient` against a real `ttscserver` and `tsc --lsp --stdio`: hover, documentSymbol, completion, `serverRequestMethods()`, and a clean shutdown all pass. Reverting only the `dispatch` fix reproduces the hover timeout.
- `go build ./...`, `go vet ./internal/lspserver ./driver`, `go test ./test/driver ./internal/lspserver -count=1`, all clean. `TestLSPServerFoldsUpstreamErrorAfterEditorExit` fails with `exit status 1` when the fold is disabled and passes with it.
- `go test ./test/ttscserver -run TestTtscserverCommandHandlesStdioShutdown` passes locally in 191s. The whole `test/ttscserver` package exceeds the 600s default timeout on this Windows machine because every case shells out through `go run`; the `go` lane owns it.
- The `ttsc native server` lane is the real gate for the new e2e test and is left to CI.

https://claude.ai/code/session_01NeJzSHWjZZU1V1EoQhvbz9


samchon:
## Self-Review, round 1

Full round over the base-to-head diff plus the first CI result. Four findings, all being acted on in the next push.

### 1. The `ttsc native server` lane is red, and it is a real defect this branch unmasks

Two tests failed: the new one, and `test_ttscserver_merges_project_plugin_diagnostics`, which I did not touch. Both failed the same way, in `shutdownTtscserverClient`:

```
stderr=context canceled
ttscserver: tsgo --lsp --stdio: exit status 1
1 !== 0
```

The other four in the lane passed, so this is a race, not a deterministic break.

Cause, read out of `typescript-go@v0.0.0-20260429010842 internal/lsp/server.go`:

```go
func (s *Server) handleExit(ctx context.Context, _ lsproto.NoParams) error { return io.EOF }
```

`dispatchLoop`'s `handleError` maps that `io.EOF` to `lspExit(nil)`, after which the loop's own `select` returns `context.Canceled`. `Run` then reaches

```go
if err := g.Wait(); err != nil && !errors.Is(err, io.EOF) && ctx.Err() != nil { return err }
```

with `context.Canceled`, and `tsc --lsp --stdio` exits 1. The alternative is that `readLoop` sees stdin EOF first, delivers `io.EOF` to the errgroup, and the process exits 0. Which of the two wins is a coin flip between the dispatch loop draining `exit` from `requestQueue` and the closed stdin reaching `readLoop`. Locally it comes up 0 twelve times out of twelve; on the CI runners it came up 1 twice out of six.

`defaultUpstreamRunner` turns any nonzero upstream status into an error, so `ttscserver` then exits nonzero too.

This is not caused by answering `client/registerCapability`; it is **revealed** by it. Before this branch no test in the repository ever completed a real LSP shutdown, because the dispatch loop was parked in `initialized` and `exit` was never processed at all. A real editor has always answered the registration, so a VS Code user quitting today already hits this: `vscode-languageclient.stop()` sends `shutdown` then `exit`, and `ttscserver` reports a failure roughly half the time on a normal quit.

Remediation, in product code: after the editor sends `exit`, the upstream process is expected to terminate and its exit status is no longer ttscserver's to report. The proxy records that the editor sent `exit`, and `RunLSPServer` drops the runner error in that case. Proxy errors are unaffected, and `TestLSPServerPropagatesRunnerError` keeps pinning the no-exit case as the negative twin.

### 2. The `client/registerCapability` assertion races the diagnostic it waits behind

Step 3 asserts `serverRequestMethods()` right after the `@ttsc/lint` `no-var` publication. That publication comes from ttsc's own sidecar, not from TypeScript-Go, so nothing orders it after TypeScript-Go's registration. It is safe in practice, since the registration follows `initialized` within milliseconds while the sidecar has a Program to load, but "in practice" is how a flake gets written.

Remediation: assert it after the hover instead, where the ordering is guaranteed by construction (a hover cannot be answered unless the dispatch loop advanced past `initialized`), and put `serverRequestMethods()` into the hover failure message so a regression still names the cause instead of reporting a bare timeout.

### 3. Answering a request can now write to an ended stdin

`dispatch` may now call `send` reactively at any time, including after `shutdownTtscserverClient` has called `endStdin()`. TypeScript-Go does send late `client/registerCapability` requests for file watchers, so the window is real. `writable.write()` after `end()` does not throw; it emits `error` on the stream, and with no listener that is an uncaught exception that would take down the whole suite process rather than fail one case.

Remediation: skip the answer when stdin is already ended or destroyed (there is nobody left to answer), and attach an `error` listener to the child's stdin so a late EPIPE cannot crash the run. A dead child is already reported through the `close` handler, which rejects every pending request.

### 4. Unbounded `serverRequests` growth

Cosmetic. Sessions see one or a handful of these, and the array is the same shape as the existing `pending` map. Recording the method names is what makes finding 2's assertion possible, so it stays; no change.

### Not changed

- Timeout constants: `DIAGNOSTICS_TIMEOUT` of 120s matches `test_ttscserver_lsp_editor_session_merges_suppresses_and_fixes`, which has the same structure and was the first test in the lane until this file sorted ahead of it. The cold `@ttsc/lint` Go build is charged to the deliberately unbounded `initialize`, because the launcher builds project plugins before it spawns the server. The CI run confirms the shape: the new test took ~125s including that build, the five after it took under a second each.
- The proxy's own handling of server-initiated requests. `TestLSPProxyRoundTripsServerInitiatedApplyEdit` already pins it, and the local probe through the real `ttscserver` binary confirmed it end to end with TypeScript-Go's string ids.

samchon:
## Self-Review, round 2

Fresh full round over the updated diff, plus the rest of the first CI run's evidence. Three findings, two applied.

### 5. All three red lanes in run 1 had the one cause, and it is the shutdown defect

I had only read the `ttsc native server` log when I wrote finding 1. The other two failures are the same thing, which widens the blast radius of the defect and confirms the fix covers it:

| Lane | Case | stderr |
| --- | --- | --- |
| `ttsc fast` | `test_ttscserver_launcher_resolves_platform_binary_without_env_override` | `context canceled` / `ttscserver: tsgo --lsp --stdio: exit status 1` |
| `ttsc native corpus-misc` | `test_project_rule_lifecycle_surfaces_through_package_discovery_cli_api_and_watch` | same |
| `ttsc native server` | the new case and `test_ttscserver_merges_project_plugin_diagnostics` | same |

Four cases across three lanes and three directories, none of them touched by this branch, all failing on the exit status of a normal quit. Nothing else failed. No change needed beyond the fold already applied; recording it because it is the honest measure of what answering the registration exposed.

### 6. The editor-exit fold test could have passed vacuously

`failAfterExit` returned the raw read error when its stream ended before an `exit` ever arrived, and `RunLSPServer` folds `ErrFrameClosed` unconditionally. A run in which the notification never reached the runner would therefore have returned nil and passed for the wrong reason.

Applied: return a distinct error in that branch, one `RunLSPServer` reports. Re-verified both directions afterwards, `exit status 1` with the fold disabled and clean with it enabled.

### 7. `RunLSPServer`'s own doc comment did not mention the fold

The comment enumerates what counts as a clean shutdown (`ErrFrameClosed`, context cancellation) and an embedder reads it before reading the body. A runner error after `exit` now joins that set and belongs in the same sentence.

Applied. The website's `Hosting ttscserver` section states it for the embedder audience too.

### Re-checked, no change

- Fold breadth. `serverErr = nil` after `exit` also swallows an upstream crash that happens to coincide with the editor quitting. Reaching that window requires the crash and the `exit` to land together, since the runner's own `defer cancel()` tears the session down and `wg.Wait()` returns before a later `exit` could arrive. Narrowing it to `*exec.ExitError` would buy little and would force the test to spawn a real process to reach the branch.
- `case methodExit` is guarded on `IsNotification()` and falls through to the forward, so `exit` still reaches tsgo unchanged. A hypothetical request named `exit` does not set the flag.
- Answering the watcher registrations leaves tsgo believing watchers exist while the client never sends `workspace/didChangeWatchedFiles`. tsgo logs it and continues; the local runs show `window/logMessage` and nothing else.
- The e2e case's inline steps renumbered cleanly after the reorder (handshake, didOpen, hover, registration, symbols, completion) and the doc comment's four-step summary matches.

samchon:
## Self-Review, round 3

Fresh full round, reading every changed file end to end in its current state. Two findings, both applied.

### 8. The responder table was an object literal keyed by wire input

`SERVER_REQUEST_RESPONDERS[message.method]` resolves through the prototype chain, so a server request named `constructor`, `toString`, or `valueOf` would have found an inherited function, called it with the params, and answered with whatever it returned. No language server sends those, but the key comes straight off the wire and the table exists precisely to make unknown methods safe.

Applied: a `Map`, which is also what the rest of the class uses for wire-keyed state.

### 9. `serverRequestMethods()` documented what it does not guarantee

It said "requests answered". Finding 3's guard means a request arriving after the shutdown sequence closed stdin is recorded but not answered. Keeping the record is the right call, since arrival is what the test is asking about, so the documentation moves to match: "received", with the post-shutdown case named. The new case's failure message follows.

### Re-checked, no change

- `answerServerRequest`'s catch calls `send` a second time. Reached only when the responder throws or its result fails to serialize; the error payload it sends is a plain object that cannot fail the same way.
- The classification in `dispatch` is unchanged and correct: JSON-RPC responses never carry `method`, so the request branch cannot swallow one.
- `mergeCompletionResponseWithRequest` leaves upstream items untouched when a plugin contributes none, which is what the new case's completion assertion relies on. Confirmed against the real binary: 1045 upstream items, `legacy` among them.
- Every Windows/POSIX-sensitive path in the new case goes through `pathToFileURL` and `TestLint.createProject`.

### Verification after these edits

`npx tsc --noEmit -p tests/test-ttsc/tsconfig.json`, `npx prettier --check`, `go build ./...`, `go vet`, `go test ./test/driver ./internal/lspserver -count=1`, and three more end-to-end runs of the real `TtscserverClient` against a real `ttscserver` built from this branch, all clean.

samchon:
## Self-Review, round 4: rebase onto current master

Rebased onto `1a4ebde16` after #866, #859, and #867 landed. One file conflicted, `packages/ttsc/internal/lspserver/lsp_proxy.go`, in two places, both textual adjacency rather than disagreement. Read both against their neighbours before resolving.

**The method constant block.** #866 added `methodDidChangeWatchedFiles`; this branch added `methodExit`. Both kept. The dispatch arms sit in the same `switch` and neither returns, so `didChangeWatchedFiles` still refreshes and forwards, and `exit` still records and forwards.

**A new function landing where `editorRequestedExit` was added.** #866's `constrainInitializePositionEncoding` and `offersEncodingOtherThanUTF16` occupy the same spot. Both kept, in that order.

### Neither neighbour changed an assumption the new case relies on

- **#866, position encoding.** `constrainInitializePositionEncoding` rewrites `general.positionEncodings` to `["utf-16"]` before the initialize reaches tsgo. The new case declares no `general.positionEncodings` at all, so the rewrite is a documented no-op for it and the session runs on the LSP default, which is the same UTF-16. Every position it sends is on ASCII text (`character: 5` inside `legacy`, `character: 2` at a body indent), where all three encodings agree, so the assertions do not depend on the negotiation either way.
- **#866, watched files and residency.** The new case sends no `workspace/didChangeWatchedFiles` and deletes no file, so `invalidateForWatchedFileChanges` and the withdrawal path are not on its route. The interaction that does exist is in the other direction and is benign: answering `client/registerCapability` means tsgo now believes its watchers are registered while the test client sends no watched-file notifications. tsgo logs that and continues, which is what the local runs show.
- **#867, the lexical completion scope.** `lexicalScopeAt` decides whether a caret is in JSDoc. The new case's completion caret is at the start of a function body, plain code in both the old and new implementation, and its project enables only `no-var`, which publishes no hints. Its assertion is on tsgo's own items after the merge, which `handleCompletionRequest` leaves untouched when the plugin contributes none.

### Verification after the rebase

- `go build ./...`, `go vet ./internal/lspserver`, `go test ./test/driver ./internal/lspserver -count=1`, all clean.
- `npx tsc --noEmit -p tests/test-ttsc/tsconfig.json`, clean.
- Three more end-to-end runs of the real `TtscserverClient` against a `ttscserver` rebuilt from the rebased tree: hover, documentSymbol, completion, `serverRequestMethods()`, and a clean shutdown, all passing.
- `prettier --check` reports all three touched files locally, and reports an untouched sibling identically: this checkout has `core.autocrlf=true` and the index holds LF. The same content passed the `typecheck` lane on the pre-rebase run.

PR #869, opened 2026-07-20
docs(agents): forbid idling on a long command during a campaign wave

samchon:
## Intent

The campaign development document tells an implementing agent what to do and in which order, but never that a long command is something to run *alongside* work rather than watch. The waves of the current campaign paid for that omission: an agent would claim a batch, wait for a dependency install, then wait again for each verification, with four batches idling in parallel.

This adds the missing rule, stated where the waiting happens.

## Scope

- `.agents/skills/issue-campaign/development.md` gains a `Never Idle On A Long Command` section, listed in the document's Flow.
- `.agents/skills/issue-campaign/SKILL.md`'s pointer sentence gains the same concern, so the index still describes what the linked document owns.

The section says to start long commands in the background and keep working, and names the four overlaps that are always available: claim before setup finishes, read and implement against the source tree, fill a verification wait with work that does not consume its result, and start co-dependent checks together.

## What it deliberately does not allow

Two overlaps are forbidden, because overlapping them destroys the evidence they exist to produce:

- **A Self-Review round must not race a change.** The review skill already requires one complete round over a final surface and a restart whenever anything changes, so a round begun while code can still move is not a round. What may move earlier is its preparation — the diff inventory and consequence-surface notes.
- **A merge must not precede its evidence.** Under a CI-suspended campaign, local verification is the only gate, so a result arriving after the merge proves nothing about what was merged.

Without those two exclusions the rule would quietly authorize reviewing a moving target and merging on a pending check, which is the opposite of what the campaign's gates exist for.

## Verification

Documentation only; no code, no generated artifact, no workflow touched.

- `git diff --stat origin/master...HEAD` — two files, both under `.agents/skills/issue-campaign/`.
- Anchor check: the new Flow entry `#never-idle-on-a-long-command` matches the section heading.
- Read back in rendered form for section order and list structure.
- `pnpm format` deliberately not run: the campaign's no-format rule holds until Post-Campaign Cleanup owns the repository-wide formatter result.
- Campaign CI cancellation gate run for the pushed SHA `26c119a5`: no workflow run was enqueued for it, recorded in `ci-state.md`.

### DIFF
--- a/.agents/skills/issue-campaign/SKILL.md
+++ b/.agents/skills/issue-campaign/SKILL.md
@@ -59,6 +59,6 @@ Use tables for repeated case mappings. Read the rendered issue back and keep its
 
 ## Develop And Repeat The Campaign
 
-Read [development.md](development.md) in full when the user authorizes implementation pull requests or ends a campaign that entered implementation. It owns per-push campaign CI cancellation, DAG-based batching, claim pull requests, implementation waves, worktree cleanup, renewed discovery, the no-format rule, and Post-Campaign Cleanup. Do not continue after a push until its cancellation gate passes.
+Read [development.md](development.md) in full when the user authorizes implementation pull requests or ends a campaign that entered implementation. It owns per-push campaign CI cancellation, DAG-based batching, claim pull requests, implementation waves, the concurrency discipline that keeps an agent from idling on a long command, worktree cleanup, renewed discovery, the no-format rule, and Post-Campaign Cleanup. Do not continue after a push until its cancellation gate passes.
 
 An audit or issue-publication-only campaign does not load this campaign's [implementation procedure](development.md) or mutate repository Actions.
--- a/.agents/skills/issue-campaign/development.md
+++ b/.agents/skills/issue-campaign/development.md
@@ -69,6 +69,7 @@ Two things must never be overlapped, because overlapping them destroys the evide
 
 - **A Self-Review round must not race a change.** The review skill requires one complete round over a final surface and a restart whenever anything changes, so a round begun while the code can still move is not a round at all. Prepare its inputs early — the base-to-head diff inventory, consequence-surface notes, the list of generated artifacts to re-check — and open the round once the batch stops changing.
 - **A merge must not precede its evidence.** Local verification is the only gate a CI-suspended campaign has, so a result that lands after the merge proves nothing about what was merged.
+- **The cancellation gate is a barrier, not a slow command.** [Cancel Campaign CI After Every Push](#cancel-campaign-ci-after-every-push) already forbids continuing while a gate is unresolved; nothing in this section relaxes that. Poll it to completion before doing anything else.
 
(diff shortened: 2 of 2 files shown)


samchon:
## Self-Review round 1 — one finding, with its remediation

**The new section contradicts an existing rule in the same document.**

`Implement And Revalidate A Batch` already says: *"Immediately pass the exact-SHA cancellation gate after each push; do not hold a completed implementation locally until handoff **or continue working while that gate is unresolved**."* The parent skill repeats the barrier: *"Do not continue after a push until its cancellation gate passes."*

My section says to overlap by default and lists two exclusions, neither of which is the gate. A reader who takes the new rule at face value would fill the gate's polling wait with the next unit of work — which is precisely what the campaign forbids, and for a good reason: the gate exists so a campaign push cannot leave someone else's CI minutes burning while the agent moves on, and an unresolved gate is the one condition under which the campaign wants everything to stop.

That is a real conflict, not a wording preference. Two rules in one document pointing opposite ways is worse than the omission I set out to fix, because the reader cannot tell which governs.

**Remediation:** add the cancellation gate as a third never-overlap item, phrased so it is clearly a barrier rather than a slow command, and cross-reference the existing sentence instead of restating its rationale. No other change; the four positive overlaps stand as written, since none of them touches a gate.

Round 2 will re-read the whole surface after the edit, per the review law.

samchon:
## Self-Review round 2 — one finding, introduced by round 1's own remediation

**The list lead-in now miscounts its own list.** It reads *"Two things must never be overlapped"* above three bullets, because round 1 added the cancellation-gate barrier without touching the sentence that introduces them.

Small, but not cosmetic in a document whose whole job is to be followed literally: a reader who counts two and stops has a fifty-fifty chance of skipping the barrier that round 1 existed to add.

**Remediation:** drop the count and let the list carry its own length. Replacing "Two things" with "These must never be overlapped" keeps the rule and removes a number that has to be maintained every time the list changes — the same failure mode this campaign has been filing issues about all week, in miniature.

Round 3 will re-read the whole surface after the edit.

samchon:
## Self-Review round 3 — clean

Full re-read of the final surface (both files, `origin/master...HEAD`): 21 insertions across two files, no code, no generated artifact, no workflow.

Checked and clean:

- The Flow entry and the section heading agree, so the anchor resolves, as does the new in-section link to `Cancel Campaign CI After Every Push`.
- The never-overlap list no longer states a count it would have to maintain.
- No rule here contradicts the campaign document's existing gate barrier, the review skill's requirement that a round cover a final surface, or the pull-request skill's merge gate — the three exclusions exist precisely to preserve them.
- Prose paragraphs are one source line each, headings carry no numbering, and each rule states itself before its reason, per the repository's writing conventions.
- `pnpm format` deliberately not run; the campaign's no-format rule holds until Post-Campaign Cleanup.

Cancellation gate passed for all three pushed SHAs — `26c119a5`, `b2fadff0`, `5e7f77e8` — and for the pull-request creation. No workflow run was ever enqueued for any of them, which is expected for a documentation-only change under this repository's path filters; recorded in `ci-state.md`.

Two rounds found something and were remediated; this round found nothing. Merging.

PR #870, opened 2026-07-20
feat(ttscserver): rediscover completion corpora without a language-server restart

### LINKED ISSUE #788: feat(ttscserver): refresh completion-hint corpora and late trigger characters without restart
## Problem

`NativePluginSource` discovers completion hints exactly once, in a background goroutine started by `NewNativePluginSource`. The resulting slice is a session snapshot:

- `didOpen` / `didSave` invalidate resident plugin Programs, but do not rediscover `lsp-hints`.
- enabling a hint-producing rule or changing a contributor's indexed inputs requires a full language-server restart;
- `initialize` can finish before the asynchronous discovery, so trigger characters that were not already advertised by TypeScript-Go never reach the client.

The current VS Code guide documents restart as the safe workaround. That is honest degradation for the first release, but it leaves project-derived corpora stale precisely when their source documents change.

## Evidence

- `packages/ttsc/internal/lspserver/lsp_native_plugin_source.go`: `go source.discoverCompletionHints()` runs only during construction; `completionHints` is later read as an immutable snapshot.
- `packages/ttsc/internal/lspserver/lsp_proxy.go`: `invalidateResidentPlugins` runs for saved/opened documents but does not touch the hint snapshot.
- completion trigger characters are merged only while rewriting the upstream `initialize` response. A corpus that arrives later cannot update that response.

This is the remaining lifecycle edge after the corpus-snapshot design in #736. It is independent of per-keystroke matching: completion requests must continue to read only an in-memory snapshot.

## Required behavior

1. A saved input or relevant config change schedules one asynchronous corpus refresh rather than blocking the editor request.
2. Concurrent invalidations coalesce, and an older refresh may not overwrite a newer generation.
3. The last known-good corpus remains available while a refresh runs; a failed plugin refresh is logged and does not erase an unrelated producer's good corpus.
4. Trigger-character changes discovered after initialization are advertised through an LSP-correct capability update, or through another proven strategy that does not restart the server or replace TypeScript-Go's completion provider.
5. Sources without `CompletionHints` and plugins without `lsp-hints` remain unchanged.

## Verification

| Boundary | Assertion |
| --- | --- |
| Refresh | Change a contributor-indexed file, send `didSave`, and observe the new item without restarting `ttscserver`. |
| Config | Enable `jsdoc/check-tag-names` after startup and observe its corpus after the configured refresh event. |
| Race | Complete generation N+1 before N and assert N cannot restore stale items. |
| Failure | Make one producer's refresh fail and assert other producers and the prior good snapshot remain usable. |
| Capability | Discover a trigger not already advertised by TypeScript-Go and assert the client begins invoking completion for it. |
| Regression | TypeScript-Go completion items and trigger characters remain present and in their original order. |

The capability test depends on fixing the shared test client's server→client request handling in #771; `client/registerCapability` cannot currently complete in that harness.

Related: #733, #736, #771.

samchon:
## Summary

Closes #788. The completion-hint corpus was a session snapshot: `NewNativePluginSource` fetched it once in a background goroutine and every later read was that same slice, so a rule enabled after startup — or a contributor index rebuilt from a saved document — could not reach the editor without **ttsc: Restart language server**.

The corpus now rediscovers itself during the session. The async-by-design stance from #736 and #790 is preserved: no editor request ever waits for a Program load, and "no hints yet" is still the honest answer before the first fetch lands.

## What changed

**Source (`lsp_native_plugin_source.go`, `lsp_refresh.go`)**

- Each producer's corpus is stored separately, stamped with the refresh generation that produced it. The flattened publication-order snapshot the completion path reads is rebuilt on every store, so completion still copies a ready slice.
- `RefreshCompletionHints()` schedules one asynchronous rediscovery through `coalescingRefresh`: at most one run in flight and one queued rerun, each run carrying a strictly increasing generation. The startup fetch goes through the same scheduler, so startup and mid-session refresh are not two independent writers.
- A store from an older generation is dropped. A producer whose refresh could not run keeps its last known-good corpus; only its own successful answer changes it, and an empty successful answer clears it — that is how a disabled rule's items stop being offered.
- `SetCompletionHintsObserver` reports a completed cycle to the host.

**Proxy (`lsp_proxy.go`)**

- `textDocument/didSave`, `workspace/didChangeConfiguration`, and `workspace/didChangeWatchedFiles` schedule a refresh. `textDocument/didChange` deliberately does not: the buffer is dirty but disk is unchanged, and a refresh spawns a sidecar per plugin while completion is on the live-buffer hot path.
- The trigger characters the editor was actually told about are recorded while `initialize` is augmented, including the session where the corpus was still empty.
- Both hooks are optional-interface assertions, so `NullPluginSource` and any source built before this are unaffected.

**Driver / docs**

- `driver.CompletionHintRefresher` and `driver.CompletionHintObserverSource` document the new optional seams.
- `lint/editor.mdx`, `faq.mdx`, `development/concepts/protocol.mdx`, `development/walkthroughs/lint.mdx`, `development/reference/driver-api.mdx`, and `packages/vscode/README.md` replace "restart to force a fresh discovery" with the save-driven refresh and state precisely what still needs a restart.

## What still requires a restart, and why

**One case: a completion trigger character that neither TypeScript-Go advertises nor the corpus contained when `initialize` was answered.**

Trigger characters are merged into the `initialize` response, which the editor has already consumed. LSP's only remedy is `client/registerCapability`, and it is not used here for two independent reasons:

1. **It would corrupt TypeScript-Go's own list.** `vscode-languageclient` implements a dynamic completion registration as a *second* `registerCompletionItemProvider` beside the static one — trigger characters only control auto-opening, while VS Code queries every matching provider on each completion request. The client would send two `textDocument/completion` requests and merge both answers, so every tsgo item would be offered twice. The static registration's id is client-generated, so it cannot be unregistered first. Issue requirement 4 explicitly forbids replacing TypeScript-Go's completion provider; offering its whole list twice is the same harm by another route.
2. **#771 blocks verification.** The shared LSP test client ignores server→client requests, so `client/registerCapability` cannot complete in that harness. Even a correct implementation could not be exercised end to end today.

Rather than leave that gap silent, the proxy sends one `window/logMessage` per character naming it and saying what to do. Everything else about such a hint already works without a restart: the items appear on explicit completion (Ctrl+Space) and after any character the session already advertises — which includes `@`, `.`, `/`, `"`, `'`, backtick, `<`, and `#` from tsgo.

## Verification

- `go test ./internal/lspserver -count=1` and `-count=5 -run 'Hints|LateCompletion'` (repeat run for flakiness) — pass.
- `go test ./internal/... ./driver ./test/... -count=1` — pass, including `test/ttscserver` after the rebase onto master.
- `go vet ./internal/lspserver ./driver` — clean.
- Go files formatted with `.vscode/gofmt-2spaces.sh -w`.

**Could not run locally, deferred to CI:** `-race` (no gcc on this Windows host; `CGO_ENABLED=1` fails, and no CI lane passes `-race` either), `node scripts/test-go-lint.cjs` and `pnpm test:typecheck` (the worktree has no `node_modules` and no built `ttsx` launcher; neither the lint Go engine nor any TypeScript source changed), and `pnpm format` for the same reason — the Markdown/MDX edits follow the existing one-paragraph-per-line style and no CI lane checks repository formatting.

## Tests

Seven new cases, each failing before this change:

| Case | Boundary |
| --- | --- |
| `lsp_hints_refresh_replaces_the_session_snapshot_test.go` | No corpus yet, a corpus arriving later, one changing, and one that stops publishing |
| `lsp_hints_refresh_keeps_each_producer_isolated_test.go` | A whole generation where no producer could answer leaves both corpora intact and logs nothing; one producer's success does not disturb another. On master `discoverCompletionHints` assigned a freshly built slice unconditionally, so the same body would observe an emptied corpus |
| `lsp_hints_refresh_ignores_a_superseded_generation_test.go` | Generation N+1 completing before N; N cannot restore stale items, and N+2 still applies |
| `lsp_hints_refresh_coalesces_concurrent_requests_test.go` | Five requests during one run produce exactly one rerun with a newer generation; a request while idle starts a fresh run |
| `lsp_hints_refresh_serves_completion_while_running_test.go` | Concurrent completion reads against concurrent refresh writes; every snapshot is a whole corpus in manifest order |
| `lsp_hints_refresh_scheduled_by_editor_events_test.go` | Which notifications refresh, and the negative twin: `didChange`, `didOpen`, `didClose`, and `completion` must not |
| `lsp_late_completion_trigger_is_reported_once_test.go` | One notice per late character, none before `initialize`, none on repeat, none for a character upstream already advertises |

## Change integrity

- `packages/ttsc/driver/lsp.go` gains two documented optional interfaces; no existing symbol changed.
- `discoverCompletionHints` now takes a generation, so the one existing caller in `lsp_hints_absence_is_not_an_error_test.go` passes `1`. Its assertions are unchanged.
- A stray `CodeActionKinds` doc line that had been left attached to the `CompletionHints` comment block was moved back onto `CodeActionKinds`.

## Follow-ups, not in this PR

- `lsp-hints` still spawns a sidecar per refresh rather than reusing the resident daemon's warm Program. Routing it through `lsp-serve` needs a new verb in `packages/lint/linthost/serve.go`, and today `serveRun` treats an unknown-verb exit code as an answered failure rather than falling back to exec, so routing it before that fallback exists would break hints against any lint binary that predates the verb.
- The trigger-character half stays blocked as described above; #771 remains its harness prerequisite.

https://claude.ai/code/session_01NeJzSHWjZZU1V1EoQhvbz9


samchon:
## Self-Review, round 1

One complete pass over the branch diff (Go source, tests, driver surface, and every documentation page that describes the corpus lifecycle), plus the consumers the change reaches: `packages/lint/linthost` (the only shipped `lsp-hints` producer), `packages/vscode` (the only shipped client), and the `test/ttscserver` Go suite.

### Findings to apply

**1. `packages/lint/linthost/hints.go` still tells the caller to ask once.** `RunLSPHints`'s doc comment says the caller "is expected to ask once and cache rather than ask per request". That was the whole contract before this branch; the host now asks once *per invalidating event* and caches between them. The sentence is the producer-side statement of the behavior this PR changes, so leaving it would leave the next reader with the old lifecycle.

**2. `lint/editor.mdx` is vaguer than the shipped client.** "change the editor configuration, or change a watched file" is accurate about the server, but this page is the VS Code guide, and the extension synchronizes exactly two things: the `ttsc` settings section and a `**/{tsconfig,jsconfig}*.json` watcher (`packages/vscode/src/extension.ts`). Naming them removes the reading that saving `lint.config.ts` takes the watched-file path — it takes the save path, which also refreshes.

### Verified consequences accepted as designed

**3. A project that does not parse clears the lint corpus until the next good refresh.** `RunLSPHints` returns an empty array with exit 0 when `parseDiags` is non-empty, and the store treats an empty *successful* answer as "this producer publishes nothing now" — which is what makes a disabled rule's items stop being offered. The wire cannot distinguish "the Program did not parse" from "no rule publishes", so the two share an answer. Keeping the clear is the better trade: the alternative reinstates exactly the staleness this issue is about, the state is self-healing on the next successful save, and a corpus is a projection of a Program that currently does not exist.

**4. `lsp-hints` still loads a cold Program per refresh.** Coalescing bounds it to one run in flight plus one queued rerun, and it never blocks an editor request, but the steady-state cost of repeated saves on a large project is real. The fix is to route the verb through the resident daemon that already holds a warm Program for `lsp-diagnostics`. That needs a new verb in `packages/lint/linthost/serve.go` **and** a change to `serveRun`, which today treats a nonzero resident exit as an answered failure rather than falling back to exec — routing before that fallback exists would break hints against every lint binary that predates the verb. Out of scope here, recorded as a follow-up.

**5. Considered and rejected: an `lsp-hints`-unsupported cache.** `serveRun` remembers a sidecar that cannot speak `lsp-serve` so it stops retrying, and the same shape would spare a legacy plugin one failed spawn per save. Rejected: `run` returns the same error for "unknown verb" and for a transient failure (a locked binary, a timeout), so the cache would convert one unlucky first attempt into a session-long silence — and it would remove the retry that makes refresh self-healing, which is the property this PR exists to add. The saving is also small next to the one producer that does implement the verb.

### Checked and clean

- Lock discipline: the observer is invoked outside `hintsMu`, and `takeUnadvertisedCompletionTriggers` reads the corpus before taking `capabilityMu`, so no source lock is held across a proxy lock or a frame write.
- `rememberAdvertisedCompletionTriggers` runs after every early return in `augmentInitializeResult` that means "there is no session", and before the merge rewrites `completionProvider`, so the recorded set is exactly what the editor receives — including the session where the corpus was still empty, which is the one that matters.
- Ordering is manifest order on every rebuild, not map order; a manifest entry repeated under two stages keeps two corpora because the key mirrors the identity `pluginOwnsCommand` compares.
- Nothing observes a half-rebuilt corpus: the flattened snapshot is replaced under the write lock in one assignment.
- No file owned by another in-flight branch was touched: `tests/test-ttsc/src/internal/ttscserver.ts`, `lsp_completion.go`, and `lsp_completion_scope*.go` are untouched.

samchon:
Rebased onto `bc5fd26` (master had advanced with the completion-scope work) and re-ran `go vet ./internal/lspserver ./driver` plus `go test ./internal/lspserver -count=1` on the new base — clean. The rebase touched no file this branch owns; the scope work lives in `lsp_completion.go` and `lsp_completion_scope*.go`, which this branch does not modify.

samchon:
## Rebased onto `1a4ebde` (#866, #859, #867) — two real conflicts, both semantic

The textual conflicts were small; the important one was invisible to git.

**1. `methodDidChangeWatchedFiles` now exists twice.** #866 introduced the constant and a dedicated `case` that localizes resident invalidation per URI and withdraws lint findings for a deleted watched file. This branch had added the same method to a combined `case methodDidChangeConfiguration, methodDidChangeWatchedFiles`. Taking both sides verbatim is a duplicate constant case — a compile error, and had it compiled, #866's handler would have been dead code and its fix silently reverted.

Resolved by giving each notification its own home: the combined case is now `case methodDidChangeConfiguration` alone, and the corpus refresh moved *inside* `invalidateForWatchedFileChanges`, where it can honor a rule this branch did not know about. #866 refuses to drop the resident daemon's warm Program when the editor reports `"changes": []` — "dropping warm state here would throw away the residency the daemon exists to provide for nothing." That reasoning is stronger for a corpus refresh, which spawns a sidecar per plugin and loads a Program in each. So an empty `changes` array now schedules no refresh, and unreadable params take the same conservative "assume everything changed" branch #866 gives the daemon.

Two test cases were added for the boundary the merge created: a watched-file notification with no params must refresh, and one with an empty `changes` array must not. `driver-api.mdx` states the same rule.

**2. `protocol.mdx` hint paragraph.** #866 added the UTF-16 position-encoding paragraph directly above the `lsp-hints` paragraph this branch rewrote. Both kept, in that order.

### Neither landed change invalidates an assumption behind the refresh path

**#867 (lexical JSDoc scope).** The scanner it added decides *where a hint applies*, per completion request, downstream of the corpus. This branch changes *what the corpus contains* and when it is refetched. The only corpus field it reads is `After`'s last rune, for trigger characters, which #867 did not touch. No overlap.

**#866 (residency and position encoding).** The encoding work constrains `general.positionEncodings` on the initialize *request* path, before the frame is forwarded upstream. `rememberAdvertisedCompletionTriggers` reads the initialize *response* capabilities, so it still records exactly the trigger list the editor receives. The residency work is the conflict above and is now integrated rather than bypassed; `didSave` continues to use `invalidateResidentPlugins(env)`, which #866 kept alongside the new per-URI helper.

### Verification on the new base

- `go build ./...`, `go vet ./internal/lspserver ./driver` — clean.
- `go test ./internal/lspserver ./test/driver -count=1` — pass.

Issue #844, opened 2026-07-20
fix(ttsc): resolve CLI flag names the way the compiler the launcher wraps resolves them


samchon:
## Problem

`README.md` states the product contract: "`ttsc` is a drop-in replacement for `tsc`. It reads the same `tsconfig.json`, takes the same flags, and emits the same JavaScript".

TypeScript's option parser, legacy `tsc` and native tsgo alike, matches option names case-insensitively. The ttsc launcher does not. Its schema index is a plain map keyed on the exact spelling, and the parser does an exact lookup.

The consequence is not a rejection, which would be visible. A case variant that tsgo treats as identical to the canonical spelling falls through the launcher's unknown-flag escape hatch and reaches tsgo, where it works, while every ttsc-side consumer of that flag never fires. The user gets a build in which tsgo honored the flag and ttsc did not, with no diagnostic.

Three verified instances:

- `ttsc --listemittedfiles` prints nothing, because tsgo emits the `TSFILE:` lines and ttsc strips them as its own internal protocol noise;
- `ttsc --showconfig` prints a resolved config carrying ttsc's injected `"noEmitOnError": true`, because the launcher did not recognize a print-and-exit flag and added its compile-only guard anyway;
- `ttsc --Watch` hands the session to tsgo's own watch loop, so the plugin-aware ttsc watch never runs and no plugin transform is applied on any rebuild.

Expected behavior: a spelling tsgo resolves to option X resolves to X in the launcher too, or ttsc rejects it. Silent forwarding of a flag ttsc claims to own is the failure.

## Evidence

`packages/ttsc/src/flags/schema.ts`, `packages/ttsc/src/flags/parser.ts`, and `packages/ttsc/src/compiler/internal/runBuild.ts` are byte-identical between the audited commit `2b724664eee71c57cd688c55122cac137774bf62` and `master` at `612abc2d20d3f99429319081c1f6520bbe34ac7a`.

### Root cause

`FLAG_BY_NAME` in `schema.ts` is built once as a `Map<string, FlagSpec>` keyed on each flag's exact name and aliases. `parseFlags` in `parser.ts` resolves a token with two exact lookups in sequence:

```ts
const flag = accepted.get(token);
if (flag !== undefined) { ... }
const globalFlag = FLAG_BY_NAME.get(token);
if (globalFlag !== undefined) { ... }
if (current.startsWith("-")) { passthrough.push(current); continue; }
```

[...]

PR #875, opened 2026-07-20
fix(ttsc): share cache-subcommand flag identity

### LINKED ISSUE #873: fix(ttsc): the cache subcommand parses flags outside the identity #844 made shared
## Problem

`ttsc cache paths` parses its own flags with a hand-rolled exact-string switch, so it is the one CLI surface #844 did not reach. `ttsc --listemittedfiles` now resolves like `--listEmittedFiles`, but `ttsc cache paths --JSON` still errors as an unknown flag.

The subcommand has no schema rows at all, so it never consults the normalization or the lookup that #844 made the single owner of flag identity. Every other lane — the parser, the `runBuild` predicates, both launcher dispatchers, and the generated Go allow-lists — now agrees on what a flag name is. This one still has its own answer.

## Evidence

Verified on `master` at `bc5fd2658`, after #844 merged.

`packages/ttsc/src/launcher/internal/runTtsc.ts` dispatches `cache` to `runCache`, which reaches `parseCachePathsArgs`. That function switches on literal strings — `--json`, `--cache-dir`, `--cwd`, `-p`, `--project`, `--tsconfig` — and sends anything else to its default arm. No call reaches `normalizeFlagToken` or `resolveFlagSpec`.

Found by the agent that implemented #844, while confirming that its normalization had reached every consumer. It is reported rather than fixed there because the `cache` subcommand sits outside that issue's declared surface.

## Consequence surface

| Input | `ttsc build` lane after #844 | `ttsc cache paths` today |
| --- | --- | --- |
| `--json` | resolves | resolves |
| `--JSON` | resolves | unknown flag, exits non-zero |
| `--Cwd value` | resolves | unknown flag, exits non-zero |
| `-P value` | resolves through the alias rule | unknown flag, exits non-zero |

The user-visible effect is a CLI that answers the same spelling differently depending on which subcommand precedes it. `cache paths --json` is documented for CI use, so the failure lands in scripts, where a case variant is easy to write and the error text does not suggest that casing is the reason.

## Approach

Give the `cache` subcommand and its `paths` verb schema rows, and let the shared identity and lookup resolve them, so no CLI surface keeps a private notion of what a flag name is.

The fix is not complete when `cache paths` accepts `--JSON`. It is complete when no launcher path can parse a flag without going through the shared owner — the same requirement #844 set for the lanes it covered, extended to the surface it did not.

- `ttsc cache paths --JSON`, `--Cwd`, `-P` and the canonical spellings all resolve identically, and the output is unchanged for the canonical forms.
- An unknown flag still fails, with the message it produces today.
- The `--json` value rejection (`--json=x` is an error) survives, since it is a real constraint rather than a parsing accident.
- A guard fails if a launcher subcommand parses flags without the shared lookup.
- The generated flag documentation and the Go allow-lists regenerate cleanly through the repository's own generator; the schema addition must not be hand-edited into generated output.
- `TTSC_BUILD_SCOPE=test-ttsc pnpm run build:current` then the `features/compiler` and `features/project` suites stay green, with any new case named and confirmed in the output.

## Coordination

- #844 established the identity rule and the lookup this issue reuses; it is merged, and nothing here changes it.
- #846 derived the project-free flag set from the schema. Adding `cache` rows should follow the same derivation rather than introducing a second classification.
- Related in kind, not in file: #871 records the other place this campaign found a rule sealed in one copy and left in another. The pattern is the reason both are filed as class issues rather than as one-line fixes.

samchon:
Campaign batch: **CLI cache-subcommand flag identity**.

Closes #873.

`cache paths` is the remaining launcher path that parses its flags with a
literal-string switch instead of the schema identity #844 made authoritative.
This batch will make its supported `--json`, `--cache-dir`, `--cwd`, and
project spellings resolve through that shared owner while preserving its
subcommand-specific value and unknown-option errors.

## Scope

- `packages/ttsc/src/launcher/internal/runTtsc.ts` and focused launcher tests.
- The regression covers canonical and case-variant spellings, `-P`, inline
  value rejection for `--json`, and the unchanged unknown-option failure.
- No generated output is hand-edited; a schema change, if evidence requires
  one, moves through the existing generator and flag checks.

## Campaign constraints

- The branch is an empty reservation at opening; implementation and local
  verification are pending.
- Campaign Actions are not modified. Exact-SHA runs for every push and this
  pull-request creation are cancelled and recorded; local verification is the
  merge gate.
- `pnpm format` is intentionally not run. Post-Campaign Cleanup owns the
  repository-wide formatter result.

### DIFF
--- a/packages/ttsc/src/flags/schema.ts
+++ b/packages/ttsc/src/flags/schema.ts
@@ -26,6 +26,7 @@
 export type TtscSubcommand =
   | "ttsc"
   | "build"
+  | "cache"
   | "check"
   | "fix"
   | "format"
@@ -221,6 +222,7 @@ export const FLAG_SCHEMA: readonly FlagSpec[] = [
       "ttsc",
       "ttsx",
       "build",
+      "cache",
       "check",
       "fix",
       "format",
@@ -237,6 +239,7 @@ export const FLAG_SCHEMA: readonly FlagSpec[] = [
       "ttsc",
       "ttsx",
       "build",
+      "cache",
       "check",
       "fix",
       "format",
@@ -336,6 +339,7 @@ export const FLAG_SCHEMA: readonly FlagSpec[] = [
       "ttsc",
       "ttsx",
       "build",
+      "cache",
       "check",
       "fix",
       "format",
@@ -344,6 +348,13 @@ export const FLAG_SCHEMA: readonly FlagSpec[] = [
     ],
     consumedBy: ["launcher"],
     description: "Override the runner and source-plugin cache root.",
+  },
+  {
+    name: "--json",
+    kind: "boolean",
+    subcommands: ["cache"],
+    consumedBy: ["launcher"],
+    description: "Print cache paths as JSON.",
   },
 
   // -------------------------------------------------------------------------
--- a/packages/ttsc/src/launcher/internal/runTtsc.ts
+++ b/packages/ttsc/src/launcher/internal/runTtsc.ts
@@ -307,8 +307,14 @@ function parseCachePathsArgs(argv: readonly string[]): {
   const rest = [...argv];
   while (rest.length !== 0) {
     const token = rest.shift()!;
-    const [flag, inlineValue] = splitInlineFlag(token);
-    switch (flag) {
+    const [rawFlag, inlineValue] = splitInlineFlag(token);
+    const flag = resolveFlagSpec(rawFlag);
+    if (flag?.subcommands.includes("cache") !== true) {
+      throw new Error(
+        `ttsc: cache paths does not support ${JSON.stringify(token)}`,
+      );
+    }
+    switch (flag.name) {
       case "--json":
         if (inlineValue !== undefined) {
           throw new Error("ttsc: --json does not take a value");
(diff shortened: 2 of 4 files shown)


samchon:
Self-Review round 1 found one missing boundary in the new regression. The
test keeps the existing inline rejection for `--json=true`, but it does not
pin the spaced boolean-looking value `--json false`. `cache paths` has always
treated that second token as unsupported, and accepting it would silently
weaken this subcommand's value contract while moving it toward the generic
boolean parser.

I will add that negative case, then commit it and restart the complete solo
Self-Review from the resulting snapshot.

samchon:
Self-Review round 2 found one more boundary that the regression should name:
`--binary` is a valid global schema flag but is not a `cache paths` flag. The
implementation correctly rejects it through the `subcommands.includes("cache")`
gate, but the test currently proves only a wholly unknown option. Add the
known-but-unaccepted twin so a future cache row or parser change cannot turn a
shared-schema lookup into accidental support for another command's option.

I will add the assertion, commit it, and restart the complete solo review on
that new immutable snapshot.

samchon:
Final solo Self-Review completed cleanly on `7827f72490fd904916a90df6d4d0feea971aafc1`.

The final surface is limited to the cache flag schema, `runTtsc`'s cache-path
parser, the generated flag reference, and one real-launcher regression. The
review rechecked flag identity, alias and case normalization, inline and
spaced boolean boundaries, known-but-out-of-scope schema flags, unknown flags,
project resolution, generated-doc ownership, ttsx and build-lane forwarding,
and Windows path execution. No further sound improvement remained.

Local verification:

- `pnpm --filter ttsc build` passed before the immutable implementation
  snapshot; later commits changed only the regression assertions.
- `TTSC_BUILD_SCOPE=test-ttsc pnpm run build:current` completed its scoped
  package build in the campaign worktree. The background command's log has no
  error output and the current `ttsc` launcher and Win32 binary were produced.
- `pnpm run check:flags` passed: generated artifacts match the schema and all
  six declared compiler-flag kinds agree with the pinned upstream option table.
- `TTSC_TEST_DIR=features/compiler pnpm --filter @ttsc/test-ttsc start --
  --include=ttsc_cache_paths_resolves_schema_flag_spellings` passed.
- The compiler subtree also executed both cache-path cases successfully. Its
  two failures were the known Windows `spawnSync ... tsc.exe UNKNOWN` fake
  preview limitation, unrelated to this batch. The project subtree's only
  failure was the known Windows `fs.symlinkSync` `EPERM` limitation; its
  cache/project configuration cases passed.

Campaign CI remains intentionally cancelled. Every run observed for the final
SHA is terminal `cancelled` in the exact-SHA records, including the one
queued `test` run that needed a force-cancel after GitHub returned HTTP 500 to
the first ordinary cancellation attempt.

PR #876, opened 2026-07-20
fix(vscode): preserve literal percent through command shims

### LINKED ISSUE #804: fix(vscode): deliver a literal percent sign to the Windows command-shim child instead of doubling it
## Problem

`@ttsc/vscode` builds a Windows `.cmd`/`.bat` launch as one pre-quoted `cmd.exe /d /s /c` payload. The helper that quotes each argument escapes `%` by doubling it. `%%` is a percent escape inside a **batch file's own source text**; it is not an escape on a `cmd /c` command line, and `cmd` performs no `%%` to `%` reduction there.

The doubling therefore fails in both directions at once:

- a **literal** `%` reaches the child as `%%`, so an argument that names a real path names a path that does not exist;
- a **`%NAME%`-shaped** segment is still expanded by `cmd`, and the doubling leaves two stray percent characters around the expansion, which is strictly worse than passing the argument undoubled.

`%` is a legal Windows path character. When the workspace, the `node_modules` tree, the tsconfig path, or the packaged `.vsix` sits under a directory containing one, the language server either starts against a nonexistent project or fails to start at all, and `npx @ttsc/vscode install` fails with a generic CLI status message.

Expected behavior: every element of the payload reaches the child process byte-identical to the string the caller passed.

Affected: Windows users of the VS Code extension whose `ttsc.serverPath` resolves to a `.cmd`/`.bat` shim (the ordinary shape of `node_modules/.bin/ttscserver.cmd`), and every Windows user of the `@ttsc/vscode` install CLI. POSIX is unaffected, and the `.exe` and `.js` launcher branches never reach this helper.

## Evidence

Both files below are byte-identical between the audited commit `2b724664eee71c57cd688c55122cac137774bf62` and `master` at `612abc2d20d3f99429319081c1f6520bbe34ac7a`. `packages/vscode/README.md` and `packages/vscode/package.json` did move on master; no code cited here did.

### The duplicated helper

Two byte-identical copies build the payload:

- `packages/vscode/src/serverResolution.ts` declares `quoteWindowsArg`, used by the exported `quoteWindowsCommand`, reached from `createServerLaunchCommand` (the `isWindowsCommandLauncher` branch) and `createServerExecutable`, which is the `Executable` handed to `vscode-languageclient`.
- `packages/vscode/bin/install.js` declares its own `quoteWindowsArg` and `quoteWindowsCommand`, reached from `createCodeCommand` (the `platform === "win32"` branch) and then `spawnCode`/`runCode`/`main`.

Both bodies read:

```js
return `"${String(arg).replace(/"/g, '\\"').replace(/%/g, "%%")}"`;
```

A fix applied to one copy leaves the other surface broken, so the duplication is part of the defect and not incidental.

### Reproduction against a real `cmd.exe` child

The probe loads the production helper, builds the payload it would hand `vscode-languageclient`, and spawns `cmd.exe` with the exact options the extension uses, printing back the argument the child actually received. `TTSC_QUOTE_SENTINEL` is set to `EXPANDED` in the child environment so the expansion half is observable.

```powershell
@'
import { pathToFileURL } from 'node:url';
import { spawnSync } from 'node:child_process';
import path from 'node:path';
const mod = await import(pathToFileURL(path.resolve('packages/vscode/src/serverResolution.ts')).href);
const literal = String.raw`C:\workspace\%TTSC_QUOTE_SENTINEL%\100%`;
const payload = mod.quoteWindowsCommand([process.execPath, '-p', 'process.argv[1]', literal]);
const env = { ...process.env, TTSC_QUOTE_SENTINEL: 'EXPANDED' };
const result = spawnSync(env.ComSpec || 'cmd.exe', ['/d', '/s', '/c', payload], {
  encoding: 'utf8', env, windowsVerbatimArguments: true,
});
console.log(JSON.stringify({ literal, payload, status: result.status, stdout: result.stdout.trim() }));
'@ | node --disable-warning=ExperimentalWarning --experimental-transform-types --input-type=module
```

Observed output:

```text
{"literal":"C:\\workspace\\%TTSC_QUOTE_SENTINEL%\\100%","payload":"\"\"C:\\Program Files\\nodejs\\node.exe\" \"-p\" \"process.argv[1]\" \"C:\\workspace\\%%TTSC_QUOTE_SENTINEL%%\\100%%\"\"","status":0,"stdout":"C:\\workspace\\%EXPANDED%\\100%%"}
```

The child should have received `literal` unchanged. It received `C:\workspace\%EXPANDED%\100%%`: the sentinel was expanded and the trailing percent was doubled.

A shell sanity check confirms the premise directly: `cmd /d /s /c "echo 100%%"` prints `100%%`, so the doubling is not undone anywhere on this path.

### Reproduction through the real launch boundary

A second probe mirrors the repository's own boundary test `tests/test-ttsc/src/features/ttscserver/test_vscode_server_launch_command_spawns_windows_command_shim.ts`, using the same recorder `.cmd` and the same `createServerExecutable` options, with one character added to the fixture directory name (`Tools & SDK (x86) 100% done`):

```text
payload : ""C:\...\Tools & SDK (x86) 100%% done\server.cmd" "--stdio"
          "--cwd=C:\...\Tools & SDK (x86) 100%% done\my project" ... "
status  : 1
stderr  : "The system cannot find the path specified."
expected: ["--stdio","--cwd=C:\\...\\Tools & SDK (x86) 100% done\\my project", ...]
recorded: null
MATCH   : false
```

The shim never executed: the launcher path itself was mangled, so `cmd` could not find it. Driving `createCodeCommand` from `bin/install.js` the same way gives the install half:

```text
payload   : ""…\code.cmd" "--install-extension" "C:\tmp & 100%%\ttsc.vsix" "--force""
child argv: ["--install-extension","C:\\tmp & 100%%\\ttsc.vsix","--force"]
expected  : ["--install-extension","C:\\tmp & 100%\\ttsc.vsix","--force"]
```

and the expansion direction on the server half:

```text
cwd sent   : C:\work\%TEMP%\app
child argv : ["--stdio","--cwd=C:\\work\\%C:\\Users\\…\\Temp%\\app", …]
```

### The existing test pins the corrupted string

`tests/test-ttsc/src/features/ttscserver/test_vscode_install_script_uses_windows_command_shim.ts` feeds the argument `"C:\\tmp & 100%\\ttsc.vsix"`, which its own doc comment describes as containing metacharacters, and then asserts the produced payload:

```ts
'""code.cmd" "--install-extension" "C:\\tmp & 100%%\\ttsc.vsix" "--force""',
```

The same expectation appears a second time for the resolved `code.cmd` path. The test never spawns anything, so it cannot observe that the child receives a different path than the caller passed. The expectation was taken from what the code emits, so it locks the defect in and must be re-derived from observed child argv rather than adjusted to keep passing.

The sibling `test_vscode_server_launch_command_spawns_windows_command_shim.ts` does spawn a real recorder `.cmd`, but its fixture path carries spaces, `&`, and parentheses and no `%`, and its spawn assertions are gated on Windows inside the test body.

### Provenance

`quoteWindowsArg` entered `serverResolution.ts` in `68016abd4` ("fix(vscode): harden ttscserver plugin bridge"); the `bin/install.js` copy arrived with the install script and was carried forward by `3bbeb8675` when the Windows smoke lane was added. Neither commit message mentions percent handling.

Not a duplicate. Closed #675 and its PR #689 own the `windowsVerbatimArguments` re-escaping defect on this same code path; their evidence and acceptance name spaces and shell metacharacters, they never verify percent delivery, and the fix left `quoteWindowsArg`'s body untouched. Closed #791 is the release script's pnpm command shim, a different file. Searches for percent, `quoteWindows`, and `cmd.exe` across open and closed issues return only #675 and #791.

## Consequence surface

| Path | Trigger | Observed effect |
| --- | --- | --- |
| `ttsc.serverPath` resolving to a `.cmd`/`.bat` shim | `%` anywhere in the launcher path | `cmd` cannot find the launcher; the language client never starts, so no plugin diagnostics, code actions, formatting, or `workspace/executeCommand` for that workspace |
| same | `%` in the workspace or tsconfig path | `--cwd=` / `--tsconfig=` name a nonexistent directory, so the server resolves the wrong project or none, silently |
| same | a `%NAME%`-shaped path segment | segment expanded to the environment value and wrapped in stray percent characters |
| `npx @ttsc/vscode install` / `uninstall` | `%` in the resolved `.vsix` path | `code --install-extension <bad path>` fails with only "VS Code CLI exited with status N" |
| `.exe` launcher, `.js`/`.cjs`/`.mjs` launcher, every POSIX platform | any | unaffected; those branches never reach the helper |

The doubling is unconditional, so a path containing `%` is corrupted every time on the two affected paths.

## Approach

Invariant: **an argument handed to `createServerLaunchCommand` or `createCodeCommand` reaches the target process byte-identical, with no environment expansion and no percent rewriting, for every launcher kind.**

Architectural owner: the pair of `quoteWindowsArg`/`quoteWindowsCommand` helpers. Both copies must move together, and consolidating them behind one shared implementation or one shared oracle is preferable to editing two bodies that have already drifted apart in review once.

Two facts constrain the design and belong in the implementer's decision rather than in a prescription here. Removing the doubling fully seals the literal-`%` class and makes the `%VAR%` case no worse than a plain `cmd` invocation. Sealing the `%VAR%` class as well is not achievable by escaping at all, because `cmd` expands on the `/c` line before any quoting the caller can apply; that half requires either not routing through `cmd` when the target is directly spawnable, or delivering volatile values through `options.env`/`options.cwd` rather than through `--cwd=`. Note that `--tsconfig=` and the launcher path itself still need a path-safe channel under either choice.

Out of scope, and verified correct today: the `windowsVerbatimArguments` decision and the `/d /s /c` shape (both owned by closed #675), the POSIX branch, and the `.exe`/`.js` launcher branches.

| Case | Expected |
| --- | --- |
| Positive | A `.cmd` launcher under a directory containing a literal `%` starts, and the child's `process.argv` equals the arguments the non-command launcher passes |
| Positive | `createCodeCommand` with a `.vsix` path containing `%` produces a child argv equal to the intended arguments |
| Negative twin | An argument containing `%NAME%` where `NAME` is a set environment variable arrives unexpanded |
| Negative twin | Spaces, `&`, `^`, and parentheses continue to arrive intact, as the existing fixture already covers |
| Boundary | `%` at the start of an argument, at the end, alone as the whole argument, and a caller string that already contains `%%` |
| Boundary | Empty argument, trailing backslash, and an embedded `"` |
| Regression | `.exe` and `.js` launcher command/args shapes unchanged; `createCodeCommand(args, "linux")` still returns `{command:"code", args, options:{}}`; `test_vscode_server_launch_command_uses_command_mode.ts` still passes |
| Test correction | The two payload expectations in `test_vscode_install_script_uses_windows_command_shim.ts` are re-derived from observed child argv, not adjusted to match new output |

Narrow proving command:

```bash
TTSC_TEST_DIR=features/ttscserver pnpm --filter @ttsc/test-ttsc start
```

That subtree is `tests/test-ttsc/src/features/ttscserver/`, and the two cases this issue changes are `test_vscode_install_script_uses_windows_command_shim.ts` and `test_vscode_server_launch_command_spawns_windows_command_shim.ts`. Both load `packages/vscode/src/serverResolution.ts` and `packages/vscode/bin/install.js` from the workspace directly, so no `@ttsc/vscode` build is required for them to exercise the change.

Two properties of this command matter. The runner exits 1 with "No tests were discovered under …" when the subtree is empty or misspelled, so a mistyped path fails rather than passing silently. But it must be run **on Windows**: the spawn assertions in the launch test are gated on the platform inside the test body, so on Linux the command passes while asserting only the generated string, which is the shape that let this defect through in the first place.

Broader:

```bash
pnpm --filter @ttsc/vscode build
pnpm test
```

The `@ttsc/vscode` build is a packaging sanity check on the bundled extension, not a proof of this behavior.

## Coordination

- The `ttsc fast` lane in `.github/workflows/test.yml` (`dir: features`) runs on `ubuntu-latest`, and the only Windows lane, `windows-go`, runs the Go suites plus the flag check. The existing Windows shim test therefore compiles everywhere but only asserts child argv when a developer runs it on Windows. A percent case asserted only inside that Windows-gated branch will not be proven by CI. Giving these cases a Windows runner is the preferred outcome. If the skip stays instead, **it must announce itself in the run output**, naming the assertions it did not perform: a silent skip on the only platform where the defect exists is how this class survives, and this repository already carries an issue about executable test bodies that no lane runs.
- Closed #675 and merged PR #689 own `windowsVerbatimArguments` propagation on this exact path. Do not re-file or revisit that behavior; it is correct and must stay.
- Closed #791 (release script pnpm command shim) is a different file and a different failure.
- Merged PR #790 touched `packages/vscode/README.md` and `packages/vscode/package.json` only. Neither `src/serverResolution.ts` nor `bin/install.js` changed on master.
- Independent of every other candidate in this audit; no file under `packages/graph` is involved.

samchon:
Closes #804.\n\nCampaign reservation for the duplicated Windows command-shim quoting surface. Implementation, Windows child-argv regression coverage, local verification, and solo self-review follow after the required exact-SHA CI gate.

### DIFF
--- a/packages/vscode/bin/install.js
+++ b/packages/vscode/bin/install.js
@@ -22,15 +22,19 @@ function createCodeCommand(
 ) {
   if (platform === "win32") {
     const launcher = findWindowsCodeCommand(env, deps);
+    const commandShim = createWindowsCommandShim([launcher, ...args]);
     return {
       command: env.ComSpec || "cmd.exe",
-      args: ["/d", "/s", "/c", quoteWindowsCommand([launcher, ...args])],
+      args: ["/d", "/s", "/c", commandShim.payload],
       // The `/c` payload is already fully quoted. Node's Windows spawn escapes
       // each arg again unless told not to, turning `""code.cmd" ...` into
       // `"\"\"code.cmd\" ...\""`, which cmd.exe sees as one un-runnable token
       // after the CVE-2024-27980 argument-escaping change. Pass the payload
       // verbatim so cmd's `/s` strips the outer quotes and runs `code.cmd ...`.
-      options: { windowsVerbatimArguments: true },
+      options: {
+        env: { ...env, ...commandShim.environment },
+        windowsVerbatimArguments: true,
+      },
     };
   }
   return { command: "code", args, options: {} };
@@ -78,12 +82,24 @@ function findWindowsCodeCommand(env = process.env, deps = {}) {
   return candidates.find((candidate) => existsSync(candidate)) ?? "code.cmd";
 }
 
-function quoteWindowsCommand(args) {
-  return `"${args.map(quoteWindowsArg).join(" ")}"`;
+function createWindowsCommandShim(args) {
+  const prefix = "TTSC_VSCODE_COMMAND_SHIM_ARG_";
+  const environment = Object.fromEntries(
+    args.map((arg, index) => [prefix + index, quoteWindowsArg(arg)]),
+  );
+  return {
+    environment,
+    // cmd expands every placeholder exactly once. The environment values are
+    // already Windows-command-line arguments, so their literal `%` segments do
+    // not expand again and their quotes continue to protect shell metacharacters.
+    payload: `"${args.map((_, index) => `%${prefix}${index}%`).join(" ")}"`,
+  };
 }
 
 function quoteWindowsArg(arg) {
-  return `"${String(arg).replace(/"/g, '\\"').replace(/%/g, "%%")}"`;
+  return `"${String(arg)
+    .replace(/(\\*)"/g, "$1$1\\\"")
+    .replace(/(\\*)$/, "$1$1")}"`;
 }
 
 function spawnCode(args, vsixForFallback) {
@@ -160,5 +176,4 @@ module.exports = {
   findWindowsCodeCommand,
   findVsix,
   main,
-  quoteWindowsCommand,
 };
--- a/packages/vscode/src/serverResolution.ts
+++ b/packages/vscode/src/serverResolution.ts
@@ -28,6 +28,7 @@ export type ServerProcessOptions = {
 export type ServerLaunchCommand = {
   args: string[];
   command: string;
+  commandShimEnvironment?: NodeJS.ProcessEnv;
   // True only for the Windows `.cmd`/`.bat` command-shim launch path, whose
   // `args` are already a single fully quoted `/c` payload. Non-command launchers
   // omit it so their argument arrays keep Node's default escaping.
@@ -167,9 +168,11 @@ export function createServerLaunchCommand(
     return { command: process.execPath, args: [launcher, ...args] };
   }
   if (platform === "win32" && isWindowsCommandLauncher(launcher)) {
+    const commandShim = createWindowsCommandShim([launcher, ...args]);
     return {
       command: env.ComSpec || "cmd.exe",
-      args: ["/d", "/s", "/c", quoteWindowsCommand([launcher, ...args])],
+      args: ["/d", "/s", "/c", commandShim.payload],
+      commandShimEnvironment: commandShim.environment,
       windowsVerbatimArguments: true,
     };
   }
@@ -200,13 +203,13 @@ export function createServerExecutable(
     args: launch.args,
     options:
       options && launch.windowsVerbatimArguments
-        ? { ...options, windowsVerbatimArguments: true }
+        ? {
+            ...options,
+            env: { ...options.env, ...launch.commandShimEnvironment },
+            windowsVerbatimArguments: true,
+          }
         : options,
   };
-}
-
-export function quoteWindowsCommand(args: readonly string[]): string {
-  return `"${args.map(quoteWindowsArg).join(" ")}"`;
 }
 
 export function createDocumentSelectorPattern<T>(
@@ -420,8 +423,27 @@ function isWindowsCommandLauncher(launcher: string): boolean {
   return [".cmd", ".bat"].includes(path.extname(launcher).toLowerCase());
 }
 
+function createWindowsCommandShim(args: readonly string[]): {
+  environment: NodeJS.ProcessEnv;
+  payload: string;
+} {
+  const prefix = "TTSC_VSCODE_COMMAND_SHIM_ARG_";
+  const environment = Object.fromEntries(
+    args.map((arg, index) => [prefix + index, quoteWindowsArg(arg)]),
+  );
+  return {
+    environment,
+    // cmd expands every placeholder exactly once. The environment values are
+    // already Windows-command-line arguments, so their literal `%` segments do
+    // not expand again and their quotes continue to protect shell metacharacters.
+    payload: `"${args.map((_, index) => `%${prefix}${index}%`).join(" ")}"`,
+  };
+}
+
 function quoteWindowsArg(arg: string): string {
-  return `"${String(arg).replace(/"/g, '\\"').replace(/%/g, "%%")}"`;
+  return `"${String(arg)
+    .replace(/(\\*)"/g, "$1$1\\\"")
+    .replace(/(\\*)$/, "$1$1")}"`;
 }
 
 /**
--- a/scripts/smoke-vscode-install.cjs
+++ b/scripts/smoke-vscode-install.cjs
@@ -65,6 +65,7 @@ function run(command, args, options = {}) {
     cwd: options.cwd ?? process.cwd(),
     encoding: "utf8",
     stdio: options.capture ? "pipe" : "inherit",
+    env: options.env,
     windowsHide: true,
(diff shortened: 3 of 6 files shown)


samchon:
Solo Self-Review completed after the final implementation surface was in place. I rechecked both command-shim producers, the extension executable handoff, the smoke runner, the cmd/bat and non-command branches, standard Windows argv escaping, and the actual child-argv fixtures. No further finding.\n\nLocal evidence: pnpm --filter @ttsc/vscode build; pnpm --filter ttsc build; TTSC_TEST_DIR=features/ttscserver pnpm --filter @ttsc/test-ttsc start (all passed on Windows). The new code.cmd recorder covers literal %, %NAME%, %%, &, ^, empty args, trailing backslashes, and embedded quotes.

PR #878, opened 2026-07-20
fix(paths): preserve module reference semantics

### LINKED ISSUE #805: fix(paths): stop rewriting global ambient module names and shadowed `require` calls
## Problem

`@ttsc/paths` rewrites `compilerOptions.paths` aliases into relative specifiers in the emit. It decides what to rewrite from syntax shape alone: every `ModuleDeclaration` whose name is a string literal, and every call whose callee is an identifier spelled `require`. Two ordinary source shapes match that test without being module references the plugin may rewrite.

1. A **global ambient external module declaration** in a script file. `declare module "@lib/message" { ... }` in a file with no top-level `import`/`export` declares an ambient module, and TypeScript forbids a relative name there. The plugin rewrites it anyway, reports success, and produces source the compiler rejects with TS2436.
2. A **shadowed `require`**. `function use(require: (id: string) => string)` binds `require` as an ordinary parameter. The call `require("@lib/message")` is application data flowing into a callback, not a CommonJS module load. The plugin rewrites the argument, so the program's observable return value changes.

Expected behavior: a string literal is rewritten only when it is a module reference of the file that contains it. A declaration that cannot legally take a relative name, and a call to an identifier that resolves to a local binding, are not module references.

Affected users: everyone who enables `@ttsc/paths`, through `ttsc` emit, `ttsx`, the `@ttsc/unplugin` and `@ttsc/metro` bundler adapters, or the wasm playground host. Case 1 turns a valid project into failing output with a success exit code. Case 2 changes runtime behavior silently.

## Evidence

`packages/paths/driver/paths.go` is byte-identical between the audited commit `2b724664eee71c57cd688c55122cac137774bf62` and `master` at `612abc2d20d3f99429319081c1f6520bbe34ac7a`. `git diff 2b724664 612abc2d2 -- packages/paths` reports one changed file, `packages/paths/package.json`, a version bump.

### Root cause

`visitModuleSpecifiers` in `packages/paths/driver/paths.go` classifies by node kind and forwards the literal to `rewriter.apply`:

```go
case shimast.KindModuleDeclaration:
  decl := node.AsModuleDeclaration()
  if decl != nil {
    visit(decl.Name())
  }
case shimast.KindCallExpression:
  call := node.AsCallExpression()
  if isModuleSpecifierCall(call) && call.Arguments != nil && len(call.Arguments.Nodes) > 0 {
    visit(call.Arguments.Nodes[0])
  }
```

```go
func isModuleSpecifierCall(call *shimast.CallExpression) bool {
  if call == nil || call.Expression == nil {
    return false
  }
  switch call.Expression.Kind {
  case shimast.KindImportKeyword:
    return true
  case shimast.KindIdentifier:
    return call.Expression.Text() == "require"
  default:
    return false
  }
}
```

`rewriter.apply` then resolves the literal through the paths table and mutates it in place. Neither branch consults the fact that decides the question: whether the containing file is an external module (which turns `declare module "X"` from an ambient declaration into an augmentation), and whether the `require` identifier resolves to a local binding rather than the CommonJS one.

The affected files are in scope: `packages/ttsc/driver/program.go::sourceFilesRaw` excludes declaration files but includes ordinary `.ts` scripts, and `plugin.ApplyProgram` walks `driver.Program.SourceFiles()`.

Create a project with these four files:

```jsonc
// tsconfig.json
{
  "compilerOptions": {
    "module": "commonjs",
    "declaration": true,
    "outDir": "dist",
    "paths": { "@lib/*": ["./src/lib/*"] },
    "plugins": [{ "transform": "@ttsc/paths" }],
    "rootDir": "src",
    "strict": true,
    "target": "ES2022"
  },
  "include": ["src"]
}
```

```ts
// src/global.ts
declare module "@lib/message" { export const declared: string; }
```

```ts
// src/custom.ts
export function use(require: (id: string) => string): string { return require("@lib/message"); }
```

```ts
// src/lib/message.ts
export const message = "ok";
```

The unmodified fixture type-checks clean under TypeScript 7.0.2 (`tsc --noEmit` exits 0). Run the real sidecar from `packages/paths`:

```bash
go run ./plugin transform \
  --cwd=<fixture> \
  --tsconfig=<fixture>/tsconfig.json \
  --plugins-json='[{"name":"@ttsc/paths","stage":"transform","config":{"transform":"@ttsc/paths"}}]' \
  --quiet
```

Observed, with the transformed `global.ts` written back out and re-checked:

```text
transform status=0
transformed global.ts="declare module \"./lib/message.js\" {\n    export const declared: string;\n}\n"
transformed custom.ts="export function use(require: (id: string) => string): string { return require(\"./lib/message.js\"); }\n"
original custom result="@lib/message"
transformed custom result="./lib/message.js"
recheck status=1
global.ts(1,16): error TS2436: Ambient module declaration cannot specify relative module name.
```

`original custom result` and `transformed custom result` are the value returned by calling `use((id) => id)` on the original and transformed sources. The transform changed a program's output data.

### Existing coverage and why it passes

`packages/paths/test/command_rewrites_all_module_specifier_forms_test.go` exercises every supported form in one fixture file, `src/main.ts`. That file contains top-level `import`/`export` declarations, so its `declare module "@ambient/thing"` is a **module augmentation**, which may legally take a relative name and must keep being rewritten. The same fixture models the CommonJS call with an ambient `declare function require(id: string): unknown;`, so the identifier resolves to a module loader. Its negative cases are `fn("@lib/message")` and `obj.require("@lib/message")`.

The two semantic boundaries this issue is about, script file versus module file, and shadowed versus real `require`, have no case in either direction.

### Provenance

`18de54176` (PR #91, "feat(ttsc): remove hard-coded 1st party plugin logics, and make 100% test coverages.") introduced both classifier branches together with the all-forms test. PR #314 ("fix(paths): terminate rootDir inference at volume roots and mirror tsc pattern precedence") refactored the visitor and explicitly kept the ambient-declaration coverage while fixing unrelated path math and pattern precedence. Neither change touched the classification question. Issue #680 and PR #691 changed emitted-extension prediction and are unrelated.

No **Forbidden** classification is asserted. The history shows an intended general syntax-coverage feature with one implementation, not fixture-only logic, a knowingly forced design, or a repeated symptom patched narrowly.

### Duplicate check

A comparison across all open and closed issues and pull requests, plus focused searches for ambient relative declarations, module augmentations, shadowed `require`, and custom `require`, found no owner. No open pull request touches `packages/paths`.

## Consequence surface

| Source shape | Containing file | Today | Required |
| --- | --- | --- | --- |
| `declare module "@lib/x" {}` | script (no top-level import/export) | rewritten to `"./lib/x.js"`, output fails TS2436 | unchanged |
| `declare module "@lib/x" {}` | module (has top-level import/export) | rewritten (augmentation) | rewritten, unchanged behavior |
| `require("@lib/x")` where `require` is a parameter, local, or imported binding | any | rewritten, return value changes | unchanged |
| `require("@lib/x")` resolving to the CommonJS or ambient-declared loader | any | rewritten | rewritten, unchanged behavior |
| `import`, `export ... from`, `import(...)`, `import type`, `import x = require(...)` | any | rewritten | rewritten, unchanged behavior |
| `fn("@lib/x")`, `obj.require("@lib/x")` | any | unchanged | unchanged |
| `namespace foo {}`, `module foo {}` | any | unchanged (the name is an `Identifier`, filtered by the `KindStringLiteral` check in `apply`) | unchanged |

Reached from every consumer of the transform: `ttsc` build and `--emit`, `ttsx`, `@ttsc/unplugin` and `@ttsc/metro`, and the wasm host. Declaration inputs are excluded by `sourceFilesRaw`, so the affected ambient-declaration surface is ordinary script source (`.ts`, `.tsx`, `.mts`, `.cts` and the JS equivalents the program parses).

Platform-independent: this is AST classification with no path or filesystem component. The path math that runs after classification is not in question here.

`website/src/content/docs/development/walkthroughs/paths.mdx` teaches the syntactic classifier as the complete rule. Its module-specifier table row for `ModuleDeclaration` says the name is a `StringLiteral` "only for the ambient `declare module "pkg"` form", and the `CallExpression` row makes no distinction about `require` resolution. Both must change with the behavior.

## Approach

Invariant: **a string literal is rewritten only when it is a module reference of the file that contains it.** A `declare module "X"` name is a rewritable module reference only when the containing file is an external module (making the declaration an augmentation), and a `require(...)` argument is a rewritable module reference only when the callee identifier is the module loader rather than a binding declared in an enclosing scope.

Architectural owner: `packages/paths/driver/paths.go`, in the classification step before `rewriter.apply` mutates anything. The path math (`rewrite`, `resolveSource`, `outputPathForSource`) is correct and stays as is.

The semantic facts are already reachable at that point and no new plugin protocol is needed: `driver.Program` carries a `Checker`, and `packages/ttsc/shim/ast` aliases the upstream `SourceFile` type directly (`type SourceFile = innerast.SourceFile`), so the parser's own module indicator is available on the node the walk already has. Confirm the exact upstream accessor before relying on it, and follow the typescript-go shim-sync skill if a re-export turns out to be missing rather than reimplementing the classification locally.

Do not solve this by pattern-matching the two witnesses (for example, skipping `declare module` when the file name looks like a global, or refusing to rewrite when a parameter happens to be named `require`). The classification, not the two shapes, is the defect.

| Case | Expected |
| --- | --- |
| Negative | A global ambient external module declaration matching a paths alias, in a file with no top-level import or export, keeps its non-relative name, and the transformed output type-checks clean |
| Positive | An external module augmentation matching the same alias, in a file that is a module, still rewrites to the correct relative emitted path |
| Negative | `require("@lib/x")` where `require` is a function parameter is unchanged, and calling the transformed function returns the original string |
| Negative | The same for a local `const require = ...` and for an imported binding named `require` |
| Positive | A genuine unshadowed CommonJS `require` still rewrites |
| Regression | Every form in `command_rewrites_all_module_specifier_forms_test.go` still rewrites, and `fn("@lib/x")` plus `obj.require("@lib/x")` stay untouched |
| Boundary | `namespace foo {}` and `module foo {}` remain untouched; a `declare module` whose name does not match any pattern remains untouched |

Tests belong beside the existing sidecar tests in `packages/paths/test/`, exercising the real parser through the `transform` command rather than a synthetic AST, with the negative twin next to each positive case.

Narrow proving commands:

```bash
# 23 Go test files under packages/paths/test, including
# command_rewrites_all_module_specifier_forms_test.go
node scripts/test-go-utility-plugins.cjs

# the @ttsc/test-paths feature suite
pnpm --filter @ttsc/test-paths start
```

When running a single Go test directly instead, add `-v` and confirm the test name appears in the output. `go test -run <name>` exits 0 when the pattern matches nothing, exactly as it does for a directory with no test files, so a green exit code alone does not prove the new case ran.

Broader: `pnpm run test:go` and `pnpm test:features`.

The walkthrough page must be updated in the same change, per the development skill's rule that a behavior change updates its matching page under `website/src/content/docs/`.

## Coordination

- Sole owner of `packages/paths/**`; no other audit candidate touches this package, so it can be implemented in parallel with everything else.
- PR #91 introduced the classifier, PR #314 retained it. Neither is a duplicate; both are context for why the syntactic rule looked complete.
- Closed #680 and PR #691 own emitted-extension prediction. Their behavior must not regress.

samchon:
## Intent

Fix #805 by rewriting a paths alias only when the string is a module reference in its containing source file.

## Scope

- classify string-named module declarations using the source file's external-module state;
- classify bare `require` calls with the TypeScript-Go checker so locally or imported bindings stay untouched;
- add real `@ttsc/paths` sidecar regression coverage and update the paths walkthrough.

## Deferred

No shim change is expected: the pinned aliases already expose the required checker, symbol, and source-file state.

## Verification

Pending. This is an implementation-free campaign reservation. Campaign Actions for this branch are deliberately cancelled by exact commit SHA; local sidecar and feature-suite verification plus solo Self-Review are the merge gates.

Closes #805.

### DIFF
--- a/packages/paths/driver/paths.go
+++ b/packages/paths/driver/paths.go
@@ -6,6 +6,7 @@ import (
   "strings"
 
   shimast "github.com/microsoft/typescript-go/shim/ast"
+  shimchecker "github.com/microsoft/typescript-go/shim/checker"
   shimcore "github.com/microsoft/typescript-go/shim/core"
   shimtspath "github.com/microsoft/typescript-go/shim/tspath"
 
@@ -32,6 +33,7 @@ func (plugin) ApplyProgram(prog *driver.Program, _ driver.PluginContext) error {
 // rewriter holds the resolved tsconfig paths configuration used to rewrite
 // module specifiers across an entire program.
 type rewriter struct {
+  checker     *shimchecker.Checker
   basePath    string
   jsxPreserve bool
   outDir      string
@@ -60,6 +62,7 @@ func newRewriter(prog *driver.Program) *rewriter {
   if prog == nil || prog.ParsedConfig == nil || prog.ParsedConfig.ParsedConfig == nil || prog.ParsedConfig.ParsedConfig.CompilerOptions == nil {
     return out
   }
+  out.checker = prog.Checker
   options := prog.ParsedConfig.ParsedConfig.CompilerOptions
   cwd := prog.Host.GetCurrentDirectory()
   out.basePath = filepath.Clean(options.GetPathsBasePath(cwd))
@@ -104,7 +107,7 @@ func (r *rewriter) apply(file *shimast.SourceFile) {
   if r == nil || file == nil || len(r.patterns) == 0 {
     return
   }
-  visitModuleSpecifiers(file.AsNode(), func(lit *shimast.Node) {
+  visitModuleSpecifiers(file, r.checker, func(lit *shimast.Node) {
     if lit == nil || lit.Kind != shimast.KindStringLiteral {
       return
     }
@@ -118,16 +121,16 @@ func (r *rewriter) apply(file *shimast.SourceFile) {
   })
 }
 
-// visitModuleSpecifiers recursively walks the AST rooted at node, calling
-// visit for every string-literal module specifier it finds. Covered nodes
-// include import/export declarations, require() calls, dynamic import()
-// expressions, import-equals declarations, and import-type nodes.
+// visitModuleSpecifiers recursively walks file, calling visit for every
+// string-literal module reference it finds. Syntax identifies candidates, then
+// SourceFile and Checker facts exclude declarations and calls whose literal is
+// not actually a module reference in the containing file.
 //
 // The recursion runs through one closure created per walk, not one per node:
 // handing ForEachChild a fresh closure at every node would allocate once per
 // AST node across the whole program on every apply pass.
-func visitModuleSpecifiers(node *shimast.Node, visit func(*shimast.Node)) {
-  if node == nil {
+func visitModuleSpecifiers(file *shimast.SourceFile, checker *shimchecker.Checker, visit func(*shimast.Node)) {
+  if file == nil {
     return
   }
   var walk func(node *shimast.Node) bool
@@ -152,24 +155,24 @@ func visitModuleSpecifiers(node *shimast.Node, visit func(*shimast.Node)) {
       }
     case shimast.KindModuleDeclaration:
       decl := node.AsModuleDeclaration()
-      if decl != nil {
+      if file.ExternalModuleIndicator != nil && decl != nil {
         visit(decl.Name())
       }
     case shimast.KindCallExpression:
       call := node.AsCallExpression()
-      if isModuleSpecifierCall(call) && call.Arguments != nil && len(call.Arguments.Nodes) > 0 {
+      if isModuleSpecifierCall(checker, call) && call.Arguments != nil && len(call.Arguments.Nodes) > 0 {
         visit(call.Arguments.Nodes[0])
       }
     }
     node.ForEachChild(walk)
     return false
   }
-  walk(node)
+  walk(file.AsNode())
 }
 
 // isModuleSpecifierCall reports whether call is a dynamic import() or a
-// CommonJS require() expression.
-func isModuleSpecifierCall(call *shimast.CallExpression) bool {
+// CommonJS require() expression that resolves to the module loader.
+func isModuleSpecifierCall(checker *shimchecker.Checker, call *shimast.CallExpression) bool {
   if call == nil || call.Expression == nil {
     return false
   }
@@ -177,10 +180,26 @@ func isModuleSpecifierCall(call *shimast.CallExpression) bool {
   case shimast.KindImportKeyword:
     return true
   case shimast.KindIdentifier:
-    return call.Expression.Text() == "require"
+    return call.Expression.Text() == "require" && isModuleLoader(checker, call.Expression)
   default:
     return false
   }
+}
+
+// isModuleLoader reports whether expression names the CommonJS loader rather
+// than a parameter, local, or imported binding named require. An unresolved
+// bare require is the runtime CommonJS global; declared loaders are ambient.
+func isModuleLoader(checker *shimchecker.Checker, expression *shimast.Node) bool {
+  if checker == nil || expression == nil {
+    return false
+  }
+  symbol := checker.GetSymbolAtLocation(expression)
+  if symbol == nil {
+    return true
+  }
+  return symbol.Flags&shimast.SymbolFlagsAlias == 0 &&
+    symbol.ValueDeclaration != nil &&
+    symbol.ValueDeclaration.Flags&shimast.NodeFlagsAmbient != 0
 }
 
 // rewrite resolves specifier from fromSource using the tsconfig paths table and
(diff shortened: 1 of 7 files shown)


samchon:
Solo Self-Review finding: the new loader fixture covers the ambient-symbol path, but not the deliberate nil-symbol path for an unbound CommonJS 
equire. I will add an unbound fixture that executes through the same createRequire runtime and asserts its emitted relative specifier before restarting the complete review on the new snapshot.

samchon:
Final solo Self-Review of 43f1012212f9cf3b64c4257bedccf4a0387c519c against current master is clean. The earlier review found the unbound-
equire semantic branch was unproven; commit 43f10122 added and reran that behavioral case. Local verification is complete: 
ode scripts/test-go-utility-plugins.cjs (banner, paths, strip), scoped uild:current (experimental), and pnpm --filter @ttsc/test-paths start (5/5, 369.141s).

PR #881, opened 2026-07-20
fix(release): verify Windows platform package contents

### LINKED ISSUE #807: fix(release): verify Windows platform package contents instead of skipping the whole package
## Problem

`scripts/assert-platform-package.cjs` is the release gate that proves a `@ttsc/<platform>` package actually carries the binaries it promises. It combines two independent invariants: **file presence** (the native helpers and the bundled Go toolchain are there) and **POSIX executable mode** (the tar headers preserved the executable bit).

Only the second invariant is inapplicable to Windows. The script disables both: in directory mode and in tarball mode it returns before any check as soon as the manifest names a `win32` package. A `@ttsc/win32-x64` package with a valid `package.json` and an empty or missing `bin/` tree passes the gate with exit 0.

That gate runs immediately before publication (`package:latest:publish`), on the `next` channel, and on each packed tarball in `experimental/tarballs`. An installable Windows platform package whose launcher cannot start can therefore be published, and nothing in the release path would say so. The earlier build steps do not cover this: they prove the build produced binaries, not that `npm pack` retained them, and a `files` or `publishConfig` change can open exactly that gap after a successful build.

## Evidence

`scripts/assert-platform-package.cjs` is byte-identical between the audited commit `2b724664eee71c57cd688c55122cac137774bf62` and `master` at `612abc2d20d3f99429319081c1f6520bbe34ac7a`.

### The two early returns

Directory mode:

```js
function inspectPackageDir(dir) {
  const manifestPath = path.join(dir, "package.json");
  const manifest = JSON.parse(fs.readFileSync(manifestPath, "utf8"));
  const platform = platformFromPackageName(manifest.name);
  if (platform === null || platform.os === "win32") return;

  const executablePaths = requiredExecutablePaths(dir);
  inspectExecutablePublishConfig(manifest, executablePaths);
  ...
```

Tarball mode repeats it, before `requiredExecutablePathsFromEntries` and every missing-entry check:

```js
  const manifest = JSON.parse(packageJson.content.toString("utf8"));
  const platform = platformFromPackageName(manifest.name);
  if (platform === null || platform.os === "win32") return;
```

Everything the script does, presence and mode alike, lives after those returns.

Two minimal package directories, each containing only a `package.json` with the platform name and no `bin/` tree at all, run against `master`:

```text
$ node scripts/assert-platform-package.cjs <dir with @ttsc/win32-x64 manifest>
exit=0

$ node scripts/assert-platform-package.cjs <dir with @ttsc/linux-x64 manifest>
@ttsc/linux-x64: publishConfig.executableFiles missing ./bin/ttsc
@ttsc/linux-x64: publishConfig.executableFiles missing ./bin/ttscserver
@ttsc/linux-x64: publishConfig.executableFiles missing ./bin/ttscgraph
@ttsc/linux-x64: publishConfig.executableFiles missing ./bin/go/bin/go
@ttsc/linux-x64: publishConfig.executableFiles missing ./bin/go/bin/gofmt
@ttsc/linux-x64: missing executable bin/ttsc
@ttsc/linux-x64: missing executable bin/ttscserver
@ttsc/linux-x64: missing executable bin/ttscgraph
@ttsc/linux-x64: missing executable bin/go/bin/go
@ttsc/linux-x64: missing executable bin/go/bin/gofmt
exit=1
```

The tarball path behaves the same way. A synthetic gzip tar whose only entry is `package/package.json` carrying the `@ttsc/win32-x64` manifest:

```text
$ node scripts/assert-platform-package.cjs <win32-empty.tgz>
exit=0
```

### What a real Windows package contains

`scripts/build-platform-package.cjs` names the Windows artifacts explicitly, so the required set is not a guess:

```js
const outFile = path.join(outDir, npmOs === "win32" ? "ttsc.exe" : "ttsc");
const serverFile = path.join(outDir, npmOs === "win32" ? "ttscserver.exe" : "ttscserver");
const graphFile = path.join(outDir, npmOs === "win32" ? "ttscgraph.exe" : "ttscgraph");
```

and for the bundled toolchain, `embedGoToolchain` and `verifyEmbeddedGoToolchain` both resolve `bin/go.exe` on Windows. The built `packages/ttsc-win32-x64` in a working tree confirms the layout: `bin/ttsc.exe`, `bin/ttscgraph.exe`, `bin/ttscserver.exe`, `bin/go/bin/go.exe`, `bin/go/bin/gofmt.exe`, and `bin/go/pkg/tool/windows_amd64/{asm,cgo,compile,cover,fix,link,preprofile,vet}.exe`.

### Provenance and why the exception is over-broad

`5a9652931` ("fix(ttsc): preserve platform package executable modes", 2026-07-05) introduced the whole verifier. Its own commit message states the two mode constraints it was built for and, third, the Windows exception it intended:

```text
Constraint: pnpm marks exact publishConfig.executableFiles paths executable in package archives
Constraint: npm package tar headers must carry executable modes for POSIX native helpers and bundled Go tools
Constraint: Windows .exe platform packages do not depend on POSIX executable bits
```

The intent is precise and correct: Windows does not need POSIX mode bits. The implementation placed that exception around the entire inspection routine rather than around the mode test, so it also disabled presence checking, which has nothing to do with mode bits.

The same distinction is visible in the build script, which skips only the POSIX-specific work on Windows:

```js
if (npmOs !== "win32") {
  chmodIfBuilt("ttsc", outFile);
  chmodIfBuilt("ttscserver", serverFile);
  chmodIfBuilt("ttscgraph", graphFile);
  syncExecutablePublishConfig();
}
```

That is why `packages/ttsc-win32-*/package.json` carries no `publishConfig.executableFiles` while `packages/ttsc-linux-x64/package.json` carries all five entries.

### Existing coverage

`tests/test-ttsc/src/features/platform/test_platform_package_tarball_requires_posix_executable_modes.ts` builds synthetic `@ttsc/linux-x64` tarballs and source manifests and asserts the good ones pass and the mode-broken and metadata-incomplete ones fail. It is POSIX-only by construction, which is why the Windows blind spot has no failing test.

### Duplicate check

No open or closed issue or pull request covers platform-package content verification on Windows. No open pull request touches `scripts/assert-platform-package.cjs`.

## Consequence surface

| Path | Consumer | Today for `win32` | Required |
| --- | --- | --- | --- |
| directory mode, `--source` | `pnpm run assert:platform-packages`, called by `package:latest:publish` before any publish command | returns immediately, exit 0 | presence of every required `.exe` and bundled Go tool file |
| directory mode, explicit target | `next.bash` | returns immediately | same |
| tarball mode | `experimental/tarballs/index.ts`, once per packed platform tarball | returns immediately | presence of every required entry inside `package/` |
| POSIX mode check | all of the above | not reached | still skipped on Windows, deliberately |
| `publishConfig.executableFiles` check | all of the above | not reached | still skipped on Windows, see the decision below |

The required Windows path set, derived from the build script rather than invented:

| POSIX required path | Windows equivalent |
| --- | --- |
| `bin/ttsc` | `bin/ttsc.exe` |
| `bin/ttscserver` | `bin/ttscserver.exe` |
| `bin/ttscgraph` | `bin/ttscgraph.exe` |
| `bin/go/bin/go` | `bin/go/bin/go.exe` |
| `bin/go/bin/gofmt` | `bin/go/bin/gofmt.exe` |
| every file under `bin/go/pkg/tool/**` | every file under `bin/go/pkg/tool/**` (`.exe` on Windows) |

In tarball mode the POSIX side derives its required list from `publishConfig.executableFiles` through `configuredExecutablePaths`, falling back to the POSIX `baseExecutablePaths` when the field is absent. Windows manifests have no such field by design, so the Windows required list must come from the `.exe` set above, not from that fallback.

One property of the current design matters when writing the Windows half. The `bin/go/pkg/tool/**` entries are derived from what is already there: `requiredExecutablePaths` walks the directory on disk, and `requiredExecutablePathsFromEntries` reads the tarball's own entries. Requiring them therefore proves nothing on its own; what makes them load-bearing on POSIX is that the manifest's `publishConfig.executableFiles`, synced at build time from the built toolchain, is an **independent** list the artifact is checked against. Do not write a Windows check that enumerates the artifact and then asserts the artifact contains what it enumerated.

**One decision belongs to the release owner**: whether Windows manifests should also declare `publishConfig.executableFiles`.

The evidence says no. That field exists so pnpm marks archive entries executable, the build script deliberately does not sync it for Windows, and the introducing commit names the reason. This issue therefore treats the `executableFiles` assertion as POSIX-only and requires presence only for Windows.

Before making that permanent, check once whether the reason still holds: confirm against the current pnpm and npm behavior that `publishConfig.executableFiles` still does nothing for a Windows package beyond mode marking. If it has gained any other effect, add the declaration to the seven-platform build sync and require it here too. The default above is inherited from a 2026-07-05 commit message, and it deserves one look before it hardens into a rule.

The cost of that choice, stated so it is chosen rather than inherited: without an independent declared list, Windows presence checking is only as strong as the five base `.exe` paths, which is still the difference between catching an empty `bin/` and catching nothing. If the release owner wants Windows toolchain completeness checked to the same depth as POSIX, the independent list has to come from somewhere (a synced manifest field, or a build-time inventory), and that is a larger change than this issue claims. Whichever way it is settled, say so explicitly in the implementation rather than leaving it implied.

## Approach

Invariant: **the release gate proves that every platform package, on every platform, contains the executables it ships. Only the POSIX executable-mode assertions are platform-conditional.**

Architectural owner: `scripts/assert-platform-package.cjs`. Restructure so the platform is a parameter of the required-path derivation, not a reason to stop inspecting, and guard `hasExecutableMode` (and the `executableFiles` metadata assertion that exists to serve it) rather than the routine. Both entry points, `inspectPackageDir` and `inspectTarball`, must go through the same derivation so they cannot drift again.

Failure messages should keep their current shape (`<name>: missing executable <rel>`, `<name>: tarball missing executable <rel>`) so existing expectations and human readers are unaffected.

Do not solve this by special-casing the three top-level `.exe` names. The bundled Go toolchain is the larger half of the payload and is exactly what the mode work of `5a9652931` was written to protect.

| Case | Expected |
| --- | --- |
| Negative | A `@ttsc/win32-x64` directory with a valid manifest and no `bin/` tree fails, naming every missing `.exe` |
| Negative | A `@ttsc/win32-x64` tarball whose only entry is `package/package.json` fails, naming every missing entry |
| Negative | A Windows package missing only `bin/go/bin/gofmt.exe` fails, naming that path |
| Boundary | A Windows package missing a single `bin/go/pkg/tool/**` entry is **not** required to fail under the decision above, because no independent declared list exists for Windows. If the release owner reverses that decision, this becomes a positive requirement instead. |
| Positive | A complete Windows directory and a complete Windows tarball pass, with no `publishConfig.executableFiles` present and no mode assertion applied |
| Positive | A Windows package whose entries carry mode 0644 still passes; POSIX mode is not asserted on Windows |
| Regression | Every existing POSIX case keeps its result: the good and partial Linux tarballs pass, the 0644 `ttscgraph` tarball fails naming the offending path, and the two metadata-incomplete source manifests fail |
| Boundary | A non-platform package name (no `@ttsc/<os>-<arch>` match) is still skipped |
| Boundary | `win32-arm64` behaves identically to `win32-x64` |

The regression belongs beside the existing POSIX case in `tests/test-ttsc/src/features/platform/`, with a Windows counterpart for both the directory path and the tarball path. Both must be runnable on any host, since the current POSIX test already builds synthetic tarballs rather than depending on the running platform.

Narrow proving command:

```bash
# runs tests/test-ttsc/src/features/**, including
# features/platform/test_platform_package_tarball_requires_posix_executable_modes.ts
# and the Windows counterparts added by this issue
TTSC_TEST_DIR=features pnpm --filter @ttsc/test-ttsc start
```

That is the `ttsc fast` CI lane. Confirm the new Windows cases appear by name in the runner output; the suite reports one line per case, so a file that was never picked up is visible rather than silently absent.

Broader: `pnpm test:features`, plus a manual `node scripts/assert-platform-package.cjs --source` against a real built tree.

Consider also running the assertion against the actual packed artifact in the release job rather than only the pre-pack directory, so the tarball path gets exercised on every release instead of only under `experimental/tarballs`.

## Coordination

- Closed #682 owns release preflight (tag and package-version coherence). It runs before publication and validates versions, not package contents. Not a duplicate.
- The commit that introduced the verifier, `5a9652931`, carries the directive "Keep platform tarball mode checks in publish and package:tgz paths unless another tar-header verifier replaces them". This change keeps those checks and adds presence checking beside them.
- Files touched: `scripts/assert-platform-package.cjs` and `tests/test-ttsc/src/features/platform/**`. The sibling packaging candidate about published dependency ranges (#842) touches `packages/*/package.json` and `tests/test-unplugin/**` only, so the two can run in parallel.

samchon:
## Intent

Fix #807 by making the platform-package release gate verify Windows package
contents without imposing POSIX executable-mode requirements.

## Scope

- require the independently derivable Windows `.exe` launcher and bundled Go
  base paths in source-package and tarball inspection;
- retain the POSIX-only `publishConfig.executableFiles` and tar-header mode
  assertions;
- add synthetic Windows directory/tarball coverage beside the existing POSIX
  regression, including the pass/fail boundaries in #807.

## Deferred

The current `pnpm 10.6.4` and `npm 11.14.1` pack comparison confirms that
`publishConfig.executableFiles` only controls pnpm tar-entry mode. This change
therefore does not add it to Windows manifests or infer completeness from the
artifact's own `pkg/tool` listing.

## Verification

Pending. This is an implementation-free campaign reservation. Actions caused
by this branch are cancelled only by exact commit SHA; local platform feature
tests, source/tarball probes, and solo Self-Review are the merge gates.

Closes #807.

### DIFF
--- a/scripts/assert-platform-package.cjs
+++ b/scripts/assert-platform-package.cjs
@@ -76,11 +76,13 @@ function inspectPackageDir(dir) {
       failures.push(`${manifest.name}: missing executable ${rel}`);
       continue;
     }
-    const mode = fs.statSync(file).mode & 0o777;
-    if (platform.os !== "win32" && !hasExecutableMode(mode)) {
-      failures.push(
-        `${manifest.name}: ${rel} has mode ${formatMode(mode)}, expected executable mode`,
-      );
+    if (platform.os !== "win32") {
+      const mode = fs.statSync(file).mode & 0o777;
+      if (!hasExecutableMode(mode)) {
+        failures.push(
+          `${manifest.name}: ${rel} has mode ${formatMode(mode)}, expected executable mode`,
+        );
+      }
     }
   }
 }
--- a/scripts/assert-platform-package.cjs
+++ b/scripts/assert-platform-package.cjs
@@ -7,7 +7,7 @@ const args = process.argv.slice(2);
 const sourceMode = args.includes("--source");
 const targets = args.filter((arg) => !arg.startsWith("--"));
 const failures = [];
-const baseExecutablePaths = [
+const posixBaseExecutablePaths = [
   "bin/ttsc",
   "bin/ttscserver",
   "bin/ttscgraph",
@@ -19,7 +19,7 @@ const minimumExecutablePaths = [
   "bin/go/bin/go",
   "bin/go/bin/gofmt",
 ];
-const requiredExecutableConfigPaths = baseExecutablePaths.map(
+const requiredExecutableConfigPaths = posixBaseExecutablePaths.map(
   (rel) => `./${rel}`,
 );
 
@@ -64,10 +64,12 @@ function inspectPackageDir(dir) {
   const manifestPath = path.join(dir, "package.json");
   const manifest = JSON.parse(fs.readFileSync(manifestPath, "utf8"));
   const platform = platformFromPackageName(manifest.name);
-  if (platform === null || platform.os === "win32") return;
+  if (platform === null) return;
 
-  const executablePaths = requiredExecutablePaths(dir);
-  inspectExecutablePublishConfig(manifest, executablePaths);
+  const executablePaths = requiredExecutablePaths(dir, platform);
+  if (platform.os !== "win32") {
+    inspectExecutablePublishConfig(manifest, executablePaths);
+  }
   for (const rel of executablePaths) {
     const file = path.join(dir, rel);
     if (!fs.existsSync(file)) {
@@ -75,7 +77,7 @@ function inspectPackageDir(dir) {
       continue;
     }
     const mode = fs.statSync(file).mode & 0o777;
-    if (!hasExecutableMode(mode)) {
+    if (platform.os !== "win32" && !hasExecutableMode(mode)) {
       failures.push(
         `${manifest.name}: ${rel} has mode ${formatMode(mode)}, expected executable mode`,
       );
@@ -92,16 +94,20 @@ function inspectTarball(file) {
   }
   const manifest = JSON.parse(packageJson.content.toString("utf8"));
   const platform = platformFromPackageName(manifest.name);
-  if (platform === null || platform.os === "win32") return;
+  if (platform === null) return;
 
-  for (const rel of requiredExecutablePathsFromEntries(entries, manifest)) {
+  for (const rel of requiredExecutablePathsFromEntries(
+    entries,
+    manifest,
+    platform,
+  )) {
     const name = `package/${rel}`;
     const entry = entries.get(name);
     if (!entry) {
       failures.push(`${manifest.name}: tarball missing executable ${rel}`);
       continue;
     }
-    if (!hasExecutableMode(entry.mode)) {
+    if (platform.os !== "win32" && !hasExecutableMode(entry.mode)) {
       failures.push(
         `${manifest.name}: tarball ${rel} has mode ${formatMode(entry.mode)}, expected executable mode`,
       );
@@ -135,8 +141,9 @@ function inspectExecutablePublishConfig(manifest, executablePaths) {
   }
 }
 
-function requiredExecutablePaths(packageDir) {
-  const paths = [...baseExecutablePaths];
+function requiredExecutablePaths(packageDir, platform) {
+  const paths = baseExecutablePaths(platform);
+  if (platform.os === "win32") return paths;
   const toolDir = path.join(packageDir, "bin", "go", "pkg", "tool");
   if (fs.existsSync(toolDir)) {
     for (const file of walkFiles(toolDir)) {
@@ -146,9 +153,10 @@ function requiredExecutablePaths(packageDir) {
   return paths;
 }
 
-function requiredExecutablePathsFromEntries(entries, manifest) {
+function requiredExecutablePathsFromEntries(entries, manifest, platform) {
+  if (platform.os === "win32") return baseExecutablePaths(platform);
   const configured = configuredExecutablePaths(manifest);
-  const paths = configured ?? [...baseExecutablePaths];
+  const paths = configured ?? baseExecutablePaths(platform);
   for (const rel of minimumExecutablePaths) {
     if (!paths.includes(rel)) paths.push(rel);
   }
@@ -163,6 +171,11 @@ function requiredExecutablePathsFromEntries(entries, manifest) {
     }
   }
   return paths;
+}
+
+function baseExecutablePaths(platform) {
+  const suffix = platform.os === "win32" ? ".exe" : "";
+  return posixBaseExecutablePaths.map((rel) => `${rel}${suffix}`);
 }
 
 function configuredExecutablePaths(manifest) {
--- a/tests/test-ttsc/src/features/platform/test_platform_package_windows_contents_require_base_executables.ts
+++ b/tests/test-ttsc/src/features/platform/test_platform_package_windows_contents_require_base_executables.ts
@@ -1,0 +1,224 @@
+import zlib from "node:zlib";
+
+import {
+  assert,
+  child_process,
+  fs,
+  path,
+  workspaceRoot,
+} from "../../internal/toolchain";
+
+const windowsBaseExecutables = [
+  "bin/ttsc.exe",
+  "bin/ttscserver.exe",
+  "bin/ttscgraph.exe",
+  "bin/go/bin/go.exe",
+  "bin/go/bin/gofmt.exe",
+];
+
+/**
+ * Verifies Windows platform package contents require the base executables.
+ *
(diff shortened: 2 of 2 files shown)


samchon:
Self-Review round 1 found three coverage/contract gaps before merge: (1) the empty Windows source/tarball assertions proved only the first missing path rather than the complete five-path inventory, (2) win32-arm64 exercised only directory mode, and (3) directory inspection still read file mode on Windows even though mode is deliberately out of scope. I will assert every required missing path, add the arm64 tarball twin, and isolate `stat`/mode checking to POSIX before restarting a complete review round.

PR #882, opened 2026-07-20
fix(graph): validate launcher arguments and handshake version

### LINKED ISSUE #800: fix(graph): reject malformed `@ttsc/graph` launcher arguments instead of graphing the wrong thing
## Problem

The JavaScript launcher in `@ttsc/graph` accepts malformed arguments and keeps going. Two hand-written parsers own the whole JavaScript-side CLI surface and neither validates anything:

- `parseViewArgs` in `packages/graph/src/view.ts` converts `--port` and `--max-nodes` with a bare `Number(...)`. There is no finite/integer/range check, no rejection of a missing value, and no rejection of an unknown option.
- `parseProjectArgs` in `packages/graph/src/index.ts` (used by the default MCP server lane and by `dump` for binary anchoring) ignores every unrecognized token and silently keeps the default when `--cwd` or `--tsconfig` is the last argument.

Expected behavior is that a launcher rejects an argument it cannot interpret, before it spends the cost of building a graph and before it answers questions about a project the user did not name. Current behavior is that `ttsc-graph view --max-nodes=oops` removes the viewer's documented node budget and serves the whole graph, and that `ttsc-graph --cxd <project>` or a dangling `ttsc-graph --cwd` starts an MCP server against the process working directory. An agent then receives graph answers about the wrong project with no diagnostic anywhere.

Affected users are everyone who runs the `@ttsc/graph` CLI: `ttsc-graph view` users, and any MCP client configuration whose argument list contains a typo.

### The viewer node budget disappears for a malformed value

Lead probe, driving the real `ttscgraph` binary for the dump and the built `packages/graph/lib/view.js` for the launcher, over a fixture graph with 6 raw nodes:

- `--max-nodes 1` serves 0 of the 6 raw nodes.
- `--max-nodes oops` serves 5 of the 6 raw nodes and still starts the viewer.

The mechanism is visible in `reduce` (`packages/graph/src/reduce.ts`): the cap is applied only inside `if (kept.length > maxNodes)`. `Number("oops")` is `NaN`, and every comparison with `NaN` is false, so the branch never runs and the complete graph is serialized into the payload handed to the browser. With `1` the cap keeps a single node whose edges are then dropped, and the final `finalDegree > 0` filter removes it, which is why a valid extreme value bounds the output while the malformed one does not. `Infinity` disables the cap the same way; a negative value falls into JavaScript's negative `slice` semantics and keeps all but a short suffix.

An independent reviewer recorded the exact runtime semantics behind the three cases:

```text
> node -e "const http=require('node:http'); for(const value of ['oops','-1','Infinity']) { const n=Number(value); const kept=[0,1,2]; const capped=kept.length>n?kept.slice(0,n):kept; let listen; try { const s=http.createServer(); s.listen(n); s.close(); listen='accepted'; } catch(e) { listen=e.code+': '+e.message; } console.log(value, 'Number=', String(n), 'reduce=', JSON.stringify(capped), 'listen=', listen); }"
oops Number= NaN reduce= [0,1,2] listen= ERR_SOCKET_BAD_PORT: options.port should be >= 0 and < 65536. Received type number (NaN).
-1 Number= -1 reduce= [0,1] listen= ERR_SOCKET_BAD_PORT: options.port should be >= 0 and < 65536. Received type number (-1).
Infinity Number= Infinity reduce= [0,1,2] listen= ERR_SOCKET_BAD_PORT: options.port should be >= 0 and < 65536. Received type number (Infinity).
```

The port half is not a silent failure but a late one: `runView` resolves the binary, spawns `ttscgraph dump` with a 1 GiB output buffer, parses and reduces the dump, and reads the bundled viewer assets before `server.listen` throws `ERR_SOCKET_BAD_PORT`. The whole graph build is paid for and discarded.

### The default MCP lane accepts a typo and graphs the process directory

Lead probe, replacing `startServer` with a recorder and calling the built `runGraph` three times:

```js
runGraph(["--cwd", project]);
runGraph(["--cxd", project]);
runGraph(["--cwd"]);
```

The first call selects the fixture project. The second and third both select the launcher's own process working directory: `--cxd` is unknown so the loop skips it, the bare `project` token is likewise skipped because the parser has no positional handling, and `--cwd` with nothing after it fails the `i + 1 < argv.length` guard and leaves the default in place. Both then resolve through `path.resolve(cwd)` and start a server that answers with facts about the wrong project.

### Root cause and ownership

The root cause is that the JavaScript launcher has no argument contract. `parseViewArgs` and `parseProjectArgs` are two independent permissive loops whose only failure mode is to fall back to a default. `parseViewArgs`, the numeric conversions, and the node cap arrived together in `bd0fba6dd` (`feat(graph): add ttsc-graph view to open a project's graph in 3D`); `parseProjectArgs` arrived with the launcher and MCP work in `7c5690163` and was reorganized by `62e9fce78`.

The native side of the same product does this correctly, which shows the intended standard: `packages/ttsc/cmd/ttscgraph/dump.go` and `packages/ttsc/cmd/ttscgraph/serve.go` both build a `flag.NewFlagSet(..., flag.ContinueOnError)` and fail on a bad flag. Only the JavaScript launcher is permissive.

`website/src/content/docs/graph/viewer.mdx` documents `--cwd`, `--tsconfig`, `--port`, `--max-nodes`, and `--no-open` for `view`, and states the budget the same number implements: "A project larger than the node budget (1,200 nodes for an uploaded dump) keeps its highest-degree nodes, so the hubs stay and the leaves go." `reduce`'s default `maxNodes` is that same 1,200, and it is what `--max-nodes` overrides. A value the parser cannot interpret currently removes the protection instead of being rejected.

The only launcher CLI coverage in the repository is inside `tests/test-graph/src/features/test_ttscgraph_dump_resolves_binary_from_target_cwd.ts`, which runs `view` with valid `--cwd`, `--no-open`, and `--port 0` to prove binary anchoring. No test passes a malformed, missing, or unknown option to either parser.

Verified against the audited commit `2b724664eee71c57cd688c55122cac137774bf62`; `packages/graph/src` is byte-identical on current `master` (`612abc2d2`), where only `packages/graph/package.json` differs by a version bump.

## Consequence surface

| Invocation | Current behavior | Required behavior |
| --- | --- | --- |
| `view --max-nodes=oops`, `--max-nodes=Infinity`, `--max-nodes` with no value | Cap silently disabled; the full graph is serialized to the browser | Usage error before the dump is spawned |
| `view --max-nodes=-1` | Negative `slice` keeps all but a short suffix | Usage error |
| `view --max-nodes=0`, fractional values, values above the safe-integer range | Accepted as-is | Documented positive-integer range enforced |
| `view --port=oops`, `--port=-1`, `--port` with no value | Full graph build, then an uncaught `ERR_SOCKET_BAD_PORT` | Usage error before the dump is spawned; `--port 0` stays valid and ephemeral |
| `view --cwd=` (empty value) | Empty string resolved against the process directory | Usage error |
| `view --unknown-flag` | Ignored | Usage error |
| MCP default lane `--cxd <project>` | Server starts against the process directory | Usage error |
| MCP default lane trailing `--cwd` or `--tsconfig` | Default silently kept | Usage error |
| MCP default lane `--unknown-flag` | Ignored | Usage error |
| `dump` subcommand | argv is forwarded to the native binary, which parses strictly and fails; `parseProjectArgs` is used only to anchor binary resolution, so a typo there degrades anchoring before the native error | Anchoring must not be derived from an argument list the launcher would reject |
| Valid invocations of every flag, including `--port 0`, `--no-open`, `-p`, and both separated and `=` forms | Work today | Unchanged |

The defect is platform-independent JavaScript. Its worst form is not the viewer but the MCP lane, where a wrong project produces confidently wrong architecture answers rather than a visibly empty viewer.

## Approach

The invariant: the JavaScript launcher must refuse any argument list it cannot fully interpret, and must refuse it before resolving a binary, spawning a graph build, or starting a server. A default is a value the user did not supply, never a substitute for a value the user supplied incorrectly.

The architectural owner is the `@ttsc/graph` launcher boundary: `packages/graph/src/index.ts` and `packages/graph/src/view.ts`, which should share one parser and validator rather than keeping two permissive loops. Numeric options need an explicitly documented accepted range; string options need rejection of an empty `=` value; unknown options and missing values need a usage message and a nonzero exit.

The native Go flag parsing in `packages/ttsc/cmd/ttscgraph` is out of scope and must not be changed: it is already strict, and `dump` continues to forward its argv to it.

| Case | Expected |
| --- | --- |
| `view --max-nodes 1200` and every valid form (`--max-nodes=1200`, separated) | Cap applied exactly at the boundary; default stays 1,200 |
| `view --max-nodes oops` / `Infinity` / `-1` / `0` / `1.5` / missing value | Usage error, nonzero exit, and no `ttscgraph` process spawned |
| `view --port 0` | Still binds an ephemeral port |
| `view --port oops` / `-1` / `70000` / missing value | Usage error before the dump |
| `view --cwd=` | Usage error |
| `view --nope` | Usage error naming the unknown option |
| MCP lane `--cwd <dir>`, `--cwd=<dir>`, `--tsconfig <file>` | Selected project unchanged from today |
| MCP lane `--cxd <dir>`, trailing `--cwd`, `--nope` | Usage error; no server started |
| `dump` with valid flags | Unchanged pass-through, native exit code preserved |
| Malformed input in any lane | Asserted that no graph dump is invoked, so the failure is cheap |

Narrow proving command: `pnpm --filter @ttsc/test-graph start`, extended with launcher argument cases that assert both the exit status and that no dump was spawned. Broader: `pnpm test`.

## Coordination

- Closed #684 and merged PR #698 own valid-`--cwd` binary anchoring in every Graph lane. They assume an already-parsed valid cwd. Do not re-implement or regress that anchoring; this issue adds the validation that runs before it.
- The native `ttscgraph serve` protocol provenance is owned by #723 and is unrelated.
- A separate issue restores the `@ttsc/graph` version in the MCP initialize handshake and also edits `packages/graph/src/index.ts` (the `VERSION` constant and the `startServer` call next to `parseProjectArgs`). The two changes must be sequenced by one owner, not landed in parallel.
- `packages/graph/src/structures/ITtscGraphApplication.ts`, every detailed or public subordinate type under `packages/graph/src/structures/`, and all of their JSDoc must remain byte-for-byte unchanged. This fix does not touch them.

### LINKED ISSUE #801: fix(graph): report the installed `@ttsc/graph` version in the MCP initialize handshake
## Problem

Every MCP client that connects to `ttsc-graph` receives `{"name":"ttsc-graph","version":"1.0.0"}` from the `initialize` response, for every release. The installed package version is read, threaded through two functions, and then discarded one line before it would be used.

Three places in the package state the opposite:

- `packages/graph/src/index.ts` reads the version with the comment "The server version reported in the MCP handshake; read from this package."
- `packages/graph/src/server/startServer.ts` declares its option as `/** Server version reported in the MCP handshake. */`.
- `packages/graph/src/server/createServer.ts` accepts `version: string` and executes `void version;`.

Clients surface `serverInfo` in their server lists, logs, and diagnostics. A user bug report therefore cannot say which `@ttsc/graph` produced an answer, and no version-sensitive client check can distinguish an upgrade from a downgrade. Nothing crashes and no graph fact is wrong, so the failure is silent provenance loss.

Affected: the default stdio MCP lane launched by `@ttsc/graph`, plus every programmatic caller of `createServer` or `startServer`, on every platform. Unaffected: `loadGraph`, `TtscGraphSession`, the one-shot native `ttscgraph dump`, and the native `ttscgraph serve` protocol.

## Evidence

Lead probe against the built product module, supplying an explicit version:

```powershell
node -e "const {createServer}=require('./packages/graph/lib/server/createServer.js'); const s=createServer(async()=>{},'0.19.3'); console.log(s.server._serverInfo ?? s.server);"
```

```text
{ name: 'ttsc-graph', version: '1.0.0' }
```

An independent reviewer reproduced the same through `createMcpServer` against a controller of the same shape, alongside the installed package version:

```text
serverInfo: {"name":"ttsc-graph","version":"1.0.0"}
@ttsc/graph installed version: 0.19.4
```

`serverInfo` is exactly what `@modelcontextprotocol/sdk`'s `Server` returns from `initialize`, so this observes the handshake source and not a separate display field.

### Verified upstream constraint

The installed `@typia/mcp@13.1.0` derives a version only for an HTTP controller and otherwise falls back to the literal:

```ts
  const version: string =
    (controller.protocol === "http"
      ? controller.application.version
      : undefined) ?? "1.0.0";
  const server: McpServer = new McpServer(
    { name: controller.name, version },
```

`createServer` builds its controller with `typia.llm.controller<ITtscGraphApplication>("ttsc-graph", new TtscGraphApplication(graph))`, which is the class-controller path, so `version` is unconditionally `"1.0.0"`. `IMcpServerOptions` exposes only `textFallback`, so there is no option to pass through today. This constraint is verified from the installed dependency, not assumed.

### Regression provenance

The behavior was bisected to two commits:

1. `a8bc16678` (PR #317, "register the MCP server via `@typia/mcp`"). Its pre-image constructed the server locally and passed the value:

   ```text
   $ git show a8bc16678^:packages/graph/src/server/createServer.ts | grep -n version
   28:  version: string,
   41:    { name: "ttsc-graph", version },
   ```

   The refactor replaced that construction with `createMcpServer`, which does not accept a version, and left the parameter, both JSDoc claims, and the `package.json` read in place.

2. `331ee27ec` (PR #398) added the bare `void version;`. That line did not restore the behavior or remove the dead path; it converted a visible unused-parameter error into a silent one, which is how the loss survived.

`git log -- packages/graph/src/server/createServer.ts` shows no later change. Verified against the audited commit `2b724664eee71c57cd688c55122cac137774bf62`; `packages/graph/src` is byte-identical on current `master` (`612abc2d2`), where only `packages/graph/package.json` differs by a version bump.

## Consequence surface

| Surface | Effect |
| --- | --- |
| MCP `initialize` response | `serverInfo.version` is `"1.0.0"` for every release, in every client |
| Client server lists, logs, diagnostics, bug reports | Cannot identify the running `@ttsc/graph` |
| Version-sensitive client caches or compatibility checks | Cannot detect an upgrade or a downgrade |
| `startServer`'s `version` option | Exists solely to feed the discarded parameter |
| `index.ts`'s `VERSION` constant and its `require("../package.json")` | Exist solely to feed that option |
| Handshake `instructions`, tool list, tool schemas, structured-only results | Unaffected and must stay unaffected |
| Graph answers | Unaffected; this is provenance, not correctness of facts |

The only version string that would be accidentally correct is `1.0.0` itself, if the package ever reached it.

## Approach

The invariant: the handshake must identify the `@ttsc/graph` release actually running, or the package must stop claiming that it does. The current half-state, a value threaded and then discarded while three comments assert it is reported, is itself the defect.

The architectural owner is `packages/graph/src/server/createServer.ts`, with the JSDoc in `packages/graph/src/index.ts` and `packages/graph/src/server/startServer.ts` corrected in the same change.

Because `@typia/mcp@13.1.0` offers no version option for class controllers, restoring the value requires either an upstream option in `@typia/mcp` (same owner, so a live possibility) or setting `serverInfo` on the returned `McpServer`'s inner `Server`. Do not treat either as prescribed; the issue fixes the contract, and the implementer chooses the mechanism after checking the current dependency version.

Whichever direction is chosen, the boundary is the whole dead-version path. Deleting only `void version;` while keeping the parameter, or deleting the parameter while keeping the comment in `index.ts`, leaves half of the false contract standing. The structured-only result behavior that `@typia/mcp` provides must be preserved: the hand-written registration was removed because it serialized the same payload twice, and that must not come back.

| Case | Expected |
| --- | --- |
| `initialize` against a real `ttsc-graph` stdio server | `serverInfo.version` equals `@ttsc/graph`'s `package.json` version; or, if the claim is withdrawn, `createServer` and `startServer` take no `version` and `index.ts` no longer reads `package.json` for one |
| Two different supplied versions | Two different handshake versions |
| `serverInfo.name` | `"ttsc-graph"`, unchanged |
| Handshake `instructions` | Still the `ITtscGraphApplication` class JSDoc, unchanged |
| Tool list and tool input/output schemas | Byte-identical |
| Tool results | Still structured-only, with no duplicated text payload |
| `void version;` | Gone in either direction |
| JSDoc in `index.ts` and `startServer.ts` | Matches the chosen behavior |
| Regression shield | A test asserting the handshake `serverInfo`; without one the next dependency bump reverts this silently, which is exactly how it regressed |

Narrow proving command: `pnpm --filter @ttsc/test-graph start`, which already spawns the real server over MCP. Broader: `pnpm test`.

## Coordination

- Closed #723 covers versioning and provenance for the separate native `ttscgraph serve` JSON-line protocol. It is adjacent evidence that this project values version provenance, not a duplicate; it does not touch the TypeScript MCP initialize response.
- #782 is the `ttscgraph` release train and npm manifest completeness. Its "version" is the published package version, not `serverInfo`.
- A separate issue makes the JavaScript launcher's argument parsing strict and also edits `packages/graph/src/index.ts`, next to the `VERSION` constant and the `startServer` call. Sequence the two under one owner rather than landing them in parallel.
- `packages/graph/src/structures/ITtscGraphApplication.ts`, every detailed or public subordinate type under `packages/graph/src/structures/`, and all of their JSDoc must remain byte-for-byte unchanged. The handshake `version` field is `McpServer` `serverInfo`, outside the reflected application surface, so a complete fix touches only `createServer.ts`, `startServer.ts`, and `index.ts`.

samchon:
## Intent

Fix the shared `@ttsc/graph` JavaScript launcher boundary for #800 and #801.

## Scope

- reject malformed, missing, empty, out-of-range, and unknown launcher options
  before binary resolution, graph construction, or MCP startup;
- retain valid `view`, default MCP, and native `dump` behavior, including
  project-root binary anchoring;
- report the installed graph package version from the real MCP `initialize`
  handshake without changing its instructions, tools, schemas, or
  structured-only tool-result behavior;
- add real-launcher and real-MCP regression coverage.

## Deferred

The native `ttscgraph` flag parser and the public graph structures are outside
this batch. `packages/graph/src/structures/**` remains byte-for-byte unchanged.

## Verification

Pending. This is an implementation-free campaign reservation. Every workflow
caused by this branch is cancelled only by exact commit SHA; package-scoped
local verification and solo Self-Review are the merge gates.

Closes #800
Closes #801


samchon:
Self-Review round 1 found the launcher regression matrix did not explicitly pin `--port=Infinity`, a `--max-nodes` value above the safe-integer range, and empty `--tsconfig` forms. I will add those no-native-work boundaries, rerun the package lane, and restart the complete solo review on the new snapshot.

samchon:
Self-Review round 2 found one compatibility hole: the new dump pre-parser would reject the native command's built-in help spellings before pass-through. I will recognize those native help flags without changing the Go parser, add a forwarded-argv regression case, rerun the package lane, and restart the complete solo review on the next snapshot.

samchon:
Final solo Self-Review on 8abf0c0c272f2b223d540f3e874102eb8d71bd3: clean. I re-read the complete merge-base diff and traced the iew, default-MCP, and native-dump paths through their consumers, the native flag contract, and the installed MCP SDK initialize path. The prior two findings are resolved: numeric/empty/missing boundary coverage is present, and native dump help spellings continue to pass through. ITtscGraphApplication.ts SHA-256 remains 6AB7451FCD17E8A8C1F0F7BAEAA6A96AA4E6DEDCB379F70F030ECEE64197BE6C; the src/structures tree remains 3c7399e3ce9f2404e64d652c7f1259ebd0db2613.\n\nVerification: pnpm --filter @ttsc/graph build; targeted launcher regression; pnpm --filter @ttsc/test-graph start (Success, 79.203s); target-file Prettier check; and the built Windows launcher preserves dump --help native usage/exit 2. No unrelated master changes overlap this graph surface.

PR #883, opened 2026-07-20
fix(graph): verify source display against reproducible bytes

### LINKED ISSUE #833: fix(graph): verify Graph source display against a digest the consumer can actually reproduce

samchon:
## Intent

Fix #833 by proving the raw source bytes against `diskDigest`, then decoding them with the TypeScript-Go source-file rules before proving the resulting text against `checkerDigest`.

## Scope

- preserve graph display facts for UTF-8 BOM and UTF-16 source files whose disk bytes and decoded checker text both match the snapshot;
- retain fail-closed behavior for edited files, source-preamble divergence, virtual or vanished sources, and snapshots without `diskDigests`;
- add source-reader and end-to-end graph regression coverage, plus the matching graph guide correction.

## Deferred

#812 owns ECMAScript line-terminator splitting after this encoding and provenance root is merged. #810 remains downstream of that work.

## Verification

Pending. This is an implementation-free campaign reservation. Campaign Actions for this branch are deliberately cancelled by exact commit SHA; local graph verification and solo Self-Review are the merge gates.

Closes #833.

### DIFF
--- a/packages/graph/src/model/TtscGraphSourceReader.ts
+++ b/packages/graph/src/model/TtscGraphSourceReader.ts
@@ -11,13 +11,17 @@ type ReadFile = (file: string) => Buffer;
  *
  * Signatures and declaration docs are display facts derived from source text,
  * but the live disk is not the snapshot the checker resolved. A file becomes
- * readable here only after its current bytes hash to the snapshot's
- * `checkerDigest`. Both success and failure are cached, so every consumer of a
- * `TtscGraphMemory` sees one adjudication and one immutable line array.
+ * readable here only after its current bytes hash to `diskDigest` and the
+ * compiler-decoded text hashes to `checkerDigest`. Both success and failure are
+ * cached, so every consumer of a `TtscGraphMemory` sees one adjudication and
+ * one immutable line array.
  */
 export class TtscGraphSourceReader {
   private readonly project: string;
-  private readonly checkerDigests: ReadonlyMap<string, string>;
+  private readonly digests: ReadonlyMap<
+    string,
+    { checker: string; disk: string }
+  >;
   private readonly read: ReadFile;
   private readonly cache = new Map<string, readonly string[] | undefined>();
 
@@ -30,40 +34,51 @@ export class TtscGraphSourceReader {
   ) {
     this.project = project;
     this.read = read;
-    this.checkerDigests = new Map(
-      provenance?.capabilities.includes("sourceDigests") === true
+    this.digests = new Map(
+      provenance?.capabilities.includes("sourceDigests") === true &&
+        provenance.capabilities.includes("diskDigests") === true
         ? provenance.sources.map((source) => [
             normalize(source.file),
-            source.checkerDigest,
+            { checker: source.checkerDigest, disk: source.diskDigest },
           ])
         : [],
     );
   }
 
   /**
-   * Return frozen lines only when the bytes still equal the checker snapshot. A
-   * missing manifest entry, read failure, or digest mismatch is a cached
-   * absence rather than a reason to mix current disk text into old graph
-   * facts.
+   * Return frozen lines only when raw disk bytes and compiler-decoded text both
+   * equal the checker snapshot. A missing manifest entry, read failure, or
+   * digest mismatch is a cached absence rather than a reason to mix current
+   * disk text into old graph facts.
    */
   lines(file: string): readonly string[] | undefined {
     const key = normalize(file);
     if (this.cache.has(key)) return this.cache.get(key);
 
-    const expected = this.checkerDigests.get(key);
-    if (expected === undefined || expected === "") {
+    const expected = this.digests.get(key);
+    if (
+      expected === undefined ||
+      expected.checker === "" ||
+      expected.disk === ""
+    ) {
       this.cache.set(key, undefined);
       return undefined;
     }
 
-    let text: string;
+    let bytes: Buffer;
     try {
-      text = this.read(path.resolve(this.project, file)).toString("utf8");
+      bytes = this.read(path.resolve(this.project, file));
     } catch {
       this.cache.set(key, undefined);
       return undefined;
     }
-    if (sha256(text) !== expected) {
+    if (sha256(bytes) !== expected.disk) {
+      this.cache.set(key, undefined);
+      return undefined;
+    }
+
+    const text = decodeSource(bytes);
+    if (sha256(text) !== expected.checker) {
       this.cache.set(key, undefined);
       return undefined;
     }
@@ -78,6 +93,23 @@ function normalize(file: string): string {
   return file.replace(/\\/g, "/");
 }
 
-function sha256(text: string): string {
-  return createHash("sha256").update(text, "utf8").digest("hex");
+function decodeSource(bytes: Buffer): string {
+  if (bytes.length >= 2) {
+    const body = bytes.subarray(2, bytes.length - ((bytes.length - 2) % 2));
+    if (bytes[0] === 0xff && bytes[1] === 0xfe) return body.toString("utf16le");
+    if (bytes[0] === 0xfe && bytes[1] === 0xff)
+      return Buffer.from(body).swap16().toString("utf16le");
+  }
+  if (
+    bytes.length >= 3 &&
+    bytes[0] === 0xef &&
+    bytes[1] === 0xbb &&
+    bytes[2] === 0xbf
+  )
+    return bytes.subarray(3).toString("utf8");
+  return bytes.toString("utf8");
+}
+
+function sha256(value: string | Buffer): string {
+  return createHash("sha256").update(value).digest("hex");
 }
--- a/tests/test-graph/src/features/test_ttscgraph_details_reads_bom_and_utf16_source_snapshot.ts
+++ b/tests/test-graph/src/features/test_ttscgraph_details_reads_bom_and_utf16_source_snapshot.ts
@@ -120,6 +120,7 @@ export const test_ttscgraph_details_reads_bom_and_utf16_source_snapshot =
       }
     } finally {
       client.endStdin();
-      await client.waitForExit();
+      const code = await client.waitForExit();
+      assert.equal(code, 0, client.stderrText());
     }
   };
--- a/tests/test-graph/src/features/test_ttscgraph_details_reads_bom_and_utf16_source_snapshot.ts
+++ b/tests/test-graph/src/features/test_ttscgraph_details_reads_bom_and_utf16_source_snapshot.ts
@@ -1,0 +1,125 @@
+import { TestProject } from "@ttsc/testing";
+import fs from "node:fs";
+import path from "node:path";
+
+import { TtsgraphClient, assert } from "../internal/ttsgraph";
+
+interface ToolResult {
+  structuredContent?: unknown;
+}
+
+interface DetailsResult {
+  type: "details";
+  nodes: { name: string; signature?: string; doc?: string }[];
+}
+
+const graphArguments = (handles: string[]) => ({
+  question: "Inspect declaration signatures from the current source snapshot.",
+  draft: {
+    reason: "The named declarations need one precise graph details request.",
+    type: "details",
+  },
+  review:
+    "Confirmed: the graph details answer is the needed source-derived fact.",
+  request: { type: "details", handles },
+});
+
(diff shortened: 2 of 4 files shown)


samchon:
Solo Self-Review round 1 found one test-completeness improvement before acceptance: the new resident graph encoding regression awaited server exit but did not assert a zero exit code. I will make a post-response launcher failure fail the case with captured stderr, then push that focused test correction and restart a complete review round.

samchon:
Fresh solo review after rebasing onto `master` is clean.

Verified the complete PR diff and the adjacent graph consumers against the issue contract:
- raw filesystem bytes must match the graph `diskDigest`;
- TypeScript-Go-compatible BOM/UTF-16 decoding must then match `checkerDigest`;
- absent provenance, virtual files, edits, and preamble mismatches fail closed;
- line splitting remains deliberately unchanged for the dependent #812 work.

Local evidence on this exact rebased head:
- `pnpm --filter @ttsc/test-graph start` — passed (including the BOM, UTF-16LE, UTF-16BE end-to-end regression and the source-reader identity suite);
- protected graph application hash and structures-tree SHA both match their campaign guards;
- `git diff --check` and the working tree are clean.

Campaign CI for this implementation SHA is intentionally cancelled after exact-SHA capture.

samchon:
Fresh solo review is clean after rebasing this PR onto `master` `494a36cf` (#882).

I re-reviewed the full four-file diff and the `TtscGraphMemory` consumer. The implementation preserves the issue boundary: it proves raw bytes with `diskDigest`, proves compiler-decoded BOM/UTF-16 text with `checkerDigest`, and continues to fail closed for unavailable provenance, virtual files, edits, and preamble divergence. Line splitting remains deferred to dependent #812.

Local acceptance on this exact head:
- rebuilt `@ttsc/graph` successfully;
- `pnpm --filter @ttsc/test-graph start` passed in 113,670 ms, including the #882 launcher regressions and the new BOM/UTF-16 end-to-end regression;
- `git diff --check`, working-tree cleanliness, and both protected graph guards passed.

The PR is draft again until this current-base review record and exact-SHA cancellation readback are complete.

PR #884, opened 2026-07-20
fix(playground): restore loader contracts

### LINKED ISSUE #802: fix(playground): fence compiler-client connection generations across overlapping reset and connect
## Problem

`createCompilerClient` in `packages/playground/src/react/createCompilerClient.ts` is the public `@ttsc/playground` entry point that boots the compiler Worker and hands callers its typed `ICompilerService` driver. It owns three mutable slots, `connectionPromise`, `pendingConnector`, and `activeConnector`, and nothing marks which connection attempt owns them.

`reset()` clears all three slots and then awaits the connector that was in flight when it started. That await is a suspension point, so a caller may begin a new `connect()` before `reset()` finishes. When the new connector settles first and the old one settles afterwards, the old asynchronous continuation still runs `activeConnector = connector` unconditionally, overwriting the live connector with the one being torn down. The rejection path is symmetric: a stale `connect()` failure runs `connectionPromise = null` with no ownership check, discarding the cached promise of a newer, already-returned connection.

The results are a live Worker that no `reset()` can ever reach, a torn-down connector that gets closed twice, and, in the rejection ordering, a third connector allocated while the second is still running. The expected behavior is that exactly one connection generation is current at a time, that `reset()` closes exactly the generation it invalidated exactly once, and that every allocated connector stays reachable until it is closed.

Affected: every `@ttsc/playground` consumer that can overlap teardown with a retry, which includes the hosted ttsc.dev playground and any embedder that fires `reset()` without awaiting it, races a retry against cleanup, or drives the client from competing UI lifecycle transitions. The leaked resource is a browser Worker holding a wasm compiler and its runtime memory.

### Reproduction

The probe loads the built `createCompilerClient` and replaces only the `tgrid` Worker transport with connectors whose settlement is controlled by the test, so the interleaving is deterministic rather than scheduler-dependent:

```js
const Module = require("node:module");
const originalLoad = Module._load;
const gates = [], instances = [];
class FakeConnector {
  constructor() { this.id = instances.length + 1; this.closeCount = 0; instances.push(this); }
  connect() { return new Promise((resolve, reject) => gates.push({ resolve, reject })); }
  getDriver() { return { connectorId: this.id }; }
  close() { this.closeCount++; return Promise.resolve(); }
}
Module._load = function (request, parent, isMain) {
  if (request === "tgrid") return { WorkerConnector: FakeConnector };
  return originalLoad.call(this, request, parent, isMain);
};
const { createCompilerClient } = require("packages/playground/lib/src/react/createCompilerClient.js");

// Ordering 1: connect A, reset A, connect B; B settles first, A settles second.
const client = createCompilerClient({ workerUrl: "worker.js" });
const first = client.connect();
const firstReset = client.reset();
const second = client.connect();
gates[1].resolve();
const secondDriver = await second;
gates[0].resolve();
await Promise.all([first, firstReset]);
await client.reset();
console.log("order-1", JSON.stringify({ secondDriver, closeCounts: instances.map((x) => x.closeCount) }));

// Ordering 2: same overlap, but A rejects after B has already been returned.
const c2 = createCompilerClient({ workerUrl: "worker.js" });
const a = c2.connect().catch((e) => `A rejected: ${e.message}`);
const r = c2.reset();
const b = c2.connect();
gates[1].resolve();
const bDriver = await b;
gates[0].reject(new Error("boot failed"));
console.log("order-2 A", await a);
await r;
void c2.connect().catch(() => {});
console.log("order-2", JSON.stringify({ bDriver, connectorsAllocatedAfterThirdConnect: instances.length, closeCounts: instances.map((x) => x.closeCount) }));
```

Observed:

```text
order-1 {"secondDriver":{"connectorId":2},"closeCounts":[2,0]}
order-2 A A rejected: boot failed
order-2 {"bDriver":{"connectorId":2},"connectorsAllocatedAfterThirdConnect":3,"closeCounts":[0,0,0]}
```

Ordering 1: connector 2 is the driver the caller holds and is never closed, while connector 1 is closed twice. Ordering 2: connector 2 is live and returned to the caller, yet the stale rejection from connector 1 cleared the cached promise, so the very next `connect()` allocates connector 3 against the same client.

### Root cause

Both settlement paths publish themselves as current with no generation check. `connect()`:

```ts
      connectionPromise = (async () => {
        try {
          await connectPromise;
        } catch (err) {
          connectionPromise = null;
          if (pendingConnector === connectPromise) pendingConnector = null;
          throw err;
        }
        if (pendingConnector === connectPromise) pendingConnector = null;
        activeConnector = connector;
        return connector.getDriver() as unknown as ICompilerService;
      })();
```

The `pendingConnector === connectPromise` test guards only the slot clear, never `activeConnector = connector` and never `connectionPromise = null`. `reset()` then snapshots and clears before suspending:

```ts
    async reset(): Promise<void> {
      const inflight = pendingConnector;
      const active = activeConnector;
      connectionPromise = null;
      pendingConnector = null;
      activeConnector = null;
      if (inflight) {
        try {
          const c = await inflight;
          if (c !== active) await c.close().catch(() => {});
        } catch { }
      }
      if (active) {
        await active.close().catch(() => {});
      }
    },
```

In ordering 1, `active` is `null` at reset time, so the `inflight` branch closes connector 1 once; the late continuation of connector 1 then reinstalls itself as `activeConnector`, and the final `reset()` closes it a second time while connector 2 is unreachable.

### Provenance, and why this recurs

The failure class already has two repairs, both committed directly with no issue and no lifecycle test matrix.

| Commit | What it fixed | What it left unowned |
| --- | --- | --- |
| `afd6a3b54` "code-review round 3" | Made `reset` asynchronous and tracked the resolved connector, fixing a Worker leak | Only the resolved slot; nothing about who owns it |
| `410faa5e2` "code-review round 4" | Added `pendingConnector` because a reset during an in-flight connect still leaked a Worker | Only the in-flight slot; still no generation identity |

The current comment records the second repair's reasoning in the source: "disposing only the resolved `activeConnector` leaks the connector when reset() is called during an in-flight `await connector.connect(...)`". Each repair added the one slot whose absence had produced the observed leak, and each left the general rule unwritten. The interleaving reproduced above has the same consequence as both, an unreachable live Worker, and is reached through the slot that the second repair introduced. Two repairs at the same site, each shaped to the symptom that surfaced, with the root ownership question still unanswered, is what makes this whack-a-mole rather than a third missing case: a third slot-shaped patch would leave a fourth interleaving.

No open or closed issue or pull request mentions `createCompilerClient`, `WorkerConnector`, playground Worker leakage, or this reset/connect overlap.

Verified against the audited commit `2b724664eee71c57cd688c55122cac137774bf62`. `packages/playground/src` is byte-identical on current `master` (`612abc2d2`); only `packages/playground/package.json` differs, by a version bump.

## Consequence surface

| Case | Current behavior |
| --- | --- |
| `reset()` with no pending and no active connector | Correct |
| `reset()` after a fully settled connection | Correct |
| `reset()` during one pending connection, no replacement started | Correct; this is the case `410faa5e2` added |
| `connect A` → `reset A` → `connect B`, B settles first | B leaks, A closed twice (reproduced) |
| `connect A` → `reset A` → `connect B`, A settles first | Final ownership depends on continuation order; the same missing guard decides it |
| `connect A` → `reset A` → `connect B`, A rejects after B settled | B's cached promise is cleared; the next `connect()` allocates a duplicate connector (reproduced) |
| Concurrent resets | Can close the stale connector repeatedly while the live one stays unreachable |
| Concurrent `connect()` calls with no reset | Correct; must keep sharing one promise |
| Retry after a genuine boot failure | Must keep working; the cache clear is intentional there |

The race is scheduler-independent and not specific to any browser or OS. The `PlaygroundShell` retry paths in this repository normally await `reset()`, which narrows the in-repository witness, but the client is a published API and its contract is what breaks.

## Approach

The invariant: at most one connection generation is current, only the current generation may publish itself into the client's state, and every allocated connector is closed exactly once by whoever invalidated it.

The architectural owner is `createCompilerClient`. Whether that is a monotonic generation token compared before each mutation, a serialization of replacement behind cleanup, or one state object replaced atomically per attempt is an implementation choice for the fix. What the fix must not do is add a fourth slot for the interleaving above.

The repository already fenced an analogous cache lifecycle in closed #662, "fix(ttsx): fence dependency-cache generations and publish complete emits atomically", which is prior art for the shape of the answer rather than a duplicate.

The public shape of `createCompilerClient` (`connect(): Promise<ICompilerService>`, `reset(): Promise<void>`) must not change.

| Case | Expected |
| --- | --- |
| `connect A` → `reset A` → `connect B`, B settles first | The driver returned for B stays current; a later `reset()` closes B exactly once |
| `connect A` → `reset A` → `connect B`, A settles first | Same outcome; A never becomes current |
| A rejects after B has been returned | B's cached promise survives; no duplicate connector is allocated |
| A is closed | Exactly once, in every ordering above |
| Every allocated connector | Reachable until closed; `closeCounts` never exceeds 1 for any connector |
| Concurrent `connect()` calls | Still share one promise and one connector |
| Retry after a genuine boot failure | Still retries, that is, the cache clear still happens for the generation that failed |
| `reset()` before any `connect()`, and `reset()` after a settled `connect()` | Unchanged |
| A `close()` that throws | Still swallowed; teardown is not blocked |

There is no playground lifecycle test today, so the regression is entirely new; `tests/test-playground/src/features/` currently holds seven cases and none touches `createCompilerClient`.

Each ordering is a multi-step interleaving, not a single call: the test must control settlement order explicitly, as the reproduction does, because a probe that awaits each step in turn passes against the defect.

Narrow proving command, after adding the cases under `tests/test-playground/src/features/`:

```text
pnpm --filter @ttsc/test-playground start -- --include=compiler_client
```

The runner matches test names by substring and prints each executed test by name. A filter matching nothing fails loudly (`No tests matched --include=<...>`, exit 1), so a typo cannot pass silently; `compiler_client` matches no test at the baseline, so that is the expected result until the cases exist. Read the printed names anyway, because a filter that happens to match some *other* existing case would run green without ever executing what you added. Confirm each new case fails at the audited baseline.

Broader: `pnpm --filter @ttsc/test-playground start`, then `pnpm test`.

## Coordination

- File-disjoint from every other open playground defect. This issue touches `packages/playground/src/react/createCompilerClient.ts` only, while the npm installer issues (semver range satisfaction, PAX byte parsing, Node builtin classification, template-substitution discovery) live under `packages/playground/src/npm/` and the CommonJS `exports` issue lives under `packages/playground/src/sandbox/`. They can be implemented in parallel.
- Commits `afd6a3b54` and `410faa5e2` are the causal history above, not duplicates: neither has an issue, and neither introduced a lifecycle test.
- No open pull request touches `packages/playground`.

### LINKED ISSUE #803: fix(playground): install only the packages and versions the manifest declared
## Problem

`installPlaygroundDependencies` is a published `@ttsc/playground` API and the engine behind the hosted ttsc.dev playground. The guide states that it "installs the imported package, its registry dependencies, non-optional peer dependencies, and a matching `@types/*` package when the package itself does not ship declarations". Nothing in the package or the guide says the declared ranges are approximate.

They are. `selectVersion` in `packages/playground/src/npm/internal/npmRegistry.ts` never evaluates a semver range. It tries the range as an exact version key, then as a dist-tag, then strips a leading run of `~^<>=` and spaces and tries that as an exact key, and otherwise returns `dist-tags.latest`, falling back to a home-grown numeric sort. No branch tests whether the returned version satisfies what was asked. A caret, tilde, comparator set, union, hyphen range, or wildcard therefore routinely installs a major the package explicitly excludes, and when no satisfying version exists at all the installer reports success instead of a failure.

A second constraint loss sits one level up. `installPlaygroundDependencies` keys its `queued` and `done` sets by package name only, so the first range seen for a name wins and every later constraint for that name is discarded with no comparison and no diagnostic.

The third loss is the most serious, because it is not a version error at all. An npm alias range (`"aliased": "npm:real-package@^1.0.0"`) passes `isRegistryRange`, which filters `file:`, `link:`, `workspace:`, `portal:`, `git+`, `github:`, and `http(s):` but not `npm:`. The installer then queries the registry for the alias key, `aliased`, rather than the aliased target, `real-package`. No version matches the range `npm:real-package@^1.0.0` and no dist-tag does either, so `dist-tags.latest` wins. **The playground downloads and mounts a package the manifest never named, at whatever version is newest.** That is not a relaxed constraint; it is a different package, executing in the sandbox under the name the source imported.

Everything selected is downloaded, unpacked, and mounted into the compiler files, the Monaco declaration libs, and the Execute runtime files. A user then type-checks and runs against a dependency the manifest rejects, or against a package it never mentioned, and the resulting errors look like the package's fault.

### Reproduction

Against the audited source, with metadata that publishes a satisfying `1.9.9` alongside `2.0.0`:

```js
const { selectVersion } = require(".../packages/playground/lib/src/npm/internal/npmRegistry.js");
const metadata = {
  name: "dependency",
  "dist-tags": { latest: "2.0.0" },
  versions: { "1.9.9": { name: "dependency", version: "1.9.9" }, "2.0.0": { name: "dependency", version: "2.0.0" } },
};
for (const range of ["^1", ">=1 <2", "<2.0.0", "^16.8.0 || ^17.0.0", "~1.9", "1.x"])
  console.log(JSON.stringify({ range, selected: selectVersion(metadata, range) }));
```

Observed:

```text
{"range":"^1","selected":"2.0.0"}
{"range":">=1 <2","selected":"2.0.0"}
{"range":"<2.0.0","selected":"2.0.0"}
{"range":"^16.8.0 || ^17.0.0","selected":"2.0.0"}
{"range":"~1.9","selected":"2.0.0"}
{"range":"1.x","selected":"2.0.0"}
```

Every one of those ranges excludes `2.0.0`, and `1.9.9` satisfies four of them. `<2.0.0` is the sharpest: the operator-stripping step turns the range into the exact string `2.0.0`, so the resolver selects the single version the comparator was written to exclude.

Four further shapes, same source:

```text
A {"range":"^1.0.0","selected":"2.0.0"}          // versions 1.1.0, 1.9.0, 2.0.0; no exact 1.0.0 published
B {"range":"^2","selected":"1.0.0-beta.2"}       // no dist-tags; the numeric sort ranks a prerelease above its release
C {"exact":"1.9.9","tag":"3.0.0-rc.1"}           // exact versions and real dist-tags are correct today
D {"range":"^1.0.0","selected":"5.0.0"}          // only 5.0.0 published; no satisfying version exists, yet a version is returned
```

Case D is the one with no honest answer available: the correct outcome is a reported failure, and the current code reports success.

The Round-1 real-registry run recorded the same defect against live metadata on 2026-07-19: `@material-ui/core@4.12.4` declares peer `react: "^16.8.0 || ^17.0.0"`, and `selectVersion` returned React `19.2.7`.

### Repeated and aliased constraints

`enqueuePackageDependencies` forwards the manifest key as the package name and the manifest value as the range. Driving it over one manifest:

```text
[{"name":"aliased","range":"npm:real-package@^1.0.0","optional":false},
 {"name":"normal","range":"^1.0.0","optional":false},
 {"name":"peerA","range":"^16 || ^17","optional":false}]
{"aliasRangeSelects":"9.9.9"}
```

The first three lines are the queue after filtering: `aliased` survived `isRegistryRange` as an ordinary entry, carrying the alias expression as its range. The fourth line is `selectVersion` given metadata for the alias key and that range: it returns `9.9.9`, the alias key's `latest`. Both halves of the alias defect are therefore in one output, the wrong package name reaching the registry and the wrong version being chosen for it.

For repeated constraints, `installPlaygroundDependencies` records `queued` and `done` by name, so two consumers of the same package resolve to whichever range the traversal reached first, whether or not the two ranges intersect.

### Provenance

The dependency loader arrived in PR #135, which described the exact/tag/latest behavior and explicitly deferred full semver resolution. The resolver and the name-keyed queue were then carried into the reusable public package by `2ae95c4ff` ("feat(playground): new @ttsc/playground package") and are substantively unchanged at the audited commit. PR #699 hardened dependency discovery and sandbox behavior without touching version selection. Closed #705 replaced Typia's version-specific pack patches with a dependency-derived closure and does not generalize to arbitrary npm packages or own `selectVersion`.

What is now in scope that #135 deferred: #135 deferred *resolution*, and the package has since been published as a reusable API with a documented installer. This issue does not ask for a package manager. It asks that the installer mount only packages the manifest named, at versions that satisfy the ranges it declared, or say precisely that it cannot, so that the deferral stops being invisible to callers.

The only npm-installer test, `tests/test-playground/src/features/test_collect_external_package_names_ignores_non_code.ts`, covers source discovery and the first registry call. There is no version-selection or multi-constraint test. Repository and GitHub searches for `selectVersion`, playground semver, dependency ranges, and npm installer versions return no owning issue or pull request.

Verified against `2b724664eee71c57cd688c55122cac137774bf62`; `packages/playground/src` is byte-identical on current `master` (`612abc2d2`).

## Consequence surface

The supported grammar has to be stated by the fix rather than inferred. The current behavior across it:

| Range form | Example | Current | Required |
| --- | --- | --- | --- |
| Exact version | `1.9.9` | Correct | Unchanged |
| Dist-tag | `next` | Correct | Unchanged |
| Caret / tilde | `^1.0.0`, `~1.9` | Latest, any major | Highest satisfying |
| Abbreviated | `^1`, `1.x`, `*` | Latest, any major | Highest satisfying |
| Comparator set | `>=1 <2` | Latest | Highest satisfying |
| Strict upper bound | `<2.0.0` | Selects the excluded bound | Highest satisfying |
| Hyphen range | `1.0.0 - 1.9.9` | Latest | Highest satisfying |
| Union | `^16.8.0 \|\| ^17.0.0` | Latest | Highest satisfying |
| Prerelease ordering | `1.0.0` vs `1.0.0-beta.2` | Prerelease can outrank its release | npm precedence |
| No satisfying version | `^1.0.0` against only `5.0.0` | Returns `5.0.0`, reports success | Explicit failure naming package and range |
| npm alias | `npm:real@^1` | Installs `latest` of the alias key, a package the manifest never named | Resolve `real` at `^1`; rejecting the form with a precise error is an acceptable fallback |
| Non-registry protocols | `file:`, `link:`, `git+`, `https:` | Skipped | Unchanged |
| Two compatible constraints for one name | `^1.0.0` and `>=1.2.0` | First wins | A version satisfying both |
| Two disjoint constraints for one name | `^1.0.0` and `^2.0.0` | First wins silently | Reported, naming both requesters and ranges |

Constraint sources are ordinary `dependencies` and non-optional `peerDependencies`; both pass through the same selector. Optional peers stay unqueued. The affected consumers are the hosted playground and every `@ttsc/playground` embedder. The failure is browser-side and platform-independent, and registry tag movement makes identical source behave differently over time even when a satisfying older version stays published.

Everything downstream must keep working unchanged: `maxPackages`, the ignored and preinstalled sets, `AbortSignal` cancellation, the optional `@types/*` lookup that may legitimately 404, tarball download and unpack, and the compiler / Monaco / runtime mount maps.

## Approach

Two invariants, and the second one is not negotiable in the way the first one's mechanics are.

**The installer never mounts a package the manifest did not name.** For an npm alias that means resolving the aliased target: `"aliased": "npm:real-package@^1.0.0"` fetches `real-package` at `^1.0.0` and mounts it under the key `aliased`, which is what npm does. Rejecting the alias form outright, with an error naming the specifier and why it is unsupported, is an acceptable fallback if resolving it is judged out of scope. Continuing to fetch the alias key and mount whatever comes back is not an option in either direction.

**A version the installer returns satisfies the range it was asked for.** That invariant belongs to `selectVersion`: it returns a satisfying version under npm semver semantics, or the call fails with a message naming the package, the requested range, and the versions available. Silently widening a constraint to `latest` is never a valid outcome.

`installPlaygroundDependencies` owns the second half: constraints for one package name must be reconciled deliberately. The browser mount is flat, so two incompatible versions of one package cannot both be installed; the requirement is that the installer choose a version satisfying the intersection when one exists and report the conflicting requesters and ranges when it does not, instead of letting traversal order decide.

The optional-`@types` skip path must keep tolerating a 404 without turning a genuine no-satisfying-version condition for a required package into a silent success.

Whether satisfaction is implemented in-package or by adopting a dependency is an implementation choice; the browser bundle size constraint is real and belongs to the fix.

Positive: exact versions, dist-tags, caret, tilde, abbreviated, comparator sets, hyphen ranges, unions, and wildcards each select the highest published satisfying version. Prerelease versions order by npm precedence and are not selected for a range that does not admit them.

Negative: a range with no satisfying version fails, naming the package and the range, and mounts nothing. Two disjoint constraints for one name are reported rather than silently resolved.

Alias, as its own required case rather than a footnote: given `"aliased": "npm:real-package@^1.0.0"`, the registry request is for `real-package`, the selected version satisfies `^1.0.0`, and the files mount under the key `aliased`; or the install fails naming the specifier. The assertion that must exist either way is that no request is ever made for the alias key as a package name, because that is the step that reaches an unnamed package. A fake-registry case can prove this by failing the test if `aliased` is fetched at all.

Boundary: a range whose only satisfying version is the lowest published one; a range satisfied only by the `latest` tag (which must still be chosen for the right reason); metadata with no `dist-tags`; a package with a single version.

Regression: the existing discovery and first-registry-call test keeps passing; `maxPackages`, abort, ignored and preinstalled sets, optional-peer skipping, and the optional `@types` 404 path are unchanged. An end-to-end fixture matching `@material-ui/core@4.12.4` must never report React 19 as satisfying `^16.8.0 || ^17.0.0`.

The new coverage is a pure resolver table over the grammar above, a fake-registry installer case asserting which tarball was fetched, and a two-parent conflict case. None exists today: the only installer test is `tests/test-playground/src/features/test_collect_external_package_names_ignores_non_code.ts`, which asserts no versions.

Narrow proving command, after adding the cases under `tests/test-playground/src/features/`:

```text
pnpm --filter @ttsc/test-playground start -- --include=select_version
```

The runner matches test names by substring and prints each executed test by name. A filter matching nothing fails loudly (`No tests matched --include=<...>`, exit 1), so a typo cannot pass silently; `select_version` matches no test at the baseline, so that is the expected result until the cases exist. Read the printed names anyway, because a filter that happens to match some *other* existing case would run green without ever executing what you added. Confirm each new case fails at the audited baseline.

Broader: `pnpm --filter @ttsc/test-playground start`, then `pnpm test`.

## Coordination

- Closed PR #135 introduced the loader and deliberately deferred semver resolution; it is causal history, not a duplicate.
- Closed #705 owns Typia's dependency closure only and must keep working: the bundled typia packs stay pinned to the exact source version linked into the playground compiler.
- Closed #699 changed dependency discovery, not version selection.
- This issue touches `packages/playground/src/npm/internal/npmRegistry.ts` and `packages/playground/src/npm/installPlaygroundDependencies.ts`. It shares `npmRegistry.ts` with the PAX byte-domain parsing issue, so those two must be sequenced under one owner. It is file-disjoint from the compiler-client generation issue (`src/react/`), the sandbox `exports` issue (`src/sandbox/`), and the builtin-classification and template-substitution issues (`src/npm/packageNameFromSpecifier.ts`, `src/npm/collectExternalPackageNames.ts`).
- Three discovery rounds reported this independently; it is one issue, not three.

### LINKED ISSUE #820: fix(playground): resolve sandbox requires through the package manifest's exports boundary
## Problem

`createSandboxRequire` (`packages/playground/src/sandbox/createSandboxRequire.ts`) is the Execute lane's CommonJS loader, listed in the `@ttsc/playground` README's API table as the Execute sandbox. It runs the compiled JavaScript against the packages the playground installed from npm, so it is the component that decides which file of a package actually executes in the browser.

It disagrees with Node in two ways, and both change which code runs.

Encapsulation. For a package subpath, `resolveSpecifier` tries physical packed paths first and consults `package.json#exports` only if none of them exists. Once a package declares `exports`, that map is the package's public boundary and an unexported file must not be importable through a package subpath. In the sandbox it is: the file is present in the pack, so the direct-path attempt finds it before the manifest is ever read.

Condition order. `pickConditionalExport` reduces every conditional target with the fixed expression `obj.require ?? obj.default ?? obj.node`. Node selects among the conditions active for the consumer in the order the manifest declares them, with `default` acting as a condition that is always active rather than a fixed-priority slot. A manifest that lists `node` before `default` therefore resolves differently in Node and in the sandbox.

The same reducer owns the root entry, the exact-subpath entry, and the wildcard-subpath entry, so the ordering defect is not confined to one manifest shape. The consequence is a playground result that is right about compilation and wrong about execution: the program runs a branch its package intended for a different condition, or reaches a file the package deliberately made unreachable.

### Reproduction

Fixture package, one manifest and three files, with `private.cjs` present but deliberately absent from `exports`:

```json
{
  "name": "fixture",
  "exports": { "node": "./node.cjs", "default": "./default.cjs" }
}
```

Real Node, with the fixture in `node_modules`:

```text
{"root":"node","private":"threw: ERR_PACKAGE_PATH_NOT_EXPORTED"}
```

The production sandbox, given the same three files as its pack:

```js
const pack = {
  "fixture/package.json": JSON.stringify({ name: "fixture", exports: { node: "./node.cjs", default: "./default.cjs" } }),
  "fixture/node.cjs": 'module.exports = "node";',
  "fixture/default.cjs": 'module.exports = "default";',
  "fixture/private.cjs": 'module.exports = "private";',
};
const req = createSandboxRequire(pack, { console });
```

```text
{"root":"default","private":"private"}
```

Two divergences from one fixture. `require("fixture")` executes a different file than Node executes, and `require("fixture/private")` succeeds where Node raises `ERR_PACKAGE_PATH_NOT_EXPORTED`.

### Root cause

Subpath resolution reads the filesystem before the manifest:

```ts
    // First try direct paths (covers the common typia/lib/internal/X case).
    const direct = tryPaths(
      `${pkg}/${subpath}`,
      `${pkg}/${subpath}.js`,
      `${pkg}/${subpath}.cjs`,
      `${pkg}/${subpath}.mjs`,
      `${pkg}/${subpath}.json`,
      `${pkg}/${subpath}/index.js`,
      `${pkg}/${subpath}/index.cjs`,
    );
    if (direct) return direct;
    // Fall back to package.json exports map.
    return resolvePackageEntry(pkg, subpath);
```

Condition selection is a fixed expression, with the intent stated in the comment:

```ts
function pickConditionalExport(value: unknown): string | null {
  if (typeof value === "string") return value;
  if (value && typeof value === "object") {
    const obj = value as Record<string, unknown>;
    // Prefer require (CJS) > default > node.
    const pick = obj.require ?? obj.default ?? obj.node;
    if (typeof pick === "string") return pick;
    if (pick && typeof pick === "object") return pickConditionalExport(pick);
  }
  return null;
}
```

`resolvePackageEntry` calls it for the root target, for the exact `./subpath` entry, and for each `./prefix/*` wildcard target.

### This is a follow-up to #670, not a duplicate

Closed #670, "fix(playground): support valid CommonJS root package.json.exports shapes in the Execute sandbox", was implemented by PR #699 (`4078876d9`). Its verified scope was the three valid *root* representations: a bare string, a `"."` subpath table, and a bare condition map. It added `rootExportTarget`, whose JSDoc documents exactly those three shapes, and it stated that subpath and wildcard handling were untouched. It retained the fixed conditional picker, and its regression pins the fixed rule as intended behavior, with the comment "require is preferred over default: only the require target exists" and the test docstring listing "`require` is preferred over `default` in a condition map (CJS precedence)".

The evidence above covers neither of #670's concerns: encapsulation of a present-but-unexported subpath, and ordered condition selection. Searches for `PACKAGE_PATH_NOT_EXPORTED`, playground export encapsulation, conditional exports, and sandbox require exports found no other issue or pull request, open or closed.

Verified against `2b724664eee71c57cd688c55122cac137774bf62`; `packages/playground/src` is byte-identical on current `master` (`612abc2d2`).

### The path is live, not dormant

`installPlaygroundDependencies` merges downloaded package runtime files into its result, `PlaygroundShell` retains them and supplies them to Execute, and `TtscWebsitePlaygroundShell` merges them with the built-in runtime pack and passes the result to `createSandboxRequire` before evaluating the emitted CommonJS. Any installed package with an `exports` map reaches this resolver.

## Consequence surface

| Manifest shape / request | Node | Sandbox today |
| --- | --- | --- |
| Root string target | Selected | Selected |
| Root `"."` table | Selected | Selected |
| Root bare condition map, `require` first | `require` | `require` |
| Root bare condition map, `node` before `default` | `node` | `default` |
| Exact `"./sub"` entry, file present at the physical path | Manifest target | Physical path, manifest never read |
| Subpath present in the pack, absent from `exports` | `ERR_PACKAGE_PATH_NOT_EXPORTED` | Executes the file |
| Wildcard `"./prefix/*"` entry | Pattern target, longest-prefix precedence | Physical path first, then a first-match loop |
| `null` target (blocked subpath) | Throws | `pickConditionalExport` returns `null`; falls through |
| Array target (fallback list) | First usable entry | `pickConditionalExport` returns `null` |
| Nested condition object | Ordered, recursively | Fixed `require`/`default`/`node` recursion |
| Package with no `exports`, `main` present | `main`, then index fallback | Same |
| Relative `./` and `../` requires | Caller-relative | Same, and must stay unchanged |

The Node column is the reference *algorithm*, not the required answer in every row. Node runs with `node` active and the sandbox does not have to; the rows that turn on which conditions are active are settled by the declared set, and only the rows that turn on the algorithm itself (manifest-before-filesystem, declaration order, `null`, arrays, pattern precedence) are required to match. The `node`-before-`default` row is in the table because today's answer comes from neither: it is what a fixed expression happened to produce.

Downstream effects: a playground program can produce a result no equivalent Node program produces, or succeed where the real program fails. Because the compile step succeeded, the divergence is attributed to the package rather than to the loader.

Behavior that must not change: relative module resolution, the module cache that lets cyclic requires settle, JSON module loading, the `main` and `index` fallback for packages without an applicable `exports` map, scoped package names, and the existing specifier-naming error message for an unavailable package.

## Approach

The invariant: when a package declares `exports`, that manifest is authoritative for every bare specifier into the package, and a conditional target is resolved by walking the object's own key order and taking the first key in an active condition set the sandbox declares.

Owner: `createSandboxRequire`'s package-resolution path, that is, `resolveSpecifier`'s bare-specifier branch, `resolvePackageEntry`, and `pickConditionalExport`.

The fix must make the sandbox **declare its active condition set explicitly**, as data the resolver reads, and justify each member by what the sandbox actually implements rather than by what reads naturally. `require` is not active because CommonJS is the format the codegen produces; it is active if and only if the sandbox executes CommonJS modules, which it does through `new Function(...)` with a `module`/`exports` wrapper. `node` is not active merely because the packages come from npm; it is active only if the sandbox provides what `node`-conditioned code expects, and the sandbox runs in a browser with no Node builtins, no real filesystem, and an injected `console`. `import` is not active if the sandbox cannot evaluate ESM. Whatever set the fix lands on, each member needs that kind of one-line justification recorded next to it, because the set is the whole answer for a package that ships several branches.

This is a decision, not a derivation: the reproduced divergence, Node selecting `node` while the sandbox selects `default`, must become a documented and deliberate choice with a stated reason, not the residue of `obj.require ?? obj.default ?? obj.node`. Selecting `default` may well be the right answer for a browser sandbox. What is not acceptable is arriving at it by accident and being unable to say why.

`default` stays a condition, matched in declaration order like any other, rather than a fixed-priority slot.

The boundary is the in-memory CommonJS `exports` contract the sandbox already claims. This is not a request for a full ESM loader, a real filesystem, or `imports`/self-reference support. Root, exact subpath, pattern subpath, `null` blockers, and array fallbacks are the shapes a manifest-first resolver must handle, and they should be handled together rather than one witness at a time, because they share the reducer.

Positive: a root condition map with `node` before `default` selects the first key in the declared active set, in the manifest's own key order, and the test asserts that target because the declared set says so rather than because a fixed expression produces it; an exact `"./sub"` entry selects its manifest target even when a physical file exists at the requested path; a `"./prefix/*"` pattern selects its expanded target; a package without `exports` still resolves through `main` and the index fallback.

Condition-set coverage, as its own group: for each condition the fix declares active, a manifest that offers only that condition resolves; for each condition it declares inactive, a manifest that offers only that condition does *not* resolve to it. Reordering the keys of a condition map changes the selected target when both keys are active, which is the assertion that distinguishes ordered selection from a renamed fixed picker and would fail against today's code either way.

Negative: a subpath present in the pack but absent from a declared `exports` map fails, and the error names the requested specifier; a `null` target fails rather than falling through to a physical path; a manifest target that names a file the pack does not contain fails rather than silently resolving elsewhere.

Boundary: nested condition objects; array fallback targets; longest-prefix precedence between two overlapping wildcard patterns; a scoped package; a manifest whose `exports` is an empty object.

Regression: the three root shapes from #670 keep resolving, with the "require is preferred over default" expectation re-derived from ordered selection rather than from the fixed expression; relative requires, the cycle-settling cache, JSON modules, and the unavailable-package error message are unchanged.

The focused cases need no network, browser, or wasm: `createSandboxRequire` takes an in-memory pack. Two files already cover this resolver and both must be updated rather than duplicated: `test_create_sandbox_require_supports_commonjs_root_exports.ts` and `test_create_sandbox_require_preserves_fallback_resolution.ts`.

Narrow proving command:

```text
pnpm --filter @ttsc/test-playground start -- --include=create_sandbox_require
```

The runner matches test names by substring and prints each executed test by name. A filter matching nothing fails loudly (`No tests matched --include=<...>`, exit 1), but this filter cannot match nothing: at the baseline it already selects the two existing cases and exits green. That is the hazard to watch here. The command proves nothing about the new coverage unless the new names are in the printed list, so read it rather than the exit code, and confirm each new case fails at the baseline.

Broader: `pnpm --filter @ttsc/test-playground start`, then `pnpm test`.

## Coordination

- Closed #670 and PR #699 are the direct predecessor. Their regression asserts the fixed picker as intended, so that expectation has to be re-derived by this fix rather than preserved verbatim. Say so in the change, because a reviewer will otherwise read it as a regression.
- This issue touches `packages/playground/src/sandbox/createSandboxRequire.ts` and its tests only. It is file-disjoint from the npm installer issues (`src/npm/**`) and the compiler-client generation issue (`src/react/`), so it can run in parallel with them.
- No open pull request touches `packages/playground`.

### LINKED ISSUE #823: fix(playground): parse npm PAX tar headers in the byte domain
## Problem

`unpackNpmTarball` in `packages/playground/src/npm/internal/npmRegistry.ts` extracts a downloaded npm tarball in the browser and mounts its files into the in-memory `node_modules` tree, the Monaco declaration libs, and the Execute runtime pack.

PAX is the tar mechanism used whenever the base 100-byte header cannot carry a path faithfully, which is exactly the case for a non-ASCII filename. A PAX extended header body is a sequence of records, each introduced by its own length in **bytes**. The implementation decodes the whole body to a JavaScript string first and then slices it by those byte counts, which are UTF-16 code-unit indices once the bytes have become a string. For any record containing a multibyte character before its boundary, the two offsets diverge.

For a `path` record the slice runs past the intended end and captures the record's terminating newline into the path. The later text-file filter then no longer recognizes the extension, and the file is dropped. The package installs, reports success, and is missing a file the registry tarball contains.

### Reproduction

A valid in-memory gzip/tar with one PAX extended header and one regular file, driven through the production `unpackNpmTarball`. The record's declared length is computed by iterating until the decimal prefix equals the record's own encoded byte length:

```js
let record = "";
for (let n = 20; n < 60; n++) {
  const candidate = `${n} path=package/한글.ts\n`;
  if (new TextEncoder().encode(candidate).length === n) { record = candidate; break; }
}
// ...type 'x' header + record, then a type '0' header + `export const value = 1;\n`
const out = await unpackNpmTarball(gz, undefined);
console.log(JSON.stringify({ files: Object.keys(out.files) }));
```

Observed, with an ASCII control of identical structure:

```text
{"record":"26 path=package/한글.ts\n","byteLength":26,"utf16Length":22}
{"files":[]}
{"control":"25 path=package/plain.ts","files":["plain.ts"]}
```

The declared length is 26 bytes; the decoded string is 22 UTF-16 code units. The multibyte record yields no files at all. The ASCII control, built by the same generator, unpacks correctly, so the failure is the byte/character mismatch and not the fixture.

### Root cause

The body is decoded before any boundary is located:

```ts
    if (type === "x") {
      paxPath = parsePaxPath(decoder.decode(body));
      continue;
    }
```

and then `length`, a byte count, is used as a string index in three places:

```ts
function parsePaxPath(text: string): string | null {
  let rest = text;
  while (rest.length > 0) {
    const space = rest.indexOf(" ");
    if (space < 0) return null;
    const length = Number(rest.slice(0, space));
    if (!Number.isFinite(length) || length <= space) return null;
    const record = rest.slice(space + 1, length - 1);
    const eq = record.indexOf("=");
    if (eq > 0 && record.slice(0, eq) === "path") return record.slice(eq + 1);
    rest = rest.slice(length);
  }
  return null;
}
```

`rest.slice(space + 1, length - 1)` is meant to stop one position before the record terminator. With a byte count applied to a shorter string, the end index is past the string, so the returned value keeps the trailing `\n`. The consumer never trims it:

```ts
    const rawPath = paxPath ?? longPath ?? readTarString(header.subarray(0, 100), decoder);
    // ...
    const rel = stripTarRoot(rawPath);
    if (!rel || !TEXT_FILE_REGEXP.test(rel)) continue;
```

`stripTarRoot` normalizes separators and strips the leading `package/` component but does not trim, and `TEXT_FILE_REGEXP` is anchored at the end of the string, so `한글.ts\n` matches nothing and the entry is skipped.

`rest = rest.slice(length)` has the same defect independently: advancing a string cursor by a byte count leaves the parser mid-record, so a multibyte record before a later `path` record can corrupt traversal even when the `path` record itself is ASCII.

### Provenance

`parsePaxPath` and `unpackNpmTarball` arrived with the dependency loader and were carried into the reusable package by `2ae95c4ff`; `git log -S parsePaxPath` traces only the loader's introduction and migration commits (`d60e9ebb9`, `2ae95c4ff`, `82d63ecbf`). PR #135 introduced the broader loader without documenting this boundary. `tests/test-playground` has no tar, PAX, or Unicode-path coverage. Searches across all issue and pull-request states for Playground tar, PAX, and Unicode path terms returned no owner.

Verified against `2b724664eee71c57cd688c55122cac137774bf62`; `packages/playground/src` is byte-identical on current `master` (`612abc2d2`).

## Consequence surface

| Case | Current | Required |
| --- | --- | --- |
| ASCII `path` record | Correct | Unchanged |
| `path` record with any multibyte character | File silently dropped from all three mounts | File present |
| Multibyte record before a later ASCII `path` record | Cursor lands mid-record; traversal can miss or misread the rest | Correct traversal |
| Multiple records in one header | Advances by byte count over a string | Advances by byte boundary |
| Declared length not a number, or shorter than its own prefix | Returns `null`, header ignored | Rejected explicitly, and the entry does not silently become a different file |
| Declared length past the end of the body | Slices past the string, no error | Rejected |
| GNU long-name (`L`) headers | Correct | Unchanged |
| Base-header paths | Correct | Unchanged |

The three affected outputs are the compiler files under the in-memory `node_modules` tree, the Monaco declaration extra-libs, and the Execute runtime files. A package whose Unicode-named `.ts`, `.js`, `.json`, or `.d.ts` file is dropped presents as a missing module or a missing type from a package that clearly contains it, or fails at runtime after a clean compile. A dropped `package.json` would additionally erase the manifest the installer reads for transitive dependencies.

## Approach

The invariant: record boundaries in a PAX extended header are located in the byte domain, and field text is decoded only after its boundaries are known.

Owner: `parsePaxPath` and its caller in `packages/playground/src/npm/internal/npmRegistry.ts`. The caller currently hands over a decoded string, so the seam moves: the parser needs the `Uint8Array` body, not `decoder.decode(body)`.

Each record's declared length must be validated against the remaining bytes and against its own prefix, and the record must be confirmed to end at its terminator, before any field is trusted. A record that fails validation is a malformed header, not a silent no-match.

Broader archive-format support is out of scope. This is the `path` keyword in a PAX extended header, parsed correctly.

Positive: an ASCII `path` record resolves as it does today; a `path` record containing multibyte characters (Korean, Japanese, emoji, combining marks) yields the exact filename and the file appears in all three mounts; a header carrying several records returns the `path` record regardless of its position among them.

Negative: a non-numeric or negative declared length, a length shorter than its own prefix, a length that overruns the body, and a record whose declared length does not land on its terminator are each rejected without inventing a path; a header with no `path` record falls back to the base header or the GNU long name as it does today.

Boundary: a record whose length exactly consumes the remaining body; a multibyte record followed by an ASCII `path` record, which is the traversal case the byte cursor breaks independently of the path value; a path at the tar 100-byte base-header limit.

Regression: GNU `L` long names, base-header paths, `stripTarRoot`'s `package/` handling, the text-file and runtime-file filters, `package.json` parsing, and abort behavior are unchanged.

The new case builds the tar in memory and asserts the unpacked file map, with the ASCII control alongside the multibyte case. `tests/test-playground` has no tar coverage today, so all of it is new.

Narrow proving command, after adding the cases under `tests/test-playground/src/features/`:

```text
pnpm --filter @ttsc/test-playground start -- --include=tarball
```

The runner matches test names by substring and prints each executed test by name. A filter matching nothing fails loudly (`No tests matched --include=<...>`, exit 1), so a typo cannot pass silently; `tarball` matches no test at the baseline, so that is the expected result until the cases exist. Read the printed names anyway, because a filter that happens to match some *other* existing case would run green without ever executing what you added. Confirm each new case fails at the audited baseline.

Broader: `pnpm --filter @ttsc/test-playground start`, then `pnpm test`.

## Coordination

- This issue edits `packages/playground/src/npm/internal/npmRegistry.ts`, the same file as the semver range-satisfaction issue. Sequence the two under one owner; they touch different functions (`parsePaxPath` / `unpackNpmTarball` versus `selectVersion`) and do not conflict semantically.
- File-disjoint from the sandbox `exports` issue, the compiler-client generation issue, the builtin-classification issue, and the template-substitution discovery issue.
- No open or closed issue or pull request owns tar parsing in this package, and no open pull request touches `packages/playground`.

### LINKED ISSUE #824: fix(playground): classify the whole pinned Node builtin surface before installing from npm
## Problem

`packageNameFromSpecifier` (`packages/playground/src/npm/packageNameFromSpecifier.ts`) decides which module specifiers in playground source are npm packages to download. Its documented contract is explicit:

```ts
/**
 * Extract the package name from a module specifier.
 *
 * Returns `null` for relative paths, hash imports, URL specifiers, and Node
 * built-in modules — the caller doesn't install those from npm.
 */
```

The builtin half of that contract is a hand-written `BUILTIN_MODULES` set of 30 names that is missing 12 of the roots the pinned Node baseline actually ships. A source file importing one of the missing names is classified as an external package, and `installPlaygroundDependencies` fetches the npm package of the same name if one exists. Several do.

The two spellings of the same import are also handled by two different rules. `node:async_hooks` returns `null` from the URL-scheme guard, before the builtin set is ever consulted, so it is skipped correctly by accident, while the bare `async_hooks` is classified as a package. The lines that strip a `node:` prefix and then look the remainder up in `BUILTIN_MODULES` are unreachable.

The result: a playground program that imports a valid Node builtin triggers a registry lookup, a tarball download, and a mount of an unrelated third-party package into the compiler files, the Monaco declaration libs, and the Execute runtime pack. The same program written with the `node:` prefix does none of that.

### Reproduction

Comparing the production classifier against `module.builtinModules` on the pinned baseline (`packages/ttsc/package.json` declares `engines.node: ">=22.15.0"`; the run used Node v22.21.0), over root builtins only, excluding the legacy underscore names:

```js
const { packageNameFromSpecifier } = require(".../packages/playground/lib/src/npm/packageNameFromSpecifier.js");
const { builtinModules } = require("node:module");
const roots = [...new Set(builtinModules.filter((m) => !m.startsWith("_") && !m.includes("/")))];
console.log(JSON.stringify({ misclassified: roots.filter((r) => packageNameFromSpecifier(r) !== null) }));
```

Observed:

```text
{"misclassified":["async_hooks","cluster","dgram","diagnostics_channel","domain","http2","inspector","repl","sys","trace_events","v8","wasi"]}
{"bare_async_hooks":"async_hooks","prefixed_async_hooks":null,"bare_fs_promises":null,"bare_http2_subpath":"http2"}
```

Twelve canonical roots are returned as npm package names. The second line shows the two-spelling asymmetry: bare `async_hooks` is a package, `node:async_hooks` is `null`. Builtin subpaths of names that *are* in the set (`fs/promises`) already resolve correctly, and subpaths of a missing name (`http2/x`) inherit the misclassification.

These are not theoretical names. A read-only registry check returned HTTP 200 for every one tested:

```text
async_hooks 200
diagnostics_channel 200
http2 200
v8 200
repl 200
domain 200
cluster 200
```

So `import "http2"` in the playground downloads and mounts an unrelated npm package named `http2`.

### Root cause

The set is a literal, and the two spellings take different exits:

```ts
export function packageNameFromSpecifier(specifier: string): string | null {
  if (
    specifier.startsWith("#") ||
    specifier.startsWith(".") ||
    specifier.startsWith("/") ||
    /^[a-z][a-z0-9+.-]*:/i.test(specifier)
  )
    return null;
  const bare = specifier.startsWith("node:") ? specifier.slice(5) : specifier;
  const first = bare.split("/")[0];
  if (first && BUILTIN_MODULES.has(first)) return null;
```

`node:` matches the scheme pattern, so the function returns at the first guard and the `bare` computation below it never sees a prefixed specifier.

### Provenance

The list and the parser arrived with the reusable playground package in `2ae95c4ff` and have not gained completeness coverage. PR #699 replaced the dependency-discovery tokenizer without changing builtin classification. There is no test asserting the set against any canonical source. Targeted searches for incomplete builtin lists and bare Node builtin installation found no duplicate issue or pull request, open or closed.

Verified against `2b724664eee71c57cd688c55122cac137774bf62`; `packages/playground/src` is byte-identical on current `master` (`612abc2d2`).

## Consequence surface

| Specifier | Current | Required |
| --- | --- | --- |
| `fs`, `path`, and the 28 other listed names | `null` | `null` |
| `async_hooks`, `cluster`, `dgram`, `diagnostics_channel`, `domain`, `http2`, `inspector`, `repl`, `sys`, `trace_events`, `v8`, `wasi` | Package name; installed if it exists on npm | `null` |
| `node:<any builtin>` | `null`, via the scheme guard | `null`, for the right reason |
| `node:<not a builtin>` | `null` | Decided by the fix; `node:` specifiers are not installable either way |
| `fs/promises`, `stream/web`, `util/types` | `null` | `null` |
| `http2/x`, `v8/tools` | Package name | `null` |
| `test`, `sqlite` (bare) | Package name | Package name; these resolve as builtins only under the `node:` prefix, so the bare form really is an npm request |
| `@scope/pkg`, `pkg/sub` | Package name | Unchanged |
| `./x`, `/x`, `#x`, `https://x` | `null` | Unchanged |

Downstream: a false package request costs a registry round trip and a tarball download, mounts unrelated declarations and runtime files into the sandbox, and can surface as a confusing type error or an unexpected runtime module. It is also avoidable supply-chain exposure: the playground fetches and executes a third-party package because the user wrote a Node builtin's name.

The guide states that "The dependency loader starts from bare imports, dynamic imports, and `require("package")` calls in the source", which is why the bare spelling is the common one.

## Approach

The invariant: the set of names the classifier treats as Node builtins agrees with the pinned Node baseline, and the bare and `node:`-prefixed spellings of the same specifier reach the same verdict through the same rule.

Owner: `packages/playground/src/npm/packageNameFromSpecifier.ts`.

The module is browser-side, so `module.builtinModules` cannot be read at runtime. A checked-in set derived from the pinned baseline is acceptable; what is not acceptable is another hand-maintained list that nothing compares against. Generating it, or asserting it against `module.builtinModules` in a Node-side test, is what keeps the two in agreement as the baseline moves. Adding the 12 names without that check reproduces this issue at the next Node bump.

Normalization has to be one rule rather than two accidental ones: strip a `node:` prefix, take the root segment before the first `/`, and consult the canonical set. The scheme guard must keep rejecting genuine URL specifiers.

A `node:`-prefix-only builtin (`node:test`, `node:sqlite` on the pinned baseline) is not resolvable in bare form, so the bare spelling must stay classified as an npm package. Any canonical set the fix derives has to preserve that distinction rather than flatten it.

Positive: every root in the canonical set returns `null` in both bare and `node:` spellings; every builtin subpath of a canonical root returns `null`; ordinary and scoped package specifiers still return their package names.

Negative: the 12 names above no longer produce a registry request; a `node:`-prefix-only builtin still returns its package name in bare form; relative, absolute, hash, and URL specifiers are unchanged.

Boundary: a bare name that is both a builtin root and a published npm package, which is the reproduced case; a subpath of a missing-today name (`http2/x`); a scoped name whose first segment coincides with a builtin.

Regression: a completeness assertion comparing the checked-in set against `module.builtinModules` on the pinned baseline, so a future Node bump fails the test rather than silently reintroducing the gap. `collectExternalPackageNames`'s existing behavior over the same specifiers is unchanged.

Narrow proving command, after adding the cases under `tests/test-playground/src/features/`:

```text
pnpm --filter @ttsc/test-playground start -- --include=package_name_from_specifier
```

The runner matches test names by substring and prints each executed test by name. A filter matching nothing fails loudly (`No tests matched --include=<...>`, exit 1), so a typo cannot pass silently; `package_name_from_specifier` matches no test at the baseline, so that is the expected result until the cases exist. Read the printed names anyway, because a filter that happens to match some *other* existing case would run green without ever executing what you added. Confirm the misclassification case fails at the audited baseline.

The completeness assertion must read `module.builtinModules` at test time rather than restating the same list a second time; a test that compares the shipped set against a copy of itself cannot fail.

Broader: `pnpm --filter @ttsc/test-playground start`, then `pnpm test`.

## Coordination

- This issue edits `packages/playground/src/npm/packageNameFromSpecifier.ts` and its tests. It is file-disjoint from the semver and PAX issues (`src/npm/internal/npmRegistry.ts`, `src/npm/installPlaygroundDependencies.ts`), the sandbox `exports` issue (`src/sandbox/`), and the compiler-client generation issue (`src/react/`).
- The template-substitution discovery issue edits `packages/playground/src/npm/collectExternalPackageNames.ts`, which calls this function. The two do not overlap in code, but they share the same acceptance surface (which specifiers become install requests), so their regressions should not contradict each other.
- No open or closed issue owns builtin classification, and no open pull request touches `packages/playground`.

### LINKED ISSUE #826: fix(playground): discover executable module loads inside template substitutions
## Problem

`collectExternalPackageNames` (`packages/playground/src/npm/collectExternalPackageNames.ts`) is the published `@ttsc/playground` API that decides which npm packages the playground fetches before compiling. Its documented contract distinguishes executable module syntax from text that merely looks like it:

```ts
/**
 * Collect the string specifiers of every executable module-loading construct in
 * `source`: `import`/`export ... from`, side-effect `import "x"`, dynamic
 * `import("x")`, and `require("x")` calls.
 *
 * The scan tokenizes `source` first so import/export/require lookalikes that
 * live inside comments, string or template contents, or regular-expression
 * literals never become package requests — only real code-level string
 * arguments are returned. Specifiers that cannot be resolved statically (a
 * template-literal or computed argument) are intentionally skipped so inert
 * text cannot drive a network install.
 */
```

The tokenizer consumes a template literal as one opaque token, including its `${...}` substitutions. A substitution is not template text: it is ordinary JavaScript that the runtime evaluates. A literal `import("pkg")` or `require("pkg")` written inside a substitution is executable module syntax, and it is skipped.

The playground therefore compiles and executes a program whose module load has no installed package behind it, and the failure presents as a missing module for a package the source clearly names.

### Reproduction

Two module loads of the same shape, one inside a substitution and one outside, driven through the production function:

```js
const source = [
  'const mod = await import(`${await import("inside-template")}`);',
  'import "outside-template";',
].join("\n");
console.log(JSON.stringify(collectExternalPackageNames(source, [])));

const source2 = 'const x = `${require("inside-require")}`;\nrequire("outside-require");';
console.log(JSON.stringify(collectExternalPackageNames(source2, [])));
```

Observed:

```text
["outside-template"]
["outside-require"]
```

The outer specifier of the first line is a template, so skipping it is correct and intended. The inner `import("inside-template")` is a literal string argument to a real dynamic import, and it is lost. The `require` form behaves identically.

### Root cause

The template branch of the tokenizer tracks `${`/`}` nesting only to find the end of the literal, and emits one `other` token for the whole thing:

```ts
    // Template literal. Contents (including `${...}` expressions) are treated as
    // opaque: a template specifier is not statically resolvable.
    if (c === "`") {
      i++;
      let depth = 0;
      while (i < n) {
        const d = source[i];
        if (d === "\\") { i += 2; continue; }
        if (depth === 0 && d === "`") { i++; break; }
        if (d === "$" && source[i + 1] === "{") { depth++; i += 2; continue; }
        if (depth > 0 && d === "}") { depth--; i++; continue; }
        i++;
      }
      tokens.push({ kind: "other" });
      continue;
    }
```

The comment states the reason as "a template specifier is not statically resolvable", which is true of the template's *value* and false of the code inside its substitutions. Those are two different things, and the branch conflates them.

`collectModuleSpecifiers` then never sees the `import` / `require` word tokens, the open paren, or the string argument, so no later stage can recover them.

### Provenance

Closed #669 and merged PR #699 replaced a raw regular expression with this tokenizer specifically to eliminate phantom dependencies from comments, strings, raw template text, and regex literals. That was a false-positive fix and its stated contract was executable module syntax only. This is the false-negative half of the same boundary, and #669's acceptance never covered it: the existing test `tests/test-playground/src/features/test_collect_external_package_names_ignores_non_code.ts` covers raw template lookalikes and a computed template argument to `require`, and contains no case with executable code inside a substitution.

Searches across open and closed issues and pull requests found no newer owner.

Verified against `2b724664eee71c57cd688c55122cac137774bf62`; `packages/playground/src` is byte-identical on current `master` (`612abc2d2`).

## Consequence surface

| Source shape | Current | Required |
| --- | --- | --- |
| `import("pkg")` at top level | Collected | Collected |
| `` `${import("pkg")}` `` | Skipped | Collected |
| `` `${require("pkg")}` `` | Skipped | Collected |
| Nested template inside a substitution containing a real import | Skipped | Collected |
| Raw template text spelling `import("pkg")` with no substitution | Skipped | Skipped |
| `` import(`pkg-${name}`) `` (computed specifier) | Skipped | Skipped |
| `import("pkg")` inside a comment, string, or regex literal | Skipped | Skipped |
| A string, comment, or regex literal *inside* a substitution that spells a module load | Skipped | Skipped |
| `obj.require(...)`, `import.meta` | Skipped | Skipped |
| Tagged template with a substitution containing a real import | Skipped | Collected |

The consumers are `PlaygroundShell`, which calls this function to decide what to install before compiling, the hosted ttsc.dev playground through `TtscWebsitePlaygroundShell`, and any embedder using the exported function directly. The guide states that "The dependency loader starts from bare imports, dynamic imports, and `require("package")` calls in the source", with no exception for interpolation position.

## Approach

The invariant: a template literal's quasis are inert text, and its substitutions are code. The tokenizer must tokenize substitutions with the same rules it applies to top-level source, recursively, and keep treating quasis as opaque.

Owner: the template branch of `tokenize` in `packages/playground/src/npm/collectExternalPackageNames.ts`, and whatever token stream shape that requires. Nesting depth is already tracked there, so the missing piece is recursion into the substitution rather than a new scanner.

The false-positive guarantees #669 established are the constraint, not an afterthought: entering a substitution must not weaken string, comment, or regex handling inside it, and a computed specifier inside a substitution must stay skipped for the same reason a computed specifier outside one is skipped.

Positive: a literal `import("pkg")` and a literal `require("pkg")` inside a substitution are collected; the same inside a nested template inside a substitution is collected; a substitution containing several statements collects each literal load; a tagged template's substitutions behave like an untagged one's.

Negative: raw template text spelling a module load is still skipped; a computed specifier (`` import(`a-${b}`) ``) is still skipped, inside and outside a substitution; strings, line and block comments, and regex literals *inside* a substitution do not produce specifiers; `obj.require(...)` and `import.meta` are still not module loads.

Boundary: an unterminated template; a substitution containing an unbalanced `}` inside a string; a template inside a substitution inside a template; a substitution containing a regex literal whose body contains a backtick.

Regression: every case in `test_collect_external_package_names_ignores_non_code.ts` keeps its current verdict.

Narrow proving command, after adding the new cases to `tests/test-playground/src/features/`:

```text
pnpm --filter @ttsc/test-playground start -- --include=collect_external_package_names
```

The runner matches test names by substring and prints each executed test by name. A filter matching nothing fails loudly (`No tests matched --include=<...>`, exit 1), but this filter cannot match nothing: at the baseline it already selects `test_collect_external_package_names_ignores_non_code` and exits green. That is the hazard here. The command proves nothing about the new coverage unless the new name is in the printed list, so read it rather than the exit code, and confirm the new case fails before the fix.

Broader: `pnpm --filter @ttsc/test-playground start` (all playground features), then `pnpm test`.

## Coordination

- Closed #669 and PR #699 are the direct predecessor and must not regress; this issue owns the opposite direction of the same boundary.
- This issue edits `packages/playground/src/npm/collectExternalPackageNames.ts` and its test. It is file-disjoint from the semver and PAX issues (`src/npm/internal/npmRegistry.ts`), the sandbox `exports` issue (`src/sandbox/`), and the compiler-client generation issue (`src/react/`).
- The builtin-classification issue edits `packages/playground/src/npm/packageNameFromSpecifier.ts`, which this function calls. No code overlap, but both change which specifiers become install requests, so their regressions must agree.
- No open pull request touches `packages/playground`.

samchon:
## Intent

Restore the public `@ttsc/playground` dependency-loader, sandbox, and
compiler-client contracts as one compatible package batch.

Closes #802
Closes #803
Closes #820
Closes #823
Closes #824
Closes #826

## Scope

- fence overlapping compiler-client connection generations;
- resolve declared npm ranges and aliases, reconcile constraints, and parse PAX
  records in bytes;
- make sandbox package `exports` authoritative with explicitly declared active
  conditions;
- classify the pinned Node builtin surface and scan executable template
  substitutions;
- add focused browser-free feature coverage, then validate the full playground
  feature package.

## Campaign verification

This is the empty reservation commit. Implementation verification is pending.
Every campaign push and pull-request mutation receives an exact-SHA Actions
cancellation record; repository workflow configuration remains unchanged.

### DIFF
--- a/packages/playground/package.json
+++ b/packages/playground/package.json
@@ -33,7 +33,8 @@
   },
   "homepage": "https://ttsc.dev",
   "dependencies": {
-    "lz-string": "^1.5.0"
+    "lz-string": "^1.5.0",
+    "semver": "^7.8.1"
   },
   "peerDependencies": {
     "@monaco-editor/react": "^4.6.0",
@@ -46,6 +47,7 @@
   "devDependencies": {
     "@monaco-editor/react": "^4.6.0",
     "@ttsc/wasm": "workspace:^",
+    "@types/semver": "^7.7.1",
     "@types/node": "catalog:utils",
     "@types/react": "^18.3.1",
     "monaco-editor": "^0.52.0",
@@ -55,5 +57,12 @@
     "tgrid": "^1.0.3",
     "typescript": "catalog:typescript"
   },
-  "keywords": ["ttsc", "playground", "typescript", "browser", "wasm", "react"]
+  "keywords": [
+    "ttsc",
+    "playground",
+    "typescript",
+    "browser",
+    "wasm",
+    "react"
+  ]
 }
--- a/packages/playground/src/npm/collectExternalPackageNames.ts
+++ b/packages/playground/src/npm/collectExternalPackageNames.ts
@@ -152,14 +152,6 @@ function tokenize(source: string): Token[] {
 
   // A `/` opens a regex only in operator/statement position — never right after
   // a value (identifier, number, string, template, regex, `)` or `]`).
-  const regexAllowed = (): boolean => {
-    const prev = tokens[tokens.length - 1];
-    if (!prev) return true;
-    if (prev.kind === "string" || prev.kind === "other") return false;
-    if (prev.kind === "word") return REGEX_PRECEDING_KEYWORDS.has(prev.value);
-    return prev.value !== ")" && prev.value !== "]";
-  };
-
   let i = 0;
   while (i < n) {
     const c = source[i]!;
@@ -189,25 +181,8 @@ function tokenize(source: string): Token[] {
       continue;
     }
     // Regular-expression literal.
-    if (c === "/" && regexAllowed()) {
-      i++;
-      let inClass = false;
-      while (i < n) {
-        const d = source[i];
-        if (d === "\\") {
-          i += 2;
-          continue;
-        }
-        if (d === "\n") break;
-        if (d === "[") inClass = true;
-        else if (d === "]") inClass = false;
-        else if (d === "/" && !inClass) {
-          i++;
-          break;
-        }
-        i++;
-      }
-      while (i < n && isIdPart(source[i]!)) i++; // flags
+    if (c === "/" && isRegexAllowed(tokens)) {
+      i = skipRegularExpression(source, i, isIdPart);
       tokens.push({ kind: "other" });
       continue;
     }
@@ -233,29 +208,26 @@ function tokenize(source: string): Token[] {
       tokens.push({ kind: "string", value });
       continue;
     }
-    // Template literal. Contents (including `${...}` expressions) are treated as
-    // opaque: a template specifier is not statically resolvable.
+    // Template quasis are opaque, but `${...}` substitutions are executable
+    // JavaScript and must receive the same lexical treatment as top-level code.
     if (c === "`") {
       i++;
-      let depth = 0;
       while (i < n) {
         const d = source[i];
         if (d === "\\") {
           i += 2;
           continue;
         }
-        if (depth === 0 && d === "`") {
+        if (d === "`") {
           i++;
           break;
         }
         if (d === "$" && source[i + 1] === "{") {
-          depth++;
-          i += 2;
-          continue;
-        }
-        if (depth > 0 && d === "}") {
-          depth--;
-          i++;
+          const start = i + 2;
+          const end = findTemplateSubstitutionEnd(source, start);
+          tokens.push({ kind: "other" }, ...tokenize(source.slice(start, end)));
+          tokens.push({ kind: "other" });
+          i = end < n ? end + 1 : end;
           continue;
         }
         i++;
@@ -284,4 +256,105 @@ function tokenize(source: string): Token[] {
     i++;
   }
   return tokens;
+}
+
+function findTemplateSubstitutionEnd(source: string, start: number): number {
+  let depth = 1;
+  for (let i = start; i < source.length; i++) {
+    const c = source[i]!;
+    if (c === "\\") {
+      i++;
+      continue;
+    }
+    if (c === "'" || c === '"') {
+      i = skipQuoted(source, i, c);
+      continue;
+    }
+    if (c === "`") {
+      i = skipTemplate(source, i);
+      continue;
+    }
+    if (c === "/" && source[i + 1] === "/") {
+      i += 2;
+      while (i < source.length && source[i] !== "\n") i++;
+      continue;
+    }
+    if (c === "/" && source[i + 1] === "*") {
+      i += 2;
+      while (i < source.length && !(source[i] === "*" && source[i + 1] === "/"))
+        i++;
+      i++;
+      continue;
+    }
+    if (c === "/" && isRegexAllowed(tokenize(source.slice(start, i)))) {
+      i = skipRegularExpression(source, i, (character) =>
+        /[A-Za-z0-9_$]/.test(character),
+      );
+      i--;
+      continue;
+    }
+    if (c === "{") depth++;
+    else if (c === "}" && --depth === 0) return i;
+  }
+  return source.length;
+}
+
+function isRegexAllowed(tokens: readonly Token[]): boolean {
+  const previous = tokens[tokens.length - 1];
+  if (!previous) return true;
+  if (previous.kind === "string" || previous.kind === "other") return false;
+  if (previous.kind === "word")
+    return REGEX_PRECEDING_KEYWORDS.has(previous.value);
+  return previous.value !== ")" && previous.value !== "]";
+}
+
+function skipRegularExpression(
+  source: string,
+  start: number,
+  isIdentifierPart: (character: string) => boolean,
+): number {
+  let index = start + 1;
+  let inClass = false;
+  while (index < source.length) {
+    const character = source[index];
+    if (character === "\\") {
+      index += 2;
+      continue;
+    }
+    if (character === "\n") break;
+    if (character === "[") inClass = true;
+    else if (character === "]") inClass = false;
+    else if (character === "/" && !inClass) {
+      index++;
+      break;
+    }
+    index++;
+  }
+  while (index < source.length && isIdentifierPart(source[index]!)) index++;
+  return index;
+}
+
+function skipQuoted(source: string, start: number, quote: string): number {
+  for (let i = start + 1; i < source.length; i++) {
+    if (source[i] === "\\") {
+      i++;
+      continue;
+    }
+    if (source[i] === quote || source[i] === "\n") return i;
+  }
+  return source.length;
+}
+
+function skipTemplate(source: string, start: number): number {
+  for (let i = start + 1; i < source.length; i++) {
+    if (source[i] === "\\") {
+      i++;
+      continue;
+    }
+    if (source[i] === "`") return i;
+    if (source[i] === "$" && source[i + 1] === "{") {
+      i = findTemplateSubstitutionEnd(source, i + 2);
+    }
+  }
+  return source.length;
 }
--- a/packages/playground/src/npm/collectExternalPackageNames.ts
+++ b/packages/playground/src/npm/collectExternalPackageNames.ts
@@ -260,18 +260,26 @@ function tokenize(source: string): Token[] {
 
 function findTemplateSubstitutionEnd(source: string, start: number): number {
   let depth = 1;
+  let previous: Token | undefined;
+  const isIdentifierStart = (character: string): boolean =>
+    /[A-Za-z_$]/.test(character);
+  const isIdentifierPart = (character: string): boolean =>
+    /[A-Za-z0-9_$]/.test(character);
   for (let i = start; i < source.length; i++) {
     const c = source[i]!;
+    if (/\s/.test(c)) continue;
     if (c === "\\") {
       i++;
       continue;
     }
     if (c === "'" || c === '"') {
       i = skipQuoted(source, i, c);
+      previous = { kind: "string", value: "" };
       continue;
     }
     if (c === "`") {
       i = skipTemplate(source, i);
+      previous = { kind: "other" };
       continue;
     }
     if (c === "/" && source[i + 1] === "/") {
@@ -286,21 +294,41 @@ function findTemplateSubstitutionEnd(source: string, start: number): number {
       i++;
       continue;
     }
-    if (c === "/" && isRegexAllowed(tokenize(source.slice(start, i)))) {
+    if (c === "/" && isRegexAllowedAfter(previous)) {
       i = skipRegularExpression(source, i, (character) =>
         /[A-Za-z0-9_$]/.test(character),
       );
       i--;
+      previous = { kind: "other" };
+      continue;
+    }
+    if (isIdentifierStart(c)) {
+      let end = i + 1;
+      while (end < source.length && isIdentifierPart(source[end]!)) end++;
+      previous = { kind: "word", value: source.slice(i, end) };
+      i = end - 1;
+      continue;
+    }
+    if (c >= "0" && c <= "9") {
+      let end = i + 1;
+      while (end < source.length && /[0-9a-fA-FxXoObBeE._]/.test(source[end]!))
+        end++;
+      previous = { kind: "other" };
+      i = end - 1;
       continue;
     }
     if (c === "{") depth++;
     else if (c === "}" && --depth === 0) return i;
+    previous = { kind: "punct", value: c };
   }
   return source.length;
 }
 
 function isRegexAllowed(tokens: readonly Token[]): boolean {
-  const previous = tokens[tokens.length - 1];
+  return isRegexAllowedAfter(tokens[tokens.length - 1]);
+}
+
+function isRegexAllowedAfter(previous: Token | undefined): boolean {
   if (!previous) return true;
   if (previous.kind === "string" || previous.kind === "other") return false;
   if (previous.kind === "word")
--- a/packages/playground/src/npm/installPlaygroundDependencies.ts
+++ b/packages/playground/src/npm/installPlaygroundDependencies.ts
@@ -228,7 +228,8 @@ function selectQueuedVersion(
     const message = error instanceof Error ? error.message : String(error);
     const requestedBy = requests
       .map(
-        ({ requester, range }) => `${requester} requests ${JSON.stringify(range)}`,
+        ({ requester, range }) =>
+          `${requester} requests ${JSON.stringify(range)}`,
       )
       .join("; ");
     throw new Error(`${message} Requested by ${requestedBy}.`);
--- a/packages/playground/src/npm/installPlaygroundDependencies.ts
+++ b/packages/playground/src/npm/installPlaygroundDependencies.ts
@@ -5,6 +5,7 @@ import { BUILT_IN_PLAYGROUND_PACKAGES } from "./BUILT_IN_PLAYGROUND_PACKAGES";
 import {
   DECLARATION_FILE_REGEXP,
   type IQueueItem,
+  type IVersionRequest,
   downloadTarball,
   enqueuePackageDependencies,
   fetchNpmMetadata,
@@ -63,8 +64,8 @@ export async function installPlaygroundDependencies(
           `Conflicting npm aliases for ${normalized.name}: ${known.registryName} and ${normalized.registryName}.`,
         );
       }
-      known.ranges ??= [known.range];
-      known.ranges.push(normalized.range);
+      known.requests ??= [toVersionRequest(known)];
+      known.requests.push(toVersionRequest(normalized));
       known.optional &&= normalized.optional;
       const completed = done.get(normalized.name);
       const metadata = metadataByName.get(normalized.name);
@@ -73,7 +74,7 @@ export async function installPlaygroundDependencies(
         // already been installed. Pin the installed version into the combined
         // solve so a compatible late constraint is accepted, but a range that
         // rejects the mounted version fails instead of being silently lost.
-        selectVersion(metadata, [...known.ranges, completed]);
+        selectQueuedVersion(metadata, known, [completed]);
       } else if (completed !== undefined && !known.optional) {
         // An optional 404 may later be reached through a required edge. Retry
         // that edge so the required dependency cannot remain silently absent.
@@ -83,7 +84,7 @@ export async function installPlaygroundDependencies(
       return;
     }
     if (installed.has(normalized.name)) return;
-    normalized.ranges = [normalized.range];
+    normalized.requests = [toVersionRequest(normalized)];
     queued.set(normalized.name, normalized);
     queue.push(normalized);
     report("queued", normalized, `Queued ${normalized.name}`);
@@ -135,7 +136,7 @@ export async function installPlaygroundDependencies(
     }
 
     metadataByName.set(item.name, metadata);
-    const version = selectVersion(metadata, item.ranges ?? [item.range]);
+    const version = selectQueuedVersion(metadata, item);
     const versionMetadata = metadata.versions[version];
     const tarball = versionMetadata?.dist?.tarball;
     if (!versionMetadata || !tarball) {
@@ -178,12 +179,13 @@ export async function installPlaygroundDependencies(
     installed.add(item.name);
     report("done", item, `Installed ${item.name}@${version}`, version);
 
-    enqueuePackageDependencies(packageJson, enqueue);
+    enqueuePackageDependencies(packageJson, enqueue, item.name);
     if (declarationCount === 0 && !item.name.startsWith("@types/")) {
       enqueue({
         name: toTypesPackageName(item.name),
         range: "*",
         optional: true,
+        requester: item.name,
       });
     }
   }
@@ -206,4 +208,29 @@ function normalizeRegistryRequest(item: IQueueItem): IQueueItem {
     registryName: match[1]!,
     range: match[2] || "*",
   };
+}
+
+function toVersionRequest(item: IQueueItem): IVersionRequest {
+  return { range: item.range, requester: item.requester };
+}
+
+function selectQueuedVersion(
+  metadata: Parameters<typeof selectVersion>[0],
+  item: IQueueItem,
+  extraRanges: readonly string[] = [],
+): string {
+  const requests = item.requests ?? [toVersionRequest(item)];
+  const ranges = [...requests.map((request) => request.range), ...extraRanges];
+  try {
+    return selectVersion(metadata, ranges);
+  } catch (error) {
+    if (requests.length < 2) throw error;
+    const message = error instanceof Error ? error.message : String(error);
+    const requestedBy = requests
+      .map(
+        ({ requester, range }) => `${requester} requests ${JSON.stringify(range)}`,
+      )
+      .join("; ");
+    throw new Error(`${message} Requested by ${requestedBy}.`);
+  }
 }
--- a/packages/playground/src/npm/installPlaygroundDependencies.ts
+++ b/packages/playground/src/npm/installPlaygroundDependencies.ts
@@ -43,8 +43,9 @@ export async function installPlaygroundDependencies(
   const installed = new Set(options.installedPackages ?? []);
   const maxPackages = options.maxPackages ?? DEFAULT_MAX_PACKAGES;
   const queue: IQueueItem[] = [];
-  const queued = new Set<string>();
-  const done = new Set<string>();
+  const queued = new Map<string, IQueueItem>();
+  const done = new Map<string, string>();
+  const metadataByName = new Map<string, Parameters<typeof selectVersion>[0]>();
   const result: IPlaygroundDependencyInstallResult = {
     packages: [],
     compilerFiles: {},
@@ -53,11 +54,39 @@ export async function installPlaygroundDependencies(
   };
 
   const enqueue = (item: IQueueItem): void => {
-    if (ignored.has(item.name) || installed.has(item.name)) return;
-    if (queued.has(item.name) || done.has(item.name)) return;
-    queued.add(item.name);
-    queue.push(item);
-    report("queued", item, `Queued ${item.name}`);
+    const normalized = normalizeRegistryRequest(item);
+    if (ignored.has(normalized.name)) return;
+    const known = queued.get(normalized.name);
+    if (known) {
+      if (known.registryName !== normalized.registryName) {
+        throw new Error(
+          `Conflicting npm aliases for ${normalized.name}: ${known.registryName} and ${normalized.registryName}.`,
+        );
+      }
+      known.ranges ??= [known.range];
+      known.ranges.push(normalized.range);
+      known.optional &&= normalized.optional;
+      const completed = done.get(normalized.name);
+      const metadata = metadataByName.get(normalized.name);
+      if (completed !== undefined && metadata) {
+        // A dependency can discover an additional range after this package has
+        // already been installed. Pin the installed version into the combined
+        // solve so a compatible late constraint is accepted, but a range that
+        // rejects the mounted version fails instead of being silently lost.
+        selectVersion(metadata, [...known.ranges, completed]);
+      } else if (completed !== undefined && !known.optional) {
+        // An optional 404 may later be reached through a required edge. Retry
+        // that edge so the required dependency cannot remain silently absent.
+        done.delete(normalized.name);
+        queue.push(known);
+      }
+      return;
+    }
+    if (installed.has(normalized.name)) return;
+    normalized.ranges = [normalized.range];
+    queued.set(normalized.name, normalized);
+    queue.push(normalized);
+    report("queued", normalized, `Queued ${normalized.name}`);
   };
 
   const report = (
@@ -77,7 +106,7 @@ export async function installPlaygroundDependencies(
   };
 
   for (const name of packageNames) {
-    enqueue({ name, range: "latest", optional: false });
+    enqueue({ name, range: "*", optional: false, requester: "source" });
   }
 
   for (let index = 0; index < queue.length; index++) {
@@ -94,23 +123,24 @@ export async function installPlaygroundDependencies(
     report("resolve", item, `Resolving ${item.name}`);
     const metadata = await fetchNpmMetadata(
       fetchImpl,
-      item.name,
+      item.registryName ?? item.name,
       item.optional,
       options.signal,
     );
     throwIfAborted(options.signal);
     if (!metadata) {
-      done.add(item.name);
+      done.set(item.name, "");
       report("skip", item, `Skipped optional ${item.name}`);
       continue;
     }
 
-    const version = selectVersion(metadata, item.range);
+    metadataByName.set(item.name, metadata);
+    const version = selectVersion(metadata, item.ranges ?? [item.range]);
     const versionMetadata = metadata.versions[version];
     const tarball = versionMetadata?.dist?.tarball;
     if (!versionMetadata || !tarball) {
       if (item.optional) {
-        done.add(item.name);
+        done.set(item.name, version);
         report("skip", item, `Skipped optional ${item.name}`);
         continue;
       }
@@ -144,7 +174,7 @@ export async function installPlaygroundDependencies(
       declarationCount,
     });
 
-    done.add(item.name);
+    done.set(item.name, version);
     installed.add(item.name);
     report("done", item, `Installed ${item.name}@${version}`, version);
 
@@ -152,7 +182,7 @@ export async function installPlaygroundDependencies(
     if (declarationCount === 0 && !item.name.startsWith("@types/")) {
       enqueue({
         name: toTypesPackageName(item.name),
-        range: "latest",
+        range: "*",
         optional: true,
       });
     }
@@ -160,4 +190,20 @@ export async function installPlaygroundDependencies(
 
   report("done", null, "Dependency install complete");
   return result;
+}
+
+function normalizeRegistryRequest(item: IQueueItem): IQueueItem {
+  if (!item.range.startsWith("npm:"))
+    return { ...item, registryName: item.name };
+  const match = /^npm:((?:@[^/]+\/)?[^@/]+)(?:@(.+))?$/.exec(item.range);
+  if (!match) {
+    throw new Error(
+      `Unsupported npm alias ${JSON.stringify(item.range)} for ${item.name}.`,
+    );
+  }
+  return {
+    ...item,
+    registryName: match[1]!,
+    range: match[2] || "*",
+  };
 }
--- a/packages/playground/src/npm/internal/npmRegistry.ts
+++ b/packages/playground/src/npm/internal/npmRegistry.ts
@@ -1,3 +1,5 @@
+import { maxSatisfying, satisfies, valid, validRange } from "semver";
+
 // Internal npm-registry helpers used by `installPlaygroundDependencies`.
 // Grouped in one file because every helper is privately coupled to the tar
 // + version-resolve + tarball-extract flow; splitting them per the public
@@ -37,6 +39,9 @@ export interface IQueueItem {
   name: string;
   range: string;
   optional: boolean;
+  registryName?: string;
+  ranges?: string[];
+  requester?: string;
 }
 
 export type FetchLike = (
@@ -79,19 +84,43 @@ export async function fetchNpmMetadata(
   return (await response.json()) as INpmMetadata;
 }
 
-export function selectVersion(metadata: INpmMetadata, range: string): string {
+export function selectVersion(
+  metadata: INpmMetadata,
+  ranges: readonly string[] | string,
+): string {
   const versions = metadata.versions;
-  if (versions[range]) return range;
-  const tag = metadata["dist-tags"]?.[range];
-  if (tag && versions[tag]) return tag;
-  const normalized = range.replace(/^[~^<>= ]+/, "").trim();
-  if (normalized && versions[normalized]) return normalized;
-  const latest = metadata["dist-tags"]?.latest;
-  if (latest && versions[latest]) return latest;
-  const all = Object.keys(versions).sort(compareVersionDesc);
-  const fallback = all[0];
-  if (!fallback) throw new Error(`No versions found for ${metadata.name}.`);
-  return fallback;
+  const requested = Array.isArray(ranges) ? ranges : [ranges];
+  const semverRanges = requested.filter((range) => validRange(range) !== null);
+  const tags = requested.filter((range) => validRange(range) === null);
+  const taggedVersions = new Set<string>();
+  for (const tag of tags) {
+    const version = metadata["dist-tags"]?.[tag];
+    if (!version || !versions[version]) {
+      throw new Error(
+        `No npm dist-tag ${JSON.stringify(tag)} exists for ${metadata.name}.`,
+      );
+    }
+    taggedVersions.add(version);
+  }
+  if (taggedVersions.size > 1) {
+    throw new Error(
+      `Conflicting npm tags for ${metadata.name}: ${requested.join(", ")}.`,
+    );
+  }
+
+  const candidates = Object.keys(versions).filter(
+    (version) =>
+      valid(version) !== null &&
+      semverRanges.every((range) => satisfies(version, range)) &&
+      (taggedVersions.size === 0 || taggedVersions.has(version)),
+  );
+  const selected = maxSatisfying(candidates, "*");
+  if (selected) return selected;
+  throw new Error(
+    `No version of ${metadata.name} satisfies ${requested
+      .map((range) => JSON.stringify(range))
+      .join(", ")}.`,
+  );
 }
 
 export async function downloadTarball(
@@ -136,7 +165,7 @@ export async function unpackNpmTarball(
       continue;
     }
     if (type === "x") {
-      paxPath = parsePaxPath(decoder.decode(body));
+      paxPath = parsePaxPath(body);
       continue;
     }
     if (type !== "0" && type !== "\0") {
@@ -256,19 +285,32 @@ function parseOctal(bytes: Uint8Array): number {
   return text ? Number.parseInt(text, 8) : 0;
 }
 
-function parsePaxPath(text: string): string | null {
-  let rest = text;
-  while (rest.length > 0) {
-    const space = rest.indexOf(" ");
-    if (space < 0) return null;
-    const length = Number(rest.slice(0, space));
-    if (!Number.isFinite(length) || length <= space) return null;
-    const record = rest.slice(space + 1, length - 1);
+function parsePaxPath(bytes: Uint8Array): string | null {
+  const decoder = new TextDecoder();
+  let offset = 0;
+  let path: string | null = null;
+  while (offset < bytes.length) {
+    let space = offset;
+    while (space < bytes.length && bytes[space] !== 0x20) space++;
+    if (space === bytes.length) throw new Error("Invalid PAX header length.");
+    const lengthText = decoder.decode(bytes.subarray(offset, space));
+    if (!/^[1-9][0-9]*$/.test(lengthText))
+      throw new Error("Invalid PAX header length.");
+    const length = Number(lengthText);
+    const end = offset + length;
+    if (
+      !Number.isSafeInteger(length) ||
+      length <= space - offset + 1 ||
+      end > bytes.length ||
+      bytes[end - 1] !== 0x0a
+    )
+      throw new Error("Invalid PAX header record.");
+    const record = decoder.decode(bytes.subarray(space + 1, end - 1));
     const eq = record.indexOf("=");
-    if (eq > 0 && record.slice(0, eq) === "path") return record.slice(eq + 1);
-    rest = rest.slice(length);
+    if (eq > 0 && record.slice(0, eq) === "path") path = record.slice(eq + 1);
+    offset = end;
   }
-  return null;
+  return path;
 }
 
 function stripTarRoot(path: string): string {
@@ -278,15 +320,4 @@ function stripTarRoot(path: string): string {
     return normalized.slice("package/".length);
   const slash = normalized.indexOf("/");
   return slash < 0 ? normalized : normalized.slice(slash + 1);
-}
-
-function compareVersionDesc(a: string, b: string): number {
-  const pa = a.split(/[.-]/).map((part) => Number.parseInt(part, 10));
-  const pb = b.split(/[.-]/).map((part) => Number.parseInt(part, 10));
-  for (let i = 0; i < Math.max(pa.length, pb.length); i++) {
-    const av = Number.isFinite(pa[i]) ? pa[i]! : 0;
-    const bv = Number.isFinite(pb[i]) ? pb[i]! : 0;
-    if (av !== bv) return bv - av;
-  }
-  return b.localeCompare(a);
 }
--- a/packages/playground/src/npm/internal/npmRegistry.ts
+++ b/packages/playground/src/npm/internal/npmRegistry.ts
@@ -39,9 +39,14 @@ export interface IQueueItem {
   name: string;
   range: string;
   optional: boolean;
+  requester: string;
   registryName?: string;
-  ranges?: string[];
-  requester?: string;
+  requests?: IVersionRequest[];
+}
+
+export interface IVersionRequest {
+  range: string;
+  requester: string;
 }
 
 export type FetchLike = (
@@ -242,9 +247,11 @@ export function enqueuePackageDependencies(
     peerDependenciesMeta?: Record<string, { optional?: boolean }>;
   },
   enqueue: (item: IQueueItem) => void,
+  requester: string,
 ): void {
   for (const [name, range] of Object.entries(packageJson.dependencies ?? {})) {
-    if (isRegistryRange(range)) enqueue({ name, range, optional: false });
+    if (isRegistryRange(range))
+      enqueue({ name, range, optional: false, requester });
   }
   for (const [name, range] of Object.entries(
     packageJson.peerDependencies ?? {},
@@ -257,7 +264,7 @@ export function enqueuePackageDependencies(
     // silent skip that leaves the wasm-side compile reporting a generic
     // "Cannot find module" with no breadcrumb back to the dep installer.
     if (!optional && isRegistryRange(range))
-      enqueue({ name, range, optional: false });
+      enqueue({ name, range, optional: false, requester });
   }
 }
 
--- a/packages/playground/src/npm/packageNameFromSpecifier.ts
+++ b/packages/playground/src/npm/packageNameFromSpecifier.ts
@@ -1,15 +1,22 @@
 const BUILTIN_MODULES = new Set([
   "assert",
+  "async_hooks",
   "buffer",
   "child_process",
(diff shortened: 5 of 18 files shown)


samchon:
Solo Self-Review found one in-scope scanner issue before merge: indTemplateSubstitutionEnd re-tokenizes the full substitution prefix for every /, making a long substitution with many division operators quadratic. I will replace that lookup with single-pass lexical state, add a division-in-substitution regression beside the regex boundary, then rebuild and restart the complete solo review.

samchon:
Second solo review found a remaining #803 acceptance gap: the new queue type has a `requester` field, but dependency edges never populate it and a disjoint-range error therefore names only ranges, not both requesters as required. I will thread requester provenance from the source/root and each resolved package through enqueueing, make the conflict diagnostic identify those edges, and add an end-to-end regression before restarting the full review.

samchon:
Fresh solo review found a #802 ownership hole in the ordinary boot-failure path: when no reset is in flight, `connect()` clears `current` but drops its connector without closing it. That violates the issue invariant that every allocated Worker remains reachable until one close, and it leaves the genuine-failure/retry acceptance case incomplete. I will give each connection an idempotent shared close operation, use it for both failure cleanup and reset, and extend the controlled lifecycle test for normal failure, retry, and a throwing close before restarting the full review.

samchon:
Final solo Self-Review of `origin/master...4559ffc7a974a2d39ae51c96d906e0d41ca668ce` is clean. It re-read the complete source, regression, lock, and documentation diff after the ordinary-failure ownership repair and the requester-provenance repair. Verified: `pnpm --filter @ttsc/playground build`; `pnpm --filter @ttsc/test-playground start` (12/12); and `git diff --check`. The broader `pnpm test` was attempted earlier but its `check:flags` prerequisite is unavailable in this sparse worktree because `.vscode/gofmt-2spaces.sh` is absent; the package-scoped acceptance suite is green. No `pnpm format` was run.

PR #886, opened 2026-07-20
fix(bundlers): preserve cache freshness across resolution changes

### LINKED ISSUE #830: fix(metro): let a clean run record the observation that clears a volatile snapshot
## Problem

`packages/metro/src/core/fingerprint.ts` folds a per-run nonce into Metro's transform cache key whenever the recorded snapshot is marked volatile, which soundly disables cross-run cache reuse for a project whose plugin declared an output that depends on non-file inputs. The module documents an exit from that state:

```ts
 * - A plugin-declared volatile output (non-file inputs; unrepresentable in any
 *   file fingerprint) marks the snapshot volatile, which also folds a nonce
 *   until a later run records the volatile declaration gone.
```

A later run cannot record that. `prepareSnapshot` clears the main `volatile` flag only when at least one worker document exists for the run; with zero workers it preserves the previous value. The worker recorder starts clean and writes only when its state becomes dirty, and the two entry points that can make it dirty are `record()`, which returns without writing for any input already covered by the project fingerprint walk, and `recordVolatile()`, which writes only the positive declaration.

So a run that transforms files cleanly, with every derived input inside the project walk, produces no worker document at all. The next config process sees zero workers, keeps `volatile: true`, and folds a fresh nonce again. There is no run that ends the cycle. Metro's persistent transform cache never reuses anything for that project again, even though the volatile plugin behavior was removed, until the user deletes the snapshot directory or an unrelated out-of-walk input happens to appear.

The affected user is any React Native or Expo project on `@ttsc/metro` that once had a volatile plugin declaration. Output stays correct; every build after that pays a full re-transform.

### Reproduction

The probe drives the real `fingerprint.ts` module through the exact lifecycle: a config process prepares the snapshot, a worker declares volatility, the next config process compacts it, then a clean worker transforms a file whose only recorded input is inside the project walk, then the next config process compacts again. The control repeats the last step with an out-of-walk input.

```ts
import {
  computeProjectFingerprint, createSnapshotRecorder, prepareSnapshot,
  readSnapshotState, resolveFingerprintBase,
} from "packages/metro/src/core/fingerprint.ts";

// project root with tsconfig.json and src/a.ts, nothing else
prepareSnapshot(projectRoot);
createSnapshotRecorder().recordVolatile({ projectRoot });
prepareSnapshot(projectRoot);                                  // run 2: compaction
createSnapshotRecorder().record({ projectRoot, input });       // clean, in-walk
prepareSnapshot(projectRoot);                                  // run 3: compaction
// control: same, with an out-of-walk input
createSnapshotRecorder().record({ projectRoot, input: externalFile });
prepareSnapshot(projectRoot);
```

Observed:

```text
run1 workers ["graph-inputs.worker-2h7k-d61a3bd61e99.json"]
run2 state {"files":[],"id":"d657e95947a7f02a298f09fcf7d2b930","volatile":true}
run2 workers after clean in-walk transform []
run3 state {"files":[],"id":"d657e95947a7f02a298f09fcf7d2b930","volatile":true}
run3 keys stable false
control workers ["graph-inputs.worker-2h7k-c18d7240793f.json"]
control state {"files":["C:\\Users\\...\\ttsc-metro-external.json"],"id":"d657e95947a7f02a298f09fcf7d2b930","volatile":false}
control keys stable true
```

The volatile declaration writes a worker document. The clean in-walk transform writes none. The flag survives compaction, and two consecutive `computeProjectFingerprint` calls differ. The control, which differs only by one out-of-walk input, writes a worker document, clears the flag, and stabilizes the key.

### Root cause

`prepareSnapshot` treats "no worker documents" as "no information", and keeps the old verdict:

```ts
    const volatile =
      workers.entries.length === 0
        ? (main?.volatile ?? false)
        : // Worker files carry the previous run's fresh observations, so they
          // own the volatile verdict: a removed volatile declaration must be
          // able to clear the sticky flag.
          workers.entries.some((entry) => entry.volatile);
```

The comment is the intended design. The recorder cannot supply the input it needs:

```ts
    record(props) {
      const base = resolveFingerprintBase(props.projectRoot);
      const state = stateFor(props.projectRoot, props.explicitProject);
      const input = path.resolve(props.input);
      if (
        state.files.has(input) ||
        state.roots.some((root) => isProjectWalkPath(root, input))
      ) {
        return;
      }
      state.files.add(input);
      state.dirty = true;
      flush(base, state);
    },
```

`flush` returns immediately unless `state.dirty`, and `dirty` is set only by adding an out-of-walk file or by declaring volatility. A clean in-walk-only run therefore ends with `dirty === false` and writes nothing.

A project with all inputs inside the walk is the ordinary minimal case, not a contrived one. The bundled typescript-go standard libraries are deliberately excluded as non-filesystem inputs (`packages/ttsc/driver/graph.go`), and a project's sources and tsconfig are inside the walk by definition. A local or source plugin can declare volatility without introducing any out-of-root dependency, so removing that declaration need not produce the incidental out-of-walk input that the control shows would rescue the state.

### Provenance

The mechanism arrived in `fe6c7f1ca` / PR #722, "sound cache invalidation from the reference graph", together with the documented recovery path quoted above. The clear path has a test, but it hand-writes the very document a real clean run never produces: `assertPrepareSnapshotCompactsWorkerFiles` in `tests/test-metro/src/internal/metro-cache.ts` writes `{"files":[recorded],"version":1,"volatile":false}` directly, and `assertCacheKeyFoldsNonceWhileSnapshotVolatile` writes `{"files":[],"version":1,"volatile":true}` directly. The transformer-side test covers the positive marker only (`test_transformer_records_volatile_declarations_in_the_worker_snapshot.ts`). No test asserts that a real clean run produces a clearing observation, which is why the gap between the documented recovery and the implemented recorder was never visible.

#716, #721, and PR #722 are the originating issue and implementation, not a report of this defect. #255 tracks the resident and shared compiler cost model, not this lifecycle. GitHub searches for volatile snapshot lifecycle and Metro cache recovery returned no duplicate.

Verified against `2b724664eee71c57cd688c55122cac137774bf62`; `packages/metro/src` is byte-identical on current `master` (`612abc2d2`).

## Consequence surface

| Run shape | Worker document written today | Effect on the sticky flag |
| --- | --- | --- |
| Plugin declares volatility | Yes, `volatile: true` | Sets it (correct) |
| Clean transform, at least one out-of-walk input | Yes, `volatile: false` | Clears it (correct) |
| Clean transform, every input inside the project walk | No | Never clears |
| No eligible file transformed at all | No | Must not clear; absence of transforms is not evidence |
| Several workers, one declares volatility | Mixed | Any positive must win |
| Several workers, all clean | Yes | Clears |
| Corrupt worker document | Swept, new epoch id | Fail-closed; must stay |
| Unwritable cache directory | None | Degrades to a nonce; must stay |

While the flag is stuck, `computeProjectFingerprint` folds a nonce on every call, so Metro's persistent transform cache shares nothing across runs for that project. Output stays sound; the cost is a full re-transform per build, indefinitely, with no diagnostic pointing at the cause. The user-visible remedies today are deleting `node_modules/.cache/ttsc-metro` or accidentally introducing an out-of-walk input.

## Approach

The invariant: a run that transformed at least one eligible file has an opinion about volatility, and records it, whether or not it also had an out-of-walk input to record. Absence of worker documents means the run observed nothing, which is not evidence that the declaration is gone.

Owner: `createSnapshotRecorder` and `prepareSnapshot` in `packages/metro/src/core/fingerprint.ts`, plus whichever transformer call site knows that an eligible file was transformed.

Everything the existing model guarantees has to survive: any positive worker observation wins the union, writes stay atomic, a corrupt document fails closed and mints a fresh epoch id, the main snapshot identity stays stable across compaction so keys do not churn for an unrelated reason, and concurrent workers never race a shared write.

Note that the observation is per epoch, not per process: writing a clean document unconditionally on every worker start would create write traffic proportional to worker count for no new information, and writing it at the wrong moment (before the transform that would have declared volatility) would clear the flag that the same run is about to set.

Positive: a project whose plugin declares volatility, then has the declaration removed, and whose every input is inside the project walk, produces stable cache keys after one further clean run.

Negative: a run that transforms nothing does not clear the flag; a run in which any worker declares volatility keeps it set even if other workers are clean; a corrupt worker document still fails closed with a fresh epoch.

Boundary: mixed workers with `true` and `false` in the same epoch; a clean run that also has out-of-walk inputs, which is today's working control and must keep working; an unwritable cache directory, which must still degrade to a nonce rather than throw.

Regression: the existing `tests/test-metro/src/features/cache/` suite keeps passing, including `test_cache_key_folds_a_nonce_while_the_snapshot_records_volatile`, `test_cache_key_is_stable_across_runs_for_an_unchanged_project`, `test_prepare_snapshot_compacts_worker_files_into_the_main_snapshot`, and `test_prepare_snapshot_heals_a_corrupt_worker_file_with_a_new_epoch`.

**This defect cannot be proven by a single run.** A one-shot command passes against it, because the stuck state only appears at the third `prepareSnapshot`. The new regression must drive the sequence: prepare, record volatile, prepare (compaction), transform cleanly with in-walk inputs only, prepare, then compare two `computeProjectFingerprint` results and assert they are equal. It must also be written so that it fails at the audited baseline; the hand-written-worker-document shortcut used by the existing clear-path helper is exactly what hid this, so the new case must go through the real recorder.

Narrow proving command, after adding the case under `tests/test-metro/src/features/cache/`:

```text
pnpm --filter @ttsc/test-metro start -- --include=volatile
```

The runner matches test names by substring and prints each executed test by name. A filter matching nothing fails loudly (`No tests matched --include=<...>`, exit 1), but this filter cannot match nothing: at the baseline it already selects `test_cache_key_folds_a_nonce_while_the_snapshot_records_volatile` and `test_transformer_records_volatile_declarations_in_the_worker_snapshot`, and both pass. That is the hazard here, and it is the same shape as the defect: a green result that never exercised the case in question. Read the printed names rather than the exit code, and confirm the new case fails at the baseline.

Broader: `pnpm --filter @ttsc/test-metro start`, then `pnpm test`.

## Coordination

- PR #722 (`fe6c7f1ca`) is the causal history and its guarantees are the constraint list above, not obstacles.
- This issue edits `packages/metro/src/core/fingerprint.ts` and possibly the transformer call site in `packages/metro/src/`. It shares that file with the issue that requires freshness inputs to include missing module-resolution candidates, which also touches `packages/metro/src/core/fingerprint.ts` and `packages/unplugin/src/core/transform.ts`. Sequence the two under one owner.
- The Windows path-identity issue touches `packages/unplugin/src/core/transform.ts` only and is disjoint from this file, but it belongs to the same bundler-freshness component; if one owner takes all three, land this one first, since it is the smallest and its guarantees are the ones the others must not break.

### LINKED ISSUE #831: fix(unplugin): compare compiler and bundler paths by filesystem identity on Windows
## Problem

`packages/unplugin/src/core/transform.ts` joins two path vocabularies. One comes from the bundler, as a module id resolved with `path.resolve`. The other comes from the compiler envelope, keyed relative to a `projectRoot` that the plugin derives itself. Every place the two must be compared uses raw JavaScript string equality, which is case-sensitive even on a case-insensitive Windows volume.

The plugin has a fast path that is not case-sensitive and a set of slow paths that are. `toProjectKey` runs `path.relative`, which on Windows compares path segments case-insensitively, so the fast lookup survives a spelling difference. Every fallback and every membership, deduplication, self-exclusion, and reachability decision compares absolute strings directly, so those do not.

The consequence is not one message. When the fast path misses for an unrelated reason and the slow path is asked to recover, it cannot; `selectTransformedSource` throws `ttsc transform did not return output for <file>` after a successful compile, and the caller evicts the generation. The membership checks are worse than a hard failure: `declaresFile` silently answers "not volatile" and "dependencies not complete", `selectFileDependencies` silently returns an empty dependency list, and `selectReachableEdges` silently finds no reachable edges. Those are all sound-looking answers that change caching and watch behavior without any error.

### The identity mismatch, reproduced

On Windows, `path.resolve` preserves whatever drive-letter case it was given, and both spellings address the same existing file. `path.relative` does not care:

```js
const upper = "D:\\github\\samchon\\ttsc\\packages\\unplugin\\src\\index.ts";
const lower = "d:\\github\\samchon\\ttsc\\packages\\unplugin\\src\\index.ts";
const toProjectKey = (root, file) => path.relative(root, file).replace(/\\/g, "/");
```

```text
{
 "bothExist": [ true, true ],
 "resolved": [ "D:\\github\\...\\src\\index.ts", "d:\\github\\...\\src\\index.ts" ],
 "rawStringEqual": false,
 "toProjectKey_sameCase": "src/index.ts",
 "toProjectKey_mixedCase": "src/index.ts",
 "toProjectKeyAgrees": true
}
```

That asymmetry is the whole defect: the one comparison that is case-insensitive is the fast path, and every comparison written to back it up is not.

### How the two spellings legitimately differ

The plugin derives the project root from the tsconfig it resolved, and resolves an explicit `project` option against `process.cwd()`:

```ts
function resolveTsconfig(file: string, tsconfig?: string): string {
  if (tsconfig !== undefined) {
    return path.isAbsolute(tsconfig)
      ? tsconfig
      : path.resolve(process.cwd(), tsconfig);
  }
  // ...walk upward from path.dirname(file)
  return path.resolve(process.cwd(), "tsconfig.json");
}
```

`projectRoot` is `path.dirname(tsconfig)`. The file side is `path.resolve(stripQuery(id))`, where `id` is whatever the bundler handed over. Neither side is canonicalized, so any drive-letter-case difference between the bundler's module ids and the cwd-derived or user-typed config path survives into every comparison below. The explicit-`project` configuration is supported and tested (`tests/test-unplugin/src/features/transform/test_transformttsc_resolves_a_relative_project_option_from_cwd.ts`), which is what makes the cwd-derived root a realistic origin for the mismatch rather than a hypothetical.

A probe that upper-cases an entire path is *not* the evidence for this issue: no bundler produces that. The drive-letter case difference is the reachable trigger, and it is the one the fix must be proven against.

### The comparisons that break

All in `packages/unplugin/src/core/transform.ts`:

```ts
  // Fast path: the compiler key matches the normalised project-relative path.
  const key = toProjectKey(props.projectRoot, props.file);
  const direct = props.result.typescript[key];
  if (direct !== undefined) return direct;
  // Slow path: resolve each candidate to an absolute path for comparison.
  for (const [candidate, source] of Object.entries(props.result.typescript)) {
    if (path.resolve(props.projectRoot, candidate) === props.file) return source;
  }
  throw new Error(`ttsc transform did not return output for ${props.file}`);
```

```ts
  return listed.some(
    (entry) =>
      typeof entry === "string" &&
      entry.length !== 0 &&
      path.resolve(props.projectRoot, entry) === props.file,
  );
```

`selectFileDependencies` repeats the same scan and then uses raw absolute strings for self-exclusion and deduplication; `selectReachableEdges` keys its edge map and visited set by raw absolute spelling; the watch and external-input merging paths and the project snapshot key computation deduplicate and exclude by raw spelling as well. A repair confined to `selectTransformedSource` would convert a loud failure into a silent loss of dependency closure, volatility, or cache validity.

### Provenance

The pipeline arrived in `2c748c24e` / PR #47. `4edb6b56f` / PR #100 ("style: repo-wide source-code quality pass") added the comment that still stands above the fallback, describing it as the path that resolves candidates "for comparison", without changing the comparison itself. No later commit addresses path identity. Open and closed searches for unplugin Windows casing, path case, and case-insensitive path comparison returned no issue or pull request.

Verified against `2b724664eee71c57cd688c55122cac137774bf62`; `packages/unplugin/src` is byte-identical on current `master` (`612abc2d2`).

## Consequence surface

| Site | Comparison | Failure mode when spellings differ |
| --- | --- | --- |
| `selectTransformedSource` fast path | `path.relative` | Survives (case-insensitive on Windows) |
| `selectTransformedSource` fallback | Raw absolute string | Throws `ttsc transform did not return output for …` after a successful compile; caller evicts the generation |
| `declaresFile` | Raw absolute string | Silently reports not volatile, not dependency-complete |
| `selectFileDependencies` | Raw absolute string, plus self-exclusion and dedup | Silently returns no dependencies; a file can also fail to exclude itself |
| `selectReachableEdges` | Raw absolute string keys and visited set | Reference graph unreachable from the bundler id; silently no edges |
| Watch and external-input merging | Raw absolute string | Duplicate or missing watch entries |
| Project snapshot keys | Raw absolute string | Cache misses or, in the other direction, two spellings treated as two files |
| POSIX, two genuinely case-distinct files | Raw absolute string | Correct today, and must stay correct |

Platform matters here: on Windows and on case-insensitive macOS volumes two spellings are one file; on Linux they are two files. A fix that lower-cases unconditionally would merge distinct POSIX files, which is a worse defect than the one being repaired.

## Approach

The invariant: two paths that name the same filesystem entry compare equal everywhere in this pipeline, and two paths that name different entries never do.

Owner: one path-identity helper in `packages/unplugin/src/core/transform.ts`, used for every equality test, map and set key, membership check, self-exclusion, deduplication, envelope lookup, reachability walk, input hashing, and temporary-file exclusion in the transform, cache, and watch paths. The original spelling must still be what is handed to filesystem reads and back to the bundler, so the normalized form is a comparison key, not a replacement value.

Case folding applies only where the host filesystem contract is case-insensitive. POSIX must keep treating case-distinct paths as distinct entries.

Whether the helper also normalizes separators, UNC prefixes, or trailing separators is an implementation choice; what the fix must not do is normalize the one site whose failure is loud and leave the sites whose failure is silent.

No new dependency, and no fixture-specific or consumer-specific special case.

Positive, on Windows: a module id whose drive letter case differs from the project root's resolves its transformed output through both the fast and the fallback path; `declaresFile` finds it in `volatile` and `dependenciesComplete`; `selectFileDependencies` returns its dependencies and excludes itself; `selectReachableEdges` reaches its edges; watch and external inputs deduplicate to one entry; the project snapshot key is the same for both spellings.

Negative, on POSIX: two files whose paths differ only by case remain two files, with distinct outputs, distinct dependencies, distinct edges, and distinct snapshot keys.

Boundary: an explicit relative `project` option resolved from a differently-cased cwd, which is the reachable trigger; a UNC path; a path with a trailing separator; a bundler id carrying a query string or hash fragment, which `stripQuery` must still remove before normalization.

Regression: the existing `tests/test-unplugin/src/features/transform/` suite keeps passing, in particular the project-cache invalidation cases (`test_transformttsc_invalidates_project_cache_when_source_changes`, `…_when_an_external_input_changes`, `…_through_an_external_graph_edge`), the watch-input cases (`test_transformttsc_narrows_watch_inputs_for_a_file_declared_complete`, `test_transformttsc_excludes_the_generated_tsconfig_from_watch_files`), and `test_transformttsc_resolves_a_relative_project_option_from_cwd`.

The new regressions must be platform-aware rather than skipped off Windows: the case-insensitive assertions run where the host filesystem is case-insensitive, and the case-distinct POSIX twin runs where it is not. A test that only ever runs on one platform proves half of this invariant, and the half it skips is the one that a naive fix breaks.

Narrow proving command, after adding the cases under `tests/test-unplugin/src/features/transform/`:

```text
pnpm --filter @ttsc/test-unplugin start -- --include=<new case name>
```

The runner matches test names by substring and prints each executed test by name. A filter matching nothing fails loudly (`No tests matched --include=<...>`, exit 1), so a typo cannot pass silently. Read the printed names anyway, because a substring that also matches an existing case runs green without executing what you added. Confirm each new case fails at the baseline on its own platform: a case that is skipped on the running platform prints nothing and cannot fail, which for this issue is the difference between proving half the invariant and proving none of it.

Broader: `pnpm --filter @ttsc/test-unplugin start`, then `pnpm test`, and note that the repository's Windows and Linux CI lanes are both load-bearing for this issue, since neither platform alone exercises both halves.

## Coordination

- This issue edits `packages/unplugin/src/core/transform.ts`. It shares that file with the issue requiring freshness inputs to include missing module-resolution candidates, which changes what goes into `selectExternalInputPaths`, `selectWatchInputs`, and `matchesCachedSource`. Sequence the two under one owner; landing path identity first gives that issue a single comparison vocabulary to extend.
- The Metro volatile-snapshot issue touches `packages/metro/src/core/fingerprint.ts` and does not edit this file, but it consumes `collectProjectInputHashes` and `isProjectWalkPath` from this package's public API, so any change to how those key their entries has to be checked against it.
- `2c748c24e` and `4edb6b56f` are causal history, not duplicates. No open issue or pull request touches `packages/unplugin`.

### LINKED ISSUE #832: fix(ttsc,unplugin,metro): track the module-resolution candidates that would supersede the current winner
## Problem

Three long-lived caches decide whether a previous compile is still valid by re-checking a recorded set of filesystem inputs. All three record the module target the compiler already selected, and none records the candidate path that does not exist yet but would outrank that target if it appeared.

Module resolution is a precedence search. When a specifier resolves, the compiler picks the first candidate that exists in a defined order: extension priority for a relative specifier, the declared order of a `paths` substitution list, `rootDirs` alternatives, the nearest `node_modules` up the directory chain, and `exports` / `imports` conditions. Creating a file at a higher-priority candidate changes which module the specifier means, without editing the importer, the tsconfig, the root set, or any file that was recorded.

None of those is a recorded input, so nothing in the recorded set moves and every cache answers "nothing changed":

- A resident `ttscgraph serve` session keeps attesting the previous snapshot, so the Graph tools answer from a program that resolved a specifier to a file the compiler would no longer choose.
- The `@ttsc/unplugin` project cache replays the previous whole-project transform.
- The adapter watch inputs and Metro's cross-run fingerprint carry the old target, so nothing schedules the importing module for a re-transform, in-process or across runs.

Affected: every `@ttsc/graph` MCP session (which is the long-lived shape by design), every bundler using `@ttsc/unplugin` in watch or long-lived mode, and every `@ttsc/metro` project across runs. The consequence is a stale answer or a stale build output that no user action short of restarting the process or clearing the cache corrects.

### Reproduction, resident Graph session

Driven against the baseline binary, `ttscgraph 0.19.3 (commit 2b724664eee7, built 2026-07-19T05:46:42.932Z, windows/amd64, go go1.26.4)`.

Project: `tsconfig.json` with `allowJs: true` and `files: ["src/main.ts"]`; `src/main.ts` containing `import { value } from "./foo";`; and only `src/foo.js` on disk. One `ttscgraph serve` child, three requests over its newline-delimited protocol, with `src/foo.ts` created between the first and the second. Nothing else is touched.

```text
resident 1: {"mode":"initial","changed":true,"dump":"present","files":{"topLevelKeys":["project","tsconfig","provenance","diagnostics","nodes","edges"],"paths":["src/foo.js","src/main.ts"]}}
resident 2 (after src/foo.ts created): {"mode":"unchanged","changed":false,"dump":"absent"}
resident 3: {"mode":"unchanged","changed":false,"dump":"absent"}
fresh one-shot dump: {"files":{"topLevelKeys":["project","tsconfig","provenance","diagnostics","nodes","edges"],"paths":["src/foo.ts","src/main.ts"]},"hasFreshOnlyInTs":true}
```

`paths` is the probe's own projection of the distinct `file` values across the dump's `nodes`; `topLevelKeys` is printed to show the projection was taken from a real dump. The session answers `unchanged` twice. A fresh one-shot `ttscgraph dump` of the same directory selects `src/foo.ts` and contains its new symbol, so the compiler's own resolution did change; only the session's freshness model failed to see it. Because Graph responses tell callers that every returned fact matches the synchronized snapshot, this is a wrong answer, not delayed indexing.

### Root cause, resident session

`auxiliaryInputs` in `packages/ttsc/cmd/ttscgraph/serve.go` contributes resolution candidates only for specifiers that failed to resolve:

```go
    for _, specifier := range sourceModuleSpecifiers(source) {
      resolved := program.TSProgram.GetResolvedModuleFromModuleSpecifier(source, specifier)
      if resolved != nil && resolved.IsResolved() {
        continue
      }
      inputs = append(inputs, unresolvedModuleCandidates(configs, directory, cwd, specifier.Text())...)
    }
```

`graphSession.Snapshot` then compares config hashes, those recorded auxiliary states, the root set, and the hashes of files already in the program. A newly created higher-priority target is in none of those sets, so the snapshot is reported unchanged. The root-set check helps only when the new file is itself a literal `files` entry.

### Root cause, unplugin and Metro

`matchesCachedSource` in `packages/unplugin/src/core/transform.ts` re-hashes exactly the recorded key universe, and states the premise that makes that sufficient:

```ts
  // Re-hash the out-of-walk inputs the compiler reported for this generation
  // over exactly the recorded key universe, so an edit to a `node_modules`
  // declaration or a monorepo sibling source invalidates the entry even in a
  // host that never clears the cache between builds. A new out-of-walk input
  // cannot appear without some recorded input changing first: a new reference
  // edge requires editing an in-walk source, and a new global or config file
  // requires a tsconfig or package manifest change, both of which the project
  // walk above already detects.
```

The resident run above is a counterexample to the sentence beginning "A new out-of-walk input cannot appear": nothing in the walk changed, no recorded input changed, and the correct resolution changed. The premise holds for reference edges and config files; it does not hold for resolution precedence, because precedence is decided by paths that do not exist.

`selectExternalInputPaths` derives that recorded set from realized envelope members only, `graph.edges` sources and targets, `graph.globals`, `graph.configs`, and plugin `dependencies`. `selectWatchInputs` registers realized reachability by the same rule. `computeProjectFingerprint` in `packages/metro/src/core/fingerprint.ts` persists and re-hashes exactly that recorded set across runs:

```ts
      hash.update(`snapshot:${snapshot.id}`);
      hash.update(stableStringify(collectExternalInputHashes(snapshot.files)));
```

Every one of those sets is a list of files the compiler chose. None of them can contain a file that does not exist.

The transform-cache half was additionally exercised with a project whose `@generated/value` mapped to two ordered out-of-root `paths` targets, `../external/generated` then `../external/fallback`. Exact one-shot compiler dumps selected `fallback.ts` while the first target was absent, and `generated.ts` immediately after that file was created. A transform-envelope probe primed one cache generation whose realized graph contained only `fallback.ts`, created the higher-priority file, and called `transformTtsc` again with the same source: the compiler invocation counter stayed at `1` and both calls returned generation `1`.

### History: two repairs, both at this seam, both claiming this class

The resident freshness model was introduced by PR #351 (`5cfd477e4`), whose purpose was exactly candidate tracking. Its commit message says it "seal[s] freshness blind spots for vanished and not-yet-created inputs" and enumerates three of them: a source that vanished between load and capture, "a literal `files` entry that does not exist yet", and config `types` packages, which "now contribute the same candidates as type directives".

Its regression suite is the strongest evidence of what that sealing covered:

| Test | Starting state | Transition covered |
| --- | --- | --- |
| `serve_session_reloads_newly_resolved_path_alias_test.go` | Specifier unresolved | unresolved → resolved |
| `serve_session_reloads_newly_resolved_root_dirs_target_test.go` | Specifier unresolved | unresolved → resolved |
| `serve_session_reloads_newly_resolved_package_export_test.go` | Specifier unresolved | unresolved → resolved |
| `serve_session_reloads_newly_resolved_package_import_test.go` | Specifier unresolved | unresolved → resolved |
| `serve_session_reloads_newly_resolved_conditional_export_test.go` | Specifier unresolved | unresolved → resolved |
| `serve_session_reloads_newly_resolved_excluded_import_test.go` | Specifier unresolved | unresolved → resolved |
| `serve_session_reloads_newly_resolved_reference_path_test.go` | Reference unresolved | unresolved → resolved |
| `serve_session_reloads_newly_resolved_type_reference_test.go` | Directive unresolved | unresolved → resolved |
| `serve_session_reloads_newly_resolved_config_types_test.go` | Types package absent | unresolved → resolved |
| `serve_session_keeps_missing_literal_root_unchanged_test.go` | Root absent | negative twin |

Every resolution mechanism in the precedence class already has a test, and every one of them starts from a specifier that resolves to nothing. `TestServeSessionReloadsNewlyResolvedRootDirsTarget` is representative: `./template` has no target at all until the test creates one. Not one case starts from a specifier that already resolves and then creates a better candidate. The `continue` quoted above is the exact line that makes that direction unreachable, and it is the line the coverage was written around.

The second repair is #716 / #721 / PR #722 (`fe6c7f1ca`), which established realized reference-graph invalidation for the transform cache and described the result as sound, writing the "cannot appear without some recorded input changing first" premise into production code. Its tests edit or delete already recorded external inputs; none creates an unrecorded alternative that changes resolution.

So the same root premise, that recording the selected target is enough, survived two repairs whose stated purpose was freshness completeness, and reaches three consumers. Fixing only the `.js` → `.ts` witness would leave the other four mechanisms in the table above in exactly the state PR #351 left them.

No open or closed issue or pull request tracks it. Searches covered `ttscgraph`, resident stale resolution, module-resolution cache, higher-priority candidates, external-input cache, `paths` fallback, and missing candidates.

Verified against `2b724664eee71c57cd688c55122cac137774bf62`. `packages/ttsc/cmd/ttscgraph/serve.go`, `packages/unplugin/src`, and `packages/metro/src` are byte-identical on current `master` (`612abc2d2`).

## Consequence surface

The full precedence class, all reachable through the same premise:

| Mechanism | Currently selected | Appears later and should win | Currently observed |
| --- | --- | --- | --- |
| Relative extension priority | `./foo` → `foo.js` | `foo.ts`, `foo.tsx`, `foo.d.ts` | unchanged |
| Ordered `paths` substitutions | second entry | first entry | unchanged |
| `rootDirs` alternatives | one virtual root | a higher-priority root | unchanged |
| `node_modules` proximity | ancestor package | nearer package | unchanged |
| Package `exports` / `imports` | one condition target | a higher-priority target | unchanged |
| Directory index versus file | `foo/index.ts` | `foo.ts` | unchanged |
| Total resolution failure → success | nothing | any candidate | correctly reloads (PR #351) |
| Recorded winner edited or deleted | recorded | itself | correctly invalidates (PR #722) |
| An irrelevant unrelated file appears | recorded | nothing | must stay unchanged |

Consumers, by owner:

| Owner | Surface | Stale effect |
| --- | --- | --- |
| `packages/ttsc/cmd/ttscgraph/serve.go` | Resident Graph session snapshot | `mode=unchanged`; every Graph tool answers from the superseded program |
| `packages/unplugin/src/core/transform.ts` | Project transform cache (`matchesCachedSource`, `selectExternalInputPaths`) | Whole-project transform replayed against superseded resolution |
| `packages/unplugin/src/core/transform.ts` | Adapter watch inputs (`selectWatchInputs`) | Creating the winning candidate schedules nothing |
| `packages/metro/src/core/fingerprint.ts` | Cross-run cache key | Old key reused in a later Metro run |

Process-lifetime consumers (Metro, Bun, Turbopack) hold the stale generation for the life of the process. The unplugin cache is additionally stale whenever the new winner is outside `collectProjectInputHashes`'s walk, which is the normal case for `node_modules` and out-of-root `paths` targets.

## Approach

The invariant: a freshness input set for a resolved specifier contains not only the chosen target but every candidate path that, if it appeared, the compiler would choose instead.

Owners, in order of dependency:

1. `packages/ttsc/cmd/ttscgraph/serve.go`, `auxiliaryInputs` and the session snapshot comparison. Candidate enumeration is compiler-owned and host-owned; it stays internal and is not a protocol change.
2. `packages/unplugin/src/core/transform.ts`, both the project-cache validity set and the watch-input set. If the enumeration lives on the compiler side, this is the transport of an additional envelope member rather than a re-implementation of resolution in TypeScript.
3. `packages/metro/src/core/fingerprint.ts`, which persists whatever set the transform layer records. It should need no independent notion of a candidate.

The complete specifier surface is what `sourceModuleSpecifiers` already recognizes: imports, exports, import-equals, import types, `require`, and dynamic import, across every resolution mechanism the compiler supports.

Cost is a real constraint. Superseding candidates for every specifier in a large program is a large set, and the fix must decide how it is bounded (for example, only candidates strictly above the selected one in the precedence order, rather than the whole search space) and must not turn every snapshot into a full resolution replay. That decision belongs in the fix, with the bound stated.

No polling, no cache-disable escape hatch, no fixture-name or package-name special case.

`packages/graph/src/structures/ITtscGraphApplication.ts`, every subordinate public Graph type, and all of their JSDoc must remain byte-for-byte unchanged.

Positive, one case per row of the precedence table, each keeping the importer and the tsconfig byte-identical and creating only the superseding candidate between two requests:

| Case | Expected after the candidate appears |
| --- | --- |
| `.js` selected, sibling `.ts` created | Resident session reports changed; the dump contains the `.ts` file's symbols |
| Second `paths` entry selected, first created | Transform cache misses; the new target is in the program |
| `rootDirs` lower-priority selected, higher created | Reload |
| Ancestor `node_modules` package selected, nearer created | Reload |
| `exports` condition target superseded | Reload |
| Same, through the unplugin cache | A second `transformTtsc` call compiles again rather than replaying |
| Same, through watch inputs | Creating the candidate schedules the importing module |
| Same, across two Metro runs | The cache key differs between runs |

Negative twins, which must not regress into over-invalidation: an unrelated file appears somewhere in or out of the walk and nothing invalidates; a candidate below the selected one in precedence appears and nothing invalidates; the existing winner alone is edited or deleted and the current behavior is preserved; a specifier that was already unresolved becomes resolved and still reloads, per PR #351's suite.

Boundary: a specifier whose candidate list is empty; a candidate path the OS cannot parse, which `5cfd477e4` already handles by recording it as absent and which must keep working; a candidate created and removed again between two requests.

**No single-run command proves this.** Every case is a two-observation sequence against one long-lived consumer, and a single request or a single build passes against the defect. Each regression must: take the first observation, create exactly one file, take the second observation, and assert the second differs. Running the same command twice from the shell is not equivalent for the resident session or for the in-process caches, because a fresh process resolves correctly, which is precisely the control in the reproduction above.

Narrow proving commands, per owner, after adding the cases:

```text
cd packages/ttsc && go test -count=1 ./cmd/ttscgraph/ -run TestServeSession
pnpm --filter @ttsc/test-unplugin start -- --include=<new case name>
pnpm --filter @ttsc/test-metro start -- --include=<new case name>
```

`packages/ttsc/cmd/ttscgraph` contains 30 `_test.go` files today, so the Go command runs real tests, and `TestServeSession` matches more than twenty of them. Two Go-side traps still apply if that command is narrowed: `-count=1` is required because the test cache otherwise reports a pass without executing anything, and a `-run` pattern that matches nothing exits **0**, printing `ok … [no tests to run]`. Add `-v` when running a single new case and confirm its name appears. The two `pnpm` runners match test names by substring, print each executed test by name, and fail loudly when a filter matches nothing (`No tests matched --include=<...>`, exit 1). Read the printed names anyway: a substring that also matches an existing case runs green without executing what you added, and this issue's whole subject is a green result produced by a check that never looked at the thing that changed.

Broader: `node scripts/test-go-graph.cjs` (which runs `go test -count=1 ./internal/graph/... ./cmd/ttscgraph/... ./cmd/graphbench/...`), `pnpm --filter @ttsc/test-unplugin start`, `pnpm --filter @ttsc/test-metro start`, then `pnpm test`.

## Coordination

- PR #351 (`5cfd477e4`) and #716 / #721 / PR #722 (`fe6c7f1ca`) are the causal history. Both are closed and neither's acceptance covers the resolved-to-different-resolved transition, so this is residue rather than a duplicate. Their existing guarantees, including the OS-invalid-candidate tolerance and the realized-input hashing, must keep holding.
- **Three owners, one issue, deliberately.** This is not a bundling convenience and it should not be split during implementation. The three surfaces share one premise, so splitting it invites each owner to fix the witness in front of them and leave the class, which is precisely the pattern the history above documents: a repair shaped to the symptom that surfaced, with the general rule left unwritten. Implement it as one sequenced change: the native session first, then the unplugin project cache and watch inputs, then Metro. A partial landing that fixes the `.js` → `.ts` witness and leaves ordered `paths`, `rootDirs`, `node_modules` proximity, or `exports` / `imports` is not this issue closed.
- Shares `packages/unplugin/src/core/transform.ts` with the Windows path-identity issue and `packages/metro/src/core/fingerprint.ts` with the volatile-snapshot issue. Sequence all three under one owner; land the two smaller ones first so this issue extends a settled comparison vocabulary and a settled snapshot lifecycle.
- Protected Graph contract: `ITtscGraphApplication` and every subordinate public type and JSDoc stay byte-identical.

samchon:
## Intent

Resolve the ordered bundler-freshness batch: #830, #831, then #832.

## Scope

- let a Metro worker record a clean observation so a removed volatile declaration can clear the next snapshot;
- compare unplugin compiler-envelope and bundler paths by filesystem identity, preserving distinct paths on case-sensitive filesystems;
- record module-resolution candidates that precede a selected target, then consume them in the resident graph session, unplugin cache and watch inputs, and Metro's cross-run fingerprint.

## Constraints

`packages/graph/src/structures/ITtscGraphApplication.ts` and its subordinate public Graph structures remain byte-identical. The source changes follow the dependency order above in this one owner batch.

## Verification

- completed a final solo Self-Review over the full `origin/master...HEAD` surface;
- `pnpm --filter ttsc shim:audit` passed;
- `go test -count=1 ./cmd/ttscgraph -run TestServeSession -v` passed;
- the complete Metro suite passed;
- `go test -count=1 ./test/driver -run '^TestTransformGraphReportsSupersedingModuleCandidates$' -v` passed;
- `pnpm --filter @ttsc/test-unplugin start -- --include=transformttsc_tracks_superseding_resolution_candidates` passed;
- `pnpm --filter @ttsc/test-unplugin start -- --include=transformttsc_uses_filesystem_path_identity` passed.

A complete unplugin suite was also retried. Its two failures are the pre-existing `transformttsc_alias_overlay_*` fixtures, which expect a non-built `@ttsc/banner/lib/index.js`; they are outside this batch's paths. Every Actions run triggered by this branch is cancelled only by its exact commit SHA, with final records required before merge.

Closes #830
Closes #831
Closes #832


samchon:
Solo self-review found one candidate-enumeration overreach before finalizing the #832 snapshot: `compilerOptions.baseUrl` is not a TypeScript-Go fallback after `paths` in the pinned resolver, yet the host probe list records it. A file created only at that path can therefore reload an unchanged session even though it cannot change resolution. I will remove that probe, add the negative resident-session regression, and restart the full review and verification round.

samchon:
The restarted review found a second boundedness correction before final verification: the generic no-extension probe list included `.mts` and `.cts` ahead of `.d.ts`, although the pinned commonjs resolver only probes `.ts`, `.tsx`, `.d.ts`, `.js`, and `.jsx` in that branch. I will split probe suffixes by explicit module extension and extend the lower-priority resident-session regression so an irrelevant `.mts` creation cannot reload the session.

samchon:
Self-review finding: `packageManifestCandidates` traverses every conditional export/import branch, including conditions that the current TypeScript-Go resolution mode cannot select. That wrongly turns a newly created inactive target (for example `browser` without the `browser` custom condition) into a graph reload. I will pass the compiler-selected usage mode and options into candidate enumeration, retain only the active TypeScript-Go conditions plus `default`, and add a session regression proving an inactive conditional target remains unchanged while the active predecessor still reloads.

samchon:
Follow-up to the same review: the existing list also applies CommonJS folder probes (`package.json` and `index.*`) to ESM relative and `paths` resolutions, even though TypeScript-Go stops after file lookup in ESM mode. The mode-aware candidate context will suppress those folder probes, while retaining them for CommonJS and type-reference lookup.

samchon:
Additional self-review finding: package-entry fields (`types`, `typings`, `main`) are resolved through a different TypeScript-Go branch from export-map targets. In an ESM importer, a CommonJS package's extensionless `main` may still perform CJS file/directory probes; `types`/`typings` are also absent when `noDtsResolution` is set. I will preserve that package-entry context and add a focused regression instead of applying the ESM export-target rule to every manifest target.

samchon:
Full `TestServeSession` exposed a test-only overassertion in the existing exports-blocked negative twin: creating the higher-priority export target correctly returns `reload`, but that fixture's extra JavaScript export is not necessarily materialized as a dump node. The regression's contract is the resident-session state transition, so I will remove only the unrelated node-shape assertion and retain the reload/changed checks.

samchon:
Self-review finding: the new unplugin cache regressions replace \TtscCompiler.prototype.transform\ to manufacture graph envelopes. That is test-only monkey-patching, which conflicts with the repository development rule and does not prove the native graph-to-adapter protocol boundary. I will replace these helpers with real compiler-backed project fixtures that exercise the emitted candidate graph, cache invalidation, watch registration, and host-specific path identity through the normal transform path, then restart the complete solo review.

samchon:
Self-review finding: the candidate list is a prefix of resolution probes, not exclusively absent files. For example, an existing package.json or directory can precede an eventual index-file winner and changing it can alter resolution. Several new public comments/docs call every candidate missing, which would understate the envelope contract. I will change that wording to resolution candidates/probes while retaining the precise statement that an absent candidate is tracked until it appears, then restart the complete review.

Issue #855, opened 2026-07-20
fix(ttsc): honor `--noEmit`, `check`, `--emit=false`, and tsconfig `noEmit` in single-file mode


samchon:
## Problem

`ttsc --noEmit src/main.ts` writes `dist/main.js`. So does `ttsc check src/main.ts`, `ttsc --emit=false src/main.ts`, and plain `ttsc src/main.ts` in a project whose `tsconfig.json` sets `"noEmit": true`. The control on the same project, `ttsc --noEmit` with no file argument, correctly writes nothing.

A flag that means "write nothing" writes files. The single-file lane never consults the emit decision the launcher already computed, and the layer beneath it forces emit on.

The affected shape is one the tool itself advertises. `ttsc --help` says "Any other flag is forwarded to tsgo as-is, so tsgo compiler options such as `--strict` or `--target` work directly (e.g. `ttsc --strict file.ts`)", and `ttsc --noEmit` is presented as the CI type-check gate in `website/src/content/docs/ttsc/compile.mdx`, `setup/ttsc.mdx`, and `execute.mdx`. A CI step written as `ttsc --noEmit src/index.ts` mutates the checkout.

Expected behavior: when emit is off, the single-file lane type-checks the file, writes nothing, and prints no emitted-file path, exactly as project mode does.

### Read this before judging the issue covered

`tests/test-ttsc/src/features/compiler/test_ttsc_no_plugin_build_honors_tsconfig_no_emit.ts` exists, is named as though it owns this behavior, and passes. It does not cover it. It runs `ttsc` on a `noEmit: true` project with **no file argument**, so it exercises the project lane, which is the lane that is already correct. It passes today, it will pass unchanged after this fix, and it would pass just as well if the single-file lane were made worse.

Anyone checking "is this already covered?" or "is this a duplicate?" will find that test first. It is the reason the defect survived: the coverage it appears to provide is coverage of the other lane. Adding a file argument to its own fixture is what exposes the write.

## Evidence

`packages/ttsc/src/launcher/internal/runTtsc.ts` and `packages/ttsc/src/compiler/internal/runSingleFileEmit.ts` are byte-identical between the audited commit `2b724664eee71c57cd688c55122cac137774bf62` and `master` at `612abc2d20d3f99429319081c1f6520bbe34ac7a`.

### Reproduction with observed output

Project A: `app.ts` plus a `tsconfig.json` with `outDir: "bin"`.

```text
[...]

Issue #852, opened 2026-07-20
fix(ttsc): make `--verbose` print the build summary and emitted files on the plugin-free lane it advertises


samchon:
## Problem

`ttsc --verbose` prints nothing on a project that declares no ttsc plugin. Not a summary, not an emitted-file list, not a byte on either stream. The build succeeds and the files are written; the flag has no observable effect.

That is the shape the guide's own walkthrough describes. `website/src/content/docs/ttsc/compile.mdx` documents it twice and unconditionally:

```text
npx ttsc --verbose                  # print every emitted file path
```

and, in the flags table, "`--verbose` | Print build summary and emitted file paths." `ttsc --help` repeats the promise: "`--verbose`  Print the build summary and emitted files".

The flag does work on the native-host lane, which runs only when the project has at least one native plugin. A documented flag whose behavior depends on an internal lane choice the user cannot see is a broken contract for the common case.

Affected: every `ttsc`, `ttsc build`, and `ttsc check` invocation with `--verbose` on a project without ttsc plugins.

## Evidence

`packages/ttsc/src/flags/schema.ts`, `packages/ttsc/src/launcher/internal/runTtsc.ts`, `packages/ttsc/src/compiler/internal/runBuild.ts`, and `packages/ttsc/cmd/ttsc/build.go` are byte-identical between the audited commit `2b724664eee71c57cd688c55122cac137774bf62` and `master` at `612abc2d20d3f99429319081c1f6520bbe34ac7a`.

### Reproduction with observed output

Probe project: `tsconfig.json` with `rootDir: "src"` and `outDir: "bin"`, three `.ts` sources, no `compilerOptions.plugins`.

```text
$ node packages/ttsc/lib/launcher/ttsc.js --verbose > v.out 2> v.err
exit=0
stdout bytes=0 stderr bytes=0
$ ls bin
argv.js  boom.js  index.js
```

Three files emitted, zero bytes printed. The same holds for `ttsc build --verbose`, `ttsc check --verbose`, `ttsc --verbose --emit`, and `ttsc --quiet=false`: all exit 0, all silent.

### Root cause

The launcher parses the flag and folds it into `quiet`:

```ts
const verbose = getBoolean(result, "--verbose");
const quietFlag = getBoolean(result, "--quiet");
const quiet = verbose === true ? false : (quietFlag ?? true);
```

[...]

PR #887, opened 2026-07-20
fix(ttsc): reconcile watch inputs from compiler topology

### LINKED ISSUE #817: fix(ttsc): watch the compiler's resolved input topology instead of a one-time directory snapshot
## Problem

`ttsc --watch` advertises "rebuild on save" in `README.md` and "Rebuild when project files change" in its own `--help`. The launcher does not watch the compiler's inputs. It takes one recursive snapshot of every directory below the selected tsconfig's directory, drops four guessed entry names (`node_modules`, `.git`, `lib`, `dist`), opens one non-recursive `fs.watch` per remaining directory, and never reconciles that set again.

One frozen filesystem guess produces two opposite failures:

1. **Missed inputs.** A directory created after startup is never watched, so later edits inside it are silent. Inputs outside the tsconfig directory (extended configs, project references, loaded dependency sources) are never watched at all. Authored sources under a directory literally named `lib` or `dist` are excluded even when the compiler includes them.
2. **Self-triggered rebuilds.** A configured output directory whose basename is not exactly `lib` or `dist` is watched. Watchers are installed before the first build, so the build's own emit produces events, each event schedules another build, and each rebuild emits again. An ordinary `outDir: "build"` or `"out"` can make an idle watch session rebuild continuously.

Expected behavior is compiler-owned invalidation: rebuild for the program's current inputs, update the watched set when the program's topology changes, and never self-trigger on the compiler's own configured outputs.

Affected: `ttsc --watch`, `ttsc build --watch`, and `ttsc check --watch`, including single-file invocations, which enter the same watcher.

## Evidence

Every anchor below is unchanged between the audited commit `2b724664eee71c57cd688c55122cac137774bf62` and `master` at `612abc2d20d3f99429319081c1f6520bbe34ac7a`.

### Verified control flow

`packages/ttsc/src/launcher/internal/runTtsc.ts`:

- `runCompatibleBuild` intercepts `--watch` and calls `runWatch` before any other lane dispatch. The flag is launcher-owned in `packages/ttsc/src/flags/schema.ts`, so TypeScript-Go's own watch implementation is never involved.
- `runWatch` resolves `root` to `path.dirname(resolveProjectConfig(...))`, calls `collectWatchDirectories(root)` exactly once, and maps that frozen array through `fs.watch(dir, { persistent: true }, () => trigger())`.
- The watcher array is built before the first `runOnce()` call, so output written by the initial build is observable by those watchers. During a build an event sets `rerun = true` and the `finally` path schedules the next build.
- `collectWatchDirectories` walks the whole tree and skips only exact entry names:

```ts
if (
  entry.name === "node_modules" ||
  entry.name === ".git" ||
  entry.name === "lib" ||
  entry.name === "dist"
) {
  continue;
}
```

It never reads `outDir`, `declarationDir`, `include`, `exclude`, resolved root files, the `extends` chain, project references, or plugin inputs, and it is never called again.

- Node's `fs.watch` is non-recursive here. A watch on a parent reports that a child was created; it does not become a watcher for later changes inside that child.

### Filesystem probes with observed output

Both probes use the same non-recursive `fs.watch(dir, { persistent: true })` shape the launcher installs, on Windows.

New-directory probe, only the initial root watcher installed, then `later/` created and `later/value.ts` created and then modified:

```json
{
  "platform": "win32",
  "creationEvents": [{ "event": "rename", "name": "later" }],
  "fileCreateEvents": [{ "event": "change", "name": "later" }],
  "fileModifyEvents": []
}
```

The parent saw the topology change; the later edit produced no event. That is the missing-reconciliation failure exactly as `collectWatchDirectories` leaves it.

Pre-existing `build` output probe, watchers on the root and on `build` as the launcher installs them, then three simulated compiler writes:

```text
emit 1: build rename index.js; root change build; build change index.js
emit 2: build change index.js; build change index.js
emit 3: build change index.js; build change index.js
```

Every write produced an event on the `build` watcher. In `runWatch` each such event calls `trigger`, and the next real build writes the same outputs, so the loop sustains itself with no source edit.

### Coverage gap

The existing real-CLI watch tests (`tests/test-ttsc/src/features/compiler/test_ttsc_watch_exits_nonzero_on_failed_build.ts`, `test_ttsc_watch_survives_throwing_rebuild.ts`, and the `fix`/`format` watch rejections) prove the latest failure status is returned, that a throwing rebuild does not crash the session, and that `fix`/`format` reject watch mode. None creates a source directory after startup, uses an arbitrary pre-existing emit directory, places authored input under `lib` or `dist`, changes an external config or project reference, or asserts the absence of a self-triggered rebuild.

### Provenance

The recursive collector and the `lib`/`dist` names originate in the initial launcher work (`7a6d1a288`) and survived later watch status and error hardening plus launcher refactoring. No compiler-derived topology was lost in a regression; the watcher was born with directory-name guesses and those guesses are still its only output boundary.

Not a duplicate. Closed #293 is forced single-file emit outside `rootDir`. #214/#716 are bundler and plugin reference-graph invalidation. #263/#268 are graph package snapshots. PRs #97 and #187 changed emit and watch behavior without giving the watcher compiler-owned topology. No open or closed item reports the frozen-directory, new-input, or arbitrary-output loop class.

## Consequence surface

| Case | Compiler-relevant expectation | Current behavior |
| --- | --- | --- |
| Existing source file in an initially present directory | Rebuild | Rebuilds |
| New source directory created after startup, then a file inside it edited | Rebuild on both changes | Parent may trigger on creation; the directory is never added, so later edits are silent |
| Pre-existing `outDir`/`declarationDir` named `build`, `out`, `bin`, `generated`, … | Ignore compiler-owned writes | Watched; each emit schedules the next build indefinitely |
| Output directory first created by the initial build | Ignore compiler-owned writes | Parent sees the creation but the frozen set never adds the directory, so behavior differs purely on whether the directory pre-existed |
| Authored input below a directory named `lib` or `dist` | Rebuild | Entire subtree silently excluded even when the compiler includes it |
| Extended config, project reference, or loaded input outside the tsconfig directory | Rebuild | Never watched |
| Relevant declaration below `node_modules` changes | Rebuild | Whole dependency tree categorically excluded rather than following loaded inputs |
| README, coverage output, cache files below the root | Ignore | Rebuilds unless an excluded basename happens to hide them |
| Nested symlinked directory used as a source root | Rebuild | `Dirent.isDirectory()` does not traverse the symlink |

Secondary effects: emit-mode users can see unbounded rebuilds with repeated plugin process startups, repeated diagnostics, and output churn. `--noEmit` avoids the self-output loop but keeps every missed-input case. Config, reference, and plugin inputs are the riskiest, because the build can be logically stale while the session keeps reporting its last successful build. Large unrelated trees inflate the watcher count and can hit platform watcher limits; the registered asynchronous `error` listener logs but cannot repair an incomplete set. Windows, Linux, and macOS all install one watcher per initially present directory; event spelling and coalescing differ, but neither the static membership nor the output ownership problem is platform-specific.

## Approach

Invariant: **the owner that resolves a build's effective inputs and outputs owns watch invalidation.** The watched set must be derived from the resolved root files, configs, project references, and plugin inputs, reconciled after every rebuild and after topology changes, and it must exclude the build's own configured emit and cache products.

Architectural owner: the launcher/compiler boundary, `runWatch` in `packages/ttsc/src/launcher/internal/runTtsc.ts` together with whatever surface exposes the resolved program inputs to it. Whether that is delegation to TypeScript-Go watch facilities, exposing the resident build graph to the launcher, or a launcher-side dynamic set reconciled from `runBuild`'s resolution result is a design decision this issue deliberately does not settle. What it does require is one source of truth rather than a parallel filesystem approximation.

The four-name exclusion list must not simply grow. A longer list of guessed basenames reproduces both directions of the current failure: it still misses arbitrary configured outputs, and it still hides authored sources that happen to use one of the names.

| Case | Expected |
| --- | --- |
| Positive | Editing an initially present source file schedules exactly one debounced rebuild |
| Positive | Creating a new included directory and file after startup rebuilds, and a later edit inside that directory rebuilds again |
| Negative twin | With `outDir: "build"` pre-existing, an emit-mode session performs the initial build and then stays quiet; the build count does not grow during an idle period |
| Negative twin | With `outDir: "build"` absent at startup and created by the first build, the same quiet period holds |
| Positive | Authored included files below directories named `lib` and `dist` rebuild normally |
| Positive | Editing an extended config, a referenced project's input, or a loaded dependency input rebuilds |
| Negative twin | Editing an unrelated file below the root (README, coverage output) does not rebuild |
| Boundary | Deleting and renaming watched inputs and directories reconciles cleanly; a missing initial root that later appears becomes watched |
| Regression | `--noEmit`, `--emit`, check mode, single-file mode, symlinked logical roots, watcher error/limit handling, and clean SIGINT/SIGTERM termination keep their current status semantics |

New cases belong under `tests/test-ttsc/src/features/compiler/`, one assertion per file, and must count build cycles and enforce a quiet period after an emit rather than only asserting an initial success line.

Narrow proving command:

```bash
TTSC_TEST_DIR=features/compiler pnpm --filter @ttsc/test-ttsc start
```

That subtree already holds the watch cases this issue must keep green: `test_ttsc_watch_exits_nonzero_on_failed_build.ts`, `test_ttsc_watch_survives_throwing_rebuild.ts`, `test_ttsc_rejects_fix_with_watch_mode.ts`, and `test_ttsc_rejects_format_with_watch_mode.ts`. The runner exits 1 with "No tests were discovered under …" if the subtree is empty or misspelled.

Each new case must actually enter `runWatch`, which means invoking the launcher with `--watch` and observing the session over time. A one-shot build assertion cannot see any of the failures in this issue, because the frozen watcher set is only built inside `runWatch`.

Broader:

```bash
pnpm run test:typecheck
pnpm test:features
```

## Coordination

- The watch lane dispatches to `runSingleFile` for every rebuild when a file argument is present. #855 (`--noEmit` ignored in single-file mode) changes that dispatch decision and touches the same `runWatch` body. Sequence the two under one owner rather than running them in parallel.
- Any work that changes the launcher/native build boundary or the resolved-project surface should land in a known order relative to this one, since this fix needs the resolved input list to be reachable from the launcher.
- Independent of the lint, LSP, and Graph work; no file under `packages/graph` is involved.

samchon:
## Intent

Close #817 by deriving `ttsc --watch` invalidation from the compiler's resolved input topology rather than one startup directory walk.

## Implementation

- Resolve TypeScript-Go program inputs with `--listFilesOnly`, then reconcile file and directory watchers after every build and topology event.
- Keep the resolved `extends` chain, recursive project references, loaded external sources, and selected Go plugin source trees in the running watch session.
- Exclude configured compiler output locations from topology discovery, so initial and later emits do not schedule another build.
- Replace literal `lib`/`dist` exclusions with compiler ownership, and update the two watch guides to describe the new contract.

## Regression coverage

- Dynamic included directories; existing and initially absent `outDir` quietness; authored `lib` and `dist` inputs; unrelated README quietness.
- External extended configs, loaded TypeScript dependencies, project references, and selected Go plugin source edits.
- Existing watch recovery, error, and shutdown cases remain in the same real-launcher suite.

## Verification

- `pnpm --filter ttsc build`
- `TTSC_TEST_DIR=features/compiler pnpm --filter @ttsc/test-ttsc start -- --include=watch_` (15 cases, including the Go plugin source case)
- `TTSC_TEST_DIR=features/project pnpm --filter @ttsc/test-ttsc start`: all exercised config/plugin cases passed; the symlink-only case is blocked locally by Windows `EPERM` symlink permission.
- A full compiler-feature run exercised all watch cases successfully; two existing consumer-local fake-`tsc.exe` cases remain blocked by Windows `spawnSync ... UNKNOWN` in the fixture executable path.

## Coordination

This is only the compiler DAG root. #855 and #852 remain unclaimed until this PR merges.

Closes #817

### DIFF
--- a/packages/ttsc/src/compiler/internal/project/readProjectConfig.ts
+++ b/packages/ttsc/src/compiler/internal/project/readProjectConfig.ts
@@ -19,6 +19,7 @@ const PATH_OPTIONS = new Set(["baseUrl", "declarationDir", "rootDir"]);
  * projected into `ITtscParsedProjectConfig`.
  */
 type ResolvedCompilerOptions = {
+  configPaths: string[];
   options: Record<string, unknown>;
   /** Directory of the tsconfig that last declared each option key. */
   optionBaseDirs: Record<string, string>;
@@ -46,6 +47,7 @@ export function readProjectConfig(
   const root = identity.physicalProjectRoot;
   const compilerOptions = readResolvedCompilerOptions(tsconfig);
   return {
+    configPaths: compilerOptions.configPaths,
     compilerOptions: {
       ...compilerOptions.options,
       outDir: compilerOptions.outDir,
@@ -135,6 +137,7 @@ function readResolvedCompilerOptions(
       ? ownPlugins.filter(isProjectPluginConfig)
       : base.plugins;
     return {
+      configPaths: uniquePaths([...base.configPaths, canonical]),
       optionBaseDirs,
       options,
       outDir:
@@ -175,6 +178,7 @@ function resolveBaseCompilerOptions(
   }
   if (!Array.isArray(extended)) {
     return {
+      configPaths: [],
       optionBaseDirs: {},
       options: {},
       pluginBaseDirs: [],
@@ -183,6 +187,7 @@ function resolveBaseCompilerOptions(
     };
   }
   let merged: ResolvedCompilerOptions = {
+    configPaths: [],
     optionBaseDirs: {},
     options: {},
     pluginBaseDirs: [],
@@ -198,6 +203,7 @@ function resolveBaseCompilerOptions(
       seen,
     );
     merged = {
+      configPaths: uniquePaths([...merged.configPaths, ...current.configPaths]),
       optionBaseDirs: {
         ...merged.optionBaseDirs,
         ...current.optionBaseDirs,
@@ -215,6 +221,10 @@ function resolveBaseCompilerOptions(
     };
   }
   return merged;
+}
+
+function uniquePaths(paths: readonly string[]): string[] {
+  return [...new Set(paths)];
 }
 
 /**
--- a/packages/ttsc/src/compiler/internal/runBuild.ts
+++ b/packages/ttsc/src/compiler/internal/runBuild.ts
@@ -27,6 +27,12 @@ type RunBuildOptions = TtscBuildOptions & {
   skipDiagnosticsCheck?: boolean;
   forceListEmittedFiles?: boolean;
   /**
+   * Receives selected native-plugin source roots after the project resolves.
+   * The watch launcher uses these roots to invalidate a sidecar when its Go
+   * implementation changes between rebuilds.
+   */
+  onWatchInputs?: (inputs: readonly string[]) => void;
+  /**
    * Emit an external source map from the direct tsgo build lane even when the
    * project configures none. Set by the ttsx runtime builds so a served emit
    * carries a map to inline under the source URL (issue #353). Applied only to
@@ -194,6 +200,12 @@ function runBuildTimed(
   if (projectFree !== null) return projectFree;
   const setupStartedAt = process.hrtime.bigint();
   const execution = resolveExecutionContext(options);
+  options.onWatchInputs?.(
+    execution.nativePlugins.flatMap((plugin) => [
+      plugin.source,
+      ...(plugin.contributors?.map((contributor) => contributor.source) ?? []),
+    ]),
+  );
   if (
     execution.nativePlugins.length > 0 ||
     execution.pluginSetupFailure !== undefined
--- a/packages/ttsc/src/launcher/internal/runTtsc.ts
+++ b/packages/ttsc/src/launcher/internal/runTtsc.ts
@@ -22,6 +22,7 @@ import {
 import type { TtscBuildOptions } from "../../structures/internal/TtscBuildOptions";
 import { getCompilerVersionText } from "./getCompilerVersionText";
 import { resolveCacheDir } from "./resolveCacheDir";
+import { WatchTopology } from "./watchTopology";
 
 /**
  * CLI entry point for `ttsc`. Dispatches argv to the appropriate build lane
@@ -680,9 +681,13 @@ function runWatch(
   checkOnly: boolean,
 ): number {
   const cwd = path.resolve(options.cwd ?? process.cwd());
+  let topology: WatchTopology | undefined;
   const invocation = {
     ...options,
     cwd,
+    onWatchInputs: (inputs: readonly string[]) => {
+      topology?.setExtraInputs(inputs);
+    },
     quiet: true,
   };
   const root = path.dirname(
@@ -700,6 +705,7 @@ function runWatch(
 
   const runOnce = () => {
     running = true;
+    let completed = false;
     try {
       if (!options.preserveWatchOutput) {
         process.stdout.write("\x1bc");
@@ -725,6 +731,7 @@ function runWatch(
               return result.status;
             })();
       lastStatus = status;
+      completed = true;
       process.stdout.write(
         `[ttsc] ${status === 0 ? "watch build complete" : "watch build failed"}\n`,
       );
@@ -734,6 +741,19 @@ function runWatch(
       process.stdout.write(`[ttsc] watch build failed\n`);
     } finally {
       running = false;
+      if (completed) {
+        try {
+          // A filesystem event can arrive after the build reports completion
+          // but before this synchronous cleanup reaches its re-resolution.
+          // Notify for a membership change here as well, otherwise that new
+          // input is silently absorbed into the watch set without a rebuild.
+          topology?.refresh(true);
+        } catch (error) {
+          process.stderr.write(
+            `[ttsc] watch error on ${path.relative(cwd, root) || "."}: ${formatError(error)}\n`,
+          );
+        }
+      }
     }
     if (rerun) {
       rerun = false;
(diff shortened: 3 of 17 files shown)


samchon:
## Solo self-review

Reviewed the complete #817 diff at `cc35ffad9633294afe311bb40e8863cf3d650f95`.

- Traced resolved program inputs, `extends` chains, recursive references, output filtering, plugin-source invalidation, lifecycle cleanup, and Windows path identity.
- Rechecked normal source edits, topology additions/removals, config recovery, emit-loop suppression, and the race between a completed build's output and watcher reconciliation.
- Found no remaining sound correction.

Local evidence is in the PR body; the full real-launcher `watch_` suite passes all 15 cases, including selected Go plugin-source invalidation.

samchon:
## Final self-review (`82ecb07b73610a9225165899934bec11317af5c1`)

No blocking or follow-up findings.

- Re-read the full diff against `c7b0a781`, including topology reconciliation, `extends`/reference traversal, output exclusion, plugin-source invalidation, watcher lifecycle, and the post-build membership race.
- Confirmed the rebase preserves the reviewed implementation exactly (`patch-id c97d285c96f6e74126756fafe49cc2d8773bbf06`) and `git diff --check` is clean.
- `pnpm --filter ttsc build` passes. The real launcher watch regression set passed 15/15 cases (dynamic included directory, output quietness, lib/dist input, external config/source/reference, unrelated README, and selected Go-plugin source).
- The broader compiler run retains two unrelated Windows fixture failures that spawn a consumer-local fake `tsc.exe`; they are outside this change and are documented in the PR body.

PR #888, opened 2026-07-20
fix(release): publish upgradeable host ranges

### LINKED ISSUE #842: fix(release): publish host and sibling ranges a consumer can actually upgrade
## Problem

Three published packages declare a cross-package dependency with pnpm's `workspace:*` protocol. On publish, pnpm rewrites `workspace:*` to the **exact** version, not a range. The published contract is therefore an exact pin:

- `@ttsc/unplugin` peers `ttsc` at exactly the version it was released with;
- `@ttsc/graph` peers `ttsc` at exactly that version;
- `@ttsc/metro` depends on `@ttsc/unplugin` at exactly that version.

`ttsc` ships frequent patch releases. Under npm 7+ strict peer resolution, any *independent* upgrade of the host makes the tree unsatisfiable: `ttsc@0.19.4` beside `@ttsc/unplugin@0.19.3` is an `ERESOLVE` failure. That is the normal shape of automated dependency tooling, which bumps one dependency per pull request, and of a manual `npm install ttsc@latest`. The consumer's escapes are `--legacy-peer-deps` or discovering the lockstep bump themselves. `@ttsc/metro` inherits this transitively and additionally hard-pins `@ttsc/unplugin`, so a metro consumer cannot take an `@ttsc/unplugin` patch at all without a metro release.

The repository already knows the correct spelling and uses it for one relationship: `packages/playground` declares `"@ttsc/wasm": "workspace:^"`, which publishes as `^0.19.3`. That contrast is what makes this a defect rather than a policy choice.

Expected behavior: a published cross-package host or sibling dependency admits every release of that host it is compatible with. A published dependency on a **bundled native artifact** stays exact, because it must match the launcher that dispatches to it.

## Evidence

All four dependency specs below are unchanged between the audited commit `2b724664eee71c57cd688c55122cac137774bf62` and `master` at `612abc2d20d3f99429319081c1f6520bbe34ac7a`; the only difference in those manifests is the workspace-wide version bump to 0.19.4. That version is prepared but unpublished, so the registry state quoted here is the current live state, re-checked on 2026-07-20.

### The repository is already paying for this pin, in its own CI

This is not a hypothetical semver argument. `fb8faac86` (PR #785, "release: prepare manifest-complete ttscgraph 0.19.4") had to change `.github/workflows/typia.yml` to inject a paired `@ttsc/unplugin` tarball and patch downstream importers, and states the cause in its own commit message:

> The typia lane previously mixed a local ttsc candidate with the published unplugin, whose exact peer contract forced a registry lookup for the unpublished version.

The workflow's inline comment says the same thing at the mechanism level:

```js
// pnpm 10 does not apply a file override to an auto-installed exact
// peer. Give importers that consume only @ttsc/unplugin an explicit
// local ttsc so they cannot resolve the unpublished candidate from npm.
```

An exact peer that the resolver insists on satisfying from the registry is exactly the failure a consumer hits when they upgrade `ttsc` alone. The repository met it first because its prerelease lane is the first consumer to hold a `ttsc` the registry does not have.

That commit accepted the pin as a constraint ("ttsc and @ttsc/unplugin publish with exact peer-version alignment") and worked around it inside CI. The workaround is sound for the lane and is not a fix for the published contract, which is what this issue changes.

### The published contract, from the live registry

```text
$ npm view @ttsc/unplugin@latest version peerDependencies
version = '0.19.3'
peerDependencies = { ttsc: '0.19.3' }

$ npm view @ttsc/graph@latest version peerDependencies
version = '0.19.3'
peerDependencies = { ttsc: '0.19.3' }

$ npm view @ttsc/metro@latest version dependencies
version = '0.19.3'
dependencies = { '@ttsc/unplugin': '0.19.3' }

$ npm view @ttsc/playground@latest peerDependencies
{ ..., '@ttsc/wasm': '^0.19.3' }
```

### The source specs

| Package | Field | Spec | Publishes as |
| --- | --- | --- | --- |
| `packages/unplugin/package.json` | `peerDependencies.ttsc` | `workspace:*` | exact |
| `packages/graph/package.json` | `peerDependencies.ttsc` | `workspace:*` | exact |
| `packages/metro/package.json` | `dependencies["@ttsc/unplugin"]` | `workspace:*` | exact |
| `packages/playground/package.json` | `peerDependencies["@ttsc/wasm"]` | `workspace:^` | caret range |

### Why the existing test passes

`tests/test-unplugin/src/features/adapters/test_packaged_manifest_declares_the_external_ttsc_host.ts`, through `tests/test-unplugin/src/internal/packaged-host-contract.ts`, already packs the real tarball with `pnpm pack` and reads the published manifest. Its assertions are:

```ts
assert.ok(typeof peer === "string" && peer.length !== 0, ...);
assert.doesNotMatch(peer, /^workspace:/, "workspace protocol leaked into the published ttsc spec");
assert.notEqual(manifest.peerDependenciesMeta?.ttsc?.optional, true, ...);
assert.equal(manifest.dependencies?.ttsc, undefined, ...);
```

It asserts the spec is concrete and not `workspace:`, and stops deliberately short of asserting anything about the range. An exact pin satisfies every one of those, which is precisely why the current contract passes. That helper is the natural home for the extended assertion, and it already proves the technique works without a network install.

### Release-order exposure

`scripts/release-preflight.cjs` requires every non-private `packages/*` manifest to carry the tag's version, so publication is lockstep by design. It is not atomic: `package:latest:publish` runs `pnpm --filter=./packages/* --filter=!ttsc -r publish` and only then publishes `ttsc`. A publish that fails partway leaves the registry with mixed versions, and an exact pin turns a partial release into an uninstallable one rather than a degraded one.

- `packages/unplugin` gained the peer in `1dbac266c` (PR #695, closing issue #679, "fix(unplugin): declare the external ttsc host in the published dependency contract") as `"ttsc": "workspace:*"`. The accompanying test pinned "concrete and non-workspace", so the exact-pin consequence was never in scope.
- `packages/graph` has carried `workspace:*` since `9cbd847e6` (PR #259).
- `packages/playground` was authored with `workspace:^` in `2ae95c4ff`.
- No open or closed issue or pull request mentions peer ranges or `ERESOLVE`. No open pull request touches these manifests.
- No **Forbidden** classification is claimed. This is an ordinary packaging defect: one implementation, no fixture-only logic, no prior partial fix of the same cause.

At the time of writing all packages are at `0.19.3` on npm, so nothing is broken right now. This is a latent contract defect with a deterministic trigger, not an outage.

## Consequence surface

Every published cross-package relationship that uses the workspace protocol, and the range each one needs. This is the complete set: eleven relationships across the workspace.

| Consumer | Field | Target | Relationship | Today | Intended |
| --- | --- | --- | --- | --- | --- |
| `@ttsc/unplugin` | `peerDependencies` | `ttsc` | external host compiler | exact | range admitting patch upgrades |
| `@ttsc/graph` | `peerDependencies` | `ttsc` | external host compiler | exact | range admitting patch upgrades |
| `@ttsc/metro` | `dependencies` | `@ttsc/unplugin` | sibling adapter | exact | range admitting patch upgrades |
| `@ttsc/playground` | `peerDependencies` | `@ttsc/wasm` | sibling runtime | `^` range | unchanged, already correct |
| `ttsc` | `optionalDependencies` | `@ttsc/linux-x64` | bundled native artifact | exact | **stays exact** |
| `ttsc` | `optionalDependencies` | `@ttsc/linux-arm` | bundled native artifact | exact | **stays exact** |
| `ttsc` | `optionalDependencies` | `@ttsc/linux-arm64` | bundled native artifact | exact | **stays exact** |
| `ttsc` | `optionalDependencies` | `@ttsc/darwin-x64` | bundled native artifact | exact | **stays exact** |
| `ttsc` | `optionalDependencies` | `@ttsc/darwin-arm64` | bundled native artifact | exact | **stays exact** |
| `ttsc` | `optionalDependencies` | `@ttsc/win32-x64` | bundled native artifact | exact | **stays exact** |
| `ttsc` | `optionalDependencies` | `@ttsc/win32-arm64` | bundled native artifact | exact | **stays exact** |

The seven platform packages ship the native binary that the launcher of that exact `ttsc` version dispatches to. A range there would allow a mismatched binary, so the boundary is not "use a caret everywhere", it is per relationship.

Consumer-facing paths affected by the three exact pins:

- `npm install ttsc@latest` in a project that already has `@ttsc/unplugin` or `@ttsc/graph`.
- A Renovate or Dependabot pull request that bumps `ttsc` alone.
- A metro consumer taking an `@ttsc/unplugin` patch.
- Any install performed while a lockstep release is partially published.

Yarn and pnpm behave less strictly than npm 7+ by default, so the failure is loudest on npm, which is the manager the setup guides use (`npm install -D ttsc typescript @ttsc/unplugin`, and the equivalents in `setup/metro.mdx` and `setup/graph.mdx`).

## Approach

Invariant: **a published cross-package host or sibling dependency admits every release of that target that is compatible with it, while a published dependency on a bundled native artifact stays exact.** The decision is made per relationship, not per file and not per protocol spelling.

Owner: the three manifests, plus the packed-manifest contract helper `tests/test-unplugin/src/internal/packaged-host-contract.ts` that turns the decision into a check.

The spelling is `workspace:^`, matching the `packages/playground` precedent that already exists in this workspace. It publishes as a caret range, which under 0.x semantics accepts later patches of the same minor and stops at the next minor. That matches how this repository releases today, where every release advances the whole set in lockstep, and it makes the three relationships consistent with the one that is already correct.

Change these three specs and nothing else:

| File | Field | From | To |
| --- | --- | --- | --- |
| `packages/unplugin/package.json` | `peerDependencies.ttsc` | `workspace:*` | `workspace:^` |
| `packages/graph/package.json` | `peerDependencies.ttsc` | `workspace:*` | `workspace:^` |
| `packages/metro/package.json` | `dependencies["@ttsc/unplugin"]` | `workspace:*` | `workspace:^` |

The `devDependencies` entries on `ttsc` in those same manifests do appear in the published manifest (`npm view @ttsc/unplugin@0.19.3 devDependencies` shows `ttsc: '0.19.3'`), but a package manager never installs a published package's dev dependencies, so they cannot make a consumer's tree unsatisfiable. Leave them as `workspace:*`: inside the workspace that spelling correctly means "the local one".

Extend the packed-manifest contract to assert the intended shape rather than only concreteness, and extend it to `@ttsc/graph` and `@ttsc/metro`, which have no packed-manifest coverage at all today. The assertion must be a property of the range, not a literal version string, so it does not need editing on every release.

Do not widen the platform `optionalDependencies`. Do not solve this by publishing with a post-processing step that rewrites versions outside the manifest; the manifest is the reviewable contract.

| Case | Expected |
| --- | --- |
| Positive | `pnpm pack` of `@ttsc/unplugin`, `@ttsc/graph`, and `@ttsc/metro` yields host specs that accept a later patch of the target (a `0.19.3` release accepts `0.19.4`) |
| Negative | An exact-pinned host spec fails the extended contract test |
| Negative | A leaked `workspace:` protocol still fails, as it does today |
| Boundary | The range still rejects an incompatible major or minor host |
| Boundary | `ttsc`'s seven platform `optionalDependencies` are still exact after packing, and a range there fails the check |
| Regression | The existing assertions in `packaged-host-contract.ts` still hold: the peer is present, non-empty, not optional, and `ttsc` is not a bundled runtime dependency |
| Regression | `scripts/release-preflight.cjs` still passes for a tag, with every non-private package carrying the tag's version |

Narrow proving command:

```bash
pnpm --filter @ttsc/test-unplugin start
```

Broader: `pnpm test:features`, plus the new coverage wherever `@ttsc/graph` and `@ttsc/metro` packed-manifest checks land (`tests/test-graph` and `tests/test-metro` are the existing suites for those packages).

A real end-to-end confirmation is worth doing once by hand before release: pack the three packages plus a `ttsc` at a higher patch, install them together with npm, and confirm no `ERESOLVE`.

## Coordination

- Closed #679 and merged PR #695 own the "declare the external host at all" step. Reference them; do not re-file. Their assertions must keep passing.
- Closed #682 owns release preflight (tag and version coherence before publication). It is unrelated to range shape.
- Merged PR #785 is the CI workaround described above. If this issue lands, revisit whether the `typia.yml` importer patching is still needed, but do not remove it in the same change without proving the lane still passes.
- The sibling packaging candidate #807 (Windows platform package content verification) touches `scripts/assert-platform-package.cjs` and `tests/test-ttsc/**`. No file overlap, so the two can run in parallel.
- `packages/graph/package.json` is edited. The protected Graph contract is unaffected: `packages/graph/src/structures/ITtscGraphApplication.ts`, every detailed and public subordinate type in that application contract, and all of their JSDoc stay byte-for-byte unchanged.
- Publish timing: this changes the published dependency contract, so it should land in a release window where the whole set is republished, which is already how this repository releases.

samchon:
## Summary

Reserves the admitted #842 packaging-contract correction.

- change only the published runtime relationships in `@ttsc/unplugin`,
  `@ttsc/graph`, and `@ttsc/metro` from `workspace:*` to `workspace:^`;
- prove each owning package's packed tarball publishes the intended compatible
  range, including negative and boundary cases;
- keep the three packages' development-only `workspace:*` edges and ttsc's
  seven native-platform optional dependency pins exact.

## Excluded

No release, tag, publication, workflow change, lockstep-version change, or
typia workflow workaround removal is part of this pull request. The protected
Graph application contract remains byte-for-byte unchanged.

Closes #842.

## Verification

Pending. This is the implementation-free campaign reservation. Actions caused
by this branch are cancelled only by exact commit SHA; the merge gates will be
the package-owned packed-tarball suites, appropriate broader feature coverage,
and a fresh solo Self-Review.

### DIFF
--- a/packages/graph/package.json
+++ b/packages/graph/package.json
@@ -23,7 +23,7 @@
     "typia": "catalog:samchon"
   },
   "peerDependencies": {
-    "ttsc": "workspace:*"
+    "ttsc": "workspace:^"
   },
   "keywords": [
     "ttsc",
--- a/packages/metro/package.json
+++ b/packages/metro/package.json
@@ -38,7 +38,7 @@
     "src"
   ],
   "dependencies": {
-    "@ttsc/unplugin": "workspace:*"
+    "@ttsc/unplugin": "workspace:^"
   },
   "peerDependencies": {
     "@expo/metro-config": "*",
--- a/packages/unplugin/package.json
+++ b/packages/unplugin/package.json
@@ -98,7 +98,7 @@
     "unplugin": "^2.3.11"
   },
   "peerDependencies": {
-    "ttsc": "workspace:*"
+    "ttsc": "workspace:^"
   },
   "devDependencies": {
     "@rollup/plugin-commonjs": "^29.0.2",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -184,7 +184,7 @@ importers:
         specifier: '*'
         version: 0.86.0(@babel/core@7.29.7)
       '@ttsc/unplugin':
-        specifier: workspace:*
+        specifier: workspace:^
         version: link:../unplugin
       metro-react-native-babel-transformer:
         specifier: '*'
--- a/tests/test-graph/src/features/test_packaged_manifest_declares_upgradeable_ttsc_host.ts
+++ b/tests/test-graph/src/features/test_packaged_manifest_declares_upgradeable_ttsc_host.ts
@@ -77,7 +77,9 @@ function assertCompatibleCaretRange(range: string, dependency: string): void {
     match,
     `${dependency} must publish a caret range, received ${range}`,
   );
-  const [major, minor, patch] = match.slice(1).map(Number);
+  const major = Number(match[1]!);
+  const minor = Number(match[2]!);
+  const patch = Number(match[3]!);
   const lower: [number, number, number] = [major, minor, patch];
   const upper: [number, number, number] =
     major > 0
@@ -99,8 +101,5 @@ function compareVersions(
   left: [number, number, number],
   right: [number, number, number],
 ): number {
-  for (let index = 0; index < left.length; ++index) {
-    if (left[index] !== right[index]) return left[index] - right[index];
-  }
-  return 0;
+  return left[0] - right[0] || left[1] - right[1] || left[2] - right[2];
 }
--- a/tests/test-graph/src/features/test_packaged_manifest_declares_upgradeable_ttsc_host.ts
+++ b/tests/test-graph/src/features/test_packaged_manifest_declares_upgradeable_ttsc_host.ts
@@ -1,0 +1,103 @@
+import { TestProject } from "@ttsc/testing";
+import assert from "node:assert/strict";
+import { spawnSync } from "node:child_process";
+import fs from "node:fs";
+import path from "node:path";
+
+/**
+ * Verifies the published @ttsc/graph manifest permits compatible ttsc upgrades.
+ *
+ * The graph server runs inside the consumer's ttsc host, so an exact peer pin
+ * makes a normal independent patch upgrade fail npm's peer solver. Reading the
+ * packed manifest tests the release artifact, where pnpm expands the workspace
+ * protocol, rather than merely restating the source manifest spelling.
+ *
+ * 1. Pack @ttsc/graph and extract its published package manifest.
+ * 2. Assert the required ttsc peer is a concrete caret range that admits the next
+ *    compatible patch.
+ * 3. Assert the range rejects its next incompatible semver boundary and ttsc
+ *    remains external to the graph package.
+ */
+export const test_packaged_manifest_declares_upgradeable_ttsc_host = () => {
+  const manifest = readPackedManifest();
+  const peer = manifest.peerDependencies?.ttsc;
+  assert.ok(
+    typeof peer === "string",
+    "published graph manifest needs a ttsc peer",
+  );
+  assert.notEqual(
+    manifest.peerDependenciesMeta?.ttsc?.optional,
+    true,
+    "published graph ttsc peer must be required",
+  );
+  assert.equal(
+    manifest.dependencies?.ttsc,
+    undefined,
+    "published graph must not bundle a second ttsc runtime",
+  );
+  assertCompatibleCaretRange(peer, "ttsc");
+};
+
+function readPackedManifest(): Record<string, any> {
+  const packageDir = path.join(TestProject.WORKSPACE_ROOT, "packages", "graph");
+  const destination = TestProject.tmpdir("ttsc-graph-pack-");
+  const pack = spawnSync("pnpm", ["pack", "--pack-destination", destination], {
+    cwd: packageDir,
+    encoding: "utf8",
+    shell: process.platform === "win32",
+  });
+  assert.equal(pack.status, 0, `pnpm pack failed:\n${pack.stderr}`);
+
+  const tarball = fs
+    .readdirSync(destination)
+    .find((name) => name.endsWith(".tgz"));
+  assert.ok(tarball, "pnpm pack produced no graph tarball");
+  const manifest = spawnSync(
+    "tar",
+    ["-xzOf", tarball, "package/package.json"],
+    {
+      cwd: destination,
+      encoding: "utf8",
+    },
+  );
+  assert.equal(
+    manifest.status,
+    0,
+    `tar extraction failed:\n${manifest.stderr}`,
+  );
+  return JSON.parse(manifest.stdout);
+}
+
+function assertCompatibleCaretRange(range: string, dependency: string): void {
+  const match = /^\^(\d+)\.(\d+)\.(\d+)(?:-[0-9A-Za-z.-]+)?$/.exec(range);
+  assert.ok(
+    match,
+    `${dependency} must publish a caret range, received ${range}`,
+  );
+  const [major, minor, patch] = match.slice(1).map(Number);
+  const lower: [number, number, number] = [major, minor, patch];
+  const upper: [number, number, number] =
+    major > 0
+      ? [major + 1, 0, 0]
+      : minor > 0
+        ? [0, minor + 1, 0]
+        : [0, 0, patch + 1];
+  const accepts = (candidate: [number, number, number]): boolean =>
+    compareVersions(candidate, lower) >= 0 &&
+    compareVersions(candidate, upper) < 0;
+
+  if (minor > 0 || major > 0) {
+    assert.equal(accepts([major, minor, patch + 1]), true);
+  }
+  assert.equal(accepts(upper), false);
+}
+
+function compareVersions(
+  left: [number, number, number],
+  right: [number, number, number],
+): number {
+  for (let index = 0; index < left.length; ++index) {
+    if (left[index] !== right[index]) return left[index] - right[index];
+  }
+  return 0;
+}
--- a/tests/test-graph/src/features/test_packaged_manifest_declares_upgradeable_ttsc_host.ts
+++ b/tests/test-graph/src/features/test_packaged_manifest_declares_upgradeable_ttsc_host.ts
@@ -14,7 +14,7 @@ import path from "node:path";
  *
  * 1. Pack @ttsc/graph and extract its published package manifest.
  * 2. Assert the required ttsc peer is a concrete caret range that admits the next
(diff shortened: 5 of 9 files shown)


samchon:
Self-Review finding: the packed-version assertions accepted prereleases but not SemVer build metadata, although release preflight accepts build metadata. I will extend the range/exact-version patterns before the next snapshot so every release form supported by the repository remains covered.

samchon:
Fresh Self-Review finding: the existing unplugin packed-manifest test comments still describe workspace:* becoming an exact version, while this PR now asserts the published workspace:^ caret contract. I will correct those test-local contract comments before the final review so the maintained rationale matches the tested artifact.

samchon:
Verification finding: pnpm test:typecheck rejects the three new caret-range helpers under noUncheckedIndexedAccess because the regex capture array and tuple loop are not narrowed. I will replace them with explicit captured components and fixed tuple comparisons, then restart the required verification and solo review on the new snapshot.

samchon:
Fresh Self-Review finding: the packed-manifest assertions prove the current caret range, but do not execute the required negative twin that an exact pin is rejected. I will add that oracle to each contract helper, then restart the affected package suites and the final solo review on the corrected snapshot.

samchon:
Final solo Self-Review on fe614b71584ce4e47557d6129c20c6e98fd374da found no remaining in-scope improvement. The final diff preserves the protected Graph application file SHA-256 and structures tree exactly.\n\nFocused verification passed: packed-manifest regressions for @ttsc/unplugin, @ttsc/graph, @ttsc/metro, and ttsc platform pins; package typechecks for unplugin, graph, and metro; and 
ode scripts/release-preflight.cjs --tag v0.19.4.\n\nBroader evidence: the full Metro suite passed. pnpm test:typecheck fails in two unchanged playground files, while the test-ttsc package typecheck fails in two unchanged watch nullability sites. The full graph suite fails two pre-existing graph snapshot assertions outside this PR's manifest/test surface. The old full unplugin and test-ttsc commands that began before the final snapshot still have owned Go children running, so their results are not used as acceptance evidence; they will be allowed to terminate before worktree cleanup. The focused final-snapshot cases are the merge acceptance evidence.

samchon:
Rebased the unchanged reviewed patch onto current master 5edb8859e2f516894ad58d3af764767a2463f0be. Fresh final verification on 1a0c47747cd0130dd4339252bb710e8a8fc093b5 passed all four real packed-manifest cases, the unplugin/graph/metro package typechecks, and release preflight for v0.19.4. The protected Graph application SHA-256 and structures tree remain unchanged. The final solo Self-Review remains clean; campaign Actions runs for the rebased SHA are being recorded and cancelled by exact SHA.

Issue #864, opened 2026-07-20
fix(factory): the Expression union omits kinds the printer prints


samchon:
## Problem

`@ttsc/factory`'s public `Expression` union does not contain every node kind the printer can print as an expression. `TsPrinter` has a `case "PartiallyEmittedExpression"` arm, and the factory exposes JSX element and fragment builders, but neither kind is a member of `Expression`.

The consequence is that correct code does not typecheck and some printer behavior is unreachable from typed callers:

- `factory.createExpressionStatement(jsxElement)` cannot be written against the public types, though the printer emits it.
- The parenthesizer path for `PartiallyEmittedExpression` cannot be exercised from typed code at all, because no typed value can reach it.

## Evidence

Found while implementing #851. The implementer began correcting a real runtime parenthesizer gap for `PartiallyEmittedExpression`, and **the fix itself failed to compile with TS2367** — a comparison against a kind the union does not admit is statically impossible. That compile error is the proof: the runtime branch exists, and the type surface says nothing can ever arrive at it.

At the audited baseline:

- `packages/factory/src/ast/expressions/Expression.ts` is a 40-member union with no `PartiallyEmitted` member and no `Jsx*` member.
- `packages/factory/src/TsPrinter.ts` carries `case "PartiallyEmittedExpression":` in its expression switch.

So the printer's kind coverage and the public union disagree, in a direction that silently narrows what a drop-in caller can express.

## Consequence surface

| Affected use | Today | Expected |
| --- | --- | --- |
| `createExpressionStatement(jsxElement)` | not typeable | typeable, and already printable |
| JSX element or fragment used in any expression position | not typeable | typeable |
| `PartiallyEmittedExpression` reaching the parenthesizer | unreachable from typed code | reachable, and correctly parenthesized |
| A printer fix for either kind | cannot be written or tested through the public API | ordinary |

The reach is the whole public type surface rather than one builder: anything typed `Expression` — arguments, initializers, returns, array and object members — inherits the omission.

## Approach

The union and the printer's expression switch must describe the same set. Whichever direction is chosen has to be deliberate rather than incidental:

1. **Widen the union** to admit the kinds the printer already prints, which is what a drop-in `ts.factory` consumer expects, since the legacy `Expression` type admits both.
2. **Narrow the printer** by removing an arm the public API can never reach, if the kind is genuinely unsupported.

[...]

Issue #834, opened 2026-07-20
fix(factory): gate the `ts.factory` drop-in promise on signatures and printed text, not on member names


samchon:
## Problem

`@ttsc/factory` is a standalone published library (`@ttsc/factory@0.19.3`) whose whole purpose is to keep `ts.factory`-based code generation working after the tsgo migration. Its README states the contract in one sentence:

> `createXxx` methods mirror the legacy `ts.factory` names and parameter order, and return concrete, fully typed _outline_ AST nodes (each with a `kind` discriminant).

The repository has exactly one gate for that promise, `tests/test-factory/src/features/parity/test_factory_completeness.ts`. It parses the public `NodeFactory` interface out of the real `typescript.d.ts` and compares **member name sets**, in both directions. Nothing compares a parameter list, a parameter type, or a byte of printed output against the oracle. Half of the sentence the README publishes is therefore ungated, and every divergence below reports green on the `factory` CI lane.

Four classes of divergence are live at the audited baseline, all invisible to that gate:

1. **Parameter order.** `createImportTypeNode` and `createJSDocImportTag` skip a parameter that upstream has in the middle of the list, so a caller ported from real `ts.factory` binds arguments to the wrong slots. `createImportClause` implements the `@deprecated` boolean overload rather than the current one, so a modern first argument is read as `isTypeOnly`.
2. **Literal escaping.** `escapeString` escapes only the backslash, LF, CR, TAB and the active quote. C0 controls, DEL, U+2028, U+2029 and unpaired surrogates are emitted raw, and an unpaired surrogate becomes U+FFFD the moment the generated text is written to a UTF-8 file, so the generated program holds a different string value than the AST the caller built.
3. **A published input that does nothing.** `createModuleDeclaration`'s JSDoc documents `flags` as the input that selects the emitted keyword; the printer never reads it, and `NodeFlags.ts` documents the opposite behavior in the same package.
4. **A printed token that no other producer emits.** `createComma` prints `a , b`. `CommaListExpression`, the legacy printer and the repository's own pinned Prettier all print `a, b`, and `createComma`'s own JSDoc example shows `(a, b)`.

Underneath all four sits one structural cause: **the parity oracle can only see names**, so nothing that is not a name can regress-detect itself. Fixing the four classes without extending the oracle leaves the class open for the next builder that is added or edited.

Affected users are external consumers of the published package. Nothing in this workspace depends on `@ttsc/factory`, which is also why no in-repo suite has ever exercised these paths.

[...]

PR #889, opened 2026-07-20
fix(factory): align expression kind coverage

samchon:
## Summary

Reserves the implementation surface for #864: align `@ttsc/factory`'s public `Expression` union and expression parenthesizer with the printable TypeScript legacy AST kinds.

## Scope

- Include `PartiallyEmittedExpression` and all printable JSX expression forms in the public union.
- Verify transparent partially emitted wrappers and JSX expressions against the pinned `ts-legacy` printer.
- Extend the existing factory kind-coverage regression checks.
- Excludes #834's maintainer-gated signature and compatibility decisions.

## Verification

Pending implementation. Planned gates are `pnpm --filter @ttsc/test-factory start` and `pnpm test:features`.

## Campaign CI

This is an issue-campaign implementation PR. Actions runs triggered by this branch are cancelled per exact commit SHA; local verification and solo self-review are the merge gates.

### DIFF
--- a/packages/factory/src/TsPrinter.ts
+++ b/packages/factory/src/TsPrinter.ts
@@ -1739,9 +1739,20 @@ export class TsPrinter {
    * re-associate — matching the legacy printer's parenthesizer rules.
    */
   private parenthesizedExpression(expression: Expression): Doc {
-    return expression.kind === "ParenthesizedExpression"
+    return this.skipPartiallyEmittedExpressions(expression).kind ===
+      "ParenthesizedExpression"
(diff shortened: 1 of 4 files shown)


samchon:
Self-review found one in-scope coverage gap: skipPartiallyEmittedExpressions intentionally unwraps nested transform wrappers with a loop, but the new regression only constructs one wrapper. I will add a nested partial comma-precedence oracle row, rerun the factory gates, then repeat the complete review on the resulting snapshot.

PR #890, opened 2026-07-20
fix(ttsx): preserve prefix-only builtin URLs

### LINKED ISSUE #880: ttsx: preserve prefix-only node builtin URLs on affected Node releases
## Problem

`ttsx` fails before executing a valid CommonJS program on affected Node.js releases when the program calls a prefix-only builtin such as `require("node:sqlite")`.

The failure is not caused by `nextLoad()` returning `source: null`. A builtin load result is allowed to carry `format: "builtin"` with a null source. The real defect occurs earlier: affected Node versions pass the synchronous resolve hook `node:sqlite` but `nextResolve()` returns the stripped URL `sqlite`. The later load validation no longer recognizes that URL as a builtin and rejects its otherwise valid null source with `ERR_INVALID_RETURN_PROPERTY_VALUE`.

`ttsx` promises Node.js 22.15.0 or later because synchronous `module.registerHooks()` first became available there, so the runtime must preserve the declared support floor rather than requiring users to discover a narrower set of working Node minors.

## Minimal reproduction

On Node.js 22.15.0:

```js
const { registerHooks } = require("node:module");

registerHooks({
  resolve(specifier, context, nextResolve) {
    const result = nextResolve(specifier, context);
    console.error(specifier, "=>", result.url);
    return result;
  },
  load(url, context, nextLoad) {
    return nextLoad(url, context);
  },
});

require("node:sqlite");
```

Output:

```text
node:sqlite => sqlite

TypeError [ERR_INVALID_RETURN_PROPERTY_VALUE]:
Expected a string, an ArrayBuffer, or a TypedArray to be returned for the
"source" from the "load" hook but got null.
```

Control cases:

- `require("node:crypto")` passes because `nextResolve()` preserves `node:crypto`.
- `import("node:sqlite")` passes because this bug is on the CommonJS builtin path.
- Replacing only the exact bad resolve result `sqlite` with the original `node:sqlite` makes the CommonJS reproduction pass without changing the load result.

The downstream generated AutoBE backend reproduces the same path through `RuntimeAuthProbeProvider.ts`, whose CommonJS emit calls `require("node:sqlite")`. Its independently installed graph uses Node.js 22.15.0, TypeScript 7.0.2, and `ttsc@0.19.3`.

## Ownership and version boundary

This is a Node.js bug already fixed upstream:

- nodejs/node#58607
- nodejs/node#58612
- nodejs/node commit `bfc68c8ae83d61d005c860449b787bb80df5d17b`

Affected version families include:

| Node.js | Behavior |
| --- | --- |
| 22.15.0 through 22.17.1 | affected |
| 22.18.0 and later 22.x | fixed upstream |
| 23.5.0 through 23.11.1 | affected; no fixed 23.x release |
| 24.0.0 through 24.3.x | affected |
| 24.4.0 and later 24.x | fixed upstream |
| 25.x and later | fixed upstream |

Raising ttsc's minimum Node version to 22.18.0 would still incorrectly admit affected Node 23 and early Node 24 releases. A narrow compatibility correction is required.

The exact ttsc exposure starts with synchronous runtime hooks in commit `7e2a1ad13b9f1c5b6ce86e4f881bbe521145b948`, first released in `v0.15.0`. The relevant `runtimeHooks.ts` behavior is unchanged across `ttsc@0.19.0` through `0.19.3`; this is not a patch-release regression.

## Consequence surface

The affected class is every `ttsx` CommonJS graph that requires a prefix-only builtin on an affected Node release. Known prefix-only builtin families include `node:sqlite`, `node:test`, `node:test/reporters`, and `node:sea`. Ordinary builtins such as `node:crypto`, whose resolved URL retains the prefix, are controls rather than failing examples.

The AutoBE authorization-runtime fixture is one release-shaped downstream consumer. Its original `node:crypto` attribution was incidental because the same graph later reaches `node:sqlite`.

## Approach

Normalize only the exact broken `nextResolve()` result in `runtimeHooks.ts`:

```ts
specifier.startsWith("node:") &&
result.url === specifier.slice("node:".length)
```

For that case, preserve the rest of the resolve result and restore `url: specifier`.

Do not synthesize builtin source, replace a valid null source, special-case the AutoBE fixture, or overwrite arbitrary custom `node:` remaps. The compatibility layer owns only Node's prefix-stripping defect.

- On Node.js 22.15.0, a real `ttsx` CommonJS entry requiring `node:sqlite` exits zero.
- Prefix-only builtins `node:sqlite`, `node:test`, `node:test/reporters`, and `node:sea` retain their `node:` URL through the runtime hook.
- `node:crypto` remains unchanged as the ordinary-builtin control.
- ESM builtin imports remain unchanged.
- A custom hook result that maps a `node:` specifier to any URL other than its exact stripped spelling is preserved.
- The load hook continues to forward the valid builtin `source: null` result rather than fabricating source.
- The focused ttsx runtime suite and broader ttsc verification pass.
- The unchanged downstream AutoBE command exits zero and prints `AUTH_RUNTIME_OK`:

  ```powershell
  pnpm --filter autobe-auth-runtime-fixture-backend exec ttsx --project test/tsconfig.json test/authRuntime.ts
  ```

## Coordination

Downstream integration gate: wrtnlabs/autobe-mcp#588.

After a fixed ttsc release exists, AutoBE must adopt it through the ordinary root catalog and `pnpm sync:versions` path, then rerun the independently installed generated-runtime fixture. No AutoBE-only loader patch or issue-specific downgrade should be added.

samchon:
## Summary

- restore the `node:` scheme only when Node's synchronous resolver returns the exact prefix-stripped URL for an actual core builtin
- preserve all other resolver results, including their metadata and object identity, while leaving builtin `load` results and valid null sources unchanged
- cover prefix-only CommonJS builtins, ordinary builtins, ESM imports, custom remaps, and non-builtin `node:` specifiers with real `ttsx` runs and a pure helper matrix
- run the failing CommonJS regression on Node.js 22.15.0 in CI and document the compatibility behavior

## Cause

Affected Node.js releases pass a prefix-only builtin such as `node:sqlite` to synchronous resolve hooks but return `sqlite` from `nextResolve()` on the CommonJS path. Returning that stripped URL prevents the following load result from being recognized as a builtin, so Node rejects its otherwise valid null source with `ERR_INVALID_RETURN_PROPERTY_VALUE`.

The correction is deliberately narrower than a general `node:` rewrite. It requires `module.isBuiltin(specifier)` and an exact prefix-stripped result, so user hook remaps and non-core `node:`-shaped specifiers remain untouched.

## Validation

- reproduced the pre-fix failure with the real `ttsx` launcher on Node.js 22.15.0
- focused helper and real `ttsx` regressions pass on Node.js 22.15.0
- `pnpm run build:current`
- `pnpm run test:typecheck`
- `pnpm run check:flags`
- `node --test packages/ttsc/scripts/check-flags.test.cjs`
- `pnpm --filter @ttsc/website build`
- `pnpm format`

The full local `features/ttsx-runtime` run passed the changed surface and all but two pre-existing Windows environment cases. One cannot spawn a copied temporary `tsc.exe` on this machine (`spawnSync ... UNKNOWN`), and one requires symlink permission (`EPERM`). Ordinary CI remains the acceptance gate for those platform-independent lanes.

## Integration chronology

- rebased the implementation onto `origin/master` at `a8288d649d62eb941ac0638b58bb5fb1b7b30abd`
- verified head `b119af085d36fba87fa740e094432c8143e605e9` with `pnpm install`, `pnpm --filter ttsc build`, and all five focused regressions on Node.js 22.15.0
- confirmed the branch is exactly one commit ahead of that master with a clean eight-file diff
- rerunning the repository-wide typecheck after the rebase reaches a pre-existing failure introduced on current master in `test_compiler_client_fences_connection_generations.ts`: its `ICompilerService`-typed driver is asserted through a test-only `connectorId`. The #880 files do not touch playground code; the earlier full typecheck passed before that upstream refresh.

Closes #880

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -72,6 +72,13 @@ jobs:
               scope: test-ttsc,
             }
           - {
+              name: "ttsx node 22.15",
+              node: "22.15.0",
+              build: "pnpm --filter ttsc build",
+              run: "pnpm --filter @ttsc/test-ttsc start -- --include=commonjs_loads_prefix_only_node_builtins",
+              dir: features/ttsx-runtime,
+            }
+          - {
               name: "ttsc native corpus-source",
               run: "pnpm --filter @ttsc/test-ttsc start",
               dir: native-plugins/corpus-source,
@@ -203,7 +210,7 @@ jobs:
 
       - uses: actions/setup-node@v4
         with:
-          node-version: 24.x
+          node-version: ${{ matrix.node || '24.x' }}
 
       - uses: actions/setup-go@v5
         with:
--- a/packages/ttsc/src/launcher/internal/runtimeHooks.ts
+++ b/packages/ttsc/src/launcher/internal/runtimeHooks.ts
@@ -1,6 +1,11 @@
 import crypto from "node:crypto";
 import fs from "node:fs";
-import { Module, registerHooks, stripTypeScriptTypes } from "node:module";
+import {
+  Module,
+  isBuiltin,
+  registerHooks,
+  stripTypeScriptTypes,
+} from "node:module";
 import os from "node:os";
 import path from "node:path";
 import { fileURLToPath, pathToFileURL } from "node:url";
@@ -294,7 +299,10 @@ function resolve(
 ): ResolveResult {
   try {
     return rememberCommonJsNamedInterop(
-      nextResolve(specifier, context),
+      restoreStrippedNodeBuiltinScheme(
+        specifier,
+        nextResolve(specifier, context),
+      ),
       context,
     );
   } catch (error) {
@@ -307,6 +315,26 @@ function resolve(
       context,
     );
   }
+}
+
+/**
+ * Restore a `node:` builtin URL when affected Node releases return the exact
+ * prefix-stripped spelling from their synchronous CommonJS resolver.
+ *
+ * Every other result passes through unchanged. In particular, a user hook that
+ * intentionally remaps a `node:` specifier to another URL retains ownership of
+ * that mapping, while ordinary and ESM builtin results already carrying the
+ * scheme avoid an unnecessary copy.
+ */
+export function restoreStrippedNodeBuiltinScheme(
+  specifier: string,
+  result: ResolveResult,
+): ResolveResult {
+  return isBuiltin(specifier) &&
+    specifier.startsWith("node:") &&
+    result.url === specifier.slice("node:".length)
+    ? { ...result, url: specifier }
+    : result;
 }
 
 /** Cache of built projects keyed by owning tsconfig path. */
--- a/packages/wasm/test/host/fountain_position_tokens_test.go
+++ b/packages/wasm/test/host/fountain_position_tokens_test.go
@@ -36,8 +36,10 @@ type fountainSymbol struct {
 // offsets, position errors, and release lifecycle errors.
 func TestFountainPositionVerbsResolveTouchingTokens(t *testing.T) {
 	api := startFountainAPI(t)
-	root := fmt.Sprintf("/fountain-position-%d", time.Now().UnixNano())
-	t.Cleanup(func() { _ = os.RemoveAll(root) })
+	root := os.Getenv("TTSC_WASM_TEST_ROOT")
+	if root == "" || !filepath.IsAbs(root) {
+		t.Fatalf("TTSC_WASM_TEST_ROOT must be an absolute wasm path, got %q", root)
+	}
 	if err := os.MkdirAll(filepath.Join(root, "src"), 0o755); err != nil {
 		t.Fatal(err)
 	}
--- a/scripts/go-wasm-exec.cjs
+++ b/scripts/go-wasm-exec.cjs
@@ -21,6 +21,7 @@ const keep = new Set([
   "PATHEXT",
   "SystemRoot",
   "TEMP",
+  "TTSC_WASM_TEST_ROOT",
   "TMP",
   "USERPROFILE",
 ]);
@@ -28,5 +29,9 @@ for (const key of Object.keys(process.env)) {
   if (!keep.has(key)) delete process.env[key];
 }
 
-process.argv = [process.argv[0], path.resolve(wasmExec), ...process.argv.slice(3)];
+process.argv = [
+  process.argv[0],
+  path.resolve(wasmExec),
+  ...process.argv.slice(3),
+];
 require(process.argv[1]);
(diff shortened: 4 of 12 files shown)


samchon:
Rebased onto current `origin/master` (`a8288d649d62eb941ac0638b58bb5fb1b7b30abd`) and force-pushed head `b119af085d36fba87fa740e094432c8143e605e9`.

Post-rebase integration evidence on Node.js 22.15.0:

- `pnpm install`
- `pnpm --filter ttsc build`
- all five focused helper/CJS/ESM/custom-remap regressions
- clean eight-file diff, exactly one commit ahead of current master

The PR body now records the unrelated repository-wide typecheck failure already present in the newly merged playground test on current master. New-head Actions are running; none were cancelled.

samchon:
The current-master playground typecheck blocker documented here is now tracked in #896 and claimed by Draft PR #897. The fix is confined to the #884 test driver's local connectorId narrowing; production ICompilerService and #890's ttsx surface remain unchanged. #897 is running ordinary CI and will be merged before treating #890's latest-target typecheck as unblocked.

samchon:
Final campaign validation for head `c1c0510169736b7337237648e1d3a1d58ad568b8`, rebased onto `master` at `cde9c6f71dd327b00a6a47e9107918afae35969b`:

- Node.js `v22.15.0`: `pnpm --filter ttsc build` and all five focused prefix-only/ordinary/ESM/custom/helper regressions passed.
- `pnpm run test:typecheck` passed every workspace typecheck target.
- `node scripts/test-go-wasm.cjs` passed with worktree-scoped `GOCACHE`/`GOTMPDIR`; the runner-created fixture count was `0` before and after.
- The wasm runner's forced failure path also preserved both fixture and go.work temporary-directory counts, proving `finally` cleanup executes.
- `pnpm --filter ttsc shim:audit:test` and `pnpm --filter ttsc shim:audit` passed.
- scoped Prettier checks and `git diff --check origin/master...HEAD` passed.
- fresh solo whole-diff Self-Review completed with no remaining findings.

Per the campaign instruction, this PR also absorbs both real red-CI causes discovered after the original implementation:

- `tests/test-ttsc/src/internal/watch.ts` now narrows the explicitly piped child streams at runtime, removing the `TS2531` baseline failure without a non-null assertion.
- The js/wasm host runner now supplies a writable POSIX-absolute fixture root on both Linux and Windows, fixing issue #907's host-root `/fountain-position-*` permission failure while cleaning the fixture on success and failure.

The earlier website module and shim-audit baseline failures are resolved by merged `master` repairs #900 and #901. The unrelated full local `ttsx-runtime` Windows-only environment failures remain the previously documented copied-executable `spawnSync UNKNOWN` and file-symlink `EPERM`; every changed runtime case passes.

Exact-head Actions runs `29750221059`, `29750221275`, `29750221617`, `29750221163`, `29750221220`, `29750221347`, `29750221465`, `29750221744`, and `29750221856` are terminal `cancelled` under the campaign policy; the two Socket checks succeeded.

Issue #848, opened 2026-07-20
fix(wasm): stamp the build metadata `@ttsc/wasm` documents into the binaries it publishes


samchon:
## Problem

`api.version()` is the one API whose purpose is to identify which wasm is running. Every published `@ttsc/wasm` binary answers with the compile-time defaults, so it identifies nothing:

```text
{
 "version": "0.0.0-dev",
 "commit": "dev",
 "date": "unknown",
 "go": "go1.26.5",
 "goos": "js",
 "goarch": "wasm"
}
```

The package documents this metadata as real. `packages/wasm/host/host.go` declares the contract beside the variables:

```go
// Build metadata. Override at link time via
// `-ldflags "-X github.com/samchon/ttsc/packages/wasm/host.version=..."`.
```

`packages/wasm/README.md` publishes the recipe and states "The `version`, `commit`, and `date` variables in `host/host.go` are all overridable. `globalThis[apiName].version()` reads from them." `website/src/content/docs/wasm/index.mdx` lists `api.version()` in its API table as "Build metadata for support panels and bug reports", and its Version stamping section closes with "`api.version()` returns the stamped version, commit, date, Go version, GOOS, and GOARCH. Surface this in bug reports; wasm playground issues are otherwise hard to reproduce."

The stamping recipe itself is written for custom binaries, and that is exactly the gap: the same page tells readers to ship the base binary verbatim (`cp node_modules/@ttsc/wasm/dist/ttsc.wasm public/compiler/ttsc.wasm`), so the artifact most consumers deploy is the unstamped one, while the API table and the bug-report instruction apply to whatever binary they run. A support panel cannot tell one release from another, and the bug reports the documentation asks for carry no usable identity.

Affected: every consumer of the published `@ttsc/wasm`, and the ttsc.dev playground, whose wasm is built the same way. No runtime behavior depends on the strings, so the blast radius is diagnosis and support.

### Observed output, published artifact

Two reviewers independently booted the published `@ttsc/wasm@0.19.3` `dist/ttsc.wasm` in Node through the package's own boot helper and called `version()`. The first recorded the block quoted above; the second recorded the same payload in one line:

```text
VERSION {"goos":"js","goarch":"wasm","version":"0.0.0-dev","commit":"dev","date":"unknown","go":"go1.26.5"}
```

The native helper from the same repository, built by the stamping script, shows what the contract looks like when it is honored:

```text
$ packages/ttsc-win32-x64/bin/ttsc.exe --version
[...]

Issue #847, opened 2026-07-20
fix(wasm): resolve the token under the cursor in the fountain position verbs


samchon:
## Problem

The three position-addressed fountain verbs in `@ttsc/wasm` never see a token. `getNodeAtPosition`, `getTypeAtPosition`, and `getSymbolAtPosition` all resolve the position to the innermost **non-token** node, so an identifier, a literal, or a keyword under the cursor is never the node handed to the checker.

The published contracts say otherwise. `packages/wasm/src/structures/ITtscApi.ts` declares:

- `getNodeAtPosition`: "Return the deepest AST node whose range covers `position` (byte offset)."
- `getTypeAtPosition`: "Run the TypeScript-Go type checker against the node at `position` and return its printed type string + flags. Returns `{type: null}` when there is no node at that position."
- `getSymbolAtPosition`: "Resolve the symbol the node at `position` refers to, including up to 16 declaration sites. Returns `{symbol: null}` when no symbol is bound."

`packages/wasm/host/fountain.go` adds that `symbolInfoOf` carries declaration sites "so callers can implement 'go to definition' without follow-up snapshot queries", and its own comment says `{symbol: null}` means the node has no associated symbol, "e.g. punctuation, whitespace". `packages/wasm/README.md` sells the family as "`embed-typescript`-style raw access to the program, diagnostics, AST nodes, the type checker at a position".

In practice:

- the "deepest node" promise is false at every position inside a name, literal, or keyword;
- `getSymbolAtPosition` returns `null` at every declaration name and every plain identifier reference, so go-to-definition is impossible;
- when a symbol is returned it can be the wrong one, because the receiver and the property of a member access collapse to the same node;
- `getTypeAtPosition` reports the enclosing expression's type, so hover over an argument reports the call's return type;
- the documented `null` conditions never fire, while `type` is non-null at positions that hold no expression at all.

Affected: every embedder of the published `@ttsc/wasm` fountain API, which is precisely the hover, go-to-definition, and symbol-panel use case the guide advertises. The first-party playground does not call these verbs, so the damage is entirely on the published embedder surface.

### Root cause, from the sources

`packages/wasm/host/fountain.go` implements all three verbs on one call:

```go
    node := ast.GetNodeAtPosition(file, pos, false)
```

`packages/ttsc/shim/ast/shim.go` links that to the internal upstream helper and repeats the claim:

```go
// GetNodeAtPosition returns the deepest AST node whose range covers position
[...]

PR #892, opened 2026-07-20
fix(wasm,website): key Go build caches by effective inputs

### LINKED ISSUE #815: fix(wasm,website): key the Go build caches by every effective build input
## Problem

Two build scripts decide whether an existing WebAssembly binary is current, and both answer with the same approximation: the newest modification time of files whose name ends in `.go`, under a fixed list of directories.

- `packages/wasm/build/build-wasm.cjs` produces the published `packages/wasm/dist/ttsc.wasm`.
- `website/build/compiler.cjs` produces the ttsc.dev playground's `website/public/compiler/playground.wasm`.

A Go build depends on far more than `.go` files. `go.mod`, `go.sum`, the nested shim module manifests, `//go:embed` payloads, the build flags in the script itself, the Go toolchain, and the paired `wasm_exec.js` all change the produced binary and none of them is examined. The website script additionally omits an entire linked Go package from its root list.

When any omitted input changes after a successful build, the next build prints that the cache is up to date and ships the old binary. `build-wasm.cjs` makes the mismatch worse on a hit: it still refreshes the vendored shim tree and rewrites the published `dist/go.mod`, so the package's module metadata advertises the new dependency graph while `ttsc.wasm` was compiled against the old one.

Affected: local `pnpm --filter @ttsc/wasm build`, website `pnpm build` and `pnpm dev`, any persistent or reused CI workspace, and release preparation on a non-clean tree. A fresh CI checkout has no cached artifact and rebuilds, which limits release exposure but does not make the predicate correct.

### Reproduction, `packages/wasm`

An isolated copy of the baseline build script was given a module manifest one hour newer than the cached binary, with all `.go` sources older than both:

```text
packages/wasm/cmd/ttsc-wasm/main.go   2026-07-19T04:46:29.8456453Z
packages/wasm/dist/ttsc.wasm          2026-07-19T05:46:29.8514768Z
packages/wasm/go.mod                  2026-07-19T06:46:29.8514768Z
```

Running `node packages/wasm/build/build-wasm.cjs` produced:

```text
build/build-wasm.cjs: cached ttsc.wasm is up to date (0.00 MiB) — skipping rebuild
build/build-wasm.cjs: vendored shim/* → shim-vendor\shim
build/build-wasm.cjs: emitted published go.mod → dist\go.mod
before=2026-07-19T05:46:29.8514768Z
after=2026-07-19T05:46:29.8514768Z
```

The first three lines are the script's own output; `before=` and `after=` are the harness printing the artifact's modification time around the call. The binary is unchanged, and the two lines after the cache hit are the shim vendoring and the published `go.mod` rewrite that run anyway. The fixture's payload was a stub, so a real `go build` would have been unmistakable; none ran.

### Root cause, both scripts

The predicate is one filename suffix. `packages/wasm/build/build-wasm.cjs`:

```js
    } else if (stat.isFile() && cur.endsWith(".go")) {
      visit(cur, stat);
    }
```

```js
if (!forceRebuild && fs.existsSync(wasmOut)) {
  const wasmMtime = fs.statSync(wasmOut).mtimeMs;
  const sourceMtime = newestMtime(packageRoot, ttscDir);
  if (sourceMtime > 0 && wasmMtime >= sourceMtime) {
```

`website/build/compiler.cjs` duplicates the same walker and the same suffix test, and applies it to a fixed root list:

```js
const wasmSourceMtime = Math.max(
  newestMtime(COMPILER_DIR),
  newestMtime(path.join(REPO_ROOT, "packages", "wasm")),
  newestMtime(path.join(REPO_ROOT, "packages", "ttsc")),
  newestMtime(path.join(REPO_ROOT, "packages", "lint", "linthost")),
  newestMtime(path.join(REPO_ROOT, "packages", "lint", "rule")),
  newestMtime(TYPIA_NATIVE_ADAPTER),
);
```

Its own comment states the requirement it does not meet: "The cache key has to track every Go source the playground links in".

### The omitted inputs are real, not hypothetical

`//go:embed` payloads. `packages/lint/linthost/polyfill_data.go` embeds `polyfill_data_gen.json` and `packages/lint/linthost/rule_codes.go` embeds `rule_codes.json`. Both are compiled into the playground binary through `website/compiler/cmd/playground/plugins.go`, which imports `github.com/samchon/ttsc/packages/lint/linthost`. Regenerating either JSON changes the binary and moves no `.go` mtime.

A linked Go package outside every root. `packages/lint/linthost/rule_codes.go` imports `github.com/samchon/ttsc/packages/lint/internal/rulecode`, whose only source is `packages/lint/internal/rulecode/rulecode.go`. That path is under none of the six website roots, so editing it does not invalidate the playground cache at all. This is not a non-`.go` gap; it is a compiled Go source the key never sees.

Module identity. `packages/wasm/go.mod`, `packages/ttsc/go.mod` and `go.sum`, the 44 tracked `go.mod` / `go.sum` files including every `packages/ttsc/shim/*` manifest, and `website/compiler/go.mod` / `go.sum` with its `replace` directives for the shim tree, `packages/lint`, `packages/ttsc`, `packages/wasm`, and typia's native adapter, all select what is compiled. None ends in `.go`.

Toolchain and pairing. `build-wasm.cjs` copies `wasm_exec.js` on a cache hit only when the destination is missing, so a Go upgrade can leave a new bridge beside an old binary, or an old bridge beside a binary the cache calls current. The Go version and the `-trimpath -ldflags "-s -w"` flags in the script are likewise outside the key.

The website script's one sidecar stamp, `.typia-version`, proves the class was already recognized once: a typia version bump busts the cache because the binary does not carry the version. Every other dependency and embed input has the same property and no stamp.

### Provenance

Both walkers arrived with the website and wasm platform in `b1ddf02ec` / PR #83 and are unchanged at the audited commit. `cbc56cfe9` / PR #87 and `75f9e4a18` / PR #708 strengthened typia dependency consistency and version stamping without broadening the Go build input set. Searches across open and closed issues and pull requests for the wasm and playground build caches, `newestMtime`, `go:embed`, and stale wasm builds return no owner.

Verified against `2b724664eee71c57cd688c55122cac137774bf62`. `packages/wasm/build/build-wasm.cjs` and `website/build/compiler.cjs` are byte-identical on current `master` (`612abc2d2`), where the `newestMtime` root list is still the six entries quoted above and `packages/lint/internal/rulecode/rulecode.go` still exists outside them.

## Consequence surface

| Changed input | Current result after a prior build | Required |
| --- | --- | --- |
| Any `.go` file under a listed root | Rebuild | Rebuild |
| `packages/lint/internal/rulecode/rulecode.go` (website) | Cache hit | Rebuild |
| `polyfill_data_gen.json`, `rule_codes.json` (website) | Cache hit | Rebuild |
| `packages/wasm/go.mod` or `go.sum` | Cache hit | Rebuild |
| `packages/ttsc/go.mod` or `go.sum` | Cache hit | Rebuild |
| Nested `packages/ttsc/shim/*/go.mod` or `go.sum` | Cache hit | Rebuild |
| `website/compiler/go.mod` or `go.sum` | Cache hit | Rebuild |
| Build script, build flags, `GOOS`/`GOARCH` | Cache hit | Rebuild |
| Go toolchain version or source `wasm_exec.js` | Cache hit; the bridge is copied only when missing | Rebuild, and publish a coherent binary/bridge pair |
| typia version (website) | Rebuild, via the `.typia-version` sidecar | Rebuild |
| Unrelated TypeScript, docs, website content | Cache hit | Cache hit |
| Exact unchanged inputs | Cache hit | Cache hit |
| `--force` / `TTSC_WASM_FORCE` / `TTSC_PLAYGROUND_FORCE` | Rebuild | Rebuild |

Downstream: a stale `dist/ttsc.wasm` reaches every `@ttsc/wasm` consumer built from that workspace, and a stale `playground.wasm` serves the browser old lint rule codes, old polyfill data, an old typescript-go, or an old build composition. A mismatched `wasm_exec.js` and `.wasm` can fail at boot or at a syscall boundary while the build reports success. On a `build-wasm.cjs` hit the package still looks current at the source and module layer because the shim vendoring and published `go.mod` rewrite run regardless.

## Approach

The invariant: a cache hit is valid only when a deterministic identity of every effective input to the `go build` invocation matches the identity recorded beside the existing artifact, and `ttsc.wasm` / `playground.wasm`, `wasm_exec.js`, and that recorded identity always belong to the same build.

Owners: `packages/wasm/build/build-wasm.cjs` and `website/build/compiler.cjs`. They are two copies of one predicate, so the fix must not repair one and leave the other; whether the answer is a shared helper or two aligned implementations is an implementation choice.

Two shapes both satisfy the invariant: a content-addressed manifest written beside the outputs and compared on the next run, or an unconditional build with the caching delegated to Go's own build cache. Extending the extension allowlist is not a fix, because the next omitted input class returns the same failure. Asking Go for its own dependency set (`go list -deps -json`, or the equivalent) is one way to stop maintaining a hand-written root list, and the website script's missing `rulecode` root is the evidence that a hand-written list drifts.

Fail toward rebuilding whenever an input cannot be resolved. Preserve a cheap hit for a genuinely unchanged tree and the explicit force override.

| Case | Expected |
| --- | --- |
| No input changed | Cache hit, no `go build` |
| A `.go` source changed | Rebuild |
| `packages/lint/internal/rulecode/rulecode.go` changed | Website build rebuilds |
| An embedded JSON changed | Website build rebuilds |
| Any `go.mod` or `go.sum` in the build graph changed | Rebuild |
| Build script or flags changed | Rebuild |
| Go toolchain changed | Rebuild, and the shipped `wasm_exec.js` comes from that toolchain |
| Unrelated website or TypeScript content changed | Cache hit |
| Force flag or environment variable | Rebuild |
| Recorded identity unreadable or corrupt | Rebuild, never a hit |
| After any hit | `dist/go.mod`, the vendored shim, `wasm_exec.js`, and the binary describe one build |

**A cache defect cannot be proven by one build.** Every row above is a two-build sequence: build, mutate exactly one input, build again, and compare the artifact identity. A single invocation of either script passes against the defect, because the first build is always a miss and the second is the one that lies.

Neither script has any test today (`git ls-files` shows no test file referencing `build-wasm.cjs` or `website/build/compiler.cjs`), so the freshness decision should be extracted into a function that can be exercised over synthetic inputs without invoking Go, and the tests written against that. The website package already runs plain Node test files (`node --test test/typia-dependency-graph.test.cjs`), which is the nearest existing pattern; `tests/test-wasm` is the nearest one for `packages/wasm`.

Narrow proving commands, after adding the tests, and naming the file that is added so the command cannot silently run nothing:

```text
node --test website/test/<new-file>.test.cjs
pnpm --filter @ttsc/test-wasm start -- --include=<new case name>
```

`node --test` on a path that does not exist fails, and the `pnpm` runner matches test names by substring, prints each executed test by name, and fails loudly when a filter matches nothing (`No tests matched --include=<...>`, exit 1). Read the printed names anyway: a filter that happens to match some other existing case runs green without executing what you added.

Broader, and required at least once by hand because the isolated tests do not invoke Go: run `node packages/wasm/build/build-wasm.cjs`, record the artifact hash, touch `packages/wasm/go.mod`, run it again, and confirm the hash moved; repeat the same two-build sequence for the website `build:compiler` against `packages/lint/linthost/rule_codes.json`. Then the website compiler smoke path.

## Coordination

- This issue and the issue that stamps `@ttsc/wasm` build metadata edit the same two files, `packages/wasm/build/build-wasm.cjs` and `website/build/compiler.cjs`, and they interact: that issue must add `-X` linker values without defeating this cache, and this issue must fold whatever identity that stamp carries into the cache key. Sequence the two under one owner, **cache key first**. The order is not a preference. Stamping the binary while a stale cache can still skip the rebuild produces an unstamped artifact out of a build that reported success, which is this same silent-cache-hit defect wearing the other issue's clothes, and it would be reported as a stamping failure.
- The `.typia-version` sidecar in `website/build/compiler.cjs` is existing correct behavior for one input class and must keep working, whether it is absorbed into the new identity or left as is.
- `b1ddf02ec` / PR #83 is the causal history. PR #87 and PR #708 are adjacent typia-consistency work, not duplicates.
- File-disjoint from the `packages/wasm/src/**` and `packages/wasm/host/**` defects (MemFS tree invariants, MemFS descriptor semantics, the `pipe2` contract, token-precise position resolution); those can run in parallel.

samchon:
## Intent

Closes #815.

Replace the two timestamp-and-`.go` build-cache predicates with one
deterministic effective-input identity contract. A cache hit must pair the
recorded identity, the wasm binary, and `wasm_exec.js` from one build.

## Scope

- `packages/wasm/build/build-wasm.cjs`
- `website/build/compiler.cjs`
- focused cache-identity regressions for both builders
- the matching playground/wasm documentation only if the published behavior
  changes

`#847` owns the independent fountain-token path. `#848` depends on this cache
root and will not begin until this PR merges.

## Verification

Pending implementation. Campaign implementation runs are cancelled by exact
commit SHA; local two-build cache tests and builder smoke checks are the merge
gate.

### DIFF
--- a/packages/wasm/build/build-wasm.cjs
+++ b/packages/wasm/build/build-wasm.cjs
@@ -2,23 +2,15 @@
 //
 // Usage: node build/build-wasm.cjs [--force]
 //
-// Behavior:
-//   * Skips the Go build when `dist/ttsc.wasm` is newer than every .go file
-//     under `packages/{ttsc,wasm}/`. Pass `--force` or set
-//     `TTSC_WASM_FORCE=1` to rebuild unconditionally.
-//   * Hard failure if the Go toolchain is missing.
-//
-// Outputs:
-//   * packages/wasm/dist/ttsc.wasm
-//   * packages/wasm/dist/wasm_exec.js
-//
-// Downstream consumers (the website, plugin-author playgrounds) point their
-// own bundle pipelines at `node_modules/@ttsc/wasm/dist/wasm_exec.js` and
-// build their own .wasm against the same Go module.
+// The cache records a content identity for every effective Go dependency,
+// module manifest, build flag, toolchain bridge, and published shim artifact.
+// It never accepts a partially staged publication directory as a cache hit.
 
 const cp = require("child_process");
 const fs = require("fs");
 const path = require("path");
+
+const { createGoBuildCache } = require("../../../scripts/go-build-cache.cjs");
 
 const packageRoot = path.resolve(__dirname, "..");
 const repoRoot = path.resolve(packageRoot, "..", "..");
@@ -31,96 +23,60 @@ const wasmOut = path.join(outDir, "ttsc.wasm");
 const wasmExecOut = path.join(outDir, "wasm_exec.js");
 const goModPath = path.join(packageRoot, "go.mod");
 const publishedGoModOut = path.join(outDir, "go.mod");
-const forceRebuild =
-  process.argv.includes("--force") || !!process.env.TTSC_WASM_FORCE;
-
-fs.mkdirSync(outDir, { recursive: true });
-
-function hasGoToolchain() {
-  try {
-    cp.execSync("go env GOROOT", { stdio: "ignore" });
-    return true;
-  } catch {
-    return false;
-  }
-}
+const cachePath = path.join(outDir, ".ttsc-wasm-build.json");
+const buildArguments = [
+  "go",
+  "build",
+  "-trimpath",
+  "-ldflags",
+  "-s -w",
+  "-o",
+  wasmOut,
+  "./cmd/ttsc-wasm",
+];
+const buildEnvironment = { GOOS: "js", GOARCH: "wasm" };
 
 function locateWasmExec() {
-  if (!hasGoToolchain()) return null;
-  const goroot =
-    process.env.GOROOT ??
-    cp.execSync("go env GOROOT", { encoding: "utf8" }).trim();
+  let goroot;
+  try {
+    goroot =
+      process.env.GOROOT ??
+      cp.execFileSync("go", ["env", "GOROOT"], { encoding: "utf8" }).trim();
+  } catch {
+    return null;
+  }
   // Go 1.24+ ships wasm_exec.js under lib/wasm/. Older releases kept it
   // under misc/wasm/. Try both so the build works across Go installs.
-  const candidates = [
+  for (const candidate of [
     path.join(goroot, "lib", "wasm", "wasm_exec.js"),
     path.join(goroot, "misc", "wasm", "wasm_exec.js"),
-  ];
-  for (const candidate of candidates) {
+  ]) {
     if (fs.existsSync(candidate)) return candidate;
   }
   return null;
 }
 
-function newestMtime(...roots) {
-  let max = 0;
-  for (const root of roots) {
-    if (!root || !fs.existsSync(root)) continue;
-    walk(root, (_file, stat) => {
-      if (stat.mtimeMs > max) max = stat.mtimeMs;
-    });
-  }
-  return max;
-}
-
-function walk(root, visit) {
-  const stack = [root];
-  while (stack.length) {
-    const cur = stack.pop();
-    let stat;
-    try {
-      stat = fs.statSync(cur);
-    } catch {
-      continue;
-    }
-    if (stat.isDirectory()) {
-      const base = path.basename(cur);
-      if (
-        base === "node_modules" ||
-        base === ".ttsc" ||
-        base === ".git" ||
-        base === "lib" ||
-        base === "dist"
-      )
-        continue;
-      for (const entry of fs.readdirSync(cur)) {
-        stack.push(path.join(cur, entry));
-      }
-    } else if (stat.isFile() && cur.endsWith(".go")) {
-      visit(cur, stat);
-    }
-  }
+function createCacheOptions({ force, wasmExecSrc }) {
+  return {
+    artifactPaths: [wasmOut, wasmExecOut, publishedGoModOut, vendorShimDir],
+    buildArguments,
+    cachePath,
+    cwd: packageRoot,
+    dependencyPackages: ["./cmd/ttsc-wasm"],
+    environment: buildEnvironment,
+    extraFiles: [__filename, goModPath, wasmExecSrc],
+    force,
+    inputDirectories: [shimSrc],
+  };
 }
 
 function buildWasm() {
   console.log(`build/build-wasm.cjs: building ${wasmOut}`);
-  cp.execFileSync(
-    "go",
-    [
-      "build",
-      "-trimpath",
-      "-ldflags",
-      "-s -w",
-      "-o",
-      wasmOut,
-      "./cmd/ttsc-wasm",
-    ],
-    {
-      cwd: packageRoot,
-      env: { ...process.env, GOOS: "js", GOARCH: "wasm" },
-      stdio: "inherit",
-    },
-  );
+  cp.execFileSync("go", buildArguments.slice(1), {
+    cwd: packageRoot,
+    env: { ...process.env, ...buildEnvironment },
+    stdio: "inherit",
+  });
 }
 
 // Mirror packages/ttsc/shim/* into packages/wasm/shim-vendor/shim/* so the
@@ -141,7 +97,7 @@ function vendorShim() {
   fs.mkdirSync(vendorShimDir, { recursive: true });
   copyTree(shimSrc, vendorShimDir);
   console.log(
-    `build/build-wasm.cjs: vendored shim/* → ${path.relative(packageRoot, vendorShimDir)}`,
+    `build/build-wasm.cjs: vendored shim/* -> ${path.relative(packageRoot, vendorShimDir)}`,
   );
 }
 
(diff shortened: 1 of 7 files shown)


samchon:
Self-review found one missing regression dimension: the cache fixture proves source, manifests, bridge, flags, targets, outputs, and corrupt records, but does not yet vary the reported Go toolchain version. I will add that assertion so a toolchain upgrade cannot regress to a cache hit.

samchon:
Second Self-review found a publication-boundary defect before merge: @ttsc/wasm publishes build/ but not the repository-root scripts/ directory, so the new build script cannot require a root helper from an installed tarball. I will relocate the single shared helper into packages/wasm/build/ (a published path) and make the website consume that same module.

samchon:
Fresh self-review found two inherited Go environment inputs not yet represented by the shared WASM cache identity: GOWASM can change js/wasm output, and GOTOOLCHAIN can select a different compiler. I will add both effective values and regression coverage, then restart the complete review and verification cycle.

samchon:
Fresh final verification found a merge consequence from #885: website/compiler/go.mod mirrors the local shim module replacements but omits the newly required shim/astnav, so go list -deps ./cmd/playground cannot resolve the fountain dependency. This blocks the website cache identity itself. I will add the matching local replace and indirect requirement, regenerate only that module's sum, and rerun the real website builder.

samchon:
Self-review follow-up before final rebase:

- `go.work.sum` participates in Go workspace resolution, so the cache identity now hashes it beside an active `GOWORK` file.
- Hashing only the `go` driver is insufficient for a same-version toolchain replacement. The identity now hashes the complete `GOTOOLDIR` content, covering the compile/link programs the build invokes, without retaining raw machine paths.

Focused `node --test scripts/go-build-cache.test.cjs scripts/go-build-cache-builders.test.cjs` passes with regressions for both boundaries. These commits are local pending the required #900 dependency merge, then will be rebased and force-with-lease pushed once with the final validation evidence.

samchon:
Self-review follow-up before the final push:

`website/build/compiler.cjs` now writes `public/compiler/.playground-wasm-build.json`, but `website/.gitignore` did not exclude that generated cache record. A normal compiler build therefore left an untracked file in the worktree. I will add that exact sidecar to the adjacent generated-output ignore list, then rerun the focused cache tests and website build on the frozen snapshot.

samchon:
Final self-review and local verification are complete on `4a85e3dad26f74121934f0200793fb767a032b73`, rebased onto `336b3255`.

- The protected Graph file/tree guards remain unchanged.
- `node --test scripts/go-build-cache.test.cjs scripts/go-build-cache-builders.test.cjs` passes.
- Actual `@ttsc/wasm` builds prove miss, unchanged hit, module-identity rebuild, artifact-pair hashing, and restoration.
- The website compiler completes after its documented `@ttsc/playground` prerequisite; an embedded `rule_codes.json` ledger change rebuilds `playground.wasm` (`10cdee...`) and exact restoration rebuilds the baseline (`b588...`).
- `pnpm --filter @ttsc/test-wasm start` passes all 29 cases, and the full scoped `test:go` runner exits 0 after rebuilding the reboot-missing `ttsx` launcher.
- The final round found and fixed the generated website cache record's missing ignore entry. No further issue remains in the #815 surface.

Campaign Actions for this SHA are deliberately cancelled after exact-SHA readback; the local evidence and cancellation records are the merge gate.

Issue #854, opened 2026-07-20
fix(release): the VS Code extension is not on the Marketplace, and no post-publish check notices


samchon:
## Problem

Three shipped documents tell users to install the VS Code extension from the Visual Studio Marketplace. `samchon.ttsc` does not exist in the gallery, so every one of those instructions fails for every user.

The release pipeline runs `scripts/publish-vscode-marketplace.sh` on every tag, `vsce publish` returns success, and the workflow step is recorded green. Nothing after that asks whether the version it just published is actually served. The npm channel does have that check: `release.yml` has `wasm-smoke` and `vscode-smoke` jobs that reinstall the just-published packages from the registry and exercise them. The Marketplace channel has no equivalent, which is why a silent no-op has repeated across every release since the step was added.

Two things are wrong and both belong in this issue:

1. **The advertised install path does not work.** Documentation and the actual distribution channels disagree. Which way that disagreement is resolved, by restoring the listing or by dropping the Marketplace from the documentation, is a product decision for the maintainer. This issue does not take a side and presents both paths with their consequences below.
2. **Nothing would have caught it.** A publish that reports success is treated as a publish that happened. This half is a defect under either product decision: a release that claims to publish to a channel and does not must fail, and today it passes.

The bundled-VSIX path does work and is smoke-tested. The extension is reachable; only the channel the documentation recommends first is dead.

### The gallery is empty

Queried five ways. Exact-name filter (`filterType: 7` is the same filter VS Code's install-by-id uses):

```text
$ curl -s -X POST https://marketplace.visualstudio.com/_apis/public/gallery/extensionquery \
    -H 'Accept: application/json;api-version=7.2-preview.1' -H 'Content-Type: application/json' \
    -d '{"filters":[{"criteria":[{"filterType":7,"value":"samchon.ttsc"}],"pageSize":5}],"flags":914}'
{"results":[{"extensions":[],"pagingToken":null,"resultMetadata":[{"metadataType":"ResultCount","metadataItems":[{"name":"TotalCount","count":0}]}]}]}
```

The same query for `esbenp.prettier-vscode` returns one result (`prettier-vscode 12.4.0`), so the probe is sound.

```text
# every extension owned by the publisher (filterType 18 = publisher name)
$ ... '{"filters":[{"criteria":[{"filterType":18,"value":"samchon"}],"pageSize":50}],"flags":914}'
count 0

# free-text search
[...]

PR #893, opened 2026-07-20
fix(release): verify Marketplace versions before npm publish

samchon:
## Intent

Make a release fail closed unless the advertised VS Code Marketplace channel is publicly reachable before any publication and serves the exact tagged version after `vsce publish`.

## Verified premise

Anonymous exact-name Gallery queries currently return no `samchon.ttsc` extension, while `esbenp.prettier-vscode` resolves with its full version history. The release workflow records `vsce publish` success but never verifies the public result.

The available evidence does **not** establish that every historical Marketplace publish was a silent no-op. The extension may have been public and later unpublished or hidden. This pull request owns the current public absence and the missing release-safety invariant, not that stronger historical claim.

## Scope

- Add a dependency-free Marketplace probe with strict response and identifier validation, exact extension/version matching, bounded retry, and fail-closed errors.
- Query all served versions so a legitimate `--skip-duplicate` retry of an older public version succeeds even when a newer version exists.
- Check that the established public channel is reachable before external publication.
- Check that the exact tagged version is publicly served after Marketplace publication and before npm publication.
- Add deterministic positive, negative, malformed, transient-failure, deadline, and workflow-ordering coverage.

## Deferred

- Restoring, publishing, unpublishing, or inspecting the extension in the owner-only Marketplace portal.
- Choosing whether the Marketplace channel is retained or removed.
- Changing Marketplace installation documentation.
- Closing #854; its product decision remains open.

## Verification

- `pnpm --filter @ttsc/test-ttsc start -- '--include=marketplace_probe,release_workflow_runs_preflight_before_publication'` — 7/7 passed.
- Changed TypeScript files passed an isolated strict `tsc --noEmit` check.
- `pnpm run assert:project-layout` — passed.
- `node --check scripts/assert-marketplace-version.cjs` — passed.
- Public control: `esbenp.prettier-vscode@12.3.0` passed, proving historical served versions are visible.
- Current negative: `samchon.ttsc` remained absent and failed after the bounded retry window.
- `pnpm run test:typecheck` could not complete because the existing unbuilt `@ttsc/graph` declarations are absent; it failed before reaching `test-ttsc` and reported only those unrelated imports.

Refs #854

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -49,6 +49,8 @@ jobs:
         run: pnpm install --frozen-lockfile
       - name: Preflight - validate release tag and package versions
         run: node scripts/release-preflight.cjs
+      - name: Preflight - verify public Marketplace channel
+        run: node scripts/assert-marketplace-version.cjs --extension samchon.ttsc --timeout-ms 60000 --interval-ms 5000
       - name: Build all packages
         run: pnpm run build
       - name: Assert dist/ttsc.wasm exists and is >10MB
@@ -79,6 +81,8 @@ jobs:
         run: SKIP_BUILD=1 bash scripts/publish-vscode-marketplace.sh
         env:
           VSCE_PAT: ${{ secrets.VSCE_PAT }}
+      - name: Verify VS Code Marketplace serves release
+        run: node scripts/assert-marketplace-version.cjs --extension samchon.ttsc --version "${GITHUB_REF_NAME#v}" --timeout-ms 300000 --interval-ms 10000
       - name: Publish to npm
         run: pnpm run package:latest:publish
   wasm-smoke:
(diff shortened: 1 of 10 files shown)


samchon:
Implementation and Self-Review chronology:

- The first complete review found that Gallery flags `914` include `IncludeLatestVersionOnly`. That would reject a legitimate `vsce publish --skip-duplicate` retry of an older still-served version whenever a newer version exists.
- The probe now requests `IncludeVersions` only (`flags: 1`). A live anonymous control proved `esbenp.prettier-vscode@12.3.0` is found beneath the latest `12.4.0`, and the deterministic test pins that query flag.
- The same round tightened extension-id parsing to the public `<publisher>.<name>` grammar and added malformed-input coverage.
- After those changes, a fresh complete review of the current base-to-head diff found no further improvement. The final immutable `17bcb9b3` snapshot passed 7 focused feature tests, changed-file strict TypeScript checking, `assert:project-layout`, `node --check`, and `git diff --check` on current `master`.
- The implementation-wave Actions for each pushed SHA were cancelled by exact commit as required by the campaign workflow; local verification and solo review are the merge gates.

The Marketplace portal/channel decision remains deliberately deferred, so #854 must stay open after this safety-half PR merges.

PR #894, opened 2026-07-20
fix(ttsc): honor no-emit in single-file mode

### LINKED ISSUE #817: fix(ttsc): watch the compiler's resolved input topology instead of a one-time directory snapshot

### LINKED ISSUE #855: fix(ttsc): honor `--noEmit`, `check`, `--emit=false`, and tsconfig `noEmit` in single-file mode

samchon:
## Reserved issue

Closes #855.

## Scope

Reserve the single-file `ttsc` emit-decision boundary. The implementation will preserve the temporary compiler output needed for diagnostics and transformed text, while suppressing every final user-tree write and emitted-path line when the effective decision is no-emit. It will carry CLI boolean forms, `check`, and resolved `tsconfig` `noEmit` through the normal and watch single-file callers.

## Verification plan

The real `native-plugins/compiler` lane will cover each no-write invocation, the no-`outDir` sibling boundary, diagnostic preservation, both live watch forms, and the explicit `--emit` override. Existing plain and explicit-`--outDir` single-file emit tests remain positive twins. Package build and the targeted compiler corpus are pending implementation.

## Coordination

PR #887 resolved #817 before this reservation. This PR does not include #852, which depends on this no-emit behavior, and does not change graph, plugin protocol, workflow, or Actions configuration. Campaign implementation CI is deliberately cancelled only by exact head SHA; local verification and solo self-review are the merge gates.

### DIFF
--- a/packages/ttsc/src/launcher/internal/runTtsc.ts
+++ b/packages/ttsc/src/launcher/internal/runTtsc.ts
@@ -109,7 +109,7 @@ type TtscMode = "build" | "check" | "fix" | "format";
 
 function runCompatibleBuild(argv: readonly string[], mode: TtscMode): number {
   const checkOnly = mode !== "build";
-  const options = normalizeBuildOptions(parseBuildArgs(argv, checkOnly));
+  const options = normalizeBuildOptions(parseBuildArgs(argv));
   if (mode === "fix") {
     if (options.emit === true) {
       throw new Error("ttsc: fix and --emit are mutually exclusive");
@@ -137,16 +137,17 @@ function runCompatibleBuild(argv: readonly string[], mode: TtscMode): number {
     }
     return runWatch(options, checkOnly);
   }
-  if (options.files.length !== 0) {
+  const buildOptions = checkOnly ? { ...options, emit: false } : options;
+  if (buildOptions.files.length !== 0) {
     if (mode === "fix") {
       throw new Error("ttsc: fix requires a project, not single-file mode");
     }
     if (mode === "format") {
       throw new Error("ttsc: format requires a project, not single-file mode");
     }
-    return runSingleFile(options);
+    return runSingleFile(buildOptions);
   }
-  const result = runBuild(checkOnly ? { ...options, emit: false } : options);
+  const result = runBuild(buildOptions);
   if (result.stdout) process.stdout.write(result.stdout);
   if (result.stderr) process.stderr.write(result.stderr);
   return result.status;
@@ -414,7 +415,7 @@ function parseProjectArgs(argv: readonly string[]) {
   };
 }
 
-function parseBuildArgs(argv: readonly string[], checkOnly: boolean) {
+function parseBuildArgs(argv: readonly string[]) {
   const result = parseFlags({
     argv,
     errorPrefix: "ttsc:",
@@ -427,17 +428,15 @@ function parseBuildArgs(argv: readonly string[], checkOnly: boolean) {
     subcommand: "build",
   });
   // Defaults: pinned by the previous hand-parser. `quiet` defaults true,
-  // `--verbose` flips it to false; `emit` defaults `undefined` in build
-  // mode (let tsconfig decide) and `false` in check/fix/format mode.
+  // `--verbose` flips it to false; `emit` defaults `undefined` so the resolved
+  // project controls ordinary build mode. `runCompatibleBuild` applies the
+  // check/fix/format no-emit decision before either execution lane runs.
   const verbose = getBoolean(result, "--verbose");
   const quietFlag = getBoolean(result, "--quiet");
   const quiet = verbose === true ? false : (quietFlag ?? true);
   const explicitEmit = getBoolean(result, "--emit");
   const explicitNoEmit = getBoolean(result, "--noEmit");
-  let emit: boolean | undefined;
-  if (explicitEmit === true) emit = true;
-  else if (explicitNoEmit === true) emit = false;
-  if (emit === undefined && checkOnly) emit = false;
+  const emit = resolveExplicitEmit(explicitEmit, explicitNoEmit);
 
   // `isPositional: looksLikeInputFile` guarantees every `result.positional`
   // token is a TypeScript input file; forwarded flag values already live in
@@ -464,6 +463,21 @@ function parseBuildArgs(argv: readonly string[], checkOnly: boolean) {
     tsconfig: getString(result, "--tsconfig"),
     watch: getBoolean(result, "--watch") === true,
   };
+}
+
+/**
+ * Collapse the two launcher-owned emit switches into the tri-state consumed by
+ * `runBuild` and the single-file lane. A specified boolean is significant even
+ * when it is `false`: `--emit=false` is analysis-only and `--noEmit=false`
+ * explicitly overrides a project's `noEmit`. `--emit` retains precedence when
+ * callers supply both switches, matching the legacy true-only resolution.
+ */
+function resolveExplicitEmit(
+  explicitEmit: boolean | undefined,
+  explicitNoEmit: boolean | undefined,
+): boolean | undefined {
+  if (explicitEmit !== undefined) return explicitEmit;
+  return explicitNoEmit === undefined ? undefined : !explicitNoEmit;
 }
 
 /**
@@ -557,12 +571,15 @@ function runSingleFile(options: ReturnType<typeof parseBuildArgs>): number {
   }
   const cwd = path.resolve(options.cwd ?? process.cwd());
   const file = path.resolve(cwd, options.files[0]!);
-  const out = resolveSingleFileOut({
-    cliOutDir: options.outDir,
-    cwd,
-    file,
-    tsconfig: options.tsconfig,
-  });
+  const emit = singleFileShouldEmit(options, cwd, file);
+  const out = emit
+    ? resolveSingleFileOut({
+        cliOutDir: options.outDir,
+        cwd,
+        file,
+        tsconfig: options.tsconfig,
+      })
+    : undefined;
   const text = runSingleFileEmit({
     binary: options.binary,
     checkers: options.checkers,
@@ -573,12 +590,35 @@ function runSingleFile(options: ReturnType<typeof parseBuildArgs>): number {
     singleThreaded: options.singleThreaded,
     tsconfig: options.tsconfig,
   });
-  if (!fs.existsSync(out)) {
+  if (out !== undefined && !fs.existsSync(out)) {
     fs.mkdirSync(path.dirname(out), { recursive: true });
     fs.writeFileSync(out, text, "utf8");
   }
-  process.stdout.write(`${path.relative(cwd, out) || path.basename(out)}\n`);
+  if (out !== undefined) {
+    process.stdout.write(`${path.relative(cwd, out) || path.basename(out)}\n`);
+  }
   return 0;
+}
+
+/**
+ * Resolve the same effective emit decision used by the project lane before
+ * entering the single-file compatibility path. The compatibility path still
+ * emits into a private temporary directory to obtain transformed text and
+ * diagnostics, but only this boundary is allowed to write into the user's
+ * tree.
+ */
+function singleFileShouldEmit(
+  options: ReturnType<typeof parseBuildArgs>,
+  cwd: string,
+  file: string,
+): boolean {
+  if (options.emit !== undefined) return options.emit;
+  const project = readProjectConfig({
+    cwd,
+    file,
+    tsconfig: options.tsconfig,
+  });
+  return project.compilerOptions.noEmit !== true;
 }
 
 function resolveSingleFileOut(opts: {
@@ -683,7 +723,7 @@ function runWatch(
   const cwd = path.resolve(options.cwd ?? process.cwd());
   let topology: WatchTopology | undefined;
   const invocation = {
-    ...options,
+    ...(checkOnly ? { ...options, emit: false } : options),
     cwd,
     onWatchInputs: (inputs: readonly string[]) => {
       topology?.setExtraInputs(inputs);
--- a/tests/test-ttsc/src/native-plugins/compiler/test_compiler_corpus_explicit_emit_booleans_control_project_mode.ts
+++ b/tests/test-ttsc/src/native-plugins/compiler/test_compiler_corpus_explicit_emit_booleans_control_project_mode.ts
@@ -1,0 +1,53 @@
+import {
+  assert,
+  commonJsProject,
+  fs,
+  path,
+  spawn,
+  ttscBin,
+} from "../../internal/compiler-corpus";
+
+/**
+ * Verifies compiler corpus: explicit emit booleans control project mode.
+ *
+ * The launcher consumes `--emit` and `--noEmit` before invoking the project
+ * lane, so dropping an explicit `false` silently changes the build decision.
+ * These twins prove both false forms remain meaningful after parsing.
+ *
+ * 1. Run `--emit=false` against a normally emitting project.
+ * 2. Run `--noEmit=false` against a `noEmit` project.
+ * 3. Assert the first writes nothing and the second explicitly restores output.
+ */
+export const test_compiler_corpus_explicit_emit_booleans_control_project_mode =
+  (): void => {
+    const disabledRoot = commonJsProject({
+      "src/main.ts": `export const value: number = 1;\n`,
+    });
+    const disabled = spawn(
+      ttscBin,
+      ["--cwd", disabledRoot, "--emit=false"],
+      { cwd: disabledRoot },
+    );
+    assert.equal(disabled.status, 0, disabled.stderr);
+    assert.equal(
+      fs.existsSync(path.join(disabledRoot, "dist", "main.js")),
+      false,
+    );
+
+    const enabledRoot = commonJsProject(
+      {
+        "src/main.ts": `export const value: number = 1;\n`,
+      },
+      { compilerOptions: { noEmit: true } },
+    );
+    const enabled = spawn(
+      ttscBin,
+      ["--cwd", enabledRoot, "--noEmit=false"],
+      { cwd: enabledRoot },
+    );
+    assert.equal(enabled.status, 0, enabled.stderr);
+    assert.equal(
+      fs.existsSync(path.join(enabledRoot, "dist", "main.js")),
+      true,
+    );
+  };
--- a/tests/test-ttsc/src/native-plugins/compiler/test_compiler_corpus_single_file_honors_tsconfig_noemit_without_outdir.ts
+++ b/tests/test-ttsc/src/native-plugins/compiler/test_compiler_corpus_single_file_honors_tsconfig_noemit_without_outdir.ts
@@ -1,0 +1,52 @@
+import {
+  assert,
+  createProject,
+  fs,
+  path,
+  spawn,
+  ttscBin,
+} from "../../internal/compiler-corpus";
+
+/**
+ * Verifies compiler corpus: single-file mode honors tsconfig noEmit without outDir.
+ *
+ * Without an `outDir`, the compatibility lane's fallback would write a JavaScript
+ * sibling beside the source. A resolved `noEmit` must suppress that write, while
+ * explicit `--emit` retains the project lane's documented override.
+ *
+ * 1. Materialize a no-emit project without an output directory.
+ * 2. Run one positional file and assert its JavaScript sibling is absent.
+ * 3. Repeat with `--emit` and assert the sibling is intentionally written.
+ */
+export const test_compiler_corpus_single_file_honors_tsconfig_noemit_without_outdir =
+  (): void => {
+    const root = createProject({
+      "tsconfig.json": JSON.stringify({
+        compilerOptions: {
+          module: "commonjs",
+          noEmit: true,
+          rootDir: "src",
+          strict: true,
+          target: "ES2022",
+        },
+        include: ["src"],
+      }),
+      "src/main.ts": `export const value: number = 1;\n`,
+    });
+    const output = path.join(root, "src", "main.js");
+
+    const suppressed = spawn(ttscBin, ["--cwd", root, "src/main.ts"], {
+      cwd: root,
+    });
+    assert.equal(suppressed.status, 0, suppressed.stderr);
+    assert.equal(fs.existsSync(output), false);
+    assert.equal(suppressed.stdout.includes("main.js"), false);
+
+    const override = spawn(
+      ttscBin,
+      ["--cwd", root, "--emit", "src/main.ts"],
+      { cwd: root },
+    );
+    assert.equal(override.status, 0, override.stderr);
+    assert.equal(fs.existsSync(output), true);
+  };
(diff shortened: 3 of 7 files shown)


samchon:
Solo self-review found one documentation boundary: the first wording implied that `--noEmit=false` restores output under every subcommand, but `ttsc check` is intentionally analysis-only even with an explicit emit flag. I narrowed the documentation to the `ttsc` and `ttsc build` lanes; the executable behavior and regression matrix remain unchanged. I will repeat the full review after this follow-up snapshot.

samchon:
Narrow native compiler corpus found a test-wiring defect before execution: the new watch case imported `commonJsProject` from `internal/toolchain`, which intentionally does not export it. The real process controls remain in `toolchain`; fixture construction will move to the existing `compiler-corpus` helper. I will amend only that test import boundary, then rerun the complete native compiler corpus and a fresh solo review.

Issue #835, opened 2026-07-20
fix(graph): drop git-ignored generated code from the `ttsc-graph view` payload, as the guide and the producer both promise


samchon:
## Problem

`packages/graph/src/reduce.ts` is the reducer that `ttsc-graph view` runs in Node before serving the payload to the bundled 3D viewer. It filters `external` nodes only. It does not filter `ignored` nodes, its local `RawNode` interface does not even declare the field, and its `counts` has no `droppedIgnored`.

Both sibling copies of the same pure transform do filter it, and three independent contracts say the payload should:

- the producer's own doc comment in `packages/ttsc/internal/graph/gitignore.go` says "the full-graph dump drops them from the viewer payload entirely";
- the shipped guide `website/src/content/docs/graph/viewer.mdx` lists, under "What is left out", "**Git-ignored generated code is dropped.**", and says explicitly that the reduction happens "in the browser for an uploaded dump, in Node for `view`";
- the website reducer and the benchmark reducer both apply `(keepExternal || !n.external) && (keepIgnored || !n.ignored)`.

So on any project that emits generated TypeScript into the source tree (a Prisma client, GraphQL codegen, protobuf output), `ttsc-graph view` renders that codegen, while the same dump uploaded to the website viewer does not.

The node cap makes it worse rather than containing it. The payload is capped by **degree**, and generated clients are large and densely connected, so they preferentially survive the cap and displace the authored graph. That is precisely the failure `gitignore.go` names as its reason for existing.

Expected behavior: the three copies of the reduction apply one policy, and the `view` payload matches the published contract.

Affected: every `ttsc-graph view` user whose project contains git-ignored generated `.ts`.

## Evidence

All anchors below are unchanged between the audited commit `2b724664eee71c57cd688c55122cac137774bf62` and `master` at `612abc2d20d3f99429319081c1f6520bbe34ac7a`.

### The divergence

`packages/graph/src/reduce.ts::reduce`:

```ts
const keptByExternal = raw.nodes.filter((n) => keepExternal || !n.external);
const projectFiles = raw.nodes.filter((n) => !n.external).map((n) => n.file);
```

`website/src/components/graph/TtscWebsiteGraphReduce.ts::reduce`:

```ts
const keep = (n: RawNode) =>
  (keepExternal || !n.external) && (keepIgnored || !n.ignored);
const keptBoundary = raw.nodes.filter(keep);
const projectFiles = raw.nodes
  .filter((n) => !n.external && !n.ignored)
[...]

PR #903, opened 2026-07-20
fix(graph): recover structured identity components

### LINKED ISSUE #810: fix(graph): recover Graph identity from structured facts, not from splitting a display id on `#` or `.`

samchon:
## Intent

Resolve #810 by recovering Graph identity components from structured facts
rather than ambiguous `#` and `.` display-string splits.

## Scope

- Reserve the canonical Graph id codec, native parser, TypeScript memory, and
  viewer-rewrite surface.
- Reserve owner derivation and its native, LSP, memory, and viewer consumers.
- Add issue-specific regression coverage after the implementation snapshot.

## Deferred

- #822 and #835 remain a separate post-#810 combined reducer batch.

## Verification

This is an implementation-free claim. Local verification is pending; campaign
Actions runs for this reservation are cancelled by exact commit SHA.

### DIFF
--- a/experimental/benchmark/graph/viewer.mjs
+++ b/experimental/benchmark/graph/viewer.mjs
@@ -101,8 +101,9 @@ function relativize(abs, root) {
 }
 
 /**
- * A node id is `<path>#<name>:<kind>`; rewrite only the path prefix so ids stay
- * a stable key and every edge endpoint (also an id) relativizes identically.
+ * A node id quotes `#` and `\\` inside its `<path>#<name>:<kind>` components.
+ * Rewrite only the decoded path so ids stay stable keys and every edge endpoint
+ * (also an id) relativizes identically.
  */
 function rewriteId(id, root) {
   const hash = graphNodeIdHash(id);
--- a/experimental/benchmark/graph/viewer.mjs
+++ b/experimental/benchmark/graph/viewer.mjs
@@ -105,9 +105,47 @@ function relativize(abs, root) {
  * a stable key and every edge endpoint (also an id) relativizes identically.
  */
 function rewriteId(id, root) {
-  const hash = id.indexOf("#");
+  const hash = graphNodeIdHash(id);
   if (hash < 0) return id;
-  return relativize(id.slice(0, hash), root) + id.slice(hash);
+  return (
+    escapeGraphNodeIdPart(relativize(unescapeGraphNodeIdPart(id.slice(0, hash)), root)) +
+    id.slice(hash)
+  );
+}
+
+function escapeGraphNodeIdPart(value) {
+  return value.replaceAll("\\", "\\\\").replaceAll("#", "\\#");
+}
+
+function unescapeGraphNodeIdPart(value) {
+  let result = "";
+  for (let index = 0; index < value.length; index++) {
+    const next = value[index + 1];
+    if (value[index] === "\\" && next !== undefined) {
+      if (next === "#" || (next === "\\" && !legacyUNCStart(value, index))) {
+        result += next;
+        index++;
+        continue;
+      }
+    }
+    result += value[index];
+  }
+  return result;
+}
+
+function legacyUNCStart(value, index) {
+  return index === 0 && value.length > 2 && value[2] !== "\\" && value[2] !== "#";
+}
+
+function graphNodeIdHash(id) {
+  for (let index = 0; index < id.length; index++) {
+    if (id[index] !== "#") continue;
+    let slashes = 0;
+    for (let slash = index - 1; slash >= 0 && id[slash] === "\\"; slash--)
+      slashes++;
+    if (slashes % 2 === 0) return index;
+  }
+  return -1;
 }
 
 /**
--- a/packages/graph/src/model/TtscGraphMemory.ts
+++ b/packages/graph/src/model/TtscGraphMemory.ts
@@ -5,6 +5,7 @@ import { ITtscGraphNode } from "../structures/ITtscGraphNode";
 import { ITtscGraphSpan } from "../structures/ITtscGraphSpan";
 import { TtscGraphEdgeKind } from "../structures/TtscGraphEdgeKind";
 import { TtscGraphSourceReader } from "./TtscGraphSourceReader";
+import { ttscGraphNodeIdPath } from "./TtscGraphNodeId";
 
 /**
  * The in-memory resident graph the MCP tools answer from.
@@ -119,10 +120,20 @@ function keyOf(node: ITtscGraphNode): string {
   return node.qualifiedName ?? node.name;
 }
 
-/** The owner key of a dotted key (`A.B.c` -> `A.B`), or "" for a top-level key. */
-function ownerKey(key: string): string {
-  const dot = key.lastIndexOf(".");
-  return dot >= 0 ? key.slice(0, dot) : "";
+/**
+ * The owner key derived from facts the producer serialized separately.
+ *
+ * A quoted member named `"a.b"` has Name `a.b` and QualifiedName `Box.a.b`.
+ * Cutting the qualified name at its last dot invents owner `Box.a`; removing
+ * the exact `.${name}` suffix instead preserves the producer's real boundary.
+ */
+function ownerKey(node: ITtscGraphNode): string | undefined {
+  if (node.qualifiedName === undefined || node.qualifiedName === node.name)
+    return undefined;
+  const suffix = `.${node.name}`;
+  if (!node.qualifiedName.endsWith(suffix)) return undefined;
+  const owner = node.qualifiedName.slice(0, -suffix.length);
+  return owner === "" ? undefined : owner;
 }
 
 /** A file's id and node name from its project-relative path. */
@@ -144,8 +155,7 @@ function spanIn(span: ITtscGraphSpan, file: string): ITtscGraphEvidence {
  * file node's id is the path itself.
  */
 function fileOfNodeId(id: string): string {
-  const hash = id.indexOf("#");
-  return hash === -1 ? id : id.slice(0, hash);
+  return ttscGraphNodeIdPath(id) ?? id;
 }
 
 function basename(file: string): string {
@@ -213,8 +223,8 @@ function synthesize(dump: ITtscGraphDump): {
     if (!node.external) byFileKey.set(node.file + "\0" + keyOf(node), node);
   }
   const owner = (node: ITtscGraphNode): ITtscGraphNode | undefined => {
-    const parent = ownerKey(keyOf(node));
-    if (parent === "") return undefined;
+    const parent = ownerKey(node);
+    if (parent === undefined) return undefined;
     return byFileKey.get(node.file + "\0" + parent);
   };
 
--- a/packages/graph/src/model/TtscGraphNodeId.ts
+++ b/packages/graph/src/model/TtscGraphNodeId.ts
@@ -19,7 +19,7 @@ export function parseTtscGraphNodeId(
   const tail = id.slice(hash + 1);
   if (tail === "") return undefined;
   const colon = tail.lastIndexOf(":");
-  if (colon === tail.length - 1) return undefined;
+  if (colon === 0 || colon === tail.length - 1) return undefined;
   return {
     path: unescapeGraphNodeIdPart(id.slice(0, hash)),
     name: unescapeGraphNodeIdPart(colon < 0 ? tail : tail.slice(0, colon)),
--- a/packages/graph/src/model/TtscGraphNodeId.ts
+++ b/packages/graph/src/model/TtscGraphNodeId.ts
@@ -1,0 +1,82 @@
+/** The structured components carried by a graph symbol id. */
+export interface ITtscGraphNodeId {
+  path: string;
+  name: string;
+  kind?: string;
+}
+
+/**
+ * Decode the graph's position-invariant `path#name:kind` identity.
+ *
+ * The producer quotes `#` and `\\` inside path and name. This reader also
+ * accepts pre-codec ordinary ids, so a current package can read an older dump.
+ */
+export function parseTtscGraphNodeId(
+  id: string,
+): ITtscGraphNodeId | undefined {
+  const hash = graphNodeIdHash(id);
+  if (hash < 0) return undefined;
+  const tail = id.slice(hash + 1);
+  if (tail === "") return undefined;
+  const colon = tail.lastIndexOf(":");
+  if (colon === tail.length - 1) return undefined;
+  return {
+    path: unescapeGraphNodeIdPart(id.slice(0, hash)),
+    name: unescapeGraphNodeIdPart(colon < 0 ? tail : tail.slice(0, colon)),
+    ...(colon < 0 ? {} : { kind: tail.slice(colon + 1) }),
+  };
+}
+
+/** Encode a symbol identity without making its component boundaries ambiguous. */
+export function writeTtscGraphNodeId(
+  path: string,
+  name: string,
+  kind: string,
+): string {
+  return `${escapeGraphNodeIdPart(path)}#${escapeGraphNodeIdPart(name)}:${kind}`;
+}
+
+/** Return the raw path component when id is a symbol identity. */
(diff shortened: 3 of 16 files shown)


samchon:
Self-review finding before remediation: the TypeScript node-id decoder accepted `path#:kind`, whereas the Go producer-side parser rejects an empty symbol component. I will make the decoder fail closed for that malformed form, keep stale legacy ids readable, then create a new snapshot and restart the full solo review as required by the campaign procedure.

samchon:
Self-review finding before remediation: the three reducer comments still describe a raw `<path>#<name>:<kind>` prefix, but the implemented contract quotes `#` and `\` inside both components. I will correct those contract comments only, then snapshot and repeat the complete solo review.

samchon:
Implementation complete for #810 only; #822 and #835 remain deferred to their combined successor batch.

Final local verification:

- `go test ./internal/graph ./internal/graphsymbols ./test/driver -run 'Test(NodeIDsPreserveHashBearingComponents|RunLSPGraphSymbolsPreserveHashBearingPaths|RunLSPAnswersReferencesForASymbol|DumpIsAFunctionOfTheSource)$' -count=1`
- direct execution of the four new graph regressions plus the existing legacy-UNC viewer regression
- `git diff --check` and the protected Graph structure hash/tree guard

The final solo Self-Review read every changed file and hunk from the final `3ac23c9e599d29d3a67282ddf56fab49430b0f10` snapshot. Earlier review corrections are recorded above; the final fresh round found no further issue. All three implementation SHAs have complete exact-SHA cancellation records in the campaign ledger.

PR #900, opened 2026-07-20
fix(website): mirror the astnav shim in the compiler module

samchon:
## Intent

Fixes #899.

Restore the website compiler's self-contained Go module graph after #885 added the in-tree `shim/astnav` module to its imported wasm host.

## Scope

- mirror the new `shim/astnav` requirement and local replacement in `website/compiler/go.mod`;
- keep the module metadata exact through Go's own module tooling;
- add a generic regression that proves every locally replaced TypeScript-Go shim required by the imported `packages/ttsc` and `packages/wasm` modules remains mirrored by the website compiler main module;
- leave #890's runtime-hook branch unchanged.

## Verification

Implementation is active. The pre-fix `master` website run fails while building `website/compiler/cmd/playground` with a missing `shim/astnav` Go sum entry.

This is an implementation pull request in the active campaign. Repository Actions remain enabled, but runs caused by each exact branch SHA are cancelled and recorded under the campaign policy. The merge gates are focused module-mirror regression coverage, the real website compiler Go build, the website build, directly affected type/module checks, diff validation, and a clean solo Self-Review.

## Campaign CI

Claim SHA `d7578224d53013075f15df82c74a7523d249106b` is under exact-SHA cancellation supervision. Implementation SHA records will be added in chronology comments after each push.

Issue #838, opened 2026-07-20
fix(lint): measure every `format/*` rule against the pinned Prettier 3.8.3 it claims to mirror


samchon:
## Problem

Every rule in the `format/*` family declares in its own doc comment that it mirrors a specific Prettier option, and `website/src/content/docs/lint/format.mdx` publishes that claim to users. No rule is checked against Prettier. Each one enumerates the node kinds and edit directions its author implemented, and its tests then feed it inputs drawn from inside that same set, so a kind the rule never visits and a direction it never emits have no witness anywhere in the repository.

This is a class defect, not three bugs. Three divergences are proven below and are required cases for the fix, but the issue is the missing oracle: as long as expectations are transcribed by hand from a reading of Prettier's documentation, the next rule will drift the same way and nothing will notice.

Affected: everyone who runs `ttsc format`, `ttsc fix`, `ttsc check` with `format.severity` set, or the editor format command, since all of them route through the same rules. The user-visible symptom is that `ttsc format` and Prettier never reach a common fixed point on the same file, which contradicts the product claim at the top of the format guide.

## Evidence

The oracle below is the repository's own pinned Prettier. `package.json` declares `"prettier": "^3.8.3"` and `pnpm-lock.yaml` resolves `prettier@3.8.3`; the runs reported here were executed against that installed copy and printed `prettier 3.8.3`.

### Divergence 1: `format/trailing-comma` misclassifies `es5` and can never remove a comma

Prettier models `trailingComma` as a level comparison, not as a per-kind allowlist. `shouldPrintComma(options, level = "es5")` returns true when the option is `"es5"` and the level is `"es5"`, or when the option is `"all"` and the level is `"all"` or `"es5"`. Only call and `new` argument lists and function parameter lists pass the explicit `"all"` level; every other list, including TypeScript tuple types, type-parameter declaration lists, and enum members, uses the default `"es5"` level and therefore does receive a comma under `trailingComma: "es5"`.

Oracle run at `printWidth: 40`, `parser: "typescript"`, `trailingComma: "es5"`:

```text
--- tupleType
type T = [
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa,
  bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb,
];

--- typeParams
function f<
  Aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa,
[...]

PR #766, opened 2026-07-18
feat(lint): rule diagnostics link to their documentation (codeDescription)

samchon:
Populates the `codeDescription` LSP field for the built-in lint rule corpus, so a rule code in the editor's Problems panel links to that rule's documentation.

Origin: #734. The wire field landed in #745 but was never populated (`findingToLSPDiagnostic` did not set it), so editors rendered the rule id as plain text. This is the producer side.

## Derivation per family

The URL is derived from the rule name; there is no 743-entry table. `packages/lint/linthost/rule_docs.go` dispatches on the name's family:

| Family | URL |
| --- | --- |
| core (149, no prefix) | `https://eslint.org/docs/latest/rules/<name>` |
| `unicorn/` (146) | `https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/<bare>.md` |
| `typescript/` (98) | `https://typescript-eslint.io/rules/<bare>` |
| 18 remaining ported families (332) | `https://ttsc.dev/docs/lint/rules/<family>#<anchor>` |

The three upstream families keep pointing at their own reference because those pages are the behavioral specification `@ttsc/lint` ports against, and the rule sources already cite them.

The website anchor reproduces the heading id Nextra assigns. Nextra 4.2.17's `remark-headings` plugin slugs the flattened heading text with `github-slugger`, which **lowercases and removes the `/`** rather than replacing it with a hyphen. So `jsx-a11y/alt-text` anchors as `#jsx-a11yalt-text`, not `#jsx-a11y-alt-text`. The transform is `strings.ToLower(strings.ReplaceAll(name, "/", ""))`, verified to reproduce the rendered id for all 725 documented rules with zero mismatches. The one rule name carrying an uppercase letter (`security/detect-pseudoRandomBytes`) falls out of the same `ToLower`; it is not special cased.

## Oracle cross-check

The in-source doc-comment URLs are the oracle. Extracted all 311 distinct upstream URLs from `packages/lint/linthost/*.go` and `packages/lint/rule/*.go` and compared them against the derivation:

- **311 of 311 matched. Zero disagreements.** No oracle URL exists that the derivation fails to produce, and no derived URL contradicts one in the source.
- Coverage of the oracle per family: core 101/149, unicorn 141/146, typescript 69/98. The remainder are rules whose source file simply carries no URL in its header comment, not rules where the two disagree.

## Deliberately left unmapped

[...]

PR #904, opened 2026-07-20
fix(lint): reject colliding contributor namespaces

### LINKED ISSUE #821: fix(lint): reject colliding contributor namespaces instead of silently dropping one plugin
## Problem

`@ttsc/lint` accepts a contributor namespace matching `/^[a-z][a-z0-9_-]*$/` and maps it to a Go sub-package name by replacing hyphens with underscores. That mapping is not injective: the two valid, distinct namespaces `a-b` and `a_b` both become `a_b`.

`resolveConfigFileContributors` deduplicates on the mapped name and keeps the first entry, so the second plugin is removed from the descriptor with no error, no warning, and no diagnostic. Declaration order in the user's lint config decides which of two validly declared plugins survives.

Expected behavior is that a config declaring two plugins gets two plugins, or an explicit failure that names what collided. What happens instead is that one plugin disappears. If the config enables a rule from the dropped plugin, the Go engine records it as unknown and prints `@ttsc/lint: ignoring unknown rule "…"`, which points the user at a rule name rather than at the plugin that was silently discarded. If the config does not enable such a rule, the plugin vanishes with no observable signal at all.

Affected: any project loading two lint contributor plugins whose namespaces differ only by hyphen versus underscore. Hyphenated namespaces are the documented ESLint-style convention (`react-hooks`), so `react-hooks` alongside `react_hooks` is the realistic collision.

### Source, `packages/lint/src/index.ts`

The accepted namespace grammar and the mapping (anchors as of `612abc2d2`, for navigation):

```ts
const NAMESPACE_PATTERN = /^[a-z][a-z0-9_-]*$/;                 // :63

function goSubpackageName(namespace: string): string {          // :72
  return namespace.replace(/-/g, "_");
}
```

`resolveConfigFileContributors` states the intent and then does the opposite:

```ts
// Dedup on the Go-subpackage form (post hyphen→underscore transform)
// so two namespaces that collapse to the same Go identifier surface
// here instead of as the contributor validator's opaque
// `duplicate name "a_b"` error.
const occupied = new Set<string>();
const out: TtscPluginContributor[] = [];
for (const entry of entries) {
  const goName = goSubpackageName(entry.namespace);
  if (occupied.has(goName)) continue;
  occupied.add(goName);
  out.push({ name: goName, source: entry.source });
}
return out;
```

The comment says the collision should "surface here". The code discards it silently. Nothing downstream can recover, because the host validator in `packages/ttsc/src/plugin/internal/loadProjectPlugins.ts` (which does throw `ttsc: plugin "…" contributors[i] duplicate name "a_b"`) receives the already-deduplicated array and sees only one entry.

A lint config declaring two distinct contributors whose namespaces collide:

```js
// lint.config.cjs
module.exports = {
  plugins: {
    "a-b": { source: ".../contributor-one" },
    a_b: { source: ".../contributor-two" },
  },
};
```

Both source directories exist and contain a non-test Go source file. Invoking the built `@ttsc/lint` descriptor factory against that config returns, with no error and exit success:

```json
[{"name":"a_b","source":"...\\contributor-one"}]
```

`contributor-two` is gone. Reversing the declaration order swaps which one survives.

### Root cause and scope

`goSubpackageName` is a lossy normalization used as a uniqueness key. The dedup loop treats a key collision between two different user-facing namespaces as a repeat of the same entry.

Every supported config format converges on `readConfigPluginEntries` and then `resolveConfigFileContributors`, so JSON, CJS, ESM, and TypeScript lint configs share the defect. `.json` is read in-process, `.js`/`.cjs` through `require`, and `.ts`/`.cts`/`.mts`/`.mjs` through a `ttsx` subprocess, but all four hand their entries to the same loop.

### Provenance

`git blame` attributes both the comment and the silent `continue` to commit `50d869b20`, merged as PR #104 during the lint config-file migration. PR #104's description and test plan cover config-file ownership, contributor discovery, and loader behavior; namespace normalization collisions are not mentioned. The comment records the intended behavior and no corresponding throw was ever written.

Searches over open and closed issues and pull requests for contributor namespace collisions, duplicate contributor names, and hyphen/underscore plugin collisions return nothing. Open #729 (contributor diagnostic data channel) and #751 (cross-file diagnostic URI vocabulary) concern rule output, not descriptor composition.

## Consequence surface

| Config shape | Today | Required |
| --- | --- | --- |
| `"a-b"` and `a_b`, distinct sources | first wins, second dropped silently | explicit failure naming both namespaces, the shared Go name, and the config path |
| `"react-hooks"` and `react_hooks` | same | same |
| Three or more namespaces collapsing to one Go name | first wins, rest dropped | one failure listing every colliding namespace |
| The same namespace declared twice | already folded before this boundary | unchanged |
| Distinct namespaces with distinct Go names | composed correctly | unchanged, hyphenated namespaces stay supported |
| Rule-name collisions between successfully linked contributors | separate existing validation | out of scope |

Downstream, only one contributor source is statically linked into the lint host binary, so the dropped plugin's rules cannot register at all. The failure is invisible in the common case and misleading in the other: `@ttsc/lint: ignoring unknown rule "a_b/some-rule"` names the rule and not the plugin that was discarded.

Platform-neutral. All four config formats are affected identically.

## Approach

Invariant: the mapping from declared contributor namespaces to linked Go sub-package names is injective, and a config that violates it fails loudly before the host binary is built.

Architectural owner: `resolveConfigFileContributors` in `packages/lint/src/index.ts`, which is the boundary that produces the descriptor's `contributors` array.

The error must name the user's own vocabulary, not the internal one: both original namespaces as written, the shared normalized Go name, and the config file path that declared them. The current `continue` deliberately routes around the host's opaque `duplicate name "a_b"` message, and that reasoning is right; the fix is to produce the better message here rather than to keep hiding the condition.

Constraints:

- Valid hyphenated namespaces must keep working. Do not narrow `NAMESPACE_PATTERN` and do not stop mapping hyphens to underscores.
- Do not relax the host's Go-identifier validation in `loadProjectPlugins.ts`; it stays as the last line of defense.
- Do not silently pick a winner under any ordering.

| Case | Expected |
| --- | --- |
| Positive | a config declaring `"a-b"` and `a_b` with distinct existing sources fails with an error naming both namespaces, `a_b`, and the config path |
| Positive | the reversed declaration order fails identically, with the same message content |
| Positive | three namespaces collapsing to one Go name report all three |
| Negative twin | `"a-b"` and `"c-d"` compose normally into `a_b` and `c_d` |
| Negative twin | a single hyphenated namespace still links and its rules still fire |
| Boundary | zero plugins, one plugin, and a config with no `plugins` key are unaffected |
| Boundary | the same namespace declared twice keeps its current folding behavior |
| Format coverage | at least one in-process format (`.json` or `.cjs`) and one subprocess-evaluated format (`.ts`) prove the collision is caught after config evaluation |

The natural test home is `tests/test-lint/src/features/plugin/`, alongside `test_descriptor_discovers_contributors_via_plugin_config_dir_for_wrapper_tsconfig.ts`, which already drives the real factory through `TestLintPlugin.loadFactory()` against a materialized project. A native-plugin case under `tests/test-lint/src/native-plugins/config/` can additionally prove the failure happens before the host build rather than surfacing later as an unknown-rule warning.

Narrow proving command:

```bash
pnpm --filter @ttsc/test-lint start
```

Broader:

```bash
node scripts/test-go-lint.cjs
pnpm test:features
```

Because the descriptor factory ships in `packages/lint/lib`, run `pnpm --filter @ttsc/lint build` (or `pnpm build:current`) before the feature test.

## Coordination

- No open or closed issue owns this, and no open pull request touches `packages/lint/src/index.ts`.
- Independent of the `format/*` conformance work (#838, #839, #856), which lives entirely in `packages/lint/linthost`. Safe to run in parallel with those.
- The user-facing error text is new product surface. If the lint guide documents contributor namespaces, update it in the same change.
- Nothing in `packages/graph` is involved.

samchon:
## Intent

Fixes #821.

Reject distinct lint contributor namespaces that map to the same Go subpackage name instead of silently retaining the first contributor.

## Behavior

- `react-hooks` and `react_hooks` now fail before a native host build with the lint-config path, both original namespaces, and their shared normalized Go name.
- Repeated occurrences of one exact namespace in flat-config arrays retain the existing first-entry fold.
- Non-colliding hyphenated namespaces still map to their Go-safe underscore names.

## Scope

- validate the complete resolved contributor entry set in `@ttsc/lint` before its existing Go-name deduplication;
- cover CJS config loading, flat-config repeats, three-way collisions, valid mappings, and an actual TypeScript `ttsx` subprocess;
- document the namespace collision contract beside the existing contributor-name transform rule.

The change does not alter #838/#839/#856 lint formatting work, #766 LSP diagnostics behavior, #900's website module repair, or Graph surfaces. #766 edits a separate `protocol.mdx` paragraph; a three-way merge against its live head has no conflict markers.

## Verification

- Pre-fix CJS reproduction: a collision failed the expected-exception assertion because the factory silently discarded the later contributor.
- `pnpm --filter @ttsc/lint build` passed.
- `TTSC_TEST_DIR=features/plugin pnpm --filter @ttsc/test-lint start` passed, including the new CJS and built-`ttsx` cases.
- `TTSC_TEST_DIR=features pnpm --filter @ttsc/test-lint start` passed in 221.862 seconds with worktree-scoped `GOCACHE` and `GOTMPDIR`.
- `pnpm exec prettier --check` passed for all four changed files.
- `git diff --check origin/master...HEAD` passed after rebasing onto `master` at `e3e5fe0299654ed9fba9c8604ac1751341cd5171`.
- Fresh solo Self-Review covered the descriptor, CJS and TypeScript loader paths, config cache, exact-repeat compatibility, Windows launcher handling, documentation, and the downstream contributor builder. No finding survived.

## Campaign CI

The implementation head is `c55d215940d6b7dfe50cdd039e81ac9d4f006b0f`. A first supervisor accidentally queried mistyped non-existent SHA `c55d21594e0c7b00c5935ae3beadc0ab08349c12` and observed no runs; it performed no cancellation. The immediately started corrective exact-SHA supervisor observed eight runs, cancelled only those runs, and recorded every one as terminal `cancelled`. The ready-state operation receives a separate exact-SHA supervision record.


samchon:
Final local validation for `c55d215940d6b7dfe50cdd039e81ac9d4f006b0f`:

- `pnpm --filter @ttsc/lint build` passed.
- `TTSC_TEST_DIR=features/plugin pnpm --filter @ttsc/test-lint start` passed, including the CJS, exact-repeat, and built-`ttsx` collision regressions.
- `TTSC_TEST_DIR=features pnpm --filter @ttsc/test-lint start` passed in 221.862s with worktree-scoped `GOCACHE` and `GOTMPDIR`.
- Targeted Prettier checks and `git diff --check origin/master...HEAD` passed.
- Fresh solo whole-diff Self-Review completed with no remaining finding.

Campaign CI chronology is complete. The initial post-push supervisor queried a mistyped non-existent SHA and cancelled nothing; the immediately started recovery supervisor for the actual SHA observed only runs `29746549495`, `29746549549`, `29746549594`, `29746549597`, `29746549621`, `29746549695`, `29746549703`, and `29746549705`, and confirmed every one terminal `cancelled`. The ready transition and this verification-body update each received a separate exact-SHA supervisor; both found only those same terminal runs and no new or active run. The complete local records are in the campaign knowledge base.

samchon:
Final review found a formatting defect before merge: targeted `prettier --check` reports drift in all four files of the current implementation diff. I will apply Prettier only to those four files, then rerun the affected build and full lint feature suite, complete a fresh solo review, and create a new exact-SHA campaign cancellation record before merging.

samchon:
Correction to the preceding note: the targeted Prettier write normalized only local CRLF checkout bytes. `git diff --cached --exit-code` and `git diff --exit-code` are both clean after staging the exact four paths, so no source change or new SHA exists. The same head `4d3d52c8e53a22e09853f333d4e70be33e491c4a` now passes targeted Prettier check. I will still finish the full feature verification and final readback on that unchanged SHA before merge.

PR #905, opened 2026-07-20
fix(ttsc): quote Go workspace comment-prefix paths

### LINKED ISSUE #857: fix(ttsc): quote go.mod and go.work tokens by upstream's rule, so a UNC plugin path does not vanish from the generated workspace
## Problem

The generated source-plugin `go.work` emits every `use` and `replace` path through a helper whose JSDoc promises it quotes "exactly as `golang.org/x/mod/modfile`'s `AutoQuote` does". Two clauses of the upstream predicate were transcribed incorrectly, and both errors fail in the dangerous direction: a token that upstream would quote is emitted bare.

The reachable consequence on Windows is silent. `formatGoWorkPath` normalizes `\` to `/`, so every UNC path (`\\server\share\…`) and every extended path (`\\?\C:\…`) becomes a token that starts with `//`. The modfile lexer treats `//` as a comment start, so the entire `use` line is consumed as a comment. `go` reports no error, the overlay module simply disappears from the workspace, and the subsequent build fails against the public proxy with an unrelated "no required module provides package …" that points nowhere near the real cause.

The second error is loud instead of silent: a path containing a non-ASCII space (U+00A0 and the other Unicode space separators) is left bare and hard-fails the workspace parse.

Expected behavior: any absolute filesystem path ttsc can produce round-trips through the modfile lexer as exactly one token.

Affected: every source-plugin build (`ttsc`, `ttsx`, `ttsc prepare`, `ttsc fix`, `ttsc format`, and the `@ttsc/lint` sidecar cold build) whose project, `node_modules`, or resolved cache root lives on a UNC share or is reached through an extended path. That is an ordinary corporate Windows layout. POSIX is reachable through any absolute path containing a doubled separator or a `/*` sequence, and both platforms through any directory name containing a non-ASCII space.

## Evidence

`packages/ttsc/src/plugin/internal/buildSourcePlugin.ts` is byte-identical between the audited commit `2b724664eee71c57cd688c55122cac137774bf62` and `master` at `612abc2d20d3f99429319081c1f6520bbe34ac7a`.

### The two divergences

Upstream, `golang.org/x/mod@v0.34.0`, `modfile/rule.go`:

```go
func MustQuote(s string) bool {
	for _, r := range s {
		switch r {
		case ' ', '"', '\'', '`':
			return true

		case '(', ')', '[', ']', '{', '}', ',':
			if len(s) > 1 {
				return true
			}

		default:
			if !unicode.IsPrint(r) {
				return true
			}
		}
	}
	return s == "" || strings.Contains(s, "//") || strings.Contains(s, "/*")
}
```

ttsc's `mustQuoteGoModToken`:

```ts
    if (!isGoGraphic(ch)) {
      return true;
    }
  }
  return s === "" || s === "//" || s === "/*";
```

| # | Upstream | ttsc | Effect |
| --- | --- | --- | --- |
| 1 | `strings.Contains(s, "//")` and `strings.Contains(s, "/*")` | whole-token equality `s === "//"`, `s === "/*"` | a token that *contains* `//` or `/*` is left bare |
| 2 | `!unicode.IsPrint(r)` | `!isGoGraphic(ch)`, where the graphic class admits `\p{Zs}` | non-ASCII Unicode spaces (U+00A0, U+2000-200A, U+202F, U+205F, U+3000) are left bare |

The file already contains the correct predicate for the second case. `isGoPrintable`, used by `goQuoteString`, is documented as "`strconv.IsPrint`: the graphic categories except that the ONLY spacing character is the ASCII space (U+0020)". Only `mustQuoteGoModToken` reaches for the graphic predicate instead.

Both divergences are load-bearing, because the modfile lexer (`x/mod@v0.34.0/modfile/read.go`) treats `//` as a comment start both at token start and mid-token, errors on `/*`, and rejects any rune that `unicode.IsSpace` accepts or `unicode.IsPrint` rejects when scanning an identifier. A bare non-breaking space therefore terminates the token.

Upstream oracle, `go run` against `golang.org/x/mod v0.34.0` from the local module cache with `GOPROXY=off`:

```text
"//server/share/proj"        -> MustQuote=true AutoQuote="//server/share/proj"
"//?/C:/Users/x/proj"        -> MustQuote=true AutoQuote="//?/C:/Users/x/proj"
"C:/Users/x//y/proj"         -> MustQuote=true AutoQuote="C:/Users/x//y/proj"
"a b"                   -> MustQuote=true AutoQuote="a b"
"a/*b"                       -> MustQuote=true AutoQuote="a/*b"
"/home/user/plugin"          -> MustQuote=false AutoQuote=/home/user/plugin
```

The ttsc implementation, imported directly from the audited source:

```text
"//server/share/proj"        -> ttsc autoQuote=//server/share/proj
"//?/C:/Users/x/proj"        -> ttsc autoQuote=//?/C:/Users/x/proj
"C:/Users/x//y/proj"         -> ttsc autoQuote=C:/Users/x//y/proj
"a b"                   -> ttsc autoQuote=a b
"a/*b"                       -> ttsc autoQuote=a/*b
"/home/user/plugin"          -> ttsc autoQuote=/home/user/plugin

-- formatGoWorkPath on real Windows UNC input --
"\\\\server\\share\\proj\\node_modules\\ttsc"    -> //server/share/proj/node_modules/ttsc
"\\\\?\\C:\\Users\\x\\proj"                      -> //?/C:/Users/x/proj
```

Five of six diverge, every one of them in the direction of emitting a token that should have been quoted.

End-to-end against the real `go` toolchain, with a `go.work` shaped exactly as `writeGoWork` emits (`go 1.26\n\nuse (\n\t.\n\t<path>\n)`):

```text
=== CASE A: unquoted //server/share/proj ===
{ "Go": "1.26", "Use": [ { "DiskPath": "." } ], "Replace": null }      <- the entry is gone

=== CASE B: unquoted NBSP path ===
go: errors parsing go.work:
go.work:5:15: unexpected input character ' '
```

Case A is the severe one: `go` exits 0 and reports nothing. The `use` line was consumed as a comment.

### The existing test asserts the wrong oracle

`tests/test-ttsc/src/native-plugins/source-plugin/test_gomod_token_quoting_mirrors_go_autoquote.ts` declares its cases as "`[input, expected]` where expected is exactly what `modfile.AutoQuote` emits", and its own doc claims it pins every branch of the logic. Three rows contradict the real oracle:

```ts
[`a${NBSP}b`, `a${NBSP}b`], // NBSP is Zs (graphic) → not forced.
…
["//", '"//"'],  // a line comment opener must be quoted to be a token.
["/*", '"/*"'],  // a block comment opener too.
```

The NBSP row asserts the opposite of upstream, and the `//` and `/*` rows exercise only whole-token equality, which is the exact case ttsc's narrowed predicate still handles. Those expectations came from ttsc's own rule rather than from the Go implementation, so they must be re-derived, not merely extended.

`a95fff991` ("fix(ttsc): quote go.work paths so workspace spaces don't break source builds (#397)") introduced the whole helper, closing #394 ("go.work generation breaks when workspace paths contain spaces"). That work attempted the general fix, a real port rather than a special case; what remains is a two-clause transcription error inside it. The file is unchanged since `v0.19.3` apart from an unrelated stderr message line.

No open or closed issue mentions UNC, `//`, `AutoQuote`, `MustQuote`, or Unicode spaces. #394 covers the ASCII-space case only. No open pull request touches `buildSourcePlugin.ts`.

## Consequence surface

| Input class | Emitted today | Result |
| --- | --- | --- |
| Windows UNC path (`\\server\share\…`) | bare `//server/share/…` | `use`/`replace` entry silently dropped, `go` exits 0 |
| Windows extended path (`\\?\C:\…`) | bare `//?/C:/…` | same |
| Any path containing a doubled separator (`C:/x//y`, `/a//b`) | bare | same |
| Any path containing `/*` | bare | `go` errors on the block-comment opener |
| Any path containing a non-ASCII space | bare | `go: errors parsing go.work: unexpected input character` |
| Path containing an ASCII space | quoted | correct, and must stay correct (#394) |
| Ordinary POSIX or drive-letter path | bare | correct, and must stay bare so space-free paths do not churn |

Both emitters are affected: the `use` block and the `replace … => <path>` line built by `sourceBuildWorkspaceReplacements`. The `//` case presents as a confusing downstream module-resolution error rather than as a diagnosable path problem, which is why it has never been reported.

## Approach

Invariant: **`autoQuoteGoModToken` is a faithful port of `modfile.AutoQuote` across the whole token domain, so any absolute filesystem path ttsc can produce round-trips through the modfile lexer as exactly one token.**

Architectural owner: `mustQuoteGoModToken` and its two helper predicates in `packages/ttsc/src/plugin/internal/buildSourcePlugin.ts`, plus the oracle rows in the existing test.

Do not change `goQuoteString` or `escapeGoRune`: they already mirror `strconv.Quote` correctly, and `isGoPrintable` is already the right shape. Only the must-quote predicate is wrong.

Do not fix this by pre-rewriting paths, for example by mapping a UNC path to a drive letter. The emitter is the owner of the quoting decision, and a path rewrite would move the failure rather than remove it.

Test expectations must be derived from the Go implementation, not from ttsc's output. The oracle is reproducible: `golang.org/x/mod` is already in the module graph, and a small Go program that prints `MustQuote`/`AutoQuote` for each row is enough to generate the table.

| Case | Input | Expected emission |
| --- | --- | --- |
| Positive | `//server/share/proj` | `"//server/share/proj"` |
| Positive | `//?/C:/Users/x/proj` | `"//?/C:/Users/x/proj"` |
| Positive | `C:/Users/x//y/proj` | `"C:/Users/x//y/proj"` |
| Positive | `a/*b` | `"a/*b"` |
| Positive | a token containing U+00A0, and one containing U+3000 | quoted, with the rune escaped by the `strconv.Quote` form |
| Negative twin | `/home/user/plugin` | unchanged, bare |
| Negative twin | `C:/Users/jsmith/proj` | unchanged, bare |
| Boundary | `(`, `)`, `,` alone | unchanged, bare |
| Boundary | `""` (empty) | `""` |
| Boundary | `//` and `/*` as whole tokens | quoted, as today |
| Regression | `/Users/John Smith/x` | `"/Users/John Smith/x"`, keeping #394 fixed |
| End to end | `writeGoWork` with a UNC `use` directory | `go work edit -json` lists the directory |
| Test correction | the NBSP row and the two comment-opener rows | re-derived from `modfile.AutoQuote`, not from ttsc output |

Narrow proving command:

```bash
TTSC_TEST_DIR=native-plugins/source-plugin pnpm --filter @ttsc/test-ttsc start
```

That is the `ttsc native source-plugin` CI lane, and it holds the test whose oracle rows this issue must correct: `test_gomod_token_quoting_mirrors_go_autoquote.ts`. The runner exits 1 with "No tests were discovered under …" if the subtree is empty or misspelled.

The unit rows exercise the exported `autoQuoteGoModToken` and `formatGoWorkPath` directly and are platform-independent. The end-to-end row is different in kind: it must emit a real `go.work` through `writeGoWork` and read it back with `go work edit -json`, which needs the real Go toolchain that this lane already provides (`TTSC_GO_BINARY` in CI). Asserting only the quoted string would not have caught the original defect, because the string looked plausible; what exposes it is that `go` accepts the file and silently omits the entry.

Broader:

```bash
pnpm test:features
```

## Coordination

- #394 and its PR #397 own the ASCII-space case in this same helper. That fix is correct and must stay green; this issue completes the port rather than replacing it.
- Independent of the CLI flag boundary candidates and of every Graph and lint candidate. The only shared file is `buildSourcePlugin.ts`, which no other candidate touches.
- The `//` case is reachable only where a path can contain a doubled separator after normalization, so a Windows runner (or a fixture that supplies the token directly) is required to prove the end-to-end row. The unit rows are platform-independent.

samchon:
Closes #857

## Scope

Reserve the Go module token-quoting repair in `buildSourcePlugin`:

- mirror `modfile.MustQuote` for embedded comment openers and non-printable Unicode spaces;
- preserve the established `strconv.Quote` escaping path;
- prove generated `go.work` remains parseable for normalized UNC and extended Windows paths.

## Verification

Pending implementation. The final snapshot will run the focused source-plugin
suite and the broader feature suite locally. Campaign-triggered Actions runs
for this exact reservation SHA are being cancelled and recorded; repository
Actions configuration remains unchanged.

### DIFF
--- a/packages/ttsc/src/plugin/internal/buildSourcePlugin.ts
+++ b/packages/ttsc/src/plugin/internal/buildSourcePlugin.ts
@@ -1473,9 +1473,9 @@ export function formatGoWorkPath(p: string): string {
  * Quote `token` for a `go.mod`/`go.work` line exactly as
  * `golang.org/x/mod/modfile`'s `AutoQuote` does: return it unchanged when it is
  * already a clean bare token, otherwise return its Go double-quoted form so the
- * value round-trips through the modfile lexer. A space-free path is therefore
- * emitted byte-for-byte as before; only tokens that would otherwise
- * mis-tokenize are quoted.
+ * value round-trips through the modfile lexer. A clean bare token is therefore
+ * emitted byte-for-byte as before; only tokens that would otherwise be split
+ * or interpreted as comments are quoted.
  *
  * Exported for unit tests.
  */
--- a/packages/ttsc/src/plugin/internal/buildSourcePlugin.ts
+++ b/packages/ttsc/src/plugin/internal/buildSourcePlugin.ts
@@ -1506,11 +1506,11 @@ function mustQuoteGoModToken(s: string): boolean {
       }
       continue;
     }
-    if (!isGoGraphic(ch)) {
+    if (!isGoPrintable(ch)) {
       return true;
     }
   }
-  return s === "" || s === "//" || s === "/*";
+  return s === "" || s.includes("//") || s.includes("/*");
 }
 
 // Mirror `strconv.Quote`: wrap in double quotes, backslash-escape `"` and `\`,
@@ -1561,13 +1561,6 @@ function escapeGoRune(ch: string): string {
       return `\\U${cp.toString(16).padStart(8, "0")}`;
     }
   }
-}
-
-// `unicode.IsGraphic`: categories L, M, N, P, S, Zs (all spacing).
-const GO_GRAPHIC_RE = /^[\p{L}\p{M}\p{N}\p{P}\p{S}\p{Zs}]$/u;
-
-function isGoGraphic(ch: string): boolean {
-  return GO_GRAPHIC_RE.test(ch);
 }
 
 // `strconv.IsPrint`: the graphic categories except that the ONLY spacing
--- a/tests/test-ttsc/src/native-plugins/source-plugin/test_gomod_token_quoting_mirrors_go_autoquote.ts
+++ b/tests/test-ttsc/src/native-plugins/source-plugin/test_gomod_token_quoting_mirrors_go_autoquote.ts
@@ -21,7 +21,8 @@ import {
  * 3. Assert each output equals the value Go would emit for the same token.
  */
 export const test_gomod_token_quoting_mirrors_go_autoquote = () => {
-  const NBSP = " "; // U+00A0: a Zs (graphic) space that is not ASCII space.
+  const NBSP = String.fromCodePoint(0x00a0);
+  const IDEOGRAPHIC_SPACE = String.fromCodePoint(0x3000);
 
   // [input, expected] where expected is exactly what modfile.AutoQuote emits.
   const autoQuoteCases: readonly [string, string][] = [
@@ -34,7 +35,6 @@ export const test_gomod_token_quoting_mirrors_go_autoquote = () => {
     [".", "."],
     ["café", "café"], // lone Unicode letter is graphic → not forced.
     ["a😀b", "a😀b"], // lone astral symbol is graphic → not forced.
-    [`a${NBSP}b`, `a${NBSP}b`], // NBSP is Zs (graphic) → not forced.
     ["(", "("], // a lone bracket/comma is a legal bare token.
     [")", ")"],
     [",", ","],
@@ -49,6 +49,8 @@ export const test_gomod_token_quoting_mirrors_go_autoquote = () => {
     ["", '""'], // empty string is not a valid bare token.
     ["//", '"//"'], // a line comment opener must be quoted to be a token.
     ["/*", '"/*"'], // a block comment opener too.
+    ["a//b", '"a//b"'], // an embedded line-comment opener too.
+    ["a/*b", '"a/*b"'], // an embedded block-comment opener too.
 
     // strconv.Quote escape forms (control runes force quoting on their own).
     ["a\tb", '"a\\tb"'], // \t
@@ -60,6 +62,8 @@ export const test_gomod_token_quoting_mirrors_go_autoquote = () => {
     ["a\x07b", '"a\\ab"'], // \a (bell)
     ["a\x01b", '"a\\x01b"'], // \xNN for other C0 controls
     ["a\x7fb", '"a\\x7fb"'], // \x7f for DEL
+    [`a${NBSP}b`, '"a\\u00a0b"'], // non-ASCII Zs is not printable.
+    [`a${IDEOGRAPHIC_SPACE}b`, '"a\\u3000b"'], // neither is U+3000.
     [`a ${NBSP}`, '"a \\u00a0"'], // \uNNNN: NBSP is not printable inside a quote.
     [String.fromCodePoint(0x10ffff), '"\\U0010ffff"'], // \UNNNNNNNN for astral non-printable.
 
@@ -81,6 +85,9 @@ export const test_gomod_token_quoting_mirrors_go_autoquote = () => {
   const formatCases: readonly [string, string][] = [
     ["C:\\Users\\John Smith\\proj", '"C:/Users/John Smith/proj"'],
     ["C:\\Users\\jsmith\\proj", "C:/Users/jsmith/proj"],
+    [String.raw`\\server\share\proj`, '"//server/share/proj"'],
+    [String.raw`\\?\C:\Users\x\proj`, '"//?/C:/Users/x/proj"'],
+    [String.raw`C:\Users\x\\y\proj`, '"C:/Users/x//y/proj"'],
     ["/Users/John Smith/x", '"/Users/John Smith/x"'],
     ["/home/user/x", "/home/user/x"],
     [".", "."],
--- a/tests/test-ttsc/src/native-plugins/source-plugin/test_writegowork_quotes_workspace_paths_with_spaces.ts
+++ b/tests/test-ttsc/src/native-plugins/source-plugin/test_writegowork_quotes_workspace_paths_with_spaces.ts
@@ -1,4 +1,5 @@
 import { TestProject } from "@ttsc/testing";
+import child_process from "node:child_process";
 
 import {
   assert,
@@ -9,7 +10,7 @@ import {
 } from "../../internal/source-build";
 
 /**
- * Verifies writeGoWork quotes go.work paths that contain spaces (#394).
+ * Verifies writeGoWork quotes unsafe go.work workspace paths (#394, #857).
  *
  * `writeGoWork` emitted `use`/`replace` paths unquoted, so an overlay resolved
  * under a directory with a space — the common `C:\Users\John Smith\...` /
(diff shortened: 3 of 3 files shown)


samchon:
Self-review finding: the utoQuoteGoModToken documentation says every space-free path remains bare, which is no longer true once embedded // and /* comment openers correctly force quoting. I will replace that with the precise clean-bare-token invariant; this is documentation-only and does not alter the implementation or acceptance matrix.

samchon:
## Campaign stopped by user direction

Current head is `206d68ca62352166e3f3e2fe87f2df6a8a75502d` (`docs(ttsc): clarify Go token quoting invariant`). The branch worktree is clean. This comment is a handoff only: no merge, rebase, push, worktree removal, branch deletion, or Go-cache cleanup was performed after the stop instruction.

### Passed local verification

- `TTSC_BUILD_SCOPE=test-ttsc pnpm run build:current` completed successfully with this worktree's scoped `GOCACHE` and `GOTMPDIR`.
- `TTSC_TEST_DIR=native-plugins/source-plugin pnpm --filter @ttsc/test-ttsc start` completed successfully. This covers the corrected `MustQuote` table and the real `go work edit -json` workspace-path parse.
- `pnpm run build:current` completed successfully with the same scoped Go cache and temporary directory.
- The attempted broad command had already completed `test-banner`, `test-factory`, and `test-graph` successfully. In `test-lint`, all output before the corpus partitions was successful, including corpus partitions `1_of_4` (1,314,043 ms) and `2_of_4` (695,967 ms). No test failure was emitted.

### Interrupted verification

The command

```text
pnpm test:features
```

was running through `pnpm --filter "./tests/test-*" -r --workspace-concurrency=1 start` in the campaign worktree with scoped `GOCACHE` and `GOTMPDIR`. At the stop request it was still executing `tests/test-lint` real `ttsc.js --cwd <temporary fixture> --noEmit` children; the last observed fixture was `unicorn-prefer-dom-node-remove`. The supervisor process was PID `50444`; its complete captured child process tree was terminated, and a post-stop process check found no remaining PIDs. Because it was deliberately interrupted, no `features-final.exit` file exists and this broad suite must not be reported as passed.

### Remaining exact steps

1. Restart the interrupted full feature command with this worktree's exact scoped `GOCACHE` and `GOTMPDIR`, and require exit code 0.
2. Reconfirm clean status, `git diff --check`, and a fresh final self-review; synchronize/rebase only if `master` advanced.
3. For every subsequent remote mutation, complete the campaign exact-SHA CI-cancellation supervisor and record it in the campaign CI state: final PR update/ready transition, any rebase push, and merge SHA.
4. Merge only after those local-verification and campaign-review gates are complete.
5. Only after a confirmed merge, remove `D:\github\samchon\ttsc@campaign-857-v2`, verify its scoped `go-cache` and `go-tmp` are gone, prune worktree metadata, and delete the local topic branch.

### Preservation boundary

The separate #773 worktree `D:\github\samchon\ttsc@pr-773` and its external-change boundary were not touched, stashed, reset, cleaned, rebased, or removed during this work. At this stop point its `git status --short` reports clean; no claim is made that a later agent may alter that worktree without separate authorization.

samchon:
## Superseded by #914

The maintainer directed that the leftover open pull requests be consolidated into one, so this branch's commits were merged into `fix/campaign-2026-07-21` with their history and authorship intact and are validated there as part of a single CI run.

Nothing from this branch is dropped. The branch itself is preserved for now; it will be deleted only after #914 merges.

Closing here to stop this pull request from consuming a parallel CI run.

PR #909, opened 2026-07-20
fix(wasm): keep the fountain test fixture off the host root

### LINKED ISSUE #907: fix(wasm): fountain position test writes its fixture under the host root
## Problem

`TestFountainPositionVerbsResolveTouchingTokens` creates its fixture directly under the host filesystem root:

```go
root := fmt.Sprintf("/fountain-position-%d", time.Now().UnixNano())
```

That passes through the real Go `js/wasm` Node runner on Windows because `/...` resolves on the current drive and that environment permits the write. The ordinary Ubuntu runner cannot create directories under `/`, so the Go lane fails before exercising the fountain API:

```text
--- FAIL: TestFountainPositionVerbsResolveTouchingTokens (0.01s)
    fountain_position_tokens_test.go:42: mkdir /fountain-position-1784552091757000000: Permission denied
```

Evidence: PR #766, workflow run `29742514194`, job `88352433438`. The failure was introduced with the runnable host regression in #885; it is independent of #766 and remains on current `master`.

## Consequence surface

This test runs a `GOOS=js GOARCH=wasm` binary through Node. Its Go `filepath` model is POSIX even when Node is hosted by Windows. A naive `t.TempDir()` is therefore not a verified fix: on Windows it yields a drive-letter path that the wasm-side path resolver does not recognize as absolute, and the project loader prepends its working directory to the resulting tsconfig path.

The fixture needs a unique, writable absolute path in the wasm process's POSIX path space whose Node filesystem target is valid on both Windows and Ubuntu. Cleanup must remain exact and test-owned. The fixture contents, snapshot lifecycle, and production fountain behavior are unrelated.

## Approach

Make the js/wasm test runner own a writable host directory and expose the corresponding POSIX absolute fixture root to the Go test, or use an equivalently explicit platform mapping. Prove the path mapping through the real Node wrapper on Windows and the Ubuntu CI runner; do not infer portability from a native Go test or from one host OS.

## Acceptance

- The focused fountain position regression passes through `node scripts/test-go-wasm.cjs` on Windows with worktree-scoped Go cache and temporary directories.
- The ordinary Ubuntu Go lane creates the fixture and reaches the API assertions instead of failing at `mkdir /...`.
- The test rejects or clearly diagnoses a missing/invalid runner-provided root rather than silently falling back to the host root.
- Production code and API result envelopes do not change.

## Change Integrity

The host-root path was never part of the contract. The fixture contents, snapshot `cwd`, API calls, position assertions, and released-snapshot boundary remain identical.

samchon:
Closes #907.

## Scope

Reserves the js/wasm test-runner surface that currently creates the fountain position fixture directly below `/`. The implementation will provide a unique writable absolute fixture root in the wasm process's POSIX path space while preserving a valid Node host mapping on both Windows and Ubuntu.

This batch changes test fixture ownership only. Production fountain behavior, snapshot lifecycle, API envelopes, and PR #766's lint implementation are excluded.

## Verification

Pending implementation:

- real Windows `node scripts/test-go-wasm.cjs` with worktree-scoped `GOCACHE` and `GOTMPDIR`;
- focused runner-path negative and boundary coverage;
- solo Self-Review over the complete diff.

Campaign implementation checks for this claim SHA are cancelled by exact SHA. Repository Actions and workflow settings remain enabled and unchanged.


samchon:
Closing this implementation-free claim because the active #890 owner was explicitly assigned the shared baseline CI repair. No tracked file was changed on this branch. Issue #907 remains the self-contained handoff; #890 owns the js/wasm runner fix and its Windows/Ubuntu verification.

Issue #734, opened 2026-07-17
feat(lint): LSPDiagnostic is missing four LSP fields — codeDescription, tags, relatedInformation, data


samchon:
Part of #733.

## Problem

`LSPDiagnostic` (`packages/ttsc/internal/lspserver/lsp_plugin.go:40-46`) carries five fields:

```go
type LSPDiagnostic struct {
  Range    LSPRange              `json:"range"`
  Severity LSPDiagnosticSeverity `json:"severity,omitempty"`
  Code     any                   `json:"code,omitempty"`
  Source   string                `json:"source,omitempty"`
  Message  string                `json:"message"`
}
```

LSP's spec `Diagnostic` has four more, and every one is absent: `relatedInformation`, `tags`, `codeDescription`, `data`. Each is a user-facing feature the host cannot express, and none of them needs new architecture — they are struct fields on a channel that already works end to end.

## Evidence, per field

**`codeDescription` — the cheapest of the four, and free for rules.** A URL attached to the diagnostic code; editors make the rule name in the Problems panel a link to its docs. `Code` is already carried, so nothing about the rule or the wire shape changes: only the host knows the docs URL for a built-in rule name, and only the host has to supply it. Roughly ten lines.

**`tags` — the most-loved lint UX in VS Code, and ttsc cannot do it.** `DiagnosticTag.Unnecessary` greys out dead code; `Deprecated` strikes it through. Ruff and Biome both set them, **and both use `Deprecated` as well as `Unnecessary`** — ESLint is the laggard here, not the standard. ttsc ships the rules that want this (`no-unused-vars` and its family) and renders them as ordinary errors.

One caution that belongs in the implementation, not just the review: **a tag is a claim about what the code *is*, not how bad it is.** `Unnecessary` says "safe to delete." Applying it to a finding that means "work not yet done" tells the user to delete the thing they have not finished. It fits `no-unused-vars`; it does not fit a coverage rule. The mapping should be per-rule and deliberate, never a severity-driven default.

[...]

Issue #910, opened 2026-07-20
epic(campaign): quota-stop handoff for #872, self-contained as of 2026-07-20


samchon:
# Campaign stop-point handoff for #872

This issue is the self-contained handoff for the 2026-07-19/20 repository-audit campaign. It supersedes the operational status in #872. The maintainer stopped the campaign because the remaining model quota is exhausted. Discovery and implementation must not continue automatically from this issue.

Do not depend on `.wiki`. It is git-ignored and may not exist on the next computer or in the next session. Every fact needed to decide the next action is recorded below. Treat #872 as historical evidence only: its counts, in-flight table, and old parallel workflow are stale.

## Exact resume baseline

| Item | Value |
| --- | --- |
| Remote target | `origin/master` |
| Remote target SHA | `43539ab39325c41a3504acc6244131aea7d3ce1e` |
| Head subject | `fix(ttsc): bound resident transform request lifetimes (#906)` |
| Campaign parent | #872 |
| New workflow contract | #908 merged as `141bee18c90de174371090b8cbed2f109eb08763` |
| Current policy | Issue campaigns and reviews are solo by default. Load `.agents/skills/multi-agent/SKILL.md` only on an explicit parallel request. Do not create a clone or worktree for solo work or Self-Review. |

The protected Graph contract still holds at the remote target:

| Surface | Required value |
| --- | --- |
| `packages/graph/src/structures/ITtscGraphApplication.ts` SHA-256 | `6AB7451FCD17E8A8C1F0F7BAEAA6A96AA4E6DEDCB379F70F030ECEE64197BE6C` |
| `packages/graph/src/structures` Git tree | `3c7399e3ce9f2404e64d652c7f1259ebd0db2613` |

Verify both before any Graph-related change and immediately before any Graph merge.

## Campaign work completed after the original handoff

| Pull request | Merge commit | Closed campaign issue or batch |
| --- | --- | --- |
| #903 | `336b3255496250560340e88e7d675b9e6317c38b` | #810 |
| #902 | `31f88636b34eecaff03b6fe3d486ca274575d2f8` | #853 implementation, but the issue itself is still open and needs explicit state verification before closure |
| #892 | `5f6e2d1ae6036e2d4acc1637ae479ea1941c1968` | #815 |
| #886 | `c86a91ee747500711449232525f61f9094d8a285` | #830, #831, #832 |
| #895 | `03104d146bf01cca5c33a8a01f1ec15b79a4d13a` | #838 |
[...]

PR #914, opened 2026-07-20
fix(campaign): consolidate the leftover pull requests and repair the red master lanes

### LINKED ISSUE #857: fix(ttsc): quote go.mod and go.work tokens by upstream's rule, so a UNC plugin path does not vanish from the generated workspace

### LINKED ISSUE #911: fix(ttsc): the transform envelope's `candidates` section escaped the programmatic API contract test
## Problem

`master` is red on the `ttsc fast` lane for every pull request opened against it, and the
programmatic transform API's `graph.candidates` field is described by a type that contradicts
what the runtime always produces.

`ITtscCompilerTransformation.IReferenceGraph.candidates` is declared optional
(`candidates?: Record<string, string[]>`), but `parseReferenceGraph` in
`packages/ttsc/src/compiler/internal/transformProjectInMemory.ts` materializes it
unconditionally: `const candidates = parseDependencyLists(section.candidates) ?? {}` and then
returns `{ candidates, configs, edges, globals }`. The Go producer omits the key when empty
(`json:"candidates,omitempty"` on `driver.TransformGraph`), so the empty case is created purely by
the JS parser. Every `TtscCompiler.transform()` caller therefore receives a `candidates` key, even
when the field's own documentation says an absent candidate set means nothing to watch.

The pre-existing contract test asserts the whole graph section with `deepEqual` and still expects
the three-key shape that predates the field, so it fails on every branch.

## Evidence

Reproduction: open any pull request that touches `packages/**` or `tests/**` and read the
`ttsc fast` lane. Two independent, unrelated branches fail identically:

| Pull request | Job | Failure |
| --- | --- | --- |
| #766 (`packages/lint`, `packages/ttsc/internal/lspserver`, website) | [88380178836](https://github.com/samchon/ttsc/actions/runs/29750714570/job/88380178836) | `+ candidates: {}` |
| #773 (`packages/lint` rules, website) | [88384446536](https://github.com/samchon/ttsc/actions/runs/29751945907/job/88384446536) | `+ candidates: {}` |

```text
AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
+ actual - expected
  {
+   candidates: {},
    configs: [
      'tsconfig.json'
    ],
    edges: {
      'src/main.ts': [
  at test_ttsccompiler_transform_surfaces_reference_graph_and_volatile_list
     (tests/test-ttsc/src/features/api/test_ttsccompiler_transform_surfaces_reference_graph_and_volatile_list.ts:38:12)
```

Neither branch touches `packages/ttsc/driver`, `packages/ttsc/src/compiler`, or
`tests/test-ttsc/src/features/api`, so neither can be the cause.

Root cause and provenance: `c86a91ee747500711449232525f61f9094d8a285`
(`fix(bundlers): preserve cache freshness across resolution changes`, #886) added
`Candidates` to `driver.TransformGraph`, added the optional `candidates` property to
`ITtscCompilerTransformation.IReferenceGraph`, and taught `parseReferenceGraph` to always emit it.
It did not extend
`tests/test-ttsc/src/features/api/test_ttsccompiler_transform_surfaces_reference_graph_and_volatile_list.ts`,
whose `assert.deepEqual(result.graph, { configs, edges, globals })` is an exact-shape assertion.

`.github/workflows/test.yml` triggers only on `pull_request`, so the merge itself never proved the
lane, and no run on `master` can surface the regression.

## Consequence surface

- Every open and future pull request touching the watched paths inherits a red `ttsc fast` lane,
  so the lane carries no signal and a real regression in it would be indistinguishable from this
  one. #766 and #773 are blocked on it today.
- The published `ITtscCompilerTransformation` type is wrong for every consumer that narrows on
  `graph.candidates === undefined`: the runtime never produces `undefined` through the
  programmatic API, while the field's own doc comment describes the absent case as meaningful.
- The `candidates` field has no positive assertion anywhere in the TypeScript API suite. The Go
  side is covered (`packages/ttsc/test/driver`, `cmd/ttscgraph`), but the JS envelope decode of a
  *populated* candidate map is unproven, so a producer- or parser-side regression in the field
  would only be caught in Go.
- `parseReferenceGraph` treats an all-empty section as "no graph" and returns `undefined`. Adding
  `candidates` to that emptiness test changed when a graph section exists at all; that boundary is
  also unasserted.

## Approach

`transformProjectInMemory.parseReferenceGraph` owns the JS-side shape of the envelope, and
`ITtscCompilerTransformation.IReferenceGraph` owns its published type. Make the declared type and
the produced value agree, in whichever direction preserves the documented meaning of an absent
candidate set, and prove the field through the programmatic API rather than only through the Go
driver.

Do not delete or weaken the existing `deepEqual` shape assertion to make the lane green; an
exact-shape assertion is what makes an unannounced envelope key visible at all.

| Case | Expected |
| --- | --- |
| Positive | A project whose import has a superseding resolution candidate exposes it through `TtscCompiler.transform().graph.candidates`, keyed by importing file, with project-relative slash paths. |
| Negative twin | A project with no superseding candidate produces the documented empty form, and the declared type matches it exactly. |
| Boundary | A project whose graph section is otherwise empty keeps `parseReferenceGraph`'s existing "no graph" answer unchanged. |
| Regression | The existing exact-shape `deepEqual` on the whole `graph` object stays exact and passes. |

Proving commands, narrow to broad:

```bash
pnpm --filter @ttsc/test-ttsc start   # TTSC_TEST_DIR=features/api
pnpm test:features                    # ttsc fast lane
```

## Coordination

- Blocks #766 and #773, whose `ttsc fast` lanes cannot go green until this lands.
- Same provenance as the Windows-only serve-session defect filed alongside this issue; both come
  from #886 landing with a cancelled CI run. They touch different files and are independently
  implementable.
- #806 separately owns the structural question of proving that every committed test has a CI
  owner. This issue is the concrete defect, not that audit.

### LINKED ISSUE #912: fix(graph): a resident serve session never sees a superseding resolution candidate appear on Windows
## Problem

On Windows, a resident `ttscgraph serve` session never notices that a superseding module
resolution candidate appeared on disk. Creating a file that outranks the currently selected
module target leaves the session in `unchanged` mode with no dump, so a Windows host keeps serving
a stale program whose import edge is now wrong.

The same scenarios pass on Linux, so this is a platform defect in the candidate disk-state watch,
not a design gap.

## Evidence

The `windows-go` lane is red on every open pull request. Six `packages/ttsc/cmd/ttscgraph`
cases fail, all with the same shape — a reload was expected and `unchanged` was reported:

```text
--- FAIL: TestServeSessionKeepsExportsBlockedFileCandidatesUnchanged (0.35s)
    serve_session_keeps_exports_blocked_file_candidates_unchanged_test.go:62:
        export-map first target = dump:false mode:"unchanged" changed:false
--- FAIL: TestServeSessionKeepsInactiveConditionalExportUnchanged (0.28s)
    serve_session_keeps_inactive_conditional_export_unchanged_test.go:64:
        active types condition = dump:false mode:"unchanged" changed:false
--- FAIL: TestServeSessionReloadsCommonJSPackageEntryForESM (0.26s)
    serve_session_reloads_commonjs_package_entry_for_esm_test.go:48:
        CommonJS package main.ts = dump:false mode:"unchanged" changed:false
--- FAIL: TestServeSessionReloadsSupersedingModuleCandidates/nearer_node_modules (0.25s)
        superseding candidate = dump:false mode:"unchanged" changed:false want-node:"nearerNodeModulesWinner"
--- FAIL: TestServeSessionReloadsSupersedingModuleCandidates/package_exports (0.25s)
--- FAIL: TestServeSessionReloadsSupersedingModuleCandidates/conditional_package_exports (0.26s)
FAIL	github.com/samchon/ttsc/packages/ttsc/cmd/ttscgraph	17.355s
```

Source: `windows-go` job
[88380178897](https://github.com/samchon/ttsc/actions/runs/29750714570/job/88380178897) on #766,
a branch that touches only `packages/lint`, `packages/ttsc/internal/lspserver`, and the website.

Platform contrast, same package, same commit range: the Linux `go` lane reports
`ok github.com/samchon/ttsc/packages/ttsc/cmd/ttscgraph 20.110s`
([88384446188](https://github.com/samchon/ttsc/actions/runs/29751945907/job/88384446188)).

Note the two `Keeps…Unchanged` cases fail on their *second* assertion, not the first: the
"must not reload" half passes, and the following "must reload once the real target appears" half
is what reports `unchanged`. Every failure is the same direction — a candidate that appeared was
not seen.

Provenance: `c86a91ee747500711449232525f61f9094d8a285` (#886) added
`packages/ttsc/driver/resolution_candidates.go`, the `auxiliaryInputs` /
`captureDiskStates` / `diskStatesChanged` watch in `packages/ttsc/cmd/ttscgraph/serve.go`, and all
six failing cases. `.github/workflows/test.yml` triggers only on `pull_request`, and #886's own
run was cancelled under the previous campaign's cancellation choreography, so the lane was never
proven before the merge.

## Consequence surface

The defect is in the *watch* path, so it is not limited to the failing test names:

- `ttscgraph serve` on Windows: any resolution change that materializes a new higher-priority
  file — a nearer `node_modules`, a package `exports` target, a conditional export, a CommonJS
  package `main` gaining a `.ts` sibling, a `paths` pattern, a `moduleSuffixes` variant — is
  silently missed. The session answers `unchanged` and consumers cache a stale graph.
- The same `driver.ModuleResolutionCandidates` / `driver.ModuleResolutionPredecessors` /
  `driver.FileCandidates` / `driver.TypeReferenceCandidates` producers feed the bundler freshness
  path (`packages/unplugin`, `packages/metro`) through the transform envelope's `graph.candidates`.
  If the cause is candidate-path production rather than the disk-state capture, Windows bundler
  caches are unsound in the same way; if it is the capture, they are not. The fix must establish
  which, and cover both consumers accordingly.
- `captureDiskStates` deliberately records an unreadable path as absent so an unparseable
  specifier cannot fail a snapshot. That tolerance also swallows Windows-specific `os.ReadFile` and
  `os.Stat` errors, so a path that is malformed on Windows looks exactly like a path that does not
  exist yet. Any repair must keep the tolerance for genuinely unparseable specifiers while making a
  real Windows path observable.
- Failure and recovery: because the session reports `unchanged`, there is no error, no diagnostic,
  and no recovery — the staleness is invisible to the caller.

## Approach

`packages/ttsc/cmd/ttscgraph/serve.go` owns the resident session's freshness decision;
`packages/ttsc/driver/resolution_candidates.go` owns the candidate paths it watches. The invariant
is that a candidate path produced on Windows must name the same filesystem object the OS creates,
under whatever separator, volume, short-name, and case form the compiler's source file names and
`t.TempDir()` roots supply.

Establish where the Windows path form diverges before choosing a repair — candidate production,
the disk-state capture, or the comparison — and fix that layer rather than normalizing at the
comparison site so the symptom disappears.

| Case | Expected |
| --- | --- |
| Positive | Each of the six failing cases reports `reload` with a dump on Windows, for the same reason it does on Linux. |
| Negative twin | A candidate strictly *below* the selected target still reports `unchanged` on Windows; the repair must not make every irrelevant file creation reload. |
| Boundary | A specifier the host OS cannot parse as a path (`./style.css?inline`, a `data:` URL) still records as absent instead of failing the snapshot on Windows. |
| Regression | Linux and macOS behavior for all `cmd/ttscgraph` serve-session cases is unchanged. |

Proving commands, narrow to broad:

```bash
go test ./packages/ttsc/cmd/ttscgraph -run TestServeSession   # on Windows and on Linux
pnpm test:go                                                   # go and windows-go lanes
```

The `windows-go` lane is the only mechanical proof of the platform half; a green Linux run is not
acceptance for this issue.

## Coordination

- Blocks #766 and #773, whose `windows-go` lanes cannot go green until this lands.
- Shares provenance with the `graph.candidates` contract-test defect filed alongside it; both come
  from #886 landing with a cancelled CI run. They touch different files and are independently
  implementable.
- If the cause turns out to be in `resolution_candidates.go`, the bundler freshness consumers in
  `packages/unplugin` and `packages/metro` are in the same consequence class and must be covered
  in the same change.

samchon:
## Intent

One cycle pull request for the 2026-07-21 campaign. It consolidates the three leftover open pull requests — #905, #766, #773 — into a single branch and adds the repair for the two lanes that are red on `master` itself, so the whole set is proven by one CI run instead of four.

The consolidation is deliberate and was requested by the maintainer. The three branches address different packages, but they were competing for the same finite local and CI capacity while none of them could reach a green lane set, because `master` itself is red.

Closes #857.
Closes #911.
Closes #912.

Advances #734 (the `codeDescription` producer half; the `data` field remains) and #768.

Supersedes #905, #766 and #773, which are closed in favour of this branch. Their commits and authorship are preserved here through merge commits.

## Scope

| Origin | Subject | Files |
| --- | --- | --- |
| #905 → #857 | Quote Go workspace comment-prefix paths: mirror `modfile.MustQuote` for embedded comment openers and non-printable Unicode spaces, keep the `strconv.Quote` escaping path, and keep a generated `go.work` parseable for normalized UNC and extended Windows paths | `packages/ttsc/src/plugin/internal/buildSourcePlugin.ts`, `tests/test-ttsc/src/native-plugins/source-plugin/**` |
| #766 → #734 | Populate the `codeDescription` LSP field so a rule code in the editor's Problems panel links to that rule's documentation, derived from the rule name's family rather than a 743-entry table | `packages/lint/linthost/{lsp,rule_docs}.go`, `packages/lint/test/{command,registry}/**`, `packages/ttsc/internal/lspserver/lsp_plugin.go`, `website/.../protocol.mdx` |
| #773 → #768 | Route what the storybook, security and solid rule families already know through the diagnostic-tag, suggestion and automatic-rename channels | `packages/lint/linthost/rules_{security,solid,storybook}.go`, `packages/lint/src/structures/rules/**`, `packages/lint/test/{fix,rules}/**`, `website/src/content/docs/lint/rules/**` |
| #911 | The transform envelope's `candidates` section escaped the programmatic API contract test | `packages/ttsc/src/{compiler,structures}/**`, `tests/test-ttsc/src/features/api/**` |
| #912 | A resident serve session never sees a superseding resolution candidate appear on Windows | `packages/ttsc/driver/resolution_candidates.go`, `packages/ttsc/cmd/ttscgraph/**` |

#911 and #912 are not optional additions. They were introduced by #886, which merged onto `master` while its CI run was cancelled, and `.github/workflows/test.yml` triggers only on `pull_request`, so nothing on `master` reports them. Every open pull request inherits both, which is why none of the three consolidated branches could reach a green lane set on its own.

## Verification

Local builds and test runs are not performed. The maintainer's standing directive is that this machine has no capacity for them and that GitHub Actions is the only build and test environment. This pull request's ordinary CI is therefore the sole acceptance gate, and every lane is consumed to completion — no campaign run is cancelled.

The `typia` and `nestia` workflows are downstream-compatibility signal and do not gate this merge.

## Deferred

Nothing from the three original branches is dropped. The remaining published campaign issues stay out of this pull request and enter the next cycle.


samchon:
## Self-Review round 3: the #912 fix is wider than the Windows lane that exposed it

Round 3 asked what the identity fallback changes for real projects rather than for the six failing tests, and the answer is larger than the issue stated.

`ModuleResolutionPredecessors` located the compiler's selected target by comparing path spellings. TypeScript resolves a `node_modules` package through the real path by default (`preserveSymlinks` is the opt-out), so the winner arrives realpath-resolved. pnpm's layout makes that a rename on every bare specifier — in this repository:

```text
node_modules/prettier -> node_modules/.pnpm/prettier@3.8.3/node_modules/prettier
```

The candidate search enumerates `node_modules/prettier/...`; the compiler reports `node_modules/.pnpm/prettier@3.8.3/node_modules/prettier/...`. No candidate could ever equal it, so the scan fell through and returned `nil` — the branch that deliberately discards the whole predecessor list.

The consequence is that **`graph.candidates` was empty for every bare specifier in a pnpm project, on every platform**, not only on Windows. #886's freshness feature was inert exactly where the repository's own package manager puts its dependencies. The Linux `cmd/ttscgraph` suite did not catch it because `t.TempDir()` fixtures contain no symlink, and the Windows lane failed only because 8.3 short-name expansion produced the same divergence inside a temp directory.

This is reasoned from the layout and TypeScript's documented default, not measured, so I am stating it as the mechanism rather than as a number.

### What that means for this pull request

- The repair is correct and the six Windows cases prove it, but reviewers should read it as a behavior change for `@ttsc/unplugin` and `packages/metro` under pnpm: those caches will now invalidate on superseding-candidate changes they previously ignored. Correctness improves; hit rate may fall where it was wrong before.
- The cost is paid only when the textual scan finds nothing, which is precisely the case that used to return `nil`. A project whose spellings agree still pays nothing.
- `captureDiskStates` already stats every candidate each snapshot, so the second pass stays within the order of work the resident session already does.

No code changed in round 3.

samchon:
## Acceptance gates, head `b3f8257df45213a4035885c4662ed11fbb5a02f9`

### CI

56 of 57 checks succeeded on this exact SHA. No check failed and none was cancelled — every run was consumed to completion.

The one outstanding check is `nestia`, which the maintainer's standing directive excludes from merge gating; it is downstream-compatibility signal only.

Notable lanes, given what this branch carries:

| Lane | Result | What it proves |
| --- | --- | --- |
| `windows-go` | success | #912. `cmd/ttscgraph` now passes on Windows; the six serve-session cases that motivated the issue report reload instead of unchanged. |
| `ttsc fast` | success | #911. The contract test that had been red on `master` since #886 passes, and the new populated-`candidates` case with it. |
| `ttsc native source-plugin` | success | #857, validated by CI for the first time — every lane on the original #905 was cancelled. |
| `go`, `lint native corpus-1..4`, `lint native fix/format/config` | success | The consolidated `@ttsc/lint` work from #766 and #773, including the `solid/no-react-deps` regression this branch repaired. |

### Self-Review

Three complete solo rounds over `origin/master...HEAD`, 39 files, +2341/-72.

- Round 1 applied two changes: the `storybook/hierarchy-separator` fallback that stops the rule going silent when it cannot bound the raw separator, and the missing "Type-aware via the Checker" marker on the two rules this branch made Checker-backed. It also rejected one proposal after tracing it — narrowing the redundant-story-name range to spare a comment would leave `{ /* note */, }`, which does not parse.
- Round 2 read every file round 1 had not and re-verified each round-1 edit against the test that governs it. It applied nothing.
- Round 3 asked what the diff does to real projects and found the pnpm consequence recorded in the previous comment. It applied nothing.

Rounds 2 and 3 changed no code, so the review is clean on this exact head.

### Protected Graph contract

Verified immediately before requesting merge, as #910 requires:

| Surface | Value | Expected |
| --- | --- | --- |
| `packages/graph/src/structures/ITtscGraphApplication.ts` SHA-256 | `6AB7451FCD17E8A8C1F0F7BAEAA6A96AA4E6DEDCB379F70F030ECEE64197BE6C` | matches |
| `packages/graph/src/structures` tree | `3c7399e3ce9f2404e64d652c7f1259ebd0db2613` | matches |

### Not verified

No local build or test was run, by the lead or by any subagent, per the maintainer's standing directive. CI is the whole of the verification above.

PR #919, opened 2026-07-21
docs(agents): state the solo campaign cycle cadence

samchon:
## Intent

The maintainer's working cadence for a solo issue campaign was not written down, so `.agents/skills/issue-campaign/development.md` left five operative rules to be inferred: every admitted issue is conquered in one cycle regardless of difficulty, implementation runs continuously with its tests, CI is consumed once instead of per commit, and the claim body, commit messages, and pull-request comments each own a different part of the issue bookkeeping.

## Scope

One file, `.agents/skills/issue-campaign/development.md`. Five added paragraphs and one reworded list item.

**`Plan One Cycle Pull Request`** gains the difficulty closure. The section already refused to split the cycle by package, invariant, or validation lane, and step 2 already listed the four dispositions that remove an issue. What it did not say is that an issue whose resolution needs a judgment call about design, invariant ownership, or an acceptable behavior change is settled and implemented inside the cycle. The new paragraph names those four dispositions as the only exits, so difficulty is excluded by construction rather than by a separate prohibition.

**`Claim The Complete Cycle`** step 4 changes from "Link every cycle issue" to "Reference every cycle issue by number", and a new paragraph keeps closing keywords out of the claim body. The body is written before any code exists, so a claim-time closing list closes whatever the cycle later drops, defers, or disproves, and buries the analysis those issues carry. The cycle's closing set is the union of the commit lines, which makes the merge close exactly what landed. This is stated as the rule rather than as a warning about the failure.

**`Implement And Write Tests`** gains three paragraphs:

- implement without interruption, write each piece's tests as that piece lands, keep committing as each unit becomes coherent, and do not pause the sequence for a check run;
- end each commit message with one `Close #n: <issue title>` line per resolved issue, so a multi-issue commit carries one legible line per issue;
- post a pull-request comment after each commit naming what landed and which issues it resolved.

The three roles are stated separately because they are mechanically different. GitHub auto-closes from a commit message or the pull-request body and never from a comment, so a reader who folded them together would put the closing keyword in the comment and close nothing. The keyword and the number are what GitHub matches, and the `: <issue title>` tail is free text, so the format closes normally while keeping the log readable without opening each number.

**`Validate With CI And Self-Review`** gains the read-once rule. The repair loop below it already operated over a settled head, but nothing said CI is not a per-commit gate. The paragraph states the cadence and the fact that enforces it: every workflow under `.github/workflows/` sets `cancel-in-progress: ${{ github.event_name == 'pull_request' }}`, so an intermediate commit's run is cancelled by the next push and waiting on it stalls implementation for a discarded result.

## Already covered, left alone

- **Commits split for diagnosis, not for pacing.** `Plan One Cycle Pull Request` already says "Keep issue-level commits when that improves diagnosis, but the pull request remains the integrated campaign unit." The cadence paragraph adds only the missing half, that the sequence is not paused for a check run, and does not restate the diagnosis rationale.
- **One pull request per cycle, and no package or lane splitting.** Stated in the flow rules and in `Plan One Cycle Pull Request`. Untouched.
- **The DAG does not defer an issue to a later pull request.** Already the closing paragraph of `Plan One Cycle Pull Request`. The difficulty paragraph follows it instead of merging into it, because predecessor order and difficulty are separate reasons an implementer reaches for a second pull request.
- **The claim pull request opens empty and prevents overlapping work.** Already steps 2 through 4 plus the rationale paragraph. Only the linking verb changed, so the following paragraph's reference-versus-close distinction is unambiguous.
- **Repair every red lane in the same pull request, including a pre-existing failure.** Already in the flow rules and in `Validate With CI And Self-Review`. Untouched.

## Deliberately not added

- **No edit to `.agents/skills/pull-request/SKILL.md`.** Its "Watch Checks After Every Ordinary Push" loop is not contradicted. This document's flow pushes the claim commit and then the integrated snapshot, and the read-once rule governs implementation pauses, not the settled head's check loop, which still runs through the existing repair cycle.
- **No edit to `.agents/skills/multi-agent/issue-campaign.md`.** Its batch pull requests link issues under a different ownership model, and the maintainer stated this cadence for the solo campaign.
- **No new section and no sibling document.** Each point landed in the section that already owns its phase, per the documentation skill's rule that always-applicable procedure stays in the owning document.

## Verification

- Prettier over the changed file with the options `prettier.config.js` applies to `*.md` (`--prose-wrap never --embedded-language-formatting off --print-width 80 --tab-width 2`), reported unchanged on the final content. The flags were passed explicitly because this worktree has no `node_modules`; the two plugins in that config apply only to TypeScript and JSDoc, not to Markdown.
- `git diff --check` clean.
- The one added link, `[CI is read once per settled head](#validate-with-ci-and-self-review)`, resolves to the existing `## Validate With CI And Self-Review` heading, the same anchor the document's Flow list already uses. No other link was touched.
- `cancel-in-progress` verified by reading every `concurrency` block under `.github/workflows/`; all eleven set it to `${{ github.event_name == 'pull_request' }}`.
- `Closes`, `closing keyword`, and issue-linking wording grepped across `.agents/` and `AGENTS.md` to confirm no other document states a conflicting rule.
- No build, test, or Go command was run. This is a Markdown-only change.

### DIFF
--- a/.agents/skills/issue-campaign/development.md
+++ b/.agents/skills/issue-campaign/development.md
@@ -34,7 +34,7 @@ Different packages, invariants, or validation lanes do not split the solo cycle.
 
 An issue whose only predecessor is another issue in the same cycle is implementation-ready for this purpose. Order the edits through the DAG instead of deferring it to another pull request.
 
-Difficulty never removes an issue from the cycle. When a resolution needs a judgment call about design, invariant ownership, or an acceptable behavior change, settle it from the issue's evidence and implement that decision here. Proved duplicate, invalid premise, out of scope, and external blocker stay the only dispositions that take an issue out of the cycle.
+Difficulty never removes an issue from the cycle. When a resolution needs a judgment call about design, invariant ownership, or an acceptable behavior change, settle it from the issue's evidence and implement that decision inside the cycle. A proved duplicate, an invalid premise, an out-of-scope finding, and an external blocker remain the only dispositions that remove one.
 
 ## Claim The Complete Cycle
 
@@ -46,7 +46,7 @@ Claim the whole cycle before implementation:
 4. Reference every cycle issue by number, mark verification pending, and state that the pull request owns the complete accepted cycle.
 5. Record the checkout, branch, pull request, head SHA, issue set, and external temporary-asset ledger in `.wiki`.
 
-Reference the issues in the claim body, and keep every closing keyword out of it. The body is written before any code exists, so a claim-time closing list closes whatever the cycle later drops, defers, or disproves and buries the analysis those issues carry. The cycle's closing set is the union of the commit trailers, which makes the merge close exactly what landed.
+Reference the issues in the claim body, and keep every closing keyword out of it. The body is written before any code exists, so a claim-time closing list closes whatever the cycle later drops, defers, or disproves, burying the analysis those issues carry. The cycle's closing set is the union of the [commit closing lines](#implement-and-write-tests), which makes the merge close exactly what landed.
 
 The empty pull request prevents overlapping contributor work before code is written. Measure official duration from its GitHub `createdAt` timestamp through `mergedAt`, including implementation, CI, review, fixes, rebases, and merge.
 
--- a/.agents/skills/issue-campaign/development.md
+++ b/.agents/skills/issue-campaign/development.md
@@ -43,8 +43,10 @@ Claim the whole cycle before implementation:
 1. Use the current checkout, update the target branch with `git pull --ff-only`, and create one topic branch. Do not create a clone or worktree.
 2. Create one implementation-free commit with `git commit --allow-empty`.
 3. Push the branch and open one draft pull request.
-4. Link every cycle issue, mark verification pending, and state that the pull request owns the complete accepted cycle.
+4. Reference every cycle issue by number, mark verification pending, and state that the pull request owns the complete accepted cycle.
 5. Record the checkout, branch, pull request, head SHA, issue set, and external temporary-asset ledger in `.wiki`.
+
+Reference the issues in the claim body, and keep every closing keyword out of it. The body is written before any code exists, so a claim-time closing list closes whatever the cycle later drops, defers, or disproves and buries the analysis those issues carry. The cycle's closing set is the union of the commit trailers, which makes the merge close exactly what landed.
 
 The empty pull request prevents overlapping contributor work before code is written. Measure official duration from its GitHub `createdAt` timestamp through `mergedAt`, including implementation, CI, review, fixes, rebases, and merge.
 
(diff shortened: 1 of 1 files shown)


samchon:
## Self-Review

Four complete solo rounds over the base-to-head diff, each a fresh full read of the whole document rather than a recheck of the last fix.

**Round 1** applied three corrections. The claim-body paragraph called the per-commit closings "trailers" while the section that defines them calls them lines, so the two paragraphs named the same mechanism differently; it now says "commit closing lines" and links to the defining section. "Drops, defers, or disproves and buries" could parse as one four-item list, so the burying clause became a participle. The surviving dispositions became parallel noun phrases, and an ambiguous "here" became "inside the cycle".

**Round 2** applied three more. The claim-body paragraph opened by repeating step 4's affirmative referencing rule, so it now leads with the prohibition that step 4 does not already exclude. "GitHub reads the title tail as free text, which keeps the log legible" attributed the legibility to GitHub's parser instead of to writing the title, so it now says GitHub ignores the tail and the log stays legible. The comment paragraph carried two negations in one sentence and now carries the exclusion with "only".

**Round 3** applied one. "Saving a test pass for the end of the cycle" reads as a passing test as readily as a pass over the code to write tests; it now says "leaving the tests for the end of the cycle".

**Round 4** found nothing. It re-verified rule-before-reason ordering in all six added paragraphs, that each negative rule prevents a failure its affirmative does not already exclude, both internal anchors against the real headings, absence of em-dashes and emoji, one source line per paragraph, and prettier reporting the file unchanged.

### Consequence surface checked in round 4

- `.agents/skills/issue-campaign/SKILL.md` describes what this document owns as "the single cycle pull request, empty claim, ... ordinary CI"; the added rules fall inside that enumeration, so the parent skill and the AGENTS.md index need no edit.
- No `.github/PULL_REQUEST_TEMPLATE`, no `CONTRIBUTING.md`, and no other file under `.agents/` states an issue-closing or CI-cadence rule, so nothing in the repository now contradicts these paragraphs.
- The last 300 commits carry 57 `Closes #n` references and no `Close #n: <title>` line, which is the convention this change replaces for solo campaigns rather than a contradiction of a written rule.
- Under squash merge the constituent commit messages reach the squash commit, so the closing lines still close on merge. A squash body edited to drop them is already caught by `Merge And Clean Up` step 1, which requires verifying every linked issue's final state, so no squash-specific caveat was added.

### Corrections to the body above

The body was written at open and two of its counts are wrong. The change adds **six** paragraphs, not five: the difficulty closure, the claim-body rule, the continuous cadence, the commit closing line, the comment ledger, and the read-once rule, plus the one reworded list item. And **ten** workflows are `pull_request`-triggered and all ten set `cancel-in-progress`; the eleventh file under `.github/workflows/`, `release.yml`, has no `pull_request` trigger and no concurrency block, so the document's "every pull-request workflow" claim holds.

Per the pull-request skill these corrections stay in a comment so the thread preserves chronology.

PR #920, opened 2026-07-21
docs(agents): add the commit early-warning pass to the solo campaign

samchon:
## Intent

Follow-up to #919, which landed the solo campaign cycle cadence. That cadence has the author committing continuously through a long cycle, and nothing read those commits until the pre-merge round. This adds the missing early reader and, more importantly, writes down why it cannot become the gate.

A defect named while its commit is the newest thing written costs little, because the author has not yet built on it. A defect found after ten more commits is a rewrite. That timing is the entire value of layer 1, and it is orthogonal to what layer 2 proves.

## The two layers

**Layer 1, delegated and parallel.** After a commit lands the author may spawn one read-only subagent over that commit and keep implementing while it runs. The pass reads that one commit and reports candidates; it never edits, commits, pushes, or takes an implementation decision.

**Layer 2, the author's own complete Self-Review over the whole base-to-head diff before merge, unchanged.** Not delegated, not partitioned, and not satisfied by any number of per-commit passes.

Both are needed because a reader holding one commit cannot see what appears only across files. The document states the class with three concrete shapes: a helper that duplicates one the package already has, a document claiming a verification that the component it describes does not perform, and a replacement whose needle cannot match because the file on disk uses CRLF. None of the three is visible inside the commit that introduces it, and all three are the kind of finding a whole-surface round exists to catch.

## Scope

Four files, each stating its own part once.

| File | What it gains |
| --- | --- |
| `issue-campaign/development.md` | The definition, in `Implement And Write Tests`: what the pass is, its read-only limits, its timing rationale, why it never reduces the gate, and the cross-file blindness that makes both layers necessary. |
| `review/SKILL.md` | A `Commit Early-Warning Pass` section owning the boundary and the name. |
| `multi-agent/SKILL.md` | One contrast paragraph after the Shared Parallelism Rules. |
| `issue-campaign/SKILL.md` | The carve-out clause on its blanket no-subagent sentence. |

The design is not repeated in three places. `development.md` carries the procedure and the rationale, `review/SKILL.md` carries only the boundary, and `multi-agent/SKILL.md` carries only the topology contrast, each linking to the others.

### The naming

The review skill says the pass is not a review under that skill, that it delegates nothing the Non-Negotiable Review Law governs, and that one commit is not a declared surface, a reported candidate is not an accepted finding, and the passes do not add up to a round. It then forbids the name directly: never call it a Self-Review and never report it as one, because a reader who sees that name concludes the gate already ran and the whole-surface round disappears without anyone deciding to drop it.

### The carve-outs, stated as carve-outs

Three documents carried a blanket prohibition the pass would have contradicted. Every one was traced and reconciled:

- `issue-campaign/SKILL.md`, "The main agent owns every phase and never spawns or delegates to a subagent", now reads "spawns no subagent other than the read-only commit early-warning pass that development.md defines".
- `issue-campaign/development.md` flow rule, "Do not spawn or delegate to a subagent", now reads "Spawn no subagent except the read-only commit early-warning pass", linked to the defining section.
- `review/SKILL.md`, "Do not spawn a subagent, delegate a concern, or load the discussion skill", is **untouched by design**. It is about Self-Review and must keep meaning exactly that, so the new section states the pass is a different activity rather than editing the law to make room for it.

`multi-agent/SKILL.md`'s "Do not let agents re-delegate" is also untouched, so a multi-agent implementation agent still may not spawn a reader. The contrast paragraph is descriptive and grants that agent nothing. `Self-Review remains solo for every author and every implementation branch` stays true and stays where it is.

## Deliberately not added

- **No change to `review/SKILL.md`'s frontmatter.** "This skill never spawns review agents" stays literally true: the pass is not a review, and the campaign author spawns it, not this skill. Editing the trigger contract would suggest the skill is a route into the pass, which it is not.
- **No change to `AGENTS.md`.** Its maintenance rule limits edits to repository-contract changes, and no skill was added, renamed, or rescoped. The index lines for review and issue campaigns both remain accurate.
- **No change to `multi-agent/issue-campaign.md` or `multi-agent/review.md`.** The pass is a solo-campaign device, and extending it to parallel implementation batches would collide with the no-re-delegation rule.
- **No procedural detail about how to brief or run the subagent.** The pass is one read-only reader over one commit; a briefing template would inflate the document without adding a rule.

## Verification

- Prettier over all four files with the options `prettier.config.js` applies to `*.md` (`--prose-wrap never --embedded-language-formatting off --print-width 80 --tab-width 2`), then `--check`, which reports all four already conform. The flags are passed explicitly because this worktree has no `node_modules`; the two plugins in that config apply only to TypeScript and JSDoc.
- `git diff --check` clean.
- Every added link resolves: `#implement-and-write-tests` and `#validate-with-ci-and-self-review` are existing headings in `development.md` and are already used by its Flow list; `#commit-early-warning-pass` matches the new `## Commit Early-Warning Pass` heading; the relative paths `../review/SKILL.md` from `issue-campaign/development.md` and from `multi-agent/SKILL.md`, and `../issue-campaign/development.md` from `review/SKILL.md`, all resolve on disk.
- `subagent`, `sub-agent`, and `delegate` grepped across `.agents/` and `AGENTS.md` to find every blanket prohibition, so the reconciliation above covers all of them rather than the two I started from.
- The development flow list still says "Four rules" and still has four bullets.
- No build, test, or Go command was run. This is a Markdown-only change.


samchon:
## Self-Review

Five complete solo rounds over the base-to-head diff, each a fresh full pass over all four files and the documents that link to them.

**Round 1** applied three corrections. The campaign document restated "never call the pass a Self-Review", which the review skill already owns in full; it now points at that boundary and names what lives there instead of repeating the imperative. The review skill's law sentence lost a comma pile and now says the whole surface rather than the base-to-head surface, since the law also governs discovery rounds. "Takes an implementation decision" became "makes".

**Round 2** applied one. Both new paragraphs in `development.md` said "the author", while that document calls the actor "the main agent" in its flow rules and everywhere else. The campaign documents now say "main agent" and the review and multi-agent skills keep "author", which is their own vocabulary.

**Round 3** applied one, and it is the finding worth reading. `multi-agent/issue-campaign.md` step 3 sends every parallel campaign agent to read the solo development procedure, so the carve-out I added to that document's flow rules reached batch implementers and contradicted `Do not let agents re-delegate`. Nothing in the four files I edited showed this; it only appears when you follow who reads what. The contrast paragraph now closes it in the file that owns the re-delegation rule.

**Round 4** applied one. The exclusion said "batch implementer"; `issue-campaign.md` says "implementation agents" and reserves "batch" for the parallel wave, so it now reads "batch implementation agent".

**Round 5** found nothing. It re-read all four hunks, re-verified both anchors against their headings and all three relative paths on disk, confirmed the actor vocabulary is now consistent per document, confirmed the design is stated once each rather than three times, and confirmed `development.md` still says "Four rules" over four bullets.

### Consequence surface

- Every blanket no-subagent rule in the repository was located by grepping `subagent`, `sub-agent`, and `delegate` across `.agents/` and `AGENTS.md`, then reconciled: two carved out, one deliberately untouched, one left binding on the workflow it governs.
- The solo-implementation fallback in `multi-agent/issue-campaign.md` lines 11 to 19 is literally the solo procedure, so the round 3 exclusion was scoped to batch implementation agents and deliberately leaves the pass available in that mode.
- `multi-agent/review.md`'s "The author always completes Self-Review alone", `multi-agent/SKILL.md`'s "Self-Review remains solo for every author and every implementation branch", and the discussion skill's exclusion list were each checked and are all still true unchanged.
- `AGENTS.md` needs no edit. Its index lines for review, issue campaigns, and multi-agent workflows remain accurate, and its maintenance rule limits edits to repository-contract changes.

### Correction to the body above

The body says the multi-agent contrast paragraph "is descriptive and grants that agent nothing". That was true when the body was written and is now understated: round 3 made the exclusion explicit rather than leaving it to inference, so the paragraph states outright that a batch implementation agent runs no pass. Per the pull-request skill the correction stays in a comment so the thread preserves chronology.

PR #921, opened 2026-07-21
docs(agents): separate per-commit early warning from Self-Review

samchon:
## Intent

A solo campaign's cycle pull request carries many commits, and the maintainer asked whether each commit's review can be handed to a subagent while the author keeps implementing. The answer is that two layers exist and neither replaces the other, so this records both and keeps them distinguishable by name.

- **Layer 1, per-commit early warning.** After a commit lands the author may spawn one subagent to read it and report candidates while implementation continues. It reads and reports; it never edits, commits, pushes, or settles a design question. Its value is timing: a defect named while the commit is the newest thing written is cheap and nothing is built on it yet.
- **Layer 2, the author's own Self-Review over the whole base-to-head diff before merge.** Unchanged, undelegated, unpartitioned, and not satisfied by any number of per-commit passes.

Naming is the load-bearing part. If the early-warning pass is called a Self-Review, a reader concludes the gate already ran and the whole-surface round quietly disappears, which is exactly the failure the fresh-start rule exists to prevent.

## Scope

| File | Share |
| --- | --- |
| `.agents/skills/review/SKILL.md` | New `Early Warning Is Not Self-Review` section: the pass is a different activity under a different name, it delegates nothing the Non-Negotiable Review Law governs, the law's four rules still bind the author's own round, and a commit-sized reader cannot see cross-commit defects. |
| `.agents/skills/issue-campaign/development.md` | Layer 1 as implementation cadence, placed next to the implement-without-interruption paragraph #919 added, plus a precise amendment to the no-subagent rule in the four governing rules. |
| `.agents/skills/multi-agent/SKILL.md` | One paragraph of contrast so a reader of that skill does not assume the same topology: several reviewers each take the whole declared surface and the lead adjudicates, versus one reader on one commit's slice with the author's final whole-surface round still the gate. |
| `.agents/skills/issue-campaign/SKILL.md` | One-sentence precision fix. Its opening said the main agent "never spawns or delegates to a subagent", which would have flatly contradicted the new carve-out for a reader entering the campaign from the parent skill. It now says the main agent delegates no decision, edit, or review round, and links to the cadence. |

The design is stated once and linked from the other documents rather than repeated. `Self-Review remains solo for every author and every implementation branch` in the multi-agent skill stays true and stays where it is; this adds a parallel early-warning reader, it does not loosen that line.

## Deliberately not added

- No change to the Non-Negotiable Review Law's own text. Its first sentence had to keep meaning exactly what it says, so the carve-out is a separate named section that leaves the law verbatim.
- No new sibling document. The material is short enough to live in the owning sections.
- No `AGENTS.md` change. This is not a new skill area, a rename, or a rule that must apply before any skill loads.
- No procedure for how to brief the early-warning subagent beyond what it may and may not do. Prescribing a brief format would outrun the settled design.
- No repository-wide sweep for other prose about delegation. The four touched files are the verified surface where the new rule changes meaning.

## Verification

- `prettier --write` on all four touched Markdown files, using the repository config. No reflow beyond the added lines; `git diff --stat` is 20 insertions and 2 deletions.
- `git diff --check` clean.
- Every link touched resolves: `../review/SKILL.md#early-warning-is-not-self-review` against the new `## Early Warning Is Not Self-Review` heading, and `development.md#implement-and-write-tests` (relative and `../issue-campaign/` forms) against the existing `## Implement And Write Tests` heading.
- No build or test was run. The change is Markdown only.
- Solo Self-Review over the base-to-head diff follows in a comment on this pull request.

### DIFF
--- a/.agents/skills/issue-campaign/SKILL.md
+++ b/.agents/skills/issue-campaign/SKILL.md
@@ -5,7 +5,7 @@ description: "Defines the default solo repository-wide issue campaign for ttsc: 
 
 # Issue Campaign
 
-An issue campaign is a repeatable solo sequence of exhaustive discovery, issue publication, one unified implementation pull request, and renewed discovery. The main agent owns every phase and delegates no decision, edit, or review round to a subagent. Implementation may run one read-only [per-commit early-warning pass](development.md#implement-and-write-tests) alongside it.
+An issue campaign is a repeatable solo sequence of exhaustive discovery, issue publication, one unified implementation pull request, and renewed discovery. The main agent owns every phase and delegates no decision, edit, or review round to a subagent. The only subagent it may spawn is the read-only [per-commit early-warning pass](development.md#implement-and-write-tests), which reads a landed commit while implementation continues.
 
 Use the [multi-agent skill](../multi-agent/SKILL.md) and its issue-campaign procedure instead only when the user explicitly asks for a parallel or multi-agent issue campaign.
 
--- a/.agents/skills/issue-campaign/SKILL.md
+++ b/.agents/skills/issue-campaign/SKILL.md
@@ -5,7 +5,7 @@ description: "Defines the default solo repository-wide issue campaign for ttsc: 
 
 # Issue Campaign
 
-An issue campaign is a repeatable solo sequence of exhaustive discovery, issue publication, one unified implementation pull request, and renewed discovery. The main agent owns every phase and never spawns or delegates to a subagent.
+An issue campaign is a repeatable solo sequence of exhaustive discovery, issue publication, one unified implementation pull request, and renewed discovery. The main agent owns every phase and delegates no decision, edit, or review round to a subagent. Implementation may run one read-only [per-commit early-warning pass](development.md#implement-and-write-tests) alongside it.
 
 Use the [multi-agent skill](../multi-agent/SKILL.md) and its issue-campaign procedure instead only when the user explicitly asks for a parallel or multi-agent issue campaign.
 
--- a/.agents/skills/issue-campaign/development.md
+++ b/.agents/skills/issue-campaign/development.md
@@ -13,7 +13,7 @@ Read this document in full when the user authorizes implementation pull requests
 
 Four rules govern the implementation phase:
 
-- The main agent performs all implementation, test authoring, CI diagnosis, review, and cleanup. Do not spawn or delegate to a subagent.
+- The main agent performs all implementation, test authoring, CI diagnosis, review, and cleanup. The only permitted subagent is the read-only [per-commit early-warning pass](#implement-and-write-tests), which decides and changes nothing.
 - Put every accepted, implementation-ready issue in the current cycle into one pull request. The issue DAG controls implementation order inside that pull request, not pull-request count.
 - Work in the current checkout and one topic branch. Do not create a clone or worktree for a solo campaign or its Self-Review.
 - The pull request's ordinary CI and a clean solo Self-Review are the acceptance gates. Repair every red CI lane in that same pull request, even when the failure predates the campaign or is unrelated to its original issues.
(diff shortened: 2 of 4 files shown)


samchon:
## Solo Self-Review

Four complete rounds under the review skill's Non-Negotiable Review Law, each over the whole base-to-head surface plus uncommitted changes, each restarted from scratch after a correction.

**Round 1, four defects, all corrected.**

- The carve-out said "the law's first sentence keeps its exact meaning", but the prohibition on spawning a subagent is the law's second sentence, and a position reference breaks the moment that paragraph is edited. It now names the prohibition.
- "what exists only between commits" mis-stated the class. Two of the three examples are cross-file inside one commit's reach, so it now reads "across files and commits".
- "Under the gate's name it reads as the gate already run" was awkward and buried the mechanism, replaced with the reader inference that causes the round to disappear.
- In `development.md`, "which is why it leaves the rule above intact" pointed vaguely at a rule two sections up while reading as the adjacent paragraph. It now states the decisions that stay with the main agent.

**Round 2, two defects, both corrected.**

- The review skill was granting the spawn ("an author may run"), which is not its share and contradicts its own frontmatter contract, `This skill never spawns review agents`. It now describes the pass and states that solo campaign development authorizes and bounds it. Permission lives in one place.
- `issue-campaign/SKILL.md` mixed "main agent" and "author" in adjacent clauses.

**Round 3, one defect, corrected.** `issue-campaign/SKILL.md` called the pass "the one exception" to delegating a decision, edit, or review round, but the pass is none of those, so it is not an exception to that sentence. It is now stated as the only subagent the main agent may spawn, matching `development.md`'s rule bullet.

**Round 4, clean.** Full reread of all four files and the consequence surface with nothing left to improve.

### Consequence surface checked

- Grepped every `subagent`, `delegate`, and `spawn` statement across `.agents/` and `AGENTS.md`. One contradiction existed beyond the three planned files and is fixed: `issue-campaign/SKILL.md` opened with "never spawns or delegates to a subagent", which would have forbidden the new pass outright for anyone entering the campaign from the parent skill. `multi-agent/SKILL.md`'s `Do not let agents re-delegate` and `Self-Review remains solo for every author and every implementation branch` are both still true and untouched.
- Frontmatter checked on all three skills. `review` still never spawns review agents, `issue-campaign` still describes solo Self-Review, `multi-agent` still keeps Self-Review solo. No trigger scope changed, so no frontmatter edit was needed.
- `AGENTS.md` skill index re-read against all four files. Its review, issue-campaign, and multi-agent pointers remain accurate, and a carve-out inside a skill is not one of the repository-contract changes that warrant editing it.
- Placement verified against the section each document owns. The cadence sits directly after the implement-without-interruption paragraph #919 added, because it is precisely how a commit gets read without pausing the sequence.

### Verification

- `prettier --write` with the repository config on every touched file after each round; final run reports all four `(unchanged)`.
- `git diff --check` clean after every round.
- Anchors verified against the actual headings after the last edit: `#early-warning-is-not-self-review` and `#implement-and-write-tests` each match exactly one heading, and all eight references to them resolve.
- No build or test run. Markdown only, and the repository's standing rule keeps validation on CI.

samchon:
Superseded by #920, which merged at 04:07 UTC while this branch was in Self-Review and carries the same two-layer design across the same four files.

I compared the two before deciding, rather than resolving the conflict and merging:

- `review/SKILL.md` gains the carve-out and the naming rule in both, as `## Commit Early-Warning Pass` on master.
- `issue-campaign/development.md` gains the cadence in both, including why a commit-sized reader cannot replace the gate.
- `multi-agent/SKILL.md` gains the contrast paragraph in both.
- `issue-campaign/SKILL.md`'s blanket "never spawns or delegates to a subagent" is carved out in both.

#920 also covers a case this branch missed: `multi-agent/issue-campaign.md` sends every parallel campaign agent to read the solo development procedure, so the carve-out reached batch implementation agents and collided with the no-re-delegation rule. Its contrast paragraph closes that.

Merging this branch would restate a settled design a second time in the same documents, against the documentation skill's rule to state each rule at its owning document. Nothing here is a needed delta on top of master, so it is closed rather than trimmed. Two details this branch carried that master states less explicitly, in case either is wanted later: hand the subagent the commit diff and the repository instructions it must read, and record a disproved candidate so a later pass does not rediscover it.

Issue #922, opened 2026-07-21
fix(lint): force-break a non-empty block in expression position, as Prettier does


samchon:
## Problem

Prettier's block printer emits a hardline after `{` for **any** non-empty block, so a block written on one line in expression position is always expanded:

```ts
// input
export const o = { m() { return 1; } };
run(() => { a(); b(); });
```

```ts
// prettier@3.8.3 --parser typescript
export const o = {
  m() {
    return 1;
  },
};
run(() => {
  a();
  b();
});
```

`ttsc format` leaves both byte-identical.

This is the half of #856 that the brace-ownership split deliberately did not take. #856 fixed the case where the cascade broke a body out and stranded the `}`; for an expression-nested block it fixed that by making both structural rules cede, because `format/indent` hands such a brace to `format/print-width` and splitting a body whose brace nothing can restore is what produced the stranded hybrid in the first place. The result is correct for those two rules and short of the oracle, because the third rule never fires.

## Evidence

`packages/lint/test/format/command_format_brace_ownership_negative_twins_test.go` records the current output as `TestCommandFormatLeavesExpressionNestedOneLineBlockWhole`, naming it a divergence rather than parity.

`format/print-width` does not reach either input:

- `run(() => { a(); b(); });` has one callback argument, so `callForcesFunctionBreak` is false; the node holds no newline and fits `printWidth`, so the fast path in `rules_format_print_width.go` returns before laying anything out.
- `export const o = { m() { return 1; } };` has no `PropertyAssignment` initializers, so `fastPathForcesBreak` is false for the same reason.

Both are under `printWidth`, so no width-driven path can be expected to reach them: the break is unconditional in Prettier, not width-driven.

## Consequence surface

| Input class | ttsc | Prettier |
| --- | --- | --- |
| object literal with a method body on one line | unchanged | expanded, trailing comma added |
| call argument arrow/function with a body on one line | unchanged | expanded |
| the same shapes nested in an array element, a conditional arm, a parenthesized expression | unchanged | expanded |
| a block already spanning lines | unchanged | unchanged |
| an EMPTY block `{}` in the same positions | unchanged | unchanged, must stay so |

Reached from `ttsc format`, `ttsc fix`, the `format/print-width` diagnostic in `ttsc check`, and the LSP and VS Code format command.

## Approach

[...]

PR #917, opened 2026-07-21
fix(campaign): cycle 2 — graph traces, factory parity, lint semantics, and the two CI gaps

samchon:
## Intent

Cycle 2 of the 2026-07-21 campaign: one pull request carrying every campaign issue this cycle could close on established evidence, plus the repairs CI found along the way.

Claimed empty first so no overlapping work started before the code was written. Each commit names the issues it closes in its own subject, so the closing record is the history rather than this description.

## What this closes

Closes #784.
Closes #786.
Closes #794.
Closes #795.
Closes #796.
Closes #797.
Closes #806.
Closes #809.
Closes #814.
Closes #816.
Closes #818.
Closes #822.
Closes #834.
Closes #835.
Closes #848.
Closes #852.
Closes #856.
Closes #858.
Closes #871.
Closes #915.
Closes #916.

| Area | Issue | What landed |
| --- | --- | --- |
| CI | #915 | `push: branches: [master]` on the seven validation workflows. Their `concurrency` blocks already carried the non-cancelling push shape — only the trigger was missing. |
| repo | #916 | The tree is formatted to the pinned Prettier and a check gates it, so the 529-file drift cannot return. |
| test | #806 | Every committed executable test has a CI owner, proven mechanically in both directions rather than by a list someone maintains. |
| lint | #858 | `checkThenables` inverted to the upstream meaning: it selects which types the rule examines, so the default leaves thenables out of scope. |
| lint | #794 | A rejection handler reached through an optional chain is credited. The optional-chain walk is reused rather than written twice. |
| lint | #795 | `no-array-delete` decides from the target's type, not the subscript's syntax — closing a false positive on `Record` / index signatures and a false negative on `delete arr[next()]` together. |
| lint | #796 | `require-await` counts a `for await` loop and exempts a promise-returning function, both per the upstream contract. |
| lint | #797 | `no-namespace` exempts `declare global`, using upstream's own discriminator. No shim change was needed. |
| lint | #856 | The shared-line brace gets an owner. `format/indent` claims a stranded `}` once its frame spans lines and breaks a member or clause header off the `{` that opened its body; `format/statement-split` cedes the one case whose brace no rule can restore. The cascade's fixed point is now a form Prettier also produces, so `ttsc check` can no longer bless a malformed file. |
| graph | #784 | A hub-suppressed dispatch fanout is carried as an omission, so an open trace reports `truncated`. |
[...]

Issue #839, opened 2026-07-20
fix(lint): measure display width the way the published `getStringWidth` contract promises


samchon:
## Problem

`packages/lint/linthost/display_width.go` is the single width authority for the formatter. Its own doc comment states the contract:

> `displayWidth` returns the number of terminal columns a string occupies, matching Prettier's getStringWidth

and `website/src/content/docs/lint/format.mdx` publishes the same contract to users:

> Width is measured in **display columns** (Prettier's `getStringWidth`), not UTF-8 bytes: a wide East Asian or emoji code point counts as two columns

Neither statement is true. `displayWidth` sums a per-code-point width, and that width comes from `isWideRune`, a hand-written 13-range table. Two faults follow.

1. The table is neither the Unicode East_Asian_Width property nor Prettier's emoji set. Its lowest emoji range starts at `U+1F300`, so the whole `U+2000`-`U+2FFF` pictograph block is charged one column, and several ranges it does claim are narrow in Prettier.
2. There is no grapheme clustering. Prettier segments text into extended grapheme clusters and charges an emoji cluster two columns; `displayWidth` charges each code point in a ZWJ sequence, skin-tone sequence, or keycap sequence separately.

The result is that `ttsc format` breaks lines at the wrong column in both directions: a line containing `⭐`, `✅`, `❌`, or `⚡` is measured short and stays flat past `printWidth`, and a line containing one of the over-claimed ranges is measured long and breaks early. Emoji in log strings, test names, and CLI banners are ordinary TypeScript content, so this is not an exotic input class.

Affected: every user-visible formatting path, since the three consumers are the layout engine's fit checks, the `format/print-width` rule, and `format/clause-join`.

## Evidence

The oracle is the repository's own pinned Prettier. `package.json` declares `"prettier": "^3.8.3"` and `pnpm-lock.yaml` resolves `prettier@3.8.3`.

### Side by side

`prettier.util.getStringWidth` from the installed 3.8.3, against `displayWidth` from `packages/lint/linthost/display_width.go` compiled verbatim (the file imports only `unicode`, so it runs unmodified in a scratch `package main`):

| Input | Code points | Prettier | ttsc | Fault |
| --- | --- | --- | --- | --- |
| `⭐` | U+2B50 | 2 | 1 | under |
| `✅` | U+2705 | 2 | 1 | under |
| `❌` | U+274C | 2 | 1 | under |
| `⚡` | U+26A1 | 2 | 1 | under |
[...]

PR #924, opened 2026-07-21
docs(agents): stop a reverted commit's closing keyword from closing its issue

samchon:
## Problem

A squash merge concatenates every commit message in the pull request into the merge commit body, and GitHub parses closing keywords from that whole body. A `Close #n` line therefore closes its issue on merge even when a later commit in the same pull request reverted the code that earned it. The keyword outlives the fix it described.

The solo campaign workflow makes this likely rather than rare. `development.md` deliberately puts one `Close #n: <issue title>` line in every commit that resolves an issue, collects many such commits into one cycle pull request, and expects CI-driven and review-driven reverts inside that same pull request.

It has now cost two incidents in two consecutive cycles:

| Issue | Cycle | Pull request | Revert | Outcome |
| #839 | 2 | #917 (`ad6e9d60d`) | `e27e79804` | closed by the merge, reopened 38 seconds later by hand |
| #922 | 3 | #923 (`806e540ef`) | `b60747e1c` | closed by the merge, reopened by hand |

In both cases the reverting commit's subject was `Revert "Close #n: ..."`, which still carries the keyword and the number into the squashed body.

## Change

One document, two insertions, in the campaign procedure that creates the hazard.

`Implement And Write Tests` owns the `Close #n` commit-line convention, so the prevention rule sits directly under it:

> A revert inside the pull request must drop the closing keyword with the code. When a later commit reverts one that carried a `Close #n` line, write the reverting commit without that phrase and remove any `Closes #n` line for that issue from the pull-request body. A squash merge concatenates every commit message in the pull request into the merge commit body, so GitHub reads the reverted commit's keyword there and closes an issue whose fix no longer exists at `HEAD`.

`Merge And Clean Up` owns the merge gate, so the detection and the correction sit there:

> Before merging, reconcile the closing keywords against what survives at `HEAD`. `git log origin/master..HEAD` still shows a reverted commit's subject, so read the whole range and confirm every issue the merge will close has a surviving fix.

and, appended to the post-merge verification step:

> Reopen any issue the squash merge closed without a surviving fix, and comment that the merge closed it mechanically.

Prevention at commit time, detection before merge, correction after merge. The last is a backstop rather than redundancy: once a reverting commit is pushed, its subject cannot be withdrawn from the range without rewriting history, so a pull request that already contains one must plan to reopen.

## Scope

The pull-request skill was considered and left untouched. The `Close #n` commit convention exists only in the campaign document, and the pull-request skill's `Issue Campaign Override` already requires completing `.agents/skills/issue-campaign/development.md` before any campaign push, pull request, or merge, so a cross-reference would duplicate an existing link. No new rule, no restructuring, no new skill.

## Verification

Documentation only, no code path touched. `npx prettier --write .agents/skills/issue-campaign/development.md` reported no reformatting beyond the edits, which keeps `--prose-wrap never` satisfied. No build or test was run, and none applies to this diff.

https://claude.ai/code/session_01LfjPNKwN63jfUhF4FuUW6X

### DIFF
--- a/.agents/skills/issue-campaign/development.md
+++ b/.agents/skills/issue-campaign/development.md
@@ -58,6 +58,8 @@ Implement without interruption. Write each piece's tests as that piece lands ins
 
 Close each issue from the commit that earns it. End the commit message with one `Close #n: <issue title>` line per resolved issue, so a commit that resolves several issues carries several lines. GitHub matches the keyword and the number and ignores the title tail, so the line closes the issue normally while the log stays legible without opening each number.
 
+A revert inside the pull request must drop the closing keyword with the code. When a later commit reverts one that carried a `Close #n` line, write the reverting commit without that phrase and remove any `Closes #n` line for that issue from the pull-request body. A squash merge concatenates every commit message in the pull request into the merge commit body, so GitHub reads the reverted commit's keyword there and closes an issue whose fix no longer exists at `HEAD`.
+
 Post a pull-request comment after each commit naming what that commit landed and which issues it resolved. The comment is the running ledger for a reader who does not read the diff, not a closing mechanism: GitHub closes an issue only from a commit message or the pull-request body.
 
 Once a commit lands, the main agent may spawn one read-only subagent as a commit early-warning pass over that commit and keep implementing while it runs. The pass reads that one commit and reports candidates. It never edits, commits, pushes, or makes an implementation decision. Its value is timing: a defect named while that code is the newest thing written costs little to correct, and nothing has been built on top of it yet.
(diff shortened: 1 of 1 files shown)


samchon:
Self-Review found two things the opening body predates.

**Round 2, a false guarantee in the rule itself.** The first wording said the reverting commit must drop the keyword, which reads as though a clean revert subject spares the issue. It does not. A revert adds a commit, it does not remove one, so the original `Close` line is still in the range the squash concatenates. The cycle 2 merge proves it: `ad6e9d60d`'s body carries the original closing line at line 865 and its `Revert` quote at line 943, so rewriting only the revert subject would have closed the issue anyway. The paragraph now says the commit-time rule is necessary and not sufficient, names `git revert`'s default subject as the trap both incidents hit, and routes to the merge gate that has to reopen the issue. The pre-merge line now tells the reader to read the whole range rather than the revert.

**Round 3, this pull request tripping its own rule.** Reconciling the range the way the new text prescribes turned up a closing keyword for the cycle 3 issue inside one of this branch's own commit messages, quoted as an example. Under the default concatenation GitHub would have parsed it and this documentation change would have claimed an issue it does not fix. This merges with an explicit squash subject and body instead of the concatenated default, so no quoted example reaches the parser. The two commits and the diff are unchanged.

A complete round over the final head found nothing further. Verification is `npx prettier --check` on the one changed file, which passes; `.agents` is inside the repository format scope and outside `.prettierignore`, so the next `pnpm format` sweep produces no diff here. No build or test lane applies to this change.

PR #935, opened 2026-07-22
feat(lint): complete statement and expression-block reflow

### LINKED ISSUE #922: fix(lint): force-break a non-empty block in expression position, as Prettier does

### LINKED ISSUE #928: feat(lint): nine statement kinds have no printer, so a reflowed block half-expands
## Problem

Nine statement kinds have no printer, so a reflow that reaches into a block body expands the body and leaves those statements frozen at whatever width the source wrote them. The result is half-expanded output — a shape Prettier never produces.

```ts
// input
run(() => {
  for (const x of xs) { f(x); }
});
```

```ts
// prettier@3.8.3
run(() => {
  for (const x of xs) {
    f(x);
  }
});
```

Today `ttsc format` leaves the `for` line alone, because `printBlock` dispatches each statement and `KindForOfStatement` falls through to `verbatim`.

## Evidence

`packages/lint/linthost/print_dispatch.go`'s `dispatchNode` handles fifteen kinds. Of the statement kinds a block body can hold, it handles two — `ExpressionStatement` and `ReturnStatement`. These nine fall through:

| Kind | Reached from a block body |
| --- | --- |
| `KindIfStatement` | yes |
| `KindForStatement` | yes |
| `KindForOfStatement` | yes |
| `KindForInStatement` | yes |
| `KindWhileStatement` | yes |
| `KindTryStatement` | yes |
| `KindSwitchStatement` | yes |
| `KindVariableStatement` | yes |
| `KindThrowStatement` | yes |

`verbatim` is correct as a fallback — it round-trips source bytes and reports `covered = !nodeSpansMultipleLines`, so a multi-line statement makes the enclosing rule abstain rather than corrupt anything. Nothing is broken today. The gap is that the printer cannot express a layout for these, so any force-break that reaches a block containing one produces the half-expanded shape above.

That is why this blocks #922 rather than being a defect on its own: #922 asks the formatter to expand a non-empty block in expression position, and doing so through a body holding any of these nine seals the expansion exactly one level deep. A first attempt at #922 was reverted (`b60747e1c`) for precisely this, among two other causes.

## Consequence surface

| Shape | Today | With a printer |
| --- | --- | --- |
| A block body holding only expression statements and returns | fully laid out | unchanged |
| A block body holding an `if`, loop, `try`, `switch`, `throw`, or declaration | that statement frozen at its source width | laid out |
| A statement already spanning lines | `covered = false`, enclosing rule abstains | laid out |
| A statement carrying comments | must keep round-tripping verbatim | unchanged |

Reached from `ttsc format`, `ttsc fix`, the `format/print-width` diagnostic in `ttsc check`, and the LSP and VS Code format command.

## Approach

Invariant: a statement the printer is asked to lay out inside a reflowed block is laid out, not carried through at its source width.

Architectural owner is `dispatchNode` in `print_dispatch.go` plus one printer per kind, alongside the existing `printBlock`, `printExpressionStatement`, and `printReturnStatement`.

Each kind is independently useful and independently testable, so this can land in pieces. A sensible order is the one that unblocks the most common block bodies first: `VariableStatement`, `IfStatement`, the three `for` forms and `while`, then `TryStatement`, `SwitchStatement`, and `ThrowStatement`.

Two constraints carried from the existing printers:

- a statement whose subtree carries a comment the minted layout would drop must keep returning `verbatim` with a hard `covered = false`, the way `printObjectLiteral` and `printParenthesizedExpression` do through `listHasInterItemComments`;
- the header of a control-flow statement — `for (…)`, `if (…)`, `catch (e)` — is the `format/clause-join` rule's surface, so a printer for these must not reflow the header and put the two in disagreement.

| Case | Expected |
| --- | --- |
| Positive | Each of the nine, written on one line inside a reflowed block, matches the pinned Prettier |
| Positive | The `#922` witness `run(() => { for (const x of xs) { f(x); } });` expands throughout |
| Negative twin | An already-Prettier-formatted body is byte-identical |
| Negative twin | A statement carrying a comment round-trips verbatim and the enclosing rule abstains |
| Boundary | `format.endOfLine: "crlf"` on every inserted break |
| Boundary | Idempotence, and convergence inside the ten-pass cascade budget |
| Regression | Every existing `format/print-width` and `command_format_*` case unchanged |

Add a conformance-corpus row measured against the pinned Prettier for each kind, not only a cascade test. The `#922` revert turned on exactly that distinction: its cascade test used a shape that passed while the shape in the issue's own acceptance table did not, and only an oracle row would have caught it.

```bash
node scripts/test-go-lint.cjs
```

## Coordination

Blocks #922, together with the break-after-`=>` arrow-body variant that `print_nodes_call.go` records as `DEFERRED (architectural, not a one-line predicate fix)`. The third prerequisite, an object-member printer, landed in #925.

samchon:
## Intent

Cycle 7 of the 2026-07-21 campaign completes the printer prerequisites for #922 and the nine missing statement printers in #928.

## What this changes

- Dispatches the three `for` forms, `while`, block/`else if` forms of `if`, `try`/`catch`/`finally`, `switch`, variable initializers, and thrown expressions through structured printers.
- Preserves control-flow headers and clause keywords as source gaps so `format/clause-join` remains their owner; braceless and unsafe comment-bearing forms abstain.
- Adds structured object-member bodies, comment-only blocks, and a break-after-`=>` layout for call/conditional/JSX expression bodies.
- Makes non-empty expression-position blocks deny the print-width flat fast path through calls, arrays, conditionals, object members, and parenthesized expressions; empty blocks remain flat.
- Documents the expanded formatter surface.

## Verification

- Pinned-Prettier conformance rows cover every #928 statement kind and every #922 expression position, including a parenthesized object body and a call-bodied arrow.
- Cascade tests cover exact output, idempotence, braceless/comment negative twins, empty blocks, and CRLF.
- `.vscode/gofmt-2spaces.sh` completed on every touched Go file and `git diff --check origin/master` passes.
- Per the campaign handoff, no local build/test was run on this resource-constrained machine; ordinary GitHub Actions is the build, test, and format gate.

Closes #928.
Closes #922.


samchon:
## Cycle scope completed at `1ca7fbca5`

The inherited head implemented the five single-body control-flow printers and left #928 open. This commit completes the same accepted dependency chain in #935:

- #928: all nine named statement kinds now dispatch their controlled blocks, switch clauses, initializers, or thrown expressions; braceless and comment-bearing negative paths remain conservative.
- #922: non-empty expression-position blocks now deny the flat fast path only after object members, statement bodies, and call-bodied arrows have structured layouts. Empty and comment-only blocks, conditional/array positions, CRLF, and idempotence are pinned.
- The conformance corpus has one pinned-Prettier row for every #928 kind and every #922 expression position. The formatter guide now describes the shipped surface.

Local evidence under the campaign's standing no-build/no-test directive: `.vscode/gofmt-2spaces.sh` completed on every touched Go file and `git diff --check origin/master` passed. An exact-file Prettier write exceeded 30 seconds and was terminated without a result; ordinary CI remains the build, test, and format acceptance gate.

Resolves #928 and #922 through the commit closing lines.

samchon:
Self-Review correction: the mixed block/braceless negative twin now starts from the canonical multi-line form for its owned block. That keeps the assertion focused on the braceless else arm instead of allowing another format rule to legitimately change the fixture. Pushed as 474863f7e; CI and the full review restart from this head.

samchon:
Second Self-Review correction (0427f6854): the break-after-=> arrow variant now has the matching argument-list close layout, so a hugged expression-bodied last argument renders its trailing comma and outer ) on separate lines instead of attaching the parenthesis to the nested call. The unconditional-break walk now also follows parenthesized expressions, with a corpus/cascade witness for () => ({ method() { ... } }). CI and the full review restart from this head.

samchon:
CI follow-up at 7481ba4a8: fixed both lint regressions reported by the Linux and Windows Go lanes. Comment-only blocks now expand only when their original interior is single-line; multiline comment-only blocks remain uncovered. Switch case blocks and clauses now inspect trivia against their actual clause/statement item ranges, so unsupported inter-item comments make the enclosing print-width edit abstain.

samchon:
Final solo Self-Review completed from the current head across the complete diff and its formatter consequence surface: no remaining actionable finding. All 51 ordinary/gating check lanes are green on $head; the separate non-gating downstream 
estia lane is still running. There are no reviews or unresolved review threads, and git diff --check origin/master...HEAD passes. Per the maintainer campaign directive, no local build/test was run; GitHub Actions is the build/test gate.

Issue #948, opened 2026-07-22
perf(ttscserver): stop scanning the whole document on every completion request


samchon:
## Problem

`ttscserver` scans the entire document, twice, on every `textDocument/completion` request, before it knows whether the plugin corpus can contribute anything at that cursor — and it does so synchronously inside the serial editor-to-upstream pump, ahead of forwarding the request to tsgo.

The cost is linear in document size and is paid whether or not the cursor sits anywhere near a hint trigger. On an ordinary file it is invisible. On a large generated file it is milliseconds of added completion latency plus head-of-line blocking on every message queued behind it.

## Evidence

`Proxy.completionItemsFor` (`packages/ttsc/internal/lspserver/lsp_completion_request.go:76-108`) performs two full-prefix walks:

```go
offset, ok := offsetForPosition(text, params.Position.Line, params.Position.Character)
...
items, filter := matchCompletionHints(
  hints,
  linePrefixAt(text, offset),
  cursorInJSDoc(text, offset),
)
```

`offsetForPosition` walks newline by newline from byte zero. `cursorInJSDoc` calls `lexicalScopeAt`, which by design scans forward from the start of the document — correctly so, since comment state is a property of the token stream, not of the bytes near the cursor.

Both arguments are evaluated eagerly. But the scope is only ever consumed by `completionHintApplies` (`packages/ttsc/internal/lspserver/lsp_completion.go:41-60`), which also requires `strings.Contains(linePrefix, hint.After)` — a test on the current line alone. When no hint's trigger literal appears in the line prefix, no hint can apply, and the document scan that produced `inJSDoc` was pure waste. That is the overwhelming majority of completion requests: the corpus published by `jsdoc/check-tag-names` triggers on `@`.

This runs on the critical path, not beside it. `pumpEditorToUpstream` (`lsp_proxy.go:284-311`) is a single serial loop: it reads a frame, calls `handleEditorEnvelope`, and only then writes the frame upstream. `handleEditorEnvelope` dispatches `case methodCompletion: return p.handleCompletionRequest(env)` with no goroutine, unlike `executeCommand`, `formatting`, and the codeAction response path, which all hand off with `go`. So the scan delays tsgo's own completion work by its full duration and blocks every subsequent editor message — the next `didChange`, a `$/cancelRequest`, a hover — behind it.

### Measurement

[...]

Issue #947, opened 2026-07-22
perf(lint): answer lsp-hints without loading a Program the corpus does not need


samchon:
## Problem

`@ttsc/lint` answers `lsp-hints` by loading a full Program, and `ttscserver` asks for it on every document save.

The verb exists so a project rule can publish an editor-completion corpus. The corpus is a projection of what a project rule's `Check` found, so the design loads a Program before asking the rules. But the host cannot know whether *any* rule has a corpus to publish until after that load, and the only built-in publisher's corpus does not depend on the Program at all.

The result is that a project which publishes nothing — the common case — pays a cold process spawn plus a full parse, bind, and (when any enabled rule is type-aware) a full type-checker build, per save, to be told `[]`.

## Evidence

`RunLSPHints` (`packages/lint/linthost/hints.go`) loads rules, builds the engine, loads the Program, and runs the whole project cycle before it ever asks which rules publish hints:

```go
prog, parseDiags, err := loadProgram(opts.cwd, opts.tsconfig, loadProgramOptions{
  forceNoEmit:      true,
  needsRuleChecker: engine.NeedsTypeChecker(),
  projectIdentity:  opts.projectIdentity,
})
...
return writeJSON(collectProjectHints(prog.runProjectCycle(engine)))
```

`collectProjectHints` only then discovers whether a rule implements `publicrule.HintRule`, and `evaluateProject` (`packages/lint/linthost/project_engine.go:169-178`) skips every project rule that is not `Declared` or whose severity is `Off`. Both facts are known from the registration tables and the resolved config — before a Program exists.

`needsRuleChecker: engine.NeedsTypeChecker()` is engine-wide, so one enabled type-aware *file* rule forces a full checker build for a verb that needs no checker. The only built-in publisher declares `NeedsTypeChecker() bool { return false }`.

The corpus that work produces, for the one built-in publisher, is a constant. `jsdocCheckTagNamesHintRule` (`packages/lint/linthost/rules_jsdoc.go:17-51`) has a `Check` whose entire body is `ctx.SetState(jsdocCheckTagNamesHintState{})`, and `Hints` returns a sorted projection of the package-level `knownJSDocTags` map. No Program input reaches the result.

[...]

PR #946, opened 2026-07-22
fix(campaign): harden runtime safety and release gates

samchon:
## Intent

Claim the complete accepted second cycle of the 2026-07-22 solo issue campaign in one pull request.

This implementation-free head reserves #939, #940, #941, #942, #943, #944, and #945 while their integrated implementation is in progress.

## Scope

- restore failed-module cache eviction in the playground CommonJS sandbox;
- reconcile exact dependency state across incremental playground installs;
- keep Metro cross-run cache keys non-reusable after snapshot persistence failures;
- refuse explicit ttsc clean targets that contain the project;
- encode Windows junction paths as literal data in all three config loaders;
- build and smoke the `@ttsc/factory` publication artifacts in ordinary CI;
- preserve JavaScript string, operator, and optional-call semantics in playground dependency discovery; and
- record PR updates and Self-Review rounds as formal `COMMENT`-event GitHub reviews.

## Verification

Implementation and regression coverage are pending.

Per maintainer direction, no local build or test command will run; ordinary pull-request CI owns all build and test verification.

## Coordination

This pull request owns the complete accepted cycle and will retain issue-level commits and acceptance matrices inside one integrated head.

The body is the claim-time intent statement and intentionally contains no closing keywords; each implementation commit will close only the issue it actually earns.

PR #949, opened 2026-07-22
perf(lsp): answer editor guidance without rebuilding the world

samchon:
Claims the complete accepted cycle for the LSP editor-guidance channel: issue #947 and issue #948. This pull request owns both; no other pull request in this campaign will carry either.

Both issues are performance defects in the channel `@ttsc/lint` opened so a project rule can guide an editor. Neither changes what the channel answers — only what it costs to answer.

## Scope

**#947 — `lsp-hints` loads a Program the corpus does not need.** The verb loads a full Program, and runs the project cycle, before it can discover whether any rule has a corpus to publish. `ttscserver` asks for it on every `didSave`. A project that publishes nothing pays a cold process spawn plus a full parse and bind to be told `[]`.

**#948 — the completion path scans the whole document per request.** `completionItemsFor` walks the document twice — once to resolve the cursor, once to decide the lexical scope — before the line-local test that is the only thing able to admit a hint, and it does so synchronously inside the serial editor-to-upstream pump, ahead of forwarding the request to tsgo.

## Measured baseline

`nestia.examples/shopping/packages/backend`, 280 files, 8.60 MB. The sidecar was built from the tree and its verbs invoked directly; `lsp-command-ids` loads no Program and is therefore the spawn floor. Min of 3, Go 1.26.4, windows/amd64.

| Path | Now |
| --- | --- |
| `lsp-command-ids` (spawn floor) | 59 ms |
| `lsp-diagnostics` one-shot | 909 ms |
| `lsp-hints` one-shot, per save | 753 ms |
| resident `lsp-diagnostics`, warm | 143 ms |
| resident `lsp-diagnostics`, after a save | 113 ms |
| completion request, 1 MB document | 3.55 ms |

The resident daemon added in #755 and made incremental in #765 already turned the per-save diagnostics cost from 909 ms into 113 ms. `lsp-hints` was never routed through it, so it reintroduced a full cold load on the same path.

## Verification

Pending. This is the claim commit; no implementation exists yet. Acceptance is the ordinary pull-request CI plus a complete solo Self-Review over the base-to-head diff, with each issue's acceptance matrix proved case by case.

Both issues require that `ttsc lint` — `check`, `fix`, and `format` — stays byte-identical, since the CLI path is structurally isolated from this channel and must remain so.


samchon (commented):
Implementation ledger for `7e2013fa9` and `7b7f80115`, both pushed to head `7b7f80115`.

**`7e2013fa9` — resolves #947.** `computeLSPHints` splits out of `RunLSPHints` the way `computeLSPDiagnostics` and `computeLSPCodeActions` already did, so the resident daemon can answer the verb. Before it touches a Program it asks the new `Engine.hintPublishers`, which reads the declared project rules and the `rule.HintRule` assertion from the registration table — both known without parsing anything. The checker decision narrows with it, and `residentProgramCache` gained a one-Program-per-project invariant so that narrowing cannot make the daemon hold two.

Tests: `lsp_hints_skips_the_program_when_nothing_publishes` (a project whose tsconfig does not exist answers `[]` silently, which is only reachable if the loader was never called), `lsp_serve_answers_hints_from_the_resident_daemon`, and the two `resident_cache_*` cases covering both arrival orders.

**`7b7f80115` — resolves #948.** `anyCompletionHintTriggers` gates the document scan behind the line-local trigger test that `completionHintApplies` already required, and `indexByteFrom` delegates to `strings.IndexByte`.

Tests: `lsp_completion_gate_never_refuses_a_matching_line` asserts the gate is never stricter than the match it precedes — a false there would silently lose a hint — and `lsp_index_byte_from_bounds_its_search_start` covers the slice bounds the delegation introduced.

**Local evidence.** The full `internal/lspserver` Go suite passes. The four new lint cases plus the pre-existing `lsp_hints_publishes_jsdoc_check_tag_names_corpus` pass through `scripts/bench-go-lint.cjs`. A sidecar built from this branch produces byte-identical corpora against the shipped one in all three rule configurations measured. CI remains the acceptance gate; Self-Review over this exact diff starts now.

**Deferred, and stated so the issues are not read as fully closed on their own terms.** #947's third reduction (a corpus declaring it needs no Program at all) is untouched: it widens the public `rule` package, and reductions 1 and 2 already satisfy every row of that issue's acceptance matrix. #948's "per-request cost no longer scales with document size" is only partly met — the scope scan is gone, but `offsetForPosition` still walks lines from byte zero. Removing that walk needs a line-start index maintained across `didChange`, where `cacheDidChangeText` already pays the same cost to splice the buffer; that is a change against a path #948 does not own, and it is recorded rather than silently skipped.

samchon (commented):
Solo Self-Review, rounds 1 through 3, over the `origin/master...HEAD` diff. Each round restarted from the whole surface after the previous round's corrections landed.

**Round 1 — three findings, all corrected in `1835b9da4`.**

The completion gate tested only a hint's trigger literal, so a hint carrying a scope this host does not know still opened it and still paid the document scan the match was going to refuse regardless. Scope is a property of the hint, not of the cursor, so it belongs on the cheap side of the split. `completionHintCouldApply` now holds every admission condition independent of the cursor and `completionHintApplies` is that predicate plus the scope decision — which makes the gate a superset of the match by construction rather than by inspection.

`computeLSPHints` dereferenced a Program that `acquireProgram` may return nil without diagnostics; `residentProgramCache.acquire` guards exactly that shape, so it is reachable. It now answers the empty corpus, matching what the parse-diagnostics branch already did and avoiding a JSON `null` that a decoder expecting an array would read as a broken plugin.

`hintPublishers` allocated a slice of rule names nothing consumed. It is a predicate and now returns one.

**Round 2 — one finding, corrected in `6df8c6b38`.** Inserting `residentProgramKey` between `acquire`'s doc comment and `acquire` left the cache's central contract documenting a two-line string builder while `acquire` documented nothing. Both comments now sit on their own functions, and `acquire`'s states the one-Program-per-project invariant this branch gave it. The resident verb list had also grown to "one per save, one per cursor, and one per save".

**Round 3 — one finding, corrected in `7bd4448ec`.** `lsp_hints_skips_the_program_when_nothing_publishes` would have passed just as well if the verb had stopped loading a Program entirely, or had started swallowing every loader failure. Its negative twin declares a publisher against the same unloadable project, so the loader must be reached and its failure must still surface.

**Verified in this round and unchanged.** `./internal/lspserver` is in `scripts/test-go-ttsc.cjs`'s package list and `packages/lint/test/command` is flattened by the lint runner, so both new suites are covered by ordinary CI rather than only locally. `acquire` still holds its mutex across `loadProgram` exactly as before. The checker-superset fallback hands a checker-bearing Program to a checker-free verb, and `evaluateProject` still passes nil to any rule that declined one, so a rule cannot silently gain a checker it disowned. The two cache keys cannot coexist, so no lookup order is ambiguous.

**Stated rather than fixed.** Routing hints through the daemon serializes them behind diagnostics and code actions on the same sidecar mutex, where they previously ran in their own process. That is a deliberate trade: the serialized total is far below the parallel one it replaces, and whichever verb runs first pays the incremental update the other then reuses.

Round 4 starts over the settled head once CI reports.

samchon (commented):
**Red lane disposition: `deploy`.**

It fails here and on `origin/master` alike, for a cause this branch does not touch:

```
Error: [typia-graph] catalogs.samchon.typia must be one exact version, found "&typia"
  at readExactTypiaPin (website/build/typia-dependency-graph.cjs:396)
```

`70a29d4c0` ("bump up typia") made the workspace catalog carry a YAML anchor where the website's graph builder expects a literal version. No file on that path appears in this diff.

The campaign rule is to repair every red lane in the same pull request rather than call it another contributor's problem, and this is not that call. The fix already exists and is in flight: #946 changes exactly `pnpm-workspace.yaml`, `website/build/typia-dependency-graph.cjs`, and `website/test/typia-dependency-graph.test.cjs`. Landing the same edit here would put it in two open pull requests and collide on all three files.

`#946` is still a draft, so it cannot go first. `master` carries no branch protection, so nothing required is blocked, and merging this leaves `deploy` exactly as red as it already is on `master` — it goes green when #946 lands, not through anything this branch could do without duplicating that work.

Stated here rather than left implicit, so the lane is not read as passing and not read as ignored.

samchon (commented):
Solo Self-Review, rounds 4 and 5, completing the ledger for head `a7941e740`.

**Round 4 — one finding, corrected in `a7941e740`.** `lsp_hints_still_loads_the_program_for_a_declared_publisher` asserted that the loader failure reached stderr but discarded the exit code, so an implementation that reported the failure and then returned success would have passed it. The corpus verb answers an empty array with exit 0 for every condition an editor should ignore — a Program that could not load is not one of those, and the test now pins that distinction.

**Round 5 — clean.** A complete fresh round over the whole `origin/master...HEAD` surface produced nothing that survived verification.

What it covered beyond re-reading the diff:

- Every document mentioning `lsp-hints` was re-read for a claim this branch made false. `driver-api.mdx`'s `CompletionHintRefresher` paragraph describes the refresh scheduling, which is unchanged. `walkthroughs/lint.mdx` lists the verb in the dispatch table only. `protocol.mdx` carries the three corrections already made.
- CI coverage of the new suites was confirmed rather than assumed: `./internal/lspserver` is in `scripts/test-go-ttsc.cjs`'s package list, and `packages/lint/test/command` is flattened into `linthost` by the lint runner, so neither new suite is local-only.
- `collectProjectHints` cannot return nil on the surviving path: `runProjectCycle` returns nil only for a nil Program or nil engine, both excluded above it, and `evaluateProject` always populates `results`. The wire keeps `[]` rather than `null`.
- The daemon's two cache keys cannot coexist under the new invariant, so no lookup order is ambiguous and `invalidate` / `applyChanges` need no change.
- Lock ordering across the new routing: `serveRun` releases `residentMu` before taking the sidecar's `mu`, so hints joining the daemon adds no new ordering.

**Carried forward as stated trade-offs, not defects.** Hints now serialize behind diagnostics and code actions on one sidecar mutex, bounded by the existing 30-second per-request timeout; previously a hung hint rule could not delay diagnostics at all. And `offsetForPosition` still walks lines from byte zero, so a very large document keeps a per-request cost proportional to its size, six times smaller than before but not removed. Both are recorded on this pull request and in the campaign ledger rather than folded into the issues as if resolved.

Five rounds, four of them producing corrections, the fifth clean. Merge waits on CI settling at this same head.

PR #956, opened 2026-07-22
fix: preserve package and filesystem boundary semantics

### LINKED ISSUE #950: fix(wasm): keep descriptor authority and file identity across namespace mutations
## Problem

`@ttsc/wasm` still lets an open descriptor lose its authority or silently change which file it addresses.

First, `fs.ftruncate` never checks `entry.writable`. A descriptor opened with the published read-only access mode can resize the file and update its mtime even though sibling `read` and `write` operations enforce the captured access mode.

Second, every descriptor stores only `entry.path`, and every descriptor operation looks the node up again in the pathname map. That makes namespace mutations rewrite file identity:

- after `unlink(path)`, a descriptor already open on that file starts returning `ENOENT` instead of retaining access to the unlinked file;
- when `rename(source, destination)` overwrites an existing destination, a descriptor already open on the old destination silently starts reading and writing the source file now occupying that pathname.

Both outcomes report successful namespace mutations and then give an existing descriptor authority over a state different from the one it opened.

## Evidence

The owner is `packages/wasm/src/createMemFS.ts`.

- `IDescriptor` retains `path`, `position`, `readable`, `writable`, and `append`, but not the opened `INode`.
- `read`, `writeThroughDescriptor`, `fstat`, and `ftruncate` all call `nodes.get(entry.path)`.
- `unlink` deletes the path from `nodes` without preserving the node for open descriptors.
- `moveSubtree` replaces the destination entry and rewrites source-descriptor paths to the destination. It does not distinguish descriptors opened on the overwritten destination, so both descriptor populations subsequently resolve the replacement node.
- `ftruncate` validates descriptor existence, node kind, and length, then resizes without checking `entry.writable`.

Closed #813 added descriptor access-mode state and enforced it for `read` and `write`; its rename regression covers a source descriptor following the moved source, not an open destination being overwritten, and it does not test read-only `ftruncate`. Closed #671 fixed path mutations but likewise has no open-unlink or overwritten-destination descriptor case. This is their unsealed descriptor-identity remainder, not a duplicate.

The published MemFS guide says descriptors carry their access mode, mutations enforce POSIX-style invariants, and a mutation that cannot satisfy its contract returns an error instead of changing state differently.

## Consequence surface

Affected operations are `read`, `write`, `writeSync`, `fstat`, and `ftruncate` after `unlink` or a replacement `rename`, plus `ftruncate` on a read-only descriptor. Source descriptors must still follow a successful rename, multiple descriptors to one live file must still share bytes with independent cursors, and descriptors 1/2 and pipe ends remain special streams rather than filesystem nodes.

## Approach

An open descriptor must retain the opened node identity in addition to its current pathname label. Descriptor operations use that retained node; the namespace map owns only path lookup. Renaming the source may update its descriptor labels, but overwriting a destination must not retarget descriptors that already refer to the displaced node. Unlink removes the name while the node remains reachable through its descriptors until they close.

`ftruncate` must enforce the same writable capability as `write`, failing before any byte or mtime change. Preserve `writeFile`/`O_TRUNC` updates to an existing mapped file node so already-open descriptors to that same file observe a real in-place truncation or replacement.

- A read-only descriptor rejects `ftruncate` with `EBADF`; bytes, mtime, and cursor stay unchanged.
- A writable descriptor still grows and shrinks through `ftruncate`, including zero-fill growth.
- After unlinking an open file, its fd can still read, write, `fstat`, and `ftruncate` according to its access mode, while path-based `stat` reports `ENOENT`.
- Renaming `source` over `destination` leaves a pre-opened source fd attached to source bytes at the new name and a pre-opened destination fd attached to the displaced destination bytes; writes through either do not cross over.
- Closing the last detached fd does not recreate a pathname.
- Existing source-rename, directory-rename, multiple-descriptor, stdio, pipe, and bad-fd regressions remain green.

Add focused cases under `tests/test-wasm/src/features/memfs`; PR CI owns build and test execution.

### LINKED ISSUE #951: fix(wasm): give low-level mkdir single-directory semantics
## Problem

The low-level MemFS `fs.mkdir(path, mode, callback)` operation is implemented by calling the host convenience helper `mkdirp(path)`. It therefore has recursive and idempotent semantics that the Node-style callback does not promise:

- `fs.mkdir("/missing/child")` creates both ancestors and succeeds instead of returning `ENOENT` for the missing parent;
- `fs.mkdir("/existing")` succeeds instead of returning `EEXIST`;
- a normalized alias reaching an existing directory likewise reports a new successful creation.

Go's wasm runtime uses this callback for filesystem operations. A caller of `os.Mkdir` cannot distinguish creating one directory from creating a whole chain or from doing nothing at all.

## Evidence

`packages/wasm/src/createMemFS.ts` exposes two deliberately different layers:

- `host.mkdirp` is the recursive, idempotent convenience helper documented for seeding a virtual project;
- `host.fs.mkdir` is part of `IWasmExecFS`, the Node-style callback surface consumed by `wasm_exec.js`.

The latter currently does only:

```ts
mkdir(p, _perm, callback) {
  try {
    mkdirp(p);
    callback(null);
  } catch (err) {
    callback(err as NodeJS.ErrnoException);
  }
}
```

Closed #811 explicitly separated low-level `fs.open` from parent-creating host helpers and added atomic tree preflight, but did not cover `fs.mkdir`. Existing tests cover `host.mkdirp` only. No open or closed issue owns the low-level mkdir contract.

## Approach

Keep `host.mkdirp` unchanged. Implement `fs.mkdir` as a single-directory mutation: normalize first, reject the root or an existing target with `EEXIST`, require the immediate parent to exist and be a directory (`ENOENT`/`ENOTDIR`), then add exactly one directory node. Every rejection must be atomic.

- Creating one child under an existing directory succeeds and creates only that node.
- A missing immediate parent returns `ENOENT` and creates nothing.
- A file-valued parent returns `ENOTDIR` and creates nothing.
- An existing file, existing directory, and `/` return `EEXIST` without mutation.
- Normalized aliases make the same decision as their canonical path.
- `host.mkdirp` remains recursive, idempotent, and atomic on a conflicting segment.
- Existing `open`, `writeFile`, `rename`, `rmdir`, and tree-invariant regressions remain green.

Add a focused MemFS feature case; PR CI owns build and test execution.

### LINKED ISSUE #952: fix(ttsc): pass Windows Go command wrappers as literal argv data
## Problem

On Windows, a `.cmd` or `.bat` Go tool selected through `TTSC_GO_BINARY` is interpolated directly into a `shell: true` command string. Quoting protects spaces and ordinary shell metacharacters, but `cmd.exe` expands `%NAME%` segments even inside quotes. Because `%` is legal in a Windows path, a wrapper stored under a directory such as `C:\tools\%SDK%\go.cmd` is executed from a different expanded path or cannot be found.

Every source-plugin Go operation uses this helper: `go mod edit`, `go env`, `go version`, and `go build`. The failure can therefore appear as a missing toolchain, a wrong compiler cache identity, or a failed plugin build depending on which call reaches the path first.

## Evidence

`packages/ttsc/src/plugin/internal/buildSourcePlugin.ts::spawnGoTool` does:

```ts
const shell = shouldSpawnGoToolThroughShell(goBinary);
return spawnSync(shell ? `"${goBinary}"` : goBinary, [...args], {
  ...options,
  shell,
});
```

`shouldSpawnGoToolThroughShell` selects this branch for Windows `.cmd`/`.bat` files. The command is consequently parsed by `cmd.exe`, where quoting does not disable percent expansion.

Closed #804 proved the same Windows rule on the VS Code command-shim boundary and fixed it by passing quoted volatile values through environment placeholders, which `cmd` expands once. The source-plugin helper is a separate implementation and was not changed by that issue. Existing fake-Go wrapper tests use ordinary temp paths and do not cover a literal or environment-shaped percent segment.

## Approach

Treat the wrapper path and every argument as data. Use an explicit `cmd.exe /d /s /c` launch with one fully quoted payload whose volatile values arrive through a per-call environment indirection, or an equivalent mechanism that guarantees the child sees every original string byte-for-byte. Do not use a string interpolation whose content is reparsed for `%NAME%` expansion.

The non-Windows and native `.exe`/extensionless branches remain direct `spawnSync` calls.

- On Windows, a fake `go.cmd` beneath a path containing a literal `%`, a set `%NAME%`-shaped segment, spaces, `&`, parentheses, and `^` receives the exact argv for `version`, `env`, `mod edit`, and `build`.
- Literal `%`, `%%`, `%` at either end, and an unknown `%NAME%` arrive unchanged.
- The caller's environment and the Go build/cache environment remain intact apart from private shim variables.
- A missing wrapper still produces the existing toolchain-not-found guidance.
- Native Windows executables and every POSIX launcher keep their current command/args shape.

The proving case must run in the Windows PR CI lane; local build/test execution is intentionally not part of this campaign.

### LINKED ISSUE #953: fix(paths): resolve aliases with the compiler host's case sensitivity
## Problem

`@ttsc/paths` ignores the compiler host's filesystem case-sensitivity when it resolves a configured alias target back to a Program source file. On Windows and case-insensitive macOS volumes, tsgo can resolve a target whose spelling differs only by case, but the plugin's map lookup misses and leaves the alias unchanged in emitted JavaScript.

For example, a project with `src/feature.ts` and `paths: { "@feature": ["SRC/FEATURE"] }` is a valid case-insensitive resolution. The compiler includes the source, but `@ttsc/paths` searches a case-sensitive Go map for the uppercase candidate and emits `require("@feature")`/`import "@feature"`, which the runtime cannot resolve without tsconfig paths support.

## Evidence

`packages/paths/driver/paths.go::newRewriter` stores each Program source as:

```go
name := normalizePath(file.FileName())
out.sourceFiles[name] = name
```

`lookupSource` then indexes the map with case-preserving normalized candidates for the exact path, extension fallbacks, and index fallbacks. `prog.FS.UseCaseSensitiveFileNames()` is available and already read by this file, but it is used only while inferring a common root and never when building or querying `sourceFiles`.

No feature or Go helper test exercises a case-only difference. Closed #831 fixed filesystem-identity comparison in the unplugin adapter, not this Go plugin.

## Consequence surface

Exact targets, extensionless targets, explicit extensions, and directory/index targets all share the same map and can miss. The returned value must retain the Program's real spelling so output-path computation still uses the compiler-owned source name. Case-sensitive hosts must continue rejecting a target whose case does not identify a real Program file.

## Approach

Give the rewriter one canonical key function derived from `UseCaseSensitiveFileNames`. Store Program source names under canonical keys and query every exact/extension/index candidate through the same function, while keeping the map value as the original normalized path.

- On a case-insensitive rewriter, exact, extensionless, explicit-extension, and index aliases resolve despite case-only spelling differences and return the Program's original path.
- On a case-sensitive rewriter, the same mismatches do not resolve.
- Drive-letter and directory-component case differences use the same rule.
- Existing target ordering, ambiguous-extension precedence, JSON handling, root/outDir mapping, and emitted suffix regressions remain unchanged.
- A Windows end-to-end project proves the emitted runtime specifier is rewritten; platform-independent helper cases prove both host modes.

PR CI owns Go and feature test execution.

### LINKED ISSUE #954: fix(ttsc,playground): honor general exports subpath patterns
## Problem

Two custom package-export resolvers claim Node-style subpath support but implement only a narrow trailing-star subset.

1. The ttsc plugin loader's dedicated `ttsc` condition resolver matches exact subpaths only. A package that exposes runtime-free plugin descriptors through a valid pattern such as `"./plugins/*": { "ttsc": "./descriptors/*.js", "default": "./runtime/*.js" }` falls through to ordinary `require.resolve`, which selects the default runtime branch or fails. This recreates the invalid-descriptor/self-hosting cycle the private condition exists to prevent.
2. The playground Execute sandbox considers only export keys ending in `/*`. Valid Node patterns with a literal suffix, such as `"./features/*.js": "./dist/features/*.cjs"`, are treated as unexported even though tsgo compiled the same package import successfully.

## Evidence

`packages/ttsc/src/plugin/internal/loadProjectPlugins.ts::selectExportTarget` detects whether `exports` is a subpath map and then returns only `record[subpath]`; it has no wildcard matcher or target substitution.

`packages/playground/src/sandbox/createSandboxRequire.ts::resolvePackageEntry` filters entries with `pattern.endsWith("/*")`, strips that fixed shape, and sorts by string length. It cannot match a `*` followed by a suffix.

Closed #820 required the playground resolver to honor the exports boundary and pattern precedence, but its implementation and tests cover only `./prefix/*`. The plugin condition tests cover only the root exact entry. No issue owns general single-star subpath patterns in either resolver.

## Approach

Implement Node-compatible single-`*` subpath matching and substitution in both owners. Match exact keys first, then eligible pattern keys using Node's specificity order; require both literal prefix and suffix to match; substitute the captured middle into every `*` in the selected target. Preserve manifest key order for condition objects and the existing active-condition sets.

Do not add ESM evaluation, package `imports`, or filesystem access to the sandbox. This issue is only about resolving the export maps both components already consume.

- Exact subpaths still beat patterns.
- `./features/*.js` matches `./features/a.js`, substitutes `a`, and rejects `./features/a.css`.
- Overlapping patterns select the Node-more-specific key, including cases where prefix and suffix specificity differ.
- Scoped packages and nested conditional/array targets retain their current behavior.
- A `ttsc`-condition pattern selects the runtime-free descriptor while ordinary package resolution still selects `default`.
- The playground sandbox resolves a suffix-bearing pattern, enforces an unmatched/`null` boundary, and keeps relative, `main`, root-export, JSON, cycle-cache, and trailing-`/*` regressions green.

Add focused ttsc plugin-loader and playground sandbox cases; PR CI owns execution.

### LINKED ISSUE #955: fix(playground): classify the complete non-registry dependency spec family
## Problem

The playground npm walker skips unsupported non-registry dependencies by testing a short list of string prefixes. Valid npm dependency specs outside that list are misclassified as registry ranges. A package with a dependency such as `git://host/repo.git`, `git@github.com:owner/repo.git`, `gitlab:owner/repo`, `bitbucket:owner/repo`, or the hosted shorthand `owner/repo` is queued as though the dependency name were on the npm registry and the source spec were a dist-tag.

The installer then fetches metadata for the wrong registry request and fails with a message such as `No npm dist-tag "owner/repo" exists`, even though the manifest never described a registry tag. This can block installing an otherwise ordinary registry package.

## Evidence

`packages/playground/src/npm/internal/npmRegistry.ts::isRegistryRange` returns false only for:

```ts
/^(file:|link:|workspace:|portal:|git\+|github:|https?:)/
```

Every other string is sent to `selectVersion`, whose non-semver branch interprets it as an npm dist-tag. Closed #803 established that this browser installer follows registry dependencies and leaves unsupported non-registry protocols uninstalled, but its implementation/acceptance witnesses cover only the listed prefixes. It does not classify the other npm git, hosted, URL, or path spellings.

## Approach

Classify dependency specs by the grammar they represent rather than by an incomplete deny-prefix list. Enqueue only valid semver ranges, exact versions, npm dist-tags, and the already-supported `npm:` aliases. Skip the complete unsupported path/URL/git/hosted family consistently. A malformed or ambiguous value must not be reinterpreted as a registry tag solely because it lacks one known prefix.

Preserve required-vs-optional handling for supported registry dependencies and all mounted-version reconciliation from #803/#940.

- Semver ranges, exact versions, ordinary dist-tags, and `npm:` aliases still enqueue.
- `file:`, `link:`, `workspace:`, `portal:`, HTTP(S), `git+*`, `git://`, `ssh://`, scp-style `git@host:`, `github:`, `gitlab:`, `bitbucket:`, and hosted `owner/repo` specs do not enqueue registry requests.
- Relative, absolute, home-relative, and Windows path specs are not treated as tags.
- A fake registry test proves no fetch occurs for each skipped dependency and that supported siblings still install.
- Required peer/dependency constraints, optional peers, optional `@types` 404, aliases, max-package limits, and abort handling remain unchanged.

PR CI owns playground build and tests.

samchon:
## Intent

Claim the complete accepted third cycle of the 2026-07-22 solo issue campaign in one pull request.

## Scope

- retain MemFS descriptor authority and node identity across unlink and replacement rename;
- give the low-level MemFS `mkdir` callback single-directory semantics while preserving `mkdirp`;
- pass Windows Go command wrappers and their arguments as literal data through `cmd.exe`;
- key `@ttsc/paths` source lookup with the compiler host's filesystem case sensitivity;
- honor general single-star package export patterns in the plugin loader and playground sandbox; and
- classify only registry-supported npm ranges, aliases, and dist-tags as registry requests.

Closes #950.
Closes #951.
Closes #952.
Closes #953.
Closes #954.
Closes #955.

## Verification

Focused regression coverage accompanies every implementation unit, including Windows-only process and case-sensitivity witnesses.

Per maintainer direction, no local build or test command ran. Ordinary pull-request CI owns all build and test verification; `typia.yml` and `nestia.yml` are not campaign gates.

## Review Ledger

Every pushed commit and every Self-Review round is recorded as a formal `COMMENT`-event GitHub pull-request review. Ordinary issue-style PR comments are not used for this ledger.

### DIFF
--- a/packages/playground/src/npm/internal/npmRegistry.ts
+++ b/packages/playground/src/npm/internal/npmRegistry.ts
@@ -278,12 +278,21 @@ export function enqueuePackageDependencies(
 }
 
 function isRegistryRange(range: string): boolean {
-  if (range.startsWith("npm:")) return true;
-  if (validRange(range) !== null) return true;
+  const spec = range.trim();
+  // Match npm-package-arg's file classification before its registry fallback:
+  // dot/home/root/drive paths and tar archive names are source specs even
+  // though their characters could also form a URI-safe dist-tag.
+  if (
+    /^(?:\.|~[\\/]|[\\/]|[a-zA-Z]:)/.test(spec) ||
+    /\.(?:tgz|tar\.gz|tar)$/i.test(spec)
+  )
+    return false;
+  if (spec.startsWith("npm:")) return true;
+  if (validRange(spec) !== null) return true;
   // npm accepts a dist-tag only when it is a non-empty URI-component-safe
   // token. Git, hosted, URL, and local-path specs all require punctuation that
   // encodeURIComponent escapes, so they cannot fall through as fake tags.
-  return range.length > 0 && encodeURIComponent(range) === range;
+  return spec.length > 0 && encodeURIComponent(spec) === spec;
 }
 
 export function toTypesPackageName(packageName: string): string {
--- a/packages/playground/src/npm/internal/npmRegistry.ts
+++ b/packages/playground/src/npm/internal/npmRegistry.ts
@@ -278,7 +278,12 @@ export function enqueuePackageDependencies(
 }
 
 function isRegistryRange(range: string): boolean {
-  return !/^(file:|link:|workspace:|portal:|git\+|github:|https?:)/.test(range);
+  if (range.startsWith("npm:")) return true;
+  if (validRange(range) !== null) return true;
+  // npm accepts a dist-tag only when it is a non-empty URI-component-safe
+  // token. Git, hosted, URL, and local-path specs all require punctuation that
+  // encodeURIComponent escapes, so they cannot fall through as fake tags.
+  return range.length > 0 && encodeURIComponent(range) === range;
 }
 
 export function toTypesPackageName(packageName: string): string {
--- a/packages/playground/src/sandbox/createSandboxRequire.ts
+++ b/packages/playground/src/sandbox/createSandboxRequire.ts
@@ -116,7 +116,7 @@ export function createSandboxRequire(
       }
       return tryPaths(`${pkg}/index.js`, `${pkg}/index.cjs`);
     }
-    // Subpath: honor exports["./subpath"] or exports["./subpath/*"] patterns.
+    // Subpath: honor exact exports and general single-star patterns.
     const exportsAny = pj.exports;
     if (
       exportsAny !== undefined &&
@@ -130,14 +130,18 @@ export function createSandboxRequire(
         return resolveExportTarget(pkg, entries[exactKey]);
       // Node resolves the most-specific wildcard, not insertion order.
       const patterns = Object.entries(entries)
-        .filter(([pattern]) => pattern.endsWith("/*"))
-        .sort(([a], [b]) => b.length - a.length);
-      for (const [pattern, target] of patterns) {
-        if (!pattern.endsWith("/*")) continue;
-        const prefix = pattern.slice(2, -1); // strip "./" and trailing "*"
-        if (!subpath.startsWith(prefix)) continue;
-        const rest = subpath.slice(prefix.length);
-        return resolveExportTarget(pkg, target, rest);
+        .map(([pattern, target]) => ({
+          pattern,
+          replacement: exportPatternReplacement(pattern, exactKey),
+          target,
+        }))
+        .filter(
+          (entry): entry is typeof entry & { replacement: string } =>
+            entry.replacement !== undefined,
+        )
+        .sort((a, b) => compareExportPatternKeys(a.pattern, b.pattern));
+      for (const { replacement, target } of patterns) {
+        return resolveExportTarget(pkg, target, replacement);
       }
     }
     return null;
@@ -256,6 +260,41 @@ export function createSandboxRequire(
     }
     return evaluate(resolved).exports;
   };
+}
+
+/** Capture the middle of one valid single-star exports key. */
+function exportPatternReplacement(
+  pattern: string,
+  subpath: string,
+): string | undefined {
+  const star = pattern.indexOf("*");
+  if (
+    !pattern.startsWith("./") ||
+    star === -1 ||
+    pattern.indexOf("*", star + 1) !== -1
+  ) {
+    return undefined;
+  }
+  const prefix = pattern.slice(0, star);
+  const suffix = pattern.slice(star + 1);
+  if (
+    subpath.length < prefix.length + suffix.length ||
+    !subpath.startsWith(prefix) ||
+    !subpath.endsWith(suffix)
+  ) {
+    return undefined;
+  }
+  return subpath.slice(prefix.length, subpath.length - suffix.length);
+}
+
+/** Node exports patterns rank longer prefixes, then longer full keys, first. */
+function compareExportPatternKeys(left: string, right: string): number {
+  const leftPrefix = left.indexOf("*");
+  const rightPrefix = right.indexOf("*");
+  if (leftPrefix !== rightPrefix) {
+    return rightPrefix - leftPrefix;
+  }
+  return right.length - left.length;
 }
 
 /**
--- a/packages/playground/src/sandbox/createSandboxRequire.ts
+++ b/packages/playground/src/sandbox/createSandboxRequire.ts
@@ -126,7 +126,11 @@ export function createSandboxRequire(
       const entries = exportsAny as Record<string, unknown>;
       // Exact match first.
       const exactKey = `./${subpath}`;
-      if (Object.prototype.hasOwnProperty.call(entries, exactKey))
+      if (
+        Object.prototype.hasOwnProperty.call(entries, exactKey) &&
+        !exactKey.includes("*") &&
+        !exactKey.endsWith("/")
+      )
         return resolveExportTarget(pkg, entries[exactKey]);
       // Node resolves the most-specific wildcard, not insertion order.
       const patterns = Object.entries(entries)
@@ -278,7 +282,7 @@ function exportPatternReplacement(
   const prefix = pattern.slice(0, star);
   const suffix = pattern.slice(star + 1);
   if (
-    subpath.length < prefix.length + suffix.length ||
+    subpath.length < pattern.length ||
     !subpath.startsWith(prefix) ||
     !subpath.endsWith(suffix)
   ) {
--- a/packages/ttsc/src/plugin/internal/buildSourcePlugin.ts
+++ b/packages/ttsc/src/plugin/internal/buildSourcePlugin.ts
@@ -1762,24 +1762,58 @@ function goToolchainNotFoundMessage(pluginName: string): string {
   );
 }
 
-function spawnGoTool(
+export function spawnGoTool(
   goBinary: string,
   args: readonly string[],
   options: SpawnSyncOptionsWithStringEncoding,
 ): SpawnSyncReturns<string> {
-  const shell = shouldSpawnGoToolThroughShell(goBinary);
-  // Under shell:true (a Windows .cmd/.bat wrapper) Node does not quote the
-  // command, so a wrapper at a path containing a space would be split by cmd.
-  // Quote it here; the go arguments ttsc passes never contain spaces, and
-  // quoting a space-free path is harmless.
-  return spawnSync(shell ? `"${goBinary}"` : goBinary, [...args], {
-    ...options,
-    shell,
-  });
+  if (!shouldSpawnGoToolThroughShell(goBinary)) {
+    return spawnSync(goBinary, [...args], options);
+  }
+  const inheritedEnv = options.env ?? process.env;
+  const shim = createWindowsGoCommandShim([goBinary, ...args]);
+  return spawnSync(
+    inheritedEnv.ComSpec ?? inheritedEnv.COMSPEC ?? "cmd.exe",
+    ["/d", "/s", "/c", shim.payload],
+    {
+      ...options,
+      env: { ...inheritedEnv, ...shim.environment },
+      shell: false,
+      // The /c payload is already one fully quoted Windows command line.
+      windowsVerbatimArguments: true,
+    },
+  );
 }
 
 function shouldSpawnGoToolThroughShell(goBinary: string): boolean {
   return process.platform === "win32" && /\.(?:bat|cmd)$/i.test(goBinary);
+}
+
+/** Pass volatile cmd wrapper arguments through one-pass environment expansion. */
+function createWindowsGoCommandShim(args: readonly string[]): {
+  environment: NodeJS.ProcessEnv;
+  payload: string;
+} {
+  const prefix = `TTSC_GO_COMMAND_SHIM_${crypto
+    .randomBytes(8)
+    .toString("hex")
+    .toUpperCase()}_ARG_`;
+  const environment = Object.fromEntries(
+    args.map((arg, index) => [prefix + index, quoteWindowsCommandArg(arg)]),
+  );
+  return {
+    environment,
+    // cmd expands each placeholder exactly once. Percent-shaped text inside an
+    // expanded value is not scanned as a second environment reference.
+    payload: `"${args.map((_, index) => `%${prefix}${index}%`).join(" ")}"`,
+  };
+}
+
+/** Quote one argv value for the Windows command-line parser used by cmd. */
+function quoteWindowsCommandArg(arg: string): string {
+  return `"${String(arg)
+    .replace(/(\\*)"/g, '$1$1\\"')
+    .replace(/(\\*)$/, "$1$1")}"`;
 }
 
 function goBuildEnv(
--- a/packages/ttsc/src/plugin/internal/buildSourcePlugin.ts
+++ b/packages/ttsc/src/plugin/internal/buildSourcePlugin.ts
@@ -1774,7 +1774,7 @@ export function spawnGoTool(
   const shim = createWindowsGoCommandShim([goBinary, ...args]);
   return spawnSync(
     inheritedEnv.ComSpec ?? inheritedEnv.COMSPEC ?? "cmd.exe",
-    ["/d", "/s", "/c", shim.payload],
+    ["/d", "/v:off", "/s", "/c", shim.payload],
     {
       ...options,
       env: { ...inheritedEnv, ...shim.environment },
--- a/packages/ttsc/src/plugin/internal/buildSourcePlugin.ts
+++ b/packages/ttsc/src/plugin/internal/buildSourcePlugin.ts
@@ -2537,11 +2537,8 @@ function resolveWindowsGoTool(
 
   const hasExtension = windowsFileNameHasExtension(binary);
   for (const candidate of candidates) {
-    if (hasExtension) {
-      if (isExecutableFile(candidate)) {
-        return { location: candidate, wrapper: false };
-      }
-      continue;
+    if (hasExtension && isExecutableFile(candidate)) {
+      return { location: candidate, wrapper: false };
     }
     for (const extension of [".com", ".exe"]) {
       const executable = candidate + extension;
@@ -2550,8 +2547,6 @@ function resolveWindowsGoTool(
       }
     }
   }
-
-  if (hasExtension) return { location: null, wrapper: false };
 
   const wrapperExtensions = windowsExecutableExtensions(env).filter((ext) =>
     /\.(?:bat|cmd)$/i.test(ext),
@@ -2580,9 +2575,7 @@ function executableSearchBases(
       ? splitWindowsSearchPath(readPathEnvironment(env))
           .map(unquoteWindowsSearchEntry)
           .filter((dir) => dir.length > 0)
-      : readPathEnvironment(env)
-          .split(path.delimiter)
-          .filter((dir) => dir.length > 0);
+      : readPathEnvironment(env).split(path.delimiter);
   const noDefaultCurrentDirectory =
     process.platform === "win32"
       ? (readWindowsEnvironmentValue(
@@ -2647,7 +2640,7 @@ function splitWindowsSearchPath(value: string): string[] {
     const character = value[index]!;
     if (quote !== "") {
       if (character === quote) quote = "";
-    } else if (character === '"' && index === start) {
+    } else if ((character === '"' || character === "'") && index === start) {
       quote = character;
     } else if (character === ";") {
       entries.push(value.slice(start, index));
@@ -2659,9 +2652,13 @@ function splitWindowsSearchPath(value: string): string[] {
 }
 
 function unquoteWindowsSearchEntry(location: string): string {
-  if (!location.startsWith('"')) return location;
-  const unquoted = location.slice(1);
-  return unquoted.endsWith('"') ? unquoted.slice(0, -1) : unquoted;
+  const first = location[0];
+  const withoutFirst =
+    first === '"' || first === "'" ? location.slice(1) : location;
+  const last = withoutFirst[withoutFirst.length - 1];
+  return last === '"' || last === "'"
+    ? withoutFirst.slice(0, -1)
+    : withoutFirst;
 }
 
 function windowsExecutableExtensions(
@@ -2680,7 +2677,7 @@ function readPathEnvironment(env: NodeJS.ProcessEnv = process.env): string {
     ? (readWindowsEnvironmentValue(env, "PATH") ??
         readWindowsEnvironmentValue(process.env, "PATH") ??
         "")
-    : (env.PATH ?? "");
+    : (env.PATH ?? "/usr/bin:/bin");
 }
 
 function readWindowsEnvironmentValue(
--- a/packages/ttsc/src/plugin/internal/buildSourcePlugin.ts
+++ b/packages/ttsc/src/plugin/internal/buildSourcePlugin.ts
@@ -194,7 +194,7 @@ export function buildSourcePlugin(opts: {
   const { dir, entry, source } = resolveSourceBuildTarget(opts);
   const overlayDirs = [...(opts.overlayDirs ?? findTtscOverlayDirs())].sort();
   const contributors = opts.contributors ?? [];
-  const goBinary = resolveGoCompiler(env);
+  const goBinary = resolveGoToolForBuild(resolveGoCompiler(env), env, dir);
   ensureExecutableGoToolchain(goBinary);
   const key = computeCacheKey({
     contributors,
@@ -1772,22 +1772,24 @@ export function spawnGoTool(
     return spawnSync(goBinary, [...args], options);
   }
   const inheritedEnv = options.env ?? process.env;
-  const resolved = findExecutablePath(
+  const resolved = resolveWindowsGoTool(
     goBinary,
     inheritedEnv,
     spawnWorkingDirectory(options.cwd),
   );
-  if (!isWindowsCommandWrapper(resolved ?? goBinary)) {
+  if (!resolved.wrapper) {
     return spawnSync(goBinary, [...args], options);
   }
   // Preserve the native spawn ENOENT contract before cmd.exe becomes the
   // actual child process. The callers use that code for the install guidance.
-  if (resolved === null) {
+  if (resolved.location === null) {
     return spawnSync(goBinary, [...args], options);
   }
-  const shim = createWindowsGoCommandShim([resolved, ...args]);
+  const shim = createWindowsGoCommandShim([resolved.location, ...args]);
   return spawnSync(
-    readWindowsEnvironmentValue(inheritedEnv, "COMSPEC") ?? "cmd.exe",
+    readWindowsEnvironmentValue(inheritedEnv, "COMSPEC") ??
+      readWindowsEnvironmentValue(process.env, "COMSPEC") ??
+      "cmd.exe",
     windowsGoCommandArgs(shim.payload),
     {
       ...options,
@@ -2289,15 +2291,19 @@ export function computeCacheKey(inputs: {
   tsgoVersion: string;
 }): string {
   const env = inputs.env ?? process.env;
+  const goBinary =
+    inputs.goBinary === undefined
+      ? undefined
+      : resolveGoToolForBuild(inputs.goBinary, env, inputs.dir);
   const hash = crypto.createHash("sha256");
   hash.update(`ttsc=${inputs.ttscVersion}\n`);
   hash.update(`tsgo=${inputs.tsgoVersion}\n`);
   hash.update(`platform=${process.platform}/${process.arch}\n`);
   hash.update(`entry=${inputs.entry}\n`);
-  if (inputs.goBinary !== undefined) {
-    hash.update(`go=${resolveGoCompilerIdentity(inputs.goBinary, env)}\n`);
+  if (goBinary !== undefined) {
+    hash.update(`go=${resolveGoCompilerIdentity(goBinary, env, inputs.dir)}\n`);
   }
-  hashGoBuildEnvironment(hash, inputs.goBinary, inputs.dir, env);
+  hashGoBuildEnvironment(hash, goBinary, inputs.dir, env);
   hashExternalGoBuildEnvironment(hash, env);
   hashSourceDirectory(hash, "plugin", inputs.dir);
   for (const [index, dir] of [...(inputs.overlayDirs ?? [])].sort().entries()) {
@@ -2389,23 +2395,42 @@ function isHashableFile(name: string): boolean {
 // content signature (byte size + nanosecond mtime). That signature changes if
 // a long-lived host rewrites the binary in place between calls, so the memo
 // re-derives the identity exactly as the un-memoized code would and the
-// cache-key bytes stay byte-for-byte identical to today. `go version` reads no
-// cwd/custom env (the spawn passes neither), so cwd is not part of the key.
+// cache-key bytes stay byte-for-byte identical to today. The selected compiler
+// path is shared by every build subprocess on Windows, while `go version` uses
+// the same effective cwd and environment as the cache-key `go env` query. The
+// memo key includes that context so an environment-sensitive wrapper cannot
+// lend its version result to another compiler invocation.
 const goCompilerIdentityCache = new Map<string, string>();
 
 function resolveGoCompilerIdentity(
   goBinary: string,
   env: NodeJS.ProcessEnv = process.env,
+  cwd: string = process.cwd(),
 ): string {
-  const resolved = resolveExecutableIdentityPath(goBinary, env);
-  const memoKey = goCompilerIdentityMemoKey(goBinary, resolved);
+  const selected = resolveGoToolForBuild(goBinary, env, cwd);
+  const resolved =
+    process.platform === "win32"
+      ? resolveRealPath(selected)
+      : resolveExecutableIdentityPath(selected, env, cwd);
+  const compilerEnv = goBuildEnv(selected, undefined, env);
+  const memoKey = goCompilerIdentityMemoKey(
+    goBinary,
+    resolved,
+    compilerEnv,
+    cwd,
+  );
   if (memoKey !== null) {
     const cached = goCompilerIdentityCache.get(memoKey);
     if (cached !== undefined) {
       return cached;
     }
   }
-  const identity = computeGoCompilerIdentity(goBinary, resolved);
+  const identity = computeGoCompilerIdentity(
+    selected,
+    resolved,
+    compilerEnv,
+    cwd,
+  );
   if (memoKey !== null) {
     goCompilerIdentityCache.set(memoKey, identity);
   }
@@ -2418,21 +2443,38 @@ function resolveGoCompilerIdentity(
 function goCompilerIdentityMemoKey(
   goBinary: string,
   resolved: string,
+  env: NodeJS.ProcessEnv,
+  cwd: string,
 ): string | null {
   try {
     const stat = fs.statSync(resolved);
-    return `${goBinary}\0${resolved}\0${stat.size}\0${stat.mtimeMs}`;
+    const context = crypto.createHash("sha256");
+    context.update(cwd);
+    for (const [key, value] of Object.entries(env).sort(([left], [right]) =>
+      left === right ? 0 : left < right ? -1 : 1,
+    )) {
+      if (value === undefined) continue;
+      context.update(`\0${key.length}:${key}${value.length}:${value}`);
+    }
+    return `${goBinary}\0${resolved}\0${stat.size}\0${stat.mtimeMs}\0${context.digest("hex")}`;
   } catch {
     return null;
   }
 }
 
-function computeGoCompilerIdentity(goBinary: string, resolved: string): string {
+function computeGoCompilerIdentity(
+  goBinary: string,
+  resolved: string,
+  env: NodeJS.ProcessEnv,
+  cwd: string,
+): string {
   if (!fs.existsSync(resolved)) {
     return "missing";
   }
   const version = spawnGoTool(goBinary, ["version"], {
+    cwd,
     encoding: "utf8",
+    env,
     windowsHide: true,
   });
   const versionText =
@@ -2446,9 +2488,20 @@ function computeGoCompilerIdentity(goBinary: string, resolved: string): string {
 function resolveExecutableIdentityPath(
   binary: string,
   env: NodeJS.ProcessEnv = process.env,
+  cwd: string = process.cwd(),
 ): string {
-  const resolved = findExecutablePath(binary, env, process.cwd());
+  const resolved = findExecutablePath(binary, env, cwd);
   return resolved === null ? binary : resolveRealPath(resolved);
+}
+
+/** Pin Windows PATH and command-wrapper lookup to one build-wide target. */
+function resolveGoToolForBuild(
+  binary: string,
+  env: NodeJS.ProcessEnv,
+  cwd: string,
+): string {
+  if (process.platform !== "win32") return binary;
+  return resolveWindowsGoTool(binary, env, cwd).location ?? binary;
 }
 
 function findExecutablePath(
@@ -2456,23 +2509,95 @@ function findExecutablePath(
   env: NodeJS.ProcessEnv,
   cwd: string,
 ): string | null {
-  if (path.isAbsolute(binary)) {
-    return findExecutableCandidate(binary, env);
-  }
-  if (hasPathSeparator(binary)) {
-    return findExecutableCandidate(path.resolve(cwd, binary), env);
-  }
-
-  const directories = readPathEnvironment(env)
-    .split(path.delimiter)
-    .map((dir) => stripPathQuotes(dir));
-  if (process.platform === "win32") directories.unshift(cwd);
-  for (const dir of directories) {
-    const candidate = path.resolve(cwd, dir, binary);
+  for (const candidate of executableSearchBases(binary, env, cwd)) {
     const resolved = findExecutableCandidate(candidate, env);
     if (resolved !== null) return resolved;
   }
   return null;
+}
+
+interface IWindowsGoToolResolution {
+  location: string | null;
+  wrapper: boolean;
+}
+
+/** Preserve libuv's native target before falling back to cmd/bat wrappers. */
+function resolveWindowsGoTool(
+  binary: string,
+  env: NodeJS.ProcessEnv,
+  cwd: string,
+): IWindowsGoToolResolution {
+  const candidates = executableSearchBases(binary, env, cwd);
+  if (isWindowsCommandWrapper(binary)) {
+    return {
+      location: candidates.find(isExecutableFile) ?? null,
+      wrapper: true,
+    };
+  }
+
+  const hasExtension = windowsFileNameHasExtension(binary);
+  for (const candidate of candidates) {
+    if (hasExtension) {
+      if (isExecutableFile(candidate)) {
+        return { location: candidate, wrapper: false };
+      }
+      continue;
+    }
+    for (const extension of [".com", ".exe"]) {
+      const executable = candidate + extension;
+      if (isExecutableFile(executable)) {
+        return { location: executable, wrapper: false };
+      }
+    }
+  }
+
+  if (hasExtension) return { location: null, wrapper: false };
+
+  const wrapperExtensions = windowsExecutableExtensions(env).filter((ext) =>
+    /\.(?:bat|cmd)$/i.test(ext),
+  );
+  for (const candidate of candidates) {
+    for (const extension of wrapperExtensions) {
+      const executable = candidate + extension;
+      if (isExecutableFile(executable)) {
+        return { location: executable, wrapper: true };
+      }
+    }
+  }
+  return { location: null, wrapper: false };
+}
+
+function executableSearchBases(
+  binary: string,
+  env: NodeJS.ProcessEnv,
+  cwd: string,
+): string[] {
+  if (path.isAbsolute(binary)) return [binary];
+  if (hasPathQualifier(binary)) return [path.resolve(cwd, binary)];
+
+  const directories =
+    process.platform === "win32"
+      ? splitWindowsSearchPath(readPathEnvironment(env))
+          .map(unquoteWindowsSearchEntry)
+          .filter((dir) => dir.length > 0)
+      : readPathEnvironment(env)
+          .split(path.delimiter)
+          .filter((dir) => dir.length > 0);
+  const noDefaultCurrentDirectory =
+    process.platform === "win32"
+      ? (readWindowsEnvironmentValue(
+          env,
+          "NoDefaultCurrentDirectoryInExePath",
+        ) ??
+        readWindowsEnvironmentValue(
+          process.env,
+          "NoDefaultCurrentDirectoryInExePath",
+        ))
+      : undefined;
+  if (process.platform === "win32" && noDefaultCurrentDirectory === undefined) {
+    directories.unshift(cwd);
+  }
+  return directories.map((dir) => path.resolve(cwd, dir, binary));
 }
 
 function findExecutableCandidate(
@@ -2480,9 +2605,7 @@ function findExecutableCandidate(
   env: NodeJS.ProcessEnv,
 ): string | null {
   if (isExecutableFile(candidate)) return candidate;
-  if (process.platform !== "win32" || path.extname(candidate) !== "") {
-    return null;
-  }
+  if (process.platform !== "win32") return null;
   // Probe every PATHEXT extension, not just `.exe`, so a compiler backed by a
   // `.cmd`/`.bat` wrapper is both launched correctly and hashed into the cache
   // key. Otherwise the wrapper reads as missing and changes do not invalidate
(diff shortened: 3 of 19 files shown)


samchon (commented):
## Commit Review — `cad995b29`

Reviewed the complete pushed commit against issues #950 through #955 and the cycle-three acceptance matrix.

- MemFS descriptors now retain opened node identity, enforce writable `ftruncate`, and keep low-level `mkdir` distinct from `mkdirp`.
- `@ttsc/paths` uses the compiler host's canonical file-name rule while retaining the Program's source spelling.
- Both custom export resolvers implement exact-first, prefix/suffix-aware single-star matching and full target substitution.
- The playground npm walker positively classifies supported registry ranges, aliases, and URI-safe tags, leaving non-registry sources unqueued.
- Windows Go command wrappers receive the wrapper path and every argument through one-pass environment indirection.
- Focused regressions cover each boundary, including Windows-only witnesses.

Static diff, consequence-surface, formatting, and whitespace review are clean. Per maintainer direction, no local build or test ran; ordinary PR CI owns execution, with `typia.yml` and `nestia.yml` excluded from the campaign gate.

samchon (commented):
## Overall Self-Review — Round 1

Reviewed PR #956 from `origin/master` through head `cad995b29` as one integrated change, including all 16 changed files and every newly discovered test entry.

### Result

Clean round: no sound blocking or non-blocking improvement remains on the published head.

### Verified surfaces

- Descriptor operations consistently use retained MemFS node identity; namespace operations still update names without retargeting displaced descriptors.
- Low-level `mkdir` rejects existing targets and missing/non-directory parents before mutation; host `mkdirp` is untouched.
- Case canonicalization is applied symmetrically to every paths source-map insertion and exact/extension/index lookup while stored values retain Program spelling.
- Export resolution keeps exact entries first, ranks matching patterns by prefix then full-key specificity, substitutes every target star, preserves condition order, and retains `null` boundaries.
- Registry classification preserves semver ranges, exact versions, `npm:` aliases, and URI-safe dist-tags while skipping the complete tested URL/git/hosted/path family.
- Windows command-wrapper launch preserves the caller environment, uses explicit `cmd.exe /d /s /c`, and carries the already quoted payload with verbatim Windows arguments; direct executables and POSIX launchers remain direct spawns.
- New tests sit under each suite's existing discovery root and use existing internal import/build conventions.

One test-order defect was found and corrected before publication: the read-only MemFS cursor witness initially consumed the cursor twice. The pushed head proves the rejected `ftruncate` leaves the original cursor available to the existing positive read.

Formatting and `git show --check` are clean. Per maintainer direction, no local build or test ran; ordinary PR CI owns execution, and `typia.yml` / `nestia.yml` are excluded from this campaign's required gate.

packages/playground/src/sandbox/createSandboxRequire.ts
```diff
@@ -258,6 +266,41 @@ export function createSandboxRequire(
...
+    !pattern.startsWith("./") ||
+    star === -1 ||
+    pattern.indexOf("*", star + 1) !== -1
+  ) {
+    return undefined;
+  }
+  const prefix = pattern.slice(0, star);
+  const suffix = pattern.slice(star + 1);
+  if (
+    subpath.length < pattern.length ||
```
samchon, line 285: Self-Review correction: Node requires `packageSubpath.length >= key.length`, which makes the single-star capture non-empty. Using the full pattern length here now rejects `./empty/.js` for `./empty/*.js` while retaining prefix/suffix matching and specificity.

packages/playground/src/npm/internal/npmRegistry.ts
```diff
@@ -278,7 +278,21 @@ export function enqueuePackageDependencies(
 }
 
 function isRegistryRange(range: string): boolean {
-  return !/^(file:|link:|workspace:|portal:|git\+|github:|https?:)/.test(range);
+  const spec = range.trim();
+  // Match npm-package-arg's file classification before its registry fallback:
```
samchon, line 282: Self-Review correction: npm-package-arg classifies dot/root/drive paths and tar archive suffixes before its registry fallback. This guard prevents URI-safe source names such as `archive.tgz` or `.local` from being mistaken for valid dist-tags.

samchon (commented):
## Commit Review — `09fe3315c`

Reviewed the resolver-precedence correction against the official Node `packageExportsResolve` / `patternKeyCompare` implementation and npm-package-arg classification order. The commit closes the two Self-Review gaps without widening either custom resolver beyond its existing package boundary. Formatting and whitespace review are clean; execution remains delegated to PR CI, excluding `typia.yml` and `nestia.yml` from the campaign gate.

samchon (commented):
## Overall Self-Review — Round 2

Fresh review of PR #956 at head `09fe3315c`, with the custom resolver rules compared directly against the official Node resolver and npm-package-arg implementation.

### Findings and resolution

1. The initial export-pattern matcher admitted an empty `*` capture because it compared against the sum of literal prefix and suffix lengths. Node compares the requested subpath against the full key length, and it excludes literal-star/trailing-slash requests from the exact branch. Both custom resolvers now apply those rules; a packed target that would expose the empty-capture error is covered by a negative regression.
2. The initial positive registry classifier correctly skipped punctuation-bearing sources but still admitted URI-safe tar archive names and dot-prefixed paths as dist-tags. npm-package-arg classifies those as files before registry fallback. The classifier now mirrors that precedence and the no-fetch regression covers dot paths and tar archives.

Both line-specific resolutions are recorded as inline comments in formal review `4754474550`.

### Result

Clean after correction: no further sound improvement remains in the full PR diff. Formatting and whitespace checks are clean. Per maintainer direction, no local build or test ran; ordinary PR CI owns execution, with `typia.yml` and `nestia.yml` excluded from the required gate.

samchon (commented):
Commit review for 39f08790b: merged origin/master after PR #957 landed, as requested. This brings the clarified issue-campaign discovery-saturation instructions onto the active branch without changing PR #956's implementation diff against the updated base. The branch is synchronized with master at 729c17e4e. A fresh complete Self-Review of the current base-to-head diff follows on this immutable head.

samchon (commented):
Commit review for 3cc578941: closes the remaining Windows command-shim literal-data boundary found during Overall Self-Review. The explicit cmd.exe invocation now forces /v:off so exclamation-mark text cannot become delayed environment expansion, and the Windows regression carries !NAME!, lone !, percent, quoting, whitespace, and metacharacter inputs through the wrapper unchanged. Refs #952. No local build or test command ran; PR CI owns execution.

samchon (commented):
Commit review for 57538435e: merged origin/master after PR #958 landed, as requested. The active branch now carries the mandatory Individual Self-Review and exhaustive Overall Self-Review policy at master 039f7c33e. This synchronization commit does not change PR #956's implementation diff against the updated base. The outstanding Individual Self-Review findings on 3cc578941 remain active and will be recorded and corrected before the next Overall Self-Review round.

packages/ttsc/src/plugin/internal/buildSourcePlugin.ts
```diff
@@ -1762,26 +1762,60 @@ function goToolchainNotFoundMessage(pluginName: string): string {
...
-  const shell = shouldSpawnGoToolThroughShell(goBinary);
-  // Under shell:true (a Windows .cmd/.bat wrapper) Node does not quote the
-  // command, so a wrapper at a path containing a space would be split by cmd.
-  // Quote it here; the go arguments ttsc passes never contain spaces, and
-  // quoting a space-free path is harmless.
-  return spawnSync(shell ? `"${goBinary}"` : goBinary, [...args], {
-    ...options,
-    shell,
-  });
+  if (!shouldSpawnGoToolThroughShell(goBinary)) {
```
samchon: [P1] Accepted: before switching an absolute .cmd/.bat wrapper to explicit cmd.exe, preserve ENOENT for a missing wrapper. Otherwise cmd.exe itself starts successfully and callers lose the existing Go-toolchain-not-found guidance required by #952.

tests/test-ttsc/src/native-plugins/source-plugin/test_spawngotool_preserves_windows_command_wrapper_arguments.ts
```diff
@@ -0,0 +1,95 @@
...
+
+/** Windows command wrappers receive literal paths and argv through cmd.exe. */
+export const test_spawngotool_preserves_windows_command_wrapper_arguments =
+  () => {
+    if (process.platform !== "win32") return;
+
+    const root = TestProject.tmpdir("ttsc-go-command-shim-");
+    const wrapperRoot = path.join(
+      root,
+      "%TTSC_GO_EXPANDS% literal %% & (parentheses) ^ caret !TTSC_GO_DELAYED!",
```
samchon, line 37: [P2] Accepted: these ! inputs cover the data surface, but delayed expansion is off by default, so this integration path can pass even if /v:off is removed. Add a production-used command-argument seam and assert the explicit /v:off contract.

tests/test-ttsc/src/native-plugins/source-plugin/test_spawngotool_preserves_windows_command_wrapper_arguments.ts
```diff
@@ -0,0 +1,95 @@
+import { TestProject } from "@ttsc/testing";
+
+import { spawnGoTool } from "../../../../../packages/ttsc/lib/plugin/internal/buildSourcePlugin.js";
+import { assert, fs, path } from "../../internal/source-build";
+
+/** Windows command wrappers receive literal paths and argv through cmd.exe. */
```
samchon: [P3] Accepted: expand this into the required three-part executable-test comment: one-line Verifies headline, non-obvious regression rationale, and a 2-4 step numbered scenario. Overall review applies the same correction to the other new test cases in this PR.

samchon (commented):
Individual Self-Review of commit 3cc578941.

The read-only reviewer reported three sound improvements, and the main agent accepted all three after reopening the implementation, issue #952 acceptance, Microsoft cmd semantics, and the repository test-documentation contract. The missing absolute .cmd path loses the existing ENOENT guidance when cmd.exe becomes the spawned process; the delayed-expansion regression needs an explicit test seam that fails if /v:off disappears; and the changed test needs the required Verifies/why/numbered-scenario documentation. Overall Self-Review extended the documentation finding to every new executable test case in this PR. Corrections will land in the next commit. No ordinary PR comment is used.

samchon (commented):
Commit review for e9df35459: resolves the accepted Individual and Overall Self-Review findings. Missing absolute Windows .cmd/.bat wrappers now retain native spawn ENOENT before cmd.exe becomes the child process, preserving the Go-toolchain install guidance. The production cmd argument seam fixes and asserts /v:off, the real missing-toolchain corpus case exercises a missing .cmd on Windows, and every new or extended executable test in this cycle now follows the Verifies/rationale/numbered-scenario documentation contract. Targeted Prettier and the repository Go formatter completed; no local build or test command ran. Refs #952.

tests/test-wasm/src/features/memfs/test_memfs_descriptors_enforce_their_access_mode.ts
```diff
@@ -44,6 +43,14 @@ export const test_memfs_descriptors_enforce_their_access_mode =
       { ...rejectedWrite, text: host.readFileText("/f.txt") },
       { code: "EBADF", n: 0, text: "abc" },
     );
+    TestValidator.equals(
+      "a read-only descriptor rejects ftruncate without moving its cursor",
+      {
+        code: await expectFsError((cb) => host.fs.ftruncate(readOnly, 1, cb)),
```
samchon, line 64: [P2] #950 requires a rejected read-only ftruncate to preserve mtime as well as bytes and cursor. Capture fstat mtime before and after this call and assert equality.

packages/wasm/src/createMemFS.ts
```diff
@@ -857,7 +870,11 @@ export function createMemFS(): IMemFSHost {
         callback(new MemFSError("EINVAL", "ftruncate"));
         return;
       }
-      const node = nodes.get(entry.path);
+      if (!entry.writable) {
+        callback(new MemFSError("EBADF", "ftruncate", entry.path));
+        return;
+      }
+      const node = entry.node;
```
samchon, line 877: [P2] The writable descriptor path reuses resizeFileData, but current ftruncate coverage only shrinks. Add an fd-based growth assertion that proves the new tail is zero-filled, as #950 acceptance requires.

samchon (commented):
Overall Self-Review finding round on e9df35459.

The fresh complete base-to-head review found two missing acceptance witnesses for #950. The read-only ftruncate regression proves EBADF, bytes, and cursor but not unchanged mtime. Existing path-truncate coverage proves zero-fill growth, while descriptor ftruncate coverage proves only shrink, so the issue's writable descriptor growth requirement is not directly exercised. The implementation returns before mutation and shares resizeFileData, but issue acceptance requires observable fd-boundary evidence. Both findings are accepted and will be corrected without local build or test execution.

samchon (commented):
Pushed-commit ledger for `a00e4343774afafe76f1f77cc73e3c8e5f1ce307` (`test: prove ftruncate capability consequences`).

This correction addresses the two accepted Overall Self-Review findings for #950:

- the read-only `ftruncate` witness now records the descriptor mtime before and after the rejected call, advances past the recorded clock value first, and still asserts unchanged bytes and cursor;
- the writable descriptor witness now shrinks and regrows the same file, asserting the exact zero-filled byte sequence after growth.

Targeted Prettier formatting completed for the two changed TypeScript files. Per the campaign instruction, no local build or test was run; ordinary PR CI owns those gates. The mandatory read-only Individual Self-Review for this commit is running separately.

packages/ttsc/src/plugin/internal/buildSourcePlugin.ts
```diff
@@ -1762,26 +1762,70 @@ function goToolchainNotFoundMessage(pluginName: string): string {
   );
 }
 
-function spawnGoTool(
+export function spawnGoTool(
   goBinary: string,
   args: readonly string[],
   options: SpawnSyncOptionsWithStringEncoding,
 ): SpawnSyncReturns<string> {
```
samchon, line 1770: shouldSpawnGoToolThroughShell examines only the original string. On Windows an override such as go can resolve through PATH/PATHEXT to go.cmd; this branch then direct-spawns the wrapper instead of selecting the hardened cmd.exe plan. Resolve executable identity under the effective cwd, PATH, and PATHEXT, and select wrapper execution from the resolved target.

packages/ttsc/src/plugin/internal/buildSourcePlugin.ts
```diff
@@ -1762,26 +1762,70 @@ function goToolchainNotFoundMessage(pluginName: string): string {
...
-  // quoting a space-free path is harmless.
-  return spawnSync(shell ? `"${goBinary}"` : goBinary, [...args], {
-    ...options,
-    shell,
-  });
+  if (!shouldSpawnGoToolThroughShell(goBinary)) {
+    return spawnSync(goBinary, [...args], options);
+  }
+  // Preserve the native spawn ENOENT contract before cmd.exe becomes the
+  // actual child process. The callers use that code for the install guidance.
```
samchon, line 1784: The ENOENT preflight applies only to absolute wrapper strings. Missing named (missing-go.cmd) and relative (.\missing-go.cmd) overrides become cmd.exe failures and bypass the callers toolchain-not-found guidance. Preflight all wrapper forms under the effective lookup environment, while keeping existing PATH-resolved wrappers on the shell branch.

tests/test-ttsc/src/native-plugins/corpus-misc/test_plugin_corpus_missing_go_toolchain_points_users_at_the_install_hint.ts
```diff
@@ -16,7 +16,8 @@ import {
  *
  * Source plugins require `go build`, and first-time users often do not have Go
  * in PATH. When neither `go` nor the `TTSC_GO_BINARY` override resolves to a
- * real binary, ttsc must emit a human-readable message (`Go toolchain was not
+ * real binary, including a missing Windows `.cmd` wrapper launched through
```
samchon: The test uses a missing absolute .cmd, which the new guard rejects before selecting cmd.exe; launched through cmd.exe states the opposite control flow. Update the narrative together with the named/relative/PATH-resolved regression matrix.

tests/test-wasm/src/features/memfs/test_memfs_descriptors_keep_open_node_identity.ts
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * Verifies MemFS: open descriptors retain identity across namespace changes.
+ *
+ * Path-only descriptors either disappeared after unlink or retargeted to a
+ * replacement rename destination. Each fd must retain its opened node while the
+ * namespace independently removes or replaces names.
+ *
+ * 1. Unlink an open file, then read, write, truncate, and stat through its fd.
+ * 2. Rename one open file over another open file.
+ * 3. Mutate both fds and assert their bytes remain distinct after close.
```
samchon: The distinct-byte assertion occurs before close; after close the test only proves that namespace state remains correct. Reword this numbered step to match the actual execution order.

samchon (commented):
Individual Self-Review result for e9df35459b7de05ce15448e050728d1fcf61239f (fix: preserve windows go wrapper failure contracts).

The mandatory read-only reviewer produced five candidates; main-agent adjudication accepted all five.

1. PATH/PATHEXT resolution must decide whether an extensionless command actually resolves to a .cmd/.bat wrapper before choosing the direct-spawn branch.
2. The ENOENT preservation guard must cover missing named and relative wrapper paths, not only missing absolute paths, while existing PATH-resolved wrappers must still run through cmd.exe.
3. The missing-toolchain test narrative contradicts the new preflight behavior and needs correction.
4. The access-mode test claimed mtime preservation without proving it. This is already corrected by follow-up commit a00e4343774afafe76f1f77cc73e3c8e5f1ce307.
5. The open-node identity test says byte separation is asserted after close although the byte assertion precedes close.

The functional and remaining documentation findings are left inline below and will be corrected on this PR. The reviewer made no edits, commits, pushes, GitHub posts, formatting runs, builds, or tests.

tests/test-wasm/src/features/memfs/test_memfs_descriptors_enforce_their_access_mode.ts
```diff
@@ -54,19 +77,20 @@ export const test_memfs_descriptors_enforce_their_access_mode =
 
     // Positive twins: the granted directions still work.
     TestValidator.equals(
-      "a read-only descriptor still reads",
+      "a rejected ftruncate leaves the read-only cursor unchanged",
```
samchon, line 85: The read-only cursor is still at offset zero when ftruncate is rejected, so a defect that resets the cursor to zero remains invisible. Advance the descriptor to a non-zero offset before the rejected call and assert the remaining bytes afterward.

tests/test-wasm/src/features/memfs/test_memfs_descriptors_enforce_their_access_mode.ts
```diff
@@ -14,20 +15,24 @@ const O_WRONLY = 1;
...
+ * later gap also let read-only descriptors truncate the file. These operations
+ * must fail before changing bytes, mtime, or the shared cursor.
  *
  * 1. Open the same file read-only, write-only, and read-write.
- * 2. Drive a write through the read-only fd and a read through the write-only fd,
- *    then exercise both directions on the read-write fd.
- * 3. Assert the forbidden operations return `EBADF` with zero bytes and no byte
- *    change, and that the permitted ones still work.
+ * 2. Reject read-only write and truncate plus write-only read operations.
+ * 3. Assert `EBADF`, unchanged bytes and cursor, and permitted mutations.
```
samchon: The numbered summary now needs to name the added mtime invariant as well as bytes and cursor.

tests/test-wasm/src/features/memfs/test_memfs_ftruncate_resizes_via_descriptor.ts
```diff
@@ -12,10 +12,9 @@ import { callMutation, expectFsError, openFd } from "../../internal/callbackFs";
  * file: the reserved stdout/stderr fds and unknown fds. The pre-fix version was
  * a no-op, so a descriptor-based truncate left the file unchanged.
  *
- * 1. Seed `/t.txt`="abcdef" and open it, then ftruncate the descriptor to 3.
- * 2. Read the file back and attempt ftruncate on an unknown fd, the stdout fd, and
- *    a negative length.
- * 3. Assert the file shrank to "abc" and each invalid call carries its code.
+ * 1. Seed `/t.txt`="abcdef", then shrink its descriptor to 3 and grow it to 5.
+ * 2. Read the bytes and reject unknown, stdout, and negative-length calls.
+ * 3. Assert shrink, zero-filled growth, and each invalid descriptor code.
```
samchon: The negative-length case is an invalid request rather than an invalid descriptor. Use each rejection code (or equivalent) so the summary matches all asserted cases.

samchon (commented):
Individual Self-Review result for a00e4343774afafe76f1f77cc73e3c8e5f1ce307 (test: prove ftruncate capability consequences).

The mandatory read-only reviewer found two candidates; main-agent adjudication accepted both.

- The mtime witness is robust against same-millisecond updates and fstat errors are propagated correctly.
- The shrink-then-grow byte assertion correctly catches both stale-tail restoration and non-zero extension.
- However, the read-only cursor starts at zero, so a defective ftruncate that resets it to zero would still pass; the witness must establish a non-zero cursor before rejection.
- Two numbered summaries need to name the exact mtime/rejection surfaces they now assert.

The accepted observations are inline below and will be corrected on this PR. The reviewer made no edits, commits, pushes, GitHub posts, formatting runs, builds, or tests.

samchon (commented):
Pushed-commit ledger for `db5bfa491` (`fix: resolve windows go wrappers before spawning`).

This correction adjudicates and fixes the accepted Individual Self-Review findings from `e9df35459` and `a00e43437`:

- Windows Go commands are resolved under the spawn's effective `cwd`, case-insensitive `PATH`/`PATHEXT`, and current-directory-first command lookup before choosing the wrapper branch;
- actual `.cmd`/`.bat` targets use the explicit hardened `cmd.exe /d /v:off /s /c` plan, while unresolved absolute, named, and relative wrapper forms fall back to native spawn so the callers retain ENOENT and the install hint;
- regression coverage exercises absolute, extensionless PATH-resolved, and relative wrappers, plus missing named and relative wrapper forms;
- the missing-toolchain and open-node identity narratives now match the real control flow;
- the read-only MemFS descriptor advances to a non-zero cursor before rejected `ftruncate`, and the numbered summaries match the mtime/cursor/rejection assertions.

The first formatter invocation could not resolve the worktree's import-sorting plugin because this isolated checkout has no `node_modules`; targeted Prettier was then run successfully from the existing primary checkout configuration over all six changed files. `git diff --check` passed. Per campaign instruction, no local build or test was run; ordinary PR CI owns those gates. The mandatory read-only Individual Self-Review for this commit is running separately.

samchon (commented):
Overall Self-Review finding round on `db5bfa49146cc998cf86c63c9a23bd08dc2758a8`.

This was a fresh complete pass over all 18 base-to-head changed files and hunks, the affected source/descriptor/namespace/package-resolution consumers, documentation and generated-output surfaces, Windows/POSIX behavior, and every acceptance item in #950?#955. Three sound improvements survived verification:

1. Windows wrapper discovery currently interleaves all PATHEXT extensions per directory. That can let a current/early `go.cmd` newly shadow the `.com`/`.exe` target that Node/libuv direct spawn would have selected later in PATH. Preserve native direct-spawn selection first, then use `.cmd`/`.bat` only as the fallback when no native target exists.
2. Trimming PATH and PATHEXT tokens changes Windows command lookup semantics for entries whose whitespace is literal. Quoting may be removed, but whitespace must remain data.
3. #950 explicitly requires path-based `stat` to report ENOENT after unlink; the current test proves namespace absence only through the host convenience `exists` helper and does not pin the low-level callback error.

The paths, exports-pattern, npm-spec classification, low-level mkdir, retained-node behavior, access-mode enforcement, mtime/cursor, and zero-fill surfaces otherwise matched their issue contracts in this round. No local build or test was run. These findings will be corrected, so this round is not the merge-gating clean Overall Self-Review and a new whole-diff round is required afterward.

packages/ttsc/src/plugin/internal/buildSourcePlugin.ts
```diff
@@ -2387,38 +2447,79 @@ function resolveExecutableIdentityPath(
...
+  candidate: string,
+  env: NodeJS.ProcessEnv,
+): string | null {
+  if (isExecutableFile(candidate)) return candidate;
+  if (process.platform !== "win32" || path.extname(candidate) !== "") {
+    return null;
+  }
+  // Probe every PATHEXT extension, not just `.exe`, so a compiler backed by a
+  // `.cmd`/`.bat` wrapper is both launched correctly and hashed into the cache
+  // key. Otherwise the wrapper reads as missing and changes do not invalidate
```
samchon, line 2602: This per-directory PATHEXT loop can select an early go.cmd before a later go.exe/go.com that Node libuv direct spawn previously selected. Resolve the native .com/.exe search across cwd and PATH first; only if that fails should the same lookup fall back to eligible .cmd/.bat wrappers.

packages/ttsc/src/plugin/internal/buildSourcePlugin.ts
```diff
@@ -2387,38 +2447,79 @@ function resolveExecutableIdentityPath(
...
+  }
+}
+
+function hasPathSeparator(location: string): boolean {
+  return (
+    location.includes(path.sep) ||
+    (process.platform === "win32" && location.includes("/"))
+  );
+}
+
```
samchon, line 2626: Windows command lookup does not trim leading/trailing whitespace from PATH or PATHEXT entries. Remove surrounding matching quotes without trimming literal whitespace; the PATHEXT parser below must preserve it for the same reason.

tests/test-wasm/src/features/memfs/test_memfs_descriptors_keep_open_node_identity.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    host.writeFile("/unlinked.txt", "KEEP");
+    const unlinkedFd = await openFd(host.fs, "/unlinked.txt", O_RDWR);
+    await callMutation((cb) => host.fs.unlink("/unlinked.txt", cb));
+    await writeFdText(host.fs, unlinkedFd, "X", 0);
+    await callMutation((cb) => host.fs.ftruncate(unlinkedFd, 2, cb));
+    TestValidator.equals(
+      "unlink removes only the name",
+      {
+        pathExists: host.exists("/unlinked.txt"),
```
samchon, line 53: Issue #950 requires the low-level path-based stat callback to return ENOENT after unlink. host.exists proves namespace absence but not the fs.stat error contract; add the explicit stat rejection beside the retained-fd read/write/ftruncate/fstat witness.

packages/ttsc/src/plugin/internal/buildSourcePlugin.ts
```diff
@@ -2387,38 +2447,79 @@ function resolveExecutableIdentityPath(
   binary: string,
   env: NodeJS.ProcessEnv = process.env,
```
samchon, line 2490: Compiler identity is currently resolved from process.cwd(), while later Go operations use explicit plugin/scratch cwd values. Resolve one compiler target under the effective build env/context and use that same target for identity, version, env, mod, and build so the cache key cannot describe a different toolchain.

packages/ttsc/src/plugin/internal/buildSourcePlugin.ts
```diff
@@ -2387,38 +2447,79 @@ function resolveExecutableIdentityPath(
...
-      }
-    }
+    return findExecutableCandidate(binary, env);
+  }
+  if (hasPathSeparator(binary)) {
+    return findExecutableCandidate(path.resolve(cwd, binary), env);
+  }
+
+  const directories = readPathEnvironment(env)
+    .split(path.delimiter)
```
samchon: Unconditional cwd insertion bypasses NoDefaultCurrentDirectoryInExePath. The effective Windows search policy must decide whether cwd participates before resolving a wrapper to an absolute path.

samchon (commented):
Individual Self-Review result for `db5bfa49146cc998cf86c63c9a23bd08dc2758a8` (`fix: resolve windows go wrappers before spawning`).

The mandatory read-only reviewer produced five candidates; main-agent adjudication accepted all five.

1. Compiler cache identity can be resolved/versioned under ambient cwd/env while `go env`, `go mod edit`, and `go build` use the effective build env and their explicit cwd, allowing identity and actual compiler selection to diverge.
2. Unconditional current-directory search bypasses `NoDefaultCurrentDirectoryInExePath`.
3. An omitted child PATH must fall back to the parent PATH the same way Node/libuv does on Windows.
4. Splitting/trim-normalizing PATH/PATHEXT changes quoted-semicolon and literal-whitespace lookup semantics.
5. Case-insensitive duplicate environment keys must use Node's lexicographically first winner, not insertion order.

Items 2, 4, and the native-executable precedence consequence were independently confirmed by the concurrent fresh Overall Self-Review. The correction in progress will settle one absolute tool target per build context, preserve native `.com`/`.exe` selection before wrapper fallback, honor the cwd-search policy and parent-PATH fallback, parse quoted Windows search entries without trimming, and choose environment-key winners deterministically. Regression witnesses will cover the competing native/wrapper, security-policy, omitted-PATH, quoted-semicolon, whitespace, and key-casing cases.

The reviewer found the MemFS cursor/mtime/zero-fill correction clean and made no edits, commits, pushes, GitHub posts, formatting runs, builds, or tests.

packages/ttsc/src/plugin/internal/buildSourcePlugin.ts
samchon, line 2626: Windows search entries preserve literal whitespace and may contain a semicolon inside matching quotes. Parse the quoted list statefully and remove only matching surrounding quotes; do not trim.

packages/ttsc/src/plugin/internal/buildSourcePlugin.ts
```diff
@@ -2427,7 +2528,22 @@ function windowsExecutableExtensions(
 }
```
samchon, line 2671: When the supplied child env omits PATH, Node/libuv falls back to the parent process PATH on Windows. Mirror that fallback before deciding a named wrapper is unresolved.

packages/ttsc/src/plugin/internal/buildSourcePlugin.ts
```diff
@@ -2427,7 +2528,22 @@ function windowsExecutableExtensions(
...
 
 function readPathEnvironment(env: NodeJS.ProcessEnv = process.env): string {
-  return env.PATH ?? env.Path ?? "";
+  return process.platform === "win32"
+    ? (readWindowsEnvironmentValue(env, "PATH") ?? "")
+    : (env.PATH ?? "");
+}
+
+function readWindowsEnvironmentValue(
+  env: NodeJS.ProcessEnv,
```
samchon, line 2682: For case-insensitive duplicate env keys, Node sorts keys lexicographically and passes the first winner. Match that selection rather than Object.keys insertion order so resolver PATH/PATHEXT/COMSPEC agrees with the spawned child.

samchon (commented):
## Pushed commit review: `de5532a36`

Reviewed the pushed correction commit against the accepted Overall and Individual Self-Review findings.

- Pins the Windows Go target once from the source-build directory and reuses it for compiler identity, `go version`, `go env`, `go mod`, and `go build`.
- Preserves native `.com`/`.exe` lookup across the full search path before `.cmd`/`.bat` fallback.
- Matches Windows PATH edge semantics for quoted semicolons, literal whitespace, omitted PATH inheritance, duplicate case-insensitive keys, and `NoDefaultCurrentDirectoryInExePath`.
- Extends Windows regressions for native precedence, lookup policy, cache identity, and ENOENT.
- Adds the missing MemFS post-unlink path `stat` ENOENT assertion required by #950.

Formatting and `git diff --check` passed. Per campaign policy, no local build or test command was run; PR CI owns execution validation. `typia.yml` and `nestia.yml` are excluded from the campaign gate.

samchon (commented):
## Individual Self-Review: `de5532a36`

Adjudication: four findings accepted.

The Windows extension and quote findings were verified against libuv's current `search_path` implementation: an already-extended name is tried exact and then with appended `.com`/`.exe`, and PATH entries may begin with either double or single quotes. The POSIX empty-PATH component and whitespace-test findings are also valid.

The follow-up correction will be handled by the main agent without another Individual Self-Review subagent, per the updated campaign instruction. Local build/tests remain delegated to PR CI.

Primary source: https://raw.githubusercontent.com/libuv/libuv/v1.x/src/win/process.c

packages/ttsc/src/plugin/internal/buildSourcePlugin.ts
```diff
@@ -2386,48 +2488,212 @@ function computeGoCompilerIdentity(goBinary: string, resolved: string): string {
...
+  if (isWindowsCommandWrapper(binary)) {
+    return {
+      location: candidates.find(isExecutableFile) ?? null,
+      wrapper: true,
+    };
+  }
+
+  const hasExtension = windowsFileNameHasExtension(binary);
+  for (const candidate of candidates) {
+    if (hasExtension) {
```
samchon: P1 accepted: libuv tries the exact extended name and then appends `.com` and `.exe`. Returning after the exact probe leaves `go.v1.exe` unresolved, so build-wide pinning and compiler identity can diverge from the processes that libuv later starts.

packages/ttsc/src/plugin/internal/buildSourcePlugin.ts
```diff
@@ -2386,48 +2488,212 @@ function computeGoCompilerIdentity(goBinary: string, resolved: string): string {
...
+  if (hasPathQualifier(binary)) return [path.resolve(cwd, binary)];
+
+  const directories =
+    process.platform === "win32"
+      ? splitWindowsSearchPath(readPathEnvironment(env))
+          .map(unquoteWindowsSearchEntry)
+          .filter((dir) => dir.length > 0)
+      : readPathEnvironment(env)
+          .split(path.delimiter)
+          .filter((dir) => dir.length > 0);
```
samchon: P1 accepted: on POSIX, an empty PATH component denotes the working directory. Filtering it out can make the identity resolver hash `missing` or a different executable while `spawnSync` starts the cwd tool.

packages/ttsc/src/plugin/internal/buildSourcePlugin.ts
```diff
@@ -2386,48 +2488,212 @@ function computeGoCompilerIdentity(goBinary: string, resolved: string): string {
...
+
+function splitWindowsSearchPath(value: string): string[] {
+  const entries: string[] = [];
+  let start = 0;
+  let quote = "";
+  for (let index = 0; index < value.length; index++) {
+    const character = value[index]!;
+    if (quote !== "") {
+      if (character === quote) quote = "";
+    } else if (character === '"' && index === start) {
```
samchon: P2 accepted: libuv recognizes both `"` and single quote when either begins a Windows PATH entry, and strips either quote at the boundary. The parser and unquoter must preserve that behavior.

tests/test-ttsc/src/native-plugins/source-plugin/test_spawngotool_preserves_windows_command_wrapper_arguments.ts
```diff
@@ -0,0 +1,360 @@
...
+
+    const whitespaceRoot = path.join(root, " leading-path-entry");
+    fs.mkdirSync(whitespaceRoot, { recursive: true });
+    fs.copyFileSync(script, path.join(whitespaceRoot, "capture.cjs"));
+    fs.copyFileSync(wrapper, path.join(whitespaceRoot, "go.cmd"));
+    const whitespaceArgs = ["version", "literal PATH whitespace"];
+    const whitespaceResult = spawnGoTool("go", whitespaceArgs, {
+      cwd: root,
+      encoding: "utf8",
+      env: { ...env, PATH: whitespaceRoot, PATHEXT: ".CMD" },
```
samchon: P2 accepted: this absolute PATH string begins with the temp-root drive/path, not whitespace. `.trim()` would therefore leave it unchanged. Use a cwd-relative PATH entry whose first character is the literal space.

samchon (commented):
Commit ledger — `53f5a78d3`

Self-Review scope: the follow-up correction for issue #952.

- Matched libuv executable search boundaries for extended Windows names, single-quoted PATH entries, and POSIX empty PATH components.
- Preserved the native-before-wrapper selection contract and added focused regression coverage for each corrected boundary.
- Targeted formatting and `git diff --check` are clean.
- Per campaign policy, no local build, test, or typecheck was run; execution validation belongs to PR CI. `typia` and `nestia` workflows are excluded from the merge gate.

No additional finding in this commit-scoped Self-Review.

packages/ttsc/src/plugin/internal/buildSourcePlugin.ts
```diff
@@ -2386,48 +2488,209 @@ function computeGoCompilerIdentity(goBinary: string, resolved: string): string {
...
+  binary: string,
+  env: NodeJS.ProcessEnv,
+  cwd: string,
+): string[] {
+  if (path.isAbsolute(binary)) return [binary];
+  if (hasPathQualifier(binary)) return [path.resolve(cwd, binary)];
+
+  const directories =
+    process.platform === "win32"
+      ? splitWindowsSearchPath(readPathEnvironment(env))
```
samchon, line 2575: [P1] Preserve quoted-empty PATH entries. libuv skips a truly zero-length PATH slice before unquoting, but `""`/`'` survives that check and becomes a zero-length directory, which means the spawn cwd. Filtering after unquoting drops that cwd candidate, so cache identity can select a later wrapper while the native spawn selects cwd. Filter raw entries before `unquoteWindowsSearchEntry`, and cover the quoted-empty case while `NoDefaultCurrentDirectoryInExePath` disables the separate implicit cwd probe.

packages/ttsc/src/plugin/internal/buildSourcePlugin.ts
```diff
@@ -2386,48 +2488,209 @@ function computeGoCompilerIdentity(goBinary: string, resolved: string): string {
...
+          .map(unquoteWindowsSearchEntry)
+          .filter((dir) => dir.length > 0)
+      : readPathEnvironment(env).split(path.delimiter);
+  const noDefaultCurrentDirectory =
+    process.platform === "win32"
```
samchon, line 2582: [P1] Do not read `NoDefaultCurrentDirectoryInExePath` from the child env. libuv calls `NeedCurrentDirectoryForExePathW` in the parent during path search, so only the parent process environment controls the implicit cwd probe. A per-spawn `options.env` override here can make this resolver skip or include cwd differently from the ensuing native spawn, breaking the compiler pin/cache identity contract.

samchon (commented):
Overall Self-Review finding round on the full 19-file base-to-head diff. Two Windows executable-search mismatches remain; both can make compiler identity selection diverge from the compiler libuv actually launches. No local build, test, or typecheck was run.

samchon (commented):
Commit ledger — `171f46e2f`

Self-Review scope: follow-up corrections for the Overall Self-Review findings on issue #952.

- Preserved quoted-empty Windows PATH entries as explicit cwd probes, matching libuv's pre-unquote empty-slice rule.
- Made `NoDefaultCurrentDirectoryInExePath` follow the parent process environment used by libuv instead of the child spawn environment.
- Added focused Windows regression cases for both selection boundaries.
- Targeted formatting and `git diff --check` are clean.
- Per campaign policy, no local build, test, or typecheck was run; PR CI owns execution validation. `typia` and `nestia` workflows remain outside the merge gate.

No additional finding in this commit-scoped Self-Review. No individual subagent was opened for this corrective commit, per the current campaign rule.

samchon (commented):
Overall Self-Review — clean saturation round on `171f46e2f`

I restarted the review from the base after the prior finding round and re-inspected the complete 19-file, 1,660-line base-to-head change surface rather than only the corrective commits.

Reviewed consequence surfaces:

- #950: MemFS descriptor capability, open-node identity across unlink/replacement rename, descriptor truncate, and low-level mkdir atomicity.
- #951: paths source identity under host case sensitivity, including exact, extension, and index lookup plus emitted-path consequences.
- #952: Windows Go wrapper argv fidelity, native-before-wrapper selection, cwd/PATH/PATHEXT/libuv boundaries, missing-tool ENOENT, compiler pinning, identity memoization, and POSIX empty PATH semantics.
- #953: npm registry-range classification across semver, tags, aliases, local/archive/URL/git/hosted source families, dependency queueing, and installer requests.
- #954: sandbox `exports` exact/pattern precedence, suffix patterns, repeated substitutions, null boundaries, and declared-exports encapsulation.
- #955: private `ttsc` export-condition pattern selection, substitution, conditional resolution, and descriptor loading.

Fresh call-site and state-transition checks found no further sound improvement or meaningful defect. `git diff --check` is clean and the worktree is clean.

Per campaign policy, no local build, test, or typecheck was run. Execution validation is delegated exclusively to PR CI; `typia` and `nestia` workflows are excluded from the merge gate.

PR #968, opened 2026-07-23
fix(unplugin): keep missing resolution candidates out of Vite's added-import graph

samchon:
This pull request owns the complete accepted cycle for the solo issue campaign covering #965. The claim commit is implementation-free; verification is pending.

## Intent

Keep missing resolution candidates out of Vite's added-import graph during `vite serve`. The transform hook's `addWatchFile()` currently forwards every derived watch input — including `graph.candidates` paths that do not exist by design — into Vite's `_addedImports`, which `vite:import-analysis` resolves as real imports and turns into a 500 on the first request of an otherwise healthy module.

## Scope

- Cycle issues: #965 (single-issue cycle; issue pre-defined by the user, discovery phase not in scope).
- In Vite serve, route watch inputs that do not currently exist to a filesystem watch that invalidates and retransforms the registered importers when the path is created; keep the ordinary `addWatchFile` contract for existing inputs, Vite build, and every other bundler.
- Regression coverage with a real Vite `createServer()` over a linked-workspace fixture whose envelope carries missing higher-priority candidates under `node_modules` paths.

## Verification

Pending. Local narrow-then-broad evidence and ordinary CI will be recorded before merge readiness.


samchon (commented):
## Individual Self-Review — commit 0b3fd7f38

**What landed:** the #965 fix itself. In Vite serve, derived watch inputs that do not exist on disk no longer enter the transform context's `addWatchFile()` (whose `_addedImports` `vite:import-analysis` resolves as real imports, turning the importer's first request into a 500). They are routed to a stat-poll registry (`core/viteServe.ts`) attached via `configureServer`; on creation the registered importers are invalidated across the server's module graphs and one full-reload is sent. Five regression tests over a junction-linked pnpm-shaped fixture (serve first-request, superseding-candidate retransform, unrelated-creation negative twin, existing-input added-import preservation, production build) plus the `setup/unplugin.mdx` paragraph.

**Findings and adjudication (reviewer advises; dispositions are the main agent's):**

1. **Major — restart detaches the watch.** Vite's `restartServer` configures the replacement server *before* closing the old one, so the old container's `buildEnd` → `dispose()` (`server = undefined` in this commit) ran after the new `configureServer` attach and re-armed the 500 for programmatic inline-plugin servers. Confirmed against `vite@7.3.3` source and reproduced by a failing test against this commit. **Fixed in 64c893e41** (dispose keeps the server reference; a same-instance serve→build reuse is excluded via `config.command`), with a restart regression test.
2. **Minor — poll-arming race.** A file created between the adapter's `existsSync` check and `fs.watchFile` registration counted as "unchanged" and never fired the listener. **Fixed in 979fa70fc** with a deferred recheck routed through the same listener.
3. **Minor — pollers accumulate until creation or close.** Accepted as the documented design: entries are bounded by the project's distinct missing candidates, release on creation, and are cleared at server close; one stat per path per 500ms is the deliberate cost of covering paths chokidar ignores.
4. **Nit — module-graph fast path dead on case-insensitive filesystems** (identity-keyed importer vs original-case graph keys). **Fixed in 979fa70fc**: importers keep their original absolute spelling; the identity scan remains the fallback.
5. **Coverage note:** two of the five tests fail against the pre-fix code (first-request 500, retransform timeout); the other three are deliberate negative/boundary guards. The untested restart lifecycle was covered together with finding 1 in 64c893e41.

**Clean areas:** public API surface (no new package exports, no `vite` type dependency in published declarations), #832 freshness contract (missing candidates are rerouted, never dropped; recompile guaranteed by the external-input `missing` hash marker), docs accuracy, Windows/POSIX path-identity coherence.

packages/unplugin/src/core/viteServe.ts
```diff
@@ -0,0 +1,271 @@
...
+
+  return {
+    attach(next) {
+      server = next;
+    },
+    dispose() {
+      for (const [identity, entry] of entries) {
+        unwatch(identity, entry);
+      }
+      // The server reference deliberately survives: `vite.restartServer`
```
samchon, line 118: Individual Self-Review of 0b3fd7f38, finding 1 (major): as first committed, `dispose()` also set `server = undefined`. Vite's `restartServer` configures the replacement server before closing the old one, so the old container's `buildEnd` cleared the freshly attached replacement and revived the 500 after every restart of a programmatic inline-plugin server. Resolved in 64c893e41 — this surviving comment block documents the corrected contract — and pinned by `test_vite_serve_survives_missing_candidates_after_a_server_restart`.

packages/unplugin/src/core/viteServe.ts
```diff
@@ -0,0 +1,271 @@
...
+        { interval: MISSING_INPUT_POLL_INTERVAL },
+        entry.listener,
+      );
+      // A poller must never keep the dev-server process alive on its own.
+      watcher.unref?.();
+      // `fs.watchFile` snapshots the path's stats at registration and fires
+      // only on a subsequent change, so a file created between the adapter's
+      // existence check and this registration would count as "unchanged" and
+      // never fire. One deferred recheck closes that window; routing through
+      // the listener keeps a single finalization path.
```
samchon, line 166: Individual Self-Review of 0b3fd7f38, finding 2 (minor): `fs.watchFile` snapshots the stats at registration, so a file created between the adapter's existence check and this registration would never fire the listener and the importer would keep serving the stale resolution winner. Resolved in 979fa70fc by this deferred recheck, which routes through the same listener so there is a single finalization path.

packages/unplugin/src/core/viteServe.ts
```diff
@@ -0,0 +1,271 @@
...
+/**
+ * Look up the module nodes registered for one importer spelling: the fast
+ * slash-normalized `getModulesByFile` lookup first, then an identity scan of
+ * `fileToModulesMap` for spellings that differ only by separator or case.
+ */
+function selectModulesByFile(
+  graph: ViteModuleGraphLike,
+  importer: string,
+): ViteModuleNodeLike[] {
+  const direct = graph.getModulesByFile?.(importer.replace(/\\/g, "/"));
```
samchon, line 235: Individual Self-Review of 0b3fd7f38, finding 4 (nit): the fast path was queried with the lowercased identity key, which never matches the module graph's original-case file keys on Windows/macOS, so every invalidation fell through to the O(n) scan. Resolved in 979fa70fc — importers now retain their original absolute spelling and only the fallback comparison uses `pathIdentityKey`.

packages/unplugin/src/core/viteServe.ts
```diff
@@ -0,0 +1,271 @@
...
+ * hash), so the retransform recompiles instead of replaying.
+ */
+export interface ViteServeMissingInputWatch {
+  /** Adopt the dev server whose module graphs creation events invalidate. */
+  attach(server: ViteDevServerLike): void;
+  /** Stop every poll; safe to call repeatedly. */
+  dispose(): void;
+  /**
+   * Report whether a dev server has ever been attached. This is not a liveness
+   * predicate — the reference intentionally survives the server's close (see
```
samchon, line 84: Individual Self-Review of 64c893e41, finding 1 (minor): after the correction, `serving()` returns true for a server that has already closed, but the doc still called it "Vite serve is running" — a future caller could treat it as a liveness predicate and bypass the `config.command` gate. Resolved in 979fa70fc with this contract note.

tests/test-unplugin/src/features/adapters/test_vite_serve_survives_missing_candidates_after_a_server_restart.ts
```diff
@@ -0,0 +1,39 @@
...
+ *    candidate, so the scenario cannot pass vacuously.
+ * 2. Serve it and request the entry module once.
+ * 3. Restart the resident server (same plugin instance, new plugin container).
+ * 4. Request the entry module again and assert it still transforms instead of
+ *    failing on the absent candidate.
+ */
+export const test_vite_serve_survives_missing_candidates_after_a_server_restart =
+  async () => {
+    const fixture = createLinkedWorkspaceFixture();
+    await assertFixtureDerivesMissingCandidate(fixture);
```
samchon, line 30: Individual Self-Review of 64c893e41, finding 2 (minor): without this guard the restart scenario would pass vacuously if the compiler stopped emitting the missing `node_modules` candidate for the fixture, silently ceasing to protect the restart path. Resolved in 979fa70fc by carrying the same guard as the sibling serve scenarios.

samchon (commented):
## Individual Self-Review — commit 64c893e41

**What landed:** the restart-lifecycle correction over the serve missing-input watch. `dispose()` now only stops the stat polls and deliberately keeps the attached server reference, because `vite.restartServer` configures the replacement server before closing the old one (new `configureServer` attach → old `buildEnd` dispose); forgetting the server at that point re-armed the #965 500 after every restart of a programmatic inline-plugin server. A plugin instance reused across a serve and a later build is excluded through the resolved config's `command` instead, so build lanes never route watch inputs into the serve-time poll. A restart regression test drives a real `server.restart()` on a middleware-mode server (Vite reuses inline plugin instances across restarts, so the same closure is exercised) and fails against the parent commit with the original `Failed to resolve import` 500.

**Findings and adjudication (reviewer advises; dispositions are the main agent's):**

1. **Minor — stale `serving()` doc** described the retained reference as "Vite serve is running", inviting a future caller to treat it as a liveness predicate and skip the command gate. **Fixed in 979fa70fc** — the contract now states it is not a liveness predicate and points at the gate.
2. **Minor — the restart test lacked the derives-missing-candidate vacuity guard** its siblings carry, so an upstream change in candidate emission could have made it pass without exercising the poll path. **Fixed in 979fa70fc.**
3. **Nit — the closed server stays reachable from the plugin closure** until the instance is released. Accepted as the documented trade-off: bounded to one server, and the alternative (forgetting it) is precisely the defect this commit removes.

**Verified clean:** the full lifecycle matrix (fresh serve; serve→restart; serve→final close; same-instance serve→build via the command flip; build-only; non-Vite bundlers with `viteCommand` undefined), poll-fire-after-close safety (listeners read the live closure server and every foreign-shape interaction is guarded), and `vite build --watch` staying on Rollup's native `addWatchFile` contract.

samchon (commented):
## Individual Self-Review — commit 979fa70fc

**What landed:** the adjudicated corrections from the first two Individual Self-Reviews. A deferred recheck (one poll interval after registration, routed through the same listener, `statSync` guarded) closes the race where a file created between the adapter's existence check and the `fs.watchFile` registration would count as "unchanged" and never fire; importers keep their original absolute spelling so the module graph's exact-key lookup can hit on case-insensitive filesystems while the identity scan stays as fallback; `serving()`'s contract states it is not a liveness predicate; and the restart regression test carries the same derives-missing-candidate vacuity guard as its siblings.

**Findings and adjudication (reviewer advises; dispositions are the main agent's):** a complete pass found no blocker, major, or minor defect. Three nits, all accepted without code change:

1. **Nit — a directory appearing at a candidate path finalizes once** (spurious invalidate + one reload). Self-limiting: the path then exists, so the adapter's existence gate routes it to the ordinary `addWatchFile` and the poll never re-arms. Shared with the pre-existing listener semantics, not introduced by this commit.
2. **Nit — two case-spellings of one importer can double-invalidate.** `invalidateModule` is idempotent and the call is guarded; redundant work only, and it requires Vite to hand the same module id under two casings.
3. **Nit — a pending recheck timer lingers up to 500ms after unwatch/dispose.** It is `unref`'d and its `entries.get(identity) !== entry` guard makes it a no-op against a removed entry.

**Verified clean:** single-finalization between the recheck macrotask and the poll (whichever runs first unwatches and deletes the entry, turning the other into a no-op); the listener guard cannot misfire (an existing epoch-0-mtime file still finalizes via the `existsSync` clause; a missing path never does); the doc-comment renumbering in the restart test matches the code; and no comment in the diff contradicts the behavior.

---

### Addendum — commit 197cbf5f1 (test infrastructure)

The serve fixtures bypass `TestUnpluginProject.createProject`, so each fixture resolved the default project-local cache and rebuilt the native compiler host into its own `node_modules/.cache`. The suite's `ensureSharedCacheDir` is now exported and called from the fixture builder, giving the process one shared host binary exactly like every other transform scenario. Behavior-neutral for the product; verified by running all six scenarios without a preset `TTSC_CACHE_DIR` (one cold host build, then warm reuse).

samchon (commented):
## Individual Self-Review — commit 197cbf5f1

**What landed:** test infrastructure only. `TestUnpluginProject.ensureSharedCacheDir` becomes exported and the serve fixture builder calls it, so the Vite serve scenarios share one native compiler host binary through `TTSC_CACHE_DIR` instead of resolving the default project-local cache and rebuilding the host into every fixture's own `node_modules/.cache/ttsc`.

**Findings and adjudication (reviewer advises; dispositions are the main agent's):** a complete pass found no material defect. Two informational notes, both accepted without change:

1. **Info** — the helper's doc abbreviates the default cache path as `node_modules/.cache` where the precise default is `node_modules/.cache/ttsc`; cosmetic wording.
2. **Info** — in the aggregated suite run, `transform-project-cache.ts` already seeds `TTSC_CACHE_DIR` at module load, making this call a guarded no-op there; its value is isolated single-file runs that never import that module. Redundancy is intentional, not a bug.

**Verified clean:** the export is a pure visibility change consistent with the namespace's existing helper surface; the `??=`-style early return honors a parent runner's pre-set `TTSC_CACHE_DIR`; the only other writer in the package cooperates (whichever runs first wins, the other no-ops); the tests that delete or override `TTSC_CACHE_DIR` live in other packages/processes, so no cross-contamination.

PR #980, opened 2026-07-23
fix: harden transform caches and playground package boundaries

### LINKED ISSUE #252: @ttsc/unplugin: per-build transform cache never hits on real projects (O(N) full recompiles) + O(N²) re-hashing

### LINKED ISSUE #969: Very slow Bun runtime startup with @ttsc/unplugin and Typia
## Environment

  - Bun: `1.3.14`
  - `ttsc`: `0.20.0`
  - `@ttsc/unplugin`: `0.20.0`
  - Typia: `13.2.0`
  - OS: macOS arm64

  ## Problem

  Using `@ttsc/unplugin/bun` as a Bun runtime plugin causes a very slow initial startup.

  The application is a regular TypeScript application using Typia-generated validators.

  Example runtime registration:

  ```ts
  import typiaPlugin from "@ttsc/unplugin/bun";
  import { plugin } from "bun";

  plugin(typiaPlugin());
  ```

  The application is started directly from a TypeScript entrypoint:

  ```bash
  bun ./src/main.ts
  ```

  On a cold process start, the first application log line may appear only after 1–5 minutes.

  The same application starts almost immediately when running a prebuilt JavaScript bundle.

  Disabling the Typia transform also starts quickly, but the application then fails because Typia requires its compile-time transformation.

  ## Expected behavior

  The Bun runtime plugin should provide one of the following:

  1. Transform only files required by the entrypoint.
  2. Reuse persistent or incremental compiler state.
  3. Provide a documented resident compiler or cache mode for development.

  A normal development startup should complete within a few seconds without disabling Typia transformations.

  ## Actual behavior

  The first runtime import appears to trigger a project-wide TypeScript transformation.

  The transformation is repeated after every process restart. The cache appears to be process-local, so it does not help when the development server is restarted.

  ## Minimal reproduction

  `preload.ts`:

  ```ts
  import typiaPlugin from "@ttsc/unplugin/bun";
  import { plugin } from "bun";

  plugin(typiaPlugin());
  ```

  `bunfig.toml`:

  ```toml
  preload = ["./preload.ts"]
  ```

  Start the application:

  ```bash
  bun ./src/main.ts
  ```

  Example Typia usage:

  ```ts
  import typia from "typia";

  export const validate = typia.createValidate<{ id: string }>();
  ```

  ## Additional observations

  - Native Bun TypeScript startup without the Typia transform is fast.
  - A prebuilt bundle starts fast.
  - The delay occurs before application bootstrap code executes.
  - A resident `TtscService` was also tested, but it still compiled the full configured project.
  - A narrow `tsconfig` did not clearly limit the resident compilation scope.

  ## Questions

  1. Is project-wide compilation expected for `@ttsc/unplugin/bun` in runtime mode?
  2. Is there a supported way to use a persistent resident compiler for Bun runtime loading?
  3. Can the runtime adapter use incremental per-file transformation instead of a full project transform?
  4. Can the project input set be restricted reliably to the entrypoint dependency graph?
  5. Is there a recommended development workflow for Bun and Typia that avoids multi-minute cold starts while preserving transformations?

  Thank you.

### LINKED ISSUE #975: fix(unplugin): track compiler inputs reached through filesystem links
## Problem

The Unplugin project cache can omit a compiler input reached through a symlink or Windows junction inside the project root.

The project-tree walk intentionally skips symbolic links. Following them would permit cycles and traversal outside the project. `isProjectWalkPath`, however, classifies paths only by lexical root and ignored directory names. A graph-reported file such as `<project>/linked/dependency.ts` is therefore considered covered by the project walk even though the walk never visits it.

That file enters neither `inputHashes` nor `externalInputHashes`. The same predicate is used by Metro's supplementary snapshot, so the input is omitted there as well.

## Evidence

The cache assembles its input universe in two parts:

- `collectProjectInputHashes` hashes regular files returned by `listProjectInputFiles`.
- `collectExternalInputHashes` hashes graph and plugin inputs for which `isProjectWalkPath` is false.

`listProjectInputFiles` accepts `Dirent.isDirectory()` and `Dirent.isFile()` entries. A symlink or junction satisfies neither branch and is skipped.

`isProjectWalkPath` checks the relative path segments but never calls `lstat` on those segments.

A Windows junction fixture with this shape reproduces the gap:

```text
project/
  tsconfig.json
  linked/ -> external/
external/
  dependency.ts
```

The observed classification was:

```json
{
  "hashKeys": ["tsconfig.json"],
  "classifiedInsideWalk": true
}
```

The compiler graph spelling `linked/dependency.ts` is therefore suppressed from the external snapshot despite being absent from the project snapshot.

## Consequence surface

- Bun, Turbopack, and other long-lived cache owners can replay a transform after a linked dependency changes.
- Metro can reuse a stale worker snapshot because its graph supplement applies the same classification.
- Standard bundler watch registration still receives the graph path, but the shared transform cache can answer before compiling again.
- File symlinks, directory symlinks, and Windows junctions have the same classification problem.
- Broken links and link cycles must fail closed without making project discovery recursive.
- Ordinary in-root files, ignored `node_modules` inputs, out-of-root files, case-insensitive identity, and generated temporary configs must retain their current behavior.

## Approach

Keep the tree walk non-following. Refine `isProjectWalkPath` so an existing relative path is considered walk-covered only when no component below the project root traverses a symbolic link or junction. A linked graph input then belongs to the external-input owner and receives an exact file hash.

The check must be platform-neutral and tolerate an absent candidate path. It must not resolve or follow the link during the recursive project walk.

- A graph input reached through an in-root file symlink is hashed externally.
- A graph input reached through an in-root directory symlink or Windows junction is hashed externally.
- Changing that target invalidates the project cache.
- Metro's snapshot supplement includes the linked input.
- Ordinary in-root files remain project-walk inputs and are not duplicated externally.
- Ignored directories and genuinely out-of-root graph inputs retain coverage.
- Broken links and cycles do not hang or escape the project root.
- Run:

```text
pnpm --filter @ttsc/unplugin build
pnpm --filter @ttsc/test-unplugin start -- --include=external
pnpm --filter @ttsc/test-metro start
pnpm --filter @ttsc/test-unplugin start
```

## Coordination

Issue #252 changes how often a generation is validated, not which inputs it contains. This issue supplies the missing linked-input invariant and must land before performance evidence is used to claim cache freshness.

### LINKED ISSUE #976: fix(unplugin): preserve Bun loader composition and build cache boundaries
## Problem

The Bun adapter claims every TypeScript-shaped `onLoad` path and returns original source for paths the shared transform excludes. That prevents a later Bun plugin from handling declarations, `node_modules` sources, or another excluded match.

The adapter also has no build-boundary reset when it is used with `Bun.build`, even though that plugin builder exposes `onStart`. Bun runtime plugin builders in exact 1.3.14 do not expose the hook, so the lifecycle must remain feature-detected.

## Evidence

`packages/unplugin/src/bun.ts` registers `sourceFilePattern` directly. Its loader then:

1. reads the matched path;
2. calls the raw Unplugin transform;
3. returns the original contents when no transform result exists.

The raw plugin's `transformInclude` is the owner that excludes declarations and `node_modules`, but the Bun adapter does not ask it before reading or returning.

Exact Bun 1.3.14 build controls used a later loader that replaces a dependency with `export default 42`:

| Plugin order | Result |
| --- | ---: |
| Replacement plugin only | 42 |
| Earlier broad loader returns `undefined`, then replacement | 42 |
| ttsc first with transforms disabled, then replacement | 1 |

The third result proves that ttsc completed the load with original contents and shadowed the later plugin.

The exact Bun build plugin builder exposed `onStart`; the exact runtime builder supplied to `Bun.plugin` did not. Calling `onStart` unconditionally in runtime setup throws.

A custom-namespace Bun build control remained composable with ttsc registered first. Namespace capture was investigated and rejected.

## Consequence surface

- A downstream loader cannot transform an excluded dependency that happens to match the TypeScript extension filter.
- ttsc performs avoidable filesystem reads for files it will not transform.
- `Bun.build` watch or repeated build sessions can retain a project cache across build boundaries, unlike the standard Unplugin adapters.
- Runtime registration must continue to work on Bun 1.3.14 without `onStart`.
- Lazy option resolution in `bun-register` must remain last-write-wins before the first load.
- Plugin-reported dependencies must remain harmless even though Bun runtime supplies no watch-registration channel.

## Approach

Build the raw plugin lazily through one accessor. Before reading a path, call its `transformInclude`; return `undefined` when the file is excluded so Bun can continue to later loaders.

Feature-detect `build.onStart`. Where present, forward the build boundary to the raw plugin's `buildStart` hook so it clears the project cache. Where absent, register only `onLoad`.

- An excluded `node_modules` TypeScript path returns `undefined` without reading the file.
- A later Bun loader handles that path.
- An included source still receives the ttsc transform and the correct `ts` or `tsx` loader.
- Bun runtime setup without `onStart` remains valid.
- Bun build setup with `onStart` clears a previously created raw cache at the next build boundary.
- Lazy explicit options in `bun-register` retain current ordering and idempotence.
- Exact Bun 1.3.14 composition is rechecked in addition to stub tests.
- Run:

```text
pnpm --filter @ttsc/unplugin build
pnpm --filter @ttsc/test-unplugin start -- --include=bun
pnpm --filter @ttsc/test-unplugin start
```

## Coordination

Issue #969 reports Bun runtime startup, while #252 owns the shared validation amplification. This issue corrects Bun-specific composition and lifecycle behavior without changing the checker-backed project transform contract.

### LINKED ISSUE #977: fix(playground): align package targets and alias self-references with CommonJS
## Problem

The Playground Execute sandbox now supports the three root `package.json#exports` forms reported in #670, but several adjacent package-target decisions still disagree with Node's CommonJS package algorithm:

- an array falls through when its first target is valid but the file is missing;
- an inactive nested condition stops an outer condition map instead of letting it continue;
- bare or escaping export targets are treated as pack keys;
- a manifest that mixes subpath and condition keys is accepted;
- a package installed under an npm alias cannot require itself by its manifest name.

Each difference changes which browser code executes or turns a valid installed dependency into an unavailable-package error.

## Evidence

Exact Node 22 controls and the current public sandbox produced:

| Case | Node | Playground sandbox |
| --- | --- | --- |
| `["./missing.cjs", "./available.cjs"]` | `MODULE_NOT_FOUND` for the selected missing target | Loads `available.cjs` |
| `["available.cjs", "./available.cjs"]` | Skips the invalid target and loads the second | Loads the second |
| `{ "require": { "development": "./dev.cjs" }, "default": "./available.cjs" }` | Loads the outer default | Reports unavailable |
| Export target `"available.cjs"` | `ERR_INVALID_PACKAGE_TARGET` | Loads the file |
| Mixed `{ ".": "./root.cjs", "require": "./other.cjs" }` | `ERR_INVALID_PACKAGE_CONFIG` | Loads a branch |
| Package mounted as `alias`, manifest name `actual`, internal `require("actual/sub")` | Resolves the same package | Searches for another mounted package and fails |

The alias path is live. `installPlaygroundDependencies` mounts an npm alias under the requested dependency name while retaining the downloaded manifest's actual `name`.

Node's package contract also requires export targets to begin with `./` and rejects traversal, `node_modules` segments, and encoded separator escapes after that prefix. The current resolver removes an optional `./` and concatenates the result with the pack package key.

## Consequence surface

The shared reducer owns root, exact subpath, wildcard subpath, nested condition, array, and `null` targets. The target validator must therefore cover every one of those shapes.

Behavior to preserve:

- the three root forms fixed by #670;
- exports encapsulation, ordered conditions, pattern precedence, and null blockers fixed by #820;
- explicit active conditions `require` and `default`, with `node` and `import` inactive in the browser CommonJS evaluator;
- `main` and index fallback for packages without `exports`;
- relative requires, scoped names, JSON modules, cycles, and failed-evaluation cache eviction;
- the stable requested-specifier error context.

This issue does not add ESM evaluation, Node builtins, package `imports`, filesystem access, or every Node resolution error code.

## Approach

Replace the flattened string-candidate reducer with a package-target result that distinguishes:

- resolved target;
- blocked target;
- unresolved conditional branch;
- invalid target.

Validate a string target before pack lookup. Arrays may skip invalid or unresolved entries, but once a valid target is selected, a missing file is a missing module rather than a reason to try another target. Conditional objects continue in manifest order when an active nested branch is unresolved.

Reject manifests that mix subpath and condition keys. For a bare request from inside a mounted package, compare the requested package name with that mount's parsed manifest `name`; matching requests resolve through the current mount so npm aliases retain self-reference.

- Invalid-first arrays fall through; valid-but-missing-first arrays do not.
- Nested inactive branches permit a later active outer condition.
- Bare, absolute, parent-traversing, `node_modules`, encoded slash, and encoded backslash targets fail.
- Root and subpath maps with mixed key families fail.
- Exact and pattern targets remain confined after wildcard replacement.
- Unscoped and scoped npm aliases resolve self-references by manifest name, while unrelated package requests do not.
- #670 and #820 regressions remain green.
- Direct controls compare the decision matrix with Node.
- Run:

```text
pnpm --filter @ttsc/playground build
pnpm --filter @ttsc/test-playground start -- --include=create_sandbox_require
pnpm --filter @ttsc/test-playground start
```

## Coordination

#670 is resolved in public releases beginning with 0.19.0. #820 is the direct manifest-boundary predecessor. This issue covers verified cases outside those issues' completed acceptance matrices and should not reopen either one.

### LINKED ISSUE #978: fix(unplugin): preserve newer cache generations during stale validation
## Problem

`transformTtsc` can delete a newer project-cache generation after an older caller finishes validating a stale generation.

The rejection and result-selection paths already evict by promise identity. The ordinary source-mismatch path still calls `cache.delete(key)` unconditionally. If another caller has replaced that key while the older promise was settling or being validated, the stale caller removes work it does not own and starts a third project compilation.

This is a correctness and work-amplification defect in the shared cache state machine. It affects every adapter that supplies a `TtscTransformCache`, including the standard Unplugin adapters, Bun, Metro, and Turbopack.

## Evidence

The mismatch branch in `packages/unplugin/src/core/transform.ts` follows this sequence:

1. Read `transformed = cache.get(key)`.
2. Await and validate that promise.
3. On mismatch, execute `cache.delete(key)`.
4. Install a fresh promise.

Unlike `awaitOrEvict` and `selectOrEvict`, step 3 does not check whether `cache.get(key) === transformed`.

A controlled deferred-promise reproduction used the public cache type:

1. Prime a valid project result to obtain the cache key and generation shape.
2. Store a deferred stale generation under that key.
3. Start a transform request whose source differs from the stale snapshot.
4. Before the stale generation resolves, replace the cache entry with a newer promise.
5. Resolve the stale promise.

The result was:

```json
{"newerSurvived":false,"newerWasReplaced":true,"size":1}
```

The final size of one hides the race. The entry is a newly started third generation, not the newer generation installed in step 4.

The existing feature `test_transformttsc_eviction_keeps_a_newer_generation_for_the_same_key` covers rejected promises. It does not exercise a resolved generation that fails source validation.

## Consequence surface

- Concurrent loaders can duplicate a full checker-backed project transform.
- A stale caller can replace a generation created from newer source.
- Long-lived Bun, Metro, and Turbopack processes retain the replacement generation until another invalidation.
- Standard adapters can encounter the same transition when concurrent transform hooks overlap inside one build.
- Error eviction and result-selection eviction must retain their existing identity behavior.

This issue does not change which files make a cache generation fresh. It only assigns cleanup ownership to the generation being discarded.

## Approach

Use the existing `evictGeneration(cache, key, generation)` invariant for a source mismatch. If a different generation now owns the key, retry against that generation instead of compiling over it. Only create a replacement when the stale generation still owns the key or the key is empty.

The cache key must continue to share one in-flight compile among concurrent callers.

- A resolved stale generation cannot delete or replace a newer promise stored under the same key.
- The waiting caller retries the newer generation and returns its selected output.
- A stale generation that still owns the key is evicted and replaced normally.
- Rejected transformations, host exceptions, compiler failures, volatile outputs, and ordinary cache hits retain coverage.
- The regression proves promise identity, not only `cache.size`.
- Run:

```text
pnpm --filter @ttsc/unplugin build
pnpm --filter @ttsc/test-unplugin start -- --include=transformttsc_eviction
pnpm --filter @ttsc/test-unplugin start
```

## Coordination

Issue #252 owns project-validation cost and issue #969 reports the resulting Bun startup symptom. This issue is a separate transition in the same cache owner and can be implemented in the same campaign pull request.

### LINKED ISSUE #979: fix(playground): authenticate, confine, and byte-bound npm archives
## Problem

The Playground's browser-side npm installer trusts downloaded archive bytes and entry paths too early.

It does not verify registry `dist.integrity` or `dist.shasum`, accepts tar entries that escape the canonical package root, and buffers compressed and decompressed archives without byte limits. `maxPackages` bounds dependency count but does not bound network or memory consumption per package.

This is the trust boundary that feeds the wasm compiler MemFS, Monaco declarations, and the Execute runtime pack.

## Evidence

Real npm install metadata for `typia@13.2.0` includes:

- a SHA-512 SRI `dist.integrity`;
- a SHA-1 `dist.shasum`;
- the tarball URL.

`INpmVersionMetadata.dist` currently retains only `tarball`. `downloadTarball` returns `response.arrayBuffer()`, and `unpackNpmTarball` immediately decompresses it.

A controlled registry response with deliberately false integrity and shasum values still installed successfully.

Tar extraction uses `stripTarRoot`, which removes the first component without validating the remaining segments. A regular entry named:

```text
package/../../outside.js
```

produced these mounted keys:

```text
node_modules/fixture/../../outside.js
fixture/../../outside.js
```

The compiler MemFS handoff later filters `..`, but the returned install result, Execute runtime pack, editor map, package file counts, and intermediate maps already contain the unconfined key.

The download and decompression paths also have no byte counters:

- `response.arrayBuffer()` buffers the complete compressed response;
- `new Response(decompressionStream).arrayBuffer()` buffers the complete expanded archive;
- tar entry sizes are trusted without a safe bounds check.

An authenticated package can still be malicious or compromised, so integrity verification does not replace resource and path confinement.

## Consequence surface

- Corrupt or substituted tarballs can be executed after a successful metadata request.
- Parent, absolute, drive-prefixed, backslash, PAX, and GNU long-path entries can escape the intended package key space.
- Highly compressed or simply oversized packages can exhaust a browser tab before the package-count cap applies.
- Optional dependencies must not silently convert an integrity or archive-format failure into a skip. Only registry absence retains optional semantics.
- Normal npm `package/` archives, Unicode PAX paths, long paths, JSON, declarations, and runtime JavaScript must continue to install.
- Aborted installs must stop download, verification, decompression, and parsing promptly.

## Approach

Treat registry metadata and archive bytes as one authenticated, bounded handoff:

1. Preserve `dist.integrity` and `dist.shasum`.
2. After download and before decompression, verify the strongest supported SRI algorithm. Use shasum only when SRI is absent.
3. Stream the response and decompression output through configurable compressed and expanded byte budgets with conservative browser-safe defaults.
4. Validate tar sizes and record bounds before slicing.
5. Require regular files to live below the canonical `package/` root after slash normalization. Reject empty, absolute, drive-prefixed, dot, parent, and other escaping segments. Apply the same rule to header, PAX, and GNU long paths.
6. Mount files only after the complete archive passes.

Errors must identify the package and version at the installer boundary without logging archive contents.

- A tarball matching SHA-512 SRI installs.
- A changed byte, malformed SRI, or mismatched strongest digest fails before decompression.
- SHA-1 shasum is used only when integrity is absent.
- A package with neither field follows an explicit compatibility policy covered by a test.
- Compressed and decompressed limits fail deterministically, including when `Content-Length` is absent or false.
- Truncated, overflowing, or out-of-bounds tar entries fail.
- Header, PAX, and GNU long paths containing absolute or parent traversal fail without mounting any file.
- Normal Unicode PAX and long-path archives retain coverage.
- Abort behavior and optional dependency semantics remain stable.
- Public documentation describes authenticity and size controls.
- Run:

```text
pnpm --filter @ttsc/playground build
pnpm --filter @ttsc/test-playground start -- --include=npm
pnpm --filter @ttsc/test-playground start
```

## Coordination

The package-target issue begins after a valid runtime pack is mounted. This issue owns the earlier registry metadata, download, decompression, tar parsing, and mount-key boundary. They can share one campaign pull request but require separate acceptance evidence.

samchon:
## Root cause

The reported Bun runtime delay combines two independent costs.

A checker-backed plugin such as Typia still needs one coherent project compile. On an empty ttsc plugin cache, ttsc also compiles the Go source plugin once. That persistent build can be moved out of application startup with `ttsc prepare`.

The minute-scale warm growth was not required compiler work. The Unplugin transform cache revalidated the complete project before returning every first module from a result that had already compiled the whole project. Exact public `ttsc@0.20.0`, `@ttsc/unplugin@0.20.0`, and `typia@13.2.0` controls grew from 4.56 seconds at 100 modules to 103.74 seconds at 800 modules. The 800-module run made 732,246 synchronous reads. The first-party Banner plugin reproduced the same class, so this is a ttsc/Unplugin defect rather than a Typia defect.

## Changes

- Treat one real bundler build, and one immutable Bun runtime setup, as one transform generation. Each module's first delivery compares its supplied source in constant time; repeated delivery and lifecycle-less hosts retain complete project and graph-input validation.
- Preserve authoritative newer generations across rejected, mismatching, and superseded cache races.
- Keep graph inputs reached through symlinks and Windows junctions in Unplugin and Metro external-input snapshots.
- Match Bun 1.3.14 loader composition and `BuildConfig.files` ownership, including runtime pass-through, build start, NUL virtual ids, in-memory files, separator and drive-letter normalization, component case, relative/absolute spelling, dot segments, and cwd changes.
- Complete the Playground CommonJS resolver's Node-compatible exports, array, target, legacy directory, npm-alias self-reference, and nearest-scope boundaries.
- Authenticate npm tarballs with the strongest supported SRI digest or legacy shasum, enforce compressed and expanded byte budgets, validate tar framing, and confine every selected path below one safe consistent archive root.
- Preserve cancellation through fetch, metadata JSON, response handoff, stream, bodyless fallback, digest, decompression, metadata caller, and archive caller boundaries while disposing unused responses and readers.
- Document the cache lifecycle, cold plugin preparation, archive trust limits, and CommonJS evaluator boundary.

## Measured acceptance

With a warm Typia source-plugin cache, the corrected Bun runtime control completed in 10.802 seconds at 400 modules and 33.892 seconds at 800 modules. This removes the retained per-module project revalidation while preserving the one required checker-backed project compile.

With exact 0.20.0, a warm Go build cache, and an empty ttsc plugin cache, `ttsc prepare` took 47.10 seconds; the subsequent one-module Bun start took 1.37 seconds without rebuilding the plugin binary.

## Verification

- complete `@ttsc/test-unplugin` suite;
- complete `@ttsc/test-metro` suite, 59 features;
- complete `@ttsc/test-playground` suite, 35 features;
- repository-wide `pnpm test:typecheck`;
- affected package builds;
- exact Bun 1.3.14 real-build path and loader controls;
- Node 22 package-resolution controls;
- live authenticated `typia@13.2.0` and `@types/node@26.1.1` archive controls;
- public Playground package checks for 0.19.0 and 0.20.1;
- all 34 changed feature documentation/export contracts;
- changed-file Prettier and `git diff --check`;
- five full Overall Self-Review rounds, ending clean; and
- one renewed discovery cycle with no additional sound candidate in the declared cache, adapter, resolver, and archive surface.

The root `pnpm test` umbrella completed project layout, generated-flag validation, current-platform builds, Go tests, and repository typecheck locally before its serial lint corpus exceeded the 20-minute command transport window. The same suites are split across the pull-request checks.

On the final head, 51 checks pass. The remaining `nestia` failure is a confirmed external baseline failure: ttsc `master` run [30012034569](https://github.com/samchon/ttsc/actions/runs/30012034569), Nestia PR [#1594](https://github.com/samchon/nestia/pull/1594) with published `ttsc@0.20.0`, and this pull request all fail at the same Nestia `openapi_v2` diagnostic. Nestia always supplies a default server description even when targeting Swagger 2.0, while `@typia/utils@13.2.0` correctly reports that the field cannot be represented. This compatibility regression is tracked as [samchon/nestia#1595](https://github.com/samchon/nestia/issues/1595); it is neither introduced nor affected by this pull request.

## Typia ownership

No Typia issue is warranted by the evidence. Banner reproduces the dominant warm defect, Typia output and integrity controls pass, and the plugin binary cache remains stable after preparation. The separate Nestia `openapi_v2` failure also originates in the document Nestia supplies at its Swagger 2.0 boundary, not in ttsc or Typia transformation. If a later isolated Typia-specific failure is demonstrated, it should be reported separately with that reproduction rather than attached to this cache defect.

## Issues

Closes #252
Closes #969
Closes #975
Closes #976
Closes #977
Closes #978
Closes #979

#670 is already fixed by #699. The first release containing that correction is v0.19.0; this pull request adds adjacent resolver and archive hardening but does not claim the original fix.


samchon (commented):
Individual Self-Review for d404d5f80: reviewed the complete fc0940159..d404d5f80 diff and the linked-input classification consumers in Unplugin and Metro. The lstat-based predicate matches the non-following project walk for directory links, file links, junctions, broken links, missing paths, and regular files. The external-input hash and Metro snapshot paths preserve the linked spelling as intended. No actionable findings; git diff --check passes.

samchon (commented):
Individual Self-Review for 55124be5b found two valid high-severity gaps, both accepted and remediated in 90a374971. First-delivery O(1) reuse is now enabled only after an explicit build-start signal; Metro, Turbopack, Bun runtime, and other lifecycle-less caches perform complete validation on every hit. Second, every awaited cache generation is identity-checked before validation or selection, so a matching waiter retries a newer authoritative generation instead of returning superseded output. Dedicated regressions cover an unrelated input changing before a lazy module's first request and concurrent matching/mismatching waiters. Focused tests and the exact Bun 1.3.14 composition control pass.

samchon (commented):
Individual Self-Review of `55124be5b..90a374971`

Two high-severity findings were reproduced and accepted.

1. The persistent-cache correction restored a complete project validation on every first Bun runtime module delivery because `Bun.plugin()` exposes no `onStart`. That reintroduced the O(N×P) path behind #969.
2. Vite serve calls `buildStart` once for the development plugin container rather than once per HMR edit. Treating it as a generation boundary allowed a previously emitted but unserved module to cross a later edit without validation.

Remediation:

- `65c5f0e16` defines one `Bun.plugin()` setup as one process-scoped module-loading session, while continuing to clear every `Bun.build()` generation through `onStart`.
- Vite serve now explicitly resets to persistent validation mode; production and real per-build hosts retain constant-time first delivery.
- Dedicated Bun runtime and real Vite serve regressions reproduce both lifecycle shapes.
- `24b064e19` removed a review-only global I/O monkey-patch and aligned the canonical website guide, source comments, and numbered test contracts with the implemented lifecycle.

Focused adapter, persistent-freshness, race, typecheck, and real Bun 1.3.14 controls pass after remediation. No evidence in this range identifies a Typia implementation defect.

samchon (commented):
Individual Self-Review of `65c5f0e16..66a0a4b4b`

Six Node-compatibility gaps were reproduced against Node v22.21.0 and accepted:

- empty and null-exhausted exports arrays incorrectly reached an outer default;
- alias self-reference crossed a nearer package scope;
- `exports: null` incorrectly suppressed legacy resolution;
- URL-style targets were looked up as raw pack keys instead of normalized pathnames;
- nested dot-prefixed target keys were treated as fatal configuration errors; and
- missing `main` targets and JSON directory indexes did not complete legacy fallback.

`d89ef2fb2` remediates the full class. It preserves the last meaningful array decision, stops at the nearest manifest, treats null exports as absent, resolves target URL pathnames with one decoding step and mount confinement, leaves inactive nested keys unresolved, and completes root/main-directory JSON fallback.

Four new feature scenarios cover the reported matrix, and the complete `create_sandbox_require` suite passes. Existing protections for valid-but-missing selected targets, mixed top-level maps, numeric conditions, wildcard specificity, encoded separators, and evaluation-cache eviction remain intact.

samchon (commented):
Individual Self-Review of `66a0a4b4b..65c5f0e16`

The cache-session implementation had no reproduced functional defect, but the review found three correctness-of-change issues:

- the canonical website guide still described Bun runtime as a persistent validating cache;
- the performance regression used a forbidden global `fs.readFileSync` monkey-patch; and
- the two new feature tests lacked the required numbered scenario summary.

`24b064e19` resolves all three. The guide and source comments now state the process-scoped Bun runtime contract and Vite serve's persistent contract. The Bun test observes the contract through generation behavior rather than replacing a global API, and both feature files carry the required three-step scenarios.

The affected Bun/Vite tests and Unplugin typecheck pass after remediation.

samchon (commented):
Individual Self-Review of `65c5f0e16..d89ef2fb2`

Two additional Node-compatibility findings were reproduced and accepted.

1. URL/path conversion failures such as encoded separators, malformed percent escapes, and a selected package-directory target were classified as skippable invalid targets. In an exports array this executed a later fallback, whereas Node selects the first target and then reports a load error.
2. Legacy CommonJS directory resolution remained asymmetric: relative directories and bare package subdirectories did not consistently load `package.json#main` or `index.json`.

`7146bd08e` remediates both findings. Invalid package-target selection is now distinct from post-selection module-specifier/path loading, so arrays cannot fall through after a target was selected. A shared legacy file-or-directory resolver now serves package roots, relative requests, and bare subpaths with consistent main and JSON-index behavior.

The expanded resolver suite passes, including exact Node 22 controls for encoded separators, `./` array targets, malformed escapes, relative JSON directories, nested manifests, and bare subdirectory manifests.

samchon (commented):
Individual Self-Review of `24b064e19..3afc77ae7`

Three functional findings and two coverage/convention findings were accepted.

- High: literal `package/` confinement rejected official DefinitelyTyped tarballs, including `@types/node`, `@types/lodash`, and `@types/react`.
- High: pending stream reads and WebCrypto digests did not reject promptly on abort.
- Medium: a Content-Length preflight rejection did not cancel the response body.
- Safe GNU long-name compatibility lacked a positive regression.
- Six feature-test comments lacked numbered scenario summaries.

`9b83bf6ae` remediates the complete set:

- extraction establishes one safe, consistent top-level root and strips it, preserving normal `package/` archives and current `@types/*` name-root archives while rejecting mixed roots and unsafe segments;
- stream reads race the abort signal and request cancellation, WebCrypto results are fenced by an abort race, and oversized declared bodies are cancelled;
- safe GNU long-name, official-root, pending-abort, cancellation, and documentation-contract cases are pinned; and
- public guidance now describes a consistent archive root rather than promising literal `package/`.

Live registry controls pass for authenticated `typia@13.2.0` and `@types/node@26.1.1`; the latter unpacks 89 selected files with its manifest and `index.d.ts` intact. The complete Playground feature suite passes after remediation.

samchon (commented):
Overall Self-Review, round 1

Reviewed the complete PR surface from `5a63e0` through the current cache, Bun, Playground resolver, archive, test, and documentation changes. This round found four actionable boundary defects.

1. Bun's `BuildConfig.files` matcher resolved relative callback paths against the callback-time cwd and described separator behavior too strictly. A real Bun 1.3.14 control failed after `process.chdir()`. Commit `477e13cab` captures the setup cwd, accepts Bun's normalized path spelling, and pins memory-only, override, relative, and cwd-change cases against the real runtime.
2. The first-delivery performance regression replaced global `fs.readFileSync`, contrary to the repository's test-isolation rule. `477e13cab` now mutates a real plugin input after the first compile and proves that the build-scoped generation is not revalidated or rerun for later first module deliveries.
3. Metro and Bun were still grouped together as persistent caches in source/test prose. `477e13cab` separates Metro/Turbopack's every-hit validation from Bun runtime's immutable setup session and aligns the package documentation.
4. A response implementation with `body === null` could stall forever in `arrayBuffer()` after cancellation. `9b58898f4` puts the fallback behind the same abort race as stream and digest work. Its Individual Self-Review then reproduced a narrower fetch-completion race where `arrayBuffer()` could start after the signal was already aborted; `f8ecec5b0` checks cancellation before invoking the fallback and asserts the post-abort call count remains zero.

Focused builds and regressions pass after each remediation. Round 1 is not a clean review because it produced findings; a fresh full-surface round follows.

samchon (commented):
Overall Self-Review, round 2

The fresh second pass was not clean. It found two further classes at public adapter boundaries, and the mandatory per-commit reviews widened both before acceptance.

1. Bun file-map identity was still broader than Bun 1.3.14. `deb245f99` first stopped Windows component-case folding, but its review proved that Bun also preserves relative versus absolute spelling and `.`/`..` segments; only separators and drive-letter case are normalized. The over-broad matcher could silently yield a real disk input and skip a required ttsc transform. `7282649e5` now compares exactly that Bun identity. Unit controls distinguish early ownership from a no-op transform through the options provider, and real Bun builds retain memory-only, override, relative, separator, cwd-change, and absolute cases while using disk for case-only, relative/absolute, and dot-segment aliases.
2. Cancellation stopped stream, digest, decompression, and bodyless fallback waits, but the public custom-fetch seam could still stall in the fetch Promise or metadata JSON. `b03aece07` added those abort races. Its review then reproduced an orphan rejected Promise when fetch synchronously aborted, uncancelled responses that fulfilled after losing the race, and tests that aborted before the intended inner phase had started. `f54a700b2` makes the abort helper own a producer before evaluating the changed signal state, disposes late Response bodies, preserves the exact abort reason, and uses explicit phase latches for metadata JSON, stream reads, bodyless reads, and fetch handoff. A late rejection after synchronous abort is also exercised under Node's unhandled-rejection policy.

The complete Playground suite, all Bun adapter cases, the exact Bun 1.3.14 matrix, and targeted package builds pass after remediation. Round 2 remains a findings round; a new full-surface review starts from the remediated head.

samchon (commented):
Overall Self-Review round 3 is not clean.

I re-read the complete 54-file change surface after the Bun and response-handoff remediations, including the Unplugin lifecycle/race paths, Metro linked-input ownership, Playground archive and package-resolution boundaries, public options, documentation, and all 33 changed feature-test contracts.

Finding:

- Medium: `installPlaygroundDependencies` can lose the exact abort reason at its archive-stage error boundary. Its catch block currently checks `signal.aborted` and rethrows the stage error. If an archive operation rejects and the signal aborts before the caller's catch continuation runs, the public install promise rejects with that older stage error even though cancellation is now authoritative. A deterministic reproduction used a successful custom tarball response whose `headers.get()` queued `controller.abort(reason)` and then threw. The result was `signal.aborted === true`, but the promise rejected with the header error rather than `signal.reason`. The catch must call `throwIfAborted(options.signal)` before wrapping a non-abort stage error, and a microtask-handoff regression must pin exact reason identity.

The prior remediation commits are independently clean: `2c8933bc9` preserves aborts across the nested response handoff and disposes unused bodies, `8ec5346b4` completes the actual Bun 1.3.14 Windows path-identity matrix, and `99d385047` restores strict typing without weakening the integrity mutation. Changed-file Prettier, `git diff --check`, and the 33 feature documentation/export contracts pass. The current CI head has no failed checks, but this review cannot be marked clean until the public install boundary is corrected and a fresh Overall Self-Review finds nothing further.

samchon (commented):
Overall Self-Review round 4 is not clean.

The archive-stage remediation in `bed99558d` correctly restores exact cancellation at the public install catch, but its mandatory Individual Self-Review found the symmetric metadata boundary immediately above it.

A real `Response` with a controlled `json()` Promise reproduced the gap deterministically. After `fetchNpmMetadata()` had selected `Error("metadata JSON failed")`, the test queued `controller.abort(reason)` between the internal rejection handler and `installPlaygroundDependencies()` resuming. On `bed99558d`, the public promise observed `signal.aborted === true` but still rejected with the stale metadata error rather than `signal.reason`.

Commit `62c026436` remediates that boundary by catching metadata resolution, calling `throwIfAborted(options.signal)`, and rethrowing only a still-authoritative non-abort error. The same feature now pins both metadata and archive handoffs with exact reason identity. All four async archive-install boundaries are fenced: metadata, download, integrity verification, and unpacking.

The Playground package build, repository-wide `test:typecheck`, the focused handoff regression, all 35 Playground features, changed-file Prettier, feature-contract scan, and `git diff --check` pass. This remains a findings round because the metadata gap was discovered here. A completely fresh full-surface review follows from `62c026436`.

samchon (commented):
Overall Self-Review round 5: clean.

This pass restarted from `62c026436` and re-read the complete 55-file change surface rather than checking only the last remediation. I revalidated:

- Unplugin generation ownership, first-delivery cost, persistent validation, stale-generation identity guards, volatile outputs, watch derivation, linked inputs, and Vite/Bun/Metro/Turbopack lifecycle distinctions;
- Bun 1.3.14 loader composition and its exact `BuildConfig.files` identity rules for separators, drive-letter case, component case, relative/absolute spelling, dot segments, cwd changes, and NUL virtual ids;
- Playground package exports selection versus loading, array blocking, target confinement, legacy directory fallbacks, alias self-reference, and nearest package scope;
- npm metadata, response, stream, digest, decompression, tar-record, root-confinement, byte-budget, error-context, and cancellation handoffs, including the exact reason at metadata and archive caller boundaries;
- public option types and package/website guidance; and
- every changed feature file's single export, `Verifies` summary, rationale, and numbered scenario contract.

The final metadata remediation has a clean Individual Self-Review. Its controlled Promise order is deterministic, late rejection is handled, optional 404 behavior is unchanged, and non-abort metadata errors retain their original identity. The archive boundary continues to preserve package/version context and original causes for non-abort failures.

Verification on the final head:

- complete `@ttsc/test-unplugin` suite: pass;
- complete `@ttsc/test-metro` suite, 59 features: pass;
- complete `@ttsc/test-playground` suite, 35 features: pass;
- repository-wide `pnpm test:typecheck`: pass;
- affected package builds: pass;
- all 34 changed feature contracts: pass;
- changed-file Prettier and `git diff --check`: pass;
- exact Bun 1.3.14 behavior controls and the warm Typia/Banner performance controls remain consistent with the implementation.

No actionable correctness, reliability, security, compatibility, documentation, or test-integrity finding remains. The evidence still identifies the dominant #969 defect in ttsc/Unplugin, not Typia.

PR #981, opened 2026-07-24
feat(lint): make project-rule inputs live across watch sessions

### LINKED ISSUE #971: feat(lint): let ProjectRules declare config-derived external inputs
## Problem

Project-scoped lint contributors can read files outside the TypeScript `Program`, but the contributor protocol cannot declare those inputs to the host. A contributor that resolves Markdown or OpenAPI paths from rule options is therefore invisible to both CLI watch topology and resident LSP invalidation.

A generic dependency contract is required before either client can react precisely. Watching the whole workspace is not acceptable: existing behavior intentionally ignores unrelated README changes, and a config glob must remain watchable even when it currently matches no file or the referenced file currently fails to parse.

## Verified behavior

- `ITtscLintPlugin` exposes native source identity but no runtime/config-derived data dependencies (`packages/lint/src/structures/ITtscLintPlugin.ts:27-32,49-55`).
- CLI watch topology is built from compiler inputs, config/output topology, and native plugin/contributor sources (`packages/ttsc/src/launcher/internal/watchTopology.ts:176-205`; `packages/ttsc/src/compiler/internal/runBuild.ts:203-207`).
- LSP receives broad watched-file events, but an unknown non-Program URI can only cause generic full resident invalidation (`packages/lint/linthost/serve.go:125-151`).
- A “files successfully read in the last cycle” snapshot is insufficient: missing exact paths, zero-match globs, parse failures, and future create/rename events must remain dependencies.
- Remote URLs cannot be represented honestly as filesystem dependencies.

This is distinct from #799: that issue established cache invalidation on open/watched changes, but did not add an external-input declaration contract.

## Proposed contract

Add an opt-in ProjectRule/check-contributor protocol that publishes config-derived external local dependencies after options and physical project identity are known.

The dependency model should support:

- exact local paths and glob populations;
- create/change/delete/rename membership;
- a physical project root and normalized cross-platform identity;
- dependencies that remain valid through missing files, zero matches, or contributor failure;
- reconciliation after config/options changes;
- a snapshot consumable by both the CLI launcher and resident LSP;
- protocol compatibility for contributors that declare no external inputs.

Remote HTTP(S) inputs must stay outside the filesystem model; a separate polling or conditional-revalidation policy can own those.

## Acceptance criteria

- A fixture contributor declares one exact Markdown path and one JSON/YAML glob derived from rule options.
- The dependency snapshot preserves both when the file is missing, the glob matches nothing, or parsing fails.
- Configuration changes replace old dependencies without leaking watchers or project identities.
- Windows case/separator normalization, symlink aliases, external project roots, and project references have explicit semantics and tests.
- Multiple rules/contributors can share dependencies without duplicate ownership or disposal leaks.
- CLI and LSP consumers can obtain the same reconciled snapshot without parsing contributor-specific options.
- Existing contributors and older protocol peers preserve behavior when the extension is absent.
- Remote URLs are documented as non-filesystem dependencies.

Follow-up consumers:

- CLI watch integration: #973
- LSP diagnostic refresh: #974

### LINKED ISSUE #972: perf(ttsc): reuse a resident incremental Program in check watch mode
## Problem

`ttsc check --watch` is correct for TypeScript-triggered cycles, but it does not reuse an incremental Program. Every event repeats plugin discovery, starts a fresh native lint sidecar, constructs a new Engine, loads a new Program/Checker, and runs one ProjectRule cycle. A long-lived watch therefore pays a cold one-shot build cost for every saved edit.

This is the opposite failure mode from stale caching: freshness is strong, but the watch optimization is absent.

## Verified behavior

- Each watch event calls `runOnce`, which invokes `runBuild` again (`packages/ttsc/src/launcher/internal/runTtsc.ts:726-817`).
- Each check plugin invocation runs a new native command and process (`packages/ttsc/src/compiler/internal/runBuild.ts:1123-1187`).
- The lint sidecar creates a new Engine/Program and executes one lint cycle (`packages/lint/linthost/compile.go:300-345,470-475`).
- The LSP already has a resident model: known saved TypeScript sources use `UpdateProgram`, recreate the Checker, and reset ProjectRule cycle state (`packages/lint/linthost/host.go:317-346`). Its incremental result is covered against cold-load corpus tests.
- `driver.Session` documents why transform reuse is unsafe; that does not preclude a dedicated no-emit/check lane (`packages/ttsc/driver/session.go:9-18`).

External-input correctness is tracked separately in #971. This issue is the Program/sidecar reuse optimization after a watch cycle has been triggered.

## Proposed direction

Introduce a resident check-mode session that reuses the parsed Program across compatible watch events. Reuse the already-tested LSP update semantics where practical:

- known TypeScript edit: incremental Program update and a fresh Checker;
- root-file, config, contributor topology, or declared external-input topology change: sound full reload;
- fresh Engine/ProjectRule state and reporter for every cycle;
- legacy cold sidecar fallback when resident mode cannot be used;
- keep transform/build watch outside the initial scope unless AST mutation isolation is proved.

## Acceptance criteria

- A no-emit/check watch fixture proves the native sidecar and Program are reused across ordinary saved TypeScript edits.
- Diagnostics after each update match a cold one-shot check for the same filesystem state.
- ProjectRule state/reporters never leak between cycles.
- Root add/delete, tsconfig change, contributor Go-source change, and external topology change select the documented reload path.
- Process shutdown and watch disposal release the resident sidecar.
- Performance telemetry or a benchmark demonstrates fewer parse/check and process-start costs across a representative edit sequence.
- Transform/build behavior remains unchanged unless separately proven safe.

### LINKED ISSUE #973: fix(ttsc): rebuild check-stage plugins when declared external inputs change
## Problem

`ttsc check --watch` does not rebuild when a configured ProjectRule input outside the TypeScript Program changes. A referenced Markdown heading or local OpenAPI JSON/YAML document can become stale on disk while the last evidence diagnostic remains visible indefinitely; touching a TypeScript/config/plugin file finally triggers a cold rebuild and reveals the new state.

Watching every Markdown/JSON/YAML file is not a valid fix. The existing unrelated-README regression test intentionally requires such edits to remain quiet. The launcher must consume contributor-declared dependencies from #971.

## Minimal reproduction

1. Configure a project rule whose TypeScript source cites `docs/spec.md#required`.
2. Start `ttsc --noEmit --watch` with a valid graph.
3. Rename or delete the Markdown heading.
4. Observe no rebuild and no changed diagnostic.
5. Touch the TypeScript source; the ensuing rebuild rereads Markdown and reports the old target unresolved and the new target missing.

A local configured Swagger/OpenAPI file has the same lifecycle.

## Verified cause

- Watch topology includes compiler-resolved inputs, config/output topology, and native plugin/contributor Go sources (`packages/ttsc/src/launcher/internal/watchTopology.ts:31-38,176-205,263-305`; `packages/ttsc/src/compiler/internal/runBuild.ts:203-207`).
- There is no contributor data-input channel today.
- `tests/test-ttsc/src/features/compiler/test_ttsc_watch_ignores_an_unrelated_readme.ts` passes, proving workspace-wide document watching would regress the intended boundary.
- Once some known event starts a rebuild, the check is fresh; this issue is missing event topology, not stale ProjectRule caching.

Depends on #971.

## Acceptance criteria

- Declared exact local dependency change triggers exactly one debounced rebuild.
- Create/delete/rename under a declared glob also triggers one rebuild, including a formerly zero-match glob.
- Markdown and local Swagger JSON/YAML real-launcher fixtures prove diagnostics update without touching TypeScript.
- An unrelated README remains quiet.
- Config changes reconcile watch membership immediately.
- Native plugin source changes retain their relink behavior; data-only changes reuse the linked plugin binary.
- Project references, external physical project roots, output-directory exclusion, atomic-save patterns, Windows paths, and watcher cleanup are covered.
- Remote URLs are not treated as filesystem watches.

### LINKED ISSUE #974: fix(ttscserver): recompute project diagnostics after external inputs change
## Problem

When an external ProjectRule input changes, the LSP may receive a workspace watched-file event, but it only invalidates compiler caches and refreshes hints. It does not immediately recompute and republish plugin/project diagnostics. The next save or lint verb becomes fresh, leaving the editor diagnostic stale in the meantime.

For unknown non-Program files, the resident lint sidecar also discards the entire TypeScript Program even though only ProjectRule data changed. This is correct on the next request but unnecessarily expensive.

## Minimal reproduction

1. Open a clean TypeScript document in a valid evidence project that cites a Markdown heading.
2. Delete or rename that heading.
3. Let the workspace watched-file event reach ttscserver.
4. Observe that the previous project diagnostic is not replaced or cleared.
5. Save/touch a TypeScript file; only then are fresh plugin diagnostics published.

Local Swagger/OpenAPI JSON/YAML has the same lifecycle.

## Verified cause

- Watched-file handling invalidates resident state and refreshes hints but does not schedule plugin/project diagnostic recomputation (`packages/ttsc/internal/lspserver/lsp_proxy.go:2267-2332`).
- Plugin diagnostics run through didOpen/didSave or the upstream push-diagnostic merge path (`lsp_proxy.go:332-345,1200-1265`).
- The deleted-watched-file test explicitly pins that an ordinary watched change publishes nothing.
- Unknown source URIs trigger full Program invalidation (`packages/lint/linthost/serve.go:125-151`).
- typescript-go commonly registers a broad workspace watcher, but that is an implementation side effect, not a precise contributor dependency contract.

Depends on #971. This is distinct from #799: that issue made resident caches invalidate, but did not republish project diagnostics after external-only changes.

## Acceptance criteria

- A declared external dependency event is debounced/coalesced and immediately recomputes affected plugin project diagnostics.
- The new publish replaces or clears the previous generation exactly once.
- External-only changes retain the TypeScript Program/Checker and rebuild fresh Engine/ProjectRule state only.
- Dirty TypeScript buffers are not diagnosed from stale disk content; the existing suppression boundary remains explicit.
- A project-level diagnostic verb or equivalent path works even when no TypeScript document is open.
- A generation guard prevents a slower previous computation from publishing after a newer event.
- Unrelated wildcard-watcher events do not run contributors.
- Markdown and created/changed/deleted local Swagger fixtures are covered, including cold-versus-resident equivalence.
- Remote URLs remain outside filesystem event semantics.

### LINKED ISSUE #982: fix(ttscserver): register declared ProjectRule input watchers
## Problem

`ttscserver` can classify and refresh a declared `ProjectRule` input after a `workspace/didChangeWatchedFiles` notification, but it never asks the LSP client to watch the exact files/globs published by `project-inputs`.

The bundled VS Code client statically watches only `**/{tsconfig,jsconfig}*.json`. Relying on a broad watcher that typescript-go may register is an upstream implementation side effect, not the contributor dependency contract, and does not prove that Markdown or local Swagger create/change/delete events reach the proxy.

## Verified cause

- `packages/vscode/src/extension.ts` registers only config JSON file events.
- `NativePluginSource` discovers and retains exact files/globs, but `Proxy` only consumes incoming watched-file notifications; it emits no `client/registerCapability` request for those declarations.
- The #974 proxy test injects `workspace/didChangeWatchedFiles` manually, so it proves refresh semantics but not event delivery.

## Acceptance criteria

- After initialize, clients with `workspace.didChangeWatchedFiles.dynamicRegistration` receive one deduplicated registration covering every declared exact file and glob, including missing exact files, zero-match globs, and paths outside the project root.
- Exact filenames escape glob metacharacters and file URIs preserve spaces, `#`, `%`, Windows drives, and UNC roots.
- A successful project-input refresh replaces the prior registration without leaking registrations or swallowing unrelated upstream request responses.
- Clients without dynamic registration get an explicit, truthful fallback rather than silently claiming live external diagnostics.
- Registration responses are consumed by the proxy and never forwarded to typescript-go.
- Driver tests prove the initialize-to-register wire sequence and a subsequent declared change recomputes project diagnostics.

Depends on #971 and complements #974.

### LINKED ISSUE #984: perf(ttsc): remove per-file polling from declared project input watches
## Problem

The replacement-safe follow-up switched declared ProjectRule inputs from event-driven directory watching to `fs.watchFile(..., { interval: 100 })` for every exact path, glob directory, and current glob match. A large generated OpenAPI/Markdown tree therefore performs O(files + directories) stat probes ten times per second while idle. This fixes Windows rename handles by replacing them with an unbounded polling tax, which conflicts with the resident-watch optimization campaign.

## Verified cause

- `WatchTopology.syncProjectInputWatchers` walks the complete glob population and installs one `fs.watchFile` per root, descendant directory, and matched file.
- Every watcher uses a fixed 100 ms interval.
- One 10,000-file declared population therefore owns roughly 10,000 persistent stat watchers even when no file changes.
- `website/src/content/docs/development/concepts/protocol.mdx` still describes recursive `fs.watch`, so implementation and protocol documentation disagree.
- A recursive `fs.watch` on a stable parent/root does not hold per-descendant handles on Windows and permits sibling-staged atomic directory replacement; the earlier EPERM came from additional direct/observed child watchers, not from the stable recursive root itself.

## Acceptance criteria

- Declared exact paths and glob populations are event-driven while idle; no per-file or fixed-interval polling scales with population size.
- Missing exact files, zero-match globs, nested changes, filename-less events, and atomic replacement remain observable.
- Windows replacement staging is not blocked by direct child handles.
- Unrelated events and compiler outputs remain filtered before population scans where possible.
- Tests assert watcher cardinality is bounded by declaration roots rather than current match count.
- Protocol documentation matches the implemented watcher ownership model.

### LINKED ISSUE #985: fix(ttsc): invalidate resident check sessions for every execution transition
## Problem

Resident check watch can keep a stale execution after lint config changes because the config is classified as a project input, and equivalent no-emit watch forms do not enter the resident path. A config can select a different contributor while the old binary and Program remain alive.

## Acceptance

- Config, plugin, contributor, and root-set transitions create a fresh resident process and execution.
- Source-only and content-only external changes reuse the resident Program where safe.
- Both check --watch and effective noEmit watch forms use the same resident path.
- End-to-end telemetry tests assert PID/Program replacement, reuse, and shutdown without leaks.

### LINKED ISSUE #986: fix(ttsc): make watch output inference match tsgo option semantics
## Problem

Compiler-output filtering still diverges from tsgo for CLI aliases, react-native JSX, inherited outFile/tsBuildInfoFile bases, and final passthrough precedence. A generated output can therefore retrigger watch, while a real external input can be suppressed as a presumed output.

## Acceptance

- Resolve one-dash, two-dash, case variants, and aliases through the canonical flag schema.
- Match tsgo JSX extensions and final argv override precedence.
- Preserve the declaring config directory for inherited path options.
- Verify effective output sets with focused tests and real watch fixtures.

### LINKED ISSUE #988: fix(ttscserver): reconcile project-input registrations after rejected requests
## Problem

If desired watchers change while register/unregister is in flight and that client request is rejected, the error path clears pending state without reconciling the newest desired snapshot. Registration can remain stale indefinitely.

## Acceptance

- Reconcile latest desired state after every success and failure completion.
- Preserve last-good registration when replacement registration fails.
- Retain and retry stale unregister work without duplicating active registrations.
- Cover B-pending/C-desired rejection races.

### LINKED ISSUE #989: fix(ttscserver): resume deferred project diagnostics after save and generation races
## Problem

An external-input refresh deferred for a dirty document is not explicitly resumed on didSave. A newer document generation can also reject a direct write while the caller still clears pending state. Old project diagnostics may remain published indefinitely.

## Acceptance

- Saving resumes every still-pending direct project refresh.
- Clear pending state only after the intended generation is published or safely superseded.
- Cover dirty/save, parse-diagnostic omission, and newer-generation rejection races.

### LINKED ISSUE #990: fix(lint): reload resident Programs when JSON module topology changes
## Problem

Resident change classification treats JSON creation/deletion as data-only. With resolveJsonModule, a new JSON file can satisfy an unresolved Program import, but the warm Program skips the unknown external path and stays stale.

## Acceptance

- Classify JSON topology changes as Program-invalidating when they can affect module resolution.
- Preserve warm reuse for safe content-only external data changes.
- Add create, delete, and edit tests with resolveJsonModule.

### LINKED ISSUE #991: fix(ttscserver): preserve project diagnostics across partial plugin failures
## Problem

When one project-diagnostic plugin succeeds and another fails, the partial aggregate replaces the full prior publication and silently clears the failed producer's findings. All-plugin failure retains the old publication, so failure semantics are inconsistent.

## Acceptance

- Never erase a failed producer's last-good diagnostics because another producer succeeded.
- Define per-plugin last-good aggregation or require an atomic successful refresh.
- Cover success/failure/recovery across multiple plugins.

### LINKED ISSUE #992: feat(ttsc): advertise direct project-diagnostic capability explicitly
## Problem

The projectInputs capability promises only topology discovery, while the host also assumes support for lsp-project-diagnostics and its resident verb. A staged or third-party sidecar can implement the documented snapshot command but reject the undocumented diagnostic command with no one-shot fallback.

## Acceptance

- Advertise direct/resident project diagnostics independently from projectInputs.
- Gate commands by their exact capability.
- Provide the same unsupported-resident fallback discipline used by hints.
- Test mixed-version and third-party capability combinations.

### LINKED ISSUE #993: fix(ttsc): reject project-input snapshots outside the physical project identity
## Problem

Snapshot validation accepts the first plugin's root without comparing it to the launcher's physical fallback root. On Windows it also accepts rooted paths without a drive or UNC volume, then path.resolve attaches the launcher's current drive. A sidecar can redirect long-lived watchers to the wrong project or volume.

## Acceptance

- Compare every snapshot root to the launcher's canonical physical project root, including the first and only snapshot.
- Require drive-qualified or UNC absolute local paths on Windows; reject volume-less rooted paths.
- Preserve case/separator alias handling and last-good snapshot retention.
- Add platform-focused parser and merge tests.

### LINKED ISSUE #994: fix(project-inputs): canonicalize Windows long and short path aliases
## Problem

The Windows CI lane proves project-input identity is not stable across 8.3 short paths and long paths. The lint sidecar publishes a long physical path while tests and editor URIs can retain RUNNER~1 aliases; LSP matching then returns false, and snapshot tests compare two spellings of the same file as different.

## Evidence

windows-go failed TestProjectInputsMatchMissingExactAndGlob and three lint snapshot tests on PR #981. The only difference is RUNNER~1 versus runneradmin in the same physical path.

## Acceptance

- Canonical identity treats Windows long/short aliases, case, slash variants, symlinks, and missing descendants consistently.
- Missing paths inherit canonicalization from the nearest existing ancestor.
- Published paths and URI matching share the same identity rules.
- Windows-specific tests cover exact files, zero-match globs, and config provenance without string-spelling assumptions.

### LINKED ISSUE #995: fix(ttsc): isolate pending resident changes for duplicate check entries
## Problem

Multiple check-stage entries can resolve to the same resident process key. The first entry consumes and deletes the key-level pending request, so the second sends undefined and receives an invalid-request failure. This regressed duplicate plugin configurations in 5ce398fad.

## Acceptance

- Every configured check entry receives a valid complete change request, or identical execution is deliberately deduplicated without losing diagnostics.
- Failure in an earlier entry does not drop later pending transitions.
- Add a regression with two entries sharing binary/name/args and verify no undefined framing or status-2 fallback.

### LINKED ISSUE #996: perf(ttsc): deduplicate nested recursive project-input watcher roots
## Problem

Retained external declaration roots are deduplicated only by exact path. When one declaration retained /external and a later declaration resolves to /external/a, both recursive handles remain active. One edit can trigger duplicate full population scans and multiple project callbacks, contradicting the protocol documentation.

## Acceptance

- Prune any desired recursive root already covered by an ancestor, including roots retained across snapshot generations.
- Preserve the more specific retained root so it can become active when the ancestor declaration is removed.
- Assert actual/pure watcher root cardinality and bounded notification behavior for the staged missing-ancestor scenario.
- Keep missing-tree and atomic replacement behavior.

### LINKED ISSUE #997: docs(ttsc): define hard-link alias limits for event-driven project inputs
## Problem

After removing per-file stat polling, an exact input can change through a hard-link alias outside every watched ancestor without producing a filesystem event at the declared path. Node's event APIs cannot observe that cross-boundary alias on every supported platform. The declared file contents change but watch does not rebuild.

## Acceptance

- State the event boundary precisely: edits must generate an event at the declared path or within its retained watcher ancestor.
- Call out hard-link aliases outside that boundary and network/filesystem event limitations.
- Keep symlink/physical-path and atomic replacement guarantees distinct from hard-link identity.
- Do not reintroduce O(files) polling as an undocumented fallback.

### LINKED ISSUE #998: fix(ttsc): preserve case-sensitive Windows project-input identities
## Problem

Snapshot merge lowercases every Windows path key. NTFS directories can opt into case sensitivity, so Project and project may be different physical roots and two inputs differing only by case may be distinct. The current root guard can therefore accept a foreign project and dedup distinct inputs.

## Acceptance

- Canonicalize through the nearest existing ancestor and filesystem-reported spelling rather than unconditional case folding.
- Continue accepting case, slash, 8.3, symlink, and extended-path aliases that resolve to the same physical root.
- Preserve distinct identities under a case-sensitive Windows directory.
- Cover root comparison, existing inputs, and missing descendants with platform-focused tests.

### LINKED ISSUE #999: fix(ttsc): classify overlapping plugin-source watcher events as reloads
## Problem

A project directory watcher can queue an event for a selected plugin source root before setExtraInputs installs the dedicated watcher. Once the root enters the compiler file map, that queued event is emitted as compiler instead of plugin. This causes an unnecessary Program cycle and can miss execution/plugin rebuild semantics if it is not coalesced with the dedicated event.

## Evidence

The focused watch_topology_classifies_config_and_plugin_reload_inputs test deterministically reports compiler(pluginRoot), config, config, plugin on Windows.

## Acceptance

- Any event at or below a selected plugin source root is classified as plugin regardless of which overlapping watcher delivered it.
- Plugin/config reload dominates narrower queued compiler/project changes.
- Focused classification test is deterministic and passes standalone and in the watch topology suite.

samchon:
## Intent

This pull request completes the accepted external-input campaign for `@ttsc/lint`, `ttsc check --watch`, and `ttscserver`.

Closes #971.
Closes #972.
Closes #973.
Closes #974.
Closes #982.
Closes #984.
Closes #985.
Closes #986.
Closes #988.
Closes #989.
Closes #990.
Closes #991.
Closes #992.
Closes #993.
Closes #994.
Closes #995.
Closes #996.
Closes #997.
Closes #998.
Closes #999.

## Result

- Project rules can publish normalized exact local files and `*` / `?` / `**` glob populations after options and physical project identity are resolved.
- CLI watch retains missing files and zero-match globs, refreshes Markdown and local Swagger diagnostics, survives atomic directory replacement, filters unrelated files and compiler outputs, and handles project references and Windows filesystem identities.
- Check-only watch uses a resident sidecar and incremental Program for compatible source and external-data edits. Config, root-set, plugin, and contributor topology changes select a sound reload path.
- `ttscserver` retains the warm Program for data-only changes, recomputes project diagnostics without an open document, guards dirty documents and stale generations, and dynamically registers precise LSP `RelativePattern` watchers after `initialized`.
- Clients without dynamic relative-pattern support receive an explicit warning. Remote HTTP(S) inputs remain contributor-owned because filesystem events cannot make them live.

## Verification

- `.github/workflows/test.yml` runs the complete `features/watch` suite on `ubuntu-latest`, `macos-latest`, and `windows-latest` with the current platform build.
- The existing split test matrix continues to cover Go, typecheck, package, native-plugin, LSP, and compiler surfaces.
- GitHub Actions is the final validation authority for this head; the PR remains draft until the matrix is green.
- Every coherent implementation group received a read-only Individual Self-Review, followed by a final overall review before merge.


samchon (commented):
Individual Self-Review — exact head `416efd6fc2106504be4cfe3389240c72eba7d4ce`

I re-read the coherent external-input implementation groups and their consequence surfaces at this exact head:

- project-rule snapshot production, exact-file/glob normalization, physical path identity, and output exclusion;
- CLI watch topology, missing/empty populations, resident Program invalidation, execution reload boundaries, and project references;
- LSP registration, producer ownership, dirty-document deferral, generation freshness, and partial failure recovery;
- final hardening in `c322dbec8`, `c02d900e0`, and `20d6f13e7`, including removal of unsound event suppression, precise filesystem identity assertions, and transitive local CJS config eviction;
- the final classifier comment correction in `4d4ae835b` and three-platform watch CI in `416efd6fc`.

Findings raised during the review campaign were converted into regression tests and follow-up commits rather than waived. I found no remaining blocking defect in the reviewed diff. Per maintainer direction, final execution validation is delegated to the GitHub Actions matrix, including the new Ubuntu/macOS/Windows watch jobs.

Issue #970, opened 2026-07-23
upgrading from 0.19 to 0.20 build never completes


khera:
I upgraded typia from 13.1.1  to 13.2.0,  ttsc and @ttsc/unplugin from 0.19.2 to 0.20.0. The build step never completes (it has been running 30 minutes and I finally killed it). Node version v26.5.0.

I tried upgrading just typia or just ttsc suite independently. They both result in the forever build. For now I'm sticking with the 0.19 unplugin, and the 13.1.1 typia.

```sh
npm install ttsc@latest @ttsc/unplugin@latest @ttsc/graph@latest typia@latest
 [ .. output skipped ]

[lappy]% npm run typecheck
npm notice run makemedia-app@1.8.24 typecheck
npm notice run react-router typegen && ttsc --noEmit
ttsc: building source plugin "typia" from /Users/vivek/Work/makemedia/makemedia-app/node_modules/typia/native/cmd/ttsc-typia (this runs once per cache key and can take several minutes on a cold Go cache) See https://ttsc.dev/docs/ttsc/compile#plugin-cache to persist it across builds.
[lappy]% npm run typecheck
npm notice run makemedia-app@1.8.24 typecheck
npm notice run react-router typegen && ttsc --noEmit
[lappy]% npm run build
npm notice run makemedia-app@1.8.24 build
npm notice run react-router build
Using Vite Environment API
vite v8.1.5 building client environment for production...
load: 5.31  cmd: node 20489 running 1.13u 1.25s
load: 5.44  cmd: node 20489 running 100.66u 377.74s
load: 6.03  cmd: node 20489 running 128.87u 484.31s
load: 5.39  cmd: node 20489 running 166.09u 619.48s
load: 4.55  cmd: node 20489 running 185.90u 689.69s
load: 5.99  cmd: node 20489 running 237.74u 882.93s
load: 4.75  cmd: node 20489 running 253.44u 942.37s
load: 6.93  cmd: node 20489 running 307.98u 1141.14s
load: 7.06  cmd: node 20489 running 349.43u 1295.37s
load: 5.95  cmd: node 20489 running 406.98u 1515.07s
^Cload: 4.48  cmd: node 20489 running 437.71u 1632.87s
load: 5.53  cmd: node 20489 running 501.91u 1869.50s
^\zsh: quit       npm run build
[lappy]%
```

[...]

PR #1009, opened 2026-07-25
fix(unplugin): bound per-delivery watch-input derivation; ci(release): fail-fast publish timeouts

### LINKED ISSUE #1007: fix(unplugin): per-delivery watch-input derivation rescans the whole reference graph, so graph-envelope builds never complete
## Problem

A Vite production build that serves modules from `@ttsc/unplugin`'s project transform cache re-derives every module's watch-input list from scratch on **every cache hit**. The derivation rebuilds the full reference-graph index — once in `selectReachableEdges` and again in `selectReachableSources` (via `selectResolutionCandidateInputs`) — and computes a filesystem identity key for every path it touches. With a graph-bearing transform envelope, the cost of one module delivery is O(edges); a build that delivers M modules pays O(M × edges) identity computations.

On macOS each `pathIdentityKey` call additionally performs real syscalls (`fs.existsSync` + `fs.realpathSync.native`) even when its case-sensitivity verdict is already cached (`packages/unplugin/src/core/transform.ts:1604-1633`); the win32/linux branches are pure JS. The per-delivery graph rescan therefore becomes a per-delivery syscall storm on macOS, and real application builds effectively never finish.

This is the residual of #970. PR #980 removed the per-module full-project revalidation, but per-delivery watch-input derivation keeps the same O(M × graph) scaling. The version boundary the reporter verified — typia 13.1.1 fast, typia 13.2.0 stalled, both on `@ttsc/unplugin` 0.21.0 — is the envelope shape: typia ≥ 13.1.19 stamps `graph` + per-file `dependencies` (samchon/typia#2100, samchon/typia#2140), and 13.1.1 stamps neither. With no `graph`, the derivation selectors return early and the per-delivery cost vanishes.

**Expected:** one transform generation pays graph-index construction once; each module delivery is bounded by that module's own reachability closure, not by the whole graph.

## Evidence

Reproduction (this repository, `experimental/unplugin-perf`, extended with a graph-envelope scenario that drives the real Rollup adapter over a synthetic project; `process.platform` is flipped to `darwin` after the compile so the macOS probe volume is observable on any host):

| modules | graph edges | whole-project compiles | fs probes | probes / module delivery | wall time (Windows) |
|--:|--:|--:|--:|--:|--:|
| 100 | 2,599 | 1 | 665,432 | 6,721 | 75.8 s |
| 100 | 5,099 | 1 | 1,195,132 | 12,072 | 87.6 s |
| 100 | 10,099 | 1 | 2,254,531 | 22,773 | 95.2 s |

probes/module ≈ 2.2 × edges, on pure cache hits with exactly one compile. A real react-router-scale program (edges in the tens of thousands across `node_modules` `.d.ts`, ~2k delivered modules) extrapolates to hundreds of millions of syscalls — hours for a build the reporter kills after minutes, with the reporter's exact CTRL-T signature (system CPU ≈ 2× user CPU, i.e. syscall-bound). The same run with a graph-free envelope (the typia 13.1.1 shape) derives nothing and shows no probes.

Root causes, all in `packages/unplugin/src/core/transform.ts`:

1. `selectReachableEdges` (transform.ts:540-578) and `selectReachableSources` (transform.ts:585-626) each rebuild the complete edge index — O(edges) `pathIdentityKey` calls — **per module delivery**; `selectResolutionCandidateInputs` (transform.ts:467-497) then rescans every candidate source per delivery.
2. `pathIdentityKey` (transform.ts:1597-1633) costs `existsSync` + `realpathSync.native` per call on darwin; only the case-verdict toggle is memoized, not the probe.
3. `declaresFile` (transform.ts:686-700) linearly scans `volatile` / `dependenciesComplete` with two identity computations per member per delivery (`isVolatileFile` runs at least twice per delivery).
4. `selectTransformedSource` (transform.ts:1482-1491) and `selectFileDependencies` (transform.ts:722-732) fall back to O(map) identity scans per delivery whenever the fast project-relative key misses (out-of-root keys — monorepo layouts).

All hosts funnel through `transformTtsc`, so every adapter (Vite, Rollup, webpack, Rspack, esbuild, Farm, Turbopack, Bun, Metro) carries the defect; macOS amplifies it into the reported stall.

## Consequence surface

- **Who:** any bundler user whose plugin stamps a `graph` envelope — typia ≥ 13.1.19 (including 13.2.0) projects are the entire known population. Worst on macOS; large Windows/Linux builds still pay the O(M × edges) pure-JS derivation.
- **Where:** watch-input derivation runs identically in one-shot production builds, `build --watch`, and dev, so the stall hits all three; dev/serve additionally keeps the deliberate per-delivery complete-input validation from #980 (measured separately: linear, not quadratic — out of scope here).
- **Freshness invariant:** the derived watch lists are per-build registrations bundlers consume for invalidation; the fix must preserve their exact content and order per module (pinned by existing tests such as `test_transformttsc_notifies_watch_files_on_cached_transforms` and the `transform-graph` suite).
- **Memoization boundary:** a memo may live at most one generation. Every derivable path is a recorded external/project input, so persistent-mode validation already proves those paths unchanged on every non-build-scoped hit; a change invalidates the generation and its memo with it.

## Approach

Owner: `packages/unplugin/src/core/transform.ts`. Introduce per-envelope derivation state (a `WeakMap` keyed by the compiler result object, dying with the generation):

- build the direct-edge index, source spellings, candidate entries, and the `volatile` / `dependenciesComplete` identity sets **once per envelope**;
- memoize `pathIdentityKey` results inside that state;
- memoize the final derived watch-input list per file;
- give `selectTransformedSource` and `selectFileDependencies` first-match identity indexes built once per envelope so their fallback scans disappear.

The invariant is: per delivery, work is O(reach of that module) with zero filesystem probes beyond the first derivation; per generation, O(edges) exactly once. Derived lists must remain byte-identical to today's derivation (same union, same order, same dedup, same exclusions of the module itself and the disposed temporary tsconfig).

- `experimental/unplugin-perf` Scenario C (already committed with this reproduction) gates `probes/module ≤ 64` under the simulated-darwin probe counter at graph sizes 2.6k/5.1k/10.1k edges — it fails with the numbers above pre-fix and must pass post-fix, with `pluginRuns == 1` preserved.
- New unit coverage in `tests/test-unplugin`: cache-hit deliveries of sibling modules from one graph-bearing generation must not re-probe the filesystem per delivery (probe counter + simulated darwin), while the derived watch list per module stays exactly correct.
- The existing `transform-graph`, `transform-dependencies`, `transform-complete`, and project-cache suites must pass unmodified (derivation parity).
- Negative/boundary: envelopes without `graph`, malformed graph members, `dependenciesComplete` narrowing, `volatile` interplay, cycles pointing back at the module, missing candidate paths.
- Broader: `pnpm --filter @ttsc/unplugin build`, the unplugin feature tests, and the PR's ordinary CI.

## Coordination

- Residual of #970 (closed; reporter confirmed 0.21.0 + typia 13.2.0 still stalls). This issue is the remaining root cause; a comment cross-links it there.
- Unrelated to the v0.22.0 release run: that run completed successfully in its historical duration (evidence gathered during discovery; no release-pipeline defect).
- The Vite serve-mode complete-validation cost with large external input sets is deliberate (#980 freshness contract) and linear; quantified during discovery and deliberately not changed here.

### LINKED ISSUE #1008: ci(release): publish steps have no fail-fast timeout, so a stalled publish holds the release silently for hours
## Problem

Every step and job in `.github/workflows/release.yml` runs with GitHub's default 360-minute timeout. The publish job's externally-connected steps — `Publish VS Code Marketplace extension` (`vsce publish`), `Publish to npm` (`pnpm -r publish`), and the `release` job's `pnpm dlx changelogithub` — perform network calls that can stall indefinitely, and nothing in the workflow bounds them. A stalled publish would burn up to six hours of runner time per job while presenting in the Actions UI as merely "in progress", indistinguishable from the normal ~21-minute build phase that produces no publish output.

This is not hypothetical in two directions:

- This exact pipeline already deadlocked once at the Marketplace boundary (d90544084, "ci(release): don't let the marketplace preflight deadlock a fresh launch").
- During the v0.22.0 release the run *appeared* frozen to the releaser, who could not tell a healthy run from a stuck one without polling the API for step timestamps. The run was in fact healthy (Marketplace step: 11 s; npm publish: ~5.5 min; total: ~35 min, matching v0.21.0's 36 min) — but the only signal available was waiting.

**Expected:** each externally-connected step fails fast with a clear timeout, so a genuine stall surfaces as an actionable failure in minutes instead of a silent six-hour hold.

## Evidence

- `.github/workflows/release.yml` declares no `timeout-minutes` at any job or step level (verified at v0.22.0, commit 74b5d414f).
- Observed healthy durations (runs 30161890750 and 30051461166): Marketplace publish 11-28 s, npm publish ~5.5 min, whole publish job ~35 min including the ~21 min build.
- Prior deadlock history in this exact workflow: d90544084.

## Consequence surface

- A stuck `vsce publish` today holds the npm publish hostage (steps are sequential) with no signal; the tag is already pushed, so the repository shows a release commit with no packages and no failure.
- `wasm-smoke`, `vscode-smoke`, and the `release` (changelog) jobs carry the same unbounded default.
- Timeouts must be generous enough never to kill a healthy run: observed durations put the Marketplace step at ~0.5 min and npm at ~5.5 min, so 10-minute and 20-minute step budgets carry >10x headroom; job-level guards (publish 60 min, smoke/changelog 10-15 min) cover the rest.

## Approach

Add explicit `timeout-minutes` to the two publish steps (Marketplace: 10, npm: 20) and job-level `timeout-minutes` to every job in the workflow (publish: 60, each smoke job: 15, changelog: 15). No behavior changes for healthy runs; no new credentials, scripts, or steps.

- `node scripts/...`-style static check or the existing workflow-order test (`tests/test-ttsc/src/features/platform/test_release_workflow_runs_preflight_before_publication.ts`) keeps passing; add coverage asserting the timeout fields exist, if a test home for workflow shape exists.
- The next real tag release completes with the guards present (observational).
- YAML validity: `pnpm format` (prettier) on the workflow file.

## Coordination

- Independent of #1007 (the unplugin derivation stall); both came out of the same campaign that investigated an apparently-frozen release.
- Keeps the d90544084 contract intact: nothing here reintroduces a Marketplace probe into the release path (the existing workflow test pins that).

samchon:
## Intent

One unified cycle for the `unplugin-watch-input-stall` campaign, opened as the implementation-free claim. This pull request owns the complete accepted cycle:

- #1007 — per-delivery watch-input derivation rescans the whole reference graph (O(modules × edges), syscall-bound on macOS), the residual root cause behind the #970 report that typia ≥ 13.1.19 builds never complete. Fix: per-envelope derivation indexes and memos in `packages/unplugin/src/core/transform.ts`, with derivation output byte-identical to today.
- #1008 — `release.yml` publish steps and jobs run unbounded (GitHub's 360-minute default). Fix: explicit fail-fast `timeout-minutes` with deliberate headroom over observed healthy durations.

Also included: `experimental/unplugin-perf` gains the graph-envelope reproduction (Scenario C, with a simulated-darwin fs-probe counter and a bounded-per-module invariant gate) and a serve-mode measurement (Scenario D). Scenario C is the regression gate for #1007 and currently fails on `master` by two orders of magnitude.

## Deferred

- Vite serve-mode complete validation with large external input sets: measured linear (207 reads/module at N=50/100 externals), deliberate #980 freshness contract. Any faster revalidation scheme is a separate design decision.

## Verification

- Final reviewed head: `976d6c3d9b4a935be4f61be4e3eca6d51d3fdde3`; local HEAD, remote branch, and PR head matched, with a clean worktree.
- Local targeted regression tests passed for release timeout ownership, the pinned Nestia checkout contract, and bounded watch-input derivation probes.
- TypeScript checks passed for `tests/test-ttsc` and `@ttsc/unplugin`; `pnpm install --frozen-lockfile --offline --ignore-scripts`, `pnpm format`, and `git diff --check origin/master...HEAD` passed.
- Scenario C changed from 6.7k/12k/22.8k filesystem probes per module and roughly 76–95 seconds for 99 deliveries to approximately 19 probes per module and roughly 100 ms, while preserving emitted watch-input ordering and deduplication.
- Every coherent implementation/correction commit received its mandatory read-only Individual Self-Review. Findings were either incorporated in a successor correction or explicitly verified clean; review links are recorded on the PR timeline.
- Final solo Overall Self-Review rechecked the complete diff, consequence surface, failure paths, workflow ownership, tests, lockfile, and issue closure contract and found no remaining sound improvement.
- GitHub Actions completed 64/64 checks successfully on the exact final head, including macOS watch coverage, Windows-Go, unplugin, graph/bench, native corpora, release/workflow contract tests, and the 47-minute pinned Nestia integration.

## CI repair

- The Nestia integration no longer clones moving `nestia/master`. PR #1009 exposed an `openapi_v2` failure already red in Nestia's own SDK lane after Nestia #1594 and reproducible from ttsc `master`; the workflow now pins Nestia's last full-green merge (`3b27e69b69dea3f102315042dce87c18d81be74a`, upstream run 29959327169) and a repository test locks that checkout contract.

## Closing

Closes #1007
Closes #1008


tests/test-ttsc/src/features/platform/test_release_workflow_bounds_publish_timeouts.ts
```diff
@@ -0,0 +1,50 @@
...
+  const workflow = source
+    .split("\n")
+    .filter((line) => !/^\s*#/.test(line))
+    .join("\n");
+
+  for (const [step, budget] of [
+    ["Publish VS Code Marketplace extension", "timeout-minutes: 10"],
+    ["Publish to npm", "timeout-minutes: 20"],
+  ] as Array<[string, string]>) {
+    const start = workflow.indexOf(`- name: ${step}`);
```
samchon: This slice does not end at the next YAML node of the same indentation, so the final npm step absorbs the following job. Together with counting job timeout lines instead of binding them to job names and values, the test permits misplaced or relaxed timeouts. Assert exact timeout ownership for every named job and publish step.

samchon (commented):
Individual Self-Review ? `4915edf6b` (`ci(release): fail fast when a publish step stalls`)

The workflow change itself is sound: the job and publish-step budgets are explicit and comfortably above observed healthy durations. The review found one test-ownership gap. The timeout test slices the final npm step through the following job and only counts four job-level keys, so valid YAML can move a timeout to the wrong node, relax every job to 360 minutes, or add an unbounded fifth job without failing.

Adjudication: accepted. The campaign lead is replacing the text count with exact indentation-scoped job/step mappings and exact budgets. Targeted verification passes; a correction commit and its own Individual Self-Review will follow.

samchon (commented):
# Overall Self-Review round 1 ? takeover audit

Reviewed head: `201f66ddf25eeb5fa758d6855b049f05c4f3f798`
Base: `74b5d414f4388cdd6dac1e040b0a2a9457987bc8` (`v0.22.0`)

This is the first exhaustive review after taking over the existing issue campaign. It is deliberately a **finding-and-correction round**, not the final clean gate. The PR remains draft until the correction ISR, the complete CI matrix, and a fresh zero-finding Overall Self-Review all pass on the same head.

## 1. Reconstructed facts and issue relationship

The apparently frozen release and the reopened #970 symptom share the user-visible shape ?progress stops,? but the evidence identifies two distinct causes.

- Release run [`30161890750`](https://github.com/samchon/ttsc/actions/runs/30161890750) completed successfully in roughly 35 minutes. The observed release was not deadlocked; the long build plus stale/quiet UI created the appearance of a freeze. Issue #1008 therefore hardens future releases against a *real* silent stall rather than claiming this successful run was stuck.
- The #970 residual is a real unplugin/macOS performance defect. With graph-bearing typia envelopes, every module delivery recomputed graph-derived watch inputs and repeated `existsSync`/`realpathSync.native` identity work across the graph. The cost grew as O(module deliveries ? graph edges), matching the reported typia-version-dependent stall. Issue #1007 tracks this verified residual.

## 2. Implementation review

### #1007 ? bounded watch-input derivation

Commit `1dab1e6c0` derives graph indexes, path identities, and per-output watch lists once per transform envelope/generation and reuses them for sibling deliveries.

I reviewed the full production diff and traced:

- cache lifetime and WeakMap reachability;
- concurrent deliveries from one envelope;
- graph cycles and repeated paths;
- dependency/output first-match fallback behavior;
- volatile/completeness propagation;
- malformed or partial envelope entries;
- temporary tsconfig exclusion;
- ordering and deduplication;
- Windows separator/case handling and the macOS realpath branch;
- project-root/result-root equivalence;
- benchmark instrumentation and the distinction between one-time derivation probes and persistent validation probes.

No sound production-code defect was found in this round. The implementation keeps one generation-local semantic result while removing repeated filesystem identity work. Scenario C?s gate remains tied to the real adapter path and catches the prior growth mode.

### #1008 ? release fail-fast budgets

Commit `4915edf6b` adds:

- `publish` job: 60 minutes;
- Marketplace publish step: 10 minutes;
- npm publish step: 20 minutes;
- `wasm-smoke`, `vscode-smoke`, and `release` jobs: 15 minutes each.

These budgets are materially above observed healthy execution while ensuring a genuinely wedged publication cannot hold the workflow indefinitely.

## 3. Findings accepted and corrected

### Finding A ? formatting gate failure

The prior head?s test workflow [`30166600875`](https://github.com/samchon/ttsc/actions/runs/30166600875) failed `typecheck` because Prettier interpreted a leading `+` in the new unplugin JSDoc as a list item. Commit `201f66ddf` rewrites the prose without changing the test or production semantics. Targeted Prettier validation now passes.

### Finding B ? timeout regression test did not prove YAML ownership

The initial test sliced the final npm step into the following job and counted four job-level timeout lines globally. It could therefore pass if a timeout moved to the wrong YAML owner, all budgets were relaxed, or a fifth unbounded release job was added.

Commit `201f66ddf` now:

- selects each YAML job/step mapping by exact indentation;
- requires the timeout line to be owned by that mapping;
- asserts the exact intended budget for every job and publish step;
- asserts the exact release-job set so a new unbounded job cannot silently bypass the contract.

The dedicated `bounds_publish_timeouts` test passes.

### Finding C ? performance fixture prose overstated graph fanout

The #1007 Individual Self-Review found that Scenario C described edges to all siblings while the fixture uses one next-sibling edge plus `K` external declarations. The edge count, fixture, measurements, and gate were correct. Commit `201f66ddf` corrects the prose to ?next sibling.?

## 4. Individual Self-Review record

- `1dab1e6c0` (#1007): no production defect; one accepted P3 documentation-accuracy finding, corrected in `201f66ddf`.
- `4915edf6b` (#1008): one accepted P3 test-ownership finding, recorded in the preceding formal review and corrected in `201f66ddf`.
- `201f66ddf` (review corrections): its mandatory read-only ISR is running now. Any sound finding will be adjudicated and corrected before the clean gate.

## 5. Verification evidence

Evidence inherited from the campaign ledger and rechecked against the committed paths includes the unplugin build/test path, the graph probe regression, the performance scenario, and the release workflow test. During this takeover I additionally ran:

- `pnpm format`: completed successfully, including generated flags and Go formatting;
- targeted Prettier check on all three correction files: passed;
- `pnpm --dir tests/test-ttsc start --include=bounds_publish_timeouts`: passed;
- `git diff --cached --check`: passed before commit;
- exact staged-file audit: only the three intended correction files were committed despite Windows `core.autocrlf` stat-cache noise.

A repository-wide `pnpm check:format` attempt exceeded a five-minute local command limit because its glob traversed the ignored benchmark `.tmp` fixture tree. This is recorded as a local timeout, not a pass. The actual correction files pass Prettier, `pnpm format` completed, and the full tracked-tree formatting gate is running in CI.

## 6. Prior red lanes and current disposition

The previous head had three red observations:

1. `typecheck`: deterministic formatting defect; corrected in `201f66ddf`.
2. `windows-go`: `TestLSPProxyDidSaveVersionlessAfterDirtyIgnoresStaleUpstream` missed a two-second upstream frame and then the proxy did not return. This path is outside the PR diff, but the campaign does not dismiss red CI. The new same-head run must pass or the failure will be reproduced and repaired here.
3. `nestia` run [`30166600883`](https://github.com/samchon/ttsc/actions/runs/30166600883): external `openapi_v2` failed because a Swagger v2 downgrade could not represent server descriptions. This is outside the changed path and resembles an existing integration failure, but it likewise remains an active gate until the new run proves it transient or we repair the responsible contract.

## 7. Current state and remaining gates

At submission time PR #1009 is draft, GitHub reports it mergeable, and the new full CI matrix for `201f66ddf` is in progress. It is not yet merge-ready. The campaign will continue through:

1. adjudication and formal recording of the `201f66ddf` ISR;
2. completion of every CI lane on this exact head;
3. diagnosis and repair of every red lane, including failures outside the immediate diff if reproducible;
4. a fresh exhaustive Overall Self-Review over the complete base-to-head consequence surface;
5. a final formal zero-finding review on the same green head;
6. ready-for-review transition, merge, issue-closure verification, branch/temp cleanup, and a renewed repository-wide discovery cycle.

### Round-1 conclusion

The takeover found no remaining production-code defect in the #1007/#1008 implementation, but it found and corrected one deterministic format failure, one material regression-test blind spot, and one benchmark-description mismatch. Because this round produced changes and CI is still running, **this is not the final clean review**.

samchon (commented):
Individual Self-Review ? `1dab1e6c0` (`fix(unplugin): derive watch inputs once per envelope generation`)

### Scope reviewed

The reviewer inspected the complete `219684bb5..1dab1e6c0` range and the affected watch-input derivation, graph traversal, cache-generation, benchmark, and regression-test paths. Read-only verification included the unplugin TypeScript check, the new probe regression test, an expanded `transformttsc` selection, and `git diff --check`.

### Result

No production defect was reproduced. The generation-scoped derivation and memoized graph/path identities preserve the expected dependency/watch semantics while removing the per-delivery macOS identity-probe multiplication that drove the #970 residual stall.

One P3 documentation-accuracy issue was found: `experimental/unplugin-perf/README.md` described Scenario C as connecting every module to its siblings, while the fixture connects each module only to its next sibling and then to `K` external declarations. The measured edge formula and benchmark behavior were already correct; only the topology prose was inaccurate.

### Adjudication

Accepted. Commit `201f66ddf` changes the description to ?next sibling.? This does not alter the fixture, gate, or reported measurements. The correction is covered by its own pending Individual Self-Review.

### Verification qualifications

- `pnpm --filter @ttsc/unplugin exec tsc --noEmit`: passed in the ISR.
- New probe regression test: passed in the ISR.
- Expanded `transformttsc` selection: 51 passed; 9 initially encountered `ENOENT` while concurrent repository-wide formatting was touching the shared worktree. The representative failing case passed alone, so those observations are not accepted as an implementation failure.
- Native macOS execution and a fresh complete performance-matrix rerun were not performed by this ISR; the campaign retains the benchmark evidence and CI/platform gates separately.

samchon (commented):
Individual Self-Review ? `201f66ddf` (`test(release): bind timeout assertions to yaml owners`)

### Finding

P3 ? the timeout test narrows mappings by indentation, but still searches the entire workflow. `selectYamlMapping(lines, "release:", 2)` can therefore select a future `on.release:` mapping instead of `jobs.release`. A valid in-memory mutation that adds `on.release.timeout-minutes: 15` and removes the actual `jobs.release.timeout-minutes` still satisfies the current assertion. The job-name scan also remains open from `jobs:` through EOF, so children of a later top-level mapping could be misclassified as jobs.

### Adjudication

Accepted. The test must first select the top-level `jobs:` mapping, select every job only inside that block, and select publish steps only inside `jobs.publish`. This closes the remaining ownership ambiguity instead of relying on globally unique header names.

The README ?next sibling? correction matches the generated graph, and the unplugin JSDoc change is a formatting-only repair with no semantic defect.

### Verification performed by the ISR

- `git diff --check 4915edf6..201f66dd`: passed.
- Targeted release-timeout test: passed before the adversarial mutation.
- Prettier check for all three changed files: passed.

A follow-up correction commit, its own mandatory ISR, and a restarted Overall Self-Review will follow.

samchon (commented):
Individual Self-Review ? `4ad0e2f91` (`test(release): scope timeout assertions to jobs mapping`)

### Result

Clean ? no sound findings in `201f66ddf..4ad0e2f91`.

### Evidence

- The test selects the top-level `jobs:` mapping before any job lookup and selects `publish:` only within that bounded block.
- Publish steps are searched only within `jobs.publish`; job assertions are searched only within `jobs`.
- Timeout keys must be direct children at the exact YAML indentation and must match exact budgets.
- The exact four-job set is derived only from the bounded `jobs` mapping, so children of later top-level mappings cannot masquerade as release jobs.
- `selectYamlMapping` stops at the next non-empty line with equal or shallower indentation.
- The adversarial cases are effective: a same-named mapping outside `jobs` cannot replace the real `jobs.release` timeout, and a trailing top-level mapping cannot pollute the job set.

### Verification

`pnpm --dir tests/test-ttsc start --include=bounds_publish_timeouts` passed with the real workflow and both adversarial assertions.

This satisfies the mandatory per-correction ISR gate for `4ad0e2f91`. Full CI and a fresh Overall Self-Review remain pending.

samchon (commented):
Individual Self-Review ? `c23e694d7` (`ci(nestia): pin the verified integration revision`)

### Result

The workflow correction is operationally sound, but the regression test had one accepted LOW-severity exactness gap.

### Workflow verification

- `.github/workflows/nestia.yml` parses with the expected `tarballs` and `nestia` jobs.
- `actions/checkout@v4` supports a full commit SHA in `ref` and a nested workspace-relative `path`; the public `samchon/nestia` repository does not require a separate PAT.
- `3b27e69b69dea3f102315042dce87c18d81be74a` is the merge of Nestia #1593.
- Nestia run `29959327169` completed all ten jobs successfully. Its tested PR head and the pinned merge have the same Git tree, so the merge pin preserves the verified source state.
- Nestia #1594's own SDK lane contains the same `openapi_v2` downgrade failure that appeared in ttsc's moving-master integration. The pin therefore removes an upstream-red input without weakening the ttsc-under-test tarball substitution.

No workflow-runtime, checkout, credential, or integration-path defect was found.

### Finding and adjudication

The test said it required exact repository and destination values but used substring membership. Values such as `repository: samchon/nestia-fork` or `path: experimental/nestia-copy` would have passed. Its single-line clone regex also missed a shell command whose Nestia URL followed a backslash continuation.

Accepted. Follow-up commit `d23b3b587` now requires each exact YAML line exactly once inside the named checkout-step mapping and normalizes CRLF/LF backslash continuations before rejecting a moving Nestia clone.

### ISR verification

- YAML parse: passed.
- Pinned revision and upstream green-tree provenance: verified.
- Targeted `pins_verified_upstream_revision` test: passed.
- Prettier and `git diff --check`: passed.

The follow-up correction has its own mandatory read-only ISR and a fresh same-head CI matrix. This review is not the final Overall Self-Review gate.

samchon (commented):
Individual Self-Review ? `d23b3b587` (`test(nestia): require exact checkout ownership`)

### Result

Two MEDIUM test-contract bypasses were reproduced and accepted. The reviewed commit improved exact-line matching and backslash continuations, but it still did not prove the values belonged to the checkout action's `with:` mapping and did not model YAML folded shell scalars.

### Findings

1. Replacing `with:` with `env:` left the exact `repository`, `ref`, and `path` lines present, so the test passed although `actions/checkout` received none of those inputs.
2. A valid folded command (`run: >`) with `git clone` and the Nestia URL on adjacent source lines executes as one shell command, but the raw-line clone regex stopped at the newline and missed it.

The ISR reproduced both bypasses in memory (`envOwnedFieldsStillPass: true`, `yamlFoldedCloneRejected: false`). LF/CRLF exact-line handling and the earlier backslash-continuation case did work as intended.

### Adjudication and correction

Accepted. Follow-up `f7db90bb6`:

- requires the named checkout step exactly once;
- requires the complete contiguous `uses: actions/checkout@v4` ? `with:` ? exact repository/ref/path mapping, with no prefix or ownership ambiguity;
- extracts inline, literal, and folded GitHub Actions `run:` values by indentation;
- normalizes LF/CRLF and shell backslash continuations;
- rejects active `git clone` commands rather than relying on the URL remaining on the same source line;
- embeds adversarial assertions for the `env:` ownership spoof and folded-scalar clone.

### Verification

- Targeted `pins_verified_upstream_revision` test: passed with both negative cases.
- `pnpm format`: passed on the complete correction state.
- `git diff --check`: passed.
- Staged-file audit: exactly the intended test file.

The follow-up commit has a separate mandatory ISR and a newly restarted same-head CI matrix. This remains an advisory commit review, not the final Overall Self-Review.

samchon (commented):
Individual Self-Review ? `f7db90bb6` (`test(nestia): bind checkout assertions to yaml owner`)

### Result

One MEDIUM false negative remained in the hand-written workflow command extractor and was accepted.

### Finding

The extractor handled inline values plus `>`/`|` block scalars, but returned immediately for a non-block `run:` value. YAML permits a plain scalar to continue on a more deeply indented following line. This valid workflow value:

```yaml
- run: git
    clone https://github.com/samchon/nestia.git experimental/nestia
```

is parsed and executed as one `git clone ...` command, while the custom extractor saw only `git`. The ISR reproduced the bypass against the exact commit.

### Adjudication

Accepted. Continuing to extend a partial YAML parser would keep creating representation gaps. Follow-up `2509b36c2` instead declares the already locked `yaml@2.9.0` package directly in `tests/test-ttsc`, parses the workflow with the standard YAML engine, and asserts the resulting semantic objects:

- the Nestia checkout owner is unique by step name, repository, or destination;
- the entire step is exactly the expected `actions/checkout@v4` plus `with.repository/ref/path` mapping;
- every parsed `run` string in every job is rejected if it contains an active `git clone` command;
- wrong `env:` ownership, a folded block scalar, and a continued plain scalar are embedded negative cases.

The lockfile correction contains only the three importer lines for `yaml`; unrelated optional-peer resolution churn from the initial local install was removed before commit.

### Verification

- `pnpm install --frozen-lockfile --offline --ignore-scripts`: passed; lockfile reported up to date.
- Targeted `pins_verified_upstream_revision` test, including all negative cases: passed.
- `pnpm --dir tests/test-ttsc exec tsc --noEmit`: passed.
- `pnpm format`: passed.
- `git diff --check` and exact three-file staged audit: passed.

The parser correction has its own mandatory ISR and a fresh same-head CI matrix. This is not the final Overall Self-Review.

samchon (commented):
Individual Self-Review ? `2509b36c2` (`test(nestia): parse workflow ownership from yaml`)

### Result

The standard YAML parser removed the prior folded/plain-scalar ambiguity and the dependency/lock changes are clean. Two MEDIUM regression-contract gaps remained and were accepted.

### Findings

1. YAML parsing does not execute shell continuation semantics. A literal `run: |` containing `git \` followed by `clone ...` parses as `"git \\\n  clone ..."`; the direct `/git\s+clone/` check misses it. The ISR reproduced the false negative under both LF and CRLF. This also regressed coverage that the direct parent had explicitly normalized.
2. Checkout owner uniqueness was calculated only from `jobs.nestia.steps`. A second step in another job with the same checkout name, `samchon/nestia` repository, or `experimental/nestia` destination was ignored. The pre-parser test had at least enforced the named marker across the complete workflow.

The ISR otherwise confirmed that `yaml` is a proper direct dev dependency, the lockfile change is limited to the three-line `tests/test-ttsc` importer entry, parser typing fails closed for the required mappings, and all parsed job `run` strings are traversed.

### Adjudication and correction

Accepted. Follow-up `15f728b4b`:

- flattens parsed step mappings from every workflow job before selecting checkout owners;
- requires exactly one owner by name, repository, or destination and proves that exact object belongs to `jobs.nestia`;
- normalizes shell backslash-newline continuations after YAML parsing;
- detects `git ... clone` within one shell command segment, including intervening Git options;
- adds explicit negative cases for a duplicate checkout in a different job and a `git \` + newline + `clone` command.

### Verification

- Targeted test, including wrong-owner, folded, plain, backslash-continuation, and cross-job-duplicate negatives: passed.
- `pnpm --dir tests/test-ttsc exec tsc --noEmit`: passed.
- `pnpm format`: passed.
- `git diff --check` and one-file staged audit: passed.

The follow-up has its own mandatory ISR and a newly restarted CI matrix. This remains a commit-level advisory review, not the final Overall Self-Review.

samchon (commented):
Individual Self-Review ? `15f728b4b` (`test(nestia): cover workflow-wide and shell ownership`)

### Result

Workflow-wide checkout ownership is correct. Two shell-matching findings remained: one MEDIUM false negative and one LOW false positive. Both were accepted.

### Findings

1. Backslash-newline continuation was normalized by inserting a space. Shells remove the pair; inserting a space changes token semantics. `git cl\` + newline + `one URL` executes as `git clone URL`, but the test normalized it toward `git cl one URL` and missed the clone. The existing negative case only split between complete `git` and `clone` tokens.
2. The broad `git ... clone` expression treated any later `clone` word in the same command segment as the subcommand. Harmless text such as `git config alias.clone ...` or `echo "git clone"` could fail despite performing no clone.

The ISR confirmed that owner selection is now workflow-wide and that the unique object is correctly required to belong to `jobs.nestia`. YAML parsing normalizes both LF and CRLF source scalars to LF.

### Adjudication and correction

Accepted. Follow-up `c6932349d`:

- removes each shell backslash-newline pair rather than replacing it with whitespace;
- detects `git` only at a command start or shell separator;
- permits recognized Git global options before requiring `clone` in the actual subcommand position;
- adds a token-internal `cl\` + newline + `one` negative case and an optioned `git -c ... clone` negative case;
- adds harmless `git config alias.clone` and quoted `echo "git clone"` positive cases.

### Verification

- Targeted test with all positive and negative forms: passed.
- `pnpm --dir tests/test-ttsc exec tsc --noEmit`: passed.
- `pnpm format`: passed.
- `git diff --check` and one-file staged audit: passed.

The follow-up has its own mandatory ISR and a fresh same-head CI matrix. This is not the final Overall Self-Review.

samchon (commented):
Individual Self-Review ? `c6932349d` (`test(nestia): preserve shell token semantics`)

### Result

The token-continuation correction worked, but the shell regex remained fundamentally unsound: one HIGH class of executable false negatives and one MEDIUM class of harmless false positives were reproduced. Both are accepted.

### Findings

- Active clones still bypassed the matcher through valid shell contexts or unenumerated Git options: `(git clone url)`, `! git clone url`, `if git clone url; then ...`, `git -P clone url`, `git --no-optional-locks clone url`, and quoted-argument `git -C "some dir" clone url`.
- Separators inside quoted/comment text were mistaken for command boundaries, rejecting harmless `echo "never; git clone url"` and `# never; git clone url`.

The ISR confirmed that LF token-internal continuations are correctly removed and that YAML parsing normalizes CRLF scalar input to LF. The defect is not another missing regex alternative; it is using a regex as a shell parser.

### Adjudication and correction

Accepted. Follow-up `976d6c3d9` removes the shell-command blacklist and all representation-specific shell probes. The regression test now asserts only the durable semantic structure supplied by the YAML parser:

- exactly one workflow-wide owner identified by the required step name, `samchon/nestia` repository, or `experimental/nestia` destination;
- that owner is exactly the expected `actions/checkout@v4` step and exact `with.repository/ref/path` mapping;
- the same object belongs to `jobs.nestia`;
- moving ownership to `env:` and adding an owner in a different job both fail.

This is sufficient for the actual regression: replacing the pinned checkout with the former moving clone removes the required exact owner and fails the test. It deliberately does not attempt to prove arbitrary shell scripts harmless or malicious.

### Verification

- Targeted structural test: passed.
- `pnpm --dir tests/test-ttsc exec tsc --noEmit`: passed.
- `pnpm format`: passed.
- `git diff --check` and one-file staged audit: passed.

The simplification has its own mandatory ISR and a newly restarted same-head CI matrix. This is not the final Overall Self-Review.

samchon (commented):
Individual Self-Review ? `976d6c3d9` (`test(nestia): assert checkout structure without shell parsing`)

### Result

CLEAN ? no sound findings.

### Evidence

- The parsed workflow requires exactly one owner across all jobs by the expected step name, `samchon/nestia` repository, or `experimental/nestia` destination.
- That owner must deep-equal the complete expected `actions/checkout@v4` step and exact `with.repository/ref/path` mapping.
- Object identity proves the unique owner belongs to `jobs.nestia` rather than merely existing elsewhere.
- The negative cases prove that moving inputs under `env:` cannot satisfy action ownership and that an otherwise identical owner in a second job violates workflow-wide uniqueness.
- Removing the shell blacklist is principled: it had attempted to classify arbitrary shell text and rejected unrelated commands, while replacing the required pinned checkout with the former moving clone necessarily removes the exact structural owner and still fails this test.
- The test documentation matches the assertions that remain.

### Verification

- Targeted `pins_verified_upstream_revision` feature test: passed.
- `pnpm exec tsc -p tests/test-ttsc/tsconfig.json --pretty false`: passed in the ISR.
- `git diff --check`: passed.
- Read-only review left the worktree clean.

This satisfies the mandatory correction ISR for `976d6c3d9`. Full same-head CI and a fresh exhaustive Overall Self-Review remain pending.

samchon (commented):
## Final Overall Self-Review — CLEAN

**Reviewed head:** `976d6c3d9b4a935be4f61be4e3eca6d51d3fdde3`

I completed the final solo Overall Self-Review after all correction commits had landed and after every required check had completed on this exact head. I found **no remaining correctness, compatibility, performance, workflow, test-integrity, or maintainability issue that warrants another change in this pull request**.

### 1. Incident findings and scope decision

The two user-visible symptoms investigated by this campaign are related only at the level of “a long-running operation appears stuck”; they have different owners and root causes.

1. Release run `30161890750` did not freeze. It completed successfully on 2026-07-25 in about 35 minutes. The publish job accounted for roughly 28.5 minutes, after which smoke and release stages continued normally. The verified defect was operational: the publish steps and their jobs inherited GitHub Actions' very large default upper bound, so an actually wedged marketplace or npm publish would provide no timely failure signal.
2. The residual #970 report is real and independent of the release workflow. A graph-bearing typia result envelope caused each transformed module delivery to derive watch inputs by walking and canonicalizing the complete reference graph again. On macOS, the repeated `existsSync` and `realpathSync.native` work made the path syscall-bound and scaled as O(deliveries × graph edges). That is the “never completes” behavior tracked precisely as #1007.

This PR therefore closes #1007 and #1008 as one verified campaign cycle without pretending they share an implementation root cause.

### 2. Production-code review

I re-read the complete `packages/unplugin/src/core/transform.ts` change against the prior behavior and traced both the steady-state and failure paths.

The implementation now creates derivation state once per graph-bearing result envelope and retains it in a `WeakMap`, so its lifetime follows the envelope rather than the process. The state lazily builds graph indexes, memoizes normalized/real path identities, constructs declared-input sets only when required, and indexes output/dependency relationships for per-file watch-input delivery. Subsequent module deliveries query this shared state instead of rescanning the full graph.

I specifically checked these semantic invariants:

- watch-input ordering and first-seen deduplication remain stable;
- cyclic dependency graphs terminate and do not duplicate output;
- malformed or incomplete graph members retain the existing fallback behavior;
- exception, transform-failure, volatile, and completeness paths do not incorrectly reuse successful state;
- temporary tsconfig artifacts remain excluded;
- output-first and dependency-derived lookup behavior retains the established first-match semantics;
- path identity handles Windows separators/case behavior and macOS realpath canonicalization without changing the public watch set;
- concurrent deliveries share immutable/lazily memoized envelope state without process-global leakage;
- the `WeakMap` does not extend the lifetime of compilation envelopes.

The optimization is therefore scoped to eliminating repeated derivation work; it does not weaken file freshness or alter emitted transform results.

### 3. Release workflow review

I reviewed every new `timeout-minutes` owner in `.github/workflows/release.yml` rather than only checking for timeout-like text.

- publish job: 60 minutes;
- marketplace publish step: 10 minutes;
- npm publish step: 20 minutes;
- each downstream smoke/release job: 15 minutes.

The limits have deliberate headroom over the observed healthy run and fail at the narrowest useful owner. The structural regression test parses YAML and verifies both value and ownership, preventing a timeout from silently moving to a sibling step/job where it would no longer bound the intended operation.

### 4. CI repair discovered during this PR

The initial Nestia failure was not caused by this diff. The same `openapi_v2` failure reproduced from ttsc `master` in run `30012034569`, and Nestia's own SDK lane after Nestia #1594 failed equivalently. I nevertheless repaired it here because this campaign requires every red CI result to be understood and made deterministic before merge.

`.github/workflows/nestia.yml` no longer consumes moving `nestia/master`. It checks out Nestia commit `3b27e69b69dea3f102315042dce87c18d81be74a`, the last merge whose upstream run (`29959327169`) passed all ten jobs. The final test uses the real YAML parser and asserts the exact checkout action, repository, ref, and path; it also enforces one workflow-wide owner and verifies that the owner belongs to the intended `nestia` job. Negative cases cover a misplaced `with` mapping and a cross-job duplicate.

Earlier review correctly rejected attempts to infer shell semantics with regexes and a hand-written YAML/run parser. The final correction removed that brittle surface and retained the enforceable structural contract. This is the smallest durable fix for the actual moving-upstream failure mode.

### 5. Regression and performance evidence

The new Scenario C fixture reproduces the graph-envelope path and counts filesystem probes under simulated Darwin behavior. On the pre-fix implementation, 99 deliveries produced approximately 6.7k, 12k, and 22.8k probes per module as the graph grew, taking roughly 76–95 seconds. On the final implementation, the same derivation is approximately flat at 19 probes per module and roughly 100 ms. The test gates the scaling invariant rather than a machine-specific wall-clock threshold.

Scenario D separately measures Vite serve-mode external-input validation. Its linear behavior is documented and intentionally not folded into #1007 because it enforces the existing #980 freshness contract; changing that policy would be a separate design decision.

Final local verification included:

- `pnpm install --frozen-lockfile --offline --ignore-scripts`;
- `pnpm format`;
- `pnpm --dir tests/test-ttsc exec tsc --noEmit`;
- `pnpm --filter @ttsc/unplugin exec tsc --noEmit`;
- `pnpm --dir tests/test-ttsc start --include=pins_verified_upstream_revision,bounds_publish_timeouts`;
- `pnpm --dir tests/test-unplugin start --include=bounds_watch_derivation_probes_per_module`;
- `git diff --check origin/master...HEAD`.

All passed. The lockfile change was also reduced to the intentional direct `yaml@2.9.0` test dependency; unrelated optional-peer churn was not retained.

### 6. Review history and corrections

Every coherent implementation or correction commit received the campaign-required read-only Individual Self-Review. The sequence was:

- `1dab1e6c0`: #1007 implementation;
- `4915edf6b`: #1008 workflow implementation;
- `201f66ddf`, `4ad0e2f91`: release test/documentation and owner-scope corrections;
- `c23e694d7`, `d23b3b587`, `f7db90bb6`, `2509b36c2`, `15f728b4b`, `c6932349d`: successive Nestia pin-test hardening in response to concrete review findings;
- `976d6c3d9`: final simplification to the real YAML structural contract, whose Individual Self-Review returned CLEAN.

The reviews are preserved on this PR as formal COMMENT reviews. In particular, the final correction's review is `#pullrequestreview-4780739984`; the preceding high/medium findings that motivated it are recorded in `#pullrequestreview-4780732673`. No advisory finding was silently waived.

### 7. Final merge gate

At the time of this review:

- local HEAD, `origin/fix/watch-derivation-scaling`, and PR head all equal `976d6c3d9b4a935be4f61be4e3eca6d51d3fdde3`;
- the worktree is clean;
- GitHub reports the PR as mergeable with a CLEAN merge state;
- **64/64 checks succeeded on this exact head**, including macOS watch coverage, Windows-Go, unplugin, graph/graph-bench, native corpora, workflow contract tests, and the pinned Nestia integration (`nestia` job: 47m41s).

**Disposition: CLEAN. Ready to leave draft and merge.**

PR #1026, opened 2026-07-27
fix(release): make Go SDK extraction portable on Windows

samchon:
## Intent

Own the complete accepted fourth-cycle campaign scope and make Go SDK
extraction portable when a Windows root build selects Git for Windows GNU tar.

## Scope

- #1025
- Invoke ambient tar from the archive cache with validated relative archive and
  destination operands, removing Git for Windows GNU tar's `D:` remote-host
  ambiguity without selecting a tar vendor.
- Preserve checksum-first authentication, stale-cache recovery, Go-binary
  validation, marker-last publication, and the Node zip extraction path.
- Exercise the complete authenticated extraction transaction, including paths
  with spaces and real GNU tar/bsdtar extraction.

This draft pull request owns every accepted issue from the cycle. No other
campaign issue remains unclaimed.

## Deferred

None.

## Verification

The discovery gate closed after a complete fresh empty round against
`bc3e2e9950fa71c5a1f328699e09766e46ede98e`.

- The previously failing real `@ttsc/darwin-arm64` Windows package build
  completed all binaries and embedded the verified Go 1.26.4 SDK with Git GNU
  tar.
- Exact-head `features/platform`: 17/17 passed.
- Exact-head test-ttsc typecheck passed.
- The focused regression passed with both Git GNU tar 1.35 and Windows system
  bsdtar 3.8.4.
- All thirteen workspace typechecks passed before the later test-only review
  corrections.
- CJS syntax, diff whitespace, project layout, and all 83 test-owner assertions
  passed.
- Ordinary exact-head pull-request CI is the remaining independent merge gate.

## Review

Every pushed implementation or correction commit received its required
read-only Individual Self-Review. Three reviews found and drove test-integrity
improvements; the final exact-head review is clean:

- Initial review:
  https://github.com/samchon/ttsc/pull/1026#pullrequestreview-4783476258
- Transaction-owner review:
  https://github.com/samchon/ttsc/pull/1026#pullrequestreview-4783496978
- Wiring/cleanup review:
  https://github.com/samchon/ttsc/pull/1026#pullrequestreview-4783497191
- Final clean Individual Self-Review:
  https://github.com/samchon/ttsc/pull/1026#pullrequestreview-4783518011
- Final clean Overall Self-Review:
  https://github.com/samchon/ttsc/pull/1026#pullrequestreview-4783522610

PR #1030, opened 2026-07-27
fix: close final watch, documentation, and CI runtime gaps

### LINKED ISSUE #1027: docs(website): stop advertising the removed public ttsc transform command
# Documentation advertises the removed public `ttsc transform` command

## Summary

The plugin-author end-to-end guide tells users to run `npx ttsc transform`, but
the public CLI deliberately rejects that command. Two related reference pages
also describe the source-to-source path as `ttsc transform`, which makes an
internal plugin-protocol verb look like a supported CLI surface.

The documentation should use the supported `TtscCompiler.transform()` API for
the runnable source-to-source example and name the programmatic or bundler host
when explaining source-to-source compatibility failures.

## Reproduction

On current `master`
(`79c76117fae71e582bdcdfa042f3db84cfe577c2`):

```console
> node packages/ttsc/lib/launcher/ttsc.js transform
ttsc: unknown command "transform"
ttsc: run "ttsc --help" to see supported commands
```

The process exits with status 2.

The conflicting instruction is in
`website/src/content/docs/development/authoring/end-to-end.mdx`:

````md
Run the transform path:

```bash
npx ttsc transform
```
````

The same removed surface is named in:

- `website/src/content/docs/development/reference/pitfalls.mdx`
- `website/src/content/docs/development/concepts/protocol.mdx`

Both say a source-to-source compatibility error fires during
`` `ttsc transform` ``.

## Product contract

`tests/test-ttsc/src/features/compiler/test_ttsc_rejects_unsupported_transform_command.ts`
explicitly records that `transform` existed only in earlier CLI drafts, was
removed before the public release, and must produce an `unknown command`
failure.

The supported source-to-source surfaces are:

- `new TtscCompiler({ cwd }).transform()` for Node embedders;
- bundler adapters that invoke the transform host internally; and
- the internal `transform` plugin-protocol verb that those hosts send to a
  plugin binary.

`website/src/content/docs/development/walkthroughs/lint.mdx` already states this
boundary correctly: there is no user-facing `ttsc transform` command.

## Consequence

A plugin author who follows the end-to-end guide reaches a guaranteed command
failure after completing the longest tutorial on the site. The reference pages
reinforce the wrong mental model by presenting an internal protocol verb beside
the real `ttsc build` and `ttsc check` commands.

This is not merely an obsolete spelling. There is no CLI replacement. The
correct replacement crosses from the CLI surface to the public programmatic API
or a bundler adapter.

## Scope

1. Replace the end-to-end guide's failing command with a runnable
   `TtscCompiler.transform()` example against the consumer project.
2. Show how to validate the success/failure result before reading the
   `typescript["src/main.ts"]` output.
3. Rewrite the two compatibility explanations to name
   `TtscCompiler.transform()` and bundler-hosted transforms rather than the
   nonexistent CLI command.
4. Preserve the distinction between the public programmatic surface and the
   internal plugin-protocol `transform` verb.
5. Link the example to the programmatic API reference so plugin authors can
   continue from the smallest working verification.

## Acceptance criteria

- No official guide presents `ttsc transform` or `npx ttsc transform` as a
  user-facing command.
- The end-to-end source-to-source verification is directly runnable with the
  dependencies the guide already installs.
- The example checks the discriminated result before accessing
  `result.typescript`.
- The expected transformed `src/main.ts` text remains explicit.
- The website documentation formats and builds successfully.
- The existing CLI rejection regression remains green.

## Duplicate and boundary check

Repository-wide issue and pull-request searches for `"ttsc transform"` find
historical API, protocol, cache, and transformation work, but no issue or open
pull request for this documentation contradiction. This issue does not add a
CLI command and does not change the plugin protocol.

### LINKED ISSUE #1028: fix(ttsc): reconcile project inputs after watcher registration
# Project-input watchers can miss the first change after registration

## Summary

`WatchTopology.setProjectInputs()` snapshots declared files and globs before it
registers their recursive filesystem watchers. A file can be created or changed
immediately after the method returns while the backend is still starting. If
the backend drops that first event, no later reconciliation observes the new
population.

This now produces intermittent red `watch (macos-latest)` jobs on otherwise
identical source trees. The fix should add one coalesced post-registration
reconciliation after the caller's current synchronous turn and prove the
behavior with a deterministic watcher that swallows its first event.

## Current failures

Exact merged master
`79c76117fae71e582bdcdfa042f3db84cfe577c2` failed:

```text
test_watch_topology_matches_remaining_tsgo_output_semantics: AssertionError
AssertionError [ERR_ASSERTION]: expected a project change after 0
```

Job:
https://github.com/samchon/ttsc/actions/runs/30237709237/job/89888563224

A recent earlier master run failed a different test at the same lifecycle:

```text
test_watch_topology_models_effective_adjacent_and_incremental_outputs: AssertionError
AssertionError [ERR_ASSERTION]: expected a project change after 0
```

Job:
https://github.com/samchon/ttsc/actions/runs/30186003847/job/89750729333

The intervening and neighboring macOS jobs pass. In particular, PR #1026 head
`9e4bf1aaf5e537a7c3207b7a17661cc85cbfc782` passed the watch job, and its tree
hash is exactly the same as the failing merge commit:

```text
7f45ddb8cbfd5ea776db650c6163ea90443b90c5
```

PR #1026 changed only Go SDK packaging and platform tests, not watch code.

## Cause

The current sequence in
`packages/ttsc/src/launcher/internal/watchTopology.ts::setProjectInputs` is:

1. normalize and publish the new declarations;
2. collect the current project-input matches;
3. fingerprint those matches;
4. install the recursive `fs.watch` roots; and
5. return.

Both failing tests synchronously create an expected project input immediately
after step 5. The positive assertion then waits for the watcher callback. A
backend event is the only path that calls `refreshProjectInputs()`.

macOS recursive `fs.watch` uses FSEvents. The watcher object can be returned
before the first filesystem notification is reliably delivered. If that
startup notification is lost, the baseline remains unchanged and nothing
causes a later scan. Increasing the assertion deadline cannot recover an event
that will never arrive.

The same window can affect a real watch session when a plugin publishes a new
missing file or empty glob and another tool materializes it in the same turn.

## Required behavior

After synchronizing a newly published project-input topology, schedule one
coalesced reconciliation after the current synchronous turn:

- compare the live population with the fingerprint captured before watcher
  registration;
- report a real file, glob-membership, reload, or compiler-membership change
  through the existing classification path;
- allow a real watcher callback and the reconciliation to race without
  reporting the same change twice;
- use the latest declarations if multiple publications coalesce;
- do nothing after `close()`; and
- do not create a timer or recurring poll that can keep the process alive.

The existing fingerprint update in `refreshProjectInputs()` provides the
deduplication boundary: whichever path reconciles first updates the baseline,
and the other becomes a no-op.

## Regression

Add a deterministic test that:

1. replaces `fs.watch` with a valid watcher object that never invokes its change
   callback;
2. publishes a missing project-input file;
3. creates that file synchronously before yielding;
4. observes exactly one `kind: "project"` change from the post-registration
   reconciliation;
5. proves a close before the scheduled reconciliation emits nothing; and
6. restores `fs.watch` in `finally`.

Keep the existing macOS matrix coverage. The test must not rely on making an
actual FSEvents race occur.

## Acceptance criteria

- The deterministic swallowed-first-event regression passes.
- Existing output-suppression tests still distinguish expected project inputs
  from compiler outputs.
- Repeated or unchanged `setProjectInputs()` calls coalesce rather than queueing
  unbounded work.
- A real watcher event and the post-registration reconciliation produce one
  logical change.
- `close()` prevents scheduled callbacks from emitting and all watcher handles
  still drain.
- Windows, macOS, and Linux watch jobs pass on the exact implementation head.

## Duplicate and boundary check

Repository issue and pull-request searches found no existing report for this
registration-to-first-event gap. This issue does not replace event-driven
watching with polling and does not weaken output suppression. It closes one
bounded handoff between the initial snapshot and the watcher backend.

### LINKED ISSUE #1029: ci: move workflow actions onto their Node 24 majors
# CI actions still declare the retired Node 20 runtime

## Summary

Most ttsc workflows pin action majors whose metadata declares
`runs.using: node20`. GitHub-hosted runners now force those actions to run on
Node 24 and print a deprecation warning in every affected job.

The repository should use the current Node-24 action majors directly instead of
depending on a runner override. The user has explicitly excluded
`.github/workflows/nestia.yml`; this issue leaves that workflow untouched.

## Current warning

Current jobs report:

```text
Node.js 20 is deprecated. The following actions target Node.js 20 but are being
forced to run on Node.js 24: actions/checkout@v4, actions/setup-go@v5,
actions/setup-node@v4, pnpm/action-setup@v4.
```

The warning appears during job setup and cleanup. One example is the exact
master watch job:
https://github.com/samchon/ttsc/actions/runs/30237709237/job/89888563224

## Repository state

The in-scope workflows use:

- `actions/checkout@v4`
- `actions/setup-node@v4` (with some `release.yml` jobs already on v6)
- `actions/setup-go@v5`
- `pnpm/action-setup@v4`

Affected in-scope files are:

- `.github/workflows/build.yml`
- `.github/workflows/bun.yml`
- `.github/workflows/experimental.yml`
- `.github/workflows/plugin-cache.yml`
- `.github/workflows/release.yml`
- `.github/workflows/source-map.yml`
- `.github/workflows/test.yml`
- `.github/workflows/typia.yml`
- `.github/workflows/vscode.yml`
- `.github/workflows/website.yml`

`.github/workflows/nestia.yml` is explicitly out of scope.

`.github/dependabot.yml` monitors only the npm ecosystem, so no GitHub Actions
dependency update owns this drift.

## Upstream contract

Verified from the action repositories on 2026-07-27:

| Action | Current release | Declared runtime |
| --- | --- | --- |
| `actions/checkout` | `v7.0.1` | `node24` |
| `actions/setup-node` | `v7.0.0` | `node24` |
| `actions/setup-go` | `v7.0.0` | `node24` |
| `pnpm/action-setup` | `v6.0.9` | `node24` |

The floating major tags `@v7` and `@v6` are the upstream-supported update
surfaces. Existing tool inputs remain valid: the repository still chooses Node
24, Go from `packages/ttsc/go.mod`, and pnpm 10.6.4 explicitly.

## Consequence

The current workflows do not run with the runtime declared by their pinned
action versions. They rely on GitHub's temporary compatibility override. That
creates noisy CI and leaves action dependencies outside the runtime line their
maintainers currently test.

This issue does not change the project Node, Go, or pnpm versions. It aligns the
action implementation runtime with the already-selected GitHub runner runtime.

## Required change

In every in-scope workflow:

1. replace `actions/checkout@v4` with `actions/checkout@v7`;
2. replace all `actions/setup-node` uses with `actions/setup-node@v7`;
3. replace `actions/setup-go@v5` with `actions/setup-go@v7`;
4. replace `pnpm/action-setup@v4` with `pnpm/action-setup@v6`;
5. preserve every existing `with`, cache, version, matrix, condition, and shell
   setting; and
6. do not modify `nestia.yml`.

## Acceptance criteria

- No in-scope workflow references the four Node-20 action majors.
- Every replacement action's metadata declares `node24`.
- Existing explicit Node, Go, and pnpm tool-version selections are unchanged.
- Workflow paths, triggers, timeouts, matrices, conditions, and permissions are
  unchanged.
- YAML formatting and repository workflow checks pass.
- Exact-head CI no longer prints the Node-20 forced-runtime warning for the
  migrated workflows.
- `.github/workflows/nestia.yml` has no diff.

## Duplicate and boundary check

There is no open issue or pull request for this migration. This issue does not
upgrade arbitrary third-party actions, change release credentials, or add
GitHub Actions Dependabot. It only replaces the four action majors whose
declared Node runtime is already being overridden by GitHub.

samchon:
## Intent

Close the final documentation, project-input watcher, and GitHub Actions
runtime gaps discovered by the fifth repository-wide campaign cycle before the
v0.23.0 release.

Closes #1027
Closes #1028
Closes #1029

## Discovery gate

The integrated baseline was
`79c76117fae71e582bdcdfa042f3db84cfe577c2`, the merge of PR #1026.
Discovery restarted from that merged tree and covered all 6,670 tracked files,
45 manifests, 19 public packages, 83 executable test owners, 13 typechecked
test workspaces, workflows, release wiring, process and watcher lifecycles,
platform paths, generated contracts, dependencies, guides, and repository
integrity.

Four complete fresh rounds were run:

1. Round 1 found the removed public `ttsc transform` command in three official
   guide surfaces and published #1027.
2. Round 2 traced intermittent exact-tree macOS watch failures to the
   snapshot-to-watcher startup gap and published #1028.
3. Round 3 verified that the common workflow actions still declared Node 20
   while GitHub forced them onto Node 24, then published #1029.
4. Round 4 repeated the full repository scope and was empty. Its three raw
   candidates were rejected as an intentional consumer fixture, a
   Windows-path bug in an audit probe, and harmless dangling local Git objects.

Implementation began only after that empty round. The abandoned
`fix/website-typia-lock-resolution` branch was rejected as superseded by the
integrated exact-pin contract. A speculative curl deadline was rejected because
the owning release job already has a hard deadline and no observed unbounded
transfer justified a shorter arbitrary cutoff.

## #1027 — supported transform surface

The end-to-end plugin guide now states that the CLI has no public transform
subcommand and provides a runnable `transform.mjs` example using the named
`TtscCompiler` export and `TtscCompiler.transform()`. The example handles
`exception` and diagnostic `failure` results before reading the project-relative
`typescript["src/main.ts"]` output.

The protocol and pitfalls references now attribute source-to-source
shared-host errors to bundler adapters or `TtscCompiler.transform()`. The
internal `transform` protocol verb remains distinct from public CLI commands.
Overall Self-Review then found four older dotted references that named a
nonexistent `ttsc.transform()` namespace method. The lint walkthrough and the
remaining protocol passages now consistently name the exported
`TtscCompiler.transform()` class method.

Verification included the unsupported-command regression, a configured
source-plugin programmatic transform, MDX formatting, link and export checks,
and the complete website production build: 89 static pages, RSS and Typia
dependency-graph tests, Pagefind indexing, and sitemap generation.

Individual Self-Review chronology:

- `7beefc021`: clean review of the primary #1027 implementation.
- `2332e30d8`: clean review of the Overall Self-Review correction.

## #1028 — bounded watcher-registration handoff

`WatchTopology` now queues one coalesced microtask after project-input watcher
registration. A map-identity comparison proves that a project or link watcher
was actually installed or replaced before scheduling the scan; an unchanged
snapshot republication does not behave as polling. The scan compares the live
population with the pre-registration fingerprint and uses the ordinary event
classifier, so a backend event and the handoff scan race through one exact-once
deduplication boundary. The queued work uses the latest publication, carries an
exact path when one member changed, creates no timer or recurring poll, and
stays silent after close.

The final implementation also handles the consequence surface found during
review:

- One unavailable external watcher no longer disables reconciliation for
  independently healthy roots.
- Existing safe fallback watchers remain live during the handoff scan, so the
  scan neither retries the failed preferred root nor forms a recovery loop.
- Declarations are re-normalized immediately before the scan. A missing
  symlink or junction that materializes in the handoff window therefore adds
  both the declared owner and its new physical target owner before later target
  events can occur.
- A one-member filename-less delta retains its exact path for reload,
  compiler-membership, and compiler-output classification.
- The public protocol guide documents the one bounded handoff scan while
  preserving the no-polling contract for later observation.

The deterministic regression replaces `fs.watch` with valid fake watchers and
proves five scenarios:

1. a swallowed startup event reports the synchronous input once;
2. a real backend event winning the race does not duplicate;
3. close cancels the queued notification and drains every handle;
4. a healthy project root still reconciles beside an `ENOSPC` external lane;
5. a newly materialized junction registers its physical target owner, and a
   second target event is observed through that exact watcher;
6. an unchanged republication after the handoff does not rescan an event that
   the established backend did not deliver.

Validation passed the focused tests, the existing rejected-root fallback chain,
the complete 34-test watch subtree on Windows, the ttsc build, and test
TypeScript compilation. Exact-head CI owns Windows, macOS, and Linux watch
proof.

Individual Self-Review chronology:

- `94874d94f`: accepted the global unavailable-root coupling, contradictory
  protocol prose, and missing numbered test scenarios.
- `0091791e3`: accepted the missing physical owner after a handoff-time
  symlink/junction transition.
- `27e7f74c2`: clean review of the physical-owner correction.
- `f2e0244a5`: clean review of the physical-path assertion correction exposed
  by Windows exact-head CI.

## Exact-head macOS red-CI correction

At `f2e0244a5`, `watch (macos-latest)` failed in the pre-existing
config-deletion recovery test even though the new project-input regression
passed. The test deletes `tsconfig.json` synchronously after
`WatchTopology.refresh(false)`. macOS FSEvents can return the compiler watcher
before its first event is deliverable, leaving the deletion and parse failure
unobserved.

Commit `0fc1097d2` extends the same bounded handoff invariant to newly
registered compiler file and directory watchers. It schedules only when a new
watcher object was installed, compares tracked files with their
pre-registration stamps in one coalesced microtask, and uses that same stamp as
the race boundary when a real event wins. Close remains silent; no timer or
recurring poll exists.

Individual and Overall review found two further consequences before merge. A
new source admitted by a tsconfig include pattern was not yet in the tracked
stamp set, and a POSIX atomic replacement could leave the per-file watcher on
the old inode after the first recovered notification. Commit `3f40f5ad4`
therefore performs one topology resolve after a new watcher registration,
rebinds changed existing per-file owners before reporting, and leaves missing
entries to the still-live parent directory watcher.

Its Individual Self-Review found two more boundaries. `mtimeMs:size` alone
cannot distinguish a replacement inode with identical content metadata, and a
per-file rearm must not schedule the full membership resolve reserved for a new
directory watcher. The final correction stores content and `dev:ino` ownership
separately, rearms an owner change without inventing a content notification,
uses stamp-only reconciliation for a new file owner, and preserves the
originating skip-unobserved context when a directory owner genuinely needs a
membership resolve. A combined config deletion and source replacement now
reports one config change, one compiler change, and one parse error rather than
repeating the failed resolve.

The deterministic regression proves a swallowed config deletion, a swallowed
new Program member, a real source event winning without duplication, a
same-stamp POSIX replacement followed by an in-place edit, a source rearm beside
a failed config refresh, close cancellation, and exact handle drainage.

Exact-head Windows CI also exposed a short-path assertion error in the new
test: backend registration uses the physical long path while notifications
retain the declared `RUNNER~1` spelling. The regression now compares each side
in its owning path domain. A later macOS exact head proved that compiler
topology canonicalizes the `/var` temp alias to `/private/var` on POSIX. The
final regression therefore expects lexical compiler paths on Windows and
physical compiler paths on POSIX without normalizing the actual assertion
value. The prior exact head's `corpus-misc` lane then exposed an unconditional
project-input republication scan as an idle
`config (unnamed)` rebuild; restricting scans to actual new watcher objects
closes that red-CI path.

Local validation passed the complete 34-test Windows watch subtree, test
TypeScript compilation, the ttsc package build, full repository formatting,
and diff checks. A Windows-only focused native-corpus rerun exposed a later
backend sequence that an experimental parent-event/digest correction did not
close; that experiment was removed rather than adding unproven hashing cost or
weakening named-event semantics. Exact-head Linux CI owns the issue-reproducing
corpus proof. Individual review accepted the new-member, POSIX inode,
same-stamp owner, and redundant membership-refresh findings. The immutable
path-domain correction received a clean Individual Self-Review at
`f3684251a`. The fresh Overall Self-Review of the complete 17-file base-to-head
diff is clean:
https://github.com/samchon/ttsc/pull/1030#pullrequestreview-4784899208.

## #1029 — actions on declared Node 24 runtimes

Only the four issue-owned action majors changed in the 10 in-scope workflows:

- `actions/checkout@v4` to `@v7`: 14 uses;
- `actions/setup-node@v4` or `@v6` to `@v7`: 14 uses;
- `actions/setup-go@v5` to `@v7`: 11 uses;
- `pnpm/action-setup@v4` to `@v6`: 13 uses.

Official upstream metadata declares `runs.using: node24` for every selected
major. Parent and child YAML were parsed and structurally compared after
normalizing those 52 scalar values. Triggers, path filters, permissions,
matrices, conditions, timeouts, shells, commands, caches, explicit Node 24, Go
module selection, pnpm 10.6.4, trusted publication, and release dependency
ordering are unchanged.

`.github/workflows/nestia.yml` is explicitly outside this campaign and its blob
is byte-identical across the implementation commit.

Individual Self-Review result: clean.

## Integrated verification

Local exact-head evidence:

- complete `features/watch`: 34/34 passed;
- all 13 test workspaces typechecked;
- ttsc package build passed;
- dependency audit passed with 0 high and 0 critical advisories;
- generated flag artifacts and all 15 upstream flag kinds agree;
- project layout accounts for every public package;
- all 10 changed workflows parse as YAML and pass Prettier;
- changed source, tests, and MDX pass Prettier and whitespace checks;
- `nestia.yml` has zero diff;
- tracked worktree and local/remote head are synchronized.

The final Overall Self-Review is clean. All 62 in-scope GitHub checks passed on
exact head `f3684251a9edc2134724c568e4960354dd2d85e2`, including the three OS
watch lanes and the issue-reproducing native corpus lane, with no failed,
cancelled, timed-out, or action-required check. Annotation inspection found no
Node 20 runtime warning in any changed workflow. `nestia.yml` is not a gate by
explicit owner direction.

Formal exact-head CI audit:
https://github.com/samchon/ttsc/pull/1030#pullrequestreview-4785036692.

## Deferred

None.


packages/ttsc/src/launcher/internal/watchTopology.ts
```diff
@@ -671,11 +674,40 @@ export class WatchTopology {
...
+   * `setProjectInputs()` would otherwise depend entirely on that startup event.
+   * The ordinary fingerprint update makes this scan and a real backend event
+   * race safely: whichever arrives first records the new population and the
+   * other becomes a no-op.
+   */
+  private scheduleProjectInputPostRegistrationReconciliation(): void {
+    if (
+      this.closed ||
+      this.projectInputRecoveryScheduled ||
+      this.projectInputUnobservedWatchRoots.size !== 0 ||
```
samchon: **P2 — one unavailable root suppresses every healthy root.** This map is topology-wide, so an exhausted external lane prevents the post-registration scan for successfully registered roots. Preserve unavailable-root reporting independently, but let healthy registrations complete their bounded handoff reconciliation. Add a mixed healthy/rejected-root regression.

tests/test-ttsc/src/features/watch/test_watch_topology_reconciles_project_inputs_after_registration.ts
```diff
@@ -0,0 +1,161 @@
...
+import fs from "node:fs";
+import path from "node:path";
+
+import {
+  type WatchInputChange,
+  WatchTopology,
+} from "../../../../../packages/ttsc/lib/launcher/internal/watchTopology.js";
+
+/**
+ * Verifies project-input publication closes the snapshot-to-watcher handoff.
```
samchon, line 12: **P3 — mandatory test scenarios are missing.** The headline and rationale are present, but the repository test contract also requires a concise numbered scenario list. Include the swallowed event, backend race, close, and mixed unavailable-root behaviors.

samchon (commented):
## Individual Self-Review — `94874d94f`

Reviewed exactly `aaa577db3..94874d94f` for issue #1028.

### Landed

- Added one coalesced microtask reconciliation after project-input watcher registration.
- Reused the existing fingerprint transition as the deduplication boundary between the reconciliation and a real backend event.
- Preserved an exact changed path when a filename-less reconciliation identifies exactly one changed input, retaining reload, compiler-membership, and output-suppression classification.
- Added deterministic coverage for a swallowed startup event, repeated publication coalescing, a backend event winning the race, exact-once notification, and close-before-reconciliation behavior.
- Passed the complete Windows watch feature suite, the ttsc build, test TypeScript compilation, and formatting checks before push.

### Findings and lead adjudication

Three findings survived review and were accepted.

1. **P2 — one unavailable root disabled every healthy root's reconciliation.** The guard used the topology-wide unavailable-root map. A declaration with no live watcher therefore prevented the bounded handoff scan for independently healthy registrations. The reviewer reproduced a healthy project-root watcher swallowing its first event while an unrelated external watcher failed with `ENOSPC`; the healthy change remained unreported.
2. **P2 — the public protocol guide still claimed that every observation requires an `fs.watch` event.** The new bounded event-independent scan made that statement materially contradictory even though recurring polling remains absent.
3. **P3 — the regression's opening comment omitted the mandatory numbered scenario list.** The rationale was sound, but repository test documentation requires headline, why, and numbered behavior.

Accepted. Correction commit `0091791e3` decouples healthy reconciliation from unavailable lanes, makes the reconciliation a scan-only continuation of the watcher synchronization that already occurred, adds a mixed healthy/`ENOSPC` regression, documents all four scenarios, and updates the protocol guide to distinguish the one bounded handoff scan from later event-driven observation. The first corrected test run also exposed an unnecessary second registration attempt; the implementation, rather than the assertion, was tightened so the final watch suite records one failure and one healthy change without a retry loop.

### Reviewer evidence

- Exact implementation commit and both changed files reviewed read-only.
- Focused deterministic regression, ttsc build, test TypeScript compilation, Prettier, and diff checks passed.
- The unavailable-root failure was reproduced with a mixed two-root topology.
- The reviewer traced watcher recovery, fingerprint deduplication, exact-path classification, close/drain behavior, and the matching website protocol surface.

Disposition: the core handoff fix was sound for fully observed topologies, but required the bounded mixed-root and documentation correction above.

samchon (commented):
## Individual Self-Review — `7beefc021`

Reviewed exactly `94874d94f..7beefc021` for issue #1027.

**Result: CLEAN.** No sound improvement was found.

### Requirement fidelity

- The end-to-end guide no longer advertises the removed public `ttsc transform` command.
- The replacement is a runnable `transform.mjs` example using the actual named `TtscCompiler` export and `TtscCompiler.transform()` discriminated result.
- The example handles `exception` and `failure` before reading `typescript["src/main.ts"]`, and the path follows the public project-relative key contract.
- The guide links to the existing programmatic API `transform()` anchor.
- The protocol and pitfalls pages now attribute the source-to-source shared-host error to bundler adapters or `TtscCompiler.transform()`, while retaining the internal transform verb as a protocol implementation detail.
- A fresh official-guide search found no remaining user instruction to invoke the removed CLI subcommand. Remaining text either explicitly says the command does not exist or uses “transform” as an ordinary bundler operation.

### Verification

- The named import was exercised under Node against the package's actual CommonJS publication shape.
- The existing unsupported-transform-command regression passed.
- A configured source-plugin `TtscCompiler.transform()` regression passed.
- All three changed MDX files passed Prettier and diff checks.
- The lead additionally ran the complete website production build: 89 static pages, RSS and Typia dependency-graph tests, Pagefind indexing, and sitemap generation all passed.

No implementation, package wiring, generated artifact, migration note, or additional test is required for this documentation-only correction.

samchon (commented):
## Individual Self-Review — `0091791e3`

Reviewed exactly `7beefc021..0091791e3` as the first correction for issue #1028.

### Landed

- Removed the topology-wide unavailable-root gate that had suppressed healthy registration reconciliation.
- Added a deterministic mixed-root case in which the project watcher succeeds, an unrelated external watcher fails with `ENOSPC`, and the project input still reports once.
- Made the handoff scan avoid the unconditional watcher synchronization that had produced a second failed registration attempt and could form a recovery/reconciliation cycle.
- Added the required numbered test scenarios and corrected the protocol guide's event-only claim.
- Passed the focused regression, the complete 33-test watch subtree, ttsc build, test TypeScript compilation, Prettier, and diff checks.

### Finding and lead adjudication

One P2 survived review: the scan-only correction could discover the first member behind a symlink or junction materialized in the handoff window, but it did not re-resolve the retained normalized declarations or install the newly required physical target owner. The first change was therefore reported while later target changes remained unobserved until a separate backend event or project-input republication happened.

The reviewer reproduced a missing `linked/**/*.md` declaration that became a junction to an external directory in the same turn. The handoff scan reported external `first.md`, but the watcher set still contained only the declared project-root owner. Republishing the identical declaration added the missing external owner, proving the handoff state was incomplete.

Accepted. Second correction `27e7f74c2` re-normalizes the latest declared spellings before the scan, prunes stale owner choices, preserves any already-live safe fallback covering a required lane, skips only lanes that remain unobserved, and registers newly introduced physical identities once. Its regression observes both the first swallowed target member and a second event delivered through the newly registered external owner, then proves every declared and physical handle drains on close. During validation the existing fallback-chain test also detected an unnecessary retry of the preferred failed root; the implementation was tightened to retain the live fallback rather than weakening that test.

### Reviewer evidence

- Exact correction commit reviewed read-only.
- The mixed healthy/`ENOSPC` case was confirmed non-vacuous and bounded.
- The missing physical-owner scenario was reproduced with a real junction/symlink identity transition.
- Focused regression, Prettier, and diff checks passed.
- Close behavior, exact changed paths, recovery microtasks, and the updated protocol claim were traced.

Disposition: the first correction fixed the global unavailable-root coupling but required the physical-owner continuation above.

packages/ttsc/src/launcher/internal/watchTopology.ts
```diff
@@ -671,11 +674,38 @@ export class WatchTopology {
...
+      this.closed ||
+      this.projectInputPostRegistrationReconciliationScheduled
+    ) {
+      return;
+    }
+    this.projectInputPostRegistrationReconciliationScheduled = true;
+    queueMicrotask(() => {
+      this.projectInputPostRegistrationReconciliationScheduled = false;
+      if (this.closed) return;
+      this.refreshProjectInputs(this.projectInputs.root, undefined, false);
```
samchon: **P2 — the scan can discover a new physical identity without registering its owner.** A missing declared symlink/junction may materialize before this microtask. Scanning reports its first target member, but disabling watcher synchronization leaves later target changes dark. Re-resolve the latest declared spellings and synchronize newly required physical owners once, while preserving bounded failed-root recovery.

samchon (commented):
## Individual Self-Review — `b87af6d08`

Reviewed exactly `0091791e3..b87af6d08` for issue #1029.

**Result: CLEAN.** No sound improvement was found.

### Exact scope

The commit changes only 52 action-major references across the 10 issue-owned workflows:

- `actions/checkout@v4` to `@v7`: 14 uses.
- `actions/setup-node@v4` or `@v6` to `@v7`: 14 uses.
- `actions/setup-go@v5` to `@v7`: 11 uses.
- `pnpm/action-setup@v4` to `@v6`: 13 uses.

The parent and commit blobs for `.github/workflows/nestia.yml` are identical (`0617fa644f3c8bbbd63dcc10470336115e7ca336`).

### Upstream and compatibility review

- The official upstream `action.yml` at each selected major declares `runs.using: node24`.
- Every retained `with` key remains supported.
- Explicit Node 24 selection, Go module-file and cache inputs, pnpm 10.6.4 selection, npm registry configuration, checkout depth, and release credentials are unchanged.
- `pnpm/action-setup@v6` continues reading the root `packageManager` field when release jobs omit an explicit version.
- The release trusted-publishing, package publication, smoke-test, Marketplace, and GitHub Release dependency order is structurally unchanged.

### Structural verification

The reviewer parsed parent and child YAML, normalized only the intended action-major scalar values, and compared the remaining structures. Triggers, path filters, permissions, matrices, conditions, timeouts, shells, commands, caches, and job dependencies have no other difference.

All 10 YAML files passed independent parsing, Prettier, and whitespace checks. The lead also ran the repository project-layout assertion. Exact-head CI is the final runtime proof for each migrated action.

samchon (commented):
## Individual Self-Review — `27e7f74c2`

Reviewed exactly `b87af6d08..27e7f74c2` as the second correction for issue #1028.

**Result: CLEAN.** No sound improvement was found.

### Physical-owner continuation

- The handoff microtask re-normalizes the latest declared spellings before scanning, then retains only current physical and declared declaration keys.
- A missing `linked/**/*.md` path that becomes a junction/symlink in the same turn now retains the declared project-root owner and installs the new external physical target owner.
- The regression reports the first target member from the swallowed-event scan, then drives a second target event through the newly registered external watcher and receives the second exact physical path.
- Every declared and physical fake watcher closes exactly once.

### Bounded recovery

- A required lane already covered by a live safe ancestor keeps that active watcher instead of retrying its preferred rejected root during the handoff scan.
- A previously unavailable lane with no active coverage is skipped, while a genuinely new physical identity remains eligible for one registration attempt.
- Compiler-membership invalidation carries the same boundary through the nested compiler refresh, so it cannot silently reawaken an unavailable lane.
- Failure of a newly required owner still enters the existing finite recovery fixpoint; its next handoff sees the resulting unavailable state and terminates.

### Preserved behavior

Watcher replacement transactions, reload/config/compiler classification, compiler-output suppression, fingerprint deduplication, exact-path reporting, and close/drain behavior remain intact. The test uses a Windows junction and a POSIX directory symlink and contains the required five documented scenarios. The protocol guide remains consistent with the bounded scan and subsequent event-driven model.

### Verification

- Focused swallowed-event and physical-owner regression passed.
- Existing rejected-root fallback-chain regression passed without an extra attempt.
- Complete `features/watch`: 33/33 passed on Windows.
- Test TypeScript compilation, ttsc build, Prettier, and exact-diff checks passed.
- Exact local and remote head matched and the tracked worktree was clean at review completion.

A later Windows CI run exposed only a test expectation spelling mismatch between an 8.3 temporary path and the same long physical path; commit `f2e0244a5` corrects that test-only identity comparison and has its own mandatory review.

samchon (commented):
## Individual Self-Review — `f2e0244a5`

Reviewed exactly `27e7f74c2..f2e0244a5` as the Windows exact-head CI correction for issue #1028.

**Result: CLEAN.** No sound improvement was found.

### Failure and correction

The failed Windows watch job compared two spellings of the same temporary file:

- test input: `C:\Users\RUNNER~1\...\swallowed.md`;
- watcher result: `C:\Users\runneradmin\...\swallowed.md`.

The project-input snapshot contract is physical normalization, and the implementation correctly returned the long physical identity. The test now computes each expected file with `fs.realpathSync.native()` before comparing it directly with the unmodified callback value. It does not normalize the actual result, so an incorrect alias or unrelated path cannot be collapsed into a pass.

The same contract is applied consistently to the swallowed-event, backend-event winner, mixed healthy/unavailable root, junction first-member, and junction follow-up event assertions. This also covers macOS's `/var` to `/private/var` physical alias boundary.

### Non-vacuity

- The junction case still separately requires an external physical watcher registration and invokes that exact listener for the second member; path normalization alone cannot satisfy it.
- The existing physical-registration/declared-notification and filesystem-identity snapshot regressions remain unchanged and pass.
- Deduplication, close/drain, unavailable-root, fallback, and exact owner assertions remain intact.

### Verification

- Corrected handoff regression passed.
- Physical-registration and snapshot-identity companion regressions passed.
- Test TypeScript compilation, target Prettier, and exact-diff checks passed.
- The review finished on exact synchronized head `f2e0244a5633bb08155f795b2d949ad1d206927a` with a clean tracked worktree.

Disposition: test-only platform-spelling correction is complete; the rerun Windows and macOS jobs remain the independent exact-runner proof.

samchon (commented):
## Overall Self-Review — `0fc1097d2`

Reviewed the complete cycle diff `aaa577db3..0fc1097d2` across the three accepted issues and the macOS red-CI correction. This round covered watcher state transitions, registration/event races, fingerprint deduplication, close and error paths, physical-path ownership, Windows/POSIX behavior, documentation API fidelity, workflow structure, packaging and generated contracts, and the new deterministic tests.

### Finding

**P2 — four official guide references still name a nonexistent `ttsc.transform()` API.** Issue #1027 correctly replaced the removed `ttsc transform` CLI command with `TtscCompiler.transform()` in the end-to-end, compatibility, and pitfalls passages, but a complete guide search found one remaining reference in `development/walkthroughs/lint.mdx` and three later references in `development/concepts/protocol.mdx`. The package root exports the `TtscCompiler` class; it does not export an object or namespace named `ttsc` with a `transform()` method. These passages therefore preserve the same user-facing dead end under a dotted spelling and contradict the corrected examples earlier in the protocol page.

The affected lines are outside the pull request's current diff hunks, so GitHub cannot accept an inline review comment for them. The finding is recorded here at commit scope with the exact locations: `walkthroughs/lint.mdx:17` and `concepts/protocol.mdx:449,473,480`.

### Lead adjudication

Accepted. Replace all four stale spellings with `TtscCompiler.transform()`, re-run documentation searches and formatting, push the correction, obtain the mandatory per-commit Individual Self-Review, and restart Overall Self-Review from the new immutable head.

No other sound finding survived this complete round. In particular, the compiler-watcher correction schedules only when a watcher object is newly installed, coalesces one microtask, races through the existing compiler-file stamp, reports config parse failure through the existing recovery contract, remains silent after close, and passes the complete 34-test watch subtree plus all 13 test-workspace typechecks. The four workflow action counts remain 14/14/11/13 at the Node-24 majors and `nestia.yml` remains byte-identical.

samchon (commented):
## Overall Self-Review — `2332e30d8`

Restarted a complete review of `aaa577db3..2332e30d8` after the transform-API documentation correction.

### Finding

**P2 — the compiler handoff scan still strands newly created Program members.** The new compiler reconciliation at `watchTopology.ts` scans only `this.files`, then returns immediately when no already-tracked file stamp changed. A directory watcher exists precisely to discover files that are not in `this.files` yet. If `src/new.ts` is created synchronously after `refresh(false)` and the backend drops that startup directory event, every tracked stamp remains unchanged, the microtask returns, and `resolveWatchTopology()` is never rerun. The new source therefore remains outside the resident Program until some unrelated later event.

This is the directory-membership half of the same snapshot-to-registration gap exposed by macOS CI. It is not speculative: a silent fake watcher plus a tsconfig include pattern reproduces zero topology callbacks after creating a new included source in the handoff window.

### Lead adjudication

Accepted. The bounded reconciliation must always re-resolve compiler topology once after a newly installed file or directory watcher, while retaining the stamp scan for content/config changes. With no mutation, the second resolve is a no-op and registers no watcher, so no further microtask is queued. If a new member registers another watcher, one additional bounded handoff pass is correct. Add a deterministic swallowed-directory-event scenario that requires the new source to produce exactly one topology transition, then re-run the complete watch subtree, formatting, Individual Self-Review, CI, and Overall Self-Review.

The documentation correction itself is complete: a repository-wide official-guide search now finds no `ttsc.transform()` spelling or runnable `ttsc transform` command. No other sound finding survived this round.

samchon (commented):
## Individual Self-Review — `2332e30d8`

Reviewed exactly `0fc1097d2..2332e30d8` as the #1027 Overall Self-Review correction.

**Result: CLEAN.** No sound improvement was found.

### API and documentation fidelity

- The `ttsc` package root exports `TtscCompiler`, whose instance owns the public `transform()` method; it does not export a `ttsc.transform()` namespace method.
- All four stale official-guide spellings were changed to `TtscCompiler.transform()` in the lint walkthrough and protocol result/dependency/graph passages.
- A complete Markdown/MDX search finds no remaining `ttsc.transform()` reference and no runnable instruction for the rejected `ttsc transform` CLI command.
- The internal plugin `transform` verb and diagnostic namespace remain unchanged because they are not public API calls.
- The wording now agrees with the runnable `new TtscCompiler(...).transform()` example and the programmatic API reference.

### Verification

- Exact four-reference diff reviewed.
- Both MDX files pass Prettier and whitespace checks.
- Full repository formatting completed with no unrelated tracked change.

No source, test, generated artifact, package wiring, or migration change is required for this naming correction.

samchon (commented):
## Individual Self-Review — `0fc1097d2`

Reviewed exactly `f2e0244a5..0fc1097d2` as the macOS red-CI correction.

### Landed

- Added one coalesced compiler-file stamp scan after a new compiler file or directory watcher is registered.
- Reused the ordinary stamp update as the race boundary when a backend source event arrives first.
- Recovered a synchronously deleted config with one config notification and the ordinary parse-error callback.
- Kept close silent and drained all fake watcher handles.
- Passed the complete 34-test Windows watch subtree, the ttsc build, all 13 test-workspace typechecks, full repository formatting, Prettier, and diff checks.

### Findings and lead adjudication

Two P1 findings survived review and were accepted.

1. **A newly created compiler Program member is still stranded.** The microtask scans only existing `this.files` and returns when their stamps are unchanged. A silent directory watcher plus `include: ["src/**/*.ts"]` therefore misses `src/added.ts` created in the registration window. The reviewer reproduced zero input or topology callbacks until an explicit later `refresh(true)` discovered the file. The bounded pass must re-resolve topology even when no tracked stamp changed.
2. **A POSIX atomic replacement leaves the per-file watcher bound to the old inode.** The scan reports the replacement from its changed stamp but does not execute the rearm path used by a real rename event. The reviewer reproduced one initial notification, no replacement file-watcher registration, and no second notification when the new inode received an ordinary content edit through the directory lane. The scan must retire and re-register changed existing per-file owners before later edits can occur.

Accepted. A follow-up correction will always perform one topology resolve after a new watcher registration, rearm changed existing compiler files before reporting, skip futile re-registration for entries that remain absent, and add deterministic new-member and POSIX replacement/second-edit scenarios. The resulting commit will receive its own mandatory Individual Self-Review and a fresh Overall Self-Review.

The reviewer made no repository or GitHub changes.

samchon (commented):
## Individual Self-Review - `3f40f5ad4`

Reviewed exactly `2332e30d8..3f40f5ad4` as the compiler-membership and POSIX watcher-rearm correction.

### Landed

- Made the bounded post-registration pass re-resolve compiler membership even when every previously tracked file stamp is unchanged.
- Rebound changed, existing POSIX files before reporting them so a normal later edit is not left on the old inode.
- Avoided futile per-file watcher registration for entries that remain absent.
- Added deterministic cases for a swallowed new include member, event-wins deduplication, POSIX atomic replacement plus a later edit, close cancellation, and exactly-once handle cleanup.
- Passed the complete 34-test Windows watch subtree locally, all 13 test-workspace typechecks, the ttsc build, repository formatting, dependency and layout audits, and the Ubuntu and macOS exact-head watch lanes.

### Findings and lead adjudication

Two findings survived review and were accepted.

1. **High - a same-`mtimeMs:size` POSIX atomic replacement is not rebound.** The post-registration scan still uses only the content stamp `${mtimeMs}:${size}` to decide which files enter `rearmFileWatchers()`. Replacing a tracked file with a new inode of the same size while preserving its mtime therefore leaves `changed` empty and the per-file watcher attached to the retired inode. The reviewer reproduced distinct inodes with an identical current stamp and no event from a later in-place edit. The added regression changes file size, so it does not cover this collision. Watch ownership must track filesystem identity separately from the cheap content stamp, rearming on identity movement without falsely reporting identical bytes as a compiler content change.
2. **Medium - per-file watcher rearm schedules an unnecessary full compiler refresh and loses skip context.** `rearmFileWatchers()` registers a replacement owner and schedules the same unconditional pass used for a new directory watcher. Its follow-up always calls `refreshCompilerInputs(true, false)`, even though only directory-watcher registration needs a membership scan. The reviewer reproduced one handoff containing config deletion and source replacement as `config, compiler, config` plus the same missing-config error twice. The hard-coded `false` also retries project-input roots that the originating refresh deliberately marked unobserved, while every POSIX atomic save pays for another native `tsgo --listFilesOnly` process. Scheduling must distinguish file-owner stamp reconciliation from directory-owner membership reconciliation and preserve the originating skip-unobserved context.

Accepted. The follow-up correction will separate content and physical-owner snapshots, rearm an identity change without inventing a content notification, split stamp-only and membership-refresh scheduling, preserve coalesced skip context, and extend the deterministic regression to same-stamp replacement and single-error/single-notification behavior. It will also include the Windows declared-path assertion correction discovered by exact-head CI. The resulting commit will receive its own mandatory Individual Self-Review and a fresh Overall Self-Review.

The reviewer made no repository or GitHub changes.

samchon (commented):
## Individual Self-Review — `0487ac8a9`

**Verdict: one accepted merge-blocking Medium test finding.** The production
watcher correction is otherwise clean.

Reviewed exact immutable diff:
`3f40f5ad4812d6e6af62194260f8b95effeee495..0487ac8a943722e94d92a14d0d3366280a4187cc`.

### Accepted finding — compiler notification path expectations crossed platform domains

`tests/test-ttsc/src/features/watch/test_watch_topology_reconciles_compiler_inputs_after_registration.ts`
changed the config and source notification assertions to the lexical
`fixture.config` / `fixture.source` spelling on every platform. That is correct
for Windows runner temp paths: the compiler topology retains the declared
`RUNNER~1` spelling while watcher registration uses the physical long path. It
is not the POSIX contract when the compiler canonicalizes an alias.

Exact-head macOS job
[`89918091563`](https://github.com/samchon/ttsc/actions/runs/30247614140/job/89918091563)
reproduced the consequence:

- actual config notification:
  `/private/var/.../tsconfig.json`;
- test expectation:
  `/var/.../tsconfig.json`.

The parent test's physical expectation had been green on POSIX and was wrong
only on Windows. A universal lexical expectation therefore turned a valid
product notification into a macOS test failure.

### Adjudication and correction

Accepted. Watcher-registration assertions continue to use the physical source
path on every platform. Notification assertions now use a dedicated reported
path:

- Windows: the lexical config/source spelling retained by compiler topology;
- POSIX: `fs.realpathSync.native(...)`, including macOS `/var` to
  `/private/var` canonicalization.

This does not normalize the actual value at assertion time, so a wrong product
spelling cannot pass by identity equivalence. The correction is locally green
under test TypeScript compilation and the complete Windows watcher subtree
(34/34).

### Remaining reviewed surface

The reviewer also traced snapshot seeding and pruning, per-file and
directory-event races, content versus owner movement, owner-only rearm,
membership-refresh flag and skip-context coalescing, project-input
new-watcher identity detection and recovery, close cleanup, and both new
regressions. No other sound correctness, lifecycle, platform, or test finding
survived.

samchon (commented):
## Individual Self-Review — `f3684251a`

**Verdict: CLEAN.** No actionable defect or regression-test weakening was
found in exact diff
`0487ac8a943722e94d92a14d0d3366280a4187cc..f3684251a9edc2134724c568e4960354dd2d85e2`.

The reviewed change is limited to separating compiler-notification spelling
from watcher-registration spelling in one deterministic regression:

- Windows expects the lexical config/source spelling that compiler topology
  retained in exact-head CI (`RUNNER~1`);
- POSIX expects `fs.realpathSync.native(...)`, matching macOS
  `/var` to `/private/var` canonicalization and the physical source list
  returned by tsgo;
- registration, directory containment, relative event paths, and POSIX exact
  file rearm continue to compare against `fixture.physicalSource`.

The assertions compare the raw actual notification with the selected expected
spelling. They do not normalize actual and therefore cannot hide a wrong
product path behind filesystem identity. Config deletion, a compiler directory
event, and the post-rearm POSIX in-place edit each fix the applicable
notification path. Existing count-based duplicate/error assertions retain
their original purpose.

Evidence reviewed:

- exact-head Windows `RUNNER~1` failure from commit `3f40f5ad4`;
- exact-head macOS `/var` versus `/private/var` failure from commit
  `0487ac8a9`;
- compiler config/source topology origins;
- fake watcher registration and event-relative path handling;
- direct Windows regression execution;
- exact diff whitespace check.

No file, commit, remote, or GitHub state was changed by the reviewer.

samchon (commented):
## Overall Self-Review — exact head `f3684251a`

**Verdict: CLEAN.** I found no remaining correctness, lifecycle, platform,
compatibility, test-integrity, workflow, documentation, or release-gating
improvement in the complete pull-request diff.

### Reviewed surface

I restarted this review from the immutable exact head
`f3684251a9edc2134724c568e4960354dd2d85e2`, not from the previous per-commit
reviews. The integrated base is the `79c76117f` merge of PR #1026; the campaign
claim `aaa577db3` is empty, so the complete implementation surface is
`aaa577db3..f3684251a`.

The diff contains 17 tracked files, 1,168 insertions, and 128 deletions:

- 10 workflow files whose only semantic changes are 52 action-major values;
- `WatchTopology` and two deterministic watcher handoff regressions;
- four official plugin-authoring and protocol guide pages.

I read every file and hunk, then traced the affected code beyond the diff
through compiler topology resolution, per-file and directory watcher
registration, project-input matching and fingerprinting, watcher recovery,
resident invalidation, close/drain behavior, public `TtscCompiler` exports,
workflow release ordering, and the existing Windows/POSIX watch tests.

### #1027 — public transform contract

The former `npx ttsc transform` instruction was not merely renamed. The
end-to-end guide now explains that no public transform subcommand exists and
provides a runnable `transform.mjs` example using the actual named
`TtscCompiler` export. The example handles `exception` and diagnostic
`failure` results before reading the project-relative
`typescript["src/main.ts"]` output. Protocol, pitfalls, and lint walkthrough
references now consistently distinguish the internal transform protocol verb
from bundler adapters and the public `TtscCompiler.transform()` class API.

I checked the named export, constructor, result union, output-key convention,
guide link target, unsupported-command regression, and remaining repository
references. No stale user-facing `ttsc transform` command or nonexistent
`ttsc.transform()` namespace API remains.

### #1028 — watcher registration handoff

The final design has two bounded and event-deduplicated handoff lanes:

1. Project-input reconciliation is scheduled only when synchronization
   actually installs or replaces a project or link watcher object. It compares
   the live population with the publication fingerprint once in a microtask.
   A backend event that wins updates the same fingerprint; an unchanged
   snapshot republication neither polls nor rescans.
2. Compiler reconciliation distinguishes a newly registered directory owner,
   which can require a Program membership resolve, from a newly registered
   per-file owner, which requires only a snapshot check. Coalescing preserves
   whether membership refresh is needed and the originating
   `skipUnobservedProjectInputWatchRoots` context.

I traced the following state and failure boundaries:

- startup events swallowed after `fs.watch()` returns;
- a new source admitted by a tsconfig include pattern;
- a real backend event winning the handoff race;
- a missing config producing one notification and one parse error;
- one unavailable external project-input root beside a healthy root;
- a symlink/junction materializing between publication and registration;
- declared notification paths versus physical watcher-registration paths;
- POSIX atomic replacement moving a path to a new inode;
- an identical same-`mtimeMs:size` replacement that must rebind without
  inventing a content change;
- a subsequent in-place edit on that new inode;
- source rearm beside a failed compiler membership refresh;
- repeated snapshot publication after watcher ownership is established;
- close before queued work and exact watcher drainage.

The compiler snapshot deliberately stores content identity
(`mtimeMs:size`) separately from physical ownership (`dev:ino`). This permits a
same-content replacement to rebind a POSIX file watcher while leaving named
content events authoritative inside filesystem timestamp resolution. Rearming
a file schedules only stamp reconciliation; it no longer repeats the full
compiler resolve, duplicates parse failures, retries unavailable roots with a
lost context, or starts an extra native `tsgo --listFilesOnly` process.

The project-input scheduler compares watcher-map object identities before and
after synchronization. Recovery and link retargets still schedule a handoff
scan when they install a real owner, while an unchanged publication remains
event-driven. The deterministic regression edits an established input without
delivering a backend event and proves that republishing the same snapshot does
not report it.

Windows path assertions now compare backend registration against the physical
long path and notification payloads against the declared path. That preserves
the product contract under `RUNNER~1`/long-name aliases instead of normalizing
the actual notification until a broken spelling would appear correct.

### #1029 — Node 24 action runtimes

The workflow change is exactly:

- 14 `actions/checkout` uses to `@v7`;
- 14 `actions/setup-node` uses to `@v7`;
- 11 `actions/setup-go` uses to `@v7`;
- 13 `pnpm/action-setup` uses to `@v6`.

The selected upstream majors declare `runs.using: node24`. After normalizing
those 52 scalar values, the old and new workflow structures are equal:
triggers, path filters, permissions, matrices, conditions, timeouts, shells,
commands, caches, Node 24, Go module selection, pnpm 10.6.4, trusted npm
publication, smoke tests, and release dependency ordering are unchanged.

`.github/workflows/nestia.yml` is byte-identical and, by explicit owner
direction, is outside this pull request's merge gate.

### Review chronology and adjudication

The mandatory Individual Self-Reviews and the earlier Overall rounds found and
closed real consequences rather than approving the first implementation:

- isolate healthy roots from an unavailable watcher lane;
- document the bounded scan without claiming a polling fallback;
- add the required numbered regression scenarios;
- register a physical owner when a symlink/junction materializes;
- preserve declared versus physical path domains on Windows and macOS;
- refresh membership for a swallowed new Program member;
- rebind POSIX files after atomic replacement;
- replace stale `ttsc.transform()` guide spellings;
- distinguish content movement from inode ownership;
- stop a per-file rearm from repeating compiler membership resolution;
- schedule project-input reconciliation only for a genuinely new watcher.
- preserve Windows lexical compiler paths and POSIX canonical compiler paths
  without normalizing away an incorrect actual notification.

I rejected and removed an experimental Windows parent-event/digest change. It
did not close the observed host-only sequence, added an unproven file-read cost,
and weakened the correct rule that a named content event is authoritative.
That sequence remains a fresh post-merge discovery candidate, not hidden scope
or speculative code in this PR.

### Verification

Local exact-head verification:

- `pnpm --filter ttsc build`;
- `pnpm exec tsc -p tests/test-ttsc/tsconfig.json --noEmit`;
- complete Windows `features/watch`: 34/34 passed in 47.667 seconds;
- `pnpm format`, including generated flags and Go formatting;
- all 13 test workspaces typechecked in the campaign gate;
- dependency audit: 0 high and 0 critical advisories;
- project-layout and 15-kind generated-flag integrity checks;
- YAML parsing and structural action-only comparison;
- Prettier and `git diff --check`;
- clean tracked worktree and synchronized local/remote exact head.

Exact-head GitHub Actions:

- the Ubuntu, macOS, and Windows watcher lanes all passed at the exact head;
- at review publication, 15 in-scope checks had succeeded, 19 were running, 28
  were queued, and none had failed;
- the issue-reproducing `ttsc native corpus-misc` lane was still queued and
  remains a merge gate together with every other in-scope check.

No verification was skipped within the issue-owned surface. Checks still
executing are not treated as review evidence or waived; they must complete
green before ready/merge. The unrelated Nestia workflow is explicitly excluded
from the gate.

### Merge recommendation

The complete final round produced no surviving finding. PR #1030 is ready to
merge once its exact-head in-scope checks are green.

samchon (commented):
## Exact-head CI audit — `f3684251a`

**Verdict: GREEN.** Every in-scope check settled successfully on exact head
`f3684251a9edc2134724c568e4960354dd2d85e2`; no stale-head result is counted.

Final check inventory:

- 62 successful in-scope checks;
- 0 failed, cancelled, timed out, action-required, or stale checks;
- Ubuntu, macOS, and Windows watcher lanes passed;
- `ttsc native corpus-misc`, the lane that reproduced the unchanged
  project-input republication defect at `3f40f5ad4`, passed;
- website production deployment, build, typecheck, Go, native compiler/plugin,
  package-manager cache, VS Code install, experimental platform, Socket, and
  remaining package/test lanes passed.

I also inspected every exact-head check annotation. None of the ten changed
workflow files emitted GitHub's “Node.js 20 is deprecated / forced onto Node
24” warning, confirming that the selected action majors execute on their
declared Node 24 runtimes. The only remaining Node 20 annotation belongs to the
unchanged `.github/workflows/nestia.yml`, which is explicitly outside this
campaign and merge gate by owner direction.

The branch head, pull-request head, and remote branch all resolve to the exact
SHA above. The tracked worktree is clean. This closes the final CI gate after
the clean Individual and Overall Self-Reviews.

PR #1052, opened 2026-07-30
fix(ttscgraph): canonicalize symlinked dump paths

### LINKED ISSUE #1050: fix(ttscgraph): preserve project-relative dump paths across filesystem aliases
## Summary

On macOS, the same temporary project can be spelled as both `/var/...` and `/private/var/...`. `ttscgraph` currently compares those lexical spellings when mapping source files into dump coordinates.

## Reproduction

1. Run the graph feature suite from a checkout under the macOS temporary directory.
2. Let the project root and a compiler-reported source resolve through different `/var` aliases.
3. Observe project files emitted outside the project-relative coordinate space.

On current `master` (047b464cd), this fails:

- `test_ttscgraph_dump_keeps_node_modules_as_external_leaves`
- `test_ttscgraph_dump_resolves_pnpm_workspace_edges`

## Expected

Filesystem aliases for the same existing path produce one project-relative graph coordinate.

## Actual

A source under the project is treated as a sibling outside it, so graph lookup cannot find the expected package or workspace nodes.

## Proposed fix

Resolve existing absolute filesystem paths through `filepath.EvalSymlinks` before applying the portable TypeScript path grammar, with lexical fallback for non-existent paths and synthetic Windows or UNC fixtures. Add a real symlink regression test.

mineclover:
Closes #1050

## Summary

- canonicalize existing absolute project and source paths before graph-coordinate mapping
- retain lexical fallback for missing paths and portable Windows or UNC fixtures
- cover a real project-root symlink regression

## Validation

- `node scripts/test-go-graph.cjs`
- `pnpm --filter @ttsc/test-graph start`

### DIFF
--- a/packages/ttsc/internal/graph/dump.go
+++ b/packages/ttsc/internal/graph/dump.go
@@ -246,8 +246,12 @@ func NewDump(g *Graph, project, tsconfig string, ignored map[string]bool, source
     diagnostics = []Diagnostic{}
   }
 
+  // Every relative coordinate above is relative to the mapper's canonical base,
+  // not to the spelling the caller happened to select. Publishing the raw
+  // spelling would leave a consumer resolving `../sibling.ts` against a symlink
+  // whose parent is a different directory entirely.
   dump := Dump{
-    Project:     project,
+    Project:     ctx.paths.project,
     Tsconfig:    ctx.rel(tsconfig),
     Provenance:  provenance,
     Diagnostics: diagnostics,
--- a/packages/ttsc/internal/graph/dump_path_mapper_caches_alias_resolution_test.go
+++ b/packages/ttsc/internal/graph/dump_path_mapper_caches_alias_resolution_test.go
@@ -1,0 +1,30 @@
+package graph
+
+import "testing"
+
+// TestDumpPathMapperCachesAliasResolution verifies repeated graph facts do not
+// repeat filesystem canonicalization for the same raw source path.
+//
+// One source rides the wire once per node, once per node id and twice per edge
+// endpoint, so canonicalizing inside mapPath without a raw-path cache would
+// turn every fact of a large graph into a fresh EvalSymlinks walk.
+//
+//  1. Install a counting canonicalizer on one mapper.
+//  2. Map the same absolute source path repeatedly.
+//  3. Require one canonicalization and one stable wire coordinate.
+func TestDumpPathMapperCachesAliasResolution(t *testing.T) {
+  mapper := newDumpPathMapper("C:/checkout/app")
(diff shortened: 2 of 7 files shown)


samchon (commented):
Accepted — the diagnosis and the direction are both right, and `EvalSymlinks` before the portable path grammar is the correct seam. Thank you for the report and the fix; the failure you found is real and it is invisible to CI, which is covered in #1050.

I pushed `4834dbd3` to this branch to complete it rather than send it back, because three parts of the same cause survived the original patch.

**Inputs that do not exist yet keep the aliased spelling.** `EvalSymlinks` fails outright on a path whose leaf is absent, so the fallback returned the fully lexical path — while its neighbours, which do exist, became canonical. Relative to a canonical project base those project as `../../../var/...`: the same "in-project source treated as a sibling" misprojection this issue reports, narrowed to the inputs that do not exist. A `rootDirs` target, a deleted-and-revisited source, and a path resolved through a removed container all reach it. Resolution now anchors on the longest existing ancestor and rejoins the remainder, so the whole class collapses instead of the subset that happens to be on disk.

**`Dump.Project` still published the raw spelling.** Every relative coordinate had become canonical, but the base they are relative to had not. `TtscGraphSourceReader` resolves wire paths with `path.resolve(dump.project, file)`, so a project selected through a symlink whose target has a different parent resolved every `../` sibling into the wrong directory. `/var` survives this only because it is a prefix alias — `~/link -> /srv/pkg` does not. The dump now publishes the base its own coordinates are relative to.

**Cost.** One source rides the wire once per node, once per node id and twice per edge endpoint, and `mapPath` was canonicalizing on each of them. On a real project that is the syscall walk repeated tens of thousands of times per dump. The mapper now resolves each distinct raw path once, keyed before canonicalization.

Two smaller things came with it: synthetic Windows and UNC fixtures are now excluded from the filesystem explicitly rather than relying on `EvalSymlinks` to fail on them, so the portable cross-root and injectivity rules keep proving the same thing on every host; and the tests moved to one `Test` function per file per the repository convention, adding the missing-input, published-base and Windows 8.3 spellings — the last reaches this same alias class through a different mechanism, so it belongs in the same regression set.

Verification on the exact head: `typecheck` (which runs the repository's `gofmt-2spaces` check over every tracked `.go` file), `go`, `windows-go` and `graph` all pass. No lane has failed.

One coordination note: #1056 carries an equivalent canonical-base and alias-cache change in `internal/graph/dump_paths.go`. Landing this focused fix first is deliberate — a correctness fix should not wait on a large feature branch — and #1056 should rebase onto it rather than carry a second copy.

PR #1054, opened 2026-07-31
feat(graph): expose semantic inspect CLI

mineclover:
## Summary
- add `ttsc-graph inspect` as a one-shot CLI projection of the graph MCP request union
- add local, exit-0 help for the root, dump, view, and inspect commands
- document the CLI and cover every graph-reading request branch end to end

## Verification
- `TTSC_BUILD_SCOPE=test-graph pnpm run build:current`
- `pnpm --filter @ttsc/test-graph exec tsc --noEmit --project tsconfig.json`
- `pnpm run check:format`
- direct inspect and launcher feature tests

## Known unrelated suite result
`pnpm --filter @ttsc/test-graph start` passes the new tests but fails two macOS `/var` ↔ `/private/var` external/workspace dump cases tracked in #1050. The focused path fix is already proposed in #1052.

### DIFF
--- a/packages/graph/README.md
+++ b/packages/graph/README.md
@@ -35,6 +35,18 @@ Start the client from the project root. The server builds one resident graph and
 
 `@ttsc/graph` reads the graph from the program `ttsc` type-checked, so the project needs `ttsc` and `typescript` installed alongside it. `ttsc` runs on the native TypeScript 7 compiler from the `typescript` package; it does not run on the legacy TypeScript v6.x compiler. There is no separate index step and no static-parser fallback: the graph is a byproduct of the type-check the compiler already runs, or it is not built at all.
 
+## CLI
+
+The default `ttsc-graph` command is the MCP stdio server. For one-off shell work, `inspect` exposes the same compiler-resolved request union and writes the same `{ audit, next, result }` JSON envelope:
+
+```bash
+npx -y @ttsc/graph inspect overview --cwd .
+npx -y @ttsc/graph inspect trace Service.run --to helper --focus execution
+npx -y @ttsc/graph inspect tour "How does the request flow work?" --hint Service.run
+```
+
+Run `ttsc-graph inspect --help` for `entrypoints`, `lookup`, `details`, `trace`, `overview`, and `tour` arguments. Every invocation synchronizes one current disk snapshot; use MCP when a client can reuse a resident session for several requests. `ttsc-graph dump --help` and `ttsc-graph view --help` are also local help pages and do not launch native graph work.
+
 ## Benchmark
 
 Each repository is measured with one headless agent run per arm (`baseline` with no MCP, `@ttsc/graph`, `codegraph`, `codebase-memory`, `serena`) on two prompt families, across two agent CLIs (`codex` and Claude Code). The corpus pins eight real TypeScript repositories.
--- a/packages/graph/src/bin.ts
+++ b/packages/graph/src/bin.ts
@@ -3,16 +3,24 @@
 import { runGraph } from "./index";
 import { GraphArgumentError } from "./launcherArgs";
 
-try {
-  const code = runGraph(process.argv.slice(2));
-  if (typeof code === "number") {
-    process.exitCode = code;
-  }
-} catch (error) {
-  if (error instanceof GraphArgumentError) {
-    process.stderr.write(`@ttsc/graph: ${error.message}\n`);
-    process.exitCode = 2;
-  } else {
-    throw error;
+async function main(): Promise<void> {
+  try {
+    const code = await runGraph(process.argv.slice(2));
+    if (typeof code === "number") {
+      process.exitCode = code;
+    }
+  } catch (error) {
+    if (error instanceof GraphArgumentError) {
+      process.stderr.write(`@ttsc/graph: ${error.message}\n`);
+      process.exitCode = 2;
+      return;
+    }
+    const message = error instanceof Error ? error.message : String(error);
+    process.stderr.write(
+      `${message.startsWith("@ttsc/graph:") ? message : `@ttsc/graph: ${message}`}\n`,
+    );
+    process.exitCode = 1;
   }
 }
+
+void main();
--- a/packages/graph/src/help.ts
+++ b/packages/graph/src/help.ts
@@ -24,13 +24,11 @@ export function printGraphHelp(): void {
 }
 
 /**
- * Print `ttsc-graph dump` help without starting the native binary.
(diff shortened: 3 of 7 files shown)


samchon (commented):
Reviewed the whole diff. The feature does not violate the graph contract — it stays an index-with-spans projection, adds no answer-quality checks, hardcodes no fixture, and correctly leaves `escape` MCP-only since it performs no graph work. The end-to-end coverage of all six graph-reading branches is real coverage, not a smoke test. Four findings below; the first is a behavioral regression outside the feature.

## 1. The MCP failure path can now hang instead of exiting

`runGraph` replaced

```ts
void startServer({ cwd, tsconfig, version: VERSION }).catch((error) => {
  process.stderr.write(...);
  process.exit(1);
});
```

with `await startServer(...)`, letting the rejection reach `main()` in `bin.ts`, which sets `process.exitCode = 1`.

Those are not equivalent. `startServer` registers `process.stdin.once("end", closeSession)` *before* it awaits `server.connect(transport)`. That listener refs stdin and keeps the event loop alive, so a `connect` rejection under the new code prints the cause and then leaves the process running with no server attached — an MCP client sees a launcher that started and never answers, where before it saw a clean exit 1. `process.exit` terminated regardless of open handles; `exitCode` only applies once the loop drains.

Restoring an explicit `process.exit(1)` on the server-start path (or unrefing stdin before rethrowing) resolves it. Worth a regression: launch with an unresolvable `--cwd` and assert the process exits rather than assert only on the message.

## 2. Nothing keeps the CLI in sync with the request union

`inspect` hand-mirrors six branches and roughly twenty fields of `ITtscGraphApplication.IProps["request"]`, and both switches fall through silently: `optionsFor` returns `SHARED_OPTIONS` for an unknown command, and `parseInvocation`'s `default` throws only at runtime. When the union gains a member or a branch gains a field, the CLI keeps compiling and keeps passing — the test hardcodes exactly the six invocations that exist today, so it cannot notice either.

This is the structural risk of the change, because the union is the product contract and it is expected to move. It is fixable at compile time rather than by another test: key the option table by the union's own discriminator with a `satisfies Record<Exclude<Request["type"], "escape">, readonly IInspectOption[]>`, so adding a member makes `tsc` fail until the CLI names it. That converts "the CLI drifted" from a silent gap into a typecheck error, which is the outcome worth having.

## 3. `dump --help` stopped forwarding, and the replacement duplicates a Go-owned contract

`test_ttscgraph_launcher_rejects_malformed_arguments` previously asserted `["dump", "--help"]` as a *valid forwarded form* preserving the native exit code; the PR deletes that assertion and answers locally instead. The motivation is sound and should be stated in the PR — help has to work when no native binary is resolvable — but the result is that `printDumpHelp` restates a contract `packages/ttsc/cmd/ttscgraph/main.go` owns, and restates it lossily: the native `printHelp` also documents `serve`, `--version`, and the "every node and edge, none of the MCP response caps" property that explains what `dump` is for. Two help texts for one command with no parity guard will drift on the first native flag added.

Either keep forwarding when the binary resolves and fall back locally when it does not, or add a check that fails when the native flag set and the local page disagree. Deleting the forwarding assertion without replacing the guarantee it held is the part that needs addressing either way.

## 4. Smaller items

- `runDump` accepts `-help` in addition to `--help`/`-h`; no other entry point does, and it is undocumented. Drop it.
- The synthesized CoT fields (`draft.reason`, `review`) are fine — only `question` is consumed downstream, by `runTour` — but the `runInspect` doc comment should say so, otherwise the next reader has to check whether the canned strings influence ranking.
- Negative coverage is good for the arithmetic and choice validators. It has no case for the `--` positional escape, the `--flag=value` inline form, `details` with several handles, or `--include-external` on any branch. Under the coverage rule these want negative twins, since `matchOption`'s prefix scan and the `value.startsWith("-")` rejection are exactly where an over-match hides.

## Scope

No issue precedes this PR, and it adds a permanent public CLI surface — six subcommands and roughly thirty flags — that has to be maintained against the MCP union for as long as it ships. That is a maintainer decision rather than a review finding, and it is worth settling before the items above are worked, since finding 2 is the price of keeping the surface honest over time.

samchon (commented):
Overall Self-Review over the current head, after applying my own earlier findings. Two rounds; the second is clean.

## Applied

- **The exit regression.** `startServer` registers `process.stdin.once("end", ...)` before it awaits `server.connect(transport)`, and that listener refs stdin. Letting a connect rejection reach the bin wrapper — which only sets `process.exitCode` — left a live process with no server attached. Restored an explicit `process.exit(1)` on that path.
- **CLI/union drift.** The inspect surface hand-mirrors the request union and both switches fall through by default, so an added member kept compiling and kept passing against tests that can only enumerate today's requests. A record keyed by the union's own discriminator plus a `never` assignment in the request switch now makes drift a type error. `tsc --noEmit` passes.
- **`dump --help`.** Answering locally is right, since help must work with no platform package installed, but the page had become an undeclared second copy of a contract `cmd/ttscgraph/main.go` owns and had dropped the property that says what `dump` is for. It now carries that property and names the native binary as the authority, so a drift degrades to a stale summary instead of a wrong answer.

## Round 2 — clean

- The `-help` flag remains in `DUMP_OPTIONS` and is still forwarded to the native binary exactly as on master; only the local help trigger dropped it. My earlier commit message overstated this and the follow-up commit corrects the record.
- The root help predicate is right at the boundaries: `ttsc-graph` with no argument still starts the server, `--cwd . --help` still reaches the root page.
- The synthesized CoT fields are inert — only `question` is consumed downstream, by `runTour` — and the doc comment now says so.

## Still open, and deliberately not mine to settle

No issue precedes this PR, and it adds a permanent public CLI surface that must track the MCP union for as long as it ships. The drift guard above lowers the cost of that commitment but does not remove it. Whether the surface is wanted remains a maintainer decision.

Negative coverage still has no case for the `--` positional escape, the `--flag=value` inline form, `details` with several handles, or `--include-external`. Those are where `matchOption`'s prefix scan and the `value.startsWith("-")` rejection could over-match; worth adding before this merges.

samchon:
@mineclover why closed? No more cli required?

mineclover:
@samchon 
Since I wasn’t familiar with the ttsc development roadmap, I judged it to be an unintended additional feature 
I restored Brunch after seeing the mention 
Is there a good point about the issue

mineclover:
Confirmed #1060's focused fix for the real `dump --help` resolution-path defect, including preserving the native binary as the authority and covering the fallback separately. This draft is based before the graph work now on master and overlaps that fix, so I am closing it rather than carrying the broader branch forward. Thank you.

Issue #1055, opened 2026-07-31
perf(graph): emit incremental graph snapshot shards from the resident Program


samchon:
# Outcome

Make `ttscgraph serve` export content-addressed graph shard deltas from its
resident compiler state instead of rebuilding and serializing one complete
project dump after every changed request.

This is the TypeScript producer half of
[`samchon/compiler-graph#63`](https://github.com/samchon/compiler-graph/issues/63)
and the mission in
[`samchon/compiler-graph#162`](https://github.com/samchon/compiler-graph/issues/162).
The consumer-side Graph Snapshot Protocol and atomic store already exist in
`@samchon/graph`; this issue owns the work that only `ttsc` can make
incremental.

## Problem

`ttscgraph serve` keeps one `driver.Session` and correctly reports initial,
reload, unchanged, incremental and rebuild modes. However, every changed
response still follows this path:

1. `graphSession.Snapshot` calls `changedSnapshot`;
2. `changedSnapshot` calls `buildDump`;
3. `buildDump` calls `graph.Build(program)` over all user source files;
4. it calls `graph.SourceTexts(program)` over all compiler sources;
5. `graph.NewDump` projects and sorts the complete graph; and
6. the serve response serializes the complete dump as one JSON value.

The downstream `@samchon/graph` adapter can partition that dump after receipt,
but it cannot recover the producer work, allocation or serialization already
paid. An ordinary private-body edit therefore reuses the compiler Program while
still performing graph work proportional to the whole project.

## Required invariant

For one immutable compiler generation, publish graph shards whose identities
bind the exact source/checker text, build universe, configuration and dependency
resolution that produced them. A no-op reuses the committed manifest without
rebuilding graph facts. An ordinary edit recomputes and transmits only the
compiler-invalidated semantic closure. A configuration, root-set, dependency
or public-API movement may invalidate a wider closure, but the producer must
name the changed and deleted shards explicitly.

The consumer must never observe a partial generation. Cancellation, malformed
output, digest disagreement, input movement, compiler failure or a stale base
leaves the previous committed generation usable.

## Approach

- Evolve the versioned `ttscgraph serve` protocol in coordination with
  `@samchon/graph`. Preserve an explicit compatibility failure for mismatched
  client and producer versions.
- Keep graph extraction beside the resident `driver.Session`; do not start a
  second compiler or reparse source that the Program already owns.
- Refactor graph construction so file/target-owned facts can be built and
[...]

PR #1056, opened 2026-07-31
feat(graph): emit incremental graph snapshot shards

samchon:
## Intent

Make the resident `ttscgraph serve` path publish deterministic, content-addressed graph shard generations and deltas so ordinary edits no longer rebuild and serialize one complete project dump.

## Implemented producer contract

- protocol-v1 atomic generations with exact base coordinates, ordered manifests, content-addressed upserts, explicit deletes, diagnostics, provenance and compatibility refusal
- immutable source/config/external/metadata shards owned by one resident compiler generation
- compiler-derived invalidation for no-op, body, declaration, dependency, configuration, create/delete/rename and reload paths
- cross-file edge ownership, external endpoint pruning, Windows/POSIX wire identity, cancellation, retry and last-good-generation preservation
- full-dump oracle parity while retaining the compatible one-shot dump path
- opt-in payload/path-free producer phase tracing through `SAMCHON_GRAPH_TTSC_PHASE_TRACE=1`

The downstream native transaction store is integrated in [samchon/compiler-graph#164](https://github.com/samchon/compiler-graph/pull/164) at `b4eb6074662bbb804e56064afe0a4accd68626b2`. It validates the producer manifest and native coordinates, applies dependency-bounded raw-fact deltas, reconstructs one atomic Graph Snapshot Protocol generation, and preserves the last committed snapshot on failure.

## Verification

Exact producer head: `2e2a796c1cce63de5ebc13031d71ad75e011d10d`.

- every GitHub check is green, including Ubuntu, native Go, Windows Go, linux/darwin/win32 x64 and arm64 platform packages, graph, typecheck, native defenses, package defenses, Nestia and the aggregate gate
- local `go vet ./...`, focused graph/driver tests, `pnpm test:go`, `pnpm test:typecheck`, changed-file formatting checks and platform package builds passed
- producer exact-head Individual Self-Review: [PASS](https://github.com/samchon/ttsc/pull/1056#pullrequestreview-4832837972)
- consumer exact-head Individual Self-Review: [PASS](https://github.com/samchon/compiler-graph/pull/164#pullrequestreview-4833206433)

## Integrated phase evidence

Measured through the real local producer/client lifecycle:

| Lane | Lifecycle | Producer detail | Consumer/readiness detail |
| --- | ---: | --- | --- |
| true cold | 40.269 s | native load 0.924 s; semantic refresh 0; shard export 34.878 s; encode 4.187 ms | normalize 1.713 s; commit 38.901 ms; ready 40.029 s |
| resident no-op | 0.524 s | producer total 380.029 ms; semantic refresh 379.493 ms | ready 380.860 ms |
| private body edit | 1.311 s | semantic refresh 876.036 ms; shard export 84.073 ms | normalize 29.612 ms; ready 1.008 s |

The body-edit lane meets the 2 s target. The no-op producer path is 380 ms and does **not** meet the 250 ms target. No native cold baseline was captured, so the 1.25x cold ratio is not claimed. Cold `shard-export` includes checker-query work and is not presented as pure serialization. Reload lanes remained dominated by 33.8–36.1 s shard export and need further profiling.

## Release boundary

No published `ttsc` release contains protocol-v1 shard snapshots. Published `0.23.0` still emits the legacy full dump; the downstream CI explicitly launches that version and proves warned fallback. This PR stays Draft: it is implemented and integrated, but no release or merge is authorized and the remaining performance misses are recorded rather than waived.

Issue: #1055


samchon (commented):
Commit review: `fd45c90f` (`feat(graph): emit incremental graph snapshot shards`)

Outcome: implementation direction accepted, with concrete follow-up findings. The original producer correctly established the versioned shard protocol, atomic generations, partial graph construction, explicit upserts/deletes, compatibility path, and deterministic recovery coverage for #1055. Review found that private-body edits still expanded the complete reverse-dependency closure; path projection repeated filesystem canonicalization and could publish a raw project base inconsistent with canonical relative coordinates; Go/JavaScript manifest ordering diverged for supplementary Unicode; and manifest/config coverage needed strict bijection checks. The pinned phase-separated real-project trace also remains a PR readiness requirement.

Resolution: duplicate manifest validation landed in `9d0ca52c`; the remaining code, path, ordering, config-universe, and proof-test findings landed in `4e58b964`. The real-project phase trace remains intentionally open before this draft can be marked ready. No issue is newly resolved by this review beyond the PR's declared #1055 work.

samchon (commented):
Commit review: `9d0ca52c` (`fix(graph): reject duplicate shard manifest entries`)

Outcome: production fix accepted. Strict key ordering correctly rejects duplicate manifest entries before reconstruction can conceal a missing shard. Review found that the first recovery test malformed the initial generation and therefore did not prove an already committed store was reset; its explanatory contract also needed the repository's why-plus-steps structure.

Resolution: `4e58b964` now commits one valid generation, rejects a malformed sequence-two delta, and proves a replacement child can publish a fresh sequence-one shard generation. The test contract was expanded accordingly. No additional issue is resolved by this correction commit.

samchon (commented):
Commit review: `4e58b964` (`fix(graph): preserve incremental shard invariants`)

Outcome: accepted after full local Self-Review. Private-body edits now compare tsgo's forced declaration shape and re-extract only the changed source, while public shape changes seed the resolved reverse-dependency closure. Path projection publishes the canonical base, caches alias resolution across graph facts and external IDs, and fails closed for provenance-only path errors. The TypeScript store now uses Go-compatible UTF-8 ordering and exact config-universe bijection; recovery tests exercise committed-state reset.

Validation on this exact worktree: focused Go graph/server tests, `go vet ./...`, `pnpm --filter @ttsc/graph build`, `pnpm test:typecheck`, all 68 `@ttsc/test-graph` E2E cases, `pnpm format`, changed-file Prettier checks, and full `pnpm test:go` passed. Repository-wide `pnpm build` remained in the pre-existing `packages/factory` rollup until the 15-minute local guard and is not claimed green here. The pinned phase-separated trace still gates draft readiness. This commit resolves Self-Review findings within #1055; it does not independently close another issue.

samchon (commented):
Individual Self-Review ? commit 4e58b964

This commit completed the previously identified private-body closure, canonical path, UTF-8 manifest, universe-bijection, and test-observability repairs. The read-only review found three sound follow-ups:

- Medium: incremental provenance normalization still canonicalized every committed source on each edit, keeping filesystem work O(all sources).
- Low: the public documentation overstated private-body ownership by omitting re-export and implementation-source seams.
- Low: one Go test file contained two tests, the public batch helper was not exercised directly, and the edited regression lacked the required numbered scenario.

All three were adjudicated as valid and fixed in follow-up commit 50f63de0. Focused go test ./internal/graph ./cmd/ttscgraph, focused go vet, repository formatting, and diff checks passed.

samchon (commented):
Individual Self-Review ? commit 50f63de0`n`nThe production change soundly bounds incremental filesystem canonicalization to changed facts and preserves physical/wire provenance, retry, and platform invariants. The review found one low-severity test-only defect: the regression paired independently sorted physical and wire provenance arrays by index. The cross-root poison still proved the production regression, but the final wire-preservation assertion could observe the wrong entry. Follow-up commit 868b7357 now uses the explicit physical-to-wire map.

samchon (commented):
Individual Self-Review ? commit 868b7357

Clean. The follow-up now resolves the independently sorted physical/wire provenance test entry through the explicit wireSources mapping used by production. The regression targets the intended unrelated source, retains the cross-root old-path failure proof, and introduced no new finding. Focused TestServeShardsReuseUnaffectedSources and diff checks passed.

samchon (commented):
Individual Self-Review of aeef990d

Findings:
- Low: the opt-out assertion inherited SAMCHON_GRAPH_TTSC_PHASE_TRACE from the caller, so the test failed when invoked with tracing already enabled.
- Low: native-load failures encoded an error and continued before phase emission, leaving failed loads without native-load, encode, or producer-total evidence.

Both findings are corrected in follow-up commit 8e54e8c1. Successful snapshot processing showed no correctness, atomicity, payload-leakage, or material performance regression. Focused Go tests and vet passed before the follow-up.

samchon (commented):
Individual Self-Review of 8e54e8c1

Findings:
- Low: the commit added a second Go Test function to an existing test file, contrary to the repository's one-test-per-file rule.
- Low: the early error encode-failure behavior had no regression test proving nonzero exit, observable stderr cause, and absence of successful phase rows.

Production behavior was sound. Both review gaps are corrected by 67154fcf; focused and consequence-surface Go tests plus vet pass.

samchon (commented):
Individual Self-Review of 67154fcf

Clean - no findings.

The commit restores exactly one Test function per file, preserves load-failure and ambient-environment coverage, and fixes the early encode-failure review gap with assertions for nonzero exit, visible stderr cause, and no phase rows. Focused shuffled tests, the complete cmd/ttscgraph package, vet, repository formatting, and diff checks pass.

samchon (commented):
PASS ? exact-commit Individual Self-Review for the Windows path-identity CI correction.

The hosted failure compared equivalent Windows long-path and 8.3 spellings as strings. The test now requires both paths to resolve and uses filesystem identity, while retaining the exact canonical project-coordinate assertion and tsconfig wire-coordinate assertion. This is cross-platform (`os.SameFile` uses platform file identity) and does not weaken the product contract.

Validation: 10 repeated focused passes, `go test -count=1 ./internal/graph`, `go vet ./internal/graph`, formatter check, and `git diff --check` all pass. Review outcome: CLEAN. This review covers commit 2e2a796c only; it does not authorize merge or release.

samchon (commented):
Independent review pass over the shard invalidation core and the full test set. The issue is sound and the producer contract it demands is what the branch implements; the invalidation design in particular holds up. One acceptance-matrix gap is verified below.

## What I checked and what holds

`invalidatedGraphFiles` is the part where a wrong answer would be silent — a reused shard that should have been recomputed serves stale facts to an agent with no error anywhere. It is correct as written:

- a body-only edit selects only the changed file, which is safe precisely because `publicChanged` is derived from tsgo's forced declaration shape rather than from source text, so an inferred return type that moves without the declaration text moving still lands in the public set;
- the reverse-dependency BFS over `GetReferencedFilePaths` closes transitively, and `reverseGraphDependencies` refuses self-edges and non-authored targets;
- `FileAffectsGlobalScope` escalates to the authored set, which is the right escalation and not an approximation;
- every unresolvable input (`graphSourceFile` nil, `advanceServeGraphProvenance` not ok, empty selection) falls back to `buildFullShardSnapshot` rather than publishing a partial delta — the failure direction is the conservative one throughout;
- the `ExportedTargets` / `ImplementationSources` expansion loop is monotone over a bounded file set, so it terminates, and re-running `BuildFiles` after each expansion is what makes every selected declaration fresh at commit.

I also checked the doc-comment shape on the new shard tests against the package's existing convention before flagging it, and did not: 23 of the 37 pre-existing `serve_session_*` tests use the same shortened form, so the new files match local practice rather than departing from it.

## Verified gap: three named acceptance cases are not demonstrated

The issue's acceptance list names "source create/delete/rename" and "cancellation". Checking every shard test's fixture against when the session becomes resident:

| Case | Status |
| --- | --- |
| delete | covered — `serve_shards_delete_removed_source_test.go` asserts the explicit `Deletes` entry and the dropped store identity |
| body edit / unchanged / public edit / dependency / config / oracle parity / recovery | covered |
| **create** | **not covered** — every shard test writes its complete fixture *before* `newGraphSession`, then edits or removes an existing file. No test introduces a source while the session is resident |
| **rename** | **not covered** — no `os.Rename` in the suite |
| **cancellation** | **not covered** — there is no cancellation path in `serve.go` or `serve_shards.go`, and no test |

Create is the asymmetry worth closing first. Deletion is proven explicitly, but its mirror — a new source appears, gets a shard upsert, and correctly invalidates the dependents that can now resolve an import that previously failed — is the case where a wrong closure would show up as an agent seeing a file that exists on disk but not in the graph. Rename is delete+create at the filesystem layer but not necessarily at the resolution layer, since the old path may still be referenced.

Cancellation looks safe by construction rather than by an abort path: the store commits after the build succeeds, so a client that abandons a request leaves the producer's committed generation ahead of the consumer's, and the next request's base-generation mismatch forces a full generation — the same path `bad_native_generation_restarts_session` already exercises. That is a sound design, but the acceptance criterion asks for it to be demonstrated, and right now the reasoning is implicit. Either a test that abandons a request mid-generation and proves the next one converges, or an explicit note in the issue that cancellation is discharged by base negotiation rather than by cancellation handling, would close it honestly.

## On the recorded performance misses

The 380 ms resident no-op against a 250 ms target is dominated by `semantic refresh` (379 ms), not by shard work, so it is a compiler-residency question rather than a shard-protocol one — worth stating explicitly in the issue so the remaining work is not mistaken for shard-layer optimization. The 33.8–36.1 s reload shard export is the item that most needs the profile split, since it is the lane where the incremental design should be paying off most and currently is not.

Scope note: this was a focused review of the invalidation core, the shard test set and the acceptance matrix, not a full base-to-head pass over all 46 files. The path-canonicalization work in `internal/graph/dump_paths.go` and `dump.go` now overlaps #1052, which landed the same canonical-base and alias-cache behavior as a standalone fix for #1050; this branch should rebase onto it rather than carry a second copy.

samchon (commented):
Overall Self-Review over the current head, covering the three acceptance cases I added. One round, clean.

## What the new tests pin

- **Creation.** Every prior shard test wrote its complete fixture before the session became resident, so deletion was proven while its mirror was not. This is the case where a wrong closure is invisible rather than loud: the consumer holds a graph that simply lacks a file, and an agent asking about it is told the symbol does not exist. The test also requires the dependent to be republished, because an import that previously failed to resolve now yields real cross-file facts even though the dependent's own text is unchanged.
- **Rename.** A delete plus a create at the filesystem layer, but not at the resolution layer — the old path can still be imported and the shard key binds the path. Pins the old shard deleted, a new key minted, and the repointed dependent replaced.
- **Cancellation.** There is no producer abort path. The store commits after the projection succeeds, so an abandoned request leaves the producer ahead of the consumer and base negotiation is what recovers. That is sound only if the abandoned generation is not the sole carrier of its facts, so the test abandons one generation and requires the next to supersede it and converge on the latest text.

## Round 1 — clean

Re-checked the assumptions each test makes against the production path rather than against the other tests.

- Creation and rename change the root set, so both take the reload lane and `buildFullShardSnapshot`. `BaseGeneration` still tracks the prior generation there — `serve_shards_delete_removed_source_test.go` already establishes that — so the base assertions hold on the reload path, not only the incremental one.
- The abandoned-generation test stays on the incremental lane: `graphSessionFixture` holds one source and only its text changes, so `Sequence+1` and the delete-then-upsert of the superseded key match what `serve_shards_reuse_unaffected_sources_test.go` establishes.
- `containsUpsertedShardKey` sits beside the first test that needs it, matching where `containsString` lives; the one-test-per-file rule covers `Test` functions, not helpers.
- The creation fixture starts with an unresolvable import, so the initial program carries a diagnostic. That is supported — `serve_shards_refresh_dependent_diagnostics_test.go` builds on the same ground — and is the only state in which the created-source transition can be observed.

## Not raised

I checked the new files' doc-comment shape against the package's existing convention before flagging it: 23 of 37 pre-existing `serve_session_*` tests use the same shortened form, so this is local practice rather than a departure.

## Unchanged from my earlier review

The performance items remain recorded rather than waived, and the reload lane's 33.8–36.1 s shard export is still the one that most needs a profile split. Nothing in these tests addresses that, and they do not claim to.

Issue #1053, opened 2026-07-30
docs(plugins): document community naming conventions for ttsc plugins and @ttsc/lint contributors


imjlk:
## Summary

ttsc already supports two related third-party extension surfaces:

1. top-level ttsc plugins loaded through `compilerOptions.plugins` or the
   package-level `ttsc.plugin` manifest;
2. rule contributor packages imported into `@ttsc/lint` through
   `lint.config.*`.

The runtime and loading contracts for both surfaces already exist.

This issue proposes documenting a consistent package naming and discovery
convention across the two surfaces. It does not propose new plugin resolution,
runtime validation, or automatic loading behavior.

## Current documentation

The general plugin publishing guide already documents:

- publishing a JavaScript descriptor and Go source in one npm package;
- the package-level `ttsc.plugin` manifest;
- the `ttsc-plugin` npm keyword;
- `ttsc` and `typescript` peer dependencies;
- clean-room package verification before publishing.

Its package-name example currently uses:

```text
my-ttsc-plugin
````

The public `ITtscLintPlugin` example separately uses:

```text
ttsc-lint-plugin-demo
```

These examples suggest related conventions, but the documentation does not
currently state whether these forms are recommended for community packages or
explain the distinction between the two plugin types.

## Documentation gap

The following questions are currently left implicit:

* Should top-level community plugins use `ttsc-plugin-*`?
* Should `@ttsc/lint` contributor packages use `ttsc-lint-plugin-*`?
* What are the equivalent forms for scoped packages?
* Is the `@ttsc/*` scope reserved for project-owned packages?
* Which npm keywords should each package type use?
* Which peer dependencies should a lint contributor declare?
* How does a lint contributor differ from an independently loaded ttsc plugin?
* How can a community package be discovered or proposed for the ecosystem page?

### Top-level ttsc plugins

Recommend:

```text
ttsc-plugin-<name>
@scope/ttsc-plugin-<name>
```

Examples:

```text
ttsc-plugin-example
@acme/ttsc-plugin-codegen
```

Recommended discovery metadata:

```json
{
  "keywords": [
    "ttsc",
    "ttsc-plugin",
    "tsgo"
  ]
}
```

These packages may expose the existing package-level auto-discovery manifest:

```json
{
  "ttsc": {
    "plugin": {
      "transform": "ttsc-plugin-example"
    }
  }
}
```

### `@ttsc/lint` contributor packages

Recommend:

```text
ttsc-lint-plugin-<name>
[...]

Issue #1051, opened 2026-07-30
fix(lint): avoid watching the filesystem root during config resolution


mineclover:
## Summary

The executable-config dependency collector can publish the filesystem root as a `watch` directory while resolving config imports. This is outside the project boundary and is especially visible on macOS where `/var` is a symlink to `/private/var`.

## Reproduction

1. Load a lint config from a macOS temporary directory.
2. Inspect the published config dependency graph.
3. Canonicalize `/var` and `/private/var` in the test assertion.

On current `master` (047b464cd), the focused config-loader test reports:

```text
Path: /
Kind: directory
Scope: watch
```

## Expected

Config watch inputs should remain scoped to the project or to exact external inputs that affect resolution.

## Investigation

A first attempt to simply suppress root directory records is not safe: `recordPackagePathCandidate` deliberately visits a symlink parent while tracing a path such as `/var/...`, and its current directory digest uses `/` to observe that link topology. The durable fix should record the symlink path and target directly, instead of fingerprinting its parent directory.

The issue is therefore intentionally left without a PR until that narrower dependency representation is designed and covered by a regression test.

samchon:
Verified and accepted. The report reproduces at the source level, and the investigation note is correct — a naive root-suppression would be wrong for the reason you give. Two things the issue does not yet name are worth carrying into the fix, because sealing only the reported branch would leave the same record reachable from two others.

## Verified root cause

`recordPackagePathCandidate` starts its component walk at `parsed.root` and holds `current = "/"` for the whole first iteration. Every early return inside that first iteration therefore publishes the filesystem root as a directory dependency:

| Branch | Condition at `index === 0` | Record |
| --- | --- | --- |
| symlink observation | `fs.lstatSync(next).isSymbolicLink()` | `recordDirectoryDependency("/")` |
| stat failure | `fs.lstatSync(next)` throws | `recordDirectoryDependency("/")` |
| non-directory ancestor | first component is a regular file | `recordDirectoryDependency("/")` |

The reported macOS case is the first row: `/var` is a symlink, so tracing `/var/folders/.../lint.config.ts` records `/` before it ever recurses into `/private/var`. The second row is reachable on every platform — any absolute candidate whose first component does not exist (a `main` field pointing outside the package, a resolution candidate that was probed and missed) produces the identical `/` record without a symlink anywhere. The third is rarer but the same class.

[...]

PR #1058, opened 2026-08-01
fix(lint): scope config watch inputs and document plugin naming

samchon:
## Intent

One cycle pull request for the two published issues that have no owning pull request of their own.

- #1051 — the executable-config dependency collector publishes the filesystem root as a `watch` directory.
- #1053 — no documented community naming convention for ttsc plugins and `@ttsc/lint` contributors.

The other campaign items are owned elsewhere and are not in this branch: #1050 by #1052 (complete, 25/25 checks green), #1055 by #1056, and #1054 carries no issue.

## Scope

**#1051.** The collector walks path components from the filesystem root and holds `current = "/"` through the first iteration, so every early return in that iteration records `/` as a directory dependency and digests it by reading the whole root. Three branches reach it, not the one the issue reports: the symlink observation, the `lstat` failure, and the non-directory ancestor. The fix records the observed entry itself rather than fingerprinting its parent, in both embedded loader scripts.

**#1053.** Documents the naming convention the authoring guides already follow in practice, adds the missing `@ttsc/lint` contributor publishing section, and links discovery from the ecosystem page.

## Verification

Pending. Verification is this pull request's ordinary CI; results are recorded as pull-request reviews rather than by rewriting this body.

No version field is touched. Release and launch are out of scope for this branch.

### DIFF
--- a/packages/lint/linthost/config.go
+++ b/packages/lint/linthost/config.go
@@ -2179,6 +2179,20 @@ function recordDirectoryDependency(location, owners) {
 // does not own -- /var on macOS is a symlink whose parent is the filesystem
 // root -- and digesting that parent both reaches outside the project boundary
 // and reads the whole directory to learn one entry's state.
+// A path candidate is observed through the directory that would own a
+// competing resolution, so a sibling winning extension resolution still
+// invalidates. That reasoning is what the parent digest is for and it stays.
+//
+// It does not reach the filesystem root. The root owns no candidate this trace
+// could pick, and a resolution path routinely passes through an ancestor
+// directly beneath it -- /var on macOS is a symlink whose parent is the root --
+// so digesting the parent there enumerates the entire filesystem root on every
+// config load, outside the project boundary. Record that one ancestor instead.
+function recordAncestorDependency(parent, entry, root, owners) {
+  if (parent === root) recordEntryDependency(entry, owners);
+  else recordDirectoryDependency(parent, owners);
+}
+
 function recordEntryDependency(location, owners) {
   recordDependency("entry", location, entryDigest(location), owners);
 }
@@ -2761,11 +2775,11 @@ function recordPackagePathCandidate(
     try {
       entry = fs.lstatSync(next);
     } catch {
-      recordEntryDependency(next, owners);
+      recordAncestorDependency(current, next, parsed.root, owners);
       return;
     }
     if (entry.isSymbolicLink()) {
-      recordEntryDependency(next, owners);
+      recordAncestorDependency(current, next, parsed.root, owners);
       try {
         const target = fs.readlinkSync(next);
         const remainder = components.slice(index + 1);
@@ -2787,18 +2801,12 @@ function recordPackagePathCandidate(
       }
     }
     if (index === components.length - 1) {
-      // A resolved file is observed through its parent, so a sibling that would
-      // win extension resolution still invalidates. That reasoning does not
-      // reach the filesystem root: it owns no candidate this trace could pick,
-      // and digesting it is the same out-of-boundary root listing the branches
-      // above were narrowed to avoid.
       if (isDirectory) recordDirectoryDependency(next, owners);
-      else if (current === parsed.root) recordEntryDependency(next, owners);
-      else recordDirectoryDependency(current, owners);
+      else recordAncestorDependency(current, next, parsed.root, owners);
       return;
     }
     if (!isDirectory) {
-      recordEntryDependency(next, owners);
+      recordAncestorDependency(current, next, parsed.root, owners);
       return;
     }
     current = next;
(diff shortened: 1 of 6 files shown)


samchon (commented):
Overall Self-Review — round 1 found a defect, round 2 is clean. CI is still running; this review is the change-surface gate, not the CI gate.

## Round 1 finding, applied

The first commit sealed three branches of `recordPackagePathCandidate` and left two. That is whack-a-mole, not a fix of the cause.

`recordPackagePathCandidate` observes a resolved *file* through its parent directory so a sibling that would win extension resolution still invalidates. Sound for an ordinary parent, wrong for the filesystem root: a candidate resolving to a root-level file reaches its last component while `current` is still the root, so the parent digest enumerates the whole root again — one branch later than the ones I had narrowed. The walk's tail is the same case for a candidate that resolves to the root itself, which names no package at all.

Both now record the observed entry, and a third fixture whose manifest main is a root-level path covers it.

## Round 2 — clean

Re-read the whole surface against the current head.

- **JS/Go digest parity.** The `entry` encoding is written by the loader script and re-derived by `configDependencyDigest`; a mismatch would silently invalidate every cached evaluation. `missing\0`, `<kind>\0` and `symlink\0<target>` agree, and both sides read link state with `lstat`/`Lstat` rather than following it.
- **The kind whitelist.** `normalizeConfigDependencyFingerprints` rejects the *entire* fingerprint set on an unknown kind, so omitting `entry` there would have turned every config load into "malformed dependency fingerprints" rather than degrading. It is admitted.
- **Cache version.** v5 entries record ancestors as directory digests, so reusing one would keep republishing the root from a stale fingerprint regardless of the new code. v6 forces the miss.
- **Both loaders.** The CommonJS and TypeScript scripts are equivalent copies; all five call sites and both digest helpers changed in each.
- **Scope semantics unchanged.** `entry` is not a directory, so it joins the exact-path watch list rather than the directory list — the same treatment `optional-file` already gets for paths that may not exist.

## Corrections to my own test, found by reading rather than by a failing run

- `index.cjs` cannot serve as a package index fallback; Node's `LOAD_INDEX` tries only `.js`/`.json`/`.node`. The two packages with a deliberately unresolvable main would have failed to require at all instead of reaching the ancestor branches.
- Asserting no `directory` record for the project root was wrong in the opposite direction: `recordResolutionTopology` walks node_modules search directories upward and records each on purpose, root included. Replaced with the watch-inputs-stay-within-project assertion the issue actually requires.

## Docs (#1053)

No behavior surface. The naming section documents what `getting-started` and `end-to-end` already build, so the only example changed is the one that was out of step. The contributor page is new and its last step — linking Go source on a consumer machine — is the failure a type-check cannot catch, which is why it is the step called out.

## Not verified locally

Go tests and `gofmt-2spaces` were not run here; CI owns both, as it did for #1052.

samchon (commented):
Overall Self-Review — round 3, clean, over head `5c805dab` with all 21 checks green.

## What rounds 1–2 got wrong, and how

Round 1 caught that I had sealed three of the collector's branches and left two. Round 2 read clean. Both rounds missed the larger error, and CI found it: replacing the parent digest on *every* early-return branch broke `TestScriptConfigLoaderTracksLocalDependencyGraph`, which requires `exports-priority/bridge` to stay in the graph.

That expectation is correct and my change was wrong. The parent digest is not the defect — it is the mechanism by which a sibling that would win extension resolution invalidates the cache. I removed real invalidation to fix an unrelated boundary, which is a worse trade than the reported bug.

Worth naming why review missed it twice: both rounds checked "is the root still recorded?" and never asked "did anything that used to invalidate stop invalidating?". A narrowing change needs the second question, and the regression I wrote had no negative twin that would have posed it.

## The corrected shape

The defect is only that a resolution path routinely passes through an ancestor sitting directly beneath the filesystem root — `/var` on macOS — where the parent digest becomes a full root listing outside the project boundary. All four branches plus the walk tail now route through one `recordAncestorDependency` that narrows to the observed entry when, and only when, the parent is the root.

This makes the change provably regression-free rather than argued to be: for `parent !== root` the emitted call is byte-identical to what master ran. Only the root case differs, and the root owns no candidate a resolution trace could pick, so nothing that made the parent digest meaningful applies there.

## Round 3 verification

- Five call sites, both embedded loaders, identical routing in each.
- JS/Go `entry` digest parity: `missing\0`, `<kind>\0`, `symlink\0<target>`, both reading link state with `lstat`/`Lstat` rather than following it.
- `normalizeConfigDependencyFingerprints` admits `entry`; without it every fingerprint set would have been rejected wholesale rather than degrading.
- Cache version v6 forces a miss on v5 entries, which recorded root ancestors as directory digests.
- The regression now carries the negative twin that was missing: an ordinary in-project candidate must still publish its parent directory digest. That assertion is what would have caught rounds 1–2.

## CI

21/21 green on this head, including `go`, `windows-go`, both lint defense lanes, all six platform rows, `typecheck` (which runs the repository's own `gofmt-2spaces` check over every tracked `.go` file), and nestia.

## Docs (#1053)

No behavior surface; unchanged since round 2 and re-read here.

PR #1060, opened 2026-08-01
fix(graph): answer dump --help when the native binary is absent

samchon:
## Intent

`ttsc-graph dump --help` was unreachable without an installed platform package. `dump` forwards every flag to `ttscgraph`, which owns the flag contract and answers `--help` itself, but resolution ran first — so the one command that could tell a caller what to install only worked for callers who had already installed it.

## Scope

Help falls back to a short summary on the resolution-failure path only.

- A resolvable binary keeps answering help, so the two texts cannot diverge for anyone who has it installed.
- The fallback names `ttscgraph` as the authority, so a drift degrades to a stale summary rather than a wrong answer.
- An ordinary `dump` with no binary still fails with the same resolution error.

## Provenance

Found while reviewing #1054, which proposed answering help locally in every case and deleting the assertion that `dump --help` forwards. That PR was withdrawn by its author. This takes only the part that was a real defect on `master` and keeps the native binary authoritative.

## Verification

Pending CI. Locally: `tsc --noEmit` on `@ttsc/graph` and `@ttsc/test-graph`, and formatting on the changed files.

Regression covers all three help spellings without a binary, the preserved failure for an ordinary `dump`, and — on hosts that can mark a sentinel executable — that `--help` still reaches the native binary unchanged.

### DIFF
--- a/packages/graph/src/index.ts
+++ b/packages/graph/src/index.ts
@@ -83,6 +83,37 @@ export function runGraph(
   );
 }
 
+/** The help spellings `ttscgraph` itself accepts, mirrored for the fallback. */
+const DUMP_HELP_FLAGS = new Set(["--help", "-help", "-h"]);
+
+/**
+ * Print a `dump` usage summary when the native binary cannot be resolved.
+ *
+ * This is a fallback, not a second contract: `cmd/ttscgraph/main.go` owns these
+ * flags and answers whenever it is installed. Keep this short and point at that
+ * authority, so a drift degrades to a stale summary rather than a wrong
+ * answer.
+ */
+function printDumpHelp(): void {
+  process.stdout.write(
+    [
+      "Usage: ttsc-graph dump [options]",
+      "",
+      "Write the whole compiler graph as JSON to stdout: every node and edge,",
+      "none of the MCP response caps.",
+      "",
+      "Options:",
+      "  --cwd <dir>        Project root (default: current directory).",
+      "  --tsconfig <path>  Project tsconfig path (default: tsconfig.json).",
+      "  --pretty           Indent the JSON output.",
+      "",
+      "The native `ttscgraph` binary owns these flags and is not installed here,",
+      "so this summary may lag it. Install `ttsc` and rerun for the exact list.",
+      "",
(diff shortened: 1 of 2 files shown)


samchon (commented):
Overall Self-Review — round 2 clean, 7/7 checks green.

## Round 1 finding, caught by CI rather than by me

The fixture set `TTSC_GRAPH_BINARY` to a path that does not exist, intending to model "no binary". It did not. `resolveGraphBinary` returns an absolute override **unchecked**, so resolution succeeded and the spawn failed instead — exit 1 from the `result.error` path, never reaching the resolution-failure branch the fallback lives on.

The production change was correct throughout. The test asserted something it was not exercising, which is the worse of the two failures: a green run would have claimed coverage that did not exist. Replaced with the real condition — an empty temporary project with no `ttsc` to resolve a platform package from and no override set.

This is the third time this campaign that a test of mine passed or failed for the wrong reason, and all three share one cause: I checked whether the production code was right and not whether the test reached it. Recording it here because the pattern is more useful than the individual fixes.

## Round 2 — clean

- **The fallback is scoped to help.** An ordinary `dump` with no binary still exits non-zero with the same resolution message; the test asserts that rather than only asserting the happy path.
- **The native binary keeps owning help.** When resolution succeeds, `--help` is forwarded unchanged — asserted via the sentinel marker on hosts that can mark it executable. The two texts therefore cannot diverge for anyone who has the binary, which is what keeps this a fallback rather than a second contract.
- **All three spellings.** `--help`, `-help` and `-h` are the set `DUMP_OPTIONS` already declared; the fallback mirrors exactly that set rather than inventing its own.
- **The summary names its authority.** Asserted, not merely written, so a future edit that drops the pointer fails.
- **Windows.** The forwarding half is skipped where a sentinel cannot be made executable, matching how the existing launcher test handles the same constraint.

## Provenance

The defect was found while reviewing #1054, which proposed answering help locally in every case and deleted the assertion that `dump --help` forwards. That PR was withdrawn by its author. This takes only the part that was a real defect on `master` and leaves the native binary authoritative.

PR #1066, opened 2026-08-04
fix(lint): read every TypeScript source the type-check pass reads

### LINKED ISSUE #1065: Lint skips sibling workspace sources the type-check pass reads
## Summary

In a pnpm TypeScript monorepo whose packages resolve to source, `ttsc` type-checks a sibling package's sources but `@ttsc/lint` does not lint them. A first-party sibling workspace source that the type-check pass reads is never walked by the lint pass, and does not appear in a project rule's `ctx.Sources`.

Still present in `0.23.0`, unchanged (see [Still present in 0.23.0](#still-present-in-0230) below).

## Observed

A file that enters the TypeScript Program through an import — a sibling workspace package resolving to its own `src` — is type-checked, but:

- no file rule runs over it, and
- it is absent from the `ctx.Sources` a project rule receives.

The same `ttsc` invocation therefore holds two different views of one Program.

## Where

`linthost/host.go`:

```go
// userSourceFiles returns the tsconfig-selected source files the lint engine
// owns. The tsconfig file list is the boundary: imported libraries, generated
// output, and JSON modules may still appear in Program.SourceFiles(), but lint
// and format should not walk them unless the project selected them as TS/JS
// source roots.
func (p *program) userSourceFiles() []*shimast.SourceFile {
	roots := p.userSourceFileNames()
	out := make([]*shimast.SourceFile, 0)
	for _, f := range p.tsProgram.SourceFiles() {
		if f == nil {
			continue
		}
		if _, ok := roots[canonicalProjectPath(p.cwd, f.FileName())]; !ok {
			continue
		}
		out = append(out, f)
	}
	return out
}

func (p *program) userSourceFileNames() map[string]struct{} {
	out := make(map[string]struct{})
	if p == nil || p.parsed == nil || p.parsed.ParsedConfig == nil {
		return out
	}
	for _, fileName := range p.parsed.ParsedConfig.FileNames {
		if isLintSourceFileName(fileName) {
			out[canonicalProjectPath(p.cwd, fileName)] = struct{}{}
		}
	}
	return out
}
```

`roots` comes only from `ParsedConfig.FileNames`, so a file that entered the Program through an import is dropped. That filtered list is what both lint paths receive: `runLintCycle` passes it to `engine.runFiles`, and `runProjectCycle` passes it to `engine.evaluateProject`, which forwards it through `projectSources` (`linthost/project_engine.go`) into a project rule's `ctx.Sources`. `projectSources` only subtracts globally-ignored files; nothing re-widens the set downstream.

## Still present in 0.23.0

`linthost/host.go` is byte-identical between the shipped `@ttsc/lint@0.22.0` and `@ttsc/lint@0.23.0` — `userSourceFiles` (line 356), `userSourceFileNames` (line 371), and both call sites (line 272, `evaluateProject(p.identity, p.userSourceFiles(), p.checker)`; line 281, `files := p.userSourceFiles()`).

Across the whole `linthost/` directory the only file that differs between the two releases is `project_inputs.go`, and that change is Windows path-case handling (`isCaseInsensitiveFilesystem()` becoming `usesWindowsPathSyntax()`). It does not touch this path.

## Why this matters in a source-entry monorepo

The comment justifies the boundary with "imported libraries, generated output, and JSON modules". That reasoning fits a published dependency read from `lib`, where the consumer neither authors nor reviews the file. The same filter also catches a case that is not that.

In a pnpm workspace it is ordinary for a package to publish source as its entry:

```json
{
  "main": "./src/index.ts",
  "exports": { ".": "./src/index.ts" },
  "publishConfig": {
    "main": "./lib/index.js",
    "types": "./lib/index.d.ts"
  }
}
```

Every workspace consumer then resolves that package to first-party TypeScript that the same repository authors, reviews, and ships. `tsc` treats it as such: importing it pulls the real sources into the Program and type-checks them.

A consumer cannot express the intent through configuration either. Adding the sibling to `include` is not equivalent — it also changes what the project emits, forcing `rootDir` above the package and moving every output path. That is a build-layout change made to obtain a lint-scope effect.

## Consequence

`@samchon/lint-plugin-evidence` contributes a project rule that declares populations by glob. A claim rooted at a sibling package — which its own configuration documents as resolving "from that project root, so a monorepo package can select an explicitly included sibling with `../api`" — receives an empty population once that sibling is reached by import rather than by `include`.

The failure is silent rather than loud, which is what makes it costly: an empty population demands nothing, so the claim reports full coverage while checking nothing. The rule passes, the build is green, and the obligation was never evaluated.

## The ask

That the reading and reporting side — the file rules' diagnostic pass and a project rule's `ctx.Sources` — be able to see a first-party sibling workspace source that the type-check pass of the same invocation already reads.

## What is *not* being asked

`format` should keep staying inside the project, and I am not asking for it to widen. It writes files, and a project must not rewrite a sibling package's sources merely because it imports them. Whatever shape a change takes, the write side staying bounded by the project seems right to me.

Worth noting for scoping: today the read side and the write side share one list. `format` (`format.go`), `lint --fix` (`fix.go`), plain `lint` (`compile.go`), and the LSP diagnostic and document-format paths (`lsp.go`) all obtain their files from the same `runLintCycle`/`userSourceFiles` call. So reporting and writing are not currently separable at this point in the code — that separation, if it is wanted, does not exist yet.

## Open question

What should the fix pass do with a diagnostic on a file the project does not own?

It writes, which puts it on `format`'s side of the line above. But handing a user a diagnostic they cannot fix in place is its own oddity. Middle positions exist — reporting widely while applying fixes only within the project, or offering a fix without applying it. I do not have a view on which is right, and it is a product decision rather than an implementation detail, so I would rather ask than assume.

## Environment

- `@ttsc/lint@0.22.0` and `@ttsc/lint@0.23.0` (both exhibit the code path above)
- pnpm workspace, packages resolving to `./src/index.ts` through `main`/`exports`
- Windows 11; workspace links materialize as junctions

## What I verified, and what I did not

Read from the shipped `linthost/host.go` and `linthost/project_engine.go` in both `@ttsc/lint@0.22.0` and `@ttsc/lint@0.23.0`, tracing `userSourceFiles` through `runLintCycle`, `runProjectCycle`, `evaluateProject`, and `projectSources`, and diffing the whole `linthost/` directory between the two releases. Line numbers cited are from `0.23.0`.

I did not build a reduced reproduction. The observation comes from the source and from a project rule receiving an empty population where its configuration says it should not.

samchon:
Closes #1065.

## Intent

`@ttsc/lint` held a narrower view of the Program than the type-check pass that loaded it. `linthost/host.go::userSourceFiles` filtered `Program.SourceFiles()` down to `ParsedConfig.FileNames`, so a first-party sibling workspace package resolving to its own `src` was type-checked and never linted: no file rule ran over it, and it never appeared in a project rule's `ctx.Sources`. An empty population demands nothing, so a coverage claim rooted at that sibling passed while checking nothing.

The lint read scope now follows the type-check pass. The write scope does not.

## Scope

**Read side widens.** `userSourceFiles` keeps every tsconfig-selected TS/JS root and adds every authored TypeScript source the Program pulled in through an import (`.ts`, `.tsx`, `.mts`, `.cts`, non-declaration). Both `runLintCycle` and `runProjectCycle` consume that set, so file rules and `ctx.Sources` see the same Program the checker does.

**Selection still governs the rest.** A `.d.ts` is typings rather than authored source, and the bundled `lib.*.d.ts` set plus every published package's typings reach `Program.SourceFiles()` the same way. JavaScript enters only under `allowJs`, where the project's own file list already selects the JS it owns. A JSON module carries no lint source. A project that selects any of those explicitly keeps them, exactly as before.

**Write side stays inside the project.** `fix`, `format`, and the LSP document fix/format commands filter findings through the new `projectOwnedFixableFindings` before applying edits, so a project never rewrites a sibling package's sources merely because it imports them. The diagnostic still prints; the owning package fixes it from its own run under its own config. One write rule, no per-path exception.

## Verification

| Case | File |
| --- | --- |
| Imported TypeScript joins the selected roots, JSON stays out | `user_source_files_span_program_typescript_sources_test.go` |
| Negative twin: imported `.d.ts` and imported `.js` stay out | `user_source_files_exclude_imported_declarations_and_javascript_test.go` |
| A file rule reports on a sibling workspace source | `command_check_reports_imported_sibling_workspace_source_test.go` |
| `fix` reports the sibling and leaves it byte-identical | `command_fix_reports_but_never_rewrites_imported_sibling_source_test.go` |
| `format` leaves the sibling byte-identical while reflowing the consumer | `command_format_leaves_imported_sibling_source_untouched_test.go` |
| `ctx.Sources` holds the sibling source | `project_rule_sources_include_imported_sibling_source_test.go` |

Local verification was `gofmt` (parse and format) over every changed Go file and `prettier` over the changed docs. The Go unit suite and the TypeScript e2e suites are deferred to this pull request's CI.

## Change integrity

`user_source_files_follows_tsconfig_roots_test.go` pinned the former boundary and is superseded by the two `user_source_files_*` cases above. Its JSON-module and selected-`.d.ts` assertions are preserved there.

`rule.ProjectContext.Sources`, the lint README, `docs/lint/index.mdx`, and the plugin-author walkthrough documented the old tsconfig-selected contract and are updated in the same change.

## Deferred, and why

**Benchmark file selection is untouched.** `TtscBenchmarkPerformanceTypeScriptFileSelector` feeds ESLint and Prettier the tsconfig file list, which mirrors the former lint scope. Parity survives this change: the widening admits only non-declaration TypeScript, and cross-package resolution in the RxJS and NestJS fixtures lands on `.d.ts`, so their file sets are unchanged. The Prettier lane compares against `ttsc format`, whose scope did not move at all. A future fixture that resolves a dependency to source would need the selector widened to match.

**A sibling source cannot be silenced from config.** `config.go::matchAnyPattern` returns false for any file above the config's `BaseDir`, so `ignores: ["../api/**"]` is inert today and stays inert. Inline `ttsc-lint-disable` comments in the owning file remain the only suppression. Changing that is a config glob-semantics decision outside this issue; raised here rather than folded in.


samchon (commented):
Overall Self-Review, rounds 1 and 2 over `master...da46aacf7`. CI is still running; this records what the review changed and what it cleared.

## Applied

**`projectOwnedFixableFindings` was misnamed.** It never inspected `Finding.Fix`, so "fixable" was false. Its actual job is selecting the findings this project may write to disk. Renamed `projectWritableFindings` in `da46aacf7`.

**The walkthrough paragraph carried two jobs.** The `ctx.Sources` population is now its own paragraph instead of being spliced into the project-rule lifecycle sentence.

## Cleared

**No other habitat for the same cause.** The defect was a user-source set derived from the tsconfig file list rather than from the Program. `ParsedConfig.FileNames` has exactly one consumer in the repository, and `packages/ttsc/driver`, `packages/ttsc/internal`, and `packages/graph` all derive their file sets from the Program. The class is closed by this change rather than one witness of it.

**Existing `userSourceFiles()` callers survive the widening.** Eight Go tests call it, seven of them asserting `len(files) == 1` or indexing `files[0]`. Every one of those fixtures writes a tsconfig selecting a single source, so neither the count nor the Program order they depend on moves.

**The fix and format cascades still converge.** A dropped sibling finding never increments `fixed`, so the loop reaches `fixed == 0` and exits through the ordinary converged path rather than the pass cap.

**The LSP gained no write surface.** `lspWorkspaceEditForCommand` already rejects a target outside `cwd` and one under `node_modules`, and the ownership guard now runs before the document filter, so the read-scope widening cannot turn a newly visible file into a newly writable one. The suggestion path returns a `WorkspaceEdit` for the editor to apply and touches no disk.

**Watch stays correct.** The files the Program reads are already watched for type-check correctness, so the read set and the watched set remain the same set.

## Accepted cost

`ttsc format` now walks imported sources and discards every finding they produce. Walking only the writable set would avoid that, at the price of a project rule receiving a different `ctx.Sources` under `format` than under `check` for the same project. One read set per invocation is the invariant this issue asks for, so the uniform cycle wins and the wasted walk is accepted. The benchmark fixtures are unaffected: their cross-package resolution lands on `.d.ts`, which the widening does not admit.

samchon (commented):
Overall Self-Review, round 3 over `master...ee2dc14b8`, restarted from scratch because the scope decision changed after round 2.

## Corrected from round 2

Round 2 accepted `format` walking the wider read scope and discarding what it produced, on the argument that one read set per invocation keeps `ctx.Sources` uniform between commands. That was wrong, and the accepted cost was not the maintainer's to pay. Format writes and reports nothing, so a finding outside the project has nowhere to go; the walk was pure waste, once per cascade pass. `runFormatCycle` now walks the tsconfig-selected set, and the LSP document commands take the same narrow cycle since they edit one target and never report.

Scope is now enforced by what a command reads rather than by discarding findings afterwards. `projectWritableFindings` survives as `fix`'s guard alone, because `fix` is the one command that must read wider than it writes: it prints the diagnostics that outlive the cascade.

## Applied this round

**The read scope had no test proving it is governable.** A global `ignores` entry drops an imported source from the walk and from a project rule's population, and nothing pinned that. The new case runs one fixture twice so the silence it asserts belongs to the ignore: without the entry the same imported file fails the command. A single-run version would have passed even if the widening did not exist.

**A population difference was undocumented.** Project rules are evaluated during a format run too, and they now receive the narrower set there. That is observable to a plugin author reasoning about coverage, so the walkthrough states it and says which run to draw a workspace-wide conclusion from.

## Cleared this round

**The scope memo has no live mixing.** `runCycleOver` memoizes the project cycle on the program, so scope is a property of the program rather than of the call. Every reporting path (`check`, `fix`, LSP diagnostics, hints, the resident cache) takes the wide cycle, and every writing path builds its own program and takes the narrow one, so no program is asked for both. The constraint is now stated where the memo lives.

**The LSP document commands are behavior-preserving.** Their target is bounded to `cwd`, and a target inside `cwd` that the tsconfig did not select produced no findings before this branch and produces none now.

## Unrelated red lane, folded in on request

Four high advisories published on 2026-08-03 turned the dependency audit red, after master's last green run on 2026-08-02, so master fails the same gate. No override was needed: every parent edge already declares a caret range admitting the patched version, so four resolutions were refreshed and the lockfile moved 16 lines with no other resolution touched.

samchon (commented):
Overall Self-Review, rounds 4 and 5 over `master...d7303540f`.

## Correction to the pull-request body

The body's deferred section claims a sibling source cannot be excluded from config. That is wrong, and the claim stands corrected here rather than by an edit, so the chronology survives.

A `files` selector does control it. `matchAnyPattern` anchors every pattern at the config's own directory and refuses a path above it, so an entry written as `files: ["src/**/*.ts"]` matches no imported sibling: rule resolution returns `OutOfScope` with an empty rule set, and the walk produces nothing for that file while the project's own files stay governed. Only a config with no selector at all applies its rules to everything it reads.

What the body says about `ignores` remains true: a global ignore cannot name a path above the config directory either, so it is the wrong tool for a sibling. The right one is the selector, and both are now documented under [Which files are linted](https://ttsc.dev/docs/lint#which-files-are-linted) and pinned by two cases whose first run proves the sibling reports without the control applied.

The body also predates two renames and the format-scope correction. `projectOwnedFixableFindings` is now `projectWritableFindings` and belongs to `fix` alone; `format` and the LSP document verbs walk the project's own sources through `runWriteScopedCycle` instead of filtering afterwards.

## Applied in round 4

**`runFormatCycle` was a lie at one call site.** The LSP fix-document verb takes the same cycle, and it is not a format command. `runWriteScopedCycle` names what unites its callers: they edit files and report nothing.

**A test comment mirrored a specific third-party plugin's scenario.** Generalized to the property under test, so the case reads as a contract about populations rather than as tailoring to one consumer.

## Applied in round 5

The `files`-selector control had no coverage and no documentation, described above.

## Cleared

**Windows lanes.** The new cases assert on basenames and on `\n`-joined format output, matching the existing format cases that already pass on the `windows-go` lane. The one substring the sibling assertions rely on, `index.ts`, cannot come from the import specifier, which is written without an extension.

**Config discovery in the two-package fixture.** The consumer owns the only `lint.config.json`, and discovery walks upward from its tsconfig directory, so the sibling's absence of a config cannot change which rules resolve.

Issue #1064, opened 2026-08-04
ttsc cache questions


webstech:
Thanks for the doc updates you provided recently on using the cache.  I have a couple of questions for github workflows.

1. Is the cache platform specific?  ie should the `runner.os` be part of the `key` for saving and restoring?
2. Does the cache need to be saved all the time or can it be managed based on `ttsc`, `@ttsc\unplugin`, `typia` and `typscript` updates?  If this is the case, example workflows would be great.  Saving all the time would be problematic if the cache is not platform specific.

samchon:
Both are good questions, and the second one has a non-obvious answer.

Worth saying up front: the `actions/cache` key only affects your hit rate, never correctness. ttsc's plugin cache is content-addressed. A cached binary is reused only when a key computed from the ttsc version, the resolved `typescript` version, `platform/arch`, the Go toolchain's content, the Go build environment, and the plugin's own source bytes matches exactly. A restored cache that doesn't match simply misses and rebuilds. A badly keyed cache can cost you a slow build, but it cannot hand you a wrong binary.

### 1. Yes, it's platform specific, and `runner.os` alone isn't quite enough

The cached artifact is a compiled native executable. `platform=${process.platform}/${process.arch}` is a cache-key input, and so are `GOOS` and `GOARCH`, so the cache is specific to OS *and* architecture. The full input list is in [Architecture, Cache Key Inputs](https://ttsc.dev/docs/development/reference/architecture#cache-key-inputs).

`runner.os` only yields `Linux`, `macOS`, or `Windows`, which no longer separates runners now that arm64 Linux runners exist and `macos-14`+ is arm64 while `macos-13` is x64. So include `runner.arch` as well:

```yaml
key: ${{ runner.os }}-${{ runner.arch }}-ttsc-${{ hashFiles('package-lock.json') }}
```

If two jobs of different architecture share one key, nothing breaks, but the first to finish wins the save, and every job on the other architecture then restores binaries it cannot use, misses on the internal key, and rebuilds cold on every run. You pay for the cache without getting anything from it.

### 2. No, it doesn't need saving every time, and the default behavior already gives you what you want

`actions/cache@v4` only writes in its post step when the primary `key` missed on restore. An exact hit means no save at all. So keying on your lockfile hash already gives you the policy you described: the cache is written only when a dependency actually moved.

On which of the four packages you named actually matter:

- `ttsc`, yes. It is an explicit version input and also contributes through the shim overlay source hash.
[...]

PR #1067, opened 2026-08-04
docs(ttsc): document the build cache for CI and containers

samchon:
## Intent

Answer the CI cache questions in #1064 in the documentation rather than only in the issue thread, and fix the three defects in the existing recipe that the questions exposed.

## Scope

Adds `website/src/content/docs/ttsc/cache.mdx`, a Build cache guide for anyone wiring `ttsc` into CI or a container image:

- what the cache holds, and that it is content-addressed, so a CI cache key affects hit rate and never correctness;
- why the key is specific to OS **and** architecture (`platform=${process.platform}/${process.arch}` plus `GOOS`/`GOARCH`);
- which dependency updates actually invalidate a plugin binary, including that `@ttsc/unplugin` does not, since it contributes nothing to the Go build;
- the `npm ci` ordering trap, worked npm and pnpm workflows, when to relocate the root with `TTSC_CACHE_DIR`, the garbage-collection asymmetry between the default and an explicit root, and how to confirm the cache actually hits;
- the container recipe, moved from the compile page unchanged.

Defects fixed, all from `ttsc/compile.mdx`:

1. The cache key used `runner.os` with no `runner.arch`. An architecture-spanning matrix shares one entry, the first job to finish wins the save, and every job on the other architecture restores unusable binaries and rebuilds cold on every run. Nothing errors, so the cost is invisible.
2. The key hashed `pnpm-lock.yaml` only, which is no help to an npm or yarn or bun project.
3. `TTSC_CACHE_DIR` was presented as the only way to persist the cache. The default `node_modules/.cache/ttsc` works when the cache step runs after `npm ci`, and relocating costs automatic pruning, which the page never said.

The root cause of all three was placement rather than wording: the cache key definition lives in `development/reference/architecture`, filed under plugin authoring, so a workflow author had nothing to reason from. The new page links there instead of duplicating it.

## Anchor stability

`packages/ttsc/src/plugin/internal/buildSourcePlugin.ts:272` prints `https://ttsc.dev/docs/ttsc/compile#plugin-cache` at runtime, and that string is baked into every released version. `compile.mdx` therefore keeps its `## Plugin cache` section and anchor, with the durable facts intact (locations, overrides, `ttsc clean`) plus one link onward to the new guide. Only the two `###` subsections below it moved.

`setup/unplugin.mdx` repointed to the new page, since its sentence is specifically about persisting the binary cache.

## Verification

Run locally against the branch:

- `prettier@3.8.3 --parser mdx --prose-wrap never --print-width 80 --embedded-language-formatting off --check` on all three changed MDX files: clean. Run with explicit options because `prettier.config.js` requires plugins not installed in this checkout; the options match the `*.md`/`*.mdx` override in that config.
- Every `/docs/...` link in the changed files resolved to an existing page, and every anchor to an actual heading in the target file.
- Repo-wide grep confirmed nothing references the two moved anchors (`#persisting-the-cache-across-split-ci-jobs`, `#persisting-the-cache-in-containers`), and that the runtime `#plugin-cache` anchor still resolves.

Not run: the website production build, and the rest of the repository test suite. This change touches MDX content and one `_meta.ts` entry only, with no source or behavior change.

## Deferred

Pre-existing em-dashes in `setup/unplugin.mdx` (lines 202, 204, 206) violate the documentation skill's voice rule. Left alone as unrelated to this change.

Answers #1064. Not using a closing keyword: that thread is a question, and the reporter may have follow-ups once the page is live.

https://claude.ai/code/session_012b4m4XtJk5jiHxqMofPTix


samchon (commented):
## Overall Self-Review

Solo, whole-diff, base-to-head. Three rounds; the third was clean.

### Round 1, content lost in the move

Moving the CI and container recipes off `compile.mdx` dropped three facts that the old page carried:

1. `cacheableRoots` adds the Go object cache as a second entry when `TTSC_GO_CACHE_DIR`, or a user `GOCACHE`, puts it outside the ttsc root. Without this a reader who redirects the object cache persists one directory and silently leaves the other behind. Verified against `packages/ttsc/src/launcher/internal/runTtsc.ts:274-277`, where `cacheableRoots` appends `goBuildRoot` only when `isPathWithin(paths.goBuildRoot, paths.root)` is false.
2. The old page stated the recipe works the same for pnpm, yarn, bun, and npm workspaces. The new page covered npm and pnpm only, narrowing claimed support without cause.
3. The container section had its own cold-build verification pointer. It was gone.

### Round 2, structure

4. The restored paths guidance sat under `### When to move the cache root`, so a reader staying on the default root would never reach the `GOCACHE` caveat, which applies to both layouts. Moved to its own `### Which paths to persist` directly after the first workflow example.
5. Cache verification was a `###` under `## GitHub Actions` while the container section needed it too. Promoted to a top-level `## Confirm the cache is working` covering both, and its guidance now splits by CI and container build.
6. The pnpm example began with a bare list at six spaces of indentation and no parent key, so it was not valid YAML standalone and would not paste cleanly. Rewritten to start from `steps:` at column zero. `embeddedLanguageFormatting: off` means Prettier cannot catch this class, so the other four blocks were checked the same way and all start at column zero.

### Round 3, voice

7. One throat-clearing preamble ("This has a consequence worth stating before...") removed in favour of stating the claim directly, per the documentation skill's "state the fact and stop".

Round 3 otherwise found nothing. Checked for the skill's banned constructions across the new page: no em-dashes, no emoji, none of the listed AI-cliche phrasings, one source line per paragraph.

### Anchor stability, rechecked

`packages/ttsc/src/plugin/internal/buildSourcePlugin.ts:272` prints `https://ttsc.dev/docs/ttsc/compile#plugin-cache` at runtime, and that string ships inside every released version, so the anchor cannot move. Confirmed `## Plugin cache` survives on `compile.mdx` with the locations, overrides, and `ttsc clean` facts intact. Repo-wide grep found nothing referencing the two `###` anchors that did move.

### Verification

- Prettier, run with the `*.md`/`*.mdx` override options explicitly because `prettier.config.js` requires plugins absent from this checkout: clean on all three MDX files.
- Every `/docs/...` and same-page `#` link in the changed files resolved to an existing page and an existing heading, rechecked after each round.
- Not run: the website production build and the repository test suite. This change is MDX content plus one `_meta.ts` entry, with no source or behavior change.

Issue #1071, opened 2026-08-04
moduleFormat() does not mirror tsgo: unset module, preserve, node20, and amd/umd/system are all misclassified


samchon:
## Problem

`ttsx` decides how to execute a served emit with `moduleFormat()` in
`packages/ttsc/src/launcher/internal/runtimeHooks.ts`. Its own doc comment
states the contract correctly — classify "exactly as tsgo chose when it emitted"
— but the implementation does not mirror tsgo, and diverges in four distinct
ways. Every divergence is a hard runtime failure, because the emitted JavaScript
and the format Node is told to use disagree.

The most damaging case needs no unusual configuration at all: **a
`tsconfig.json` that simply omits `module`, in a package with no `"type"`
field.** That is the most common TypeScript project shape there is, and `ttsx`
fails on it outright.

## Evidence

All four cases reproduced against `v0.24.0` (`edcea3e0c`) on win32-x64,
Node v24.18.0, `typescript@7.0.2`. In every case the emitted file is shown
alongside the failure, so the mismatch is directly visible.

### 1. `module` unset

`moduleFormat()` treats an empty module option as "defer to the nearest
`package.json` `type`". tsgo does not: `GetEmitModuleKind()` derives the emit
kind from `target`, and TypeScript 7's default target is the latest standard, so
an omitted `module` emits ES modules.

```jsonc
// package.json: no "type"
// tsconfig.json
{ "compilerOptions": { "strict": true, "outDir": "lib", "rootDir": "src" }, "include": ["src"] }
```

```
$ npx ttsx src/index.ts
SyntaxError: The requested module './dep' does not provide an export named 'dep'
    at ModuleJobSync.runSync (node:internal/modules/esm/module_job:528:17)
    at loadESMFromCJS (node:internal/modules/cjs/loader:1690:24)

$ npx ttsc -p tsconfig.json && cat lib/dep.js
export const dep = 1;
```

The emit is an ES module; `ttsx` runs it as CommonJS.

The reverse direction is broken too: `target: "ES5"` with no `module` makes tsgo
emit CommonJS, which `nearestPackageType()` classifies as `"module"` inside a
`"type": "module"` package.

### 2. `module: "preserve"`

`moduleFormat()` sends `preserve` to `nearestPackageType()`. tsgo preserves ESM
syntax under `preserve` regardless of the package type:

```
$ npx ttsx src/index.ts       # module: preserve, moduleResolution: bundler, no "type"
[...]

Issue #1070, opened 2026-08-04
ttsx cannot run an entry outside the project's include, so no project script or config file is runnable


samchon:
## Problem

`ttsx` cannot run a TypeScript entry that its owning `tsconfig.json` does not
include. A repository whose `tsconfig.json` sets `rootDir` / `include` to `src`
so that `ttsc` emits only `src` into `lib` — the correct and documented layout —
cannot run any of its own out-of-tree scripts:

```
$ npx ttsx clear.ts
ttsx: emitted entry not found for D:\project\clear.ts
```

This is the normal shape of a real project:

```
build/release.ts     <- must run under ttsx, must NOT land in lib/
clear.ts             <- must run under ttsx, must NOT land in lib/
lint.config.ts       <- must NOT land in lib/
nestia.config.ts     <- must NOT land in lib/
src/**/*.ts          <- the only thing ttsc compiles into lib/
package.json
tsconfig.json
```

`ttsc -p tsconfig.json` already behaves correctly here: it emits `lib/index.js`
and nothing else, and the out-of-tree files are neither type-checked nor
compiled. That half of the contract is right and must stay right. The broken
half is that `ttsx build/release.ts` and `ttsx clear.ts` must still run, under
that same `tsconfig.json`'s compiler options — this is precisely the difference
between `ttsc` and `ttsx`, and `ttsx` currently gets it wrong.

## Evidence

Reproduced against `v0.24.0` (`edcea3e0c`) on win32-x64, Node v24.18.0,
`typescript@7.0.2`.

```jsonc
// package.json — no "type" field
{ "name": "repro", "version": "1.0.0", "private": true }
```

```jsonc
// tsconfig.json
{
  "compilerOptions": {
    "target": "ESNext",
    "module": "NodeNext",
    "moduleResolution": "NodeNext",
    "strict": true,
    "outDir": "lib",
    "rootDir": "src",
    "types": ["node"]
  },
  "include": ["src"]
}
```

```ts
// src/index.ts
console.log("src dirname:", __dirname);
// clear.ts
console.log("root dirname:", __dirname);
```

| Command | Observed |
| --- | --- |
| `npx ttsc -p tsconfig.json` | ✅ emits `lib/index.js` only |
| `npx ttsx src/index.ts` | ✅ `src dirname: D:\repro\src` |
| `npx ttsx clear.ts` | ❌ `ttsx: emitted entry not found for D:\repro\clear.ts` |

### Root cause

[...]

Issue #1069, opened 2026-08-04
@ttsc/banner and @ttsc/strip config loaders repeat #1068: ESM-only .ts evaluation and a Program with no @types


samchon:
## Problem

`@ttsc/banner` and `@ttsc/strip` synthesize the same ephemeral config-loader
`tsconfig.json` that #1068 reports for `@ttsc/lint`, with the same two defects
baked in:

1. `module` is hardcoded to `"ESNext"`, so a `banner.config.ts` / `strip.config.ts`
   is executed as an ES module even when the owning package is CommonJS. Any use
   of `__dirname`, `__filename`, or `require` throws
   `ReferenceError: __dirname is not defined in ES module scope`.
2. The loader Program declares no `types` and no `typeRoots` and does not
   `extends` anything, so under TypeScript 7 it contains **zero** ambient type
   packages. `@types/node` is invisible and the config fails to type-check with
   `TS2304: Cannot find name '__dirname'` before it ever runs.

#1068 fixes the `@ttsc/lint` witness only. This issue covers the rest of the
class so the same cause cannot keep producing failures in sibling packages.

## Evidence

The loader-tsconfig factory is copied verbatim across four sites:

| Package | Site | Hardcoded `module` |
| --- | --- | --- |
| `@ttsc/lint` (JS factory) | `packages/lint/src/index.ts` → `evaluateTtsxConfigPlugins` | `"ESNext"` |
| `@ttsc/lint` (Go host) | `packages/lint/linthost/config.go` → `typeScriptConfigLoaderTsconfig` | `"ESNext"` |
| `@ttsc/banner` (Go driver) | `packages/banner/driver/banner.go` → `typeScriptConfigLoaderTsconfig` | `"ESNext"` |
| `@ttsc/strip` (Go driver) | `packages/strip/driver/config.go` → `stripTypeScriptLoaderTsconfig` | `"ESNext"` |

All four list the user's config file in `files`, so the config becomes part of
the loader Program and inherits its `module` option. The runtime classifier
`moduleFormat()` in `packages/ttsc/src/launcher/internal/runtimeHooks.ts` maps
every non-`node*` / non-`preserve` module target to `"module"` unconditionally,
so `nearestPackageType()` — the function that mirrors Node's own `package.json`
`"type"` walk — is unreachable for an ambiguous `.ts` config.

Reproduced against `v0.24.0` (`edcea3e0c`) on win32-x64 with `typescript@7.0.2`,
[...]

Issue #1068, opened 2026-08-04
@ttsc/lint evaluates a .ts lint config as ESM regardless of the package's "type", and its loader Program cannot see @types/node


samchon:
## Summary

A `lint.config.ts` that lives in a **CommonJS** package (no `"type": "module"` anywhere in the tree) is evaluated by `@ttsc/lint`'s config loader as an **ES module**, so `__dirname` throws at runtime. Separately, the ephemeral loader `tsconfig.json` that type-checks the config inherits nothing from the owning project and declares no `types`, so the config cannot see `@types/node` by any project-level configuration.

The two defects are independent and stack: fixing the first error reveals the second.

Notably, `moduleFormat()` already contains the package-type lookup needed to get this right — it is simply unreachable for a `.ts` config because the loader tsconfig hardcodes `module: "ESNext"`. And `lint.config.cts` **does** work, which shows the CommonJS lane is functional and only the ambiguous `.ts` extension is misclassified.

## Environment

| | |
| --- | --- |
| `ttsc` / `@ttsc/lint` | **0.24.0** and **0.22.0** (both reproduce, independently) |
| `typescript` | 7.0.2 (`@typescript/typescript-win32-x64`) |
| `@types/node` | 26.1.2 |
| Node | v24.18.0 |
| Platform | win32-x64 |

## Minimal reproduction

```bash
mkdir ttsc-lint-cjs-repro && cd ttsc-lint-cjs-repro
npm init -y
npm install ttsc@0.24.0 @ttsc/lint@0.24.0 @types/node typescript
mkdir src
```

`package.json` — note there is **no** `"type"` field, here or anywhere above it. By Node's own rules a `.ts`/`.js` file in this tree is CommonJS.

```json
{
  "name": "ttsc-lint-cjs-repro",
  "version": "1.0.0",
  "private": true
}
```

`tsconfig.json` — `include` is `["src"]`, so `lint.config.ts` is deliberately **outside** the package Program (the documented layout: the config sits next to `tsconfig.json`).

```json
{
  "compilerOptions": {
    "target": "ESNext",
    "module": "NodeNext",
    "moduleResolution": "NodeNext",
    "strict": true,
    "outDir": "lib",
    "rootDir": "src",
    "skipLibCheck": true,
    "plugins": [{ "transform": "@ttsc/lint", "configFile": "./lint.config.ts" }]
  },
  "include": ["src"]
}
```

`lint.config.ts`

```ts
[...]

PR #1072, opened 2026-08-04
fix: mirror tsgo's module-format decision and target files outside the project include

samchon:
Unified issue-campaign cycle for module-format classification and tsconfig
targeting of TypeScript sources that the owning project's `rootDir` / `include`
does not contain.

## Cycle scope

| Issue | Subject |
| --- | --- |
| #1071 | `moduleFormat()` does not mirror tsgo: unset `module`, `preserve`, `node20`, `amd`/`umd`/`system` |
| #1068 | `@ttsc/lint` evaluates a `.ts` config as ESM regardless of the package's `"type"`, and its loader Program cannot see `@types/node` |
| #1069 | `@ttsc/banner` and `@ttsc/strip` config loaders repeat #1068 |
| #1070 | `ttsx` cannot run an entry outside the project's `include` |

Implementation order follows that dependency DAG: #1071 first, because the
config loaders (#1068, #1069) and the `ttsx` entry lane (#1070) both hand their
inputs to that classifier, and their boundary cases would otherwise pass for the
wrong reason.

## Discovery record

Two complete full-scope discovery rounds ran against `edcea3e0c` before this
branch was claimed. Round 1 produced six accepted candidates (published as the
four issues above); round 2 was empty, which opened the development gate.

## Verification

Local reproduction of every defect was performed against a
`TTSC_BUILD_SCOPE=test-lint` build of `edcea3e0c` before implementation.
Acceptance for the implementation itself is this pull request's ordinary CI plus
a clean solo Overall Self-Review; no local build or test run is claimed as
evidence for the fix.

## Deferred

Nothing deferred. The full accepted set is implemented in this pull request.


samchon (commented):
## Individual Self-Review — `508dbd1c7`

`fix(ttsx): mirror tsgo's emit-format decision and evaluate configs under Node's rule`
Resolves #1071, #1068, #1069, #1070 (source only; tests and docs land in `ec4b6677d`).

A read-only reviewer took the parent-to-commit diff and traced it against the pinned
typescript-go (`v0.0.0-20260429010842-56ab4af42157`). Adjudication of every candidate below.

### Accepted and fixed

**Loader `module: "CommonJS"` versus top-level `await` (blocker).** All four config loaders
generate a `loader.mts` whose body is a top-level `await`, and TS1378 keys off the
program-wide `module` option rather than the file's emit format — so a `.mts` that still
emits as ESM is rejected the moment the option says CommonJS, which is every config in an
ordinary package. Confirmed independently by running the original #1068 reproduction against
a locally built compiler, which failed with TS1378 at `loader.mts(139,7)`. Fixed in
`f071b5952` by wrapping each loader body in an async IIFE; the same reproduction now exits 0.

The reviewer's suggested alternative — `module: "preserve"` — was **rejected**. Under
`preserve` tsgo emits CommonJS only when the manifest states `"type": "commonjs"` outright
(`getImpliedNodeFormatForEmitWorker` tests `PackageJsonType == "commonjs"`), and the packages
this issue is about omit `type` entirely. Verified by compiling a `preserve` project: the
emitted `lib/dep.js` was `export const dep = 1;` in a package with no `type`.

**`entryEmitPath` on the resolve hook's path.** `owningModuleOptions` reaches
`resolveEmittedJavaScript` once per import specifier, and a miss there walks the whole emit
tree. Memoized in `f071b5952`.

**The `node_modules` package `"type"` override.** `getImpliedNodeFormatForEmitWorker`
consults `packageJsonType` for every module kind, not only the `node*` family, and
`packageJsonType` is populated for paths under `node_modules`. The classifier's comment
asserted the opposite. Fixed in `f071b5952` with `declaredNodeModulesPackageType`, plus an
e2e that fails without it.

**`entryModuleFormat`'s `?? null` and the unreadable-tsconfig `{}`.** Both broke the
`null` versus `{}` distinction this commit introduced. Fixed in `f071b5952`.

**Go `nearestPackageType` and relative paths.** `filepath.Dir` of a relative path ends the
walk at `.` after one step. Fixed in `f071b5952` with `filepath.Abs`.

**The fuzzy matcher used as an ownership predicate.** `ttsx build/release.ts` could match the
`release.js` emitted for an unrelated `src/release.ts` and run the wrong file. Fixed in
`f071b5952` by applying the same `isWithin` guard `serveEntryEmit` uses.

### Accepted with a wider fix than reported

The reviewer noted `amd`/`umd`/`system` are unreachable in TypeScript 7. Checking
`internal/compiler/program.go`'s "Removed in TS7" block showed the same is true of
`target: ES5`, `baseUrl`, `outFile`, and `moduleResolution: Classic`. Both dead branches were
removed from the classifier in `f071b5952` rather than left as untestable code. This also
means **#1071's own claim about `amd`/`umd`/`system` was wrong**: they are not misclassified,
they are rejected before emit.

### Noted, not changed

- **Commit granularity.** Tests and documentation landed in the next commit rather than this
  one. Accepted as a branch-level property of the campaign's DAG order.
- **`amd`/`es3` comment inaccuracies.** Resolved by deleting the branches.

### Confirmed clean by the reviewer

The classifier matrix against `GetEmitModuleKind` / `GetEmitScriptTarget` row by row;
`moduleResolution: "bundler"` with `module: "commonjs"` being legal in tsgo (unlike TS 5.9);
`types: ["*"]` being genuinely supported by `UsesWildcardTypes`; the removal of the child
pre-check in `shouldExposeCommonJsNamedExports` (the authoritative re-check in
`resolveRuntimeSource` makes over-inclusion inert, and `commonJsNamedInteropUrls` has exactly
one reader); `ENTRY_PROJECT_EMIT_DIR` not colliding with the virtual layout's volume labels;
`files` + `include: []` + `exclude: []` displacing the inherited file set; and the
`projectRoot: context.root` → `context.runtimeRootDir` change carrying no regression.

samchon (commented):
## Individual Self-Review — `ec4b6677d`

`test(ttsx): pin the module-format matrix and out-of-include entries`

A read-only reviewer traced every new test against the pinned typescript-go and the test
harnesses. Adjudication below.

### Accepted and fixed — three tests could not have passed

**`module: "umd"` is removed in TypeScript 7**, not merely misclassified.
`internal/compiler/program.go` raises `Option_0_1_has_been_removed` for AMD, System, and UMD,
and `IgnoreDeprecations` is parsed but never consumed. `test_ttsx_classifies_module_umd_as_commonjs`
would have failed at the build gate without reaching the classifier. Deleted in `f071b5952`,
along with the now-unreachable `amd`/`umd`/`system` arm it was written for.

**`target: "ES5"` is removed too**, and `ScriptTargetES3` no longer exists. The ES5 test is
deleted and its purpose — proving the target-derived kind outranks the manifest — is now
carried by a `"type": "commonjs"` fixture folded into
`test_ttsx_classifies_an_unset_module_option_from_the_target`, which is a case TypeScript 7
still supports and which genuinely fails without the fix.

**`baseUrl` is removed (TS5102)**, which this repository already documents in
`tests/test-unplugin`. `test_ttsx_runs_an_excluded_entry_under_the_project_compiler_options`
declared it unnecessarily; `paths` with relative targets is the only form TS7 accepts, and the
test now uses that alone.

**Same test, second defect:** its `globals.d.ts` sat outside the synthesized entry program
(`files: [entry]`), so `__dirname` would not resolve. The declaration moved into the entry.

### Accepted and fixed — test quality

**Tautological wildcard-`types` assertions.** Three Go tests asserted the loader emits the
literal `types: ["*"]` three lines from where it writes it. Deleted; the behaviour is proven
end to end by the lint e2e that type-checks `__dirname` with no triple-slash directive, which
only passes because the wildcard makes `@types/node` visible.

**The "CommonJS package" Go fixture did not declare CommonJS.** `{"name":"cjs"}` has no
`"type"`, so under the walk as written it fell through to the root fallback — the same path as
the no-manifest row, asserted twice under different labels, and dependent on the machine's real
directory tree above the temp root. Rewritten in `f071b5952` into a deterministic matrix that
covers each step of Node's lookup, including nested manifests overriding an ancestor in both
directions. Investigating this is what surfaced the deeper bug: the walks did not match Node
at all, since Node stops at the first manifest and reads a missing `"type"` as CommonJS.

### Accepted, recorded rather than changed

**"Every case here failed before the fix" overstates it.** The `import.meta` lint test and the
`node20` module half are anti-constant-swap twins: they pass before and after, and exist so a
fix cannot be a swapped hardcode. That is a deliberate requirement of the development skill's
"a negative twin for every positive", but the commit message should not have claimed otherwise.
Recorded here rather than by rewriting a pushed message.

**`test_ttsx_reports_a_type_error_in_an_entry_the_project_include_excludes` is loose** — it
asserts the entry is named and that the internal "emitted entry not found" is gone, not the
specific diagnostic. Left as is: pinning TS2322's exact rendering would couple the test to
compiler message text.

### Coverage gaps accepted as out of scope

`module: "none"`, the `es6` target spelling, and `.cts`/`.mts` precedence over a contradicting
project `module` remain unpinned by a dedicated case. The first two are spellings tsgo rejects
or normalizes before emit; the third is covered indirectly by the existing `.cts` and `.mts`
entry tests. The dependency-cache marker gap the reviewer noted is closed by `b44202562`.

### Confirmed clean

One export per file with names matching filenames; correct `TestProject` and `TestLint` API
use, including that `extraSources` accepts a root `package.json` and that
`linkNodeModules: ["@types/node"]` resolves; `module: "node20"` + `moduleResolution: "node16"`
and `module: "preserve"` + `moduleResolution: "bundler"` both legal; extensionless ESM imports
in the `preserve` fixture rescued by the resolve hook; declaration files exempt from
`moduleDetection: force`, so the `node20` fixture's global stays global; and the Go test
plumbing — linkname helpers, package clauses, and both runners picking the new files up with
no registration step.

samchon (commented):
## Individual Self-Review — `b44202562`

`fix(ttsx): bound the entry-project emit and reject an undescribed cache marker`

Adjudication of the read-only reviewer's findings.

### Accepted and fixed

**`commonAncestorDirectory` degrades to the volume root on ordinary hosts.** `context.root`
arrives realpath-resolved from `resolveProjectIdentity` while `path.resolve(entryFile)` does
not, and the walk compared them with raw `path.relative`. On macOS the temp root is
`/var/folders/...` against a canonical `/private/var/folders/...`, so containment never holds
and the loop reaches `/` — reintroducing exactly the bound this commit set out to remove. The
Windows twin is the drive letter: `path.parse().root` compares case-sensitively while
`path.relative` on win32 does not, which this repository's own
`test_iswithin_matches_roots_and_slash_normalized_rootdirs` documents as a real condition.
Fixed in `f071b5952` by asking containment through the same filesystem-identity predicate the
runtime hooks use to consume the value, which also retires the third hand-rolled containment
helper (`isWithinDirectory`) and its `..foo` prefix bug.

**`readDependencyCache` accepted an array as a marker's `moduleOptions`.** Cosmetic and
unreachable from any writer, but the check exists precisely to reject "silence read as no
options". Tightened in `f071b5952`.

**A stale comment.** `owningModuleOptions` still cited "the volume-root `rootDir` a
config-loader **or out-of-`include` entry** project uses" after this commit narrowed the
latter. Corrected in `f071b5952`; the config-loader lane genuinely still publishes a
volume-root `rootDir`, which is why the emit-membership ownership test matters there.

### Accepted as a correction to this commit's own message

The message claimed the narrowed `rootDir` fixes a `node_modules/pkg/src/index.ts` mis-serve.
It does not: `node_modules` sits under the project root, so it is inside the new bound and the
trailing-stem matcher still reaches it. What the change actually buys is excluding `.ts` files
**outside** the project root from that matcher. Recorded here rather than by rewriting a
pushed commit message. The narrower ownership question — "did this project emit this entry" —
is separately fixed in `f071b5952` by the `isWithin` guard in `emittedEntryOf`.

### Rejected

**"The new `rootDir` can fail with TS6059 where the old one never could."** Not true for the
pinned compiler. `checkSourceFilesBelongToPath` is unimplemented in
`internal/compiler/program.go` — the site carries a `!!!` TODO marker — and grepping the whole
`internal` tree finds no emitter for `File_0_is_not_under_rootDir_1`. The in-repo comments the
reviewer cited (#299, #304) describe an older toolchain. An input outside `rootDir` gets an
output path with `..` segments rather than a diagnostic, and such a file falls through to the
dependency lane, which builds it under its own project.

### Noted, out of scope

The fallback dependency cache under the system temp directory is keyed by
`sha256(tsconfig path)` alone — not by version, content, or mtime — so an edited dependency
source is never rebuilt once a marker exists. Pre-existing, same root cause as the marker
staleness this commit fixes, but a separate defect that belongs in its own issue.

### Confirmed clean

The returned root always contains the entry, on every branch including the cross-volume
fallback; the loop cannot spin (verified `dirname` fixpoints for `C:\`, `/`, and UNC);
manifest and emit agree for every input under the root; `runtimeHooks.ts` has no module-level
executable statements and was already value-imported by `runTtsx`, so the new value import
adds no side effect and no cycle; `publishDependencyMeta`'s single call site can never write a
marker the new validation rejects, and rejecting a legacy one costs exactly one rebuild with
no livelock; the fixture migration was required rather than cosmetic, since the old markers
would now be rejected and the test would fail; and `projectModuleOptions` is semantically
identical to the `emitModuleOptions` it replaced.

samchon (commented):
## Individual Self-Review — the deadline and output-ceiling removals

Covers `32e20400b` through `c9e975142`, reviewed as three overlapping passes as the
work landed. Adjudication of every candidate below.

### What this range does

The repository owner rejected two classes of limit as uncontracted and
machine-dependent: the deadlines that killed a user's own config, descriptor, or
resident request, and the `maxBuffer` ceilings that killed a build for printing too
much. Both are gone — and the ceilings are gone because the *collection* they
bounded is gone, not because the number was raised.

Human output now streams to the parent's stderr as the child writes it. Payloads a
parent must read go to a file descriptor rather than a pipe, so no byte passes
through this heap on its way out of the child and there is nothing left to bound.

### Accepted and fixed — defects that would have shipped

**The `.mts` loaders could not compile.** Each declares its own ambient `process`
with a one-argument `stdout.write`; adding a flush callback made every TypeScript
banner and strip config fail with TS2554 under the loader's own `strict`. Fixed by
admitting the callback in the declaration — and verified by extracting the real
generated `loader.mts` for both packages and type-checking it against the pinned
compiler: no diagnostics at all.

**The failure envelope was unreachable in all three `.mts` loaders.** The inner
`catch` ends the process, so the trailing `.catch` that carried the reason could
never run. Only the CommonJS loaders delivered it — which is exactly the half the Go
tests cover, so nothing caught it. The envelope moved into the catch that actually
runs.

**`process.exit` abandons a pending pipe write**, which is where the envelope
travels for banner and strip. Each loader now sets `exitCode` before the write and
exits from its completion callback, so a callback that never fires still fails the
load rather than exiting zero.

**Two cleanup paths could replace a result with their own error.** The ttsx
descriptor evaluation and the lint config evaluation both `fs.rmSync` from a
`finally`; on Windows a scanner or an inherited handle turns that into an EBUSY that
surfaces in place of a successful build. Same treatment as the capture helper.

**The plugin-descriptor loader had no reason channel at all.** It lost the stderr
tail with the rest of the collection and reported only an exit status. It writes to a
private result file already, so the reason travels there now.

**`captureProcessOutput` leaked on a failed second `openSync`** and could throw from
`dispose()`. Both fixed, in both copies.

**A `paths` claim the product does not honour.** An excluded-entry test asserted a
path alias resolving at run time. It never could — `paths` is a compile-time mapping
tsgo does not rewrite into the emit — and an entry *inside* `include` fails
identically. Confirmed by running both.

**A test that pinned nothing.** Its `strict` case compiled identically with and
without `strict`, verified both ways. `strict` moved to the twin, where its entry's
only error is one `strict` alone produces, so an entry project compiled under
defaults would run the script instead of stopping it.

### Accepted as corrections to this range's own claims

- A commit message called the 16 MiB ceiling uncontracted. `protocol.mdx` did
  promise it. The page is corrected and the correction recorded.
- `captureProcessOutput` claimed "there is nothing to bound". Reading a result
  materializes a string, so V8's maximum string length is still an outer edge — a
  property of the runtime, identical everywhere, rather than a budget chosen here.
- Stale comments retired: a raised graph buffer, a lint forwarding block, a
  heartbeat that no longer exists, and a `finally` rationale for loaders that have no
  `finally`.

### Rejected

**"The new `rootDir` can fail with TS6059."** Not for the pinned compiler:
`checkSourceFilesBelongToPath` is an unimplemented TODO and no emitter for that
diagnostic exists anywhere in the tree.

**"`module: preserve` would keep the CommonJS lane."** Under `preserve` tsgo emits
CommonJS only when a manifest states `"type": "commonjs"` outright, and the packages
this issue is about omit `type` entirely — confirmed by compiling one.

### Kept deliberately

The 16 MiB output caps are gone, but two bounds remain and are not the same thing.
The plugin-build lock steal budget is how a dead lock holder is reclaimed; removing
it replaces a bounded wait with a permanent one and deletes the recovery path. And
V8's string limit is the runtime's, not ours.

### Verified by running it, not by reading it

A config that throws now streams its own stack to the user's stderr while it runs,
and the build exits non-zero. A healthy one exits 0. `ttsc -p` still emits only what
`include` selects; `ttsx clear.ts` and `ttsx build/release.ts` run. The
banner, paths, and strip Go suites pass.

samchon (commented):
## Overall Self-Review — round 1 (`f943f6bb3` → `9e5a61cd2`)

Solo, whole base-to-head diff (85 files). Round 1 found one executable defect and
four stale claims, all applied in `9e5a61cd2`. A fresh complete round follows.

### CI gate

`lint defense 1` was the only red lane at `f943f6bb3`
(`ci` is the aggregator). `test_descriptor_rejects_malformed_cjs_contributors`
failed because the `@ttsc/lint` descriptor's plugin-entry extractor was the one
loader in this pull request that never received a failure envelope: four of five
got one in `65aea694d` / `c9e975142`, and this one still reduced
`contributor "demo" must resolve to an object with a non-empty "source" string`
to a bare exit status. `446b4e5b6` closes that gap and adds the negative twin —
a config that ends the process itself writes no envelope and must keep the bare
status.

### Accepted and fixed

**Dead timeout branch in `runWithStdin`** (`packages/ttsc/internal/lspserver/lsp_native_plugin_source.go`).
`context.WithTimeout` became `context.WithCancel`, but `if ctx.Err() != nil { …
"timed out" }` survived after `cmd.Run()`. Nothing cancels that context before
`Run` returns, so the branch is unreachable — dead code that reports a removed
feature at the exact place a reader checks whether LSP plugin commands are still
bounded. Removed; the context's remaining purpose (reaping the child on an early
return) is now stated.

**`TtscGraphSessionOptions` kept the doc comment of its deleted field.** The
`requestTimeoutMs` property was removed but its `/** Maximum time … Defaults to
five minutes … */` block stayed, trailing the interface and documenting nothing —
while `website/src/content/docs/graph/index.mdx` now states that a snapshot
request has no deadline. Removed.

**Both typia pack loaders still promised a deadline.**
`loadTypiaSourcePack` ("A caller abort or deadline cancels that shared attempt …
The deadline covers both the response and its JSON body") and
`loadTypiaRuntimePack` document exported functions whose deadline is gone.
Rewritten to say what remains true: only a caller abort ends the shared load.

**`new TtscService(...)` example kept the shape of its removed argument.**
`programmatic-api.mdx` still spread a now-single-argument call across three lines
with a dangling comma. Collapsed.

**`declaredNodeModulesPackageType` justified itself with an over-broad upstream
rule.** The comment claimed tsgo populates `packageJsonType` only under
`node_modules`. `loadSourceFileMetaData` also fills it for an extensionless file
whose project sets a `node16`-family `moduleResolution` — the conclusion still
holds, because `program.go` rejects that resolution unless `module` is in the
same family, and a `node*` `module` reaches the package-`type` answer anyway.
The reason now matches the source.

### Verified against the pinned typescript-go, not the commit messages

`v0.0.0-20260429010842-56ab4af42157`, read from the module cache:

- `ast.GetImpliedNodeFormatForEmitWorker` does consult `PackageJsonType` for
  every module kind, and only when it agrees with `ImpliedNodeFormat` — which is
  exactly the shape `moduleFormat` mirrors (declared `"type"` wins; a silent
  manifest does not).
- `compiler/program.go` emits `Option_module_must_be_set_to_0_when_option_moduleResolution_is_set_to_1`,
  which is what makes the `node_modules` gate complete rather than a guess.
- `GetEmitModuleKind` derives the kind from `target` when `module` is absent, so
  publishing `target` alongside `module` in the runtime manifest is required, not
  decoration.

### Rejected suspicions

- **`commonAncestorDirectory` argument order.** `FilesystemPathIdentityContext.isWithin(root, candidate)`
  and `runtimeHooks.isWithin(real, directory)` take opposite orders; both call
  sites use their own correctly, and the walk returns the project root for the
  `build/release.ts` layout it exists for.
- **`shouldExposeCommonJsNamedExports` no longer checks the child's format.**
  The authoritative check is at the serve site (`format === "commonjs" &&
  commonJsNamedInteropUrls.has(url)`), so the dropped guess can only over-mark a
  URL that is then re-tested, never expose a file wrongly.
- **`entryEmitPathCache` is never invalidated.** It lives in the spawned runtime
  child, whose entry emit is written before the process starts and never changes
  under it.
- **Removing the Go-side `TTSC_LINT_DEBUG_CONFIG_GRAPH` re-forward.** The loader
  gates that write on the env var itself, so streaming the child's stderr already
  delivers it; the three Go tests that set the variable read it from the stream.
- **`spawnGoTool`'s `stdio: ["ignore", …]` and a caller's `input`.** No call site
  passes `input`.
- **`nativePluginCommandStdoutLimit` / `StderrLimit` surviving.** They truncate a
  message and say so; they never fail a build, so they are not the class of
  ceiling this cycle removed.

### Public API change, called out

`TtscServiceOptions` and `TtscService`'s second constructor argument are gone, as
are `TtscGraphSessionOptions.requestTimeoutMs` and the wasm boot deadline option.
This is the repository owner's explicit rejection of uncontracted, machine-chosen
limits, and every affected page under `website/src/content/docs/` is updated in
the same pull request.

samchon (commented):
## Overall Self-Review — round 2 (`9e5a61cd2` → `a5ef56d61`)

Fresh complete round over the whole base-to-head diff. Two findings, both in the
failure-envelope class this cycle introduced, both applied in `a5ef56d61`.

### Accepted and fixed

**The descriptor shim's `import` sat outside its own try.**
`packages/ttsc/src/plugin/internal/loadProjectPlugins.ts` generates
`load-descriptor.mts` as:

```
const mod = await import(pathToFileURL(process.env.TTSC_PLUGIN_ENTRY).href);
…
try { const descriptor = candidate(context); writeFileSync(out, …) }
catch (error) { …envelope…; process.exit(1) }
```

A descriptor whose module body throws, or whose own `import` names a missing
module, therefore never reaches the catch: no envelope is written and the caller
gets `ttsc: plugin descriptor "…" evaluation through ttsx failed with exit code
1` alone — while the same class of failure one line later yields a named reason.
`Cannot find module "./missing"` is exactly the kind of fact the envelope exists
to carry. The import, the context parse, and the candidate pick now sit inside
the same try. (Top-level `await` inside `try` is unchanged in an `.mts` module.)

**The envelope reader had no test on the ttsc side.**
`descriptorFailureReason` shipped in `c9e975142`;
`test_plugin_descriptor_failures_propagate_and_cleanup_ttsx_temp` drove only
`nonzero`, `stdout-nonzero`, `missing`, `malformed`, and `success` — every one of
which writes no envelope. The branch that exists to carry a reason was therefore
never executed by any test, so the whole feature could have been reverted with
the suite still green. Two fake-ttsx modes now cover it:

| mode | result file | asserts |
| --- | --- | --- |
| `envelope` | `{"__ttscLoaderError":"  the descriptor said why  "}`, exit 4 | `failed with exit code 4\nthe descriptor said why\n` — proves the join *and* the trim |
| `foreign-result` | `{"name":"not-an-envelope",…}`, exit 6 | status message present, `not-an-envelope` **absent** |

The second is the negative twin: without it, "only a well-formed envelope is
honoured" is an unproven claim, and any descriptor output could become the
failure message.

### Verified this round, no change needed

- **Envelope coverage is now complete across all five loaders.** banner and strip
  prove the positive direction through `resolve_banner_text_branches_test.go`
  (a real `.cjs`/`.ts` config whose validation message reaches the error) and the
  negative through `script_config_loader_test.go` /
  `typescript_config_loader_test.go` (a loader that only writes stderr must *not*
  have that text repeated in the error). The `@ttsc/lint` extractor proves the
  positive through `test_descriptor_rejects_malformed_cjs_contributors` and the
  negative through the new
  `test_descriptor_keeps_the_exit_status_when_the_evaluator_names_no_reason`.
- **`configModuleOption` returns `ESNext` for `.cts`/`.cjs`.** Not a defect:
  tsgo's `GetImpliedNodeFormatForEmitWorker` returns CommonJS for a `.cts`/`.cjs`
  file whose `ImpliedNodeFormat` is CommonJS regardless of the module option, and
  `GetImpliedNodeFormatForFile` assigns that from the extension. The option only
  has to decide the ambiguous `.ts`/`.js`/`.tsx` case, which is what it does.
- **`loader.mts` still emits ESM when the option resolves to `CommonJS`.** Same
  upstream rule from the other side (`.mts` → ESNext), which is why the loaders
  need the IIFE wrapper rather than a top-level `await` — TS1378 keys off the
  module option, not the emitted format.
- **`runTtsx` no longer deletes the evaluator-boundary environment variables.**
  Nothing writes them any more; `readEvaluatorBoundary` and its three variables
  are gone from the same file.
- **`BootCancellationReason` is now a single-member union.** `reason?.kind ===
  "abort"` still discriminates correctly and the timeout arm of
  `bootCancellationError` is gone with it.
- **`spawnNative`'s `stdio: ["ignore", …]` replaces a default `"pipe"` stdin.**
  Both give the child immediate EOF; no call site passes `input`.

### Known coverage limit, stated rather than hidden

The `@ttsc/lint` evaluator's own foreign-result case — a result file that parses
but is not an envelope — is not directly exercised, because that loader bakes its
output path into the generated source rather than passing it through the
environment, so a fake evaluator cannot address it cheaply. The shared shape of
the reader is covered by the descriptor `foreign-result` case above.

samchon (commented):
## Individual Self-Review — `446b4e5b6`

`fix(lint): give the config evaluator the reason channel every other loader has`

A read-only reviewer took the parent-to-commit diff and traced it against the pinned typescript-go and the ttsx exit-status chain. Adjudication below.

### Accepted and fixed (in `83fa83ed5`)

**The new test's comment claimed a branch it does not reach.** The negative twin exits from the config body, so the result file is never created and only `readFileSync`'s throw is exercised — the three shape guards inside `configEvaluatorFailureReason` were reached by no test in the tree, while the comment said the parent must not attach "whatever the file holds". Both halves are now true: the reader moved to `packages/lint/src/internal/configEvaluatorFailure.ts` beside `configEvaluatorProcessFailure`, and `test_config_evaluator_reason_honours_only_a_well_formed_envelope` pins all of its shapes — envelope, blank envelope, non-string envelope, payload, array, scalar, `null`, half-written, empty — at unit cost, instead of one of them costing a full ttsx build.

**The envelope had two spellings.** `__ttscLoaderError` in four loaders, `error` in the lint Go sidecar. Each writer/reader pair was self-consistent, so this was a trap rather than a live defect — but the two lint config loaders sat in one package doing one job with different keys, in a file that spends `"error"` on rule severity. Unified on the namespaced spelling.

### Rejected

**`configEvaluatorFailureReason` duplicating `descriptorFailureReason`.** Real, and not worth removing: `@ttsc/lint` must not reach into `ttsc`'s internals to share eleven lines. Recorded so the next edit to either is made to both.

### Confirmed clean

The emitted extractor still parses — no backslash, backtick, or `${` in the added lines, `quotesPair` and the control-character rule both hold, and all three placeholders survive. No TS2554: nothing was added to the loader's ambient `process` surface, unlike the banner/strip case `c9e975142` fixed. `return process.exit(1)` is well-typed against `exit(code?: number): never`. The envelope cannot be written on a success path, and the parent reads it only after `configEvaluatorProcessFailure` returned an error. Exit status 7 propagates end to end — `registerRuntimeHooks` does not intercept, `runTtsx` inherits stdio and returns `result.status ?? 1`, the launcher assigns it, and the config is never type-checked (`allowJs` without `checkJs`) so the build gate cannot remap it. A failed evaluation is never cached. And the class is closed rather than the witness: every loader that streams child stdio and reads a private result now carries an envelope.

samchon (commented):
## Individual Self-Review — `f071b5952` … `1a59a808d`

Covers the five commits whose individual pass was not yet on the record: `f071b5952`, `368cc7d38`, `1e6e9d3d9`, `bf61471cf`, `1a59a808d`. Every claim was re-verified against the pinned typescript-go rather than taken from the commit messages.

### Blocker — accepted and fixed in `83fa83ed5`

**The entry project emitted into the user's source tree.** `1a59a808d` resolved `commonAncestorDirectory`'s inputs through the filesystem-identity context so the manifest `rootDir` would match what `serveEntryEmit` compares against. It did not resolve `entry`, which is what goes into the synthesized tsconfig's `files`. The two spellings then came from different rules — the defect that commit set out to remove, moved one line down.

Upstream makes that fatal rather than cosmetic: `outputpaths.GetSourceFilePathInNewDir` strips `commonSourceDirectory` only when `ContainsPath` says the input is under it, and `commonSourceDirectory` is `options.RootDir` verbatim. `ContainsPath` is lexical. When it says no, `CombinePaths(outDir, absoluteSourcePath)` hits the `GetRootLength(trailingPath) != 0` branch and *replaces* `outDir` outright. The output is then the source path with the extension changed.

Failing input: any host where `path.resolve` and the physical path differ — macOS `/var/folders/...` against `/private/var/folders/...`, a Windows 8.3 `TEMP`, a symlinked checkout. `ttsx clear.ts` wrote `clear.js` and `clear.js.map` beside the user's `clear.ts`, plus one per relatively imported sibling, none of it under `processDir`, so `removeRuntimeOutput` never deleted it. The run still reported success, because emitted-entry lookup falls back to trailing-stem matching and found the stray file — and the ttsx lane runs on `ubuntu-latest`, where realpath equals resolve, so CI could not see it.

Fixed by resolving both halves through one identity context, and by asking the post-build lookup in the same spelling the build answered in. `test_ttsx_keeps_an_out_of_include_entry_emit_out_of_the_source_tree` reaches the project through a symlink — reproducing on every platform what realpath divergence does on macOS — and asserts the real source tree gains no emitted file.

### Accepted and fixed

**Two tests could not fail if their fix were reverted.** `test_ttsx_runs_an_entry_the_project_include_excludes` created `build/release.ts` but no `src/release.ts`, so deleting the `isWithin` ownership guard in `emittedEntryOf` broke nothing — the very failure the guard exists for was absent from its own fixture. The sibling now exists and prints a distinguishable string. `test_ttsx_classifies_a_node_modules_package_type_over_the_project_module_option` had no miscased twin, so folding case in the segment test broke nothing; `src/Node_Modules` is now the negative, because upstream's test is a case-sensitive substring search and that manifest must lose to the project's own `module`.

**A comment corrected in one place and left standing twelve lines above its own correction.** `moduleFormat`'s inline note still said tsgo reads `packageJsonType` only under `node_modules`, contradicting the function doc directly below it. Also narrowed: the escape is not "a file with no explicit extension" but a file whose extension does not itself state the format.

### Accepted, not fixed — recorded

Three branches added in `f071b5952` remain untested: the array-valued `moduleOptions` marker rejection, `owningModuleOptions` returning `null` for an unreadable owning tsconfig, and `entryModuleFormat`'s `?? {}` fallback. The `nearestPackageType` "no manifest anywhere above" terminal branch lost its only test when the environment-dependent `bare` case was replaced; restoring it needs a filesystem seam rather than a `t.TempDir()` fixture. Recorded here rather than left to be rediscovered.

### Rejected

**"`1e6e9d3d9` rests on the assumption `c9e975142` refuted."** It does not. The outer `.catch` covers the success-then-`finally`-throws path, where the inner catch never ran — genuinely reachable for the two loaders that have a `finally`, both of which still carry it. The banner and strip loaders never had one, and that pasted comment was already corrected downstream.

**Forbidden-section violation in the repeated `commonAncestorDirectory` rewrites.** Three iterations inside one pull request has the shape of whack-a-mole, but each named and addressed a different cause — raw path arithmetic, then the identity predicate, then the returned spelling. The correct classification is an incomplete consequence trace, which the blocker above records and closes.

### Confirmed clean

The package-scope walk matches `ast.GetImpliedNodeFormatForFile` in all four implementations; `effectiveModuleKind` matches `GetEmitModuleKind` including the unset-`target` default of `ScriptTargetLatestStandard`; extension precedence matches for every module kind; all four `node_modules` override combinations agree with upstream; TS1378 really is keyed on the program-wide module kind, so the async-IIFE rewrite is the right response; `amd`/`umd`/`system`/`ES5` really are removed, so the deleted tests covered configurations that cannot reach emit; `bf61471cf` weakens no assertion; and the new Go loader-tsconfig tables fold back the wildcard-`types` assertion the merged test carried, so no coverage was lost.

samchon (commented):
## Individual Self-Review — `9e5a61cd2`

`fix: retire the deadline claims that outlived the deadlines`

A read-only reviewer checked the round-1 correction against the pinned typescript-go and swept the whole repository — not just this pull request's touched files — for surviving claims. Four blockers, all accepted, all fixed in `83fa83ed5`.

### Accepted and fixed

**The corrected claim survived twelve lines above its own correction.** `moduleFormat`'s inline comment still said tsgo reads `packageJsonType` only for paths under `node_modules`, which the same commit had just corrected in the function directly below. A reader hits the false one first, because it sits on the call site.

**`packages/wasm/README.md` still passed `timeoutMs` in its published example.** The option no longer exists on `IBootTtscOptions`, so copying that block yields TS2353 — and this pull request had already deleted the identical line from the website mirror while leaving the README. Its prose restated the removed 60-second default too.

**`protocol.mdx` stated a 30-second LSP verb timeout twice**, once for the spawn-per-verb path and once for the resident path. Both constants were removed by this pull request. The 4 MiB / 1 MiB message limits in the same sentence are real and were kept, now labelled as bounds on the reply rather than on the work.

**`lsp_native_resident.go`'s `call` doc comment still promised a per-request timeout** — the resident twin of the exact defect the commit fixed on the spawn path, and contradicted by the "No deadline" comment in its own body forty lines down. Its vestigial bare block, left behind when the `select` went, is gone too.

**Five more user-facing statements about the removed 30-second pack deadline** in `packages/playground/README.md` (twice), `website/README.md`, and `playground.mdx` (twice). One of them instructed readers to pass `createTypiaSourcePackMount({ url, signal, timeoutMs })`.

**Two wording corrections in the commit's own new text.** `exec.CommandContext` kills its child rather than reaping it, and the early-return path the comment justified does not exist today — the comment now says what the context is actually for. And the upstream escape is a file whose extension does not itself state the format (`.mts`/`.cts`/`.mjs`/`.cjs`), not "a file with no explicit extension".

### Declined, recorded

Four test symbols still read `..._bounds_...` while their doc comments say there is no deadline to exercise. Those names describe what the tests still assert — cancellation and cache-recovery bounds — and renaming exported test symbols across four files is churn this cycle does not need.

### Confirmed clean

The deleted `ctx.Err()` branch was provably unreachable: the context carries no deadline, `cancel` is stored nowhere and passed nowhere, and its only invocation is the `defer` that runs after the check's position. No compile or `lostcancel` regression. Nothing in the repository asserts the deleted message. All eighteen removed identifiers have zero surviving references. The bounds deliberately kept — LSP message limits, `MaxFrameBytes`, the npm installer's tarball and unpacked limits, the plugin-build lock steal budget — are all still accurately described, and were not swept with the rest.

samchon (commented):
## Individual Self-Review — `a5ef56d61`

`fix: give a descriptor that never loaded the same reason as one that threw`

No blockers. The reviewer reconstructed the emitted shim byte for byte, confirmed it parses and type-checks, and traced both new fake-ttsx modes to the exact bytes they assert. Four candidates accepted; all are fixed in `91a5b7d16`.

### Accepted and fixed

**The headline fix shipped with no coverage of itself.** The commit moved `await import(...)` inside the shim's try, then tested only the parent-side reader through a fake ttsx that never runs the shim. Reverting the fix left the whole suite green. The objection I had held — that a real-shim failure test would be first of its kind and unverifiable here — was simply wrong: three tests already drive the real shim through ttsx (`test_plugin_ts_barrel_descriptor_entry_loads_through_ttsx`, `test_plugin_corpus_factory_context_dirname_resolves_source_through_ttsx`, `test_loadprojectplugins_ttsx_descriptor_reads_context_env`), all with descriptors that import successfully. `test_loadprojectplugins_ttsx_descriptor_names_a_failure_from_its_module_body` now uses that same harness with a descriptor that throws while evaluating, and reads the thrown message off **stdout** so the shim's own stack on stderr cannot satisfy the assertion.

**The reason reader's shape coverage lagged its lint twin from this same pull request.** The lint side got all nine result-file shapes at unit cost; the ttsc side had two, and the reader was module-private so nothing else could reach it. `pluginDescriptorFailureReason` now sits beside `pluginDescriptorProcessFailure` — the same move the lint side made one commit earlier — and `test_plugin_descriptor_reason_honours_only_a_well_formed_envelope` pins all nine.

**The import-failure path lost its trailing newline.** Before the commit, ttsx's bootstrap `fail()` ended the stack with `\n`. Afterwards the shim catches first and wrote the stack bare, and since its fd 2 is the parent's fd 2, the last stack frame and `ttsc: plugin descriptor ... failed with exit code 1` arrived glued into one line. Both this loader and the `@ttsc/lint` extractor now terminate the stack.

**The doc comment enumerated six of seven modes.** `stdout-nonzero` was missing — pre-existing, and this commit had the list open in its hands.

### Declined, recorded

The failed-envelope-write branch (`try { writeFileSync(...) } catch {}`) is still unexercised. Making `writeFileSync` throw inside a real shim is not something a fixture can arrange cheaply, and the branch's whole purpose is that `process.exit(1)` runs regardless — which it does unconditionally.

### Confirmed clean

TS1378 is excluded twice over: the shim's own tsconfig hard-pins `module: nodenext` (unlike the banner/strip loaders, whose module option follows the user's package), and moving a top-level `await` into a `try` keeps it in top-level context, so the diagnostic surface is identical before and after. `"absent" in testCase` narrows correctly — `config/tsconfig.json` sets `noUncheckedIndexedAccess` but not `exactOptionalPropertyTypes`, and `absent` is a required property of one `as const` union member rather than an optional property of a shared shape. The `envelope` regex pins trimming on both sides and cannot be satisfied by untrimmed output. `foreign-result` stops inside the `if (processFailure)` block, so neither the "produced no descriptor output" nor the "produced invalid JSON" branch is reachable for it. No CRLF path exists (`.gitattributes` is `text=auto eol=lf`). No state leaks between loop iterations: markers are per-mode and the result file lives in a per-call `mkdtemp` directory removed in the `finally`. And no catch between the loader and the worker wraps the error, so `error.message` is exactly what the parent built.

samchon (commented):
## Overall Self-Review — rounds 3 and 4 (`a5ef56d61` → `91a5b7d16`)

Two fresh complete rounds over the whole base-to-head diff, each also adjudicating the Individual Self-Review results that arrived during it. The per-commit adjudications are recorded in the four Individual reviews above; this is the round-level result.

### Round 3 — one blocker, six stale claims

The blocker is a data-integrity defect that CI structurally could not see: `buildEntryProject` gave tsgo a `rootDir` in the identity-resolved spelling and a `files` entry in the raw one. Upstream compares them lexically and, when the input is not under `rootDir`, builds the output path from the source path instead of `outDir` — so on macOS, on a Windows 8.3 `TEMP`, or in any symlinked checkout, `ttsx clear.ts` wrote `clear.js` and its map **into the user's source tree**, and the run still reported success because emitted-entry lookup fell back to trailing-stem matching and found the stray file. The ttsx lane runs on `ubuntu-latest`, where realpath equals resolve, so no lane could fail. Fixed by resolving both halves through one identity context; guarded by a test that reaches the project through a symlink, which reproduces the divergence on every platform.

Two tests in the cycle could not have failed if their fix were reverted — the out-of-`include` ownership guard had no same-stem sibling to guard against, and the `node_modules` package-type override had no miscased twin. Both now have one.

The stale claims were all outside the diff, which is why round 2 missed them: a published `@ttsc/wasm` README example still passing a removed option, a 30-second LSP timeout stated twice in the protocol guide and once in a Go doc comment, the 30-second pack deadline in two READMEs and the playground guide, and the `node_modules` comment corrected in one place while its twin stood twelve lines above the correction.

### Round 4 — coverage of the previous round's own fix

Round 3's descriptor-shim fix was tested only through a fake ttsx that never runs the shim, so reverting it left the suite green. Three tests already drive the real shim; the fix now has one too. Both reason readers moved beside their status classifiers and each gained a nine-shape proof, and both loaders now terminate their stack line so the parent's message no longer starts mid-line.

### Standing record

Four items are accepted but deliberately not fixed in this cycle, recorded so they are not rediscovered as new: three untested branches from `f071b5952` (array-valued marker rejection, `owningModuleOptions` returning `null` for an unreadable tsconfig, `entryModuleFormat`'s `?? {}` fallback), the `nearestPackageType` terminal branch that lost its only test when an environment-dependent fixture was replaced, the descriptor shim's failed-envelope-write branch, and four test symbols still named `..._bounds_...` whose bodies now say there is no deadline to exercise.

A further fresh round follows over `91a5b7d16`.

samchon (commented):
## Individual Self-Review — `83fa83ed5` … `91a5b7d16`

The most valuable review on this pull request. It found that the round-4 commit **broke every ttsx descriptor load**, and it found it by reconstructing the emitted bytes rather than reading the source.

### Blocker — accepted, fixed in `6e5fb7b95`

**The trailing-newline fix put a raw line terminator inside a JavaScript string literal.** `loadProjectPlugins.ts` assembles the shim from an array of template literals, so the template consumes one level of escapes. Written as `+ "\n"`, the cooked value is a real newline inside `"…"`, and the emitted `load-descriptor.mts` stops parsing. `@ts-nocheck` suppresses semantic errors, not syntax errors, so `runBuild` returns non-zero and **every** descriptor evaluation fails at the build gate — the successful ones as much as the failing ones. Verified by evaluating the exact source line: `HAS RAW LF: true`. The `@ttsc/lint` twin, written `+ "\\n"` in the same commit, was correct.

Fixed, and the emitted shim now verifies clean: 18 lines, quotes paired on every line, no control characters, and it parses as a real ES module.

**The reason it shipped is the finding underneath the finding.** `@ttsc/lint` has carried `test_lint_config_descriptor_extractor_emits_parseable_source` since this exact defect shipped once on the Go side; the descriptor shim had no equivalent, because its source was built inline and could not be inspected. It is now `PLUGIN_DESCRIPTOR_SHIM_SOURCE`, and `test_plugin_descriptor_shim_emits_parseable_source` applies the same three checks to the emitted bytes. Sealing the class, not the witness.

### Accepted and fixed

**Resolving the entry through the identity context resolved the entry *file*, not just its spelling.** A symlinked `clear.ts` was retargeted into whatever tree it pointed at: relative imports then resolved from there, and `rootDir` widened to the ancestor the two trees share — precisely what `buildEntryProject`'s own doc comment forbids, since a wide `rootDir` offers every raw `.ts` on disk to a lookup that falls back to stem matching. Only the directory is resolved now; a link file keeps the name the user ran.

**The fix was incomplete at `emittedEntryOf`.** `isWithin` folds spellings but `resolveEmittedJavaScript`'s exact-mirror lane compares lexically, so a raw entry passed the guard and then silently degraded to `bestStemMatch` — which keeps the first maximum, so `--listEmittedFiles` order decided which file ran. The commit message claimed one identity context resolved both halves; the diff only delivered it for `buildEntryProject`. The spelling is now decided once in `prepareExecution` and used by `createProjectContext`, both `emittedEntryOf` calls, and `buildEntryProject`.

### Accepted, recorded

The trailing-newline treatment covers 2 of 8 sibling loaders; the four Go loaders in `linthost/config.go`, `banner.go`, and `strip/driver/config.go` still write an unterminated stack into a stream their parent then writes to. Same cosmetic glue, not fixed here.

### Rejected

**The symlink test being unable to fail without the fix.** Verified end to end instead: `readProjectConfig` returns a realpath'd root while `runTtsx` does a bare `path.resolve`, TS6059 is defined but never emitted at this pin, and `sourceTreeEmits` walks the physical directory where the stray `clear.js` and its map land. It fails.

### Confirmed clean

The `isWithin` guard is genuinely load-bearing now that `src/release.ts` exists — deleting it makes `bestStemMatch` score the whole-project `release.js` above `index.js`, ttsx prints `compiled-into-lib`, and the assertion fails. The miscased `Node_Modules` negative genuinely bites, traced through `fileloader.go`'s case-sensitive `strings.Contains` and `ModuleResolutionKindBundler = 100 > NodeNext = 99`. The Go envelope key change is consistent across both writers and the reader, no test asserts the old key, and `loaderFailureReason` is only consulted on a failed run. Both nine-shape envelope-reader tests are exhaustive and correct.

samchon (commented):
## Individual Self-Review — `e0420d236` … `a059b72f5`

No blockers. One finding goes to the heart of what the commit claimed.

### Accepted and fixed in `6e5fb7b95`

**The restored test could not detect the break on its second call.** The fake emitted its unknown-id frame with `changed: false`, so under the mutation that defeats both id guards, call 1 happened to fail only because no resident graph existed yet — and call 2, plus all four assertions, passed. Detection rested on an accident of ordering, and the doc comment's stated harm ("resolving on it would hand the caller another request's graph") was never demonstrated, because the frame carried no graph.

The unknown frame now carries one. A client that settled on it answers with a graph the request never asked for, so both calls fail under the mutation. A correct session still drops the frame at the id check before any semantic handling, so nothing else changes.

**The doc comment's third paragraph was edit history.** It referred to "the original case", which no longer exists in the tree, so a future reader could not resolve the reference. Replaced with what the fixture actually does and why.

### Accepted, recorded — not fixed in this cycle

**A second timer-independent invariant is still unpinned.** The same deletion took `this.current = undefined` in `failChild` with it. Deleting that line today leaves the whole graph suite green: every restart test's replacement child answers `initial: true`, so `current` is overwritten before a stale value can be observed. Its sibling on the next line — the shard-store reset — *is* pinned. Pinning `current` needs a fake that can answer unchanged after a restart, which is a new fixture mode rather than an assertion.

**Dead fixture surface.** `respond-then-hang` (and the `stallAfterFirst` branch and request counter it feeds), `delayMs`, and `stderr` all lost their only consumers with the timeout tests. `unknown-then-respond` going dead is what made the id-pairing loss visible at all, so the same signal is still standing on three other knobs.

### Rejected

**"The unknown frame is dropped by a path that would also drop a legitimate mispairing."** No — the drop *is* the invariant path, and every earlier exit in `onLine` calls `failChild`, which rejects the live request, so a frame dropped earlier would make the awaited call throw rather than pass. **"Mutating the id routing does not fail the test."** True only of the naive one-site mutation, which `settlePending` re-verifies; that mutation does not break the invariant either. **"`assert.equal(second, first)` only passes because the graph is empty."** No — `refresh` reassigns `this.current` only under `response.changed`, so identity holds for any payload, and `node:assert/strict` makes `equal` a reference compare.

### Confirmed clean

The fake emits the unknown id on every request, not just the first; ordering is guaranteed by one encoder on one pipe and readline's arrival order; the request is pending before the write, so the precondition is real. `readPids` reads what the fake writes, `pendingCount` reads the real private map, and the `delay` un-export has no remaining importer. `a059b72f5`'s two rewrapped headlines still begin with the identifier, are byte-identical to the originals once whitespace is normalized, and are gofmt-stable — and `format-check.cjs` gates them in the always-selected `typecheck` lane even though those packages do not select the `go` lane.

samchon (commented):
## Individual Self-Review — `6e5fb7b95`

The reviewer confirmed the raw-newline repair by evaluating the emitted array and parsing it (`PARSE OK`, and `Invalid or unexpected token` for the reverted variant), then found that the round-6 spelling fix had traded one failure for another.

### Blocker — accepted, fixed in `7a9eb0916`

**A symlinked entry *file* went from a sloppy success to a hard failure.** Round 6 resolved the entry's directory but deliberately not the entry file, so a link keeps the name the user ran. The ownership guard did not agree: `isWithin` delegates to the filesystem-identity context, whose `resolve` calls `fs.realpathSync.native` on the **whole** path, final component included. So the guard folded the link, placed the entry at its target, found it outside a `rootDir` that was never widened to reach there, and threw `ttsx: emitted entry not found`.

Before round 6 the second lookup received `buildEntryProject`'s fully-resolved return value against a widened root and passed — with the over-wide `rootDir` round 6 rightly removed. And no test ran a symlinked entry *file*: the round-6 test symlinks the **directory**, which `resolveEntrySpelling` resolves, so it stays green. CI could not have caught this.

The guard is now the same lexical containment `resolveEmittedJavaScript` applies. Both sides already carry one spelling, so there was nothing left to fold — the folding *was* the defect. `test_ttsx_runs_an_entry_that_is_itself_a_symlink` links to a script outside the project, which is the only shape where the two tests disagree.

### Accepted and fixed

**Both descriptor-source guards skipped `\r`.** The class covers `-`, ``, ``, `-` — deliberately not `\n`, because the line split plus the quote scanner catch a stray newline. A carriage return is caught by neither: it does not split the line and quotes still pair around it, yet `"<CR>"` is a syntax error. The `@ttsc/lint` twin that has guarded this class since it first shipped had the same hole; both now carry `\r`.

**`resolveEntrySpelling` resolved a relative entry against `process.cwd()`** rather than the caller's `--cwd`, while everything else in the function used the latter. Unreachable through the CLI, which always passes an absolute path; one line either way.

**`commonAncestorDirectory`'s doc premise was stale** — it explained that the entry does *not* arrive resolved, which stopped being true when its only caller started passing a resolved directory.

### Accepted, recorded

The two error messages in the same function now name different spellings (`entryFile` as the caller passed it, `entry` as resolved). Naming the caller's spelling to the user is the better choice; the divergence is deliberate and now stated rather than accidental.

### Rejected

Two commit-message criticisms are fair and are corrected here rather than by rewriting a pushed message. The phrase "was written `"\n"`" was used for both the broken ttsc side and the correct lint side, which reads as self-contradictory — the lint side was written `"\\n"` and was right from the start. And "only one of the two loaders had a guard" is incomplete: `91a5b7d16` also added an end-to-end test that would have failed, so the sharper lesson is that round 4 was pushed without waiting for CI on it.

### Confirmed clean

The emitted shim parses as an ES module with top-level `await`; all 18 lines pass both checks; the guard genuinely catches the reverted escape (`quotesPair` returns false on the split line). The test file holds zero bytes below `0x20` other than LF. The import path is proven by an existing test that `require()`s the same built file, and `tsc` emits `lib/**` per file — there is no bundler for `packages/ttsc`. Exporting the constant widens nothing: `./lib/plugin/internal/*` is not a subpath export, `sideEffects: false` stays honest, and the name appears nowhere else. The graph fake's `changed: true` frame passes every earlier gate and is dropped at the id lookup for the right reason, and `unknown-then-respond` has exactly two references in the tree.

samchon (commented):
## Individual Self-Review — `7a9eb0916`

This review found that the fix and the fix before it were both aimed at the wrong target, and it is why rounds 6 through 8 exist at all.

### Blocker — accepted, fixed in `1b0d2b5b2`

**The gate owned an emit the runtime refuses to serve, and the entry ran through the orphan lane.** Round 7 made `emittedEntryOf` lexical so it would stop rejecting a symlinked entry file. Nothing in `runtimeHooks` changed: `entryEmitPath` still tests `isWithin(real, m.rootDir)` with `real = fs.realpathSync(filename)`, and `runTtsx` spawns the child with the **source** entry, so the hooks decide what gets served. For a symlinked entry the gate said "owned", the runtime said "not owned", `serveDependencyEmit` found no owning tsconfig above the target, and the file fell to `transformOrphanSource` → tsgo with `--ignoreConfig --noCheck`.

That silently drops the project's transform plugins, `target`, `paths`, and inlined source map. A project on `typia` gets `typia.random` with no transform applied. A build that looks like it worked and did something else is worse than the rejection round 7 was fixing.

**And the round-7 test could not see it** — its script was `console.log(...)`, the one program shape where both lanes are indistinguishable.

### The real correction

Round 6 decided a symlinked entry *file* should keep the name the user ran, and I accepted that framing without checking what the rest of the system does with a path. It disagrees with all of it: the project root arrives physically resolved, tsgo emits from the path it opens, and the runtime hooks key a served file by `fs.realpathSync` because that is how Node identifies a module without `--preserve-symlinks`. The preserved basename was a fourth answer nobody else gives, and rounds 6 and 7 were both patches around it.

The entry is resolved fully now. `rootDir` widening to the shared ancestor is the requirement, not the cost: the file genuinely lives outside the project, and no root that excludes it can compile it. Gate, emit, manifest, and runtime read the same path again.

### Accepted and fixed

**The regression leaked a temp directory on the early-return path.** `outside` was created before the `symlinkSync` attempt and removed only in the `finally` that the early return skips, and it was not harness-tracked — the accumulation `TestProject` documents as a cause of tmpfs exhaustion. It now uses `TestProject.tmpdir`.

**The guard was "nearly" equivalent to `resolveEmittedJavaScript`, not equivalent.** It omitted the `relative === ""` arm. Unreachable in practice, but the commit's whole thesis was exact equivalence, so it now carries that arm too.

**A stale comment.** `prepareExecution` still explained that only the directory is resolved, twelve lines above the function that resolves the whole path.

### Accepted, recorded

Two shapes the commit message wrongly called "the only" disagreement are genuinely repaired and remain untested: a symlinked source file *inside* `rootDir` and inside the project's file set (which round 7 fixed by accident), and a symlinked `rootDir` *directory*, where the old identity guard admitted an entry the lexical mirror then could not mirror — routing straight to the stem matcher, the wrong-file hazard the guard's own doc warns about.

### Rejected

**A case-sensitivity or separator divergence between the guard and the mirror.** Both call `path.relative` in the same argument order and feed the same `isOutsideRelativePath`; `path.relative` resolves both operands itself, and Windows case folding happens identically on both sides — convergence, not divergence.

**`\r` failing on a legitimate emitted line via a CRLF checkout.** `.gitattributes` is `text=auto eol=lf`, a dedicated test asserts a fresh checkout carries no `\r\n` under either `core.autocrlf`, and a byte scan of both loader files finds no `0x0D`. The only escapes in either literal are `﻿`, `` \` ``, `\$`, and `\\`.

**A missing TAB in the class.** Correct as written: `U+0009` is legal inside a string literal; `\r` is not.

### Confirmed clean

`runtimeHooks.isWithin` keeps its consumers, so its "exported for direct exercise by the ttsx e2e suite" note stays true. `resolveEntrySpelling`'s cwd argument matches how `readProjectConfig` resolves a relative `file`, and changes nothing for the absolute path the CLI always passes. The two guard-test twins remain identical in their checks, and the `\r` gap was real and pre-dated both.

samchon (commented):
## Individual Self-Review — `1b0d2b5b2` … `bb594fcd5`

Two reviews cover this range. Together they ended a four-attempt loop, and the second one delivered the verdict that mattered: *not a fifth patch — the diagnosis is right and the split is the design — but implemented one layer too shallow.*

### Blocker — accepted, fixed in `bb594fcd5`

**Resolving the entry before project discovery retargeted the tsconfig walk.** `1b0d2b5b2` handed the resolved path to `createProjectContext`, which is what *finds the project*. A symlinked entry therefore started `findUp` in the target's tree: with no tsconfig above it the run died with `could not find tsconfig.json`, and in a monorepo a link from `packages/app/tools/gen.ts` to `packages/shared/tools/gen.ts` silently compiled under **shared's** options and plugins, wrote the synthesized tsconfig into `shared/`, and cached there. Nothing reported the substitution.

There were never one entry path and one rule. There are two questions: *which project compiles this* (the path the user named, because discovery walks up from it) and *where does the output go and what does the runtime load* (the physical path, because Node keys a module by its real path). `createProjectContext` takes both.

### Blocker — accepted, fixed in `78a038981`

**The split left `runtimeRootDir` outside its own rule.** `project.root` arrives through plain `fs.realpathSync`, which leaves a Windows 8.3 component alone; the entry arrives through `fs.realpathSync.native`, which expands it; `path.relative` folds case but not 8.3. A declared `rootDir` was worse — joined verbatim, so a `rootDir` that is itself a symlinked directory never resolved at all. The gate then read an ordinary in-project entry as outside its own root, ran the entry-only build for it, and published a root wider than the project has. With an ambient `.d.ts` in the project that second build fails outright, because the synthesized config declares only the entry.

The invariant is therefore not "the entry has one spelling" but **every path the entry is compared against must be produced the same way the entry is**. `resolveRuntimeSourceRoot` now produces it that way, and `emittedEntryOf`'s comment says so — it previously asserted the pair was already honest, which is the assumption that hid this.

### Accepted and fixed

**The fifth spelling, confirmed in scope.** The bootstrap asked `entryModuleFormat` from the unresolved path, so `moduleFormat` walked from the *link's* directory to find the deciding `package.json`. A `nodenext` project whose link target sits in a `"type": "module"` package got its ESM emit handed to `require` and died on `Unexpected token 'export'`. It asks from the physical path now; `process.argv[1]` deliberately keeps the name the user typed, and Node's own realpathing already made the `require`/`import` calls correct.

**Eight tests had drifted from the product.** The source-map and stack-trace cases build expectations with plain `fs.realpathSync` while the product now answers with the native form. They share one `physicalRealpath` helper.

**`realPath` in `runtimeHooks` used the plain form** while the launcher used `.native`, so the two disagreed on exactly the 8.3 hosts.

### Corrections to my own claims

Two, recorded rather than quietly dropped. **tsgo does not "emit from the file it opens"** — it takes `files` verbatim and never resolves them, which is precisely why it must be handed the spelling Node will use; making it a second independent witness overstated the case and is part of why `runtimeRootDir` looked already-settled. And the 8.3 divergence is **a real-user fix, not a CI fix**: no Windows lane runs ttsx at all, so the commit message borrowed authority from a runner that never executes this code.

### Accepted, recorded

`prepareExecution`'s returned `entryFile` is dead — `runPreparedEntry` never reads it, so `emittedEntryOf` is purely an ownership gate and the path it returns is discarded. And `scriptEndingWith("clear.ts")` in the symlink regression matches only because link and target share a basename; the hooks record the target's URL.

### Rejected

**That the wide-`rootDir` hazard returned.** On the default CLI path it cannot widen at all — discovery walks up from the entry's own directory, so `context.root` is already an ancestor and `commonAncestorDirectory` returns it exactly. Widening needs an explicit `--tsconfig` pointing at an unrelated tree. And the residue is bounded: `resolveEmittedJavaScript` only ever answers from `emittedFiles` or a scan of `outDir`, never the source tree.

**That the new regression proves nothing.** It fails at `1b0d2b5b2` on discovery and at `7a9eb0916` on the source-map assertion — the orphan lane emits with `--ignoreConfig` and no `--sourceMap`, while the entry lane inlines one because `forceRuntimeSourceMap` is on for a project that configures neither.

**That routing discovery through the unresolved path breaks a symlinked project directory.** `findUp` uses `existsSync`, which follows links, and `resolveProjectConfig` realpaths what it finds. Only a symlinked `rootDir` *target* was affected, which the `78a038981` fix closes.

samchon (commented):
## Overall Self-Review — final round (`e73cf0d03`), clean

Nine complete rounds and thirteen Individual Self-Reviews. This round produced nothing that survived verification, and the head has 27/27 green checks. Recording the merge gate.

### What the review rounds found that CI structurally could not

Every one of these was invisible to the pipeline, and each is now pinned by a test that fails if its fix is reverted:

| Defect | Why CI could not see it |
| --- | --- |
| The entry project wrote `clear.js` and its map **into the user's source tree** | The ttsx lane runs on `ubuntu-latest`, where realpath ≡ resolve. The divergence needs macOS, a Windows 8.3 `TEMP`, or a symlinked checkout. |
| The descriptor shim emitted a **raw newline inside a string literal**, breaking every ttsx descriptor load — the working ones too | `@ts-nocheck` suppresses semantic errors, not syntax errors, so it failed at the build gate. The `@ttsc/lint` twin had carried a guard for this class since it first shipped; this shim had none, because its source was built inline. |
| A symlinked entry fell to the **orphan type-strip lane**, dropping the project's transform plugins, `target`, `paths`, and source map | Both lanes print and both exit zero. |
| A symlinked or 8.3-spelled `rootDir` sent an ordinary in-project entry through the entry-only fallback | No Windows lane runs ttsx at all. |
| The response-id pairing invariant was deleted along with the deadline it happened to be tested beside | Nothing else asserted it. |

### The design failure worth naming

The entry-path decision was rewritten four times inside this pull request. That is exactly what the development skill's *"No forcing a broken design"* rule describes, and I spent three of those attempts on symptoms because I accepted a reviewer's framing — "a symlinked entry file should keep the name the user ran" — without checking what the rest of the system does with a path.

The root cause was that there is no single answer. **Two questions are being asked**, and they have two answers:

- *Which project compiles this?* — the path the user named, because discovery walks up from it.
- *Where does the output go, and what does the runtime load?* — the physical path, because Node keys a module by its real path and tsgo takes `files` verbatim.

And the invariant that follows: **every path the entry is compared against must be produced the same way the entry is.**

### Corrections to my own earlier claims, on the record

- tsgo does not "emit from the file it opens" — it never resolves `files`, which is precisely why it must be handed Node's spelling.
- The 8.3 fix is a real-user fix, not a CI fix; no Windows lane executes this code.
- A wide `rootDir` does not "offer every raw `.ts` on disk" to the lookup — `resolveEmittedJavaScript` only ever answers from `emittedFiles` or a scan of `outDir`. That sentence was a safety claim that was really a precision claim, and it was the stated reason for a decision that took three attempts to undo.

### Deferred, recorded rather than dropped

- `buildDependency` holds the same mixed pair for the dependency lane. No wrong-file outcome was constructible — two files in one emit tree cannot share a trailing stem chain — so the cost is the exact-mirror lane being dropped and a recursive scan per served file. Own issue.
- Three untested branches from `f071b5952`, `nearestPackageType`'s terminal branch, the shim's failed-envelope-write branch, and four test symbols still named `..._bounds_...`.

### Merge gate

Required checks green on `e73cf0d03`; no commit in `origin/master..HEAD` is a revert, so the squash closes exactly #1068, #1069, #1070, and #1071, each with a surviving fix.

Issue #1077, opened 2026-08-05
Document @ttsc/evidence in the README, the website, and the skill index


samchon:
## Problem

Once `@ttsc/evidence` ships from this repository, nothing here tells anyone it
exists. The root README lists the packages ttsc ships; `website/src/content/docs`
is the canonical home for guides and is what https://ttsc.dev serves; and
`AGENTS.md` indexes the skills an agent needs before touching a package.

The evidence graph is also the one package in this repository whose *domain
model* an agent cannot infer from the code: nodes and edges, the
coverage-versus-integrity split, activation gates, and the `@evidence` tag
grammar. The source repository keeps that in a dedicated skill, and losing it in
the move would leave every future change to the rules guessing.

## Evidence

The source repository's `README.md` opens with the product claim this package
exists to deliver:

> When Claude Code or Codex works unattended, it can skip a requirement and still
> report "done." Evidence Graph makes every configured requirement demand an
> explicit acknowledgement from the code, test, or document that claims to
> satisfy it. … **An agent can still lie. It cannot lie by omission.**

and demonstrates it with a diagnostic that is a real compile error:

```
error TS16411: [evidence/graph] Missing acknowledgement for
  'docs/discount.md#coupon-stacking' (Markdown H2 'Coupon Stacking' at docs/discount.md:3)
```

That framing is the reason the package is worth vendoring, so it should survive
the move rather than being reduced to a package listing.

The source repository also carries `.agents/skills/project/evidence/SKILL.md`,
described there as "the domain model this product exists to enforce … Read before
changing rule semantics, the tag grammar, the configuration surface, or any
diagnostic message."

## Consequence surface

- `AGENTS.md` is the shared entry point for both Claude Code and Codex CLI, and
  its maintenance rules say to update it only for repository-contract changes —
  a new skill area is explicitly one of them.
- The documentation skill owns README and website writing rules, including its
  prose-line conventions; the vendored text was written under a *different*
  repository's documentation skill and must be re-read against this one rather
  than pasted.
- The website is a Nextra site with its own navigation; a new page has to be
  placed in it, not just added as a file.
- `og.jpg` and the badge row in the upstream README point at
  `samchon/lint-plugin-evidence`. Every one of those URLs is wrong once the code
  lives here.

## Approach

Place a guide under `website/src/content/docs/` beside the other plugin guides,
carry the product framing and the worked `@evidence` example into it, and give
the root README the same short entry the other shipped packages get.

[...]

Issue #1076, opened 2026-08-05
Relocate the evidence benchmark into benchmarks/evidence and its suite into tests/


samchon:
> **Destination changed during implementation.** This issue was written when the
> benchmark would nest inside the existing `experimental/benchmark` package. It
> ships instead as its own package at `benchmarks/evidence`, alongside `graph`
> and `performance`, because the three differ in kind and one package could not
> describe that. Paths below are updated to what shipped.

## Problem

The evidence repository carries a benchmark that measures what the product
claims: whether an agent working unattended actually satisfies a requirement, and
what it costs when the evidence graph is enforcing versus when it is not. It is
13 runner sources, four requirement corpora (`erp`, `reddit`, `shopping`,
`todo`), two instruction sets (`evidence`, `plain`), three workspace templates,
and a 6-case suite proving the runner itself.

Vendoring the package without it leaves the claim in the README unmeasurable
here. But it cannot simply be copied into place: this repository already has a
benchmark home with its own contract.

## Evidence

`benchmarks` is a private workspace package that already holds
"short, export-free performance and graph executable bootstraps, reusable
`TtscBenchmark*` / `ITtscBenchmark*` implementations, and graph prompt assets".
The evidence benchmark's runners are named `EvidenceBenchmark*` and follow the
same shape, so they belong under that roof rather than at the repository root
where the source repository keeps them.

The 6-case suite (`tests/test-benchmark` upstream) is an ordinary feature suite
and belongs under `tests/`, but its upstream name is generic enough to be
misleading here, where a benchmark already exists that it does not test.

`benchmarks` is already a `pnpm-workspace.yaml` member, so no
workspace wiring is needed for the runners themselves.

## Consequence surface

- Naming: `tests/test-benchmark` would read as "the ttsc benchmark's tests". The
  suite tests the *evidence* benchmark and should say so.
- Asset size: the requirement corpora and workspace templates are checked-in
  fixtures, not generated. They enlarge the repository permanently.
- The graph benchmark contract in `.agents/skills/project/graph.md` governs the
  existing benchmark assets; the evidence benchmark must not be conflated with
  it or with the graph benchmark website.
- Aggregated results (`benchmark/aggregate/**`: `summary.json`, `time.svg`,
  `tokens.svg`) are measurements. Carrying them over means carrying numbers
  produced on another machine, under another ttsc version.

## Approach

Relocate the runners under `benchmarks`, in their own subtree so they
[...]

Issue #1075, opened 2026-08-05
Bring the evidence contributor's e2e and Go suites in, with a CI lane that runs them


samchon:
## Problem

The evidence contributor's behaviour is proved by two suites that do not exist in
this repository: 66 end-to-end feature cases in `tests/test-evidence` and 436 Go
rule tests in `packages/evidence/native`. Vendoring the package without them
would import the code and leave its proof behind — every rule would be
unverified here, and the first `@ttsc/lint` contributor-API change would break it
silently.

Bring both suites in and give them a CI lane.

## Evidence

The e2e suite drives the real `ttsc` binary through `ttsx --no-plugins
src/index.ts`, the same shape `tests/test-lint` uses. Its cases cover the graph
rules (claim activation, reference policy, hierarchy cascade, exclusion carriers,
overlapping claims, re-exported populations), the `documented` rule, the
`singular` rule, the `todo` rule, and the Markdown, Prisma, Swagger, and
TypeScript materializers.

The Go suite sits beside the sources it tests, which is already this
repository's convention for `packages/*/test`. It is denser than ttsc's own Go
suites: 436 `Test*` functions across 65 files, where `packages/*/test` here
keeps one per file.

This repository runs Go suites through per-package runners orchestrated by
`scripts/test-go.cjs`, and that orchestrator's own comment records why the list
matters:

> Derived from the ownership map rather than listed here. A literal array is the
> finite list `scripts/ci/test-owners.cjs` exists to replace: a sixth entry
> someone forgot to add would have run nowhere and reported nothing, which is the
> failure #622 produced and #806 was filed to end.

So a new suite that is not claimed in `test-owners.cjs` is a suite that does not
run — and nothing says so.

## Consequence surface

- `scripts/test-go.cjs`: a new `test-go-evidence.cjs` runner joins `runners`.
- `scripts/ci/test-owners.cjs`: both suites must be claimed, which is the single
  act that both runs them and proves they run.
- `scripts/ci/validation-plan.cjs`: the lane and its path selectors decide which
  pull requests pay for the suite.
- The e2e suite needs `@ttsc/evidence` resolvable from a fixture's
  `node_modules`, which is what `scripts/build-current.cjs` already arranges for
  `tests/lint-contributor-demo`.
- Test runtime: 66 cases each spawning a real compile is a real cost. The lane
  choice has to be deliberate rather than appended to an existing one by default.

## Approach

Copy both suites verbatim, renaming only the symbols the package issue renames
[...]

Issue #1074, opened 2026-08-05
Vendor @samchon/lint-plugin-evidence into packages/evidence as @ttsc/evidence


samchon:
## Problem

`@samchon/lint-plugin-evidence` is an evidence-graph lint contributor for
`@ttsc/lint`, developed in its own repository
(`https://github.com/samchon/lint-plugin-evidence`, v0.5.0). It makes provenance
declarable with a JSDoc `@evidence` tag, resolvable against Markdown sections,
Swagger/OpenAPI operations, Prisma schema members, and TypeScript symbols, and
enforceable as a compile error under rules the consumer configures in
`lint.config.ts`.

It belongs in this repository as a first-party package. Today it is a downstream
consumer: it pins `ttsc` and `@ttsc/lint` through a pnpm catalog and resolves the
Go shim modules through its own `node_modules`, so every `@ttsc/lint` contributor
API change reaches it only after a release, and its 436 Go rule tests never run
against an unreleased ttsc.

Vendor the package as `packages/evidence`, published as `@ttsc/evidence`.

## Evidence

Source inventory (`D:/github/samchon/evidence`, the local checkout of that
repository):

| Path | Contents |
| --- | --- |
| `packages/evidence/src/**` | 24 TypeScript files: the plugin descriptor (`index.ts`), three `internal/` loaders (Prisma models, Swagger operations, Swagger normalization), 12 `structures/`, 3 `typings/` symbol unions, and the two `@ttsc/lint` module-augmentation types |
| `packages/evidence/native/**` | 23 Go rule/engine sources and 65 `*_test.go` files carrying 436 `Test*` functions |
| `packages/evidence/build/prepack.js` | publish-time manifest rewrite |
| `packages/evidence/{go.mod,go.sum}` | local Go tooling only — see below |

The one non-obvious integration constraint is recorded in the source repository's
own `go.mod` header, and it is load-bearing:

> ttsc does NOT use it: at build time ttsc copies the ./native package into
> @ttsc/lint's own Go module as a sub-package (contrib/evidence) and supplies
> every dependency from a generated go.work. A contributor therefore ships Go
> SOURCE, never a module — ttsc rejects a go.mod that sits inside the `source`
> directory itself, which is why this file lives one level above ./native.

Its `replace` directives currently resolve through `./node_modules`. In-workspace
they must point at the sibling packages instead, the way `packages/lint/go.mod`
already does for the shim modules.

## Consequence surface

[...]

PR #1078, opened 2026-08-05
feat(evidence): vendor the evidence-graph lint contributor as @ttsc/evidence

samchon:
Vendors `@samchon/lint-plugin-evidence` into this repository as `@ttsc/evidence`, and gives every benchmark its own package.

## Cycle scope

| Issue | Subject |
| --- | --- |
| #1074 | `packages/evidence` as `@ttsc/evidence`: source, Go rules, manifest, Go module wiring, type renames |
| #1075 | the 66-case e2e suite and 436 Go rule tests, with a CI lane that runs them |
| #1076 | the benchmark and its suite |
| #1077 | README, website page, the two skills, and the AGENTS.md index |

## What this is

`@samchon/lint-plugin-evidence` makes provenance declarable with a JSDoc `@evidence` tag, resolvable against Markdown sections, Swagger/OpenAPI operations, Prisma schema members, and TypeScript symbols, and enforceable as a compile error under rules the consumer configures in `lint.config.ts`.

Today it is a downstream consumer: it pins `ttsc` and `@ttsc/lint` through a pnpm catalog and resolves the Go shim modules through its own `node_modules`, so every contributor-API change reaches it only after a release and its Go rule tests never run against an unreleased ttsc.

## One package per benchmark

`experimental/benchmark` held two unrelated benchmark systems, and the vendored evidence benchmark arrived nested inside it as a third. The three differ in kind: `graph` and `performance` measure ttsc against competing tools, while `evidence` measures a coding agent against itself. One package could not describe that.

```
benchmarks/graph         @ttsc/benchmark-graph
benchmarks/performance   @ttsc/benchmark-performance
benchmarks/evidence      @ttsc/benchmark-evidence
```

No shared runtime module spans them; each carries what it uses and resolves its own work root. The source-contract checker is the one thing neither ttsc harness owns, so it moved out to `scripts/ci/benchmark-source-contract.mts` as a repository-level static check that deliberately skips the vendored package.

## Parity is measured, not asserted

The vendored trees were copied once and then hand-corrected, so nothing could answer whether a difference from upstream was an adaptation this workspace needs or a change the copy missed.

`experimental/evidence-vendor/parity.cjs` answers it. It applies exactly the rewrites `readapt.cjs` declares to upstream's bytes, normalizes both sides through this repository's Prettier so import order and line wrapping cannot masquerade as content, and reports every residual. It resolves upstream's open campaign branch on each run rather than pinning a commit, because a stale pin compares clean against bytes upstream has already replaced.

```
upstream master plus origin/campaign-luna-0.6.0-cont at 4d66b9e75
compared 438 text files and 12 binary files against upstream
declared adaptations: 18 files
parity: clean
```

Building it found a defect the eye had missed for the life of the branch: the delivered benchmark template was never in the adaptation scope, so the Evidence arm's `lint.config.ts` still imported `@samchon/lint-plugin-evidence` while `injectEvidence` installed the packed `@ttsc/evidence`. The arm that exists to exercise the plugin could not resolve it.

## The one non-obvious constraint

A lint contributor ships Go **source**, not a module — `ttsc` copies `native/` into `@ttsc/lint`'s own module as a sub-package and rejects a `go.mod` inside the `source` directory. That is why the upstream `go.mod` sits one level above `native/` and exists only for local tooling. In-workspace its `replace` directives move from `./node_modules` to the sibling packages.

## Deliberately left out

Upstream carries two skills this branch does not bring across: `wiki`, which is about maintaining upstream's own wiki, and `lint-rule-authoring`.

The second is not a clean call. It is not a restatement of `packages/lint/README.md`: it carries Go-authoring guidance about which failures are silent — a duplicate rule name dropped with a stderr warning, overlapping `TextEdit`s discarded without a diagnostic, a `ProjectInputRule` whose signature drift is skipped by type assertion. But its own header names `website/src/content/docs/development/walkthroughs/lint.mdx` — a document in this repository — as its authoritative source, so vendoring it would import a derived copy back into the repository that owns the original. Whether ttsc wants a Go rule-authoring skill, and whether the walkthrough should absorb the silent-failure list instead, is a call for its owner rather than something to fold into a vendoring pull request.

Upstream's aggregate benchmark results are also left out. They are another repository's measurements, and this one runs its own campaign.

## Verification

This pull request's ordinary CI, plus the branch-local gates run before each push: `parity.cjs` clean, `audit.cjs` with `xref`, `link`, and `bin` empty, `readapt.cjs` fully idempotent, the source contract over both ttsc harnesses, `tsc --noEmit` over all five affected packages, the validation-plan tests, and a prompt-manifest regeneration that reproduces the committed file.


samchon (commented):
## Individual Self-Review — the vendoring commits

A read-only reviewer diffed the four vendored trees against the upstream checkout, resolved every relative specifier against disk, and called `planForPaths` to check the lane wiring. Six blockers, all mechanical, all self-inflicted by the move rather than by the code being moved. All are fixed in `e7a3ec6eb` and `3eadeb301`.

### Blockers, accepted and fixed

**The rename edited text but not filenames.** `structures/index.ts` exported `./ITtscEvidenceDocumentedConfig` against a file still called `IEvidenceDocumentedConfig.ts`. 109 specifiers did not resolve, across `packages/evidence/src`, both benchmark trees, and two test internals. Forty files moved with `git mv`. Verified after: 375 relative specifiers resolved, 0 unresolved — the 26 that still look broken are fixture strings the tests generate, matching the reviewer's own count exactly.

**`packages/evidence/tsconfig.json` was the single upstream file the copy missed.** Not cosmetic: `build` is `rimraf lib && tsc`, so `lib/` was deleted and never rebuilt, and the feature suite refuses to start without it, two Go tests `t.Fatalf` on it, and the Prisma bridge resolves through it at run time.

**Three paths were never re-based for their new locations.** The benchmark suite imported the benchmark from the repository root; the benchmark's tsconfig extended `../config` from two levels deeper than upstream; and the template asked the repository catalog for seven names. That last one is the interesting one: `ttsc`, `@ttsc/lint`, and `@ttsc/unplugin` were catalog entries upstream because they were *external* there, and a workspace never lists itself in a catalog. The resolver now reads workspace manifests rather than asking this repository to duplicate its own version numbers; nestia is a genuine external and joins the catalog like any other.

**The `go` lanes ran the evidence Go tests without building the package those tests stat**, and the local `full` scope had the same gap.

### Also accepted and fixed

`tests/test-evidence-benchmark/**` and `experimental/benchmark/evidence/**` fell through to the full 13-lane matrix instead of selecting the lane that proves them — the benchmark suite imports that source directly, so editing it has to run the suite. And `build/prepack.js` copied the repository README over the package's own, which was right upstream (they were the same document) and wrong here; the package now tracks its README like every other ttsc package, and the prepack step is gone.

### What CI then found, which the review could not

Three more, each the same shape — an assumption upstream could make and this repository cannot:

- `@ttsc/evidence` augments `@ttsc/lint`'s module and type-checks against its emitted declarations, so lint has to build first.
- `scripts/build-platforms.cjs` skipped a publishable package that has a build script. `package defenses` refused it, correctly.
- Both suites invoked `ttsx` as a command. Upstream `ttsc` was an installed dependency with a linked bin; here it is a workspace package whose bin exists only after a build. They now start through Node with the workspace loader, the way every other suite here does.

### Rejected

**Go tabs converted to two spaces looked like corruption.** It is this repository's convention: all 3510 tracked `.go` files are the same, and `format:go` sweeps every tracked `.go` through `gofmt-2spaces.sh`, which `format-check.cjs` gates.

**`go.sum` looked stale.** It is byte-identical to upstream and correctly so: the upstream replace target's `go.mod` is identical to in-tree `packages/lint/go.mod`, and every shim pins the same `typescript-go` the require block names.

**The `.gitignore` exception looked like it could not work**, given git's rule that a file cannot be re-included under an excluded parent. `native/` excludes the directory and `!packages/evidence/native/` re-includes that exact directory, so git descends. Empirically all 88 Go files are tracked, and `packages/evidence/lib/` correctly stays ignored.

**47 unresolvable catalog refs under `template/base/**`.** False positives: the template carries its own `pnpm-workspace.yaml`, and it is not a workspace member.

### Confirmed clean

Rule ids (`evidence/graph|singular|documented|todo`), the `evidence` namespace, and every diagnostic string are untouched — a renamed rule id would have silently broken every consumer config. `ITtscLintContributorRules` and `ITtscLintRuleOptionsMap` correctly kept their names. 86 of 88 Go files are byte-identical after detab; the two that differ do so only in the `node_modules/@ttsc/evidence` probe path. `experimental/benchmark/evidence` matches upstream's `benchmark` membership exactly, and the deliberate omissions (`aggregate/`, `output/`, `.cache/`) held.

samchon (commented):
## Overall Self-Review — round 1 (`920415387` → `3eadeb301`)

Whole base-to-head diff, 454 files. This round is mostly the adjudication of the individual review plus three CI failures, all recorded above and in the commits.

### Change Integrity, called out explicitly

`scripts/ci/validation-plan.test.cjs` pinned twelve lanes and seven scoped native builds. Those numbers are a contract, not an incidental, and this pull request changes it deliberately by adding a lane. The development skill requires that be stated rather than quietly edited, so: **the lane count moves from 12 to 13 and the scoped-build count from 7 to 8, because `@ttsc/evidence` ships with 79 e2e cases and 64 Go tests that need somewhere to run.**

Four selections are now pinned explicitly rather than left to inference, because three of the four directory names are not the lane id:

| Path | Lanes |
| --- | --- |
| `packages/lint/**` | …, `lint-1`, `lint-2`, **`evidence`** |
| `packages/evidence/**` | `go`, `typecheck`, `evidence` |
| `tests/test-evidence/**` | `typecheck`, `evidence` |
| `tests/test-evidence-benchmark/**` | `typecheck`, `evidence` |
| `experimental/benchmark/evidence/**` | `typecheck`, `evidence` |

The first row is the one that matters. The evidence rules link into the lint engine, so a contributor-API change that breaks them has to fail in this repository rather than after a release. Without that selector the whole justification for vendoring would have been a claim rather than a fact.

`scripts/build-platforms.cjs` also gained `@ttsc/evidence`. That is a release-path change, and it was `package defenses` that caught its absence — a publishable package with a build script that the release path skipped.

### The pattern in every CI failure

Three rounds of red, and each failure was the same shape: an assumption the upstream repository could make and this one cannot.

- `@nestia/e2e` came from a `samchon` catalog there; here it lives in `utils`. A catalog reference is a name that must exist in the workspace it lands in, not a version.
- `ttsc` was an installed dependency there, so `ttsx` was on the path. Here it is a workspace package whose bin exists only after a build.
- `ttsc`, `@ttsc/lint`, and `@ttsc/unplugin` were catalog entries there because they were external. A workspace does not list itself in a catalog, so the benchmark's version resolver reads workspace manifests instead.
- `@ttsc/lint` was a published dependency there, so its declarations always existed. Here it has to be built before the package that augments its module.

None of these is a defect in the vendored code. All four are the cost of the move, and each is now encoded somewhere a future reader will find it.

### Discretion exercised, and why

- **`native/` keeps its name.** This repository ignores that name as a build output; evidence ships Go *source* under it. Renaming would have meant diverging from upstream and rewriting the descriptor's `source` path for a cosmetic reason, so the ignore rule stays and this one package opts back in, with the reason beside it.
- **No measurements came across.** The `aggregate/` results were produced on another machine under another ttsc version. The benchmark skill's measurement-integrity rule makes carrying them and presenting them as this repository's numbers the wrong default.
- **Two upstream skills did not come across.** `lint-rule-authoring` and `wiki` are already covered by this repository's development, project, and issue-campaign skills, and AGENTS.md's merge rule forbids a second skill saying the same thing. `evidence-graph` came because the domain model is the one thing an agent cannot infer from the code.
- **`build/prepack.js` is gone.** It copied the repository README into the package, which was right upstream where the two were one document.

### Open

CI is still settling on `3eadeb301`. Further rounds follow until a complete one produces nothing.

This pull request is **not for merge**, at the repository owner's instruction.

samchon (commented):
## Overall Self-Review — round 2 (`3eadeb301` → `59cc1c618`)

This round I verified the migration's central claim myself rather than taking the individual reviewer's word for it, because the honest read of my own reviewing this session is that it has been too dependent on the subagent finding my mistakes.

### Fidelity, measured

A script normalizes both sides by undoing the intended renames (and the repository's Go detab), then compares.

| Result | |
| --- | --- |
| Files compared | 272 |
| Present upstream, absent here | **0** |
| Present here, absent upstream | **0** |
| Content differing after normalization | 93 |

Every one of the four differences I sampled is a Prettier reflow forced by names four characters longer: a wrapped union gaining a leading `|`, the import sorter reordering statements, symbol order inside one import, and JSDoc prose rewrapping. The individual reviewer reached the same conclusion independently on a different sample.

**Stated precisely, because the distinction matters:** this establishes that the file sets match exactly and that the sampled differences carry no meaning. It does **not** prove byte-level equivalence across all 93. A token-level comparison would, and I judged that cost not worth paying against two independent samples agreeing — but the claim on the record is the weaker one.

### What this round fixed

Three failures, one of which I caused in the previous round.

Switching the suites to a bare Node loader fixed `ttsx: not found` and broke something worse: the contributor descriptor resolves its Go source from `__dirname`, which exists only under CommonJS evaluation, so an ESM loader died with `__dirname is not defined` before a case ran. That is fixing a symptom and cracking a different layer, and it is the same pattern that cost the previous pull request three rounds. They now invoke the launcher by path — no bin link required, CommonJS evaluation preserved.

`assert-project-layout.test.cjs` requires every published package to appear in the project skill's Layout. The entry earns its space by recording the least obvious thing about this package: a contributor ships Go *source*, and a `go.mod` inside the copied directory is rejected outright.

The benchmark corpora were formatted rather than exempted. They are agent inputs, so reformatting is not free — but no measurement has been produced from them here, the existing graph benchmark prompts are formatted, and every entry in `.prettierignore` rests on a stronger reason than convention (an owner instruction, a parser that breaks). Aligning the baseline now beats carrying an exemption that would have to justify itself later.

### Standing record

Deliberate, and stated rather than buried: the lane contract moved from 12 to 13 lanes and 7 to 8 scoped builds; `scripts/build-platforms.cjs` gained a package the release path was skipping; and five path-to-lane selections are pinned explicitly, the important one being that `packages/lint/**` now selects the evidence lane.

CI is settling on `59cc1c618`. Further rounds follow until one produces nothing.

**Not for merge**, at the repository owner's instruction.

samchon (commented):
## Overall Self-Review — round 3 (re-vendored tree, `5ad98465e` → `982d748eb`)

The upstream repository had moved on substantially, so this round is over a fresh copy rather than a patched one. Three findings, one of which no CI lane could ever have produced.

### The adaptation is now a script, not a sequence of edits

Last time I moved the tree by hand and the rename reached identifiers but not filenames — 109 broken specifiers that only surfaced when a reviewer resolved them. This time the whole adaptation is one idempotent script that ends by resolving every relative specifier in all four trees and printing the count. It reports 407 checked, 0 unresolved.

It also caught its own defect immediately: `TREES` listed `tests/test-evidence/src` but not the package root, so the suite manifest kept `@samchon/lint-plugin-evidence` and `pnpm install` refused it by name. That is the difference between a step that ends in a measurement and a step that ends in an assumption.

### The finding CI cannot produce

Flattening the upstream benchmark skill from two levels to one rewrote the cross-group links (`../measurement/integrity.md`) and the top-level ones, and silently missed the links a document makes to a **peer in its own directory** — upstream `measurement/running.md` points at `plain-review.md`, which flattened is `measurement-plain-review.md`.

Twenty-one links pointed at files that do not exist. Nothing compiles, lints, or tests a relative link inside a skill document; it simply fails the next agent that follows it. Checking all 47 against disk is what found it, and the check now reports 0 broken.

### Two CI failures, two different problems that look identical

Both read as "a module cannot be found", and the causes share nothing.

`windows-go` runs `pnpm run test:go`, which now includes the evidence rule tests, and those stat `packages/evidence/lib`. I had given the `go` lane the build for exactly that reason and not its Windows twin, so the same test failed one lane over for the same reason.

The feature fixtures junction the package's `lib` and `native` in, then a Prisma or Swagger claim asks the loader to `require` its parser. Walking that junction back into the workspace is not something a fixture can rely on here, because the loaders are reached through ttsc's runtime hooks, which serve a module under the path it was requested by rather than its physical one. The fixture now links `@prisma/prisma-schema-wasm` and `yaml` beside the package, the way an install would, and the suite declares both. The two failing cases — `test_evidence_watch_reuses_an_unchanged_prisma_schema` and `..._swagger_document` — both go through the `createProject` helper that was changed, so the fix lands on them directly.

### Fidelity, measured again after the fresh copy

| | |
| --- | --- |
| Files compared against upstream | 272 |
| Present upstream, absent here | 0 |
| Present here, absent upstream | 0 |
| Relative specifiers unresolved | 0 of 407 |
| Skill links broken | 0 of 47 |
| Validation-plan spec | 11 pass, 0 fail |

The 92 files whose content differs after normalization are Prettier reflow from four-character-longer names, sampled and confirmed on both this copy and the previous one. Stated precisely: this proves the file sets match and the sampled differences carry no meaning; it is not a byte-level equivalence proof.

### New in this pass

The upstream benchmark **operation** skill (710 lines) arrives as `evidence-benchmark`, separate from this repository's `benchmark` skill. That one owns ttsc's own performance and graph harnesses; this one measures how a coding agent behaves with and without the evidence graph, and the two share no structure, so AGENTS.md's merge rule does not apply. `lint-rule-authoring` and `wiki` still stay out, because that rule does apply to them.

**Not for merge**, at the repository owner's instruction.

samchon:
## Handover

Written for whoever picks this up next. Read it end to end before touching the
branch: most of what follows is a fact that cost a CI cycle to learn, and
several are invariants that look like dead code until you delete one.

---

### 0. Standing instructions from the repository owner

These override anything you would otherwise infer.

1. **Do not merge this pull request.** Stated explicitly and repeated. It is
   open for review only. Everything else here is subject to change; this is not.
2. **Do not build or test locally.** Push and let GitHub Actions do it. The
   exceptions that have proven safe and are used below are *checks* rather than
   builds: `npx prettier --check`, `node --test scripts/ci/validation-plan.test.cjs`,
   `node scripts/ci/format-check.cjs`, `gofmt`, and the branch-local audit
   script. A full `pnpm install`/`pnpm build` here is not.
3. **Answer the owner in Korean.** Repository artifacts — commits, PR text,
   issues, code comments, this comment — stay in English.
4. **Commit and push every edit.** Do not accumulate uncommitted work.
5. **Do not play whack-a-mole.** This was said directly, after several rounds of
   fix-one-CI-failure-push-repeat. Sweep for the whole class before pushing.
   `experimental/evidence-vendor/audit.cjs` exists because of that instruction.
6. Do not launch multi-agent workflows unless asked by name.

---

### 1. What this branch is

Vendors `samchon/evidence` (published as `@samchon/lint-plugin-evidence`) into
this repository as `@ttsc/evidence`, with its benchmark and both feature suites.

| Issue | Scope |
| --- | --- |
| #1074 | `packages/evidence` as `@ttsc/evidence`: TS source, Go rules, manifest, Go module wiring, type renames |
| #1075 | the 66-case e2e suite and 436 Go rule tests, with a CI lane that runs them |
| #1076 | the benchmark into `experimental/benchmark/evidence`, its suite into `tests/test-evidence-benchmark` |
| #1077 | README, website page, `.agents/skills/evidence-graph`, AGENTS.md index |

Branch `feat/evidence-package`. 27 commits, 479 files, ~73k insertions.

The counts above were stale for most of this branch's life (79/64/29 from the
first copy). They are re-measured; if you re-vendor and the counts move, fix
them in the issues and the PR body too.

---

### 2. Where things are

```
packages/evidence/                     the plugin: src/ (24 TS), native/ (23 Go + 65 _test.go)
experimental/benchmark/evidence/       the benchmark: src/ + template/ requirements/ instructions/
tests/test-evidence/                   66 e2e feature cases
tests/test-evidence-benchmark/         6 benchmark feature cases
.agents/skills/evidence-graph/         domain model skill (1 file)
.agents/skills/evidence-benchmark/     benchmark operation skill (11 files, flattened from upstream's 2 levels)
experimental/evidence-vendor/          branch-local tooling: readapt.cjs, audit.cjs, flatten-skill.cjs, fix-skill-links.cjs
website/src/content/docs/lint/evidence.mdx
```

Upstream is expected at `D:/github/samchon/evidence` on the owner's machine.

**Read `experimental/evidence-vendor/README.md` before refreshing anything.** It
gives the copy order and the flags that matter.

---

### 3. Invariants — delete one of these and CI goes red in a way that reads like something else

**3.1 Two roots, not one.** Upstream's benchmark sits at
`<repository>/benchmark`, so a single value answered both *which repository owns
the git history, the pnpm catalog, and `packages/evidence`* and *where are the
benchmark's own template/requirements/instructions/output*. Here those are
different directories. `EvidenceBenchmarkLayout` holds the one literal that
knows where the package sits and derives both roots. Every call site goes
through `assetsRoot(repository)`; the executables use `repositoryRoot` instead
of each counting `../..` from `__dirname`.

**3.2 `workspacePackageVersions` must stay.** Upstream deleted it, correctly for
itself: upstream lists `ttsc`, `@ttsc/lint`, and `@ttsc/unplugin` in its own
pnpm catalog because there they are external dependencies. Here they *are* the
workspace, and a workspace never lists itself in a catalog — so a catalog lookup
cannot answer `{{version:ttsc}}` and `prepareWorkspace` throws before a cell
writes anything. `readapt.cjs` step 3b re-adds it. All twelve template tokens
were verified to resolve.

**3.3 The frozen assets are Prettier-exempt.** `.prettierignore` covers
`experimental/benchmark/evidence/{template,requirements,instructions}`. The
template carries `{{name}}` substitution tokens; Prettier reads `{{name}}` in a
`.tsx` as an object literal and writes `{{ name }}`, which no longer matches the
token, survives substitution, and reaches the delivered workspace as JSX that
does not type-check. That is exactly how two cases failed, in both arms. The
requirements are the frozen inputs a run is identified by, so reflowing their
tables changes what the measured agent read. The benchmark's own `src` is this
repository's code and stays formatted.

**3.4 Never exclude a directory named `lib` when copying the template.** It
ships frontend sources in `src/lib`. Excluding it delivers a workspace whose
entry cannot resolve `@/lib/client`. `.gitignore`'s `lib/` rule would also hide
them, so there is an explicit `!experimental/benchmark/evidence/template/**/src/lib/`
exception with the reason attached. `packages/evidence/native/` has a sibling
exception for the same `.gitignore` rule, for a different reason: a lint
contributor ships Go **source**, and ttsc rejects a `go.mod` inside the `source`
directory, so the sources sit in a plain `native/` that collides with the
per-platform packages' binary drop.

**3.5 Go indentation.** Upstream Go is tab-indented (gofmt standard). This
repository pins `.vscode/gofmt-2spaces.sh` — gofmt, then tabs to two spaces —
and `scripts/ci/format-check.cjs` enforces it over every tracked `.go`. Run the
wrapper after every copy or the typecheck lane fails.

**3.6 Bare `ttsx`/`ttsc` in a package script does not work here.** pnpm links
those bins from the workspace `ttsc` at install time, and `ttsc/lib` does not
exist yet then — the install log says so, once per suite. Every script names the
launcher by path: `node ../../../packages/ttsc/lib/launcher/ttsx.js`. Depth
differs by package; `pnpm --filter <pkg> <script>` runs in the package
directory, so the documented commands still work unchanged.

**3.7 The fixture links dependencies from the manifest, not from a list.**
`tests/test-evidence/src/internal/createProject.ts` reads
`packages/evidence/package.json` and links every entry in `dependencies`. The
package's `lib` is junctioned in and Node cannot be relied on to walk that
junction back into the workspace, because the loaders are reached through ttsc's
runtime hooks, which serve a module under the path it was *requested by* rather
than its physical one. Hardcoding the list is how `@typia/utils` was missed
after `@prisma/prisma-schema-wasm` was fixed.

**3.8 The suites inherit the benchmark's dependencies.**
`tests/test-evidence-benchmark` imports the benchmark's *source* across a
package boundary, so it must declare what that source imports — currently
`yaml`, via `EvidenceBenchmarkWorkspace`.

**3.9 The benchmark's executables are read by no suite.** `test-typecheck.cjs`
discovers `tests/*` only, and nothing imports
`experimental/benchmark/evidence/src/executable/**`. The `evidence` lane runs
`pnpm --filter @ttsc/evidence-benchmark check` ahead of the two suites so that
code is read at all. Do not drop that step; it caught an unparseable file within
one commit of being added.

**3.10 Path literals inside the benchmark describe the workspace it
*generates*.** `packages/api`, `packages/backend`, `packages/frontend`,
`.agents/skills/backend`, `config/lint.config.ts`, `tests/journeys/` — these are
correct as written and must not be rewritten to this repository's layout. Only
`benchmark/...` prefixes were re-pointed.

---

### 4. CI

Lane `evidence` in `scripts/ci/validation-plan.cjs`:

```
build: pnpm run build:current      (TTSC_BUILD_SCOPE=test-evidence)
run:   pnpm --filter @ttsc/evidence-benchmark check
    && pnpm --filter test-evidence start
    && pnpm --filter test-evidence-benchmark start
```

`scripts/build-current.cjs` gained `test-evidence: ["ttsc", "@ttsc/lint", "@ttsc/evidence", PLATFORM]`
and `@ttsc/evidence` at the end of `full`. `go` and `windows-go` both build the
package because `test:go` now includes its rule tests. Selectors route
`packages/evidence/**` to `["evidence", "go"]`, `packages/lint/**` to a set that
includes `evidence`, and both benchmark trees to `evidence`.
`scripts/ci/validation-plan.test.cjs` pins 13 lanes, 8 scoped builds, and four
evidence selections; `scripts/ci/test-owners.cjs` claims both suites and the Go
package.

**Last fully green run: `3197a8357`, 28/28.** That is the commit that restored
the frozen assets. Two commits have landed since (`6b561a82c` re-vendor,
`5b5d20bdb` tooling); at the time of writing their run is still in flight. Check
it before assuming anything.

Known flake: `platform (darwin-x64)` has failed once with
`ttsc: platform-specific TypeScript binary not found (@typescript/typescript-darwin-x64)`.
That is npm skipping typescript's optional dependency in the clean-room install;
nothing on this branch touches it, and the same lane passed on the neighbouring
commit. Re-run the job rather than changing code. `ci` is the aggregate gate job
(`needs: plan, platform-integrations, test`) — it fails as a consequence, never
independently.

---

### 5. Refreshing from upstream

The owner edits upstream while this branch is open, and asked directly whether
their latest prompt and logic changes had landed. They had not: the prompts had,
the runner had not. Assume drift on every session.

Procedure and rationale: `experimental/evidence-vendor/README.md`.
Short form: robocopy each tree, `node experimental/evidence-vendor/readapt.cjs`,
re-flatten the skill, run both formatters, then `node experimental/evidence-vendor/audit.cjs`.

`readapt.cjs` is idempotent and its `edit()` helper refuses an insertion whose
replacement text contains its own anchor unless given an explicit marker —
three edits had that shape and silently duplicated `benchmarkRoot` and
`linkEvidenceRuntimeDependencies` on a second run.

Verify with measurements, not inspection:

- File sets against upstream, **compared through `git ls-files`, not the
  filesystem** — a `.gitignore` rule hid two template sources from a
  disk-to-disk comparison that reported a clean 0/0.
- Every relative specifier resolves (`readapt.cjs` step 5 prints the count;
  expect exactly one false positive, `./functional.js`, a fixture string).
- All 47 skill links resolve (`fix-skill-links.cjs` re-verifies).
- `audit.cjs`: `xref`, `link`, and `bin` should be empty; one `dependency` hit
  (`@org/api`, a fixture string) is expected.

---

### 6. What is left

1. Confirm the run on `5b5d20bdb` settles green. If `evidence defenses` fails,
   read the case's `Actual output:` block — `assertStatus` includes the real
   command output, and grepping only for `Error` misses it.
2. Keep running Overall Self-Review rounds until one produces nothing.
3. Re-check the numeric claims in #1074/#1075/#1076 and the PR body if any
   re-vendor changes the file counts.
4. **Leave the pull request unmerged.**

Open question for the owner, recorded in the PR body under *Deliberately left
out*: upstream's `lint-rule-authoring` skill is not vendored. It is **not** a
restatement of `packages/lint/README.md` — it carries Go-authoring guidance this
repository lacks (that `As*` accessors are type assertions and panic rather than
returning nil, so a nil check after one is dead code; which defaults cost the
most; the failure policy; the reporting shape). Whether this repository wants a
Go rule-authoring skill is the owner's call, not something to fold into a
vendoring pull request. Upstream's `wiki` skill is about upstream's own wiki and
stays out.

---

### 7. Mistakes made here, so they are not repeated

- **Accepting a review framing without verifying it.** Cost two extra rounds on
  the entry-path work in the preceding cycle. AGENTS.md says evidence precedes
  correction; it applies to reviewer claims too.
- **Fixing the symptom that failed first.** Two of four runtime dependencies,
  then the third. One lane's build, then its Windows twin. The habit that
  produced the owner's "stop playing whack-a-mole" instruction.
- **Silencing the check that would have caught it.** Prettier was run over the
  tree and piped through `tail -1`, discarding a parse error it had just
  printed. The file shipped unparseable.
- **Asserting a document's contents without opening it.** Twice: the exclusion
  that handed the Go rule API to a skill containing no mention of it, and the
  claim that `lint-rule-authoring` merely restates the lint README — written one
  commit after committing a fix for exactly that mistake.
- **Verifying against the filesystem when git is what ships.** Reported 0
  missing while `.gitignore` was dropping two template sources.
- **Trusting a blanket re-copy.** It cannot distinguish an upstream change from
  an adaptation this workspace requires; §3.2 is the case that would have
  broken every benchmark cell.

Issue #1084, opened 2026-08-06
graph benchmark publish resolves a relative --from against the repository root, then reports success after folding nothing


samchon:
## Problem

`benchmarks/graph`'s publisher resolves a relative `--from` against the
**repository root**. Every other path a graph-benchmark operator types —
`--out` on both runners, `--dir` / `--report` / `--out` on the trace auditor —
resolves against the **current working directory**, which `pnpm --dir
benchmarks/graph run <script>` sets to `benchmarks/graph`.

So the exact relative string that names a run directory when it is written names
a different, non-existent directory when it is published. And publishing from a
directory that holds no report is not an error: the run reports success, folds
nothing, and with `--reset` overwrites the committed dashboard with an empty
agent-cell list.

This is the documented parallel-sweep workflow (`benchmarks/graph/README.md:141`:
"For parallel `--no-website` runs, publish afterward with `pnpm --dir
benchmarks/graph run publish -- --from <out-dir>`").

### The two bases

`benchmarks/graph/src/TtscBenchmarkGraphPublisher.ts:567-590`:

```ts
function parseSourceDirs(
  argv: readonly string[],
  repositoryRoot: string,
  workRoot: string,
): string[] {
  const dirs: string[] = [];
  for (let i = 0; i < argv.length; i++) {
    const arg = argv[i]!;
    if (arg === "--from" || arg === "--source") {
      const next = argv[++i];
      if (!next) throw new Error(`${arg} requires a directory`);
      dirs.push(path.resolve(repositoryRoot, next));      // <- repository root
    } else if (arg.startsWith("--from=")) {
      dirs.push(path.resolve(repositoryRoot, arg.slice("--from=".length)));
    } else if (arg.startsWith("--source=")) {
      dirs.push(path.resolve(repositoryRoot, arg.slice("--source=".length)));
    }
  }
  const selected = dirs.length > 0 ? dirs : [path.join(workRoot, "graph")];
  ...
}
```

Sibling entrypoints, all `path.resolve` with no base — that is cwd:

| Entry | Flag | Base | Source |
| --- | --- | --- | --- |
| agent runner | `--out` | cwd | `TtscBenchmarkGraphRunner.ts:293` |
| cold-index runner | `--out` | cwd | `TtscBenchmarkGraphIndexTime.ts:151` |
| trace auditor | `--dir` | cwd | `TtscBenchmarkGraphTraceAuditor.ts:382` |
[...]

Issue #1083, opened 2026-08-06
Benchmark run reports land under a root neither harness designates for them


samchon:
## Problem

Both first-party benchmark harnesses resolve a work root, then write their
report somewhere that root does not govern.

- `benchmarks/graph` keeps two roots on purpose: the **fixture clone** root
  (`TTSC_GRAPH_BENCH_WORK`, default `<repo>/../graph-benchmark-work`) and the
  **artifact** root (`benchmarks/graph/.work`). The agent runner, the suite, and
  the structural harness all write reports under the artifact root.
  `index-time` writes its report under the fixture clone root instead.
- `benchmarks/performance` keeps one root (`TTSC_BENCH_WORK`). Clones and the
  resume checkpoint follow an override of it; the report and its sibling `.json`
  do not — they stay pinned to `benchmarks/performance/.work`.

Neither is a wrong number. Both put a run's output where the operator did not
ask for it, and the performance case makes the documented isolation mechanism
stop isolating.

### `benchmarks/graph`: the cold-index report lands among the fixture clones

`benchmarks/graph/src/TtscBenchmarkGraphIndexTime.ts`:

```ts
const workDir = TtscBenchmarkGraph.resolveWorkDir(repoRoot);   // line 120
...
const outDir = path.resolve(                                   // line 151
  parsed.values.out ?? path.join(workDir, "graph-index", timestamp()),
);
const reportPath = path.join(outDir, "report.json");           // line 154
```

`TtscBenchmarkGraph.resolveWorkDir` is the **fixture clone** root:

```ts
export function resolveWorkDir(repositoryRoot: string): string {
  return (
    process.env.TTSC_GRAPH_BENCH_WORK ??
    path.resolve(repositoryRoot, "..", "graph-benchmark-work")
  );
}
```

Every sibling harness writes its report under the package artifact root instead:

| Harness | Report root | Source |
| --- | --- | --- |
| agent runner | `TtscBenchmarkConstant.WORK_ROOT` + `graph/<timestamp>` | `TtscBenchmarkGraphRunner.ts:293-297` |
| suite | `TtscBenchmarkConstant.WORK_ROOT` + `graph/…` | `TtscBenchmarkGraphSuite.ts:162-170`, `:264-269` |
| structural | `TtscBenchmarkConstant.ROOT` + `.work` | `TtscBenchmarkGraphStructural.ts:40-43` |
[...]

Issue #1082, opened 2026-08-05
@ttsc/graph shards exclude external nodes but keep the edges leaving them, so a project declaring globals in its own .d.ts cannot be queried


konradekk:
**Disclaimer:** this bug report was written with a help of AI/LLM.

---

From `0.24.0`, the MCP server answers no request at all against our project:

    @ttsc/graph: ttscgraph: edge source node is absent from shard facts:
    src/function.d.ts#assert:function

`lookup`, `trace` and `details` fail alike, and `includeExternal: true` changes nothing — the refusal comes from shard assembly, before the requested operation runs. `ttsc-graph dump` is unaffected: it emits the whole graph (21 MB, schemaVersion 6, one diagnostic) without complaint.

`0.23.0` answers the same request, against the same project, **using the same native binary**:

| driver               | native binary   | one `lookup` over stdio                       |
| -------------------- | --------------- | --------------------------------------------- |
| `@ttsc/graph` 0.24.0 | 0.24.0          | `edge source node is absent from shard facts` |
| `@ttsc/graph` 0.23.0 | the same 0.24.0 | answers                                       |

Both releases locate `ttscgraph` the same way — from `ttsc`, resolved at the project root — so with
`--cwd` at one project they load one binary. Calling `resolveGraphBinary` from each package confirms a single path:

    0.23.0 JS → <project>/node_modules/@ttsc/darwin-arm64/bin/ttscgraph   (0.24.0)
    0.24.0 JS → <project>/node_modules/@ttsc/darwin-arm64/bin/ttscgraph   (0.24.0)

The refusal text belongs to the native engine, and `TtscGraphSession.js:169` forwards it verbatim — but the engine is the component both versions share. What changed is the JavaScript driving it, or what it asks for.

## Reproducing

A project that declares ambient globals in its own `.d.ts`, alongside implementations that assign them:

```ts
// src/function.d.ts
declare global {
    function ignore(error: unknown): void;
    function unexpected(label: string, value: never, fallback: unknown): never;
    function assert(condition: unknown | (() => unknown), message: string): void | never;
}
export {};
```

```ts
// src/function/index.ts
window.assert = (condition, message) => {
    /* throws new Error(message) */
};
```

Any `inspect_typescript_graph` request then fails under `0.24.0` and succeeds under `0.23.0`.

## What the graph holds

The node the message names is present and well-formed:

```json
{
    "id": "src/function.d.ts#assert:function",
    "kind": "function",
    "name": "assert",
[...]

Issue #1080, opened 2026-08-05
Stale references left by the benchmark package split


samchon:
Two read-only diagnostic passes over `benchmarks/graph` and `benchmarks/performance` after the `experimental/benchmark` split (#1078) found no functional defect in either harness. Both were verified by running their entrypoints, recomputing every constant, and diffing each moved source against its pre-move revision.

What survived is a set of stale references the move left behind. None changes behavior; each misleads whoever follows it.

## `benchmarks/graph`

| File | What it says | What is true |
| --- | --- | --- |
| `src/TtscBenchmarkGraphPublisher.ts:22-25` | usage header spells `pnpm --dir benchmarks/graph publish` | `publish` is a pnpm subcommand, so that invokes pnpm's publisher against `@ttsc/benchmark-graph`. The package README already says to spell these `pnpm run <script>`; the workflow and the skill were corrected and this header was missed |
| `src/TtscBenchmarkGraphIndexTime.ts:5` | "the agent benchmark (`src/executable/graph/index.ts`)" | `src/executable/index.ts` |
| `src/TtscBenchmarkGraphRunner.ts:36` | "moving this implementation out of `src/executable/graph`" | that directory no longer exists; the rationale the sentence states is still true and still enforced |
| `src/TtscBenchmarkGraphIndexTime.ts:117` | `const benchmarkRoot = TtscBenchmarkConstant.ROOT;` | assigned, never read. It was the base for the `repoRoot` computation the root-resolution fix removed, and `noUnusedLocals` is not set so nothing reports it |
| `.agents/skills/benchmark/graph.md:59` | "Use the `graph:audit` package command" | the script is `audit`; every command block in the same file was rewritten and this prose line was missed |

## `benchmarks/performance`

| File | What it says | What is true |
| --- | --- | --- |
| `.agents/skills/benchmark/performance.md:18` | the flag and environment table lives in `benchmarks/README.md` | no such file. The table is in `benchmarks/performance/README.md` |

## Noted, pre-existing, not part of this

Both predate the split and are byte-identical to their pre-move form. They are recorded here so a later reader does not rediscover them as move damage.

- `benchmarks/performance/src/TtscBenchmarkPerformanceExecutable.ts:46-48` — `TTSC_BENCH_OUT` defaults against the package constant rather than the resolved `workRoot`, so `TTSC_BENCH_WORK=/elsewhere` relocates the clones and the checkpoint but leaves `report.md` in `benchmarks/performance/.work/`.
[...]

PR #1085, opened 2026-08-06
fix: ttsx dependency path identity, graph external edge sources, and benchmark harness references

samchon:
Single cycle pull request for the issue campaign over #1073, #1080, and #1082, plus the two issues its discovery rounds published.

## Cycle scope

| Issue | Subject |
| --- | --- |
| #1073 | ttsx's dependency lane compares an unresolved `rootDir` against a resolved source path |
| #1080 | Stale references left by the benchmark package split |
| #1082 | `@ttsc/graph` shards exclude external nodes but keep the edges leaving them |
| #1083 | Benchmark run reports land under a root neither harness designates for them |
| #1084 | graph benchmark `publish` resolves a relative `--from` against the repository root, then reports success after folding nothing |

## Discovery gate

Two complete full-scope rounds ran against `master` @ `9450025b4`. Round 1 published #1083 and #1084; Round 2 produced no meaningful candidate, which froze the accepted set above.

## What landed

- **#1082** — `assignedFunctionTarget` no longer re-attributes an assigned function body to a symbol declared outside the workspace. That single cause produced all three reported symptoms: the unassemblable shard, the evidence spans pointing at neither file, and the unset implementation source. Two Go tests cover both branches of the externality classifier, plus one serve-level test for the reported symptom.
- **#1073** — a dependency's `rootDir` is now produced by the same pass the served source is, in both the declared and the undeclared branch, and a marker's root is resolved on read. One e2e test covering both branches, one deterministic test for the reader.
- **#1080** — six stale references corrected, and the usage-header rule widened to the two script names pnpm actually shadows.
- **#1083** — each harness's report now lands under the root that run resolved, and the graph environment table lists every variable the package consults.
- **#1084** — a cell report path is absolute at the writer, so the publisher, the trace auditor, and the suite share one base; a source directory that contributes no measurement is refused instead of publishing an empty dashboard; the auditor no longer drops an unreadable cell in silence.

## Review

One read-only Individual Self-Review per pushed issue-implementation commit, each adjudicated and recorded as a formal `COMMENT` review below. One of them caught a blocking regression before it shipped: dropping the publisher's repository-root probe would have aborted every publish, because the field it resolves is written by the agent runner rather than by the suite runner. Overall Self-Review runs solo over the whole base-to-head diff and restarts after every correction.

Closing keywords are carried by the commits that earn them, so the merge closes exactly what landed.

### DIFF
--- a/benchmarks/graph/README.md
+++ b/benchmarks/graph/README.md
@@ -123,7 +123,7 @@ Use `pnpm --dir benchmarks/graph run audit -- --compare=<before>,<after>` on aud
 
 ## Environment overrides
 
-This table is the list. Every variable either runner reads is here.
+This table is the list. Every environment variable this package consults, `PATH` aside, is here.
 
 | Variable | Default | Meaning |
 | --- | --- | --- |
--- a/benchmarks/graph/README.md
+++ b/benchmarks/graph/README.md
@@ -23,6 +23,8 @@ pnpm --dir benchmarks/graph run publish -- --from <out-dir>             # publis
 
 `publish` is also a pnpm subcommand, so always spell these as `pnpm run <script>` rather than `pnpm <script>`.
 
+Every directory these commands take — `--out`, `--from`, `--dir`, `--report` — resolves against the current directory, which `pnpm --dir benchmarks/graph` sets to this package. A `--from` directory holding no report is refused rather than published as a run that contributed nothing.
+
 ## Layout
 
 - `src/executable/`: Short, export-free CLI bootstraps.
--- a/benchmarks/graph/README.md
+++ b/benchmarks/graph/README.md
@@ -21,9 +21,9 @@ pnpm --dir benchmarks/graph run index-time -- --all                     # cold i
 pnpm --dir benchmarks/graph run publish -- --from <out-dir>             # publish a completed --no-website suite
 ```
 
-`publish` is also a pnpm subcommand, so always spell these as `pnpm run <script>` rather than `pnpm <script>`.
+`publish` and `audit` are also pnpm subcommands, so always spell these as `pnpm run <script>` rather than `pnpm <script>`. `pnpm audit` reports dependency vulnerabilities; it is not this package's trace auditor.
 
-Every directory these commands take — `--out`, `--from`, `--dir`, `--report` — resolves against the current directory, which `pnpm --dir benchmarks/graph` sets to this package. A `--from` directory holding no report is refused rather than published as a run that contributed nothing.
+Every directory these commands take (`--out`, `--from`, `--dir`, `--report`) resolves against the current directory, which `pnpm --dir benchmarks/graph` sets to this package. A source directory that contributes no measurement, named or defaulted, is refused rather than published as a run that added nothing.
 
 ## Layout
 
@@ -123,12 +123,23 @@ Use `pnpm --dir benchmarks/graph run audit -- --compare=<before>,<after>` on aud
 
 ## Environment overrides
 
+This table is the list. Every variable either runner reads is here.
+
 | Variable | Default | Meaning |
 | --- | --- | --- |
 | `TTSC_GRAPH_BENCH_WORK` | `../graph-benchmark-work` | Fixture clone directory, checkouts only. It defaults outside the repo so a measured agent does not inherit ttsc's `CLAUDE.md` / `AGENTS.md` from a parent directory. Reports never follow it; they stay under `.work/`. |
 | `TTSC_GRAPH_BENCH_OUT` | `.work/graph/<timestamp>` | Report directory for `src/executable/index.ts`, below `--out` and above the default. |
-| `TTSC_GRAPH_BENCH_TIMEOUT_MS` | `1800000` | Per-child timeout for an agent sample (`src/executable/index.ts`) and a cold-index build step (`src/executable/index-time.ts`). |
+| `TTSC_GRAPH_BENCH_TIMEOUT_MS` | `1800000` | Timeout on every child either runner spawns: an agent cell, a comparator index build, a fixture clone or install, the trace-audit pass, and the `ttscgraph` build. Not a per-sample budget: one agent child runs `arms × --runs` samples. |
+| `TTSC_BENCH_CONCURRENCY` | unlimited | Cap on concurrently launched agent samples. A low cap keeps the host quiet enough for per-run timings and token counts to settle. |
+| `TTSC_BENCH_REQUIRE_QUIET` | - | `1` turns `src/executable/index-time.ts`'s host-load warning into a hard error. Set it for every publication run. |
+| `TTSC_BENCH_SKIP_LOAD_CHECK` | - | `1` disables that host-load check. It only ever bites on POSIX; `os.loadavg()` reports zeros on Windows. |
+| `TTSC_CLAUDE_RUN_TIMEOUT_MS` | `900000` | Budget for one Claude Code sample, when `--claude-run-timeout-ms` is not passed. |
+| `TTSC_CLAUDE_STARTUP_GRACE_MS` | `5000` | Delay before the prompt reaches a graph-arm Claude Code sample, so its MCP server is up first. `--claude-startup-grace-ms` overrides it; `0` disables the delay. |
+| `CODEX_MCP_STARTUP_TIMEOUT_SEC` | Codex's own | MCP startup timeout written into the generated Codex config, when `--mcp-startup-timeout-sec` is not passed. |
+| `CODEX_MCP_TOOL_TIMEOUT_SEC` | Codex's own | MCP tool-call timeout written into the generated Codex config, when `--mcp-tool-timeout-sec` is not passed. |
 | `CODEBASE_MEMORY_MCP_BINARY` | `codebase-memory-mcp` | Binary used by the `codebase-memory` comparator when `--codebase-memory-binary` is not passed. |
+| `TTSC_BENCH_CBM_MODE` | codebase-memory's own (`full`) | Index mode passed to `codebase-memory-mcp` (`full`, `moderate`, `fast`). `fast` is the only mode that indexes `vscode` without the full mode's memory blowup, and the mode is recorded on the cell. |
+| `CBM_LOG_LEVEL` | `warn` | Log level of the `codebase-memory` comparator's index build. |
 | `SERENA_MCP_COMMAND` | `uvx` | Command used to launch Serena when `--serena-command` is not passed. |
 | `SERENA_MCP_ARGS` | Serena's `uvx --from git+https://github.com/oraios/serena ...` args | Argument list used to launch Serena when `--serena-args` is not passed. |
 | `SERENA_SOURCE` | `git+https://github.com/oraios/serena` | Package source the Serena launcher installs from when `--serena-source` is not passed. |
--- a/benchmarks/graph/README.md
+++ b/benchmarks/graph/README.md
@@ -123,7 +123,7 @@ Use `pnpm --dir benchmarks/graph run audit -- --compare=<before>,<after>` on aud
 
 ## Environment overrides
 
-This table is the list. Every environment variable this package consults, `PATH` aside, is here.
+Every environment variable this package consults, `PATH` aside, is here.
 
 | Variable | Default | Meaning |
 | --- | --- | --- |
--- a/benchmarks/graph/README.md
+++ b/benchmarks/graph/README.md
@@ -123,7 +123,9 @@ Use `pnpm --dir benchmarks/graph run audit -- --compare=<before>,<after>` on aud
 
 | Variable | Default | Meaning |
 | --- | --- | --- |
-| `TTSC_GRAPH_BENCH_WORK` | `../graph-benchmark-work` | Fixture clone directory. It defaults outside the repo so a measured agent does not inherit ttsc's `CLAUDE.md` / `AGENTS.md` from a parent directory. |
+| `TTSC_GRAPH_BENCH_WORK` | `../graph-benchmark-work` | Fixture clone directory, checkouts only. It defaults outside the repo so a measured agent does not inherit ttsc's `CLAUDE.md` / `AGENTS.md` from a parent directory. Reports never follow it; they stay under `.work/`. |
+| `TTSC_GRAPH_BENCH_OUT` | `.work/graph/<timestamp>` | Report directory for `src/executable/index.ts`, below `--out` and above the default. |
+| `TTSC_GRAPH_BENCH_TIMEOUT_MS` | `1800000` | Per-child timeout for an agent sample (`src/executable/index.ts`) and a cold-index build step (`src/executable/index-time.ts`). |
 | `CODEBASE_MEMORY_MCP_BINARY` | `codebase-memory-mcp` | Binary used by the `codebase-memory` comparator when `--codebase-memory-binary` is not passed. |
 | `SERENA_MCP_COMMAND` | `uvx` | Command used to launch Serena when `--serena-command` is not passed. |
 | `SERENA_MCP_ARGS` | Serena's `uvx --from git+https://github.com/oraios/serena ...` args | Argument list used to launch Serena when `--serena-args` is not passed. |
@@ -136,6 +138,7 @@ Use `pnpm --dir benchmarks/graph run audit -- --compare=<before>,<after>` on aud
 | `.work/graph/<timestamp>/report.json` | Per-cell agent samples: tokens, cached and reasoning tokens, turns, tool calls, reads, greps, shell calls, graph calls, source touches, cost, wall time, and attempt counts. |
 | `.work/graph/<timestamp>/codex-trace-audit.json` | Codex trace audit written automatically for Codex cells: full exposed message timeline, tool-call ledger, reasoning token counts, visible-input ledger, baseline-median savings, duplicate-output exact savings, graph-replaceable shell-output surface, candidate MCP overfetch estimates, and observed later-turn prompt replay exposure. |
 | `.work/graph/structural/report.json` | Deterministic structural metrics from `src/executable/bench.ts`: load time, graph build time, and the share the build adds on top of the load it rides. |
+| `.work/graph-index/<timestamp>/report.json` | Cold index build time per (tool × fixture) from `src/executable/index-time.ts`, with the per-fixture scale block and the host it ran on. |
 | `website/public/benchmark/graph.json` | Graph dashboard data consumed by https://ttsc.dev/benchmark. `src/executable/index.ts` upserts measured agent cells by the visible axes: harness, tool, repo, prompt id or family, stable model tier, and daemon mode. Fixture branch, effort, and setup time remain metadata and never fork a visible cell. `src/executable/publish.ts` folds local reports in, replacing the `structural` block whole and upserting agent cells by key. `src/executable/index-time.ts` owns only the top-level `index` key (`{ host, scale, cells }`), upserted by (project, tool). |
 
 `.work/` is git-ignored; results are an ephemeral artifact and never committed. For parallel `--no-website` runs, publish afterward with `pnpm --dir benchmarks/graph run publish -- --from <out-dir>`; never let concurrent runners write `graph.json` directly.
--- a/benchmarks/graph/src/TtscBenchmarkGraphRunner.ts
+++ b/benchmarks/graph/src/TtscBenchmarkGraphRunner.ts
@@ -32,9 +32,9 @@ import { ITtscBenchmarkGraphWebsiteAgentCell } from "./structures/ITtscBenchmark
  * Runs the complete graph-agent benchmark while keeping its executable a
  * location-only bootstrap.
  *
- * All harness paths derive from the supplied executable directory so moving
- * this implementation out of `src/executable/graph` does not change which child
- * entrypoints are launched.
+ * All harness paths derive from the supplied executable directory, so
+ * relocating this implementation never changes which child entrypoints are
+ * launched.
  */
 export namespace TtscBenchmarkGraphRunner {
   /**
--- a/benchmarks/graph/src/TtscBenchmarkGraphRunner.ts
+++ b/benchmarks/graph/src/TtscBenchmarkGraphRunner.ts
@@ -32,9 +32,11 @@ import { ITtscBenchmarkGraphWebsiteAgentCell } from "./structures/ITtscBenchmark
  * Runs the complete graph-agent benchmark while keeping its executable a
  * location-only bootstrap.
  *
- * All harness paths derive from the supplied executable directory, so
- * relocating this implementation never changes which child entrypoints are
- * launched.
+ * Every child entrypoint it launches is derived from the supplied executable
+ * directory, so relocating this implementation never changes which bootstrap
+ * runs. Roots the harness reads or writes — the repository, the fixture work
+ * directory, the report directory, the website JSON — come from
+ * `TtscBenchmarkConstant` instead and are unaffected either way.
  */
 export namespace TtscBenchmarkGraphRunner {
   /**
@@ -622,7 +624,15 @@ export namespace TtscBenchmarkGraphRunner {
           repoDir,
           tsconfig: spec.tsconfig,
           log: path.relative(repoRoot, `${path.join(outDir, logStem)}.out.log`),
-          report: path.relative(repoRoot, copyPath),
+          // Absolute, like the suite runner already records the same field.
+          // Every reader of a cell report - the publisher, the trace auditor,
+          // the suite - then needs no base at all, and a run directory named by
+          // `--out` stays readable from whatever directory the publish is typed
+          // in. Recorded relative to the repository root it was readable from
+          // exactly one place, and only by a reader that knew to try that base.
+          // `log` stays repository-relative: it is provenance a human reads,
+          // and nothing resolves it.
+          report: copyPath,
           summary: summarize(data),
         },
         websiteCell,
--- a/.agents/skills/benchmark/graph.md
+++ b/.agents/skills/benchmark/graph.md
@@ -56,7 +56,7 @@ Never let concurrent runners write `graph.json` directly.
 
 ## Trace Audit
 
-Codex suites write `codex-trace-audit.json` automatically. Use the `graph:audit` package command only to re-audit existing output or compare before and after runs.
+Codex suites write `codex-trace-audit.json` automatically. Use the `audit` package command only to re-audit existing output or compare before and after runs.
 
 The audit records exposed assistant messages, shell and MCP calls in timeline order, per-turn usage, and `reasoning_output_tokens`. Codex does not expose hidden reasoning text; never invent it.
 
--- a/.agents/skills/benchmark/graph.md
+++ b/.agents/skills/benchmark/graph.md
@@ -52,7 +52,7 @@ pnpm --dir benchmarks/graph run publish -- --from <out-dir>
 
 Never let concurrent runners write `graph.json` directly.
 
-A relative `--out` and a relative `--from` resolve against the same directory, so name the run the same way in both. Publication refuses a source directory that holds no report instead of writing a dashboard nothing contributed to; that refusal is what keeps `--reset` from replacing the served cells with an empty set.
+A relative `--out` and a relative `--from` resolve against the same directory, so name the run the same way in both. Publication refuses a source directory that contributes no measurement, the defaulted one included, instead of writing a dashboard nothing added to; that refusal is what keeps `--reset` from replacing the served cells with an empty set.
 
 `benchmarks/graph/src/TtscBenchmarkGraphWebsiteCell.ts` is the single published-cell key. Key only by fields the website renders. Metadata such as fixture branch, reasoning effort, or setup time must not create a second visible copy of the same cell.
 
--- a/.agents/skills/benchmark/graph.md
+++ b/.agents/skills/benchmark/graph.md
@@ -52,6 +52,8 @@ pnpm --dir benchmarks/graph run publish -- --from <out-dir>
 
 Never let concurrent runners write `graph.json` directly.
 
+A relative `--out` and a relative `--from` resolve against the same directory, so name the run the same way in both. Publication refuses a source directory that holds no report instead of writing a dashboard nothing contributed to; that refusal is what keeps `--reset` from replacing the served cells with an empty set.
+
 `benchmarks/graph/src/TtscBenchmarkGraphWebsiteCell.ts` is the single published-cell key. Key only by fields the website renders. Metadata such as fixture branch, reasoning effort, or setup time must not create a second visible copy of the same cell.
 
 ## Trace Audit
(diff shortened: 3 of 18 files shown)


benchmarks/graph/src/TtscBenchmarkGraphRunner.ts
```diff
@@ -32,9 +32,9 @@ import { ITtscBenchmarkGraphWebsiteAgentCell } from "./structures/ITtscBenchmark
  * Runs the complete graph-agent benchmark while keeping its executable a
  * location-only bootstrap.
  *
- * All harness paths derive from the supplied executable directory so moving
- * this implementation out of `src/executable/graph` does not change which child
- * entrypoints are launched.
+ * All harness paths derive from the supplied executable directory, so
```
samchon: Individual Self-Review finding, accepted. "All harness paths" is broader than the code: only the child entrypoints derive from `entrypointDirectory`. `repoRoot`, `workDir`, `websiteJson`, and `outDir` all come from `TtscBenchmarkConstant`, so a maintainer reading this could believe relocating the bootstrap also relocates the work root. Tightened in a follow-up commit on this branch.

samchon (commented):
**Individual Self-Review — `edf153aa1` (docs(benchmark): correct the references the package split left behind), resolving #1080.**

One read-only reviewer over that commit's parent-to-commit diff only.

**Clean on the commit's own contract.** All six corrections in #1080's tables were verified against the tree rather than against the issue text: `publish` is a real pnpm subcommand, `benchmarks/graph/src/executable/index.ts` is the agent bootstrap and `src/executable/graph/` does not exist, the package script is `audit`, and `benchmarks/README.md` does not exist while the flag and environment tables live in `benchmarks/performance/README.md`. The widened row (the Claude and Codex usage headers) is applied. The `benchmarkRoot` removal leaves `TtscBenchmarkConstant` still used in the file and leaves the comment above it describing the line it now precedes. Repository-wide sweeps for `src/executable/graph`, `experimental/benchmark`, `@ttsc/experimental-benchmark`, `graph:audit`, `benchmarks/README.md`, and any `pnpm --dir benchmarks/<pkg> <script>` without `run` return nothing outside this campaign's own git-ignored notes. The source contract (35 + 28 files), both `tsc --noEmit` lanes, and `prettier --check` pass.

**Two advisory candidates, both accepted and corrected in a follow-up commit on this branch:**

1. `benchmarks/graph/README.md:24` states a narrower reason than the truth. `audit` is also a real pnpm subcommand (`pnpm audit` reports dependency vulnerabilities), so a reader who takes the stated reason literally concludes only `publish` needs `run` and gets the wrong command with no error. That is the same silent-wrong-command failure the `publish` row was filed for, so it belongs to the class this commit seals rather than to a follow-up issue.
2. `benchmarks/graph/src/TtscBenchmarkGraphRunner.ts:35` says "All harness paths derive from the supplied executable directory", which is broader than the code: only the child entrypoints do, while the repository root, fixture work directory, report directory, and website JSON all come from `TtscBenchmarkConstant`. The clause predates the commit, but the commit rewrote that sentence for accuracy and left the one imprecise word in it.

**Explicitly rejected as a widening:** `benchmarks/evidence` spells `pnpm --filter @ttsc/benchmark-evidence <script>` without `run`. None of those script names collides with a pnpm subcommand, and that package is vendored and exempt under `.agents/skills/benchmark/SKILL.md`.

benchmarks/graph/README.md
```diff
@@ -123,7 +125,9 @@ Use `pnpm --dir benchmarks/graph run audit -- --compare=<before>,<after>` on aud
 
 | Variable | Default | Meaning |
 | --- | --- | --- |
-| `TTSC_GRAPH_BENCH_WORK` | `../graph-benchmark-work` | Fixture clone directory. It defaults outside the repo so a measured agent does not inherit ttsc's `CLAUDE.md` / `AGENTS.md` from a parent directory. |
+| `TTSC_GRAPH_BENCH_WORK` | `../graph-benchmark-work` | Fixture clone directory, checkouts only. It defaults outside the repo so a measured agent does not inherit ttsc's `CLAUDE.md` / `AGENTS.md` from a parent directory. Reports never follow it; they stay under `.work/`. |
```
samchon, line 130: Individual Self-Review finding, accepted. This row read "Per-child timeout for an agent sample", but the timed child is a whole cell: the runner passes `--runs=N` to one child that then runs `arms × runs` samples concurrently. The per-sample budget is `TTSC_CLAUDE_RUN_TIMEOUT_MS` (default 900000), which this table did not list. The row was also under-inclusive: the same timeout bounds the trace-audit child, every comparator index build, all fixture clone/fetch/reset/clean/install steps, and the `ttscgraph` build. Corrected, and the nine remaining undocumented variables added, in a follow-up commit on this branch.

samchon (commented):
**Individual Self-Review — `e6cad9e72` (fix(benchmark): anchor each run's report to the root that run resolved), resolving #1083.**

One read-only reviewer over that commit's parent-to-commit diff only.

**Verified correct.** Both defaults do what the commit claims, and `--out` and `TTSC_BENCH_OUT` keep their precedence exactly. `reportJson` follows `outputMarkdown`, `checkpointJson` already followed `workRoot`, `websiteJson` stays repository-anchored, and `TtscBenchmarkPerformanceRunner` creates the relocated directory. `loadPrevious` reads only the website JSON and the checkpoint, never the report, so moving the report loses no merge source. No consumer of either old path exists anywhere in workflows, skills, scripts, or the website. `workDir` is still used for exactly the fixture work, so the README's new "reports never follow it" claim holds package-wide. The relocated `runRoot` / `goCache` / `goTmp` stay run-specific and cleaned, land under a git-ignored directory, and match what the structural harness already does under `.work` in CI. Both `--list` lanes still exit before any directory is created, and nothing became unused.

**Three findings, all accepted and corrected in a follow-up commit on this branch:**

1. The new `TTSC_GRAPH_BENCH_TIMEOUT_MS` row misdescribed what the timeout bounds. The value is right, the scope was not: the timed child is a whole cell running `arms × --runs` samples, not one sample, and the same timeout also bounds the trace-audit child, `codegraph init`, `codebase-memory index_repository`, Serena project create and index, every fixture clone, fetch, reset, clean and install, and the `ttscgraph` build. An operator reading it as a per-sample budget would leave the 30-minute default on a five-run cell and lose the whole cell's work.
2. **The table was still not the list, which is the defect the commit set out to close.** Nine more variables the package honours were absent: `TTSC_BENCH_REQUIRE_QUIET`, `TTSC_BENCH_SKIP_LOAD_CHECK`, `TTSC_BENCH_CBM_MODE`, `CBM_LOG_LEVEL`, `TTSC_CLAUDE_RUN_TIMEOUT_MS`, `TTSC_CLAUDE_STARTUP_GRACE_MS`, `TTSC_BENCH_CONCURRENCY`, `CODEX_MCP_STARTUP_TIMEOUT_SEC`, and `CODEX_MCP_TOOL_TIMEOUT_SEC`. Each was re-verified at its read site before being documented. A `vscode` index-time run in particular was not operable from the README alone, because `TTSC_BENCH_CBM_MODE=fast` is the only mode that indexes it without the full mode's memory blowup.
3. `<WORK>` became the primary spelling of the performance Output table while being defined nowhere the reader passes on the way. The `TTSC_BENCH_WORK` row now defines it.

**One advisory rejected, with reason.** `.work/graph-index/` is a sibling of `.work/graph/` rather than nested under it. Nesting would put index-time's artifacts inside the directory `publish` folds by default (`.work/graph`), so the sibling keeps the two harnesses' outputs from meeting in the publisher's default source tree.

**Noted, not a finding.** `outDir` also governs index-time's `codebase-memory-cache`, which now lands on the repository volume rather than beside the fixture clones. It is removed in the `finally` block, and the agent runner already places its equivalent under `.work`, so this is consistency rather than a new class.

samchon (commented):
**Individual Self-Review — `a40ac7565` (fix(graph): stop attributing an assigned implementation to a declaration-file symbol), resolving #1082.**

One read-only reviewer over that commit's parent-to-commit diff only. It verified with `go test -overlay` against scratchpad copies and did not touch the tree.

**No correctness defect in the guards.** `target.External` is the right predicate: `declarationNode` prefers a non-declaration-file declaration with a body, so a workspace symbol merged with a `.d.ts` still resolves `External=false` and keeps its assigned-implementation attribution (`value_call_edges_attribute_function_assignments_to_members_test.go` still passes). No legitimate `from` can be external: all five producers were enumerated, external node ids key on `.d.ts` / `node_modules` paths that are disjoint from workspace container paths, and instrumenting the guard to panic across the whole Go corpus never tripped it. A `.d.ts` can never enter the incremental `selected` set, so `lookupNode` cannot be blinded by it. Not materializing the node is safe: `installExternalShard` emits an external node only at reference count > 0, and the dropped node had none. **Both new tests were confirmed to fail before the fix**, with the exact messages `edge leaves the external boundary: …/globals.d.ts#patched:variable -> …/main.ts#helper:function (value-call)` and `ttscgraph: edge source node is absent from shard facts: src/globals.d.ts#patched:variable`.

**Three findings accepted, corrected in `9ada513d2`:**

1. **The `addEdgeAt` guard had zero coverage.** Three independent runs proved it: removing either guard alone leaves both new tests green, and the panic instrumentation never fires. Nothing pinned what the commit message called "the invariant that outlives any single producer".
2. **Worse, its predicate could not deliver that invariant.** The test was node existence plus an `External` flag, not the classification of the file the id names. A producer that mints an id directly from a declaration-file path passes straight through — and that is how `heritageEdges` already builds a `from`, safe today only because `addEdges` filters `IsWorkspaceSourceFile`. In the one future shape the guard was written for, #1082's deferred "reclassify a project-owned `.d.ts`" change, `Build` still records no declaration-file node, `lookupNode` answers `!ok`, and the guard stays silent. On the incremental lane it is weakest in exactly the shape it exists to catch, because an external node whose only relationship is an outgoing edge has zero incoming references and is pruned from `baseNodes`.

   Both point one way, so the guard is removed rather than repaired. A silent `return` would also have degraded a loud, diagnosable refusal — shard assembly already names the offending node — into a missing fact. The root-cause guard in `assignedFunctionTarget` stays, and it is the one both tests discriminate.
3. **Coverage breadth.** `IsWorkspaceSourceFile` refuses a file for two disjoint reasons and only the declaration-file branch was pinned. A second test now covers a symbol declared in a dependency's raw `node_modules` source.

**Recorded, no action.** The fabricated evidence spans have no direct assertion; they disappear as a consequence of removing the edge. `website/src/content/docs/graph/` documents no assigned-implementation attribution and `viewer.mdx` is already consistent, so no documentation is owed.

samchon (commented):
**Individual Self-Review — `cd94c380a` (fix(ttsx): resolve a dependency's rootDir the way its served sources are resolved), resolving #1073.**

One read-only reviewer over that commit's parent-to-commit diff only.

**No critical finding.** The two questions that would have been critical are both clean. For a path that exists, `resolvePhysicalPath(x)` reduces to `fs.realpathSync.native(x)` normalized, which is exactly what `realPath()` produces, so the two sides of `path.relative(projectRoot, sourceFile)` do agree; using `.path` rather than `.key` is the correct choice, because the key lowercases the win32 drive root and that would reach the returned emit path. The missing-suffix fold only fires for a path that does not exist, and a `rootDir` that does not exist has no sources under it either way. The new test was traced end to end and does fail before the commit: the marker carries `<dep>/src` while `realpathSync.native` answers `<dep>/sources`, which fails on Linux from the symlink alone and twice over on Windows (8.3) and macOS (`/var` to `/private/var`). The fixture type-checks with no `@types/node`. The consequence surface is closed: `DependencyCacheMeta.rootDir` has exactly one writer and one reader, `BuiltProject.rootDir` one producer pair and one consumer, and no sibling lane carries the same mixed pair (`runSingleFileEmit` uses plain `fs.realpathSync` on both sides, which is a consistent pair).

**Findings accepted, corrected in `9ada513d2`:**

1. **`readDependencyCache` could newly throw.** `throwOnRealpathError: false` silences a failed realpath, but the case-sensitivity probe rethrows a non-ENOENT `lstat` failure of its own, and no caller guards it, only `serveDependencyEmit`'s outer catch, which silently downgrades the dependency to the orphan type-strip lane. That is a new failure mode on a reader whose whole contract is to answer `null`. `resolvePhysicalPath` is now total, falling back to the unresolved spelling, which is what both callers had before the pass.
2. **The read side had no test.** #1073's second acceptance item is exactly the `readDependencyCache` change, and nothing pinned it; no test in the suite reads `readDependencyCache(...).rootDir`. A deterministic, spawn-free case now seeds a marker whose root is a symlinked directory and asserts the read answers the physical path.
3. **The undeclared-`rootDir` branch was claimed fixed but untested.** It diverges only on Windows, where every `TestProject` root inherits the runner's 8.3 temp root. The e2e fixture gains a second dependency that declares no `rootDir`, so the `project.root` branch is exercised on that lane and the assertion loop covers both markers.
4. **Three inaccurate rationale sentences.** The cost claim ignored the missing-path branch, which costs a directory read and can cost an `fsutil` query on Windows, and that is the branch both existing cache regressions actually take, since they write `rootDir: "/root"`. The reuse claim was wrong in kind: `rootDir` never gated reuse, so a stale-spelled marker was already being reused and merely served every file through the whole-tree stem rescan. And the retained `path.isAbsolute` branch is a mirror of the entry lane, not a live guard, since `readProjectConfig` absolutizes every path option. All three now say what the code does.

**Rejected, with reason.** A symlink *below* `rootDir` still drops the exact mirror, but that is not the same defect: the real source genuinely sits outside `rootDir`, tsgo strips the common source directory instead, and `resolveEmittedJavaScript`'s stem matcher is the documented answer for that shifted-prefix case. A non-absolute `meta.rootDir` resolving against cwd is a property of a malformed marker only; every marker this code writes is absolute.

samchon (commented):
**Individual Self-Review — `3476e255e` (fix(benchmark): resolve a relative publish source like every sibling, and refuse an empty one), resolving #1084.**

One read-only reviewer over that commit's parent-to-commit diff only. It found a blocking regression that this branch's next commit, `9ada513d2`, fixes.

**Blocking, accepted and fixed.** Dropping `resolveReportPath`'s repository-root probe broke every runner-produced suite report, including the absolute `--from` case the commit message claimed was unaffected. The commit's premise, "suites this repository writes record absolute paths", is true of `TtscBenchmarkGraphSuite` but that is the wrong producer: `foldSuite` reads the **runner's** `<outDir>/report.json`, and `TtscBenchmarkGraphRunner` recorded each cell's report as `path.relative(repoRoot, copyPath)`. Verified against the code: the runner writes `cells[].project`, `.branch`, `.tool`, the exact fields `foldSuite` reads, which `TtscBenchmarkGraphSuite` never writes. With the probe gone, the base joined the repository-relative path onto `sourceDir` and the publish aborted with `missing suite cell report` before folding anything. The new JSDoc's reachability claim was false in the same way: the relative branch was the only branch a runner report took.

The fix is the single base the commit's own reasoning wanted, not a restored probe. The runner now records the field absolute, the way the suite runner already does, so the publisher, the trace auditor, and the suite all need no base at all, and a run directory named by `--out` is readable from wherever the publish is typed. The sibling `log:` field stays repository-relative on purpose: it is provenance a human reads and nothing resolves it.

**Second finding, accepted and fixed.** `folded` measured file presence, not contribution. Confirmed there was no false failure in the other direction: every artifact the publisher reads sets it, including a nested structural report and an agent report whose samples are all filtered, and the `||` operands are ordered so no fold is short-circuited away. The gap ran the other way: a suite report with `cells: []`, a suite whose every cell `continue`s, and an agent report whose samples are all rejected by `validMeasuredSample` all counted as contributions, so `publish --from <dir> --reset` against a run with zero usable cells still wrote an empty agent block and exited zero. That is precisely the failure #1084 names. `foldSuite` now returns its upsert count and `foldAgent` a boolean, and the error text says "contributed no measurement" rather than "holds no report".

**Third finding, accepted.** The refusal also applies to the defaulted `.work/graph`, which the commit message and the file's usage header both denied. Not a CI regression, `publish` is invoked by no workflow and by no documented flow without `--from`, but the record was wrong; the usage header, the README, and the skill now say it.

**Fourth finding, accepted and already fixed.** The README sentence used two em-dashes, against `.agents/skills/documentation/SKILL.md` Voice.

**Fifth finding, accepted and fixed.** The same `cell.report` field had a third reader disagreeing: `TtscBenchmarkGraphTraceAuditor.rawReportsFromSuiteReport` resolved it against process cwd and then silently dropped whatever failed to exist. It worked only because the runner spawns the auditor with the repository root as cwd; a human running the documented `pnpm --dir benchmarks/graph run audit -- --report ...` audited zero cells with no error. It now resolves against the report that recorded it, the same base as the publisher and the suite.

**Rejected, with reason.** With several `--from` flags the throw names only the first empty directory and later ones are never folded. Failing fast before any write is the safe behaviour, and naming the first offender is enough to act on.

**Clean.** The throw is strictly before `mkdirSync`/`writeFileSync`, so the committed dashboard is untouched on failure under `--reset` and `--dry-run` alike; `--dry-run` still writes nothing. Duplicate `--from` detection and `foldSuite`'s `seenReports` check are unaffected by the cwd base. Nothing became unused, and the added JSDoc is on private functions, which the benchmark source contract neither requires nor forbids.

samchon (commented):
**Individual Self-Review — `9ada513d2` (fix: adjudicated Individual Self-Review results for the cycle).**

One read-only reviewer over that commit's parent-to-commit diff only. It ran `go vet`, the whole `internal/graph` suite, the changed `cmd/ttscgraph` test, `tsc --noEmit` on `benchmarks/graph`, the benchmark source contract, and `prettier --check`, and reproduced each claim it makes.

**Verified sound.** Both new Go tests genuinely pin the fix: rebuilding through `go test -overlay` with only `target.External` removed makes both fail, and the `node_modules` failure names `node_modules/dep/globals.ts`, so that fixture really does travel the `node_modules` branch of `IsWorkspaceSourceFile` rather than the declaration-file branch. Removing the `addEdgeAt` guard is safe: every `from` producer was re-enumerated and only `assignedFunctionTarget` can yield an external id. `resolvePhysicalPath`'s totality is not theoretical — the identity resolver rethrows a non-`ENOENT` `lstat` failure that `throwOnRealpathError: false` does not cover — and the catch placement restores exactly the pre-change value for both callers. On a Windows host the reviewer confirmed an unprivileged junction succeeds, that `realpathSync.native(link) !== link`, and that the identity resolver answers the real path, so the marker-read test discriminates. `roots.length === 2` holds: `dep2` has its own tsconfig, the `.tmp` marker temp names and the `<key>.lock` directory are both excluded by the `.json` filter. The throw precedes the write under `--reset` and `--dry-run` alike.

**Findings accepted, corrected in `73769e156`:**

1. **The contribution count still over-counted.** `upsertAgentCell` returns early on the skip-thinner branch without touching the document, and both callers counted that as an upsert, so a directory whose every cell is thinner than what is published passed the refusal. It now reports whether the document changed. The error direction was permissive, so no publish ever failed wrongly, but this is the accounting the commit set out to make honest.
2. **The auditor still dropped an unresolvable cell report silently.** Only the base was fixed; `fs.existsSync` still discarded a cell with no diagnostic and no non-zero exit, so any future mismatch reproduces the same zero-cell audit. It now fails, naming the cell and the report that recorded it.
3. **The auditor's own writer disagreed with its new reader on the same field.** `cells[].report` was written against the current directory while being read against the recording report, and an audit JSON is itself accepted as a `--report` input. It now records the path absolute, like the runner.
4. **The README completeness claim was false.** "Every variable either runner reads" is both too narrow — the agent harnesses and comparators read most of the table — and untrue, since `PATH` is read and unlisted.
5. **The commit message did not describe all of its scope**: it did not mention that this branch rewrote the graph environment table, added the `pnpm audit` disambiguation, or defined `<WORK>` in the performance README. Recorded in the follow-up message.

**Accepted as deliberate, with reason.** A `report.json` written before the base change now fails in the publisher rather than resolving through a repository-root probe. Reports are per-run and git-ignored, and restoring a probe would restore the second base this cycle removed; with the auditor's silent drop closed, both readers now say so instead of one failing loudly and the other quietly.

**Correction to the earlier record.** `sanitizeSamples` throws on an invalid sample rather than yielding empty ones, so the case the refusal guards is a report whose sample arrays are empty, not one whose every sample was rejected.

**Noted, no action.** `driver.LoadProgram` returns CLI and tsconfig-parse diagnostics only, so `if len(diags) != 0` in the graph tests does not actually assert the fixture type-checks. It is a limitation shared with every sibling test in that directory, and the reviewer proved both new tests fail before the fix and pass after by running them.

samchon (commented):
**Individual Self-Review — `73769e156` (fix(benchmark): count a real upsert, refuse an unreadable cell, complete the env table).**

One read-only reviewer over that commit's parent-to-commit diff only. It ran every validation gate locally: `tsc --noEmit` on `benchmarks/graph`, the benchmark source contract, `prettier --check`, and `audit -- --self-test`, all passing, and reproduced each finding with `--dry-run` against scratch fixtures.

**It found that I introduced a regression, now reverted in `117a8dfd7`.**

Making `upsertAgentCell` report whether it wrote turned a designed normal path into a fatal one. Declining to replace a published cell with a thinner sample set is the publisher's own rule, stated in this file's header and announced with a `console.warn`; a directory whose cells are all thinner than what is published then counted as contributing nothing and aborted the publish. Reproduced: `publish --from A --from B` fails while `publish --from B --from A` succeeds on the same inputs, producing an identical document, so the exit status depended on argument order. It also broke the N=1 smoke flow the README recommends, whose whole point is a thinner run, and could leave a parallel-sweep loop partially published and failed.

The refusal exists to catch a directory that measured nothing. A cell the publisher measured and then kept out on purpose measured something. `foldSuite` counts measured cells again, which is what the error message already said, and #1084's acceptance asks for nothing beyond that. My earlier claim that "the error direction was permissive, so no publish ever failed wrongly" described the state before that commit and did not account for the failure it newly made reachable.

**Also accepted and fixed:**

- The auditor's seal was half done. A cell whose `report` field is absent or not a string was still filtered away three lines above the throw that claimed no listed cell is ever dropped, so `{"cells":[{"harness":"codex","promptId":"x"}]}` still produced a zero-cell audit at exit 0. Both cases fail now.
- The comment on the auditor's writer understated the old bug: recording against the current directory broke whenever the run directory was not the directory the audit was typed in, which is every documented invocation, not only a read from elsewhere.
- The self-test's synthetic cell recorded the superseded relative spelling, so the only automated check of that file modelled a shape the file no longer produces. It records the absolute spelling now.
- `foldSuite`'s empty-sample cell vanished into the difference between two counts; it is named.
- The mixed bases inside one audit JSON now say why `traceDir` and `runsDetail[].file` stay relative, the way the runner's `log` field already does.
- The README's completeness sentence repeated the one after it.

**Rejected, with reason.** `--dir` pointed at an empty directory still audits zero cells at exit 0. That entry point scans a directory for reports rather than reading a manifest that promises cells, so nothing was dropped; a manifest listing a cell it cannot open is a broken promise, an empty directory is a true answer.

**Corrections to the record.** The throw names the report path and the file that recorded it, not the cell's repo/promptId identity. `upsertAgentCell` writing `{...existing, ...cell}` on an equal-count re-publish means "accepted", not "the document changed" — the reason the accounting is now phrased as measurement rather than mutation.

samchon (commented):
**Individual Self-Review — `117a8dfd7` (fix(benchmark): stop a deliberate thinner-cell skip from failing a publish) and `36bc196a2` (finish the two classes this cycle left one link short).**

Two read-only reviewers, one per commit, each over its own parent-to-commit diff. Both executed every check against scratchpad fixtures with `--dry-run`; `website/public/benchmark/graph.json` was verified unchanged by hash.

**The regression is gone, proven by execution.** With a published cell as incumbent, a `--from` directory holding only a thinner cell exits 0, printing `skip thinner agent cell` then the suite line. Both orders of two `--from` directories succeed and produce the same document. An N=1 smoke directory publishes.

**#1084's guarantee is intact, also by execution.** A directory with no report, a suite `report.json` whose only cell has empty sample arrays, and a directory holding only `codex-trace-audit.json` each throw `graph benchmark source directory contributed no measurement`, before the write, identically with and without `--reset`.

**The two new `auditCell` throws cannot be reached by a real run.** Both harnesses `mkdirSync` the trace directory before any run and write `${arm}-run-${n}.stream.jsonl` unconditionally right after each spawn, before validation; `runs` is at least one and at least one arm is required, so `arms × runs >= 1`. A retry overwrites under the same name. No `--keep-*` option and no cleanup step touches the trace directory. The report that names a cell is written only after the loop, so a harness that dies early leaves no cell to audit. The runner's automatic post-run audit derives each cell's trace directory from a distinct per-cell report and cannot newly fail a completed run. Measured old against new: a missing trace directory was already exit 1 on a bare ENOENT, so only the empty-directory case changes behaviour, and its old behaviour was `runs: 0` at exit zero with a median of zero folded into the suite summary.

**Findings accepted across both rounds:**

1. The empty-sample skip was named in `foldSuite` and still silent in `foldAgent`, so a directory holding one measured report and one that measured nothing published the first and lost the second without a word. Same condition, same warn - and moved up to `foldAgentFile`, which has the path, because the harness name alone does not say which `--from` directory lost its cell.
2. The auditor's chain stopped one link short of its own comment. A cell whose trace directory is gone died on a bare ENOENT naming neither the cell nor the recording report; a cell whose trace directory is empty audited as `runs: 0` at exit zero. Both fail now, named, and the message says it looked for `*.stream.jsonl` - which matters because a Claude trace directory legitimately holds `.stderr.log` files and `.home` directories beside them.
3. `traceDir` still resolved against the current directory, the one base this cycle spent two commits removing, three lines below the `cells[].report` that had just been re-based onto the recording report. Dormant, because both harnesses record it absolute, but the new throw is what surfaced it - naming a path the report never contained.
4. The self-test's synthetic cell is a hand-copied duplicate of the real cell literal, which is how the recorded spelling drifted apart to begin with. The property it drifted on is now pinned by an assertion, verified to fail when the old relative spelling is restored.
5. Wording: a tense that disagreed with its own sibling paragraph; `upsertAgentCell`'s decline left undocumented at its definition after the return value went away, and described as whole-set when the test is per arm; a throw calling a cold-index report a suite report while refusing it as one; `(N of M cell(s))` not saying which of measured or written it counts; "a published number" where these two files use "published" for the served website document; and a dangling clause.

**Deferred to the next cycle, with reasons in the campaign ledger.** The auditor's own zero-cell audit still exits zero - #1084's silent-success shape one entrypoint over - because refusing `cells.length === 0` needs decisions about `--compare` with an empty side and `--dir` on a Claude-only run directory. And `upsertAgentCell` merges rather than replaces, so a key the stored cell carries and the new one omits survives from an older measurement; pre-existing, outside every diff this cycle touched.

**Corrections to my own earlier record.** "The error direction was permissive, so no publish ever failed wrongly" described the state before the commit it was defending and did not account for the failure that commit newly made reachable. A finding rated "no false failure, low severity" is a signal that the permissive direction was chosen on purpose; changing it needs the same evidence any behaviour change needs.

samchon (commented):
**Individual Self-Review — `c2832faef` (fix(benchmark): name the report that lost its cell, and resolve traceDir like every sibling).**

One read-only reviewer over that commit's parent-to-commit diff only, executing every check against scratchpad fixtures.

**The `traceDir` base change is a no-op for every real run, verified at the producers.** Both harnesses resolve `traceDir` before recording it, including an operator-supplied relative `--trace-dir`, and both write it into the report unconditionally; the runner passes the parsed object through verbatim into its per-cell copy, and its `outDir` is itself resolved. The only other `RawCellReport` constructed anywhere is the self-test's, which is absolute and reaches `auditSyntheticCell` rather than `auditCell`. Both `--compare` entry points funnel through `buildAudit` to `auditCell`, so the base is uniform. Measured: an absolute `traceDir` audits identically; a relative one now resolves against the recording report, where the old code resolved it against `benchmarks/graph` and died naming a path the report never contained.

**The warn move changes no control flow.** `foldAgent` has exactly one `return false` and one call site, so the trigger set is identical. Confirmed by `--dry-run`: an empty-sample directory warns naming the file and is still refused; a thick directory followed by a thin one with the same key still prints `skip thinner agent cell` and is still accepted; a directory with no agent report file produces no warn, which is right, since an absent file is not a report that measured nothing.

**Gates.** `pnpm --dir benchmarks/graph run check`, the benchmark source contract (35 + 28 files), `audit -- --self-test`, and `prettier --check` all pass. No test, document, or CI harness substring-asserts any changed string.

**Message accuracy verified, not assumed.** `median([])` returns `0` and `summarizeSuite` folds every cell's median unfiltered, so "folding a median of zero into the suite summary" is exact; "recorded" rather than "published" is right, since the audit writes to `--out` and never to the website document. The `upsertAgentCell` JSDoc matches the per-arm test in the code.

**One defect, fixed in `4da7e44cc`.** The comment locator read "three functions up"; the base it names is in `rawReportsFromSuiteReport`, immediately above `auditCell`. A reader following the count lands past `buildAudit` in unrelated code.

**Left to separate issues, recorded in the campaign ledger.** `TtscBenchmarkGraphTraceAnalyzer` reads `report.traceDir` with no base at all, the one remaining sibling reader off the recording-report base, dormant for the same reason and in a file outside this cycle's surface. A Windows driveless-rooted spelling would differ between `auditCell`'s bare `path.resolve(dir, x)` and its siblings' `isAbsolute` guard, unreachable from any producer.

**Rejected for this cycle, with a reason rather than deferred.** `foldAgent` logs its `agent: … (N graph runs)` line even when `upsertAgentCell` declined the cell, so a decline prints both the skip warning and a line that reads as a successful fold. Gating that log on the upsert outcome means restoring the return value whose propagation produced this cycle's one regression, and the decline is already announced on its own line immediately before. Reworking the pair belongs to its own change.

samchon (commented):
**Overall Self-Review — final clean round over `origin/master...4da7e44cc`.**

Solo, whole surface, fresh start. Earlier rounds were each invalidated by the correction that followed them; this one covers the settled head.

Read in full this round: every changed file and hunk across all 18 files. `packages/ttsc/internal/graph/edges.go` and its three Go tests, `packages/ttsc/src/launcher/internal/runtimeHooks.ts` and its two ttsx tests, `TtscBenchmarkGraphPublisher.ts`, `TtscBenchmarkGraphTraceAuditor.ts`, `TtscBenchmarkGraphRunner.ts`, `TtscBenchmarkGraphIndexTime.ts`, both agent usage headers, `TtscBenchmarkPerformanceExecutable.ts`, both package READMEs, and both benchmark skill documents. Consequence surface re-checked: the writer/reader pairs for every recorded path field, the `--reset` and `--dry-run` write ordering, the incremental shard path for the graph change, Windows and POSIX behaviour for both path-identity changes, and the `dependency-cache` helper's re-exports that the new marker test relies on.

**Nothing to improve.**

**Verification that could not run here.** Local build and test are closed in this repository. The two benchmark packages have no automated test lane beyond their type check and the source contract, so #1080, #1083, and #1084 carry no new test; their resolved paths were read from the source and both READMEs reconciled against them, and the reviewers exercised the publisher and the auditor with `--dry-run` against scratch fixtures. Everything else is proven by the 27 green checks on this exact head.

**Gate state.** All 27 required checks green on `4da7e44cc`. Nine Individual Self-Reviews recorded, one per pushed issue-implementation commit. This round is clean on the same immutable head. Closing keywords reconciled: five `Close #n` lines, one per accepted issue, no revert in the range.

**Recorded for the next cycle's discovery, not fixed here.** The trace auditor's own top-level zero-cell audit still exits zero, which is #1084's silent-success shape one entrypoint over; `TtscBenchmarkGraphTraceAnalyzer` reads `traceDir` with no base at all; and `upsertAgentCell` merges rather than replaces, so a key the stored cell carries and the new one omits survives an upsert. Each needs its own evidence and its own consequence surface.

Issue #1101, opened 2026-08-06
The evidence vendor parity script cannot run: its upstream checkout path does not exist


samchon:
## Problem

`experimental/evidence-vendor/parity.cjs` is the script that proves every vendored `@ttsc/evidence` file either matches upstream byte for byte or is a declared adaptation. It cannot run, because it looks for the upstream checkout at a path that does not exist.

The consequence is not a broken script. It is that the one mechanical check on a hand-vendored tree has been unavailable since the tree was vendored, so drift between this repository and `samchon/lint-plugin-evidence` is currently invisible.

## Evidence

```js
// experimental/evidence-vendor/parity.cjs:21
const UP = "D:/github/samchon/lint-plugin-evidence";
```

```
$ ls -d D:/github/samchon/lint-plugin-evidence
ls: cannot access 'D:/github/samchon/lint-plugin-evidence': No such file or directory
```

The checkout is `D:/github/samchon/evidence`. Its `origin` is `https://github.com/samchon/lint-plugin-evidence.git`, so the repository is right and only the directory name differs.

Lines 27-33 resolve `BRANCH_REF` through `execFileSync("git", ["-C", UP, "rev-parse", ...])` at module scope, so the script throws on that line before comparing a single file. Nothing downstream is reached.

`experimental/evidence-vendor/README.md` states the same path in prose ("Upstream is expected at `D:/github/samchon/lint-plugin-evidence`") and in the `robocopy` recipe's `<upstream>` placeholder, so a reader following the document copies from a directory that is not there either.

`BRANCH_REF` is `origin/campaign-luna-0.6.0-cont`, and that ref does still exist upstream, so it is not the fault here.

## Consequence surface

- The parity script is the acceptance gate for any change to a vendored file. A repository-only module added under `benchmarks/evidence/src` needs an `EXCEPTIONS` entry, and the script is what proves the entry is complete.
- `audit.cjs` and `readapt.cjs` operate on the vendored trees in place and do not read the upstream path, so they are unaffected. Only `parity.cjs` and the README hold it.
- The path is machine-local by design. Making it overridable, rather than replacing one hardcoded absolute path with another, is what keeps the script usable when the checkout moves again.
- The scripts are branch-local tooling and part of no build, lane, or published package, so nothing in CI covers this.

## Approach

[...]

Issue #1098, opened 2026-08-06
Track the published evidence benchmark cohort in this repository


samchon:
## Problem

`benchmarks/evidence` is vendored from `samchon/lint-plugin-evidence`, and upstream tracks the published GPT-5.6-Luna cohort under `benchmark/aggregate`. This repository tracks none of it.

`benchmarks/evidence/aggregate/` does not exist on `master`. Every document that describes it (`benchmarks/evidence/README.md`, `.agents/skills/benchmark/evidence/measurement/aggregate.md`, `website/src/content/docs/benchmark/evidence.mdx`) says the aggregate is the only tracked measurement and the only place a published figure may come from, and then there is no aggregate. A reader who clones this repository can read how the benchmark works and cannot see what it measured, and nothing here can be checked against the numbers the package's own benchmark produced.

## Evidence

Upstream tracks fourteen artifacts:

```
$ git -C D:/github/samchon/evidence ls-files benchmark/aggregate
benchmark/aggregate/cells/gpt-5.6-luna/erp/arms.svg
benchmark/aggregate/cells/gpt-5.6-luna/erp/evidence.json
benchmark/aggregate/cells/gpt-5.6-luna/erp/plain.json
... (the same three per subject for reddit, shopping, todo)
benchmark/aggregate/summary.json
benchmark/aggregate/summary.svg
```

Nine of them are JSON: `summary.json` plus one file per cell under `cells/<model>/<subject>/<arm>.json`.

`summary.json` is exactly the `ITtscEvidenceBenchmarkReport` this repository already declares. Its eight cells carry the field set `ITtscEvidenceBenchmarkReportCell` declares, key for key:

```
engine, subject, arm, runId, benchmarkRevision, model, effort, status, stage,
launchedAt, tokens, tokenUsage, inspection, apiCost, suspendedMs, suspensions,
workElapsedMs, worktree, reviewVerdicts, stages
```

Cohort shape:

| Property | Value |
| --- | --- |
| `generatedAt` | `2026-08-06T11:33:32.869Z` |
| Model | `gpt-5.6-luna`, effort `high`, engine `codex` |
| Subjects | `todo`, `reddit`, `shopping`, `erp` |
| Arms | `plain`, `evidence`, one cell each, eight total |
| Status | every cell `completed` |
| `apiCost` | present on all eight |
[...]

Issue #1100, opened 2026-08-06
The evidence benchmark page publishes no result


samchon:
## Problem

`https://ttsc.dev/docs/benchmark/evidence` documents how the evidence benchmark is run and publishes no result from it.

The other two benchmark tracks do the opposite. `benchmark/performance.mdx` and `benchmark/graph.mdx` each import React components that read a committed JSON report and draw it, so a reader sees the measurement. The evidence page is 170 lines of method with a `Publish` section pointing at files nobody outside the repository can open, and it closes by saying "It documents the measurement, not a result."

Once the cohort is tracked here (#1098) and the charts render from it (#1099), the page can show what the benchmark found, in the same shape as the other two tracks.

## Evidence

- `website/src/content/docs/benchmark/evidence.mdx` contains no component import and no figure.
- `website/src/content/docs/benchmark/graph.mdx:1-3` imports `TtscWebsiteBenchmarkGraphCommon`, `TtscWebsiteBenchmarkGraphDedicated`, and `TtscWebsiteBenchmarkGraphTime` and places them between prose sections.
- `website/src/components/benchmark/graph/useTtscWebsiteBenchmarkGraphData.ts:13` fetches `/benchmark/graph.json` once, memoizes the promise, and exposes `{report, loading, error}`. `TtscWebsiteBenchmarkGraphData.ts` reshapes the flat cell list into groups. `TtscWebsiteBenchmarkGraphUi.tsx` holds the shared panel, eyebrow, section header, and accent colours.
- `website/src/structures/ITtscWebsiteBenchmarkGraph.ts` and `ITtscWebsiteBenchmark.ts` declare the shapes of those committed reports.
- `website/package.json` `prebuild` already runs a benchmark asset step (`build:benchmark:png`), so a data step has a place to sit.
- `website/src/content/docs/benchmark/_meta.ts` already lists `evidence: "Evidence Graph"`.

The tracked aggregate carries everything a chart needs: per cell `tokens`, `tokenUsage` (total, input, cached input, cache write, output, reasoning), `inspection`, `apiCost.amountUsd`, `workElapsedMs`, `suspendedMs`, `worktree` (files, additions, deletions), `reviewVerdicts`, and `stages[]` with per-stage tokens and elapsed time.

## Consequence surface

- `.github/workflows/website.yml` triggers on `website/**` and runs the full `pnpm run build`, which ends in `next build`, Pagefind, and the sitemap. A component that throws on missing data fails that build.
[...]

Issue #1099, opened 2026-08-06
Evidence benchmark charts render only from the ignored run tree, and the chart set is two generations behind the aggregate


samchon:
## Problem

The evidence benchmark's charts can only be produced by an operator who still holds the raw run tree, and the chart set they produce is two generations behind the aggregate it describes.

`writeEvidenceBenchmarkReport` collects from `benchmarks/evidence/output/`, which is gitignored and exists on exactly one machine. In any clone the same command writes nothing usable and, worse, is destructive: it deletes `aggregate/cells` and every top-level `*.svg` before writing, so running it in a checkout with no run tree empties the tracked aggregate. Nobody but the machine that ran the cohort can regenerate, verify, or re-style a chart from the data this repository tracks.

The renderer itself carries three defects and one dead configuration block.

### The input is unreachable

`benchmarks/evidence/src/EvidenceBenchmarkReport.ts:22` calls `collectEvidenceBenchmarkReport(options.repository, ...)`, which walks `benchmarks/evidence/output/<subject>/<engine>/<arm>/runs/`. `.gitignore` carries `/benchmarks/evidence/output/`. Lines 30-38 then remove `aggregate/cells` recursively and every top-level `*.svg` before any write, unconditionally, including when the collection returned no cell.

### The stacked bar under-fills its own total

`cell.tokens` and `cell.workElapsedMs` include `cell.inspection`, the cost of judging a Plain review, which belongs to no stage. `phaseValues` sums only `cell.stages`, so the five segments always total less than the number printed at the end of the same row.

`:237` compounds it: `maximum` is taken from `metric.cellValue` alone. A stage record can also sum above the cell total, because the total excludes idleness the stage records keep, and such a bar runs past its own track.

Against the tracked cohort this is visible rather than theoretical: every cell has a non-zero `inspection.tokenUsage.totalTokens` (Todo Plain alone is 5,076,126 tokens and 677,418 ms).

### The model label is a string nothing accepts

`:488-491` rewrites `gpt-5.6-luna` to `GPT-5.6-Luna`. The runner, the Codex session, and the price list all take `gpt-5.6-luna`, and the engine that ran it is not in the label at all. A reader trying to reproduce a figure cannot copy what the chart prints.

### Dead chart configuration

[...]

PR #1102, opened 2026-08-06
feat(benchmark-evidence): publish the evidence benchmark cohort, its charts, and its website page

samchon:
Solo issue campaign, cycle 1. This pull request owns the complete accepted cycle: the published cohort of the `@ttsc/evidence` benchmark becomes tracked data in this repository, its charts are drawn from that data instead of from a run tree that exists on one machine, the website publishes the result, and the vendor parity script that proves the port is honest is made runnable.

Campaign scope: the evidence benchmark **result publication surface** only. The runner, workspace preparation, instructions, requirements, and templates are out of scope, and so is the live campaign in #1079.

## Cycle issues

| Issue | Subject | Predecessor |
| --- | --- | --- |
| #1098 | Track the published cohort's nine JSON artifacts under `benchmarks/evidence/aggregate` | none |
| #1101 | Make `experimental/evidence-vendor/parity.cjs` runnable against the upstream checkout that exists | none |
| #1099 | Render the charts from the tracked aggregate, and port upstream's current chart set | #1098 |
| #1100 | Publish the result on `/docs/benchmark/evidence` | #1098, shares chart vocabulary with #1099 |

Implementation order follows that DAG. Verification is pending; nothing is implemented at open.

## Discovery gate

Two complete full-scope discovery rounds ran against baseline `8d03c3450 chore: release v0.25.0`, which is also `origin/master`. Round 1 accepted seven candidates and published them as the four issues above. Round 2 produced no meaningful in-scope candidate, which closed the gate and opened this cycle.

## Known limitation carried deliberately

Upstream draws a coverage block from a per-subject edge table hardcoded in its renderer. This port makes that block read coverage data from the aggregate and omit itself when there is none, rather than carrying the table. Upstream retained exact populations for `shopping` and `erp` only, so re-typing its rounded `todo` and `reddit` rates would publish a population nothing here measured. #1079 introduces the `coverage.json` artifact that fills the block; the reasoning is on #1099 and in #1088.

## Coordination

#1079 (`campaign/evidence-luna-0.24.0`) is the live campaign for the same cohort. It commits no aggregate artifact and does not touch `EvidenceBenchmarkReport.ts`, so there is no file overlap with this branch.


samchon (commented):
**Individual Self-Review — `57a4ce7f5` "track the published GPT-5.6-Luna cohort"** (closes #1098)

A read-only reviewer read the parent-to-commit diff. Every acceptance check passed: the nine files equal upstream byte for byte after CRLF normalization, each cell file is deep-equal to its `summary.json` entry including key order, the report matches `ITtscEvidenceBenchmarkReport` with no missing or excess property at any level, every one of the sixteen distinct stage names classifies under `stagePhase` with `REVIEW_SUPPLEMENT_LIMIT = 4`, and all nine are byte-identical to `JSON.stringify(parsed, null, 2) + "\n"`. Nothing unwanted is in the diff.

Adjudication of the candidates:

| Finding | Disposition |
| --- | --- |
| The aggregate is outside `parity.cjs`'s `TREES`, and nothing records why | **Accepted.** It is correct to leave it out, since this repository publishes its own measurement and a comparison would report a residual on every regeneration. The reason was unrecorded; `004ae4c18` writes it into the vendor README. |
| The commit message's "a regeneration is a no-op rather than a diff" is not literally true, because `generatedAt` is restamped | **Accepted as a correction to the record.** The verified claim is narrower: the serialization form and key order are exactly what the writer emits. Noted here rather than rewritten into history. |
| The cohort mixes two `benchmarkRevision` values: `reddit/evidence` and `shopping/evidence` ran at `0f465f52c`, the other six at `cdbda4211` | **Accepted.** Faithful to upstream, so not a copy defect, but a publication caveat. It will be stated where the figures are published rather than left for a reader to derive from run ids. |
| `erp/plain` carries a replayed stage sequence with repeated stage names, no review verdicts, and zero inspection | **Accepted as a data fact.** The charts group stages into phases and sum them, so a repeated name is attributed correctly rather than double-counted. Recorded because a consumer keying by stage name would not be. |
| Stage sums do not reconcile to cell totals uniformly: `todo/plain` leaves a residual close to but not equal to `inspection`, `shopping/plain`'s stage sum equals its total while carrying 16.2M inspection tokens, and two cells' stage elapsed sums exceed `workElapsedMs` | **Accepted, and it changes #1099.** The unattributed remainder is not the inspection figure, so naming that segment after inspection would assert something the data does not support. The acceptance criterion on #1099 that segments plus remainder equal the printed value is not uniformly satisfiable and is corrected there. |
| #1099's premise that every cell has non-zero `inspection.tokenUsage.totalTokens` is false: five of eight are zero | **Accepted.** Corrected on #1099. The dependent acceptance case still holds, since the three Plain cells that do carry inspection exercise it. |
| `shopping/plain` publishes as `completed` with a terminal `quality-failed` backend verdict | **Accepted as a publication caveat**, same treatment as the mixed revision. |
| Three documents still name `tokens.svg` and `time.svg` | **Already owned by #1099**, in flight on this branch. |
| #1098's acceptance line about `pnpm format` is vacuous, because the formatter globs only `.ts`, `.md`, and `.mdx` | **Accepted.** The files are Prettier-clean regardless, which the reviewer confirmed. |

samchon (commented):
**Individual Self-Review — `0a4e6cc24` "resolve the upstream checkout that exists"** (closes #1101)

A read-only reviewer read the parent-to-commit diff. The passing checks: hoisting and initialization order are sound, `path.resolve` runs before `process.chdir(ROOT)` so a relative argument anchors on the invocation directory, no stale copy of the old path survives anywhere in the repository, Windows separators are safe at every consumer, the README is Prettier-clean, the documentation voice holds, and the blast radius is nil because no workflow, script, or package entry invokes these files.

Four findings were accepted and are fixed in `004ae4c18`:

| Finding | Disposition |
| --- | --- |
| `EVIDENCE_UPSTREAM=""` passes `??`, and `path.resolve("")` is the current directory, so an empty override silently retargets the comparison at this repository and fails with a missing-ref message | **Accepted, substantive.** This reproduces the exact misleading-error shape the commit existed to remove. An empty value is now unset. |
| The environment beat `argv[2]`, so the hint told a reader to pass an argument that could not take effect while the variable was set | **Accepted.** The argument now wins: naming a path in one run is the more specific statement than an ambient variable. |
| git's own `fatal:` still printed above the wrapped error, because `execFileSync` inherited stderr | **Accepted.** Captured with an explicit `stdio`. |
| The same hardcoded-root defect survives in `ROOT` across all three scripts, which the commit's own comment argues against | **Accepted**, although the reviewer marked it out of #1101's scope. The development skill's rule against patching one witness of a cause applies: all three now derive the root from `__dirname`. |

Two were recorded without a code change. The `.git` existence check is looser than git for a bare clone or a subdirectory of a checkout, which is acceptable for machine-local tooling and now fails with a message that names what it looked for. The `=== false` versus `!` inconsistency inside the file is pre-existing style, and this change follows the predicate spelling already used at the top of the file.

Verified after the fix: default resolution, empty environment, argument over environment, and a checkout without the campaign ref, with no raw git output reaching the console.

Issue #1118, opened 2026-08-07
The Swagger and Prisma bridges carry unit identities, not content, so no per-unit review fingerprint exists


samchon:
## Problem

The Swagger and Prisma process bridges carry unit **identities** to the Go side, not unit **content**. Nothing downstream can therefore compute a per-unit content fingerprint for either kind, which is what #1115 needs in order to expire a stale review.

This is one root cause with two witnesses and two different consequences.

For Swagger there is no per-operation content on the Go side at all. For Prisma there is a strict subset of each declaration: a field's name, its symbol kind, and its documentation comment, and nothing else. A field's type changing from `String` to `Int`, a removed `@unique`, or a changed `@default` never reaches Go, so a digest over the current payload reports a fresh review for exactly the class of change a specification review exists to catch.

Expected behavior: each bridge returns a per-unit content digest, computed on the JavaScript side from the normalized value it already holds, under a canonical serialization that is stable across runs and across normalizations. `requireReview` then works for all four artifact kinds.

Affected users: any consumer whose citations target a Swagger operation or a Prisma model or member and who wants `requireReview` on that reference. Until this lands, #1115 rejects the flag on those references at decode.

### The boundary is deliberate and documented

`packages/evidence/src/internal/loadSwaggerOperations.ts`, on `loadSwaggerOperations`:

> The native contributor is Go, while the version converter is JavaScript. This function is the narrow process boundary between them: it accepts only source locations and returns only operation identities.

The payload type states the same thing structurally:

```ts
interface ISwaggerOperation {
  method: string;
  path: string;
}
```

`operationsOf` builds each entry through `operationOf(method, operationPath)`, which reads nothing from the `OpenApi.IOperation` value it just looked up. The operation object is in scope and discarded.

On the Go side, `native/swagger.go`, `swaggerOperationUnit` receives `swaggerOperation` and builds a unit from `Method` and `Path` alone. There is no third field.

### The Prisma payload is a subset, and the missing part is the important part

`native/prisma.go`:

```go
type prismaModel struct {
  Name          string        `json:"name"`
  Documentation string        `json:"documentation"`
  Fields        []prismaField `json:"fields"`
}

type prismaField struct {
  Name          string `json:"name"`
  Symbol        string `json:"symbol"`
  Documentation string `json:"documentation"`
}
```

`Symbol` is the `model` / `column` / `relation` classification, not a Prisma type. So the following schema changes are invisible to any Go-side digest built from this payload:

[...]

Issue #1117, opened 2026-08-07
The evidence README says evidence/documented takes no options


samchon:
## Problem

`packages/evidence/README.md` states that `evidence/documented` takes no options. It takes `ITtscEvidenceDocumentedConfig`.

The README rules table is the shortest description of the rule set and the first place a consumer looks before configuring anything. A reader who trusts it never learns that the rule's population is selectable, so they either enable the rule on a wider surface than they meant or conclude the rule is unusable for their project and leave it off.

## Evidence

`packages/evidence/README.md`, the rules table:

```
| `evidence/graph` | [`ITtscEvidenceGraphConfig`](...) | The graph itself. Project-scoped, so its entry declares no `files`. |
| `evidence/documented` | nothing | Requires a JSDoc block on every selected export, since a block is the only place a citation can live. |
| `evidence/singular` | nothing | Keeps one public identity per file, named after the file. |
| `evidence/todo` | nothing | Fails on every remaining JSDoc `@todo`, with its own text. |
```

The row's own description says "every **selected** export", so the table contradicts itself in one line: something must do the selecting.

Four independent confirmations that the rule is configurable:

- `packages/evidence/native/documented.go` declares `documentedOptions struct { Symbol json.RawMessage \`json:"symbol"\` }` and `decodeDocumentedOptions` decodes it, rejects unknown fields, and reports "expected an ITtscEvidenceDocumentedConfig object, or a bare severity for the default selection."
- `packages/evidence/src/structures/ITtscEvidenceDocumentedConfig.ts` exists.
- `packages/evidence/src/typings/ITtscLintRuleOptionsMap.ts` declares `"evidence/documented": ITtscEvidenceDocumentedConfig`.
- `website/src/content/docs/evidence/rules.mdx` documents it with `"evidence/documented": ["error", { symbol: ["type", "function"] }]` and an option table.

`documentedRule` also does not implement `AcceptsTtscLintOptions()`, unlike `todoRule` which declares `false` explicitly. The unimplemented marker defaults to accepting options, which is correct for this rule and is the runtime confirmation that the host passes an options object through.

The `singular` and `todo` rows are correct: `singular.go` and `todo.go` take nothing, and `todo.go:37` pins the refusal.

## Consequence surface

The README only. `rules.mdx`, `src/index.ts`, and the options map are all correct, so no other document inherits the error. No code change, no behavior change, no test.

## Approach

[...]

Issue #1116, opened 2026-08-07
The evidence skill names the reference policy noExclude; the public key is noEvidenceExclude


samchon:
## Problem

`.agents/skills/project/evidence/SKILL.md` names a reference policy `noExclude` three times. No such configuration key exists. The public key is `noEvidenceExclude`, and `noExclude` is only the internal Go field name.

The skill is the document an agent reads before changing evidence rule semantics or the configuration surface, so a wrong key name there is copied into configuration that the decoder then rejects as an unknown field. The failure is loud rather than silent, but the repair path leads through the wrong document.

## Evidence

Occurrences in `.agents/skills/project/evidence/SKILL.md`:

| Line | Text |
| --- | --- |
| 50 | "an `@evidenceExclude` target does the same unless its reference declares `noExclude`" |
| 56 | "and under `noExclude`, not even the second one" |
| 127 | "A reference may strengthen its own acknowledgement relation with `noExclude`, `uniqueEvidence`, and `singleEvidencePerSymbol`, declared flat on the reference object." |

The public key, three independent confirmations:

- `packages/evidence/src/structures/ITtscEvidenceGraphReferenceBase.ts:39` declares `noEvidenceExclude?: boolean`.
- `packages/evidence/native/config.go:185` lists `"noEvidenceExclude"` in the reference field allow-list, so any other spelling is rejected by `rejectUnknownFields`.
- `packages/evidence/native/config.go:306` decodes `decodeFlag("noEvidenceExclude", &policy.NoExclude)`.

`noExclude` is the Go struct field at `packages/evidence/native/model.go:83`. Line 127 states the name in a sentence explicitly about what is "declared flat on the reference object", which is the public key and not the internal field, so the spelling is wrong there on the document's own terms. Lines 50 and 56 are prose about what a consumer declares, so they carry the same error.

The other two policies in the same sentence, `uniqueEvidence` and `singleEvidencePerSymbol`, match their public keys exactly, which is what makes the third one read as correct.

## Consequence surface

The skill only. `website/src/content/docs/evidence/claims.mdx`, `packages/evidence/README.md`, and the TypeScript structures all use `noEvidenceExclude`, so nothing downstream inherits the error. No code change, no behavior change, no test.

## Approach

Replace all three occurrences with `noEvidenceExclude`. Keep the surrounding sentences intact; the claim each one makes is correct, only the identifier is wrong.

[...]

Issue #1115, opened 2026-08-07
A review never expires: fingerprint the cited scope's content and add a requireReview reference policy


samchon:
## Problem

Nothing in `@ttsc/evidence` detects that a cited unit changed after its citation was written.

A citation is checked for resolution, host eligibility, and coverage, and all three are properties of the graph as it stands right now. Rewrite the body of a cited Markdown section, change what a cited Swagger operation returns, add a column to a cited Prisma model, or change the contract of a cited TypeScript symbol, and every existing citation of it stays green. The prose reason still reads plausibly, because it was written about the previous content and nothing compares the two.

That makes a review permanent, which is the same as making it worthless. Once #1114 lands, a project can require an `@evidenceReview` beside every citation, but the review is written once and stays green forever. On the second pass over a large citation set there is no way to tell which reviews were written against content that has since moved, so the only options are re-reading everything or trusting all of it.

Expected behavior: a reference may declare `requireReview: true`. Every citation acknowledging that reference must then carry an `@evidenceReview` whose `#`-prefixed fingerprint matches a digest of the cited scope's current content. When the content changes, the fingerprint stops matching and the build fails again with the new value stated in the diagnostic.

Affected users: any consumer whose cited specifications change over time, which is all of them, and specifically anyone who needs to know which subset of a large citation set is stale.

### What this proves and what it does not

It proves that the review statement was written against the cited content as it now stands. It does not prove that anyone read that content, because the diagnostic states the expected fingerprint and a fingerprint can be pasted from an error message. It does not prove the prose is sincere, which `.agents/skills/project/evidence/SKILL.md:37` forbids attempting.

The value is expiry, and expiry is what makes a second pass over a citation set finite. State it that way in every document this issue touches.

### Units carry no content and no extent

`packages/evidence/native/model.go`, `evidenceUnit`, in full: `ID`, `ParentID`, `Target`, `Identity`, `Aliases`, `Type`, `Symbol`, `Path`, `Line`, `Readable`, `Hidden`.

`Line` is a start position. There is no end position, no extent, and no content. Every unit construction site confirms it: `markdown.go:226-235` sets `Line: index + 1` and nothing further, `markdown.go:129-135` gives a file unit `Line: 1`, `swagger.go:276` builds an operation unit, and `typescript.go:849` builds from an AST node without retaining any span.

### The existing digests are cache keys, not unit content

[...]

PR #1119, opened 2026-08-07
feat(evidence): require a reviewed, expiring statement beside every evidence citation

### LINKED ISSUE #1092: Requirement coverage can be discharged without delivery: ancestor scope bypasses noEvidenceExclude, and hedged citations substitute for exclusions
## What the audits found

A claim reference may set `noEvidenceExclude: true` so that no unit of it can ever be excused. The benchmark's frontend graph uses this on the operation reference of `frontend-hooks` and on the hook reference of `frontend-screens`, and it works exactly as intended — across three audited codebases, 23/23, 43/43 and 100/100 published operations reached a hook, with no false citation anywhere.

The requirement references do not set it, and they are where coverage is discharged without delivery. Two distinct mechanisms appeared, and only the second is a semantic question about the rule itself.

## Mechanism 1 — an ancestor citation discharges its whole subtree

`todo`'s `src/components/home/home-page.tsx:8` cites

```
@evidence docs/analysis/01-actors-and-auth.md#actors-authentication-and-account-authority
```

which is the **H1** of that file, not one of the H2/H3 units the claim selects. Because an ancestor stays resolvable as an aggregate scope and acknowledges every selected descendant, this single tag discharges **all fourteen** H2/H3 headings in the file for the screens claim. `tests/journeys/authentication.spec.ts:10` does the same for the journeys claim.

The consequence is concrete: `REQ-AUTH-BOUNDARY` and its two sub-rules are named by no screen at all. They are genuinely implemented — by a route guard in `App.tsx:41-47` — but that file is outside the screen population, so the ledger never sees the obligation being met or missed.

`reddit` shows the sharper version, where the ancestor is cited by a host that does not own the subtree:

- `post-page.tsx:15` cites `#req-func-post-post-operations`, whose subtree includes **POST-001 Create a Post** — but PostPage does not create posts; FeedPage does.
- `account-page.tsx:17` cites `#req-rule-identity-account-identity-rules`, whose subtree includes the registration rules — registration is AuthPage.
- `moderation-page.tsx:37` cites `#req-rule-report-reporting-rules`, whose subtree includes report submission — which lives on PostPage.

The workspace's own skill document already forbids this: "A parent target is truthful only when the host owns the complete selected subtree." It is a stated rule with no mechanical enforcement, so it holds only as far as the author's honesty.

**The question for the rule set:** should an ancestor citation be able to discharge descendants of a reference that refuses exclusions? Today `noEvidenceExclude` closes the front door while ancestor scope leaves a side door open. Candidate answers, none obviously right: restrict aggregate scope to references that permit exclusions; require an ancestor citation to name the subtree it claims; or leave it and treat subtree ownership as reviewer-enforced, as it is now.

## Mechanism 2 — hedged citations that no code change could falsify

Where an exclusion would have to state a reviewable reason, a vague citation costs nothing. `reddit` wrote zero exclusions — both carriers are empty — and instead attached citations like these to requirements a browser cannot deliver:

| Citation | What the requirement says | What the code does |
| --- | --- | --- |
| `moderation-page.tsx:21` — "Keeps resolution history scoped to moderation." | owners and moderators inspect **resolved** history: outcome, reporter, reason, acting moderator, decision time | renders only the **unresolved** queue; no SDK operation exposes resolved history at all |
| `moderation-page.tsx:25` — "Shows the active history boundary." | "Former bans remain private history outside the active banned-user list." | renders only **active** bans |
| `post-page.tsx:74` — "Shows author context." | vote transitions apply +1, ±2, or the inverse to author karma | renders `u/{username}` |
| `feed-page.tsx:31` — "Uses bounded feed requests." | a supplied page size is 1–100; zero, negative, fractional or >100 are refused | `limit` is hard-coded; there is no page-size control, so nothing is ever validated |

The skill asks for "a specific responsibility current code could falsify." None of these is falsifiable by any change to the code, which is what lets them stand in for delivery.

`shopping` took the opposite route — 47 exclusions in each carrier, covering the same 47 headings — and 46 of them are defensible: a browser genuinely does not own a persistence aggregate or an authorization boundary, and each reason is written as a falsifiable condition. One is not. `#req-search-policies` was excused as "the backend enforces search and listing policy", but its sub-rule **REQ-SEARCH-POLICIES-4 is titled "Render the standard product card"** and specifies exactly what the browser must paint. The DTO supplies every field; the catalog renders none of the three. A rendering obligation was excused as an authorization one.

Worth recording as a structural observation rather than a fault: `shopping` excluded three of its five requirement documents **in full** from both frontend claims. Each individual reason is plausible; the sweep is what stands out.

## What this does not claim

The graph is not being reported as broken. It did the thing it can do, in all three codebases, against an agent that had every incentive to cut corners: a complete and truthful operation-to-hook mapping that no note in `omissions.md` could have excused. The gap is narrower and worth stating exactly — the rule can enforce that a citation **resolves**, and cannot enforce that the citing artifact **delivers what the cited heading describes**. Both mechanisms above live in that gap.

Whether either warrants a rule change is the maintainer's call. Mechanism 1 is a genuine semantic question about aggregate scope interacting with `noEvidenceExclude`. Mechanism 2 may not be addressable in a linter at all.

samchon:
Claims the complete `evidence/review` campaign cycle. Opened implementation-free, before any code exists, because #1112 currently edits every file this cycle needs and that overlap has to be visible before either side writes more.

Verification is pending. Nothing is implemented at this head.

## What this cycle implements

A citation is checked for resolution, host eligibility, and coverage today. All three are properties of the graph as it stands, so nothing records that anyone verified a citation, and nothing detects that a cited unit changed after its citation was written. This cycle adds both, and claims nothing beyond them.

### 1. `@evidenceReview`, a third tag (#1114)

```text
@evidenceReview <target> <description>
@evidenceReview <target> #<fingerprint> <description>
```

A new no-option file rule `evidence/review` requires that every `@evidence` and `@evidenceExclude` on a selected TypeScript export carry a review naming the same target. It judges pairing, orphans, duplicates, and malformed tags inside one comment on one identity, reusing `documentedHosts` and `hostNodesOf` so its population cannot drift from the population the graph selects.

The tag is **not** a third `tagKind`. A review must reach neither the acknowledgement maps in `evaluateEvidenceGraph` (`acknowledged`, `evidenceByUnit`, `exclusionByUnit`, `evidenceByHostAndScope`, `evidenceUnitsByHost`, `evidenceHostsByUnit`) nor the non-participating pass. A review counted as evidence is a green build earned by a review, which is the one failure this cycle must not ship.

The `#` prefix on the fingerprint is required rather than inferred, for the reason `declaration.go` already records about the target grammar: the token stays self-discriminating through a boundary character instead of a guess. Ordinary prose supplies the counter-examples (`cafe`, `deadbeef`, `beefed`).

### 2. A content fingerprint and `requireReview` (#1115)

`evidenceUnit` gains an own-content digest, and a scope digest is composed over the cited unit and its **structural** subtree via `ParentID`. A reference may then declare `requireReview: true`, and every citation acknowledging it must carry a fingerprint matching that digest. When the cited content moves, the fingerprint stops matching and the build fails again.

Three properties are load-bearing:

- **The fingerprint is a function of the citation, not of the reference.** `UnitsByScope` is per reference, so a covered-set digest would demand two different values from one token. A structural subtree digest is reference-independent.
- **The digest excludes every position where a tag can live.** Otherwise writing the review changes the digest the review is checked against, and the repair never terminates. This is not only self-citation: a property's JSDoc is interior text of the interface containing it.
- **The digest is normalized before hashing.** Line endings to `\n`, trailing whitespace per line removed, so a CRLF and an LF checkout of one commit agree and a reformat does not expire a review.

The diagnostic states the expected fingerprint. It has to: `hints.go` publishes a corpus only on a cycle where the rule passes, so the cycle that most needs to offer the value is exactly the cycle that publishes nothing.

No autofix, no `--fix` path. A fixer that inserts the current fingerprint removes the only thing the feature adds.

### 3. Per-unit digests across the process bridges (#1118)

The Swagger and Prisma loaders reach Go through a JavaScript bridge that carries unit identities, not content. `ISwaggerOperation` is `{method, path}` and discards the operation object it just looked up; `prismaModel` carries name, documentation, and field names, so a field's type, `@unique`, `@default`, and `@relation` arguments never cross. Each bridge gains a per-unit digest computed on the JavaScript side under a canonical serialization, since `JSON.stringify` key order follows insertion order and is not stable across normalizations.

The whole-source digests both bridges already return stay exactly as they are. They are cache keys. Reusing one as a fingerprint makes every unit in a document share a value, so one endpoint change expires every review in a 100-operation document.

### 4. Two documentation defects found in the lines this cycle edits (#1116, #1117)

- `.agents/skills/project/evidence/SKILL.md` names the reference policy `noExclude` at three occurrences. The public key is `noEvidenceExclude`; `noExclude` is only the Go field.
- `packages/evidence/README.md` states that `evidence/documented` takes "nothing". It takes `ITtscEvidenceDocumentedConfig`, and the same row's own description says "every **selected** export".

Both sit inside the exact paragraph and table this cycle rewrites.

## What this cycle does not claim

It does not judge whether prose is sincere. `.agents/skills/project/evidence/SKILL.md:37` forbids that, and #1092 mechanism 2 is the field evidence for the warning: a second prose block is as hedgeable as the first. What is mechanically proved is that a separately addressed statement exists and that it was written against the cited content as it now stands. Expiry is the product, and expiry is what makes a second pass over a citation set finite.

It does not close #1092 mechanism 1. A subtree digest expires an ancestor citation's *review* when the subtree changes; the citation still discharges coverage for descendants the host does not own.

## Overlap with #1112

#1112 edits `declaration.go`, `config.go`, `model.go`, `graph.go`, `markdown.go`, `typescript.go`, `prisma.go`, `hints.go`, `documented.go`, `ITtscEvidenceGraphReferenceBase.ts`, the evidence README, all three evidence website guides, and the same `SKILL.md` sections. Two of the collisions are semantic rather than textual, and both are recorded as a review on that pull request with the adjustments it needs. This cycle lands first; #1112 rebases onto it.

## Issues in this cycle

Implementation order follows the DAG, not the pull-request count.

| Issue | Subject | Predecessor |
| --- | --- | --- |
| #1114 | the `@evidenceReview` tag and the `evidence/review` rule | none |
| #1115 | the content fingerprint and the `requireReview` reference policy | #1114 |
| #1118 | per-unit digests across the Swagger and Prisma bridges | #1115 |
| #1116 | `SKILL.md` misnames `noEvidenceExclude` | none |
| #1117 | the README says `evidence/documented` takes no options | none |

Because #1118 lands in this same cycle, the decode-time refusal of `requireReview` on a Swagger or Prisma reference that #1115 specifies is an intermediate state inside this pull request, not shipped behavior.

## Verification

Pending. Local builds and tests are not being run at the user's instruction, so this pull request's ordinary CI is the only execution gate for every lane. Every command that could not be run will be named in the final report.


samchon (commented):
## #1118 deferred to a following cycle

At the author's direction, #1118 leaves this cycle and becomes its own pull request. This one is completed and merged first.

The branch is coherent without it. #1115 already refuses `requireReview` at decode on a Swagger or Prisma reference, naming #1118 as the reason, so the shipped behavior is honest rather than partial: Markdown and TypeScript references enforce the review and expire it, and the two kinds whose loaders report unit identities rather than unit content say so instead of comparing against nothing.

What #1118 still owes, unchanged: a per-operation digest from the Swagger bridge under a canonical serialization, a per-model and per-field digest from the Prisma bridge covering the type, attributes, and defaults its payload currently drops, the matching Go payload fields, the cache-interaction proof, and removal of the decode refusal.

### Cycle scope as it now stands

| Issue | State |
| --- | --- |
| #1114 | implemented, `ae1225af` |
| #1116 | implemented, `ae1225af` |
| #1115 | implemented, `5c847c37` |
| #1117 | implemented in `ae1225af`, closing line recorded in `58ab7770` |
| #1118 | deferred, next pull request |

This departs from the campaign procedure's one-pull-request-per-cycle rule, which would keep #1118 here because its only predecessor is #1115. The departure is the author's decision and is recorded rather than silent.

samchon (commented):
﻿## Overall Self-Review, round 1

Solo, over the complete `master..feat/evidence-review` diff at `5ba81133`. Four findings, none of them a build break, and two verifications that came out clean and are worth recording because I expected them to fail.

### Verified clean

**Every TypeScript unit does get a digest, including a `package` reference's.** I expected a silent hole here: if entry-traversed or package-selected units bypassed the scanner that assigns `Digest`, `scopeFingerprint` would return an empty string, `reviewProblems` would return nil, and `requireReview` would silently enforce nothing on exactly the reference kind the option is most useful for. Traced instead: `scanTypeScriptInventory`, `typescript_cache.go:70`, and `typescript_cache.go:88` are the only constructors, all three route through `scanTypeScriptInventoryAt`, and that calls `collectTypeScriptDeclarations` where the digest is assigned. No hole.

**`documentedHost.Nodes` really does hold both halves of a merged identity.** `recordUnitNode` runs during `collectTypeScriptStatements`, so `UnitNodes` is populated in the inventory `documentedHosts` builds, and an exported namespace registers the same `type` symbol an interface does, so `hostNodesOf` keeps both. A review on `namespace ISale` therefore answers a citation on `interface ISale`, which is what the merged-identity case asserts.

### Findings

**1. `evidence/review` is wider than #1114 says it is.** MEDIUM.

The issue's Problem section says the rule requires a review on every citation "on a selected TypeScript export". The rule as built judges every public identity in every linted file, because a file rule has no claim context and cannot know what a claim selects. A citation sitting in a file no claim matches still owes a review.

This is the right behavior for a no-option file rule and it matches how `evidence/documented` and `evidence/singular` scope themselves, through the outer `files` setting. But the issue is the operative handoff and it disagrees with the code, so the issue text is what has to change, not the rule. I will correct #1114 rather than leave a reader to discover the difference.

**2. Two `requireReview` references over one cited scope produce two findings for one tag.** MEDIUM, accepted.

`reviewProblems` runs inside the per-reference loop, so a declaration acknowledging two such references is reported twice. Each message names its own claim and reference, which is how every other per-obligation diagnostic in this rule behaves, and the alternative would be to report one finding that names an obligation the author then has to guess at. Accepted as consistent rather than changed, and recorded here so it is not rediscovered as a defect.

**3. Enabling `evidence/review` beside `evidence/documented` runs the TypeScript collector twice per file.** LOW.

Both rules call `documentedHosts`, which runs `collectTypeScriptStatements` from scratch. This is a real cost on a large project and it is not a correctness problem. Sharing the collected population would mean a cross-rule channel, and the only one the contributor API offers is graph project state, which is published solely on a cycle where the graph reports nothing. That would make the review rule go silent exactly when the graph is broken, which is worse than the duplicated walk. Left as is, deliberately.

**4. A TypeScript reference unit's own documentation block is excluded from its digest.** LOW, deliberate, and worth stating plainly.

Excluding every position a tag can occupy is what stops a review from invalidating itself, and for TypeScript that is the JSDoc blocks of the whole subtree. The consequence is that a contract change expressed only in prose, with no change to the declaration itself, does not expire a review of it. That trade is recorded in the skill's Reference Policies section. The alternative reintroduces a non-terminating repair loop, so the trade stands.

### Not yet verified

Nothing in this diff has been compiled or executed. `gofmt` parses every file in `packages/evidence/native` cleanly, which establishes syntactic validity and nothing else: types, identifier resolution, arity, nil maps, and every behavioral claim in the new tests are unverified locally at the author's instruction. CI on this head is the first execution of any of it.

Round 1 applied no code change, so it does not restart itself. The next round runs against whatever CI and the outstanding Individual Self-Reviews produce.

samchon (commented):
﻿## Individual Self-Review results, adjudicated

Three read-only reviews, one per pushed commit. They found real defects and I have taken almost all of them. Repaired in `69f954bb`.

### Accepted and fixed

**1. The fingerprint was a function of the reference, not of the cited address.** CONFIRMED, and it made this feature's central documented claim false in four documents.

`reference.Units` and `reference.Scopes` are both narrowed by the reference's `symbol` selector, so an unselected descendant appears in neither. Three reachable consequences: a Markdown reference selecting only `h2` fingerprinted a cited section without the H3 bodies inside it and never expired however much of that subtree was rewritten; `symbol: ["file"]` fingerprinted the preamble alone; and two `requireReview` references over one scope under different selectors demanded two different digests from a tag that carries one token, so **no value an author could write would go green**. The reviewer also caught that my own subtree test hid the defect by widening the selector to `["h2","h3"]` — it was written around the limitation instead of exposing it.

`referenceState` now keeps `Population`, the complete materialized set, and the digest walks that. Two new cases pin it: one holds the selector at `h2` and asserts a subtree rewrite expires the review, one gives a citation two `requireReview` references with different selectors and asserts one token satisfies both.

**2. A Prisma-hosted citation under `requireReview` was an unfixable dead end.** CONFIRMED. `parseReviews` was never called from the Prisma loader, so `state.Reviews` was always empty for a Prisma claim and every citation reported unreviewed forever while the repair the message named had no effect. A diagnostic must never name a repair the author cannot perform. Reviews are read from `///` runs now.

**3. Reformatting expired a review.** CONFIRMED, and it violated the invariant I wrote three commits ago. A TypeScript node's position is its *full* start, so an undocumented declaration's span begins at the previous token and carried the blank lines and `//` comments above it. Inserting a blank line elsewhere expired the review, and inconsistently: the same edit above a *documented* declaration was neutral, because the whole leading run through the block is excluded as a tag position. Leading blanks and a leading line comment normalize away now, so both cases agree.

**4. A Markdown review was swallowed into the citation above it.** CONFIRMED, and #1114's Evidence section asserted the opposite. The Markdown declaration scan ran with tag boundaries off, so no `@`-opening line closed the declaration before it: the review vanished and the reason grew a sentence addressed to a different question. Latent when #1114 landed, live the moment #1115 started reading Markdown reviews.

**5. Content under an H5 or an anchorless heading belonged to no digest.** CONFIRMED. Those headings advance the current host to an ID no unit carries, so their lines accumulated in a bucket nothing read, and rewriting an `##### Details` section under a cited H4 expired nothing. They fold into the nearest ancestor unit now.

**6. A mid-line HTML comment stayed in the digest**, so writing a review after prose on the same line invalidated it. Fixed by removing only the comment span, never the line, which the reviewer's own follow-up note about losing content beside `-->` is what steered.

**7. An uppercase fingerprint alone was accepted as a description.** CONFIRMED. `#A3F9C1D` failed the lowercase check and was returned as prose, so a review stating nothing satisfied the non-empty test. Case was all that separated the caught form from the uncaught one. A rejected token is still kept when prose follows it, because a requirement anchor is spelled the same way.

**8. One review per declaration of a merged identity was reported as a duplicate.** CONFIRMED. Citations were deduplicated across the blocks of one identity and reviews were not, so an overload set written the normal way — by copying the block onto each signature — was reported as duplicated while every citation was reviewed exactly once. A duplicate is two reviews inside one block.

**9. "four rules" survived in four website documents**, one of them a file I had edited, and the evidence overview enumerated the rule set without `evidence/review`. Corrected.

**10. The test helper read the wrong value out of a Stale diagnostic.** CONFIRMED and latent: that message names the review's own outdated fingerprint before the expected one, both quoted, so a positional scan returns the value already in the source. Every case built on it would fail as stale rather than as a broken helper. The helper now refuses any message that is not asking for a value.

### Accepted, not fixed here

**Findings anchor on the identity's founding declaration, discarding the position the parser captured.** An orphan review written on `namespace ISale` underlines `interface ISale` above it, and `Duplicate @evidenceReview` underlines neither competing tag while saying "remove the other". `parsedReview.LineOffset` is computed and never read. Anchoring the *citation* findings on the identity is right, since that is `evidence/documented`'s boundary; anchoring malformed, orphan, and duplicate there is not. Real, and a separate change: it needs `ctx.ReportRange` and a position carried from the parser through to the rule.

**No end-to-end coverage.** `evidence/review` is the only evidence rule never run through the real binary, so nothing proves the `rules` tuple binds to the Go registration or that the host refuses a configured payload. `tests/test-evidence/src/features/` is where that belongs.

**The digest cost is unconditional.** Every TypeScript and Markdown unit is hashed on every scan whether any reference wants it. Gating it would remove the whole regression for projects not using the option.

**Acceptance cases still untested**, all named in #1114 and #1115: an exclusion under `requireReview`; a unit reachable by two addresses; an unhealthy population; a healthy-but-empty population; a withdrawn descendant inside a cited scope; a review split across continuation lines; `@evidenceReviews`; a class beside a same-named namespace.

I am recording these rather than closing them silently. They are the honest remainder of this cycle.

### Rejected

**A `#` token that is exactly seven lowercase hex characters could shadow a real digest in the test helper.** Shape-only guard, no fixture in the package reaches it, and the alternative is parsing the diagnostic grammar. Left, recorded.

**`readHostTags`' comment misstates why its block dedupe exists.** The reviewer is right that the variable-statement rationale does not apply. The dedupe is still correct defensively. Comment left as a known inaccuracy rather than churned in a commit already this wide.

### Verified clean, worth recording

The reviewers proved two things I expected to be broken. `documentedHost.Nodes` really does hold both halves of a merged identity, because an exported namespace registers the same `type` symbol an interface does. And the TypeScript self-invalidation guarantee holds exactly, because a JSDoc node's `Pos()` is the host's full start rather than the offset of the opening slash, so excluding the span also swallows the whitespace before it and adding a block anywhere inside a declaration is byte-neutral.

They also confirmed no duplicate top-level declaration, no arity error, and no unused import across the thirteen new Go files.

samchon (commented):
## Overall Self-Review, final round

Solo, over the complete `master..331bf36d` diff. Clean: nothing survived verification this round.

The rounds that preceded it are the substance, so they are recorded rather than summarized away. Rounds 3 through 6 each found a defect and one of them was mine twice over: round 3's Markdown attribution fix introduced the false expiry that round 4 had to correct, and round 4's withdrawn-content claim was unachievable against a digest that is a declaration's own text, which CI proved by failing the assertion that stated it. Round 6 found the worst of them, a review matched by `HostID` — a source position — so the package's two rules disagreed about the same file while `model.go` warned against exactly that use of the field.

What CI added on top: a Go parse error, a broken helper that would have failed as stale rather than as broken, a pre-existing dependency advisory, a lockfile refresh of mine that downgraded `sharp` past its own fix, a stale validation-plan mirror, and the one that mattered most — `stripInternal` deleting `requireReview` from the published declarations because a doc comment named a withdrawal tag. Go accepted the option, every Go case passed, and no consumer could declare it. Only the end-to-end fixture that type-checks its own config could catch that, and that fixture exists because a reviewer argued for it after I had deferred it.

The last change was a grammar split the author asked for and I had missed: reviews are attached per acknowledgement, so an exclusion cannot share a review tag with a citation. `@evidenceExcludeReview` answers an exclusion, `@evidenceReview` answers a citation, the pairing key is tag-and-target in both the file rule and the graph, and a review filed under the wrong question has its own finding rather than an orphan message that would send an author after a typo. That split also expresses a case the single tag could not: one target cited by one claim and excluded for another owes two reviews.

### Verification actually performed

Local, once permitted: `go test ./native/ -count=1` passes; `go build` and `go vet` clean; `gofmt` parses every file; `node scripts/ci/dependency-audit.cjs` reports `high=0, critical=0`; `node --test scripts/assert-project-layout.test.cjs` passes; the repository formatter leaves the tree unchanged; and `pnpm --filter @ttsc/evidence build` emits `requireReview`, which is what proved the `stripInternal` defect.

CI on this head: 28 checks pass, none failing, including `go`, `windows-go`, `typecheck`, `evidence defenses`, all six platform rows, and every downstream integration lane.

### Outstanding, and deliberately so

**#1118 is deferred to its own pull request.** The Swagger and Prisma bridges report unit identities rather than unit content, so `requireReview` is refused at decode on those reference kinds with a diagnostic naming that issue. The shipped state is honest rather than partial.

**Findings anchor on the identity's founding declaration.** `parsedReview.LineOffset` is collected and unread, so an orphan review written on `namespace I` underlines `interface I`. Correct for the citation findings, wrong for malformed, orphan, mismatched, and duplicate. It needs `ctx.ReportRange` and a position threaded from the parser.

**The digest cost is unconditional.** Every Markdown and TypeScript unit is hashed on every scan whether or not any reference asks for it.

**A TypeScript citation expires on any change inside its declaration**, including one behind a withdrawal tag, because a declaration's digest is its own text and a nested member sits inside it. Documented on the property so a consumer can choose a narrower symbol; removable only by adding a second exclusion mechanism.

None of these is a defect in what ships. They are the honest remainder, recorded here and in `.wiki` so the next cycle does not rediscover them as new.

Issue #1135, opened 2026-08-09
docs(lint): the public rule.TextEdit autofix contract describes the pre-#605 per-edit applier


samchon:
## Problem

The public contributor contract for autofix conflicts describes the applier `@ttsc/lint` replaced in #605. `rule.TextEdit`'s doc comment, which the website walkthrough names as the authority, says a conflict drops individual edits and keeps the rest. The host has been per-finding atomic since #605: a finding whose edits collide with an already-selected finding is dropped whole, so nothing half-applies.

A contributor reading the shipped contract designs against the wrong failure mode. The doc's own advice, "design fixes so each finding emits one contiguous TextEdit covering the entire replacement region", exists to dodge a partial-application hazard the host no longer has, and it is advice a rule like `no-import-type-side-effects` cannot follow.

## Evidence

`packages/lint/rule/rule.go`, on `TextEdit`:

> Application policy: a rule may emit several `TextEdit`s in one `ReportFix` / `ReportRangeFix` call, in any order. The host treats the per-pass edit set as a candidate list. Within a single fix pass, edits must not overlap each other; when two edits cover overlapping ranges (either from one rule emitting multiple edits in one call, or from two different rules in the same pass), the host applies the earliest-starting / shortest edit and silently drops the rest. There is no diagnostic for dropped edits, and the host does not currently report when a comment falls inside a deletion range. Design fixes so each finding emits one contiguous TextEdit covering the entire replacement region.

The host in `packages/lint/linthost/fix.go` does something else. `selectTextEditGroups` builds a candidate from the already-selected edits plus the whole group, runs the edit-level selector, and accepts the group only when nothing was dropped:

```go
kept := selectTextEdits(sourceLen, candidate)
// Accept the group only if nothing was dropped: a shorter result means one
// of its edits collided with an already-selected edit (or a sibling edit of
// its own group), so the group cannot apply atomically here.
if len(kept) == len(candidate) {
  selected = kept
}
```

`packages/lint/test/fix/fix_select_text_edit_groups_drops_partially_overlapped_finding_whole_test.go` pins exactly the opposite of the doc: group B has one overlapping edit and one far-away disjoint edit, and the test fails if the disjoint edit survives.

Three separate statements are wrong:

1. **Per-edit dropping.** For a finding with several edits, the host drops the whole finding, not "the rest". The finding then re-fires on the next cascade pass, which is the behavior the doc never mentions and which a contributor needs to know when reasoning about `maxFixPasses`.
[...]

Issue #1133, opened 2026-08-09
fix(lint): format/clause-join skips the else, do, labeled, and with bodies Prettier joins


samchon:
## Problem

`format/clause-join` joins a single-statement clause body back onto its header line, which is what Prettier does. It applies that only to clause bodies whose header ends in `)`, so the `else` branch, the `do` body, a labeled statement's body, and a `with` body are left broken across lines forever. `ttsc format` therefore produces output a Prettier run would change, and a project running both never reaches a fixed point.

## Evidence

`packages/lint/linthost/rules_format_clause_join.go` registers for five kinds:

```go
func (formatClauseJoin) Visits() []shimast.Kind {
  return []shimast.Kind{
    shimast.KindIfStatement,
    shimast.KindWhileStatement,
    shimast.KindForStatement,
    shimast.KindForInStatement,
    shimast.KindForOfStatement,
  }
}
```

`clauseControlledBody` returns the `if` *then* branch and the three loop bodies, and its comment states the exclusion explicitly:

> The `else` branch is intentionally excluded: its body is anchored after the `else` keyword rather than a `)`, so it does not share this rule's `)`-anchored join shape.

`joinClauseBody` then hard-requires that anchor:

```go
if gapStart == 0 || src[gapStart-1] != ')' {
  return
}
```

`KindDoStatement`, `KindLabeledStatement`, and `KindWithStatement` are absent from `Visits()` with no comment at all.

Measured against Prettier 3.8.3, running the real `format` subcommand on a project with an empty `format: {}` block:

| Source | Prettier 3.8.3 | `ttsc format` |
| --- | --- | --- |
| `if (ready)\n  run();\nelse\n  stop();` | `if (ready) run();\nelse stop();` | `if (ready) run();\nelse\n  stop();` |
| `do\n  tick();\nwhile (ready);` | `do tick();\nwhile (ready);` | unchanged |
| `outer:\nfor (const item of items) {\n  break outer;\n}` | `outer: for (const item of items) {` | unchanged |
| `outer:\nrun();` | `outer: run();` | unchanged |
| `with (scope)\n  run();` | `with (scope) run();` | unchanged |
| `if (ready)\n  run();\nelse\n  ;` | `if (ready) run();\nelse;` | `else\n  ;` |
| `do\n  ;\nwhile (ready);` | `do;\nwhile (ready);` | unchanged |

[...]

Issue #1132, opened 2026-08-09
fix(lint): seven published functional/* options are accepted and silently do nothing


samchon:
## Problem

Seven option fields published in `@ttsc/lint`'s typings and documented on the website as working knobs are accepted by the config layer and then never read by the rule. Setting any of them changes nothing, and nothing warns. The user's `lint.config.ts` says one thing and the engine does another.

Three sibling fields in the same family (`allowReturningBranches`, `allowToRejectPromises`, `immutability`) already carry an explicit "Reserved for upstream-compatible configs" note, so the package already has a way to say "accepted, not honored". These seven do not use it.

## Evidence

| Rule | Option | Where it is promised | Why it cannot work |
| --- | --- | --- | --- |
| `functional/no-mixed-types` | `checkInterfaces` | `ITtscLintFunctionalNoMixedTypesRuleOptions` ("Check interface member kinds."), `website/src/content/docs/lint/rules/functional.mdx` | `func (functionalNoMixedTypes) Check` in `packages/lint/linthost/rules_functional.go` calls no `DecodeOptions` |
| `functional/no-mixed-types` | `checkTypeLiterals` | same | same |
| `functional/no-return-void` | `allowNull` | `ITtscLintFunctionalNoReturnVoidRuleOptions` ("Permit a function that returns `null` to satisfy the rule."), same mdx | `func (functionalNoReturnVoid) Check` calls no `DecodeOptions` |
| `functional/no-return-void` | `allowUndefined` | same | same |
| `functional/no-return-void` | `ignoreInferredTypes` | same | same |
| `functional/prefer-tacit` | `checkMemberExpressions` | `ITtscLintFunctionalPreferTacitRuleOptions` ("Check member expressions such as `x => service.map(x)`."), same mdx | `func (functionalPreferTacit) Check` calls no `DecodeOptions` |
| `functional/type-declaration-immutability` | `comparator` | `ITtscLintFunctionalTypeDeclarationImmutabilityRule` ("Comparator applied to the immutability level above.") | `functionalImmutabilityRule` declares only `Identifiers`; the struct has no comparator field |

Mechanical confirmation: no file under `packages/lint/linthost/` contains the quoted strings `"checkInterfaces"`, `"checkTypeLiterals"`, `"allowNull"`, `"allowUndefined"`, `"ignoreInferredTypes"`, `"checkMemberExpressions"`, or `"comparator"`, and no Go identifier of the corresponding name exists anywhere in the package.

[...]

Issue #1131, opened 2026-08-08
fix(lint): format/parameter-properties omits override, so it abstains where Prettier breaks


samchon:
## Problem

`format/parameter-properties` enumerates TypeScript's parameter-property modifiers as four, and TypeScript's own mask holds five. A constructor whose only property modifier is `override` is not recognized as declaring a parameter property, so the rule abstains where Prettier breaks, and `ttsc lint --fix` leaves output Prettier would rewrite.

## Evidence

`anyParameterProperty` in `packages/lint/linthost/rules_format_parameter_properties.go` switches on `KindPublicKeyword`, `KindPrivateKeyword`, `KindProtectedKeyword`, and `KindReadonlyKeyword`, and omits `KindOverrideKeyword`. The doc comment above it and the rule header restate the same four.

The pinned typescript-go defines the mask as five:

```go
ModifierFlagsParameterPropertyModifier = ModifierFlagsAccessibilityModifier | ModifierFlagsReadonly | ModifierFlagsOverride
```

Two sibling rules in the same package already take the mask or name all five: `rules_no_useless_constructor.go` uses the constant, and `rules_ts_extra.go` lists every keyword.

Measured against Prettier 3.8.3, the oracle this rule's own header names:

```ts
class Ctrl extends Base {
  constructor(rate: number, kind: string) { super(); }        // Prettier keeps inline
}
class Only extends Base {
  constructor(override rate: number, kind: string) { super(); } // Prettier breaks one per line
}
```

Prettier force-breaks the second exactly as it does for a `readonly`-only parameter property. `format/parameter-properties` leaves it inline, and the flat form is 51 columns, so width-driven reflow does not explain the difference.

## Consequence surface

- Any class using `override` on a constructor parameter without an accessibility modifier, which is legal TypeScript that emits the field assignment.
- `ttsc lint --fix` produces output a subsequent Prettier run changes, so a project running both never reaches a fixed point.
- The published rule documentation in `packages/lint/src/structures/rules/ITtscLintTypeScriptRules.ts` and `website/src/content/docs/lint/rules/typescript.mdx` describe the four-item behavior; those already list five for the sibling rules, so the package contradicts itself.
- `rules_format_arrow_parens.go` carries a similar comment but rejects any modifier, which is broader and correct for its purpose. It is not affected.

## Approach

[...]

PR #1136, opened 2026-08-09
fix(lint): close the formatter clause gaps and the inert functional options

samchon:
## Intent

One `@ttsc/lint` issue-campaign cycle. This pull request owns the complete accepted set for the cycle; every issue below was published from a full-scope discovery round against `b3779c941` and frozen by a following empty round.

- #1131 `format/parameter-properties` omits `override`, so it abstains where Prettier breaks
- #1132 seven published `functional/*` options are accepted and silently do nothing
- #1133 `format/clause-join` skips the `else`, `do`, labeled, and `with` bodies Prettier joins
- #1134 `ttsc format` never places `else`/`catch`/`finally`/`while` against the closing brace
- #1135 the public `rule.TextEdit` autofix contract describes the pre-#605 per-edit applier

Closing keywords are deliberately absent from this body. The cycle's closing set is the union of the commit closing lines, so an issue this cycle later drops or disproves is not closed by the merge.

## Scope

`packages/lint` and what it owns: the native engine under `linthost/`, the public `rule` package, the published rule typings under `src/`, `packages/lint/README.md`, the lint pages under `website/src/content/docs/lint/`, the Go cases under `packages/lint/test/`, and the `lint defense 1` / `lint defense 2` lanes.

## Verification

Pending. Local evidence and CI results are recorded as formal reviews on this pull request as each commit lands.

### DIFF
--- a/packages/lint/linthost/fix.go
+++ b/packages/lint/linthost/fix.go
@@ -289,10 +289,12 @@ func selectTextEdits(sourceLen int, edits []TextEdit) []TextEdit {
   // insert at that same offset must be dropped: two coincident inserts both
   // pass the `edit.Pos < lastEnd` gate, then apply in reverse sort order and
   // concatenate at one point — silently corrupting the source (e.g. a `;`
-  // insert and a `\n` insert at EOF yielding `\n;`). The host contract keeps
-  // one winner and drops the rest (see rule.TextEdit). A zero-width insert
-  // sitting at the end of a prior NON-empty edit is left alone: it applies
-  // cleanly after the replacement and is a legitimate adjacency.
+  // insert and a `\n` insert at EOF yielding `\n;`). This edit-level selector
+  // keeps one winner and drops the rest; selectTextEditGroups turns that drop
+  // into a whole-finding skip, which is the contract rule.TextEdit states. A
+  // zero-width insert sitting at the end of a prior NON-empty edit is left
+  // alone: it applies cleanly after the replacement and is a legitimate
+  // adjacency.
   lastInsertAt := -1
   for _, edit := range sorted {
     if edit.Pos < lastEnd {
--- a/packages/lint/linthost/rules_format_clause_join.go
+++ b/packages/lint/linthost/rules_format_clause_join.go
@@ -155,11 +155,12 @@ func joinClauseBody(
   if body.Kind == shimast.KindBlock && !target.alwaysJoin {
     return
   }
-  // An empty-statement body (`while (x)\n;`) glues directly to the header with
-  // NO space: Prettier's adjustClause special-cases EmptyStatement and returns
-  // the bare `;` (`while (x);`), only prepending a space when the empty
-  // statement carries a leading comment. This rule's gap->" " rewrite cannot
-  // produce the spaceless `);` glue, so abstain and leave the source shape.
+  // An empty-statement body glues directly to the header with NO space:
+  // Prettier's adjustClause special-cases EmptyStatement and returns the bare
+  // `;` (`while (x);`, `else;`, `do;`, `label:;`), only prepending a space when
+  // the empty statement carries a leading comment. This rule joins with a
+  // single space and deliberately does not take on the spaceless variant, so it
+  // abstains and leaves the source shape rather than emitting `while (x) ;`.
   if body.Kind == shimast.KindEmptyStatement {
     return
   }
@@ -178,12 +179,6 @@ func joinClauseBody(
   }
   anchorStart := gapStart - len(target.anchor)
   if anchorStart < 0 || src[anchorStart:gapStart] != target.anchor {
-    return
-  }
-  // A keyword anchor must be a whole token: `else` may not be the tail of
-  // `myelse`, and `do` may not be the tail of `undo`.
-  if isClauseAnchorWord(target.anchor) &&
-    anchorStart > 0 && isClauseIdentifierByte(src[anchorStart-1]) {
     return
   }
   gap := src[gapStart:bodyStart]
@@ -232,17 +227,12 @@ func isClauseGapByte(c byte) bool {
 }
 
 // isClauseAnchorWord reports whether an anchor is a keyword rather than
-// punctuation, which decides whether it needs an identifier-boundary check and
-// whether it starts the header line the width budget is measured against.
+// punctuation, which decides where the width budget starts measuring. The
+// anchor needs no identifier-boundary check: it is read backward from the AST
+// body's own position, so the token before the gap is the clause's real keyword
+// and an identifier ending in `else` or `do` cannot reach it.
 func isClauseAnchorWord(anchor string) bool {
   return anchor == "else" || anchor == "do"
-}
-
-// isClauseIdentifierByte reports whether `c` can appear inside an identifier,
-// so a keyword anchor is only accepted as a whole token.
-func isClauseIdentifierByte(c byte) bool {
-  return c == '_' || c == '$' ||
-    (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9')
 }
 
 // visualWidth returns the display-column width of `s`: a tab expands to a flat
--- a/packages/lint/linthost/rules_format_clause_join.go
+++ b/packages/lint/linthost/rules_format_clause_join.go
@@ -13,13 +13,20 @@ import (
 //  if (a)
 //    b();
 //
-// becomes `if (a) b();` when the joined line fits printWidth. The same
-// applies to `for`, `for-in`, `for-of`, and `while` headers. A braced
-// body, a body that already shares the header line, a multi-line body,
-// or a join that would overflow printWidth is left untouched.
+// becomes `if (a) b();` when the joined line fits printWidth. The covered set
+// is Prettier's, not the set whose header happens to end in `)`: the `if`
+// consequent and alternate, the `for`, `for-in`, `for-of`, and `while` bodies,
+// the `do` body, the `with` body, and a labeled statement's body. A braced
+// body, a body that already shares the header line, a multi-line body, or a
+// join that would overflow printWidth is left untouched.
 //
-// The rule only ever rewrites the whitespace gap between a header's
-// closing `)` and the controlled statement, so its edits never overlap
+// Two clauses join unconditionally because Prettier gives them no group of
+// their own: an `else` whose alternate is another `if` (the `else if` chain),
+// and a labeled statement, which prints `label: statement` on one line whatever
+// the statement is, braces included.
+//
+// The rule only ever rewrites the whitespace gap between the clause's own
+// header token and the controlled statement, so its edits never overlap
 // `format/indent` (leading whitespace of a line) or `format/print-width`
 // (interior reflow). Idempotent: once joined the gap holds no newline
 // and the rule abstains.
@@ -46,6 +53,9 @@ func (formatClauseJoin) Visits() []shimast.Kind {
     shimast.KindForStatement,
     shimast.KindForInStatement,
     shimast.KindForOfStatement,
+    shimast.KindDoStatement,
+    shimast.KindWithStatement,
+    shimast.KindLabeledStatement,
   }
 }
 
@@ -63,24 +73,67 @@ func (formatClauseJoin) Check(ctx *Context, node *shimast.Node) {
   if opts.TabWidth != nil && *opts.TabWidth > 0 {
     tabWidth = *opts.TabWidth
   }
-  joinClauseBody(ctx, ctx.File.Text(), node, clauseControlledBody(node), printWidth, tabWidth)
+  src := ctx.File.Text()
+  for _, target := range clauseControlledBodies(node) {
+    joinClauseBody(ctx, src, node, target, printWidth, tabWidth)
+  }
 }
 
-// clauseControlledBody returns the single controlled statement of a
-// control-flow header, the `then` branch for `if`, the loop body for
-// the iteration statements. The `else` branch is intentionally excluded:
-// its body is anchored after the `else` keyword rather than a `)`, so it
-// does not share this rule's `)`-anchored join shape.
-func clauseControlledBody(node *shimast.Node) *shimast.Node {
+// clauseJoinTarget is one joinable clause body plus the token that ends the
+// clause header immediately before it. The anchor is what generalizes the rule
+// past the `)`-headed statements: `else` and `do` end in a keyword, a labeled
+// statement ends in `:`, and requiring the exact token still keeps a comment
+// between header and body from being swallowed.
+type clauseJoinTarget struct {
+  body   *shimast.Node
+  anchor string
+  // alwaysJoin marks a clause Prettier keeps on the header line regardless of
+  // the body's own line count or the joined width: an `else if` chain and a
+  // labeled statement.
+  alwaysJoin bool
+}
+
+// clauseControlledBodies returns every clause body of `node` that Prettier
+// keeps on its header line, paired with the token that header ends in. The set
+// is Prettier's, not the AST's: the `if` consequent and alternate, the loop
+// bodies, the `do` body, the `with` body, and a labeled statement's body.
+func clauseControlledBodies(node *shimast.Node) []clauseJoinTarget {
   switch node.Kind {
   case shimast.KindIfStatement:
-    return node.AsIfStatement().ThenStatement
+    stmt := node.AsIfStatement()
+    if stmt == nil {
+      return nil
+    }
+    targets := []clauseJoinTarget{{body: stmt.ThenStatement, anchor: ")"}}
+    if stmt.ElseStatement != nil {
+      targets = append(targets, clauseJoinTarget{
+        body:   stmt.ElseStatement,
+        anchor: "else",
+        // Prettier prints an `else if` chain flat, so the alternate being an
+        // `if` joins even when that nested statement spans lines.
+        alwaysJoin: stmt.ElseStatement.Kind == shimast.KindIfStatement,
+      })
+    }
+    return targets
   case shimast.KindWhileStatement:
-    return node.AsWhileStatement().Statement
+    return []clauseJoinTarget{{body: node.AsWhileStatement().Statement, anchor: ")"}}
   case shimast.KindForStatement:
-    return node.AsForStatement().Statement
+    return []clauseJoinTarget{{body: node.AsForStatement().Statement, anchor: ")"}}
   case shimast.KindForInStatement, shimast.KindForOfStatement:
-    return node.AsForInOrOfStatement().Statement
+    return []clauseJoinTarget{{body: node.AsForInOrOfStatement().Statement, anchor: ")"}}
+  case shimast.KindDoStatement:
+    return []clauseJoinTarget{{body: node.AsDoStatement().Statement, anchor: "do"}}
+  case shimast.KindWithStatement:
+    return []clauseJoinTarget{{body: node.AsWithStatement().Statement, anchor: ")"}}
+  case shimast.KindLabeledStatement:
+    // A label carries no group of its own in Prettier: `label: statement` is
+    // one line whatever the statement is, so the body may be a block and may
+    // span lines.
+    return []clauseJoinTarget{{
+      body:       node.AsLabeledStatement().Statement,
+      anchor:     ":",
+      alwaysJoin: true,
+    }}
   }
   return nil
 }
@@ -89,11 +142,17 @@ func joinClauseBody(
   ctx *Context,
   src string,
   node *shimast.Node,
-  body *shimast.Node,
+  target clauseJoinTarget,
   printWidth int,
   tabWidth int,
 ) {
-  if body == nil || body.Kind == shimast.KindBlock {
+  body := target.body
+  if body == nil {
+    return
+  }
+  // A braced body already owns its own line layout, except behind a label,
+  // where Prettier writes `label: {`.
+  if body.Kind == shimast.KindBlock && !target.alwaysJoin {
     return
   }
   // An empty-statement body (`while (x)\n;`) glues directly to the header with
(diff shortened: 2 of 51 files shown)


samchon (commented):
**Individual Self-Review, commit `5aee6911f` "fix(lint): take the parameter-property modifier mask from the compiler"** (resolves #1131).

What landed: `anyParameterProperty` in `packages/lint/linthost/rules_format_parameter_properties.go` now reads `shimast.ModifierFlagsParameterPropertyModifier` instead of switching on four keywords, plus an `override`-only positive case, a decorated negative twin, and an `override` case in the Prettier conformance corpus.

The reviewer verified the fix against primary evidence rather than the diff: it read `ModifierFlagsParameterPropertyModifier` in the pinned typescript-go module (`Public | Private | Protected | Readonly | Override`, bits 0 through 4, with `Decorator` at bit 15 and `Static` at bit 8 outside the mask, so no over-match), confirmed `Node.ModifierFlags()` returns `ModifierFlagsNone` on a nil modifier list so the deleted nil guard is subsumed, confirmed the JSDoc-derived modifier bits only ever come from `GetEffectiveModifierFlags` which this rule does not call, and ran the pinned Prettier 3.8.3 itself on both fixtures. Prettier breaks the `override`-only constructor and leaves the decorated one inline, so both expectations are the oracle's. It also read Prettier's own predicate (`params.length > 1 && params.some(p => p.type === "TSParameterProperty")`) and confirmed it is the same condition the rule applies.

Four findings, all adjudicated:

**F1, accepted and fixed in `ce2c7d404`.** The commit's own comment named restating the modifier set as the cause, and `isParameterProperty` in `packages/lint/linthost/rules_ts_extra.go` still restated it. That predicate is not incidental: it is the package's shared one, feeding `typescript/parameter-properties`, `typescript/no-useless-constructor`, `unicorn/no-unused-properties`, `unicorn/prevent-abbreviations`, `no-param-reassign`, and the extra-rules group. Its list was complete today, so this was not a live bug, but leaving it would have left the same drift five more chances and would have left `linthost` carrying two definitions of "parameter property". It now reads the mask and `anyParameterProperty` defers to it.

**F2, accepted and fixed in `ce2c7d404`.** `format_parameter_properties_ignores_plain_params_test.go` still described "no accessibility/readonly modifier", the exact four-item framing the commit removed. It is the one-property-away twin of the new positive, so it is where a reader looks.

**F3, accepted and fixed in `ce2c7d404`.** The new decorated negative twin justified itself against a check ("does this parameter carry modifiers?") that neither the old nor the new code ever performed, so the stated reason was a strawman even though the case is worth keeping. Its doc now states the verifiable claim: the mask excludes `ModifierFlagsDecorator`, and Prettier leaves the same constructor inline.

**F4, accepted as a record correction, not actionable.** The commit message says "the two sibling rules in this package already read it". It is two call sites in one rule file (`rules_no_useless_constructor.go`), serving one rule. The other in-repo mask reader is `packages/evidence/native/typescript.go`. The message is already pushed and is not worth a rewrite; this note is the correction.

One reviewer observation deliberately not acted on: the rule fires on an ambient constructor signature where Prettier refuses to format at all. That is invalid TypeScript, `private` already triggered it before this commit, and `override` behaves identically, so it is neither introduced nor in scope here.

One claim in #1131's own evidence did not hold and is corrected here: the issue said `packages/lint/src/structures/rules/ITtscLintTypeScriptRules.ts` and `website/src/content/docs/lint/rules/typescript.mdx` "describe the four-item behavior". They do not. Neither mentions a modifier set for this rule, and `website/src/content/docs/evidence/claims.mdx` already states all five. No documentation change was needed.

Not run by the reviewer: the Go suite, deliberately, because it would have required touching a shared checkout. It hand-traced both fixtures instead. I ran the affected cases through `scripts/test-go-lint.cjs`'s scratch materialization; they pass.

samchon (commented):
**Individual Self-Review, commit `bcc31a1a9` "docs(lint): state the per-finding autofix conflict rule the host enforces"** (resolves #1135).

What landed: the `rule.TextEdit` doc comment and the contributor walkthrough now describe the per-finding atomic applier `selectTextEditGroups` has enforced since #605, plus two tests for contract statements that had no coverage.

The reviewer checked every sentence of the new contract against `packages/lint/linthost/fix.go` and confirmed each: per-finding grouping, earliest-group-first ordering, the whole-group skip and why the `len(kept) == len(candidate)` gate is an exact "nothing was dropped" test, the intra-finding non-overlap consequence, the exact-duplicate collapse, and the re-fire on the next cascade pass through `runFix`'s reload loop. It also confirmed both new tests fail if the behavior they pin is reverted, and that the self-overlap case is not a restatement of its peer: an implementation that checked group members only against the already-selected set passes the peer and fails this one.

Four findings, all adjudicated:

**Finding 1, accepted and fixed in `ce2c7d404`.** Both documents claimed `format/sort-imports` ships multi-edit fixes. It does not: both of its report sites in `rules_format_sort_imports.go` pass a single wide replacement, which makes it the archetype of the shape the new prose argues against. The error came from #1135's own body, where "passes an edit slice" was true only in the trivial one-element sense; the commit upgraded it to a false claim. Both lists now name `typescript/no-import-type-side-effects`, `format/whitespace`, `format/indent`, `unicorn/prevent-abbreviations`, and `unicorn/template-indent`, each of which the reviewer verified at its emit site.

**Finding 2, accepted and fixed in `ce2c7d404`.** The Go doc comment wrapped mid-identifier, so pkg.go.dev would have rendered `no-import-type-side- effects`. The name is also now written in its registered namespaced form, matching every other entry in the list.

**Finding 3, accepted and fixed in `ce2c7d404`.** The walkthrough asserted the skipped finding re-fires on the next cascade pass without the convergence case the Go comment carries, and contradicted its own next bullet, which says a self-overlapping finding can never apply. It is also false when a pass applies nothing and the cascade breaks. The bullet now carries the same hedge as `rule.TextEdit`.

**Finding 4, accepted and fixed in `ce2c7d404`.** `fix.go` still pointed at `rule.TextEdit` for a "keeps one winner and drops the rest" policy that comment no longer states. The sentence now distinguishes the edit-level selector's local behavior from the whole-finding skip the public contract describes.

Two nits accepted with the same commit: the new bullet now uses the `- **Sentence.**` shape its siblings use, and the two documents name the same rule list.

The reviewer confirmed nothing else in the repository still states the old per-edit policy: `packages/lint/README.md` carries no `TextEdit` prose, the `website/src/content/docs/lint/` tree has none, and `rule.Suggestion`'s cross-reference to `TextEdit`'s policy stays true and is better supported by the new text.

samchon (commented):
**Individual Self-Review, commit `8272f3437` "fix(lint): make every published functional option change a decision or say it does not"** (resolves #1132).

The reviewer built its own scratch module, byte-verified it against `git show 8272f3437:packages/lint/linthost/rules_functional.go`, and ran real engine probes. Every finding below is a measurement it took, not a reading of the diff.

**Blocking finding 1, accepted and fixed in `e1040db47`.** `ignoreClass` was marked reserved on the claim that "the native subset visits no class-member position, so class fields are already never reported". Both clauses are false: `functionalPreferReadonlyType.Visits()` registers type-node kinds and `lintFileWalker.walk` dispatches by kind with no scope filter, so a mutable type inside a class body reports like any other. The reviewer measured seven class shapes, all reporting with `ignoreClass: true` and with `"fieldsOnly"`. This is precisely the defect #1132 exists to kill, and the commit had blessed it in the new guard. It is now honored, with `"fieldsOnly"` skipping only field declarations and keeping methods, accessors, and constructor parameters checked, and the website page corrected.

**Blocking finding 2, accepted and fixed in `e1040db47`.** `allowMutableReturnType` listed six declaration kinds. The reviewer measured a call signature, a construct signature, a constructor type, and a get accessor still reporting with the option on, while the package's own `isFunctionLikeKind` already covers accessors and constructors. Same option, same position, different answer depending on how the signature was spelled. The kind list is now every signature that can declare a return type, extracted into `signatureReturnTypeNode`, with a call-signature case added.

**Blocking finding 3, accepted and fixed in `e1040db47`.** `functionalEnclosingReturnTypeText` stopped at four kinds, so a bare `return;` inside a constructor or accessor walked past it to an enclosing function and inherited that function's annotation. Measured: `function outer(): number { class A { constructor() { return; } } return 1; }` reported with `ignoreInferredTypes: true`. The walk now stops at the nearest function-like of any kind through `isFunctionLikeKind`, with a regression case.

**Finding 4, accepted and fixed in `e1040db47`.** The website page was updated for four rules and not for `prefer-readonly-type` or `prefer-immutable-types`, so it still promised what the typings now disclaim. All seven of their option entries now carry either a default marker or the reserved note.

**Finding 5, accepted and fixed in `e1040db47`.** The guard had three holes, two demonstrated by running it. Hole A, a field name colliding with an unrelated Go string literal, is the lower bound the follow-up commit already documented; the reviewer sharpened the reason, since short generic names like `types` and `case` do collide today. Hole B was a parser bug: an interface-level doc comment carrying the reserved sentence was inherited by its first undocumented member. Hole C collapsed the field map by bare name across files, and 21 names are declared in more than one family file, so one rule's decoder exempted every other rule publishing the same key. The sweep now keys per declaration site, records the owning interface in its failure message, and resets the doc association on any intervening line.

**Finding 6, accepted and fixed in `e1040db47`.** `if reserved == 0 { t.Fatalf }` would fail the suite the day every reserved option becomes implementable, which is the outcome the sweep exists to encourage. The marker recognizer is now asserted directly against a synthetic doc, positive and negative.

**Findings 7 through 9, accepted and fixed in `e1040db47`.** `checkTypeLiterals: false` gained its "still checks interfaces" twin, and `allowUndefined` gained the default twin `allowNull` already had.

Verified clean by the reviewer and left alone: default preservation for all nine honored options, including the specific question of whether the new `switch` in `functionalNoReturnVoid.Check` can report where the old `if` did not (it cannot, measured); `isTacitWrapperMemberCallee` against a non-member dot, which `isTacitWrapperText` has already excluded through `isFunctionalMemberChain`; `isFunctionalCollectionTypeNode`; the reserved reasons for `enforcement` and `comparator`, both true; the defensibility of the `ignoreInferredTypes` narrowing; test shape and one-property-away twins across all 16 files; and change integrity, with no existing test, fixture, workflow, or dependency touched.

One reviewer observation not reproduced: it reported three `TestFormatClauseJoin*` failures from the later commit `bee2cfebf`. Those cases pass here at `bee2cfebf` and at every head since, both in isolation and in a full `node scripts/test-go-lint.cjs` run whose only failures are the three known Node-22 config-loader cases. The reviewer's scratch appears to predate the same commit's test-source correction.

`allowLocalMutation` stays reserved. The reviewer called its stated reason true but weak, since an `export`-modifier test would be as syntactic as `no-let`'s `allowInFunctions`. Upstream's "locals" means bindings declared inside a function, not un-exported declarations, and this rule inspects type annotations rather than bindings, so implementing the guess would put an invented semantic behind a published name. Reserved with a truthful reason is the honest state until the upstream oracle is available.

samchon (commented):
**Individual Self-Review, commit `bee2cfebf` "fix(lint): join every clause body Prettier keeps on its header line"** (resolves #1133).

The reviewer built two private scratch modules, one at this commit and one at its parent, drove the pinned Prettier 3.8.3 directly, and probed 27 clause shapes through both the single-rule fixer cascaded to a fixed point and the real `ttsc format` command.

Verified clean: all eight new test expectations and the new conformance corpus case reproduce byte-for-byte under Prettier 3.8.3; the anchor test could not be made to accept a wrong token or reject a right one across `case 1:`, `a ? b : c`, a broken type annotation, an object-literal key, arrow-body labels, and `with` inside a template literal; the two findings a single `if` now emits are disjoint and both survive `selectTextEditGroups`; every probe reached a fixed point in at most two passes; `format/statement-split` and `format/indent` only visit statement-list members, so the new labeled-block shape cannot contend for the same bytes; and the parent commit actively mangled the labeled-loop shape that this commit now formats to Prettier's exact output.

**Finding 2, accepted and fixed in `e1040db47`.** The braced-body gate was pinned only in the direction that joins. Mutating it to let `else`, `do`, and `with` block bodies join left the whole format suite green. A negative twin now covers a braced `else` and a braced `do`.

**Finding 3, accepted and fixed in `e1040db47`.** The keyword anchor's identifier-boundary check is unreachable: the anchor is read backward from the AST body's own position, so the token before the gap is always the clause's real keyword and `myelse` cannot occur. The check and its ASCII-only helper are gone, and the comment now states why no boundary test is needed instead of claiming one that the code could not deliver.

**Finding 4, accepted and fixed in `e1040db47`.** The overlong-body case did not pin what its doc claimed, and printWidth 20 against a 22-column join is not the boundary the testing rules ask for. It now walks the exact limit from both sides. I measured the oracle rather than the code: Prettier 3.8.3 joins `else stopEverything();` at printWidth 22 and breaks it at 21, and the rule agrees on both.

**Finding 5, accepted and fixed in `e1040db47`.** The empty-statement abstention's stated reason was checkable-false. A zero-length replacement is the same gap rewrite and would produce Prettier's `else;`, `do;`, and `label:;` exactly. The comment now says the rule joins with a single space and deliberately does not take on the spaceless variant, and names the labeled form the old comment omitted.

**Finding 1, accepted as a real divergence and published separately as its own issue rather than folded in here.** `alwaysJoin` is the only path that joins a multi-line body, and hoisting its first line leaves the continuation lines at their old column, so `ttsc format` converges on output Prettier would still reindent. The reviewer verified all three witnesses diverge at the parent commit too, and that the owner is `format/indent`, which cedes braceless control bodies and never visits a nested `else` line. It is a different rule and a different invariant from the one this commit closes, and it deserves its own case matrix rather than a tail-end patch. Blocking it behind this commit would also have withheld the clause joins that are correct today.

The minor note that `packages/lint/README.md` still describes clause joining without bounding it is deliberate: #1133 asked for one of the two documents to carry the covered set, and `website/src/content/docs/lint/format.mdx` does. The README is the short-form index and links onward.

samchon (commented):
**Individual Self-Review, commit `e1040db47` "fix(lint): honor ignoreClass, cover every return-type signature, and close the sweep's holes"**, plus the follow-up corrections in `ca3bb78fd` and `a27168f5e`.

**Finding 1, accepted and fixed in `a27168f5e`.** The `ignoreInferredTypes` correction was incomplete. `functionalEnclosingReturnTypeText` was widened to stop at the nearest function-like of any kind, and `functionalReturnTypeText` still read annotations from its own four-kind table, so a get accessor, the one function-like that both stops the walk and may annotate its return type, came back annotation-less. `class A { get value(): number { return; } }` was spared with `ignoreInferredTypes: true` while the identical function-declaration shape reported. Both now read `signatureReturnTypeNode`, the duplicated switch is gone, and a case pins the accessor.

**Finding 2, accepted and fixed in `a27168f5e`.** The braced-body negative twin claimed "a label is the one clause whose braced body Prettier pulls up; every other clause leaves `{` where it is". The reviewer ran the pinned Prettier: it pulls the brace up for `if`, `while`, `else`, `do`, and a label alike. The assertion is still right, but as a scope decision rather than an oracle claim, and the comment now says so.

**Finding 3, accepted and fixed in `a27168f5e`.** `ignoreClass` had no negative twin, and the reviewer showed a position test that answered "in a class" unconditionally passes both positive cases. A module-scope case now fails on that mistake.

**Finding 4, already fixed in `ca3bb78fd`.** A blank line between a doc comment and its property orphaned the doc, which would have reported a reserved field as inert.

**Findings 5 and 6, accepted and fixed in `a27168f5e`.** The `checkImplicit` reason differed between the typings and the website, and "whole class body" overstated an ancestor test that also covers a class's heritage clause and type parameters. Both surfaces now say the same thing.

**Finding 7, accepted and fixed in `a27168f5e`.** The width case asserts both sides of the limit, so its name now says it charges the budget at its limit rather than only that it keeps an overlong body broken.

Verified clean by the reviewer and left alone: `signatureReturnTypeNode` covers all ten legal return-type positions, with `Constructor` and `SetAccessor` correctly excluded as grammar errors and `IndexSignature`'s type correctly not a return type, every `As*` accessor checked against the pinned typescript-go; the `interface{}` comparisons match what `encoding/json` produces, with an unexpected value falling through to off; `isFunctionalClassMemberPosition` returns at the innermost class and classifies a static block, a constructor parameter property, a nested class, and a class expression correctly; the reworked sweep still catches a genuinely inert field, proved by replicating its parser against the real corpus (187 fields, 7 reserved recognized) and injecting both an ordinary inert field and an undocumented member under a marker-carrying interface doc, each flagged; dropping the `reserved == 0` fatal opens no hole, since a reworded marker turns the field inert; the printWidth boundary is Prettier's own; the removed identifier-boundary check was genuinely dead; and all six new tests fail when their behavior is reverted.

The reviewer noted `allowMutableReturnType`'s get-accessor, construct-signature, and constructor-type arms and `"fieldsOnly"`'s accessor and parameter-property arms are covered by one case each rather than all. They share one table and one ancestor test with the arms that are pinned, so the marginal case would pin the table's completeness rather than a distinct branch; the completeness claim is what the reviewer verified directly against the compiler's own node set.

samchon (commented):
**Overall Self-Review, round 1**, over the complete `origin/master...HEAD` diff at `a27168f5e`.

Surface read in full: 6 Go source files (`fix.go`, `rules_format_clause_join.go`, `rules_format_parameter_properties.go`, `rules_functional.go`, `rules_ts_extra.go`, `rule/rule.go`), 1 published typings file, 3 documentation pages, and 33 test files.

**Verified.**

- **Closing set matches what survives.** `git log origin/master..HEAD` carries exactly four `Close #` lines, for #1131, #1132, #1133, and #1135, each with a fix present at `HEAD`. #1134 has no closing line and is deferred with its reasoning recorded on the issue, so the squash merge cannot close it mechanically.
- **Diff boundary.** Every changed path is under `packages/lint` or `website`, which is the declared campaign scope. No workflow, fixture, dependency, or generated baseline outside it was touched. `.wiki` is untracked and stayed out.
- **Generated artifacts.** `go run ./tools/rulecodes` reproduces `packages/lint/linthost/rule_codes.json` byte-for-byte; no rule was added or removed, so the append-only ledger is unchanged.
- **Local evidence.** `node scripts/test-go-lint.cjs` at this head fails only the three config-loader cases that fail identically on unmodified `origin/master` in this environment. The CI `go` lane passes on Node 24, which settles those three as a local Node-22 artifact rather than a defect. `node scripts/ci/format-check.cjs` is clean and `pnpm --filter @ttsc/lint exec tsc --noEmit` is clean.
- **Oracle discipline.** Every format expectation in the diff was taken by running the pinned Prettier 3.8.3, not from what the code emits: the `override` parameter-property break, the `else` / `do` / `with` / labeled joins, the `else if` chain, the comment-in-gap abstention, and the printWidth 22/21 boundary.
- **Cross-rule safety.** A whole-corpus probe over all 726 fixtures under `tests/test-lint/src/cases/` with every AST-only built-in enabled, driven through the production `selectTextEditGroups` applier and re-parsed, produces no new parse diagnostic and reaches a fixed point on every file, before and after this diff.
- **Default preservation.** Nine `functional/*` options became live and every one of them is off by default, so no project's existing diagnostics move. The three pointer-typed options distinguish "absent" from "explicitly false" precisely for that reason, and the two `no-return-void` arms that could have widened the rule fire only on a non-nil, false pointer.

**Applied during this round.**

- `hasAncestorKind` was a second ancestor walk beside the `hasAncestor` the same file already owned; it is gone and the caller uses the existing helper.
- The option sweep's failure message now names the owning interface and declaration site, so a future failure points at one rule rather than a bare field name.
- The sweep's doc-association parser tolerates a blank line between a doc comment and its property.

**Nothing else in this round survived verification.** Every candidate I formed while reading either reproduced as already-correct behavior or turned out to be a documented, deliberate scope decision recorded in `website/src/content/docs/lint/format.mdx`.

Two things this cycle deliberately does not carry, both recorded rather than hidden: #1134, the continuation-keyword placement, is unimplemented and stays accepted for the next cycle; #1139, the continuation indentation of a hoisted multi-line clause body, was found by an individual review of this branch, verified to diverge at the parent commit as well, and published as its own issue because its owner is `format/indent`'s ceded surface rather than the rule this cycle changed.

samchon (commented):
**Individual Self-Review, commit `a27168f5e`**, plus the follow-up `6a52f462c`, and **Overall Self-Review round 2** over the complete `origin/master...HEAD` diff.

The reviewer found no defect in `a27168f5e` and proved the behavior delta exactly: `functionalNoReturnVoid.Visits()` returns five kinds and the engine dispatches strictly by them, so the `switch functionalReturnTypeText(...)` call site is bit-for-bit unchanged by the delegation. The only widened path is the enclosing walk, and intersecting `isFunctionLikeKind`'s seven kinds with the six newly covered by `signatureReturnTypeNode` leaves `GetAccessor` alone. The rule can therefore report more in exactly one shape and never less. It confirmed both new tests fail on revert and are the only cases that catch their respective mistakes, ran the pinned Prettier 3.8.3 to confirm every corrected comment is now true (Prettier does pull a brace up for `if`, `else`, `do`, `while`, `for`, `with`, `try`/`catch`, and a label alike), and re-verified the printWidth 22/21 boundary.

**Accepted and fixed in `6a52f462c`.** The reviewer warned against a plausible future refactor: `signatureReturnTypeNode` looks like a hand-maintained mirror of typescript-go's `FunctionLikeData().Type`, but `IndexSignatureDeclaration` embeds the same base, so swapping to that accessor would silently widen `allowMutableReturnType` to spare an index signature's value type. The three kinds the table omits, a constructor, a set accessor, and an index signature, are exactly the ones whose `Type` field means something else, and the comment now says so. One test doc sentence that a function declaration also satisfies was tightened in the same commit.

**Recorded, not fixed here.** Three typings-versus-website gaps the reviewer swept: `allowMutableReturnType`'s signature-coverage sentence exists only on the website; `ignoreIdentifierPattern` and `ignoreCodePattern` are published on six functional interfaces and appear nowhere under `website/`; and `type-declaration-immutability`'s `identifiers` is named but not described. Adjacent and pre-existing in a file this branch touched, the `functional/immutable-data` headline claims the rule is "Configurable to allow mutation inside constructors or initialization expressions", and `functionalImmutableData.Check` accepts only `ignoreMapsAndSets` plus the shared pattern options. That last one is a false claim in shipped documentation and belongs in the next cycle's discovery, not in a tail-end patch to a cycle whose accepted set it is not part of.

**Overall Self-Review round 2, over the whole diff at `6a52f462c`.** Re-read the full base-to-head change from scratch. Re-confirmed: the closing set is exactly #1131, #1132, #1133, and #1135, each with a surviving fix; every changed path is inside the declared scope; the rule-code ledger regenerates byte-identically; `node scripts/ci/format-check.cjs` is clean; `functionalReturnTypeText` has exactly two callers and both are correct under the delegation; and the local Go suite's only failures are the three config-loader cases that also fail on unmodified `origin/master` here and pass on CI's Node 24. The round found nothing further that survived verification.

Issue #1130, opened 2026-08-08
fix(evidence): a type-only re-export across a module boundary withholds nothing


samchon:
## Problem

A type-only export withholds value-space only when it is written in the module that declares the symbol. The identical intent written across a module boundary withholds nothing, so a barrel re-exporting a class type-only publishes every one of its members as evidence units.

Measured on `@ttsc/evidence` at `af1d511c4`:

| Export form | Members published |
| --- | --- |
| `export type { Sale }` in the declaring module | none |
| `export { type Sale }` in the declaring module | none |
| `export type { Sale } from "./sale.js"` in a barrel | `Sale.prototype.*`, `Sale.*` |
| `export { type Sale } from "./sale.js"` in a barrel | `Sale.prototype.*`, `Sale.*` |

The same table holds for a plain class, for a class merged with an interface, and for a namespace's data and functions. The dividing line is the module specifier rather than the spelling, and there is no semantic justification for that: both forms erase at emit and expose no value.

## Evidence

`collectLocalExportNames` in `packages/evidence/native/typescript.go` skips every export declaration carrying a `ModuleSpecifier`, so `target.TypeOnly` is only ever set by a local export list. `moduleExport` in `packages/evidence/native/exports.go` has no type-only field at all, so `exportDeclarationEntries` records nothing for a re-export and the reachability traversal has nothing to honor.

Reproduction, with a reference selecting the barrel module:

```ts
// src/sale.ts
export class Sale {
  price: number = 0;
  charge(): void {}
}

// src/index.ts
export type { Sale } from "./sale.js";
```

The population holds `Sale`, `Sale.prototype.price`, and `Sale.prototype.charge`. Writing `import { Sale } from "./sale.js"; export type { Sale };` in the same barrel instead yields `Sale` alone.

## Consequence surface

- **False obligations.** A consumer whose public entry re-exports classes type-only owes acknowledgements for every member, and the intent they expressed was that no value crosses that boundary.
- **Documentation cannot state the rule without a caveat.** Five surfaces state type-only projection: the published typings, `website/src/content/docs/evidence/claims.mdx`, `.agents/skills/project/evidence/SKILL.md`, and two Go comments. Each must currently say "among type-only exports the criterion is the module specifier rather than the spelling", which teaches a distinction with no meaning behind it. This is what made the same paragraph reopen across five review rounds on #1124.
- **The class guard and the merged-interface guard share the hole.** Both are correct within a module and both leak across one.
[...]

Issue #1128, opened 2026-08-08
test(evidence): host over-breadth has no negative case for four declaration kinds


samchon:
## Problem

A declaration that registers itself as a claim host under a symbol kind it does not belong to would let a claim silently accept a citation it never asked for, and for four of the six TypeScript declaration kinds nothing in the suite notices.

## Evidence

Host eligibility is registered per declaration in `packages/evidence/native/typescript.go` through `addTypeScriptHost`, and each kind is meant to register exactly the selectors it legitimately hosts. Adding one extra registration to a kind is a one-line change, and for four kinds it leaves the whole Go suite green.

| Mutation | Suite |
| --- | --- |
| interface also registers `"function"` | green |
| interface also registers `"property"` | green |
| type alias also registers `"function"` | green |
| namespace also registers `"function"` | green |
| module-scope variable **declarator** also registers `"type"` | green |
| module-scope variable **statement wrapper** also registers `"type"` | **1 case fires** |
| function declaration also registers `"type"` or `"property"` | **2 cases fire** |
| class also registers `"function"` or `"property"` | **1 case fires each** |
| class member registers a second kind | **3 cases fire** |
| interface property signature also registers `"function"` | **1 case fires** |

So the guard exists for function declarations, for the class, for class members, and for interface property signatures, and is absent for the interface, the type alias, the namespace, and the module-scope variable.

Concrete failing scenario. With the interface registering `"function"`:

```ts
/** @evidence docs/spec.md#contract An interface is not a callable. */
export interface ISale {}
```

under `{"symbol":"function", reference: markdown h2}`, the citation is accepted and discharges the section. Unmutated it is refused with `host kind 'type' is not selected (function)` and the section is reported unacknowledged. No test distinguishes the two.

## Why it matters

The failure is silent in the direction that matters. An over-broad host does not error; it accepts a tag the claim's selector was written to exclude and counts an acknowledgement against a reference. A project that narrowed `symbol` deliberately, which is the whole point of the selector, would have that narrowing quietly stop applying.

This is not a live defect: no shipped code registers a wrong host. It is missing coverage for a class of regression that a one-line edit produces, in a function every new declaration kind has to touch.

## Consequence surface

| Declaration kind | Legitimately hosts | Negative case exists |
| --- | --- | --- |
| interface | `type` | no |
| type alias | `type` | no |
| namespace | `type` | no |
| class | `type` | yes, for both other selectors |
| function declaration | `function` | yes |
[...]

Issue #1126, opened 2026-08-08
fix(evidence): a variable declarator hosts a citation its unit does not know about


samchon:
## Problem

A module-scope variable declarator is registered as a claim host but is not one of its unit's recorded nodes. That single asymmetry produces three separate wrong answers, and all three are silent.

1. `@internal` written on an inner declarator withdraws nothing.
2. A withdrawn variable identity keeps that declarator as a live claim host and exclusion carrier.
3. A citation written on an inner declarator has no semantic host identity, so `singleEvidencePerSymbol` reports the statement's identities as citing zero units while the same run reports the obligation as satisfied.

## Evidence

`collectTypeScriptVariables` in `packages/evidence/native/typescript.go` registers a host on the `VariableDeclaration`:

```go
if hidden == "" {
  addTypeScriptHost(supportedHosts, declaration, symbol)
}
```

and records two other nodes against the unit, the binding identifier through `addTypeScriptUnit` and the statement wrapper explicitly:

```go
// The binding names the unit, but TypeScript attaches a
// variable's leading JSDoc to the statement wrapper, so that
// is where a citation for this unit actually lives.
inventory.recordUnitNode(unit.ID, statement)
```

The declarator is in neither set. `inventory.UnitNodes` is therefore not a superset of the nodes the host map is keyed by, and every consumer that walks from a unit to its declarations misses that one.

### 1. An inner declarator's own tag is never read

`memberHidden := typeScriptHidingTag(file, statement, hidden)` reads the statement wrapper only.

```ts
export const first = 1,
  /**
   * @internal
   */
  second = 2;
```

`second` materializes with no withdrawal tag. It is a selected reference unit and an accepted claim host, and an `@evidence` in that same block discharges an obligation with no diagnostic. The published contract says the opposite: "A declaration whose documentation comment carries `@internal`, `@hidden`, or `@ignore` materializes no unit, and neither does anything nested inside it."

### 2. A withdrawn identity keeps the declarator host

`withdrawHiddenHosts` reconciles the host set against withdrawn identities by walking `inventory.UnitNodes`, so it cannot reach a host node no unit records.

```ts
export namespace N {
  /**
   * @internal
   */
  export var price: number;
}
export namespace N {
  export var other: number,
    /** @evidence docs/spec.md#pricing A withdrawn declarator hosts this. */
    price: number;
}
```

[...]

PR #1124, opened 2026-08-08
feat(evidence): make a class and its member variables citable units

samchon:
## Intent

Make a class a citable unit of the evidence graph, together with its member variables, so an object-oriented project can put an obligation on the subject rather than only on what the subject does.

This pull request owns the complete accepted cycle of the `evidence-class-unit` campaign. Both issues are implemented here; the issue DAG orders the edits inside the branch rather than splitting it into more than one pull request.

| Issue | Title | Order |
| --- | --- | --- |
| #1121 | a class is not a citable unit, and neither are its member variables | first |
| #1123 | a constructor parameter property is a class member variable the graph cannot see | after #1121 |

## Scope

- `packages/evidence/native/typescript.go`: `KindClassDeclaration` registers a `type` unit, a public non-callable member registers a `property` unit, and the class becomes the containment parent of its own members.
- The published selector contract, the `evidence/documented` host population, merged class-and-namespace identity, and type-only projection move with it.
- Go unit tests, the `tests/test-evidence` end-to-end suite, the package typings, the README, the website evidence guides, and the `project/evidence` skill are updated in the same change.

## Deferred

- Enum citability. `KindEnumDeclaration` still materializes nothing. It is a separate design question with its own unanswered sub-decisions, and #1121 scopes itself to classes.
- Constructors and accessors stay unselected, which is the resolution #1121 states for the constructor question.

## Verification

Pending. Local `go test ./native/...` from `packages/evidence`, `pnpm test:go`, `pnpm test`, and the pull request's ordinary CI all run against the settled head; results are recorded as formal reviews on this pull request rather than by editing this body.

No campaign CI is disabled and no check is skipped.

PR #1137, opened 2026-08-09
fix(evidence): make a host position answer for the unit that owns it

samchon:
## Intent

Cycle 2 of the evidence campaign. Its accepted set was published during cycle 1's implementation, after that cycle's discovery gate had closed, so it belongs here rather than to #1124.

| Issue | Subject | Order |
| --- | --- | --- |
| #1126 | a variable declarator hosts a citation its unit does not know about | first |
| #1128 | host over-breadth has no negative case for four declaration kinds | after #1126 |
| #1130 | a type-only re-export across a module boundary withholds nothing | independent |

#1126 lands first because #1128's negative cases include the module-scope variable, and the host set that case pins is the one #1126 changes.

## Scope

- `packages/evidence/native/typescript.go`: the unit-to-node association for a variable declarator, the withdrawal reconciliation that follows it, and the type-only mark across a module boundary.
- `packages/evidence/native/exports.go`: `moduleExport` gains what a re-export must carry for the traversal to honor a type-only edge.
- Negative host-eligibility cases for the interface, the type alias, the namespace, and the module-scope variable.
- The published typings, the README, the website evidence guides, and the `project/evidence` skill move with whatever behavior changes, and the caveat #1130 forced onto five surfaces comes off them.

## Deferred

Nothing yet. Anything this cycle disproves or externally blocks is recorded here and in the issue rather than dropped silently.

## Verification

Pending. Local `go test ./native/...` from `packages/evidence`, `pnpm test:go`, and this pull request's ordinary CI all run against the settled head; results are recorded as formal reviews rather than by editing this body.

No CI is disabled and no check is skipped.


samchon (commented):
## Overall Self-Review — complete round over `origin/master...d574879bf`

Solo, whole-diff, from scratch. **This round produced nothing that survives verification.**

Recording the method correction first, because it governs every round on this pull request. A round is one complete pass over the entire surface that collects every finding, and only then applies them together. Earlier rounds here were run as find-one, fix, commit, restart, which is the whack-a-mole the development skill forbids outright and which is why this branch carries more commits than it should. The last two rounds were run correctly: one pass, findings collected, one repair.

### What the round covered

All fourteen changed files: the collector and its space marking, the export traversal and its mark, the unit model, `documented.go`, five Go test files, the domain model, the published typings, the claims page, and the end-to-end case. Plus the consequence surface each reaches.

### Checks this round ran

- **Traversal arity and memoization.** Every `traverseEntryExports` call site carries the mark. `surfaceOf` starts its inner walk at `false` and the caller combines the edge's own mark with the nested one, so the per-target memo stays path-independent and each branch still applies its own edge. The visited set deletes on exit, so sibling branches both traverse and only cycles are cut.
- **Serialization.** The two new unit fields carry no struct tags, so no cache key or wire format moves with them.
- **Assertion strength, swept across the package rather than the diff.** No case now matches a refusal by selector alone or by carrier phrase alone; the last one outside this branch's files was closed in the same repair, since leaving it was the same cause in a third place.
- **Silence-only assertions.** None remain in any file this branch touches.
- **Doc comment shape** on every new case, and paragraph length on both prose surfaces.
- **Indentation.** Repository-wide check for tab-indented Go files after the reindent this branch had to repair.
- `go vet` clean, working tree clean, format check clean on tracked files.

### State

Fourteen commits, three closing lines matching the three accepted issues, no revert anywhere in the range. `master` advanced with the lint campaign's own cycle and is merged in, so CI runs against the real integration.

### Verifications that could not run locally

The end-to-end suite and every packaging, platform, and downstream lane run only in this pull request's CI. Local evidence is the package's Go suite after every change, `go vet`, and the pinned formatter over every touched file.

Issue #1142, opened 2026-08-09
fix(evidence): a variable unit reports the previous token's line, not its own


samchon:
## Problem

A variable unit reports the line of the token before it, not its own. Every diagnostic that names such a unit points an author at the wrong line whenever trivia separates the declaration from the previous token, which is exactly the two shapes where finding the declaration matters most: an inner declarator carrying its own documentation block, and every leaf of a multi-line destructuring pattern.

## Evidence

Measured on `@ttsc/evidence` at `0b863812f`, reading `unit.Line` after collection.

```ts
 1  export interface ISale {
 2    /** doc */
 3    price: number;
 4  }
 5
 6  /** doc */
 7  export class Sale {
 8    /** doc */
 9    charge(): void {}
10  }
11
12  /** doc */
13  export function draw(): void {}
14
15  /** doc */
16  export const limit = 1;
17
18  export const {
19    gamma,
20    delta,
21  } = source;
```

| Unit | Declared on | Reported |
| --- | --- | --- |
| `ISale` | 1 | 1 |
| `ISale.price` | 3 | 3 |
| `Sale` | 7 | 7 |
| `Sale.prototype.charge` | 9 | 9 |
| `draw` | 13 | 13 |
| `limit` | 16 | 16 |
| `gamma` | 19 | **18** |
| `delta` | 20 | **19** |

And with an inner declarator:

```ts
1
2  export const alpha = 1,
3    /** First wording. */
4    beta = 2;
```

`alpha` reports 2, correctly. `beta` reports **2**, two lines above itself.

The cause is in `lineAtNode`:

```go
func lineAtNode(_ string, node *shimast.Node) int {
  if node == nil {
    return 0
  }
  if name := node.Name(); name != nil && name.End() > 0 {
    return name.End() - 1
  }
  return node.Pos()
}
```

The `Name()` branch is correct and is why every other kind is right: it lands inside the declaration's own name. The fallback is not. A node's `Pos()` is its **full start**, which begins immediately after the previous token and therefore precedes the node's leading trivia. A variable unit is created from the binding identifier, an identifier has no `Name()` child, so a variable unit always takes the fallback.

The fallback happens to be right only when nothing separates the declaration from the previous token: for `export const limit = 1;` the previous token is `const` on the same line. For `gamma` the previous token is `{`, one line up. For `beta` it is the `,` two lines up, with a documentation block in between.

## Consequence surface

[...]

Issue #1141, opened 2026-08-09
fix(evidence): a sibling declarator's edit moves a variable unit's review fingerprint


samchon:
## Problem

A variable unit's digest is computed over the whole variable statement, so it contains every sibling declarator of that statement. Editing one declarator's initializer moves the digest of every other identity the statement declares, and a `requireReview` fingerprint written against any of them expires for a change that did not touch it.

## Evidence

Measured on `@ttsc/evidence` at `0b863812f`, by collecting the same file three times and reading `unit.Digest` for `alpha`:

```ts
export const alpha = 1,
  /** First wording. */
  beta = 2;
```

| Edit | `alpha` digest |
| --- | --- |
| as written | `c261e2ab…` |
| `beta`'s documentation reworded, code untouched | `c261e2ab…` (unchanged) |
| `beta = 2` becomes `beta = 3` | `15706ade…` (**moved**) |

`alpha` is untouched in both edits. The second one moves its digest.

The cause is which nodes the unit records. `typescript.go` computes `unit.Digest = typeScriptUnitDigest(file, inventory.UnitNodes[unit.ID])`, and `collectTypeScriptVariables` records three nodes per variable unit: the binding identifier, the variable statement, and the declarator. Driving the collector directly and printing each recorded node's text shows both identities of the statement above recording

```
typescript::property:5:alpha; <- KindIdentifier          "alpha"
typescript::property:5:alpha; <- KindVariableStatement   "export const alpha = 1,\n  /** First wording. */\n  beta = 2;"
typescript::property:5:alpha; <- KindVariableDeclaration "alpha = 1"
```

The statement wrapper is the whole statement, so `beta`'s text is inside `alpha`'s digest and `alpha`'s is inside `beta`'s. Every other unit kind records exactly one node and it is the declaration that owns the identity's content, so no other kind has this.

The statement is recorded for a real reason and must stay recorded: TypeScript attaches a variable's leading JSDoc to the statement wrapper, so it is a host position, and the withdrawal reconciliation and the semantic-host binding both walk `UnitNodes` to reach it. What is wrong is not that the statement is recorded, but that the digest treats a position as content.

## Consequence surface

- **A review expires for an edit to a different identity.** This is the same mass false-expiry #1118 records for the whole-source Swagger and Prisma digests, at a smaller radius: one statement rather than one document.
- It is worse than a stale review, because the diagnostic names a unit the author did not touch, and the repair is to re-review a declaration that did not change.
[...]

Issue #1138, opened 2026-08-09
fix(evidence): an inner declarator's own documentation block is inside its unit's digest


samchon:
## Problem

A unit's digest is meant to exclude every position a tag can occupy, so that writing a review cannot change the fingerprint that review is checked against. An inner variable declarator's own documentation block is not excluded, and this cycle made that block a real tag position.

## Evidence

Measured on `@ttsc/evidence` at `9f359cf74`. Editing only the comment text, leaving the code identical:

```ts
export const alpha = 1,
  /** First wording. */
  beta = 2;
```

`beta`'s digest moves on every rewording (`f0f5c2ce…` → `18f950b8…` → `ee50be4d…`). `alpha`, whose block would sit above the statement, is stable.

The cause is which node carries the unit's text. `addTypeScriptUnit` records the binding identifier for a variable unit, and `binding.JSDoc(file)` does not report a block attached to the enclosing declarator, so `withoutDocumentationSpans` in `packages/evidence/native/fingerprint.go` has no span to cut. `withoutLeadingTrivia` removes blank lines and `//` comments only. The statement-wrapper path excludes its block correctly, which is why the first declarator of a statement is unaffected.

This predates the cycle: the same values appear before and after the change that records the declarator against its unit. What the cycle changed is the contract it violates. Before, a citation written on an inner declarator resolved to no semantic host, so the position was not one a tag could usefully occupy. Now it is, and the domain model's rule that a digest excludes every such position is false for it.

## Consequence surface

- **`requireReview` cannot terminate on an inner declarator.** An author writes `@evidenceReview <target> #<fingerprint>` in the block; the block is part of the digest; the digest moves; the review is stale the moment it is written. The repair loop the rule is built to avoid is exactly this one, and `markdown.go` and the domain model both name it as the reason documentation spans are excluded.
- Only an inner declarator is affected. A single-declarator statement and the first declarator both take their block on the statement wrapper, which is excluded.
- A destructured pattern's leaves share the declarator, so a block on one reaches every leaf's digest.
- No fixture in this repository hardcodes a fingerprint, so nothing here fails today.

## Approach

[...]

PR #1143, opened 2026-08-09
fix(evidence,lint): variable-unit identity, per-unit review fingerprints, and the formatter clause gaps

samchon:
Cycle 3 of the solo issue campaign. One unified pull request for every accepted issue left unresolved after cycle 2.

## Scope

| Issue | Package | What it repairs |
| --- | --- | --- |
| #1142 | evidence | a variable unit reports the previous token's line, not its own |
| #1138 | evidence | an inner declarator's own documentation block is inside its unit's digest |
| #1141 | evidence | a sibling declarator's edit moves a variable unit's review fingerprint |
| #1118 | evidence | the Swagger and Prisma bridges carry identities, not content, so no per-unit review fingerprint exists |
| #1134 | lint | `ttsc format` never places `else`/`catch`/`finally`/`while` against the closing brace |
| #1139 | lint | a hoisted multi-line clause body keeps its old indentation |

#1138, #1141, and #1142 share one root cause: the collector records positional nodes for a variable unit, and the digest and the reported line each treat one of them as the declaration that owns the identity. They are repaired together.

#1118 lifts the `requireReview` refusal #1115 installed for Swagger and Prisma references.

## Method correction

Cycle 2 recorded #1138 as an issue instead of fixing it inside the cycle, on the grounds that the repair needed a design decision. `development.md` forbids exactly that: difficulty never removes an issue from the cycle, and a duplicate, an invalid premise, an out-of-scope finding, or an external blocker are the only dispositions that do. #1141 and #1142 are the neighbouring symptoms that deferral hid. Every finding this cycle produces is resolved in this pull request.


samchon (commented):
## Individual Self-Review — `3980317f0`

**What landed.** `fix(evidence): let a variable identity answer from its own declarator`, resolving #1138, #1141, and #1142. A variable unit used to be created from its binding identifier while also recording the statement wrapper, and three answers were taken from whichever of those nodes was at hand. `artifactInventory` now separates the two questions: `UnitNodes` keeps every position an identity owns, which the withdrawal reconciliation and the semantic host binding walk, and `UnitContent` holds the nodes whose text is the identity's content, which the digest hashes. `addTypeScriptUnit` takes `named` and `content` separately, and every declaration site states which of its nodes is which. `lineAtNode` treats an identifier as its own name. Nine cases across two new files.

**Verification of the repair, independent of the tests.** Reproducing the pre-change recording and digest source, four of the nine cases fail and pass after; the other five are the deliberate must-stay guards and controls. The reviewer independently dumped every `evidenceDeclaration` and `evidenceReview` across twelve fixtures at the parent and at this commit — inner declarator, block on a binding element, statement block, namespace variables, array pattern, nested pattern, a `var` merge with `@internal`, a mixed function/property statement, an ambient `export declare const`, a renamed export, a parameter property, and an overload run — and every declaration row is byte-identical. Removing the binding identifier from the recorded set cost nothing, because no probed shape attaches documentation to a binding identifier.

**Adjudication.** Ten candidates, nine accepted and applied in `f9909a986`, one recorded as consistent with the package.

| --- | --- | --- |
| 1 | `hostNodesOf`'s comment describes three recorded nodes and a binding that is not a host; the binding is no longer recorded and the filter now removes nothing | Accepted. Rewritten against the withdrawal reconciliation, which is where owning a position and being able to hold a tag genuinely part |
| 2 | `addTypeScriptUnit`'s "a declarator it shares with nothing" contradicts a destructuring pattern, and the commit's own case is named for that sharing | Accepted, corrected |
| 3 | Every variable unit's digest moves, not only the two reported shapes, and nothing records the migration | Accepted. `claims.mdx` and the domain model now state it |
| 4 | The domain model still says a variable's fingerprint moves with the node association | Accepted. Replaced, and the rule this change establishes is now stated |
| 5 | `TestASingleDeclaratorStatementAnswersAsBefore` implies value invariance that the change does not preserve | Accepted, renamed and rewritten to claim only the two behavioral properties |
| 6 | A comment between two declarators changed hands without anyone deciding it | Accepted. Pinned by a case: a `//` comment is leading trivia of the declarator below and reaches neither digest, a `/* */` comment is interior to the declarator below and reaches only that one |
| 7 | A documentation block written between the braces of a destructuring pattern is content, and a tag in it is discarded in silence | Accepted as a defect, published as #1144. Confirmed independently: no node reports that block as documentation, so it is invisible to the tag scan and to the exclusion alike. Making a binding element a host would mean inventing a documentation association the parser does not provide, so the repair is to report the unreadable tag, as the Prisma bridge already does for one buried behind a fourth slash |
| 8 | `lineAtNode`'s remaining fallback answers with a full start, a position outside the node | Accepted. Both branches now answer with a position the declaration occupies |
| 9 | The subset `UnitContent ⊆ UnitNodes` holds only while `UnitNodes` is non-nil | Accepted. Content is appended only when the position record it depends on was taken |
| 10 | Two files hold more than one `Test*` function | Recorded, no change. Every peer in the package does the same; the deviation is a package convention rather than drift |

**Coverage the reviewer confirmed correct.** The subset holds on every path; the semantic host binding, `withdrawHiddenHosts`, and `documentedHosts` answer identically; per-kind content nodes are right for interface, type alias, function, class, namespace, a dotted namespace's inner declaration, class members, parameter properties, interface and type-literal members, and variables; merged identities and overload runs keep every declaration in source order with unchanged digests; the identifier branch of `lineAtNode` is reached only for binding identifiers and every other kind reports what it reported before; no case asserts silence; no expectation is hardcoded; the fixtures compile as real TypeScript.

**Method note.** This review's probes were run in a throwaway worktree. `development.md` line 19 forbids a worktree for a solo campaign or its Self-Review, so subsequent Individual Self-Reviews are instructed to verify in the current checkout without modifying it.

samchon (commented):
## Individual Self-Review — `f9909a986`

**What landed.** `fix(evidence): finish what the variable-identity split changed`, which resolved no issue of its own and applied the adjudicated findings of the review of `3980317f0`: eight prose rewrites, two behavior-neutral code edits (`lineAtNode`'s last fallback, `recordUnitContent`'s guard), and one new case pinning where a comment between two declarators lands.

**Result.** No defect in behavior. The reviewer confirmed `recordUnitContent` is behavior-identical on every path (the position record runs before the new guard, so the `documented.go` inventory still records its content nodes), enumerated all nine `addTypeScriptUnit` call sites to prove no production node reaches the widened fallback, verified that a review naming a namespace containing a variable really does expire by tracing `scopeIndex.collect` through `ParentID`, and confirmed the new case is a genuine pin of `3980317f0`'s behavior rather than a vacuous one.

Eleven findings, all in the account rather than the behavior, all accepted and applied in `b3d9d1264`.

| --- | --- | --- |
| 1 | `hostNodesOf`'s rewritten rationale names the withdrawal reconciliation, and its only caller never runs it: `documentedHosts` skips a withdrawn unit before consulting the host set | Accepted. The filter removes nothing today, so it now states the premise it protects |
| 2 | The digest comment claims leading trivia does not reach the digest, which the same commit's new case disproves for a non-JSDoc block | Accepted, narrowed to blank lines, `//` comments, and text outside the span |
| 3 | `orderIdentityDeclarations` still explains a variable's ordering by a binding identifier that is no longer recorded, and one test doc still says the fallback is reached by more than variables | Accepted, both corrected |
| 4 | The domain model dropped the migration clause and added no replacement, leaving a paragraph whose closing instruction argues against its own contents | Accepted. The migration is stated and the direction restored |
| 5 | The two new domain-model paragraphs were inserted where they severed the referent of the paragraph after them | Accepted, moved below it |
| 6 | `lineAtNode` calls a full start a position outside the node, while the digest reads the node's own text from that offset | Accepted. It lies in the leading trivia rather than among the node's tokens |
| 7 | `node.End()-1` answers a nameless multi-line declaration with its closing line, which is #1142's defect pointing the other way | Accepted. The fallback is restored to what it was, on a branch no unit kind reaches |
| 8 | The website paragraph carried the first em-dashes in that documentation tree | Accepted, removed |
| 9 | The same paragraph narrowed the sibling problem to `const`, which `let` and `var` share | Accepted, widened |
| 10 | The commit message omitted a test rename, misattributed the destructuring case to this commit, and said "both branches" of a three-branch function | Accepted. Recorded here rather than by rewriting a pushed message: the rename is `TestASingleDeclaratorStatementAnswersAsBefore` to `…AnswersTheSameWay`, the destructuring case belongs to `3980317f0`, and `lineAtNode` has three branches of which the last is unchanged |
| 11 | Paragraph shape: one 1,100-character paragraph, a dangling "them", a sentence that states a false rule and retracts it, and a test sentence false for a JSDoc block | Accepted, all rewritten |

**Method note.** This review ran read-only in the current checkout with no worktree, per `development.md` line 19.

samchon (commented):
## Individual Self-Review — `0c586b426` and `b3d9d1264`

Two commits, recorded together because the second's findings were adjudicated while the first's repairs were still landing. Both reviews ran read-only in the current checkout with no worktree, per `development.md` line 19.

### `0c586b426` — `fix(evidence): report a tag written where nothing can read it`, resolving #1144

The reviewer confirmed the parser-aware enumeration is sound: no false positive across nineteen attached-JSDoc shapes, none inside a string, template, regex, or JSX literal, none on prose that merely mentions a tag, correct line numbers through the byte-offset conversion, and health semantics that leave a genuinely failed population still suppressing its derivative findings. It found two real defects and eleven smaller ones.

| --- | --- | --- |
| 1 | **`packages/lint/linthost` no longer compiles.** The move took `commentToken`, `scanCommentGap`, and their helpers out of the host and left three rules and a shared test calling them | Accepted, blocker. `astutil` now exports `CommentToken` and `ScanCommentGap`, the host aliases both. `go vet`/`go test` in `packages/evidence` cannot see this, because evidence imports `rule` and never `linthost`; the repository's own lint Go lane now passes |
| 2 | **A `//` tag above a documented declaration was silent.** A documentation node's reported start is its full start, so its span swallows every comment back to the previous token, and the containment test dropped them | Accepted. The comparison is now against where the block's own text begins. That shape is the likeliest one in a codebase that documents its exports |
| 3 | `///` citations and exclusions were silent while `/// @evidenceReview` was reported, because the review parser strips a triple slash and the declaration parser does not | Accepted. Every run of slashes now comes off |
| 4 | Commented-out documentation was reported, with a repair that is wrong for it | Accepted. A line whose text still opens like a block after the slashes come off is left alone |
| 5 | `readableCommentBody`'s stated premise about the declaration parser was false | Accepted, corrected |
| 6 | The commit message claimed more than the code did | Accepted. Corrected in the repair commit's own message |
| 7 | The domain model was not updated | Accepted. It now states where a TypeScript tag is readable |
| 8 | An em dash in the new `tags.mdx` paragraph | Accepted, removed |
| 9 | `spanIsAttached`'s comment described the mechanism and drew the wrong conclusion from it | Accepted, rewritten with the repair |
| 10 | The dedupe and sort duplicated what `reportProblems` already does | Accepted, reduced to a defined ordering |
| 11 | The repair sentence told a reference-only file to move a tag into a declaration "the claim selects" | Accepted, reworded |
| 12 | Test strength: substring assertions where exactly one diagnostic was available, and no case for the shapes above | Accepted. Three cases added for the repaired shapes |
| 13 | Stale fixture prose calling a line-comment tag invisible | Accepted |
| 14 | `packages/lint/README.md` called `astutil` a re-export | Accepted |
| 15 | Cost: a second walk and scanner pass per file, `spanIsAttached` linear per comment | Recorded, no change. Bounded and cached per source file; a watch rebuild pays only for changed files |

### `b3d9d1264` — `fix(evidence): correct the reasons the last commit gave for itself`

No behavioral defect. The reviewer verified the reverted fallback is byte-identical to `3980317f0` and that nothing depends on the widened form. Fifteen findings, all in prose, all applied.

The three that mattered: the domain model claimed a citation on an inner declarator now discharges an obligation it did not before, when coverage never went through the semantic host and the pre-fix failure was a silent green; the digest comment was widened into a rule whose exception list is still not exhaustive, because a variable's content node is its declarator and everything above it was never in the span; and `hostNodesOf` stated as a global invariant that every recorded position is a host, which a withdrawn identity disproves and `TestWithdrawnDeclaratorIsNotAClaimHost` already pins.

The rest: `lineAtNode` named the `{` as the token every destructuring leaf's full start lands on, true only for the first; `addTypeScriptUnit` missed that two aliases of one local in an export list also share a declarator; `claims.mdx` said "both fingerprints" where the paragraph introduces one; the full-start sentence was left as an unqualified universal; the line case claimed to guard a fallback no unit kind reaches, so it now records that gap; the digest case's comment stopped being true once #1144 made an unreadable tag reported rather than ignored; the two migration notes needed one clause to keep the second from reading as a violation of the first; the website page had no migration note at all for the host-recording change #1137 shipped, which is now written; the em dash finding was already resolved by the repair commit; line endings were verified correct in the committed blob, and the CRLF working copy is normalized; two minor comment-shape nits are fixed.

PR #1145, opened 2026-08-09
test(wasm): report what node holds when the host suite overruns

samchon:
`packages/wasm/test/host` hangs after `PASS` on about 4% of runs and is killed at eleven minutes with nothing to read, because the kill lands on node and node cannot dump goroutines living inside the wasm module. Four occurrences have now turned unrelated pull requests red.

The wasm program runs inside `scripts/go-wasm-exec.cjs`'s own process, so its pending work is node's pending work. A watchdog now reports `process.getActiveResourcesInfo()` and exits when the program overruns.

Measured both directions:

| case | result |
| --- | --- |
| deliberate hang after `PASS` | `go-wasm-exec: node is still holding ["PipeWrap","Timeout"]`, exit 1 |
| healthy program | untouched, exit 0 |

The timer is unreferenced, so it cannot hold open a process that would otherwise close, and the knob is read before the environment is trimmed so it never reaches the wasm program.

**This is evidence capture, not a fix.** #1089 stays open.

### DIFF
--- a/packages/wasm/test/host/main_test.go
+++ b/packages/wasm/test/host/main_test.go
@@ -1,0 +1,84 @@
+//go:build js && wasm
+
+package host_test
+
+import (
+  "fmt"
+  "os"
+  "runtime"
+  "syscall/js"
+  "testing"
+  "time"
+)
+
+// suiteBudget bounds one whole run of this suite, teardown included.
+//
+// Every measured green CI run finished the package in 0.44s-0.93s, so the
+// budget is a ~130x margin that only a wedged runtime can reach. It has to sit
+// well below `go test`'s own kill, because that kill produces no evidence
+// here: the -exec wrapper is node, so the SIGQUIT lands on node and dumps a
+// node process, never the Go goroutines running inside the wasm module.
+const suiteBudget = 120 * time.Second
+
+// TestMain arms a teardown guard that outlives the tests themselves.
+//
+// Three CI runs -- 30661299811 on feat/incremental-graph-snapshots,
+// 30970615435 on feat/evidence-package, and 31080957168 on
+// campaign/evidence-luna-0.24.0 -- ended identically: every test passed, the
+// binary wrote "PASS", and then the process never exited until `go test`
+// killed it at 11m0s, 662s into a package that otherwise takes under a second.
+// testing's own -test.timeout alarm never fired, which places the stall after
+// M.Run stopped that alarm: in the epilogue that writes "PASS" and calls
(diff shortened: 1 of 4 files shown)


samchon (commented):
## Individual Self-Review — `5708e2931`, `c63df8c7b`, `cec8d3b12`

Recorded after merge, for the reason given on #1146: the findings were adjudicated and applied first, the ledger entry second.

**Result.** The review found the first attempt technically safe and substantively empty, which is the more useful verdict.

| --- | --- | --- |
| 1 | The reading could not distinguish the hang from a healthy run. `getActiveResourcesInfo()` returns bare type names, and this suite always writes to stdout and always keeps `host.Expose`'s hourly timer alive, so `["PipeWrap","Timeout"]` is what a healthy run reports at any moment. The one measurement backing the diagnostic came from a deliberate hang that did not reproduce the mechanism under investigation | Accepted. The wrapper now reports per-handle detail; measured on a deliberate hang it reads `Socket(fd=1 pending=0 written=5)`, which tells a wedged pipe from a drained one |
| 2 | Exiting at 120 seconds preempted two strictly better readings: `testing`'s own ten-minute alarm, whose goroutine dump is the inference #1089 is built on, and a Go-side stack dumper written for this exact stall, stranded on an unmerged branch since `bec74bd6c` and arming at the same 120 seconds but always a module load later | Accepted, and it decided the design. The Go-side dumper lands here and is the primary artifact; the node side waits longer so the richer one always wins the race, and `testing`'s alarm becomes redundant rather than lost, since the dumper prints the same stacks eight minutes earlier |
| 3 | The branch shipped with no test | Accepted. Four cases pin the wrapper in both directions, and the ownership gate refused the file until it was claimed, which is the gate working |
| 4 | Every malformed knob value silently disabled the watchdog, and an empty string is one of them | Accepted, refused now. A lane that had lost its only guard read exactly like one that had it |
| 5 | A knob above node's 32-bit millisecond ceiling clamps the delay to 1ms, so a value meant as "effectively never" failed a healthy run instantly | Accepted, refused |
| 6 | `process.stderr.write` before `process.exit` can drop the message on macOS | Accepted. `fs.writeSync`, which is what the Go-side guard chose for a neighbouring reason |
| 7 | An unreachable fallback branch that rendered as if a string were a held resource | Accepted, removed |
| 8 | Prose fusing two independent guarantees into one causal claim, and a measurement window inconsistent with a later occurrence | Accepted |

**Standing note.** #1089 remains open. This is evidence capture, not a repair, and the next occurrence is the first that has a route to a diagnosis.

PR #1146, opened 2026-08-09
fix(evidence): report an unreadable tag only where the graph governs

samchon:
Found by a discovery round over the state merged in #1143.

`evidence/graph` reported an unreadable tag for every scanned TypeScript file rather than for the files a claim or reference glob selects. A base is a directory and a population is a glob inside it, so a claim over `src/**` also spoke about `tools/`, `unrelated/`, and `node_modules/`.

Measured before the fix, with a claim selecting `src/**` only:

| file | reported |
| --- | --- |
| `tools/scratch.ts` | yes |
| `node_modules/vendor/index.ts` | yes |
| `unrelated/legacy.ts` | yes |

A consumer with a stray `// @evidence` in a vendored package failed their build over a file they did not write, and the diagnostic named a repair they could not perform.

The selection is now recorded in the loop that already holds the base a path was reached through and the path relative to it, which is where the Markdown side has compared the base before its globs since it was written. Three cases pin the confinement and one pins that the governed population still speaks.


samchon (commented):
## Individual Self-Review — `e0516ccc7`, `f821582ce`, `f6cca2caf`

Recorded after merge, because the review's findings were adjudicated and applied before the merge but the ledger entry was not written. The obligation is the record, not its timing.

**Result.** The mechanism was confirmed right: recording selection per address key inside the loop that holds both the base and the base-relative path is the correct identity, the write sits deliberately before the already-scanned early-continue, writes are monotonic so nothing depends on map iteration, and the three confinement cases fail on `master`. Four findings, two of them behavioral, all applied in `f6cca2caf`.

| --- | --- | --- |
| 1 | A discovery round's scratch file was committed: three `TestRound*` functions that assert nothing and pass unconditionally, one probing another branch's behavior | Accepted, deleted. It was swept in by a `git add -A` and never mentioned in the message |
| 2 | The website page still said a tag inside commented-out documentation is left alone, which the code, the domain model, and a case have contradicted since #1143 — and this pull request appended to that very sentence | Accepted, corrected |
| 3 | Governance was read from the configuration *after* activation, so a file whose every declaration an author commented out produced no unit, deactivated its claim, left the population, and went unreported. A regression against `master` and an asymmetry with the Markdown loader, which has always reported before activation | Accepted. Selection is taken from the declared configuration in a pass of its own |
| 4 | A package reference's globs resolve against the installed package, not the project, and were matched against project-relative paths, so `**/*.ts` claimed `node_modules` — the one file set this confinement exists to release | Accepted. A package reference governs nothing here |
| 5 | One value carried three names across one data flow, and the godoc block had two summary sentences | Accepted |
| 6 | The domain-model sentence generalized a TypeScript-only mechanism | Accepted |

Both behavioral findings are pinned: a deactivated claim still governs what it declared, and a package reference governs no project file.

Issue #1144, opened 2026-08-09
fix(evidence): a citation in a comment the parser attaches to nothing is discarded in silence


samchon:
## Problem

A citation written in a TypeScript comment that the parser attaches to no node is discarded without a word. The rule exists to remove silence, and this is the quietest failure it has: the comment is real, the file keeps it, and an author reading the source sees a citation that does nothing.

## Evidence

Measured on `@ttsc/evidence` at `3980317f0`.

```ts
declare const source: { gamma: number; delta: number };
export const {
  /** @evidence docs/spec.md#pricing A tag between the braces. */
  gamma,
  delta,
} = source;
```

Evaluating a `symbol: "property"` claim over that file produces exactly one diagnostic, and it is about something else:

```
Missing acknowledgement for 'docs/spec.md#pricing' … Use @evidence on a selected typescript host or @evidenceExclude on an eligible carrier.
```

Nothing names the citation the author wrote. Walking every node of that file and printing `node.JSDoc(file)` returns **no node at all** for that block: TypeScript attaches no documentation to a binding element. The collector reads tags only from blocks a node reports, so the block is invisible to the declaration scan, and invisible in the same way to `withoutDocumentationSpans`, which is why it is also content rather than an excluded tag position.

A destructuring pattern is the shape that makes it easy to hit, not the boundary of the defect. Any comment the parser leaves unattached carries the same silence.

## Consequence surface

- **An author believes an obligation is discharged when it is not.** The coverage diagnostic that follows names the reference, not the ignored citation, so the repair the message suggests is to write another citation — which the author already wrote.
- The same silence swallows `@evidenceExclude`, where it is worse: the reason field makes an exclusion read as a reviewed decision, and neither the decision nor its absence is reported.
- It also swallows `@evidenceReview`, so a review can be written where it can never expire and never satisfy anything.
- `evidence/todo` and `evidence/documented` read the same attached-block set, so an unattached tag is absent from every ledger as well.
- Prisma already refuses this class rather than ignoring it: `prismaBuriedTagLines` reports a tag buried behind a fourth slash, naming the move that fixes it. TypeScript has no counterpart.

## Approach

`collectTypeScriptDeclarations` already knows every documentation block the parser attached, keyed by span. Scan the source text for the tag tokens outside that set and report each one, naming the position and the move that fixes it — the same shape and the same wording discipline as the Prisma buried-tag diagnostic.

[...]

PR #1148, opened 2026-08-09
fix(evidence): report a Markdown tag written as prose

### LINKED ISSUE #1147: fix(evidence): a Markdown citation written as prose is discarded in silence
## Problem

A Markdown citation written as prose rather than inside an HTML comment is discarded without a word. TypeScript reports such a tag since #1144 and the Prisma bridge has always reported its own unreadable placements; Markdown is the one artifact kind left silent.

## Evidence

Measured on `@ttsc/evidence` at `d2accbcc6`.

`docs/claim/plan.md`:

```md
## Plan {#plan}

@evidence docs/spec/rules.md#pricing Written as prose, not in a comment.
```

Evaluating a Markdown claim over `docs/claim/**` against a Markdown reference over `docs/spec/**` produces exactly one diagnostic, and it is about something else:

```
Missing acknowledgement for 'docs/spec/rules.md#pricing' … Use @evidence on a selected markdown host or @evidenceExclude on an eligible carrier.
```

Nothing names the citation the author wrote. The scan reads declarations only from `<!-- -->` spans, so a prose line carrying the tag reaches no host, and the coverage diagnostic that follows tells the author to write the citation they already wrote.

## Consequence surface

- **An author believes an obligation is discharged when it is not**, and the message they get names the repair they already performed.
- The same silence swallows `@evidenceExclude`, where it is worse: the reason field makes an exclusion read as a reviewed decision, and neither the decision nor its absence is reported.
- It swallows `@evidenceReview` too, so a review can be written where it can never expire and never satisfy anything.
- Markdown is the artifact kind whose authors are least likely to know the comment requirement, because the tag renders invisibly either way in the source they are reading.

## What must stay silent

A fenced block is an example, not a citation, and it is already correct: a fenced `<!-- @evidence ... -->` discharges nothing, verified. The evidence documentation itself shows tags inside fences, so reporting them would make the product's own docs fail. An indented code block is the same case in another spelling.

## Approach

The Markdown scan already tracks, per line, whether it is fenced and whether it lies inside an HTML comment. Report a tag opening a line that is neither. Word it as the other two kinds do: name the position and the move that fixes it.

Require the tag to open its line, which is the discrimination every other reader in this package already performs, so prose that merely mentions a tag mid-sentence is not reported.

- A citation, an exclusion, and a review written as prose are each reported at their own line.
- A tag inside a fenced block is not reported, in both fence spellings.
- A tag inside an indented code block is not reported.
- A tag inside an HTML comment is unaffected, including a multi-line comment.
- Prose that mentions a tag mid-sentence is not reported.
- A document with no tags produces no new diagnostic.
- Narrow: `go test ./native/...` from `packages/evidence`. Broader: the `evidence` CI lane.

## Coordination

Found by the discovery round that followed #1143. Same class as #1144, which repaired the TypeScript half; this is the Markdown half.

samchon:
Found by the discovery round that followed #1143. Same class as #1144, which repaired the TypeScript half.

A Markdown declaration is read from an HTML comment. Written as prose the tag reached no host and was discarded without a word, so the only diagnostic was the unrelated coverage finding telling the author to write the citation they had already written.

An example stays silent, which is not a concession: this product's own documentation shows tags inside fenced blocks. A fenced block in either spelling, an indented block, and a mention mid-sentence each have a case.

Close #1147


samchon (commented):
## Individual Self-Review — `74b22197f` and `906d32025`

**What landed.** `fix(evidence): report a Markdown tag written as prose`, resolving #1147, and the correction batch that followed this review.

**Result.** No defect in the core mechanism. The reviewer confirmed all three positive cases fail against `master`, six of eight hand-built mutants of the reporter are killed by the new file alone, the multi-line-comment case pins the `inHTMLComment` continuation rather than any comment, emission is confined to configured globs with no duplication across the two loader passes, and — the measurement that mattered most — **the reporter fires on zero of this repository's 242 `.md`/`.mdx` files**, including all 23 that carry `@evidence`, all six `website/src/content/docs/evidence/*.mdx`, and all 28 `.agents/skills` documents.

Eight findings, all accepted and applied in `906d32025`.

| --- | --- | --- |
| 1 | A citation written as a bullet, a numbered item, or inside a quote still reproduced #1147 verbatim, and both documents framed their exceptions as exhaustive | Accepted. The line is read without the markers that merely carry it, and the tag still has to open what remains. Five spellings pinned, with the mid-sentence row as the twin the marker rule must not break |
| 2 | An example that renders as code without being a fence was reported: an MDX component handed a template literal, and an HTML `<pre>`. The named repair would delete the example from the rendered page | Accepted. Both edges are recognized, and a case pins each |
| 3 | A review was told to write a reason, which a review does not carry | Accepted. It now asks for what was checked, as every other review diagnostic in the package does |
| 4 | The domain-model paragraph sat above the question it answers | Accepted, moved below it |
| 5 | The commit message named two conditions the scan already tracked and omitted the third it hand-rolls | Accepted. Recorded here rather than by rewriting a pushed message: the indentation check is new, and the fenced and comment checks are the pre-existing ones |
| 6 | Two behaviors the test file left unpinned: a prose tag after a closed fence, and every tag in one document rather than only the first | Accepted, both pinned |
| 7 | A reference-only population is reported with a repair that changes nothing there | Recorded, no change. The reviewer verified the TypeScript half behaves identically, so it is a pre-existing shared property rather than something this change introduced |
| 8 | An unterminated comment or fence swallows everything after it | Recorded, no change. Both follow from the pre-existing scanner state, and in the unterminated-comment case the real declaration scan is equally blind, so no reader disagrees with another |

**Method note.** The review ran read-only with no worktree. Mid-review the shared checkout was switched by this agent to another branch; the reviewer finished by pinning the reviewed tree through a `go test -overlay` map from `origin/fix/evidence-markdown-prose-tag`, so every result above was produced against the reviewed state. That interference was mine, and later reviews should be started against a branch this agent is not about to leave.

PR #1149, opened 2026-08-09
test(wasm): read both sides of the stall from one stop

samchon:
Follow-up to #1145, which cannot finish the diagnosis it was written for.

#1089 has two open causes — node never fired the write completion, or fired it and the runtime failed to route it. Both park the main goroutine in the same place, a channel receive inside `syscall.fsCall`, so the goroutine stacks read identically. The discriminator is whether node still holds a pending file request.

The node wrapper takes that reading and never gets to. Whichever guard fires first ends the process, and the Go guard's budget is half the wrapper's by design, so the wrapper's reading is structurally unreachable on every occurrence the Go guard catches. The two halves were mutually exclusive.

The Go guard now asks node directly, through a synchronous call in the sense `fs.writeSync` is, so it cannot depend on the delivery already under suspicion.

Measured by firing the guard deliberately:

| state | reading |
| --- | --- |
| nothing in flight | `getActiveResourcesInfo=[]` |
| a write outstanding | `["FSReqCallback","Timeout"]`, request visible through `_getActiveRequests` |

Still evidence capture, not a repair. #1089 stays open.

### DIFF
--- a/packages/wasm/test/host/main_test.go
+++ b/packages/wasm/test/host/main_test.go
@@ -59,11 +59,11 @@ func reportStallAndExit() {
   buffer := make([]byte, 1<<20)
   buffer = buffer[:runtime.Stack(buffer, true)]
   // The stacks go out on their own write, before anything else is attempted.
-  // Reading the JS side means calling into node, and a JS exception crossing
-  // back through syscall/js is a panic; composing both halves into one message
-  // would let that panic destroy the half already in hand, and the runtime's
-  // own panic output travels through os.Stderr, the asynchronous write path
-  // this guard exists because it is already suspect.
+  // Reading the JS side means calling into node, and the runtime does not
+  // catch every way that can end: a throwing property accessor reached through
+  // `Value.Get` leaves as an uncaught JS exception and takes the process with
+  // it, printing nothing from Go. Composing both halves into one message would
+  // let that destroy the half already in hand.
   writeStderr(fmt.Sprintf(
     "\nwasm host suite: no exit within %s.\n"+
       "Every goroutine stack follows, and what node was still holding after\n"+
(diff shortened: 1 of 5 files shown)


samchon (commented):
## Individual Self-Review — `22da87f9d`, `281e51316`, `fdcdf34dc`

**What landed.** The Go-side stall guard now takes the node-side reading itself, because the two guards were mutually exclusive: whichever fires first ends the process, and the Go budget is half the wrapper's by design, so the wrapper's reading never happened on an occurrence the Go guard caught.

**Result.** The shape was confirmed sound: no effect on a healthy run, `go vet` clean, the suite green through its real runner, the three JS bindings synchronous so none can re-enter Go or park the caller, and the budget ordering intact because no workflow sets `TTSC_WASM_EXEC_TIMEOUT`. The reviewer independently confirmed the discriminator is real: a pending `fs.write` surfaces as `FSReqCallback` and is absent once the completion fires.

Two findings were mine, found while the review ran, and are already in `281e51316`. Nine came from the review, all accepted, applied in `fdcdf34dc`.

| --- | --- | --- |
| — | The reading was an argument to the `Sprintf` that composes the report, so it ran before a byte was written and a failure destroyed the stacks already in hand | Found before the review returned. Stacks written first, on their own call |
| 1 | The recover covers less than its comment claimed: `Value.Call` routes a JS throw back as a panic, but `Value.Get` is a bare `Reflect.get` with no catch, so a throwing accessor leaves as an uncaught JS exception and ends the process without reaching Go | Accepted. The comment says so, and it is the real reason the stacks go first |
| 2 | The recover threw away the reading that matters: `getActiveResourcesInfo` is the discriminator and is taken first, and a failure in the two internals after it replaced the whole answer | Accepted. Readings accumulate and a failure appends itself. Fixing this exposed a second defect of my own — a plain return is computed before the defer runs, so the reading came back empty until it became a named return, which the new case caught immediately |
| 3 | `[]` meant both "node holds nothing" and "the reading did not work", and the first is the affirmative verdict for one of the two causes | Accepted. A list that is not a list says so |
| 4 | `scripts/go-wasm-exec.cjs` asserted the opposite of this change, that the type-name summary distinguishes nothing | Accepted. That was measured against the wrapper's own test stub, which writes through `process.stdout` and materializes a socket; the wasm program writes through `fs.write` and never creates one |
| 5 | The commit message's "which is all of them" overreaches, and quotes an output format the code does not print | Accepted, recorded here rather than by rewriting a pushed message. The code comment was already hedged correctly, and the wrapper covers exactly the occurrences the Go guard cannot |
| 6 | The stated reason for writing the stacks first was wrong: a js/wasm panic prints through the synchronous `fs.writeSync`, not an asynchronous `os.Stderr` | Accepted. The reorder is right for finding 1's reason |
| 7 | Two of the three readings duplicate the first at this observation point, because the pending file request carries none of the detail fields | Accepted as a note. They are read anyway because a held socket would carry them, and a socket appearing here at all would itself be the finding |
| 8 | No automated case, while the wrapper it replaces got four | Accepted. Two cases, neither of which makes the guard fire, so neither can hang the suite it protects |
| 9 | One function without a doc comment where every other has one | Accepted |

**Standing note.** This remains evidence capture. #1089 stays open, and the next occurrence now reads in one shot: `FSReqCallback` present means node never fired the completion and the repair is below Go; absent means the runtime failed to route it, and the event machinery becomes the place to look.

samchon (commented):
## Individual Self-Review — `f472cf7f2`, `66925a916`, `de83f1068`

The review was scoped to behavior for this round, since three earlier rounds had already swept the prose. It returned two blocking findings, both real, both applied in `de83f1068`.

| --- | --- | --- |
| 1 | The new `summary == "[]"` assertion asserts something the system does not guarantee, and encodes the wrong diagnosis. `[]` is the honest reading of an empty loop, is what this file says two functions above, and is the likeliest thing a real occurrence prints — the affirmative verdict for one of the two causes. Measured green in the full suite and red 5 of 5 alone, because the green depended on an earlier heavyweight case leaving a request in flight | Accepted. The case asserts only that a list came back, which no ambient state can move. Calling `[]` broken would have sent whoever reads the dump to discard the answer the guard exists to produce |
| 2 | The two doc blocks were misattached: a blank line left the first floating, so the guard's own contract documented a test-only helper and the function the guard calls had none | Accepted, both reattached |
| 3 | The commit message overstated the split: it moved the per-handle exposure off one case, not off the suite | Accepted, corrected in the message of the fix. Two other cases call the full reading on the healthy path deliberately, because what they pin lives in it |
| 4 | The gate is correct but not total: a Proxy over an array passes `Array.isArray` and then dies uncatchably in `Length`. Not a regression — the previous gate admitted the same shape and died the same way, and node's own APIs cannot produce it | Recorded, no change |
| 5 | `TestStallReadingNamesWhatNodeHolds` is a forward guard, not a regression pin; `TestStallReadingRefusesAForgedVerdict` is the pin that makes "reverting the gate fails the suite" true | Recorded |

**Two defects I found myself while the review ran**, both already fixed:

- `describeJSList` handed every entry to `describeJSHandle`, which asks for a constructor; asking anything that is not an object panics, and a panic there costs the readings taken after it. Fixing it, I dropped the string branch and rendered the discriminator as `[string, string]` where `[FSReqCallback, Timeout]` belongs — the diagnosis erased. The probe caught that in the same minute.
- Hardening the element step removed the panic `TestStallReadingSurvivesABrokenSource` was pinning, so the suite went red and I pushed it before reading the result. The case now breaks its source by throwing, which is the path that still ends in a recoverable panic, and it is verified in both directions.

**Where this leaves #1089.** Still open, still evidence capture. The guard writes every goroutine stack on its own call, then asks node what it holds, and the two halves separate the causes: `FSReqCallback` present means node never fired the completion and the repair is below Go; absent, read together with the main goroutine parked in `syscall.fsCall`, means the runtime failed to route it. Local reproduction was attempted and failed — 108 runs, sequential and under load, against a recorded 3.7%.

PR #1154, opened 2026-08-10
fix(ttsc): honor removeComments in PrintFileWithSourceMap

xiamu33:


samchon:
Thanks for this, and sorry for how it went.

The one-line `RemoveComments` forward you identified was right. It reached `master` ahead of this PR through #1158, which widened the same construction to every `PrinterOptions` field the pinned emitter sets — that work should have been built on top of your PR instead of alongside it, and it left yours conflicting. That was our mistake, not yours.

What is still missing on `master` is your test. `TestEmitWithPluginTransformerHonorsRemoveComments` pins something the parity test that landed with #1158 does not name: that with `removeComments: true` and `sourceMap: true` together, the authored comment goes and the `sourceMappingURL` trailer stays. That interaction is exactly where the bug lived.

So we rebased this branch onto `master` (thank you for leaving "allow edits by maintainers" on). The commit is unchanged in authorship and now carries the test alone, since the code change is already in. Verified locally against the current tree:

```
=== RUN   TestEmitWithPluginTransformerHonorsRemoveComments
--- PASS: TestEmitWithPluginTransformerHonorsRemoveComments (0.10s)
```

Merging once the checks finish.

Issue #1162, opened 2026-08-10
fix(lint): ttsc format normalizes no token spacing, while claiming to replace Prettier


samchon:
## Problem

`ttsc format` normalizes no token spacing. Not around `=`, not around binary or
comparison operators, not after `if`/before `{`, not around a type annotation's `:`,
not inside a parameter list, not around `=>`. The format guide opens with

> This is what replaces Prettier in a `ttsc` project.

and a formatter that leaves `if(x){` and `a=1` exactly as written does not replace
Prettier for ordinary hand-written input. Nothing in the guide says it will not: the
only exclusion list it carries ("What it does NOT touch (yet)") is scoped to the
`printWidth` rule and lists node *shapes* — JSX, binary expressions, destructuring,
decorators, template literals — not token spacing.

The sharpest case is not an abstention but a half-format. `format/bracket-spacing`
fires alone on a type literal and produces a shape neither the author wrote nor
Prettier emits:

```
input     export type J = {alpha:number;bravo:string};
prettier  export type J = { alpha: number; bravo: string };
ttsc      export type J = { alpha:number;bravo:string };
```

**Affected users:** anyone who adopts `ttsc format` as their Prettier replacement and
runs it over code that was not already Prettier-formatted.

## Evidence

Ten single-purpose files, formatted twice: once by `ttsc format` with
`format: { severity: "error", jsDoc: false }` (Prettier defaults), once by Prettier
3.8.3 with `printWidth: 80, semi: true, tabWidth: 2, trailingComma: "all",
parser: "typescript"` and no plugins. `ttsc format` exited 0 in every case.

| case | input | Prettier | `ttsc format` | verdict |
| --- | --- | --- | --- | --- |
| assignment | `export const a   =  1;` | `export const a = 1;` | unchanged | no-op |
| binary | `const b = 1+2;` / `const c = 1   +   2;` | `1 + 2` for both | unchanged | no-op |
| comparison | `const d = 1===1;` | `1 === 1` | unchanged | no-op |
| call | `run (1);` | `run(1);` | unchanged | no-op |
| keyword | `if(x > 0){` | `if (x > 0) {` | unchanged | no-op |
| parameters | `g( a : number , b : number )` | `g(a: number, b: number)` | unchanged | no-op |
| arrow | `(x: number)=>x*2` | `(x: number) => x * 2` | unchanged | no-op |
| annotation | `const i : number = 1;` | `const i: number = 1;` | unchanged | no-op |
[...]

Issue #1161, opened 2026-08-10
fix(ttsc): a source preamble shifts every diagnostic the plugin lane reports


samchon:
## Problem

When a plugin injects a source preamble, every TypeScript diagnostic the plugin lane
reports carries a line number shifted down by the preamble's height, and the user is
shown the same error twice with two contradictory positions.

`@ttsc/banner` is the first-party plugin that does this. On a project whose only error
is on **line 5**:

```
$ ttsc check           # no plugins
src/index.ts(5,14): error TS2322: Type 'string' is not assignable to type 'number'.
# exit 1
```

```
$ ttsc check           # the same project with @ttsc/banner
src/index.ts:11:14 - error TS2322: Type 'string' is not assignable to type 'number'.

11 export const bad: number = "not a number";
                ~~~

Found 1 error in src/index.ts:11

src/index.ts(5,14): error TS2322: Type 'string' is not assignable to type 'number'.
# exit 2
```

Line 11 does not exist in `src/index.ts`; the file is six lines long. An editor,
a CI annotation, or a human following the first block lands on the wrong line — or on
no line at all.

**Affected users:** every project whose plugin chain contains a
`driver.SourcePreamblePlugin`. `@ttsc/banner` is the shipped one; any third-party
plugin implementing that interface has the same effect.

### Reproduction

`src/index.ts` (6 lines):

```ts
export interface IUser {
  id: string;
}

export const bad: number = "not a number";
```

`tsconfig.json` with `compilerOptions.plugins: [{ "transform": "@ttsc/banner" }]` and a
`banner.config.json` carrying any `text`. Run `ttsc check` with and without the plugin
entry; the transcript above is verbatim (ANSI stripped).

The offset is exactly the preamble's height, measured by varying the banner and
counting the lines it prepends to the emitted output:

| `banner.config.json` `text` | preamble lines | reported line | source line | shift |
| --- | --- | --- | --- | --- |
| one line | 6 | 11 | 5 | 6 |
| three lines | 8 | 13 | 5 | 8 |

### Mechanism

1. `driver.LoadProgram` wraps the filesystem in `sourcePreambleFS`
   (`packages/ttsc/driver/program.go:360-364`), so tsgo parses every non-declaration
   source with the preamble prepended. Every position tsgo records is shifted down by
   the preamble's line count.
2. The native host prints those positions verbatim:
   `driver.WritePrettyDiagnostics(opts.stderr, prog.Diagnostics(), opts.cwd)`
[...]

Issue #1160, opened 2026-08-10
fix(experimental): the release rehearsal packs eight of thirteen published packages


samchon:
## Problem

`pnpm package:tgz` is the release rehearsal — the flow the shim-sync skill tells a
contributor to run before trusting a change in a real consumer, and the one
`typia.yml` / `nestia.yml` drive in their current-platform form. Its own header says
what it is for:

> default — used by `pnpm package:tgz` for release rehearsals. Builds **every
> workspace tarball** + every platform tarball (~15-20 min on CI).

It does not. `experimental/tarballs/index.ts` packs a hand-written list of eight
non-platform packages, while `pnpm package:latest` publishes thirteen:

| published (non-platform) | packed by the full rehearsal? |
| --- | --- |
| `ttsc`, `@ttsc/banner`, `@ttsc/lint`, `@ttsc/paths`, `@ttsc/strip`, `@ttsc/unplugin`, `@ttsc/vscode`, `@ttsc/wasm` | yes |
| **`@ttsc/evidence`** | **no** |
| **`@ttsc/factory`** | **no** |
| **`@ttsc/graph`** | **no** |
| **`@ttsc/metro`** | **no** |
| **`@ttsc/playground`** | **no** |

A packaging fault in one of those five — a `files` entry that misses a directory, an
`exports` target that does not exist in the tarball, a package published before its
`lib` is built — is not exercised by the rehearsal and reaches
`pnpm run package:latest:publish` unrehearsed.

## Evidence

- The list: `experimental/tarballs/index.ts`, `listTargets`

  ```ts
  const corePackages = CURRENT_ONLY
    ? ["ttsc", "banner", "lint", "paths", "strip", "unplugin"]
    : ["ttsc", "banner", "lint", "paths", "strip", "unplugin", "vscode", "wasm"];
  ```

- The publish set: root `package.json`

  ```
  "package:latest:publish": "pnpm run assert:platform-packages &&
     pnpm --filter=./packages/* --filter=!ttsc -r publish … &&
     pnpm --filter ttsc publish …"
  ```

  which is every non-`private` package under `packages/`.

- Enumerating `packages/*/package.json` for `private !== true` yields the thirteen
  non-platform names in the table above.

- The list has not been revisited since `580a375e8` (`ci(experimental): scope current
  tarball builds`, #332). `fb8faac86` touched the file afterwards but only changed the
[...]

Issue #1159, opened 2026-08-10
test(lint): nothing verifies the website anchors the rule documentation links carry


samchon:
## Problem

`@ttsc/lint` ships a `codeDescription.href` with every diagnostic whose rule belongs to
a website-documented family, and that URL carries a `#`-anchor pointing at a level-3
heading on `https://ttsc.dev/docs/lint/rules/<family>`. Nothing in the repository reads
those pages, so the day a rule joins a documented family without its heading, every
editor gets a deep link to an anchor that does not exist — and the whole test suite
stays green.

The contract is written down. `packages/lint/linthost/rule_docs.go` states it as part
of the definition of `websiteDocumentedRuleFamilies`:

> lists the families whose whole rule set is catalogued on the ttsc website, as one
> level-3 heading per rule (the rule name in a code span) in
> `website/src/content/docs/lint/rules/<family>.mdx`.

`TestRuleDocumentationURLsCoverTheBuiltinCorpus`
(`packages/lint/test/registry/rule_documentation_urls_cover_the_builtin_corpus_test.go`)
proves the *derivation* — that every built-in rule maps to a well-formed URL with a
correctly slugged anchor — and explicitly stops there. It never opens an `.mdx` file.

**Affected users:** every editor user who follows a rule's documentation link.

## Evidence

- Derivation: `ruleDocumentationURL` (`rule_docs.go:62`) returns
  `websiteRuleDocsBaseURL + family + "#" + websiteRuleAnchor(name)` for any family in
  `websiteDocumentedRuleFamilies`; `lspCodeDescriptionForRule` (`rule_docs.go:108`)
  puts it on the wire.
- The anchor transform (`websiteRuleAnchor`) reproduces github-slugger:
  `jsx-a11y/alt-text` → `#jsx-a11yalt-text`.
- Coverage: the only test over this surface asserts URL *shape*. Grepping the
  repository for a check that reads `website/src/content/docs/lint/rules/*.mdx`
  against the registry finds none.

### Current state — measured, and the invariant holds today

Cross-checking every website-documented family's `### <rule>` headings against the
rule-code ledger, all 18 families match:

```
boundaries 6/6, cypress 13/13, functional 20/20, jest 19/19, jsdoc 13/13,
jsx-a11y 37/37, nextjs 21/21, playwright 31/31, promise 17/17, react 29/29,
react-perf 4/4, regexp 22/22, security 14/14, solid 20/21*, storybook 16/16,
[...]

PR #1163, opened 2026-08-10
fix: the full-scope 0.26.1 campaign cycle

samchon:
Cycle 1 of a full-scope issue campaign over the whole repository. This pull request
owns the **complete accepted cycle** — every issue below lands here, and no second pull
request splits it.

## Cycle issues

| # | what it fixes |
| --- | --- |
| #1157 | the Go lint config evaluator resolves both its `ttsx` launcher and its compiler from the environment alone, where the JS twin asks the project |
| #1158 | the hand-assembled plugin-transform emit drops `removeComments`, `noEmitHelpers`, and `target` from the `PrinterOptions` the pinned tsgo emitter builds |
| #1159 | nothing verifies that the website headings the LSP's `codeDescription` anchors point at actually exist |
| #1160 | the release rehearsal packs eight of thirteen published packages, and nothing keeps its list in sync |
| #1161 | a source preamble shifts every diagnostic the plugin lane reports, and the shift also defeats the duplicate filter, so one error prints twice at two positions |
| #1162 | `ttsc format` normalizes no token spacing while the guide opens with "this is what replaces Prettier" |

Closing keywords are deliberately absent from this body. It is written before any code
exists, so a claim-time closing list would close whatever the cycle later drops, defers,
or disproves. The cycle's closing set is the union of the per-commit `Close #n:` lines,
which makes the merge close exactly what landed.

## How the set was arrived at

Six discovery rounds against `16860301739bccd105c6e35e0840eff2e9a66d76`, each with a
different instrument, implementation kept closed throughout:

1. source reading plus every mechanical gate → #1157, #1158, #1159
2. a Prettier 3.8.3 oracle over 500 real files → #1160
3. generalizing each accepted issue's root cause → widened #1157 to its launcher branch
4. driving the shipped CLIs → #1161
5. per-command dogfooding and a token-spacing oracle → #1162
6. cross-package integration dogfooding on Windows → **empty**

Twelve premises were withdrawn during fact-checking rather than filed, among them a
suspected `baseUrl` mishandling in `@ttsc/unplugin` (TypeScript 7 removed the option), a
suspected precedence gap in `@ttsc/factory`'s printer (the tokens it appeared to omit do
not exist in that package's `SyntaxKind`), and a local `@ttsc/test-unplugin` failure that
a discriminating experiment traced to vite 7.3.6 on an 8.3 short-name Windows temp path,
reproducing identically with **no plugins at all**.

## Coordination

**#1158 overlaps open pull request #1154 by @xiamu33**, which fixes the `removeComments`
field of exactly the line this cycle changes and carries its own regression test. That
pull request is what exposed the class; this cycle implements the whole class and credits
it. Whoever lands first, the other should be rebased or folded — that is the maintainer's
call, not this pull request's.

No overlap with #1129, #1153, or #1155.

## Plan

Implementation runs as three parallel batches on this one branch, cut along disjoint file
sets:

| batch | issues | owns |
| --- | --- | --- |
| A | #1158, #1161 | `packages/ttsc/**`, `tests/test-banner/**` |
| B | #1157, #1159, #1162 | `packages/lint/**`, `website/src/content/docs/lint/**` |
| C | #1160 | `experimental/tarballs/index.ts`, `scripts/ci/factory-package.test.cjs` |

after the batches report.


samchon (commented):
## Batch C — #1160 — Batch Self-Review, recorded

Commits `bb53fabe1` and `844192886`.

**What landed.** The full-mode pack list moved out of `listTargets` into documented top-level constants (`FULL_PACKAGES`, `FULL_EXCLUSIONS`, `CURRENT_PACKAGES`) and now names all thirteen publishable packages. A `--print-plan` flag prints both plans as JSON and packs nothing, short-circuiting before `listTargets`/`preparePackages` so it is safe on any host. The header's "every workspace tarball" claim was corrected to what the code does. The gate went into `scripts/ci/factory-package.test.cjs` beside its precedent — "the canonical full plans cover every publishable package build" — and audits the plan in three directions: `unrehearsed`, `stale`, `contradictory`. Negatives cover a new published package with no entry, a dropped entry, a blank reason, a reasoned exclusion passing, and a ghost entry; the boundary asserts platform packages stay out of the plan, which `listTargets`' own discovery and `assert-platform-package.cjs` continue to own.

**All five missing packages are packed; the exclusion set is empty and documented as a decision.** Per-package reasons: `@ttsc/evidence`'s `files` ships Go source through a `native/*.go` plus `!native/*_test.go` negation that only a real pack can answer; `@ttsc/factory`'s `prepack` path is never exercised by `factory-package.test.cjs`'s simulated install, which copies `lib` directly; `@ttsc/graph` ships a `bin`; `@ttsc/metro`'s `./transformer` subpath resolves to rollup output; `@ttsc/playground`'s nested `lib/src` shape is consumed from the registry by two websites, and full mode already packs `@ttsc/wasm` and `@ttsc/vscode`, whose consumers are the same class.

**Batch Self-Review.** Round 1 produced `844192886`: an exclusion now counts only when its reason is a non-empty **string**, so a `null` placeholder leaves the package unrehearsed and named rather than throwing a bare `TypeError`. Round 2 was a fresh complete pass over both files and the consequence surface — the five manifests, `build-platforms.cjs` ordering, `pnpm pack` prepack/postpack behaviour, the tarball consumers (`experimental/install`, `test-unplugin`, `source-map`, `benchmarks/performance`), `typia.yml`/`nestia.yml`, `release.yml`, `next.bash`, `release-preflight.cjs`, `test-owners.cjs` ownership, line endings, the Windows spawn shape, and the docs that mention `package:tgz`. It produced nothing that survived verification.

**Per the campaign topology, this batch ran no build, no test, no formatter, and no `pnpm package:tgz`.** Every check run belongs to the finishing phase.

### Lead adjudication of the risks it reported

- **Accepted, being fixed in this cycle:** `scripts/ci/validation-plan.cjs` routes every `experimental/**` path to `continue`, so a change touching only `experimental/tarballs/index.ts` selects `typecheck` — not `package-defenses`, where the new gate lives. A gate CI does not run for the file it guards is half a gate, and keeping the list in sync is the whole point of #1160, so this is inside its consequence surface rather than a new issue. #1160's Acceptance section named the wrong lane and has been corrected, with the chronology in a comment there.
- **Acknowledged, no action:** the gate proves the list, not the pack. The first full `pnpm package:tgz` after merge is the first time those five are packed; if one fails, that is the packaging fault the issue exists to expose. The published `@ttsc/evidence@0.26.1` tarball was checked independently and does carry `native/*.go` with `*_test.go` excluded, so the sharpest case is healthy today. No workflow runs full mode, so the added packs cost no CI minutes.
- **Finishing-phase task:** `experimental/tarballs/README.md` gains a line for `--print-plan`.
- **No action, to be re-checked in the Overall Self-Review:** the gate spawns `node --experimental-strip-types` on a `.ts` file from a `.cjs` test. That is already how `pnpm package:tgz` runs in CI and the lane is Node 24.x, so it adds no new requirement.

samchon (commented):
## Overall Self-Review — round 1, over `origin/master...054f7239` — clean

Solo, whole-diff: 44 files, +2742 / −84, sixteen commits including the empty claim.

### CI

Every check on this exact head passes, including all six `platform` rows, `go`, `windows-go`, both lint lanes, `ttsc core defenses`, `ttsc native defenses`, `bundler defenses`, `package defenses`, `evidence defenses`, `graph`, `shim-audit`, `typecheck`, the `typia` and `nestia` integration workflows, and the `ci` aggregate. The planner selected the full matrix because `scripts/ci/validation-plan.cjs` is a fail-open input, which is what this cycle wants.

That settles the two risks the batches could not verify themselves, since neither ran anything: the byte-identity assertion between the plugin emit lane and the plain tsgo lane, and the preamble diagnostic correction on Windows.

### What the round checked and what it found

**`driver/diagnostics_preamble.go` (#1161).** Read in full against `ApplySourcePreamble`, whose insertion point it mirrors — after a BOM, after a hashbang line — and whose one documented off-by-one it repeats deliberately. `authoredPos` / `authoredEnd` were checked at every boundary, including a range that opens before the injected region and closes inside it. The nil paths were traced: `remapDiagnostic(nil)` returns `(nil, false)`, `convertProgramDiagnostics` guards before calling it, and `remapRelated` drops such an entry instead of dereferencing it.

I chased one suspicion to the source and it did not survive: `unanchoredPreambleDiagnostic` sets `Severity: SeverityError` explicitly where `convertDiagnostic` leaves the field alone. `SeverityError` is `iota`, so those are the same value — the explicit spelling is consistent, not a divergence. The anchor-less render path was verified end to end: with neither `raw` nor `lint`, `WritePrettyDiagnostics` routes the entry to its plain list and `CountErrors` counts it, so the exit code is preserved exactly as the code comment claims.

**A newly created asymmetry, accepted with its reasoning.** Under a preamble, diagnostics are now authored-relative while node spans in the graph dump stay checker-relative. `SourceDigest` already anticipated exactly this divergence and publishes both a Checker and a Disk digest so a consumer can detect it, and the shipped `ttscgraph` links no plugin, so its `SourcePreamble` is empty and the two systems cannot disagree there. The alternative — leaving diagnostics shifted — is strictly worse. Documented on the type rather than left implicit.

**`shim/compiler/emit_sourcemap.go` (#1158).** The seven forwarded fields match `emitJSFile`'s construction in the pinned module field for field. The header now names both oracles and both guard tests, which is what makes the next pin bump fail here rather than ship another silently-ignored option.

**`linthost/config.go` (#1157).** The resolution walk reproduces Node's `_nodeModulePaths` (skips a directory already named `node_modules`, absolutizes a relative anchor) and chases the install through its directory link before the platform hop. I verified that second subtlety against this repository's own layout rather than trusting the comment: the real `typescript` lives at `.pnpm/typescript@7.0.2/node_modules/typescript`, and `@typescript/typescript-win32-x64` sits beside it under the same `node_modules` — which is precisely the candidate the walk reaches after the `node_modules` skip. A walk that started at the project's link would have climbed past it.

**Tests.** All 24 new Go cases: one `Test*` per file, the three-part doc comment, a numbered scenario. The `#1158` matrix was read in full — positives each carry a negative twin one property away, the ES2021 numeric-separator threshold is pinned from both sides (es2022 / es2021 inclusive, es2020 / es5 exclusive), the `removeComments` case asserts the `sourceMappingURL` trailer survives as the one comment the option may not strip, and every row ends by comparing both lanes' artifact sets byte for byte with the plain lane as the oracle. A wrong-but-consistent emit cannot pass.

**Packaging (#1160).** `node --test scripts/ci/factory-package.test.cjs` run locally: 3/3. `--print-plan` short-circuits before `listTargets`, so it packs nothing and cannot throw on an unsupported host.

**Closing keywords.** `git log origin/master..HEAD` carries a `Close #n:` line for each of #1157, #1158, #1159, #1160 (twice, harmlessly), #1161, #1162. No revert exists in the range, so no closing line survives without its fix.

**Documentation.** The two pages this cycle's own behaviour made inaccurate were corrected in it: the root README's "reproduces Prettier's output", and the banner walkthrough's "the one thing source-level injection costs is source maps".

### The one judgement call worth stating plainly

**#1162 was resolved by its documented-boundary fork, not by building the missing pass.** The issue offered both and required the choice be made rather than deferred. A general token-spacing normalizer must abstain inside every template literal, string, regular expression, JSX text child and comment, and compose as disjoint edits with eighteen existing passes — written without the ability to run anything, it would ship as an unexercised source rewriter that corrupts values rather than layout. The guide now carries a `Scope` section listing every measured case, documents the three passes that do write canonical spacing inside a region they rebuild, and states the `bracketSpacing` partial-normalization outcome as a decision with its cost. The README and the lint index were brought in line.

### Nothing survived verification

No improvement was applied during this round, so it stands as the clean round over this head.

Two findings from the batches were real but outside this cycle's frozen set, and are filed rather than folded in: #1164 (`@ttsc/banner` and `@ttsc/strip` still carry #1157's two branches character for character — verified, not assumed) and #1165 (`emitBOM` and `WriteFileData` on the same hand-assembled lane). The campaign cannot be reported complete while either is open.

Issue #1166, opened 2026-08-10
fix(lint): ttsc format splits an interface body and leaves its last member unterminated


samchon:
## Problem

`ttsc format` splits a one-line interface body onto multiple lines — an action, not an
abstention — and leaves the last member without the terminator Prettier emits.

```ts
// input
export interface Alpha { alpha: number }

// prettier 3.8.3
export interface Alpha {
  alpha: number;
}

// ttsc format
export interface Alpha {
  alpha: number
}
```

This is not the token-spacing boundary #1162 documented. There the tool abstains and the
guide says so. Here the tool rewrites the construct and produces a shape Prettier never
emits, and nothing documents it.

**Affected users:** anyone running `ttsc format` or `ttsc fix` over an interface written
on one line — a migrating codebase, a paste, a generator's output.

## Evidence

Five single-purpose files through `ttsc format` (Prettier defaults, `jsDoc: false`) and
through Prettier 3.8.3 with no plugins, on the merge of #1163:

| input | ttsc | prettier | verdict |
| --- | --- | --- | --- |
| `export interface Alpha { alpha: number }` | splits, **no terminator** | splits, `alpha: number;` | **diverge** |
| `export interface Bravo { alpha: number; bravo: string }` | splits, keeps the interior `;`, **last member has none** | both members terminated | **diverge** |
| `export type Charlie = { charlie: number };` | unchanged | unchanged | agree |
| an interface already split, members terminated | unchanged | unchanged | agree |
| `export class Echo { echo: number = 1 }` | splits **and terminates** the field | same | agree |

The class-field row is what makes this look like an oversight rather than a decision:
the same split, the same missing terminator, and the class field gets one.

### Cause

`packages/lint/linthost/rules_format_semi.go`'s `Check` short-circuits every type member
before the insert branch:

```go
// Interface / type-literal members and class fields carry their own
// ASI rules … inserting a missing interface/type member terminator is out of scope
// for this strip fix, so type members short-circuit in always mode.
isClassField := node.Kind == shimast.KindPropertyDeclaration
isTypeMember := isTypeMemberKind(node.Kind)
if preferNever && (isClassField || isTypeMember) {
  stripMemberSemicolon(ctx, src, node, isClassField)
  return
}
if isTypeMember {
  return
}
```

`isTypeMemberKind` covers `PropertySignature`, `MethodSignature`, `IndexSignature`,
[...]

Issue #1165, opened 2026-08-10
fix(ttsc): the plugin emit lane still skips emitBOM and never populates WriteFileData


samchon:
## Problem

#1158 closed the `PrinterOptions` half of the divergence between the hand-assembled
plugin-transform emit and the pinned tsgo emitter. Two more parts of the same
hand-assembly are still not reproduced, and both are outside the field set that issue
declared:

1. **`emitBOM` is never applied on the plugin lane.** tsgo's `printSourceFile` applies
   `EmitBOM` *outside* `PrinterOptions`, so forwarding the whole options struct does not
   pick it up. A project that sets `"emitBOM": true` gets a byte-order mark on a plain
   build and none once a plugin transform runs.
2. **`WriteFileData` is always `nil`.** `Program.writePluginEmitOutput` calls
   `writeFile(fileName, text, nil)`, so a consumer's `WriteFile` callback receives no
   `SourceMapUrlPos` and no emitter diagnostics on this lane, where tsgo's own emit
   populates both.

**Affected users:** every project whose plugin chain contains an `EmitTransformPlugin`
— typia, `@ttsc/banner`, any third-party transform plugin — that sets `emitBOM`, and
every embedder that reads `WriteFileData` from a `WriteFile` callback.

## Evidence

- The emit assembly: `packages/ttsc/driver/emit_plugin.go`'s
  `Program.EmitWithPluginTransformers`, which prints through
  `packages/ttsc/shim/compiler/emit_sourcemap.go`'s `PrintFileWithSourceMap` and then
  writes through `writePluginEmitOutput`.
- The oracle: `internal/compiler/emitter.go` in the pinned
  `typescript-go@v0.0.0-20260429010842-56ab4af42157`. `emitJSFile` builds the
  `PrinterOptions` (now matched field-for-field by #1158) and hands them to
  `printSourceFile`, which is where `EmitBOM` and the `WriteFileData` population live —
  outside the struct, which is exactly why forwarding the struct did not reach them.
- Found while implementing #1158 in #1163, and deliberately left out of that change: it
  is the same hand-assembly and the same oracle, but not the class the issue's evidence
  and acceptance table declared, and widening a frozen cycle's scope silently is what the
  campaign's claim discipline forbids.

Neither part has been reproduced end to end yet; they are read from the two sources side
by side. Reproducing each is the first acceptance item below.

## Consequence surface

- Every ttsc lane that emits JavaScript through a plugin transform: `ttsc build`,
[...]

Issue #1164, opened 2026-08-10
fix(banner,strip): the other two config evaluators still resolve their toolchain from the environment alone


samchon:
## Problem

#1157 fixed one of **three** copies of the same config evaluator. `@ttsc/banner` and
`@ttsc/strip` still resolve both tools they spawn — the `ttsx` launcher and the
TypeScript compiler it is handed — from the process environment alone, exactly as
`@ttsc/lint`'s Go evaluator did before #1157.

Leaving them is the shape the development skill's **No whack-a-mole** rule names: one
witness patched, the class still open. This issue exists so the class is sealed rather
than sampled.

**Affected users:** any project whose `banner.config.ts` or `strip.config.ts` is
evaluated in a process that does not inherit `TTSC_TSGO_BINARY` / `TTSC_TTSX_BINARY` —
the same reachability #1157 documents, most concretely an editor wired to the shipped
`ttscserver` binary through its `--tsgo <path>` flag, and any embedder of
`lspserver.NativePluginSource` / the `driver` package.

## Evidence

Three copies, one shape.

### The compiler branch

`packages/banner/driver/banner.go:396` and `packages/strip/driver/config.go:344`, both
inside the temp-dir loader that hands the child `--cwd <ephemeral dir>`:

```go
if tsgo := os.Getenv("TTSC_TSGO_BINARY"); tsgo != "" {
  args = append(args, "--binary", tsgo)
}
```

This is character-for-character what `packages/lint/linthost/config.go:3031` held before
#1157.

### The launcher branch

`packages/banner/driver/banner.go:686` (`ttsxCommandContext`) and
`packages/strip/driver/config.go:541` (`stripTtsxCommandContext`):

```go
ttsx := os.Getenv("TTSC_TTSX_BINARY")
if ttsx == "" {
  ttsx = "ttsx"
}
```

No project anchor between the variable and a bare `PATH` lookup. `ttsx` is on `PATH`
only for a **global** `ttsc` install; for the ordinary project-local install the spawn
fails with a not-found error instead of using the `ttsc` the project has.

### What #1157 established

`packages/lint/linthost/config.go` now asks the project through one shared anchor list —
the config file first, then the resolution root's manifest — with an explicit
environment variable still winning and both last resorts unchanged (no `--binary` when
nothing resolves; the bare `"ttsx"` when no `ttsc` is reachable). It also had to
reproduce two things a naive walk gets wrong, and both apply verbatim here:

[...]

Issue #1158, opened 2026-08-10
fix(ttsc): the plugin-transform emit drops compiler options the pinned emitter forwards


samchon:
## Problem

When a project compiles through a ttsc plugin transform, three `compilerOptions` are
silently ignored: **`removeComments`**, **`noEmitHelpers`**, and **`target`** (for
numeric-literal fidelity). The same project compiled without a plugin honours all
three.

`packages/ttsc/driver/emit_plugin.go`'s `EmitWithPluginTransformers` hand-assembles
tsgo's JavaScript emit pipeline, because tsgo's `Program.Emit` has no hook to inject a
transformer. It prints through the hand-maintained shim
`packages/ttsc/shim/compiler/emit_sourcemap.go`, whose `PrintFileWithSourceMap` builds
`printer.PrinterOptions` with **four** fields where the pinned emitter builds
**seven**. Whatever the shim omits takes the Go zero value, so the option has no
effect on the plugin path.

**Affected users:** every project with an `EmitTransformPlugin` in the chain — typia,
`@ttsc/banner`, and any third-party transform plugin — that sets `removeComments`,
`noEmitHelpers`, or writes numeric literals with separators.

## Evidence

Pinned typescript-go (`v0.0.0-20260429010842-56ab4af42157`),
`internal/compiler/emitter.go:194`, `emitter.emitJSFile`:

```go
printerOptions := printer.PrinterOptions{
	RemoveComments:  options.RemoveComments.IsTrue(),
	NewLine:         options.NewLine,
	NoEmitHelpers:   options.NoEmitHelpers.IsTrue(),
	SourceMap:       options.SourceMap.IsTrue(),
	InlineSourceMap: options.InlineSourceMap.IsTrue(),
	InlineSources:   options.InlineSources.IsTrue(),
	Target:          options.Target,
}
```

`packages/ttsc/shim/compiler/emit_sourcemap.go:55`:

```go
printer := innerprinter.NewPrinter(innerprinter.PrinterOptions{
  NewLine:         options.NewLine,
  SourceMap:       options.SourceMap.IsTrue(),
  InlineSourceMap: options.InlineSourceMap.IsTrue(),
  InlineSources:   options.InlineSources.IsTrue(),
}, innerprinter.PrintHandlers{}, emitContext)
```

The file's own header says "Keep it in sync with that emitter source when the pin is
bumped", and nothing mechanically checks that.

### Reproduction

[...]

PR #1167, opened 2026-08-10
fix: the full-scope campaign cycle 2

samchon:
Cycle 2 of the full-scope issue campaign. This pull request owns the **complete accepted
cycle** — every issue below lands here, and no second pull request splits it.

## Cycle issues

| # | what it fixes |
| --- | --- |
| #1164 | `@ttsc/banner` and `@ttsc/strip` still resolve both their `ttsx` launcher and their compiler from the environment alone — the other two copies of the evaluator #1157 fixed |
| #1165 | the hand-assembled plugin emit lane skips `emitBOM` and always passes `nil` `WriteFileData`, the parts of the same oracle that live outside the `PrinterOptions` struct #1158 matched |
| #1166 | `ttsc format` splits a one-line interface body and leaves its last member unterminated — a pass that acts and produces a shape Prettier never emits |

Closing keywords are deliberately absent from this body; the cycle's closing set is the
union of the per-commit `Close #n:` lines, so the merge closes exactly what landed.

## How the set was arrived at

#1164 and #1165 were found **while implementing cycle 1** (#1163) and filed rather than
folded into its set, which an empty round had already frozen. Each is the same root cause
as an issue that cycle fixed, one instance further out:

- #1157 fixed one of three copies of the config evaluator; #1164 is the other two,
  verified character for character rather than assumed.
- #1158 matched the `PrinterOptions` struct field for field; #1165 is what `printSourceFile`
  does *around* the printer, which forwarding the struct cannot reach.

#1166 came from cycle 2's own discovery, by checking the format guide's new Scope section
against the tool. Round 2 then generalized it across twenty constructs written in shapes
Prettier never produces: fifteen reach Prettier's output exactly, two abstain, and the
three that diverge are all the same cause — `PropertySignature`, `MethodSignature`,
`IndexSignature`. Nothing outside `isTypeMemberKind` acts-and-diverges, and the two
value-carrying guards (a template literal with trailing spaces, a string with interior
spaces) came back byte-identical.

## Coordination

**#1154 (`xiamu33`) is now superseded.** It fixes `RemoveComments` in
`packages/ttsc/shim/compiler/emit_sourcemap.go`, which #1163 already implemented as part
of the whole field set — crediting that pull request for the witness. #1165 touches the
same file again for `emitBOM`. Whether #1154 is closed as superseded is the maintainer's
call, not this pull request's.

No overlap with #1129, #1153, or #1155.

## Plan

Three parallel batches on this one branch, cut along disjoint file sets:

| batch | issue | owns |
| --- | --- | --- |
| D | #1164 | `packages/banner/**`, `packages/strip/**` |
| E | #1165 | `packages/ttsc/**` |
| F | #1166 | `packages/lint/**` |

after the batches report.


samchon (commented):
## Overall Self-Review — round 1, over `25a360c63...c1632d60` — clean

Solo, whole-diff: 74 files, +3595 / −167, nineteen commits including the empty claim.

### CI

**All 28 checks pass** on this head: every `platform` row, `go`, `windows-go`, both lint lanes, `ttsc core defenses`, `ttsc native defenses`, `bundler defenses`, `package defenses`, `evidence defenses`, `graph`, `shim-audit`, `typecheck`, the `typia` and `nestia` integration workflows, and the `ci` aggregate.

That settles what the batches could not verify themselves, since none of them ran anything: three Go packages' worth of new resolution code, a Prettier conformance case that runs real prettier, and two command-level cases that depend on the whole format cascade converging.

### What the round checked

**#1165, against the oracle rather than the report.** I read the pinned `internal/compiler/emitter.go::printSourceFile` line by line and compared. Three things had to be right and are:

- `sourceMapUrlPos` is taken from `writer.GetTextPos()` **before** `WriteComment("//# sourceMappingURL=")`, and defaults to `-1`;
- `EmitBOM` is applied to `writer.String()` **after** that offset is computed, so the recorded offset is deliberately three bytes short of the trailer's position in a BOM file;
- the JavaScript write gets a `WriteFileData` and the map write gets `nil`, matching `writeText(sourceMapFilePath, sourceMap, nil)`.

The pre-mark offset reads like an off-by-three until you read the oracle. It is the emitter's own behaviour, "correcting" it here would make a plugin build disagree with the plain build of the same project, and it is pinned by a test rather than left to a comment. The one reasoned divergence is `Diagnostics`, left empty because this lane has no `emitterDiagnostics` bag — a write failure returns as an `error` — and that is documented on the field rather than implied.

**#1164, on the two things that are easy to get subtly wrong.** Both packages carry the `node_modules` skip and the `resolveDirLink` → `realpathIfPossible` chase before the platform hop, which is the pair I verified against this repository's own pnpm layout during cycle 1. The second anchor differs from lint's and is better for these packages: `driver.PluginConfigBaseDir`, which is the anchor the host itself declares for config discovery and which honours `TTSC_PLUGIN_CONFIG_DIR`, so a generated wrapper tsconfig in a temp directory does not misdirect the walk.

**#1166, on the reasoning rather than the diff.** The rule now separates the two roles a member `;` plays in Prettier's object printer — a separator between two members, printed in both layouts, and a trailing terminator inside an `ifBreak`, printed only where the list breaks. That distinction is what makes `type T = { a: number }` correct output with no terminator, and it is the batch's own round-4 finding: the first implementation decided from the member's line rather than the list's wrap and inserted there. The edit stays a zero-width insertion at `End()` so it composes with `format/indent`, `format/statement-split`, and `format/print-width` instead of contesting a range with them.

**Closing keywords.** `git log 25a360c63..HEAD` carries a `Close #n:` line for each of #1164, #1165, #1166. No revert exists in the range, so no closing line survives without its fix.

**Tests.** 51 new Go cases across three packages: one `Test*` per file, the three-part doc comment, a numbered scenario. Every regression case for #1164 sheds both environment variables first, which is the property that stops `scripts/test-go-utility-plugins.cjs`'s forwarding from masking the defect the way it did before.

**Formatting.** `pnpm format` over the integrated snapshot produced **no content change** — the batches wrote already-formatted output.

**Documentation.** The two pages this cycle's behaviour made inaccurate were corrected in it: the `semi` key's description in the format guide, which described only the statement half, and the banner walkthrough's claim that ttsx is "injected as `TTSC_TTSX_BINARY`", which was the whole truth only while the environment was the only place the evaluator asked.

### One incident, checked rather than trusted

A batch reported that its `git add` left files staged and another batch's pathless `git commit` swept them in, caught and re-cut before anything was pushed. I audited every pushed commit's file set independently: **each belongs to exactly one batch's territory**, with no stray file anywhere in the range.

### Nothing survived verification

No improvement was applied during this round, so it stands as the clean round over this head.

Four findings from the batches were real but outside this cycle's frozen set, and are filed rather than folded in:

- **#1168** — three member-separator divergences `format/semi` still carries, each measured against the pinned Prettier while implementing #1166: the strip direction removing a flat list's separator, mapped-type members never being visited, and a `,` separator never normalised.
- **#1169** — the TypeScript config loader exists three times in Go, and this campaign has now paid for it twice: #1157 fixed one copy and #1164 was filed a cycle later for the other two.

Both are the next cycle's work, and the campaign cannot be reported complete while either is open.

Issue #1169, opened 2026-08-11
refactor: the TypeScript config loader exists three times in Go, and the campaign has now paid for it twice


samchon:
## Problem

The TypeScript config-file loader exists **three times** in Go, wholesale, plus once in
TypeScript. #1157 and #1164 have now written the same ~200-line toolchain resolution into
all three Go copies, which is the point at which the duplication stops being incidental.

The Go copies:

- `packages/lint/linthost/config.go`
- `packages/banner/driver/banner.go`
- `packages/strip/driver/config.go`

Each carries its own spelling of: `loaderTempBase`, `resolveDirLink`,
`realpathIfPossible`, `findNearestNodeModules`, `linkNearestNodeModules`,
`nearestPackageType`, `shouldRunTtsxThroughNode`, `ttsxCommand` /
`ttsxCommandContext`, the loader shim source, `nodeConfigLoaderEnv`, the failure-reason
reader — and now `configToolAnchors`, `resolveConfigTsgo`, `tsgoBinaryFrom`,
`resolveTtsxLauncher`, `ttsxLauncherFrom`, `nodePackageManifestFrom`,
`nodePlatformPair`, `nodePlatformPairFor`.

The cost is already measurable: **#1157 fixed one copy and #1164 was filed to fix the
other two**, a full cycle later. That is the duplication's failure mode, not a hypothetical.

## Evidence

- #1157 (`packages/lint`) and #1164 (`packages/banner`, `packages/strip`) fix the same
  two branches with the same resolution. Both are merged or in flight in this campaign.
- `packages/lint/src/index.ts` is the fourth copy, in TypeScript, and is where the policy
  originated (`e0cd1adaa`).
- The three Go copies' new blocks each open with a comment naming the other siblings, so
  a future change has a checklist. A checklist is a mitigation, not a fix.

## Why it was not done inside #1164

The implementing agent recorded the reasoning, and it is sound enough that this issue
should start from it rather than relitigate it:

1. **The only obvious shared home is `packages/ttsc/driver`, a public seam** third-party
   plugins compile against. Putting a *resolution policy* there is a permanent contract
   addition. `driver/windowsjunction` is precedent for an OS primitive, not for this.
2. **It would not unify all three.** `packages/lint`'s `go.mod` deliberately carries no
   requirement on the `packages/ttsc` module, with its own written reason. A `driver`
   home serves two of three.
3. **Extracting only the newest 200 lines creates a fourth structural pattern** and still
[...]

Issue #1168, opened 2026-08-11
fix(lint): three member-separator divergences format/semi still carries


samchon:
## Problem

#1166 gave `format/semi` its insert direction for members that end their line. Three
member-separator divergences from Prettier 3.8.3 remain, all measured against the
repository's pinned prettier while implementing it, none fixed because each is a
different decision from the one that issue declared.

1. **The strip direction removes a separator Prettier keeps.** Under `semi: false`:

   ```ts
   // input
   type T = { a: number;
     b: string };

   // prettier 3.8.3            // ttsc format
   type T = { a: number;        type T = { a: number
     b: string };                 b: string };
   ```

   `memberSemicolonRedundant` treats the `;` as redundant, but between two members in a
   *flat* list it is the separator, not a trailing terminator. #1166's contract required
   `prefer: "never"` to behave exactly as it did, so it was left alone.

2. **Mapped-type members are never terminated.**

   ```ts
   // input                     // prettier            // ttsc format
   type M = {                   type M = {             type M = {
     [K in string]: string        [K in string]: string;  [K in string]: string
   };                           };                     };
   ```

   `KindMappedType` is not in `format/semi`'s `Visits()` list at all, so the rule is
   never asked about it.

3. **A `,` separator is never normalized to `;`.** TypeScript accepts either in a type
   member list; Prettier always prints `;`. `insertMemberSemicolon` abstains on a member
   ending in `,` deliberately — appending would emit `a: number,;` — and no pass owns
   the rewrite, so a comma-separated body stays comma-separated, and a body that mixes
   both stays mixed.

## Evidence

Each row above was produced by formatting the input with `ttsc format` (Prettier
defaults) and with the pinned Prettier 3.8.3, and comparing. They were found while
implementing #1166 and are recorded in its batch report rather than inferred.

The relevant code: `packages/lint/linthost/rules_format_semi.go` —
`memberSemicolonRedundant` for (1), the `Visits()` list for (2), and
`insertMemberSemicolon`'s `case ','` for (3).

## Consequence surface

- All three need a body that is not already in Prettier's shape, which is why the
  repository's own corpus and the 500-file fixed-point probe never surface them.
- (1) only fires under `semi: false`, which is not the default.
- (2) affects every mapped type in a broken body — a common shape in a typed codebase.
- (3) produces a body mixing `,` and `;` after #1166's insert runs on the members that
  do end their line, which is legal TypeScript and closer to the oracle than before, but
  is not the oracle.
[...]

Issue #1157, opened 2026-08-10
fix(lint): the Go config evaluator resolves its launcher and compiler from the environment alone


samchon:
## Problem

`@ttsc/lint`'s **Go** config evaluator resolves both tools it needs — the `ttsx`
launcher it spawns and the TypeScript compiler it hands that launcher — from the
process environment alone. Its **JS** twin resolves both from the project being
linted. When the environment does not carry them, evaluating a `lint.config.ts` (or
`.mts`/`.cts`) fails before a single line of the config is read.

With the compiler missing:

```
ttsc: typescript is required.
Install the native TypeScript compiler in the consuming project:
  npm i -D typescript
@ttsc/lint: load TypeScript config file <path>/ttsc-lint.config.ts: exit status 2
```

The JS evaluator had the same compiler branch and was changed in e0cd1adaa
(`fix(lint): resolve the config evaluator's compiler from the project, not the
environment`, v0.26.1). That commit named the Go twin explicitly and left it alone on
purpose:

> The Go evaluator in `linthost/config.go` reads the same variable through the same
> branch and would need the same treatment; no run here reached it, because the
> failure aborts before `ttsc check`, so it is reported rather than changed blind.

This issue is that follow-up, with a reproduction — and it covers the launcher branch
too, which is the same asymmetry one step earlier in the sequence.

**Affected users:** anyone whose lint host evaluates a TypeScript lint config in a
process that does not inherit `TTSC_TSGO_BINARY` / `TTSC_TTSX_BINARY` — most concretely
an editor wired to the shipped `ttscserver` binary through its documented
`--tsgo <path>` flag, and any embedder of `lspserver.NativePluginSource` / the `driver`
package.

### 1. The compiler

`packages/lint/linthost/config.go:3031`, inside `loadTypeScriptConfigEvaluationWithin`:

```go
if tsgo := os.Getenv("TTSC_TSGO_BINARY"); tsgo != "" {
  args = append(args, "--binary", tsgo)
}
```

The child is launched with `--cwd <ephemeral temp dir>`, so it cannot discover
`typescript` the way an ordinary invocation does; `linkNearestNodeModules` is the only
thing that can put the project's modules within its reach, and it finds nothing when
the config's ancestry has no `node_modules`.

The JS twin now asks the project first (`resolveConfigTsgo` →
`packages/lint/src/index.ts`), anchored on the config file, then the project root, then
[...]

PR #1170, opened 2026-08-11
fix: the full-scope campaign cycle 3

samchon:
Cycle 3 of the full-scope issue campaign. This pull request owns the **complete accepted
cycle**; no second pull request splits it.

## Cycle issues

| # | what it fixes |
| --- | --- |
| #1168 | the three member-separator divergences `format/semi` still carries — the strip direction removing a flat list's separator, mapped-type members never being visited, and a `,` separator never normalised |
| #1169 | the TypeScript config loader exists three times in Go, and this campaign has now paid for it twice |

Closing keywords are deliberately absent from this body; the cycle's closing set is the
union of the per-commit `Close #n:` lines.

## How the set was arrived at

Both were measured while implementing cycle 2 (#1167) and filed rather than folded into
its set, which an empty round had already frozen.

- #1168's three rows were each produced by formatting an input with `ttsc format` and
  with the repository's pinned Prettier 3.8.3 and comparing, while writing #1166's insert
  direction.
- #1169 is the duplication #1157 and #1164 both paid for: one issue fixed one copy of the
  evaluator, and a second issue was needed a cycle later for the other two.

Cycle 3's own discovery round was **empty**. It re-verified cycle 2's three fixes in the
shipped tooling rather than trusting CI alone: the twenty-construct formatter sweep went
from three divergences to zero, the plain-versus-plugin emit agrees on all six option
sets including `emitBOM` and its negative twin, and the newest interaction — `emitBOM`
plus a source preamble plus a corrected source map, with a shebang and an inline map as
variants — agrees structurally with the plain lane on every property.

## Plan

Two parallel batches on this one branch, cut along disjoint **file** sets inside the
shared `packages/lint` package:

| batch | issue | owns |
| --- | --- | --- |
| G | #1168 | `packages/lint/linthost/rules_format_semi.go`, `packages/lint/test/format/**` |
| H | #1169 | `packages/lint/linthost/config.go`, `packages/lint/test/config/**`, `packages/banner/**`, `packages/strip/**`, and whatever shared home it chooses |

after the batches report.


samchon (commented):
## Overall Self-Review — full-scope cycle 3

Solo review of the whole declared surface (`cbfecfb4a...HEAD`, 9 commits, 31 files), run in fresh rounds until a complete round produced nothing.

### The one thing a reviewer should look at first

**#1168 reverses a judgment this repository made deliberately.** The test deleted by `5b76f8f21` was not an oversight — its doc comment said the expectation was *"deliberately short of the oracle: normalizing an authored separator is a rewrite no ttsc pass owns"*. This PR takes the opposite position and should be read as a decision, not a bug fix.

The reason to reverse it: the old behavior was not abstention but half-action. Given

```ts
interface Shape {
  alpha: number,
  bravo: string
}
```

the rule wrote `bravo: string;` and left `alpha: number,` beside it — a mixture neither the author nor Prettier produces. Once `format/semi` is named the owner of the member separator, it owns both spellings of it.

### Oracle table, measured rather than argued

Prettier 3.8.3, `--parser typescript`:

| input | `semi: false` | `semi: true` |
| --- | --- | --- |
| `type Flat = { a: number, b: string }` | `{ a: number; b: string }` | same, `;` after the alias |
| broken `interface` members ending `,` | separators dropped | `;` |
| `type Single = { a: number, }` | `{ a: number }` | `{ a: number }` |

Both directions land on one table, which is why batch G's extension of the comma handling to `prefer: "never"` — beyond the literal wording of issue row (3) — is in scope rather than an overreach. A rule that owned the separator under `always` and abandoned it under `never` would emit non-oracle output from its own pass.

### The collateral case that would have made this dangerous

An object literal's accessors are `,`-separated and spelled with the same `GetAccessor`/`SetAccessor` kinds a bodiless interface accessor uses, with the body's `}` sitting right where both directions scan. Rewriting that comma produces syntax the grammar rejects; deleting it splices two properties together.

`memberTakesSemicolonTerminator` excludes it by parent kind — only an interface body, type literal, or class body qualifies — and `format_semi_keeps_object_literal_accessor_separators_untouched_test.go` asserts both directions. `Visits()` carries no `KindEnumMember` and no `KindPropertyAssignment`, so enum and object-literal commas are out of reach entirely.

One finding here, fixed in `ff60ce03a`: that guard test's doc comment described the guard as skipping "a member carrying a body, or one whose list is `,`-separated". It never inspects separators. The wording sent the next reader looking for a comma check that does not exist.

### #1169's gate is not decorative

Verified empirically rather than by reading: planting a one-character semantic drift in `packages/banner/driver/banner.go` (`err == nil` → `err != nil` inside `nodePackageManifestFrom`) makes the gate fail, name the function, and name both disagreeing files. Restored, it agrees again.

The normalization cannot launder a difference: the error-prefix canonicalization is scoped to each copy's own package name, and `a copy borrowing another copy's error prefix fails` is a test for exactly that. Undeclared functions, deletions, moves outside the markers, and inverted regions are each a failure, so the table cannot be satisfied by omission.

`planForPaths` selects `typecheck` for all three guarded files, which is the lane the gate runs in — it fires on every change it guards.

Batch H's edits outside its declared file set were forced by the repository's own gates: `test-owners.cjs` is an enforced registry, and a new test file that claims no lane fails `test-owners.test.cjs`. Cycle 1's `experimental/tarballs/` planner clause is intact.

### Limits this PR accepts, stated rather than hidden

- **No reflow.** `type T = { a: number\n , b: string }` normalizes the comma in place and yields `{ a: number\n ; b: string }` — valid, correct separator, a layout Prettier would reflow. That is the boundary `memberListBreaks` already documents for object and mapped types: no ttsc pass reflows a braced body, so the rule fixes the separator it owns and leaves the wrap the author wrote.
- **A fourth copy no Go gate can reach.** `resolveConfigTsgo` / `resolveTtsxLauncher` in `packages/lint/src/index.ts` hold the same policy in another language. The new marker comment names this rather than implying the gate covers it. Whether those anchors can arrive relative is unverified and carried to the next discovery round; the worst case there is bounded — `createRequire` throws `ERR_INVALID_ARG_VALUE` on a relative path and the surrounding `catch` takes the next anchor, so it would be a miss, not the wrong-tree walk the Go copies could do before `filepath.Abs`.

### Test contract

Every new Go test file carries one `Test*` named after what it asserts, a three-part doc comment, and numbered steps. Behavior changes arrive in twins: flat/broken mapped type, strip/normalize the comma, and the object-literal exclusion as the negative case for both directions.

### Verified in the rebuilt shipped binary, not only in CI

`ttsc format` run against a fixture whose `@ttsc/lint` plugin was built from this branch's Go source, then the same input through Prettier 3.8.3:

| construct | `ttsc format` | oracle |
| --- | --- | --- |
| `type Flat = { a: number, b: string }` | `{ a: number; b: string };` | identical |
| broken `interface` with `,` and a bare member | both `;` | identical |
| object literal's `get first() {...},` | untouched | identical |
| `type M = {\n  [K in string]: string };` | `[K in string]: string; };` | `;` in the same place, `}` reflowed to its own line |

The first ten lines match byte for byte. The mapped type differs only by the wrap Prettier reflows and ttsc deliberately does not, which is the boundary named above.

Issue #1173, opened 2026-08-11
fix(ttsc): --build is forwarded after -p, so tsgo tells the user it must be first when it already was


samchon:
```console
$ ttsc --build .
error TS6369: Option '--build' must be the first command line argument.
```

`--build` *was* the first argument. The diagnostic describes a command line the user did not write.

## Why it is structurally impossible to satisfy

`createTsgoBuildArgs` opens the forwarded argument list with the project it resolved, and appends passthrough flags after it:

```ts
// packages/ttsc/src/compiler/internal/runBuild.ts:1236
const args = ["-p", execution.tsconfig];
...
// passthrough flags are pushed later
```

So a forwarded `--build` always lands after `-p <tsconfig>`, and tsgo's TS6369 is the inevitable result. No spelling of the ttsc command line can produce a passing invocation.

Measured on `efb94e107` with `typescript@7.0.2`, against a two-package solution (`{ "files": [], "references": [{ "path": "./pkg-a" }, { "path": "./pkg-b" }] }`):

| command | result |
| --- | --- |
| `tsgo --build .` | exit 0, emits `pkg-a/dist`, `pkg-b/dist`, both `.tsbuildinfo` |
| `ttsc --build .` | exit 1, TS6369 |
| `ttsc build` | exit 0, emits nothing — correct, `ttsc build` is the single-project lane per `--help`, and the solution project has no files |

## The argument order is not the bug

`-p <tsconfig>` first is deliberate and load-bearing. ttsc resolves the project itself — `extends` chains, plugin config discovery, cache keys, the resident session's identity — and pins the result rather than forwarding raw argv. Solution mode compiles many projects, while ttsc's plugin, cache, and emit architecture is built around one resolved project, so **supporting `--build` may well be out of scope on the merits.** This issue does not ask for it.

## What is the bug

Whether or not solution builds are ever supported, ttsc should not forward a flag it structurally cannot honor and let the user read an upstream diagnostic that blames their argument order. Two honest shapes, either acceptable:

1. Recognize `--build` / `-b` in the launcher and refuse it with a message that names the real reason (ttsc pins one resolved project; solution mode is not supported), pointing at whatever the supported path is for a monorepo.
2. Support it properly, which means `--build` leads the forwarded argv and the plugin/cache/emit layers learn about the multi-project shape — a much larger change, and one worth its own design discussion rather than being assumed here.

A regression case should pin whichever is chosen, since today nothing exercises `--build` through the launcher at all.

Issue #1172, opened 2026-08-11
fix(ttsx): a tool-injected outDir demands a rootDir the project never declared, so ttsx cannot run a check-only project


samchon:
`ttsx` refuses to run a project whose sources sit under a subdirectory and which declares no `compilerOptions.rootDir` — a project that `tsgo` and `ttsc` both accept, because it declares no output at all.

```
tsconfig.json(1,3): error TS5011: The common source directory of 'tsconfig.json' is './src'.
The 'rootDir' setting must be explicitly set to this or another path to adjust your output's file layout.
```

The user is being told to configure the layout of output they never asked for and never see.

## Reproduction

```jsonc
// tsconfig.json
{ "compilerOptions": { "module": "commonjs", "target": "ES2020", "noEmit": true }, "include": ["src"] }
```

```ts
// src/run.ts
declare const process: { argv: string[] };
console.log(JSON.stringify({ argv: process.argv.slice(2) }));
```

Measured on `efb94e107`, `typescript@7.0.2`:

| project | command | result |
| --- | --- | --- |
| as above | `tsgo -p .` | exit 0 |
| as above | `ttsc` | exit 0 |
| as above | `ttsc --emit` | exit 0, emits `src/run.js` in place |
| as above | `ttsc src/run.ts` | **TS5011, exit 2** |
| as above | `ttsx src/run.ts` | **TS5011, exit 2** |
| plus `"rootDir": "src"` | `ttsx src/run.ts` | exit 0, runs |
| sources at the project root instead | `ttsx run.ts` | exit 0, runs |

Adding `rootDir` is the only change between rows 5 and 6, and flattening the layout is the only change between rows 5 and 7.

## Why it happens

TS5011 fires when an `outDir` is in play and the common source directory is inferred rather than declared. This project declares no `outDir`, which is why `tsgo`, `ttsc`, and even `ttsc --emit` never raise the question — `--emit` with no `outDir` emits in place, so there is no layout to adjust.

`ttsx` and single-file emit each inject an output directory of their own, and it is that injection that makes tsgo demand an explicit layout. For `ttsx` the injected directory is a PID-isolated temp dir it fully controls:

```ts
// packages/ttsc/src/launcher/internal/prepareExecution.ts — buildProject
const result = runBuild({
  emit: true,
  outDir: context.emitDir,   // never the consumer's outDir
  ...
});
```

[...]

PR #1174, opened 2026-08-11
fix: the full-scope campaign cycle 4

samchon:
Cycle 4 of the full-scope issue campaign. Draft while the batches land; the
closing set is the union of the per-commit `Close #n:` lines, so nothing closes
from this body.

## Cycle issues

- **#1172** — `ttsx` and single-file emit inject an output directory of their
  own, and that injection makes tsgo demand an explicit `rootDir` from a project
  that declares no output at all. A check-only project that `tsgo`, `ttsc`, and
  `ttsc --emit` all accept is refused with TS5011.
- **#1173** — `--build` is forwarded after `-p <tsconfig>`, so tsgo reports that
  it must be the first command line argument when the user already put it first.

## How the set was frozen

Two full-scope discovery rounds against `efb94e107`. Round 1 filed both issues
and withdrew two premises (the JS fourth copy of the config-loader policy, and
`ttsc build` on a solution tsconfig). Round 2 swept the forwarded CLI flag
surface, `ttsc cache paths --json`, watch mode including the non-TTY and
`--preserveWatchOutput` cases, and the mechanical gates, and produced nothing.


samchon (commented):
## Overall Self-Review — full-scope cycle 4

Solo review of the whole declared surface (`efb94e107...HEAD`), run in fresh rounds until a complete round produced nothing.

### The batch was right to disobey its instruction, and the pinned source is verifiable

The lead's brief for #1172 told the batch to synthesize `rootDir` from the rule the linked-layout path already uses, `commonAncestorDirectory(path.dirname(entry), context.root)`, so the codebase would end with one policy. **That instruction was wrong**, and the batch said so with a source citation rather than following it.

Checked against the pinned typescript-go (`v0.0.0-20260429010842-56ab4af42157`) rather than taken on trust:

```go
// internal/outputpaths/commonsourcedirectory.go — GetCommonSourceDirectory
} else if options.ConfigFilePath != "" {
    // If the rootDir is not specified, then the common source directory is the directory of the config file.
    commonSourceDirectory = tspath.GetDirectoryPath(options.ConfigFilePath)
}
```

The file list is consulted only for a config-less program. So the entry's own directory agrees with tsgo only when the entry sits beside the tsconfig — exactly the flat-versus-nested split in the issue — and pinning the **config directory** reproduces tsgo's own layout byte for byte, which is why nothing moves.

TS5011 itself is confirmed as a *layout-changed advisory*, not a requirement: `internal/compiler/program.go` raises it only when the config-dir rule and the legacy computed-common-dir rule disagree while an `outDir` is in play.

### Verified in the rebuilt launcher, not only in CI

`packages/ttsc` rebuilt from this branch, then run against fixtures:

| check | result |
| --- | --- |
| `ttsx src/run.ts hello` — nested `src/`, `noEmit`, no `rootDir` (the #1172 repro) | runs, prints `["hello"]`, exit 0 |
| `ttsc src/run.ts` — same project, single-file emit | exit 0 |
| `.js` leaked into the source tree by either | none |
| `ttsc --build .` | ttsc's own refusal, exit 2, no TS6369 |
| `ttsc -b .` | same message, so the alias is really wired |
| `ttsx --build src/index.ts` | ttsx's own refusal, exit 2 |
| `ttsx src/run.ts --build` | the program receives `["--build"]`, exit 0 |

### One risk the batch accepted did not survive measurement — in the fix's favour

Batch I accepted that a no-`rootDir` project whose `include` reaches *above* its config directory "now runs instead of failing TS5011, and tsgo writes `.js` beside those above-root sources."

Measured: it does not run, and it writes nothing. The pinned root makes tsgo enforce containment, so the project is refused with **TS6059** naming the exact file it cannot place — where before the pin it was refused with **TS5011**. Refused before, refused after; only the diagnostic changed, and the new one names the offending file and why it is in the program. Declaring the containing `rootDir` runs it and still leaves no `.js` in the source tree, so the escape hatch is real.

That measurement did produce a finding against my own work: the `execute.mdx` paragraph I had written said a project that declares no `rootDir` needs none, which is false for exactly that shape. Corrected in `b39c52eb0` to name it.

### The lead closed the gap batch J could not

Batch J correctly refused to edit `runTtsx.ts` — the other batch's file — and reported `ttsx --build` as a named gap instead. Measured on the shipped launcher, it was the identical defect with the identical misleading text, so it is part of the same verified class rather than a follow-up. The lead applied it after batch I finished: the message now lives in one place parameterized by the launcher's prefix, `--build` is declared for the `ttsx` subcommand so the parser hands it to the refusal, and the case asserts both that `ttsx --build <entry>` is refused and that `ttsx <entry> --build` still reaches the program.

Declaring the subcommand without the refusal would have swallowed the flag silently, which is why both halves landed together.

### The generated-output claim was checked with the gate, not the claim

Batch J edited `website/src/content/docs/ttsc/flags.mdx` — generated output, outside its declared set — by hand and said so. `check-flags.test.cjs` passes either way, so it proves nothing here; `node packages/ttsc/scripts/check-flags.cjs` is the drift gate, and after the `ttsx` subcommand was added it reported the row stale. Regenerated, and the gate now reports `generated output matches schema` and `15 declared kinds agree with the upstream option table`. The two `flags_gen.go` files are untouched, as a launcher-only flag should leave them.

### Cross-batch hygiene

Every commit's file set matches its batch's territory; no file appears in both batches' commits. The claim commit carries `Cycle issues:` and no `Close #n:` line, so the closing set is exactly the union of the implementation commits' lines — one for #1172, two for #1173.

`TestProject.commonJsProject`'s new `withoutUndeclaredOptions` is inert for every existing fixture: `tsconfig()` is a bare `JSON.stringify`, which already omitted `undefined`. The change makes the removal the helper's stated behaviour instead of a serializer side effect, which is what let the new cases build the project shape the default `rootDir: "src"` had been hiding from all 97 ttsx cases.

The three files batch I touched outside its declared set — `runSingleFileEmit.ts`, `runtimeHooks.ts`, `compileProjectInMemory.ts` — are one option each, and they are the reason the fix covers the class rather than the reported command: every lane that injects an `outDir` the project never declared now pins the same root.

### Adjacent lanes, measured rather than reasoned

Batch J argued from the code that the new schema row is inert everywhere else. Run against the rebuilt launcher: `ttsc cache paths --json`, `ttsc check`, `ttsc format`, and `ttsc fix` all exit 0; `ttsc cache paths --build` still fails with its own `cache paths does not support "--build"` rather than the new refusal; and ordinary passthrough is untouched — `ttsc --listFiles` still prints the file list and exits 0.

PR #1176, opened 2026-08-12
feat(shim): expose checker type-instantiation mappers

### LINKED ISSUE #1175: Expose checker type-instantiation mappers to Go source plugins
# Expose checker type-instantiation mappers to Go source plugins

## Problem

The checker shim exposes the pieces needed to inspect a generic reference, including `Checker_getTypeArguments` and construct signatures, but it does not expose the checker operation that applies those arguments to types read from the generic declaration.

A Go source plugin can therefore observe `CustomClass<Arg1, Arg2>` and read the class constructor, but the constructor parameter types remain declaration types such as `A`, `A[]`, or `[A, B]`. Exploring those uninstantiated types either produces incorrect reflection metadata or fails on an unbound type parameter.

The shim is the only legal route from an external Go module into `github.com/microsoft/typescript-go/internal/checker`, so consumers cannot solve this without a ttsc shim addition.

## Evidence

A downstream typia transform reproduces the missing surface while reflecting generic class constructors:

1. Read the concrete arguments from a reference such as `CustomClass<Arg1, Arg2>`.
2. Read the class constructor signature from the class symbol's static type.
3. Read each constructor parameter type from the declaration.
4. Attempt to substitute `A -> Arg1` and `B -> Arg2` before exploring the parameter.

Step 4 requires these unexported typescript-go primitives from the pinned checker:

- `(*Checker).instantiateType(type, mapper)`
- `newSimpleTypeMapper(source, target)`
- `(*Checker).combineTypeMappers(first, second)`

The local consumer at `packages/vybia/native/core/factories/internal/metadata/emplace_metadata_object.go` uses that exact sequence. Its runtime transform probe demonstrates the required consequence: a constructor parameter `A[]` must become `Array<Arg1>`, while bare `A` and `B` positions retain their type-parameter slot metadata alongside the concrete reflected type.

The checker shim is hand-maintained and already uses `//go:linkname` for unexported checker operations, so this surface belongs there rather than in consumer-specific compiler-host logic.

## Consequence surface

- Go source plugins reflecting generic constructor, method, or callable declaration types cannot apply a reference's concrete type arguments.
- The failure affects direct type parameters and parameters nested in arrays, tuples, maps, and other constructed types.
- Exposing `TypeMapper` also exposes its `Kind()` result, so the complete `TypeMapperKind` enum family must be re-exported to preserve shim closure.
- A compile-only test proves symbol linkage but not substitution. The shim needs a runtime probe over a real Checker.
- The checker shim's standalone module needs its transitive checksums recorded; a clean standalone `go build ./...` otherwise fails before compilation.

## Proposed approach

Add hand-maintained checker shim wrappers for:

- `Checker_instantiateType(checker, type, mapper)`
- `Checker_newSimpleTypeMapper(source, target)`
- `Checker_combineTypeMappers(checker, first, second)`

Alias `TypeMapper` and `TypeMapperKind`, re-export every `TypeMapperKind` member, and remove the resolved `ESCAPE|checker|TypeMapper` shim-audit baseline entry.

Preserve the upstream boundaries:

- instantiation returns nil for a nil checker or type and preserves the original type for a nil mapper;
- a simple mapper requires non-nil source and target types;
- combining with a nil first mapper returns the second mapper without requiring a checker;
- composing two non-nil mappers requires a checker and a non-nil second mapper.

Record the checker shim's tidy `go.mod`/`go.sum` state so the module builds independently.

## Acceptance criteria

- A real-checker runtime probe instantiates a constructor parameter `T[]` as `string[]` through a single mapper.
- The probe composes two mappers and instantiates `[A, B]` as `[number, boolean]`.
- The probe covers the nil-input boundaries above.
- `TypeMapperKind` is closed under `pnpm --filter ttsc shim:audit`.
- The checker shim and `packages/ttsc` build successfully.
- The lint Go runner passes with the new runtime probe.
- The plugin-author shim documentation lists the three wrappers.

## Verification commands

```text
pnpm --filter ttsc shim:audit
go build ./...                         # packages/ttsc/shim/checker
go build ./...                         # packages/ttsc
node scripts/test-go-lint.cjs
pnpm run format
pnpm run test
pnpm run build
```

CristianPi:
# feat(shim): expose checker type-instantiation mappers

Closes #1175

## Intent

Let Go source plugins instantiate types read from a generic declaration with the concrete arguments carried by a reference. This closes the generic-reflection gap where a plugin can inspect `CustomClass<Arg1, Arg2>` but constructor parameter types still appear as declaration types such as `A`, `A[]`, or `[A, B]`.

The implementation is grounded in a downstream typia transform that reflects generic class constructors. It maps each declared class type parameter to the corresponding reference argument, then instantiates each constructor parameter before exploring it.

## Scope

- Alias the checker's `TypeMapper` and complete its reachable `TypeMapperKind` enum family.
- Bridge `instantiateType`, `newSimpleTypeMapper`, and `combineTypeMappers` from the pinned typescript-go checker through the hand-maintained shim.
- Preserve upstream nil-mapper and mapper-composition boundaries in the public wrappers.
- Remove the resolved `ESCAPE|checker|TypeMapper` shim-audit baseline entry.
- Record the checker shim's tidy indirect dependencies and checksums so it builds as a standalone module.
- Add a real-checker runtime probe for the downstream constructor-parameter path:
  - `T[]` becomes `string[]` through a simple mapper;
  - `[A, B]` becomes `[number, boolean]` through composed mappers;
  - nil-input boundaries remain defined.
- List the wrappers in the plugin-author AST and Checker guide.

## Design notes

`Checker_combineTypeMappers` composes mappers. It does not construct a `MergedTypeMapper`; the resulting `CompositeTypeMapper` reports `TypeMapperKindUnknown`. The public contract asserted here is substitution behavior rather than a mapper-kind value.

This remains a general shim capability. No typia-specific branch, fixture name, or expected consumer output is embedded in compiler-host logic.

## Deferred

None for this shim addition. Generic method/function reflection and consumer metadata representation remain owned by the downstream plugin.

## Local verification

- `go build ./...` in `packages/ttsc/shim/checker`: pass.
- `go build ./...` in `packages/ttsc`: pass.
- `pnpm --filter ttsc shim:audit`: pass, with exposed enums, reachable symbols, and producers closed.
- `node scripts/test-go-lint.cjs`: pass, including `TestTypeInstantiationMappersReachRuntimeEndpoints`.
- `pnpm run format`: pass.
- `pnpm run test` under Node v24.19.0: all prerequisite, build, typecheck, Go, and preceding feature suites passed. One unrelated utility-host case hit a transient cold plugin-build failure after roughly 35 minutes; its exact filtered rerun passed in 42.7 seconds.
- `TTSC_GO_VERSION=go1.26.5 pnpm run build` under Node v24.19.0: pass for every platform package and the graph bundle. The version override removes Fedora's local `-X:nodwarf5` suffix when selecting official cross-platform Go archives.

## Exclusions

Ten one-line `go.sum` files generated in minimal shim modules during local builds were removed and are not part of this pull request. Upstream does not track sums for those minimal modules. Only `packages/ttsc/shim/checker/go.sum` is included because the checker module now records the full indirect dependency graph and its clean standalone build requires those checksums.


CristianPi (commented):
Overall Self-Review — clean final round

Reviewed the complete `master...28d4a52` consequence surface after all fixes and formatting.

Verified:

- the public shim aliases and `TypeMapperKind` enum family are closed under the shim audit;
- the linkname wrappers match the pinned typescript-go endpoints;
- mapper composition preserves documented nil boundaries and rejects unusable nil inputs;
- the runtime probe follows the downstream declaration-type-parameter path and instantiates constructor parameters `T[] -> string[]` and `[A, B] -> [number, boolean]` through a real Checker;
- the checker shim builds independently with its recorded module checksums;
- the plugin-author documentation lists the new public wrappers;
- no `.agents`, `AGENTS.md`, or `CLAUDE.md` file changed;
- the PR contains only the six feature-owned files documented in the body.

No remaining correctness, compatibility, test, documentation, packaging, or scope finding was identified in the final fresh round.

samchon (commented):
A green probe only settles the behavior it actually witnesses. The existing concrete substitutions proved `T[] -> string[]` and `[A, B] -> [number, boolean]`, but they did not cover parallel substitution when a target argument is itself one of the declaration parameters.

A fresh full-diff review found that combining independent simple mappers uses upstream `combineTypeMappers`, which is sequential composition rather than a parallel source-to-target table. The valid self-reference `Pair<A, B> -> Pair<B, A>` therefore instantiated `[A, B]` as `[A, A]` instead of `[B, A]`.

Follow-up commit `30a4405e` closes that consequence:

- exposes upstream `newTypeMapper(sources, targets)` as a checked parallel `Checker_newTypeMapper` wrapper;
- rejects empty, length-mismatched, and nil-containing mappings before they can create an unusable mapper;
- adds a real-Checker regression probe for `Pair<B, A>` and the input boundaries;
- lists the wrapper in the plugin-author guide.

Local verification on the updated head:

- the new regression failed before the fix with `[A, A]`, expected `[B, A]`;
- both mapper runtime probes pass through the repository scratch-module runner;
- `go build ./...` passes in `packages/ttsc/shim/checker` and `packages/ttsc`;
- `pnpm --filter ttsc shim:audit` passes;
- `pnpm format` passes.

The full local lint Go suite also ran until its 10-minute global timeout in the unrelated existing `TestLoadRuleConfigTypeScriptFactoryMergesReturnedDefaultWrapper`; the two affected probes were then isolated through the same official runner and both passed. Remote checks for the new head remain the merge gate.

samchon (commented):
Final fresh-round follow-up on `17d5e48c`:

- corrected the public wrapper comment that still described sequential `Checker_combineTypeMappers` as the general way to map several declaration parameters;
- documented `Checker_newTypeMapper` for parallel parameter/argument slices and reserved `Checker_combineTypeMappers` for sequential mapper stages;
- reran `pnpm format`, both mapper runtime probes through the official scratch runner, and `pnpm --filter ttsc shim:audit`; all pass.

This removes the remaining guidance path that could reproduce the `[A, A]` capture defect despite the new safe wrapper.

samchon (approved):
Approved after a fresh review of exact head 17d5e48c. The reported shim gap is valid; the parallel-substitution capture bug and the mapper-usage guidance found during review are fixed, the regression probe passes, and all current required checks are green. No remaining actionable finding.

PR #1187, opened 2026-08-12
fix: bound cache validation and source-build storage

### LINKED ISSUE #1184: fix(unplugin): bound persistent transform validation by each delivered module's inputs
## Problem

The follow-up in https://github.com/samchon/ttsc/issues/970#issuecomment-5258998680 reports that `ttsc@0.26.2` remains extremely slow with Typia 13.2 versus 13.1.1. The preceding measurement in the same thread reports a Vitest run rising from 12.59 s to 313.53 s, with transform/import time dominating.

This is not the build-scoped whole-project validation defect fixed by #980, nor the per-delivery graph-index derivation fixed by #1009. Both fixes are present in 0.26.2.

## Verified cause

Vite `serve` (and therefore Vitest's Vite-backed transform path) calls `resetTtscTransformCache`, selecting the persistent correctness mode. On every cache hit, `matchesCachedSource` currently:

1. walks and hashes every project input; and
2. rereads and hashes the union of every out-of-walk graph member and plugin dependency in the transform envelope.

The work is therefore paid once per delivered module, even when most inputs cannot affect that module.

A platform-preserving run of the committed Unplugin performance fixture measured:

| Modules | External graph inputs | Synchronous reads | Reads/module | Delivery time |
| ---: | ---: | ---: | ---: | ---: |
| 20 | 0 | 540 | 27 | 0.374 s |
| 20 | 50 | 1,940 | 97 | 1.650 s |
| 20 | 100 | 2,940 | 147 | 2.871 s |
| 20 | 250 | 5,940 | 297 | 4.824 s |
| 50 | 100 | 10,350 | 207 | 10.8 s |

This establishes a current linear per-module cost in the exact persistent path. It does not claim the private customer project has the same graph shape, but it provides a repository-owned reproduction of the reported failure class.

The dedicated experiment also currently aborts on Windows before its serve-mode scenario: Scenario C mutates the process-global `process.platform` after cache state exists, which changes path identity and asks a Windows-produced binary to satisfy a Darwin-shaped delivery. Normal hosts cannot change `process.platform`; this is a regression-gate defect, not a production platform transition.

## Required behavior

Persistent hosts must retain freshness while validating only inputs that can affect the requested file:

- the current in-memory source;
- the file's reachable reference-graph closure;
- globals and the config chain;
- superseding resolution candidates;
- plugin-reported dependencies, honoring `dependenciesComplete`;
- volatile-file semantics.

A new or changed input that can affect a lazily delivered file must still invalidate before that file's first post-change delivery. Build-scoped adapters must keep their current first-delivery optimization.

## Acceptance criteria

- Persistent cache-hit validation is bounded by the delivered file's relevant input set rather than the union of the whole transform envelope.
- Edits to reachable sources, ambient/global inputs, configs, resolution candidates, and declared plugin dependencies invalidate the correct file.
- Unreachable sibling/external edits do not force unrelated per-module validation or transformation.
- Mixed `dependenciesComplete` and `volatile` envelopes preserve their documented conservative semantics.
- Vite serve/Vitest, Turbopack, Metro, Bun, and build-scoped adapters retain their respective lifecycle contracts.
- The Windows performance fixture no longer mutates global platform identity and reaches every scenario.
- A committed regression measurement asserts bounded per-module reads/work without publishing machine-specific wall-time as a benchmark.

## Coordination

This issue was discovered in the cache lifecycle/performance campaign opened from #970 and #1064.

### LINKED ISSUE #1185: fix(ttsc): give the ttsc-owned Go build cache a bounded lifecycle
## Problem

The follow-up in https://github.com/samchon/ttsc/issues/1064#issuecomment-5255950844 reports a Vite + Typia workspace where `node_modules/.cache/ttsc/go-build` grew by about 2.9 GB per cold rebuild and reached 9.9 GB, while `plugins/` remained about 30 MB. Deleting only `go-build` preserved a 3.3-second warm transform with no source-plugin rebuild.

The report matches the current ownership boundary.

## Verified cause

`resolveSourceBuildCachePaths` defaults to:

- `<root>/plugins`: content-addressed source-plugin executables;
- `<root>/go-build`: a ttsc-selected Go object cache used only when an executable misses.

`maybePrunePluginCache` applies the 30-day/LRU/2 GB policy only to `plugins/`. A local control seeded one stale plugin entry and one stale Go object; the same maintenance call removed the plugin entry and preserved the Go object.

Go 1.26's own disk-cache trim:

- runs from Go cache close, so an idle cache is never maintained;
- scans at most daily;
- removes entries unused for five days (with its mtime approximation);
- has no size ceiling.

Version/toolchain/plugin churn can therefore keep several multi-gigabyte object generations resident. The predecessor machine-global layout was moved into the workspace by #327 after issue #326 reported growth to hundreds of gigabytes, but the Go-object lifetime itself remained unbounded.

The public CI contract compounds the consequence. `ttsc cache paths --json` calls `cacheableRoots` the minimal set needed to skip cold plugin rebuilds, but it includes `go-build` whenever nested under the root. The compiled executable under `plugins/` is sufficient to skip that rebuild; `go-build` is an optional accelerator for the next content-key miss. The cache guide also describes automatic age/LRU cleanup without making clear that it covers only `plugins/`.

## Required behavior

The default, ttsc-owned Go object cache must have a bounded and concurrency-safe lifecycle. ttsc must never prune or delete a user-provided `GOCACHE`. Explicit `TTSC_CACHE_DIR`, `--cache-dir`, and `TTSC_GO_CACHE_DIR` ownership must remain deliberate and documented rather than inferred.

The solution must avoid turning a healthy single cold-build cohort into a repeated cold-build loop and must not delete entries while another source-plugin build is using them.

The CLI/docs must distinguish:

- the executable cache required to avoid a cold source-plugin build; and
- the Go object cache that only accelerates a future executable-cache miss.

## Acceptance criteria

- Repeated dependency/toolchain churn cannot grow the default ttsc-owned `go-build` cache without a documented bound.
- At least one useful current build cohort can remain warm under the policy; crossing a limit does not guarantee another full cold build on every invocation.
- Concurrent source-plugin builds and maintenance cannot corrupt builds or delete an active cache generation.
- Ambient/user `GOCACHE` is never maintained or removed by ttsc.
- Explicit cache-root ownership and `ttsc clean` behavior stay compatible and are covered for nested and external Go caches.
- `ttsc cache paths` exposes a clear stable distinction between required executable-cache paths and optional Go-build accelerator paths (with compatibility considered for existing JSON consumers).
- The cache guide and short compile guide state exactly which subcache is automatically pruned, how the Go cache is bounded, and what CI should persist.
- Tests seed stale/oversized Go cache state and cover default, explicit, user-owned, and concurrent cases without depending on the developer machine's real cache.

## Coordination

This issue was discovered in the cache lifecycle/performance campaign opened from #970 and #1064.

### LINKED ISSUE #1186: perf(ttsc): reuse unchanged GOROOT content identities across source-plugin cache keys
## Problem

Every source-plugin cache lookup computes the binary's content key before it can test whether the cached executable exists. The key intentionally includes the effective Go toolchain by content, but the current GOROOT implementation recursively rereads the selected standard-library/tool corpus on every `computeCacheKey` call.

An unchanged warm executable therefore still pays multi-second toolchain hashing once per source plugin.

## Reproduction

In one Node process on the current `0.26.2` baseline, a minimal unchanged source plugin called `computeCacheKey` three times with the same installed Go compiler and environment:

| Call | Elapsed | Key |
| ---: | ---: | --- |
| 1 | 2.774 s | `a8ba7e4a827827e7649b2e101179c692` |
| 2 | 3.011 s | same |
| 3 | 2.918 s | same |

The committed Unplugin performance fixture independently records about 139.4 MiB read for every otherwise-warm transform generation, constant across 10?100 project modules.

## Correctness constraint

The full content identity is deliberate. PR #93 made source-plugin executables shareable by content across install paths and added tests that require:

- equal GOROOT contents at different paths to share a key; and
- an in-place standard-library source edit in a long-lived process to change the next key.

Removing GOROOT from the key or memoizing only by its path would reintroduce stale/cross-toolchain binaries.

There is already a sound precedent for the Go compiler executable: a per-process memo is keyed by resolved path plus a cheap size/mtime signature, so unchanged content is reused and an in-place rewrite invalidates it. GOROOT currently has no corresponding per-file identity cache.

## Required behavior

Preserve the exact content-derived cache-key contract while avoiding repeated reads of every unchanged GOROOT file. File identity reuse must be scoped to the effective toolchain/environment and must rehash any member whose relevant metadata changes during a long-lived host.

## Acceptance criteria

- Repeated unchanged `computeCacheKey` calls in one process do not reread the GOROOT corpus byte-for-byte.
- A GOROOT file created, removed, renamed, or changed in place invalidates the reused identity and changes the key as appropriate.
- Identical toolchain contents installed at different paths continue to share the same source-plugin key.
- Different effective Go build environments/wrappers do not borrow each other's identity accidentally.
- Existing key bytes remain stable for unchanged inputs; deployed warm executable caches are not invalidated solely by the optimization.
- A deterministic regression test counts file reads/bytes or rehashes rather than asserting machine-specific wall time.
- The optimization applies to multiple source plugins in the same long-lived compiler/bundler process.

## Coordination

This issue was discovered alongside the customer cache/performance follow-ups in #970 and #1064.

### LINKED ISSUE #1190: Native utility config imports are absent from persistent cache inputs
## Problem

The persistent-cache invalidation added for #1184 tracks the selected `banner.config.*` / `strip.config.*` path and all discovery candidates, but not modules evaluated by those configs.

Both native plugins evaluate script configs in a child process:

- `packages/banner/driver/banner.go::loadBannerScriptConfigFile`
- `packages/strip/driver/config.go::loadStripScriptConfigFile`

A config such as:

```js
const selection = require(../shared/banner-selection.cjs);
module.exports = { text: selection.text };
```

therefore depends on `banner-selection.cjs`, but the JavaScript descriptor can only report the config path. The native transform output has no channel for the files its config loader actually evaluated. If the helper sits outside the TypeScript project walk, editing it leaves a long-lived unplugin generation valid and serves the old banner/strip result.

The same gap applies to TypeScript/ESM config imports. It is not sound to evaluate the config a second time in the JavaScript descriptor merely to discover dependencies: factories/configs may have side effects or environment-dependent behavior, and the second evaluator could disagree with the native one.

This was found while tracing the consequence surface of #970 comment 5258998680 and #1184 during the campaign Overall Self-Review.

## Expected direction

- Let native linked plugins report absolute universal host inputs they actually consumed.
- Carry those paths in the native transform JSON envelope and merge them with JavaScript-host inputs.
- Instrument banner/strip script and TypeScript config evaluators so the exact runtime file graph is reported without a second config evaluation.
- Preserve stdout protocol integrity and existing config error behavior.
- Add persistent-cache regressions proving an external imported config helper edit replaces the generation and changes output.
- Document the native host-input side channel in the plugin protocol.

### LINKED ISSUE #1196: fix(cache): preserve empty graph leaf adjacency through the JavaScript boundary
# Preserve empty graph leaf adjacency through the JavaScript boundary

## Problem

TypeScript-Go publishes every source file as a reference-graph node, including leaf entries shaped as `edges[file] = []`. The JavaScript API decoder drops those empty entries, so an import-free root file disappears from the node universe even though its compiler-time hash and realpath proof remain in the envelope.

`@ttsc/unplugin` binds input proof only to structural graph members. A leaf removed by decoding cannot contribute its compiler-time proof. During an A-to-B-to-A edit, equal pre/post snapshots for A can therefore be paired with output compiled from B and accepted as a stable persistent generation.

Expected behavior is to preserve every well-formed adjacency-array key, including an empty or string-filtered-to-empty array, while continuing to drop malformed non-array entries.

## Evidence

Verified at PR #1187 HEAD `91ce971828228ac4ee6f1383217e55ad86e8fa3b`.

- `packages/ttsc/driver/graph.go::BuildTransformGraph` explicitly retains leaf entries so `attachInputProof` can bind every source file.
- `packages/ttsc/src/compiler/internal/transformProjectInMemory.ts::parseDependencyLists` keeps a key only when at least one string element survives filtering.
- `parseReferenceGraph` reuses that parser for `graph.edges`.
- `packages/unplugin/src/core/transform.ts::envelopeGraphIndexes` constructs membership from structural graph fields and binds proof only to identities already in that membership set.
- The proof completeness check compares against the already-shrunk member set and therefore cannot detect the omitted leaf.

An actual tsgo transform over two unrelated roots, `src/main.ts` and `src/isolated.ts`, produced:

```json
{
  "edges": {},
  "hashes": [
    "src/isolated.ts",
    "src/main.ts",
    "tsconfig.json"
  ]
}
```

The current A-to-B-to-A tests do not expose this boundary because their source files are interconnected. A leaf remains a member as another source's target.

## Consequence surface

| Case | Current result | Expected result |
| --- | --- | --- |
| Source with dependencies | Preserved as a key or target | Unchanged |
| Independent root leaf with `edges[file] = []` | Key removed | Key and empty array preserved |
| Edge value is not an array | Entry removed | Unchanged |
| Array contains only non-string elements | Key removed | Well-formed array key preserved with `[]` |
| Path exists only in proof maps | Not promoted to a graph member | Unchanged |
| Independent leaf changes A-to-B-to-A during compilation | Stale output can qualify for persistent reuse | Torn generation is not reusable |

## Approach

Give `graph.edges` a parser distinct from the dependency-list parser. The graph parser retains every array-valued key and filters only its elements. Do not promote arbitrary proof keys into membership. The structural graph remains the node-universe owner.

- A JavaScript API integration test proves that unrelated root sources retain exact `[]` edge entries.
- Malformed non-array edges are still dropped.
- An array containing only non-strings retains its key as `[]`.
- An unplugin integration test proves that an independent root leaf changed A-to-B-to-A during native compilation cannot serve stale sibling output and replaces the generation.
- Existing connected graphs, optional candidates, globals/configs, graph-less sidecars, and proof-tolerant compatibility remain unchanged.
- Focused API and unplugin regressions, `pnpm test:typecheck`, and the owning feature lanes pass.

## Coordination

This is a correctness correction for PR #1187 and #1184. Proof-key promotion and fixture-specific branches are excluded.

### LINKED ISSUE #1197: fix(ttsc): create cleanup-owned temporary directories from canonical parents
# Create cleanup-owned temporary directories from canonical parents

## Problem

PR #1187's `packages/ttsc/src/plugin/internal/loadProjectPlugins.ts::createEvaluationTempDir` calls `mkdtempSync` through the ambient `TEMP` or `TMPDIR` spelling, then canonicalizes the returned child. The parent alias can be retargeted between those operations. If the new target contains the same random basename, later writes and recursive cleanup bind to a different directory from the one the process created.

Expected behavior is to freeze the parent identity before creation and keep creation, writes, and cleanup bound to one verified physical direct child.

## Evidence

Verified at PR #1187 HEAD `91ce971828228ac4ee6f1383217e55ad86e8fa3b`.

The current order is:

```ts
const directory = fs.mkdtempSync(path.join(os.tmpdir(), prefix));
return fs.realpathSync.native(directory);
```

A deterministic Windows junction schedule reproduced the ownership switch without monkey-patching:

1. Create `alias -> A`.
2. Call `mkdtempSync(alias/descriptor-)`, creating the child under A.
3. Retarget the alias to B and create the same child basename under B.
4. Call `realpathSync.native` on the original lexical child.

Observed:

```text
createdPhysicalChild = ...\a\descriptor-hUzy8v
existsInA = true
resolvedAfterRetarget = ...\b\descriptor-hUzy8v
resolvesIntoB = true
```

If B lacks the same basename, canonicalization fails and leaks the A child. If B has it, subsequent operations switch ownership to B. The unplugin scratch implementation in the same PR already demonstrates the intended parent-first invariant.

## Consequence surface

| State | Current risk | Expected result |
| --- | --- | --- |
| Stable physical temp parent | Normal | Unchanged |
| Symlink or junction parent | Retarget between create and canonicalize | Parent frozen before create |
| New target lacks the child | Canonicalization failure and leak | Alias retarget is irrelevant |
| New target has the same child | Writes and cleanup switch targets | Original physical child only |
| Postflight failure | Alias-spelled cleanup can be unsafe | Clean only a verified physical direct child |
| Windows junction, drive alias, or case alias | Lexical containment is insufficient | Physical identity and direct containment |
| POSIX symlink | Same retarget class | Same invariant |

## Approach

- Canonicalize the parent with `realpathSync.native` before calling `mkdtempSync`.
- Create under that physical parent.
- Verify the child's postflight physical identity is a direct child of the frozen parent.
- Use only the verified physical spelling for writes and recursive cleanup.
- Converge ttsc call sites on one internal helper. Package-local tools that cannot cross package boundaries must implement the same invariant locally.

- A deterministic test observes `realpath(parent) -> mkdtemp(physical parent) -> postflight(child)` through an explicit operations seam or equivalent mechanism.
- Windows junction and POSIX symlink parents return a child of the original physical parent.
- Retargeting the lexical alias after creation cannot remove a same-named directory under the new target.
- Non-directory parents and postflight mismatches fail closed.
- Normal and exceptional cleanup remains correct for descriptor evaluation, compiler output, LSP manifests, runtime orphan emit, source-plugin scratch, and every converted caller.
- Focused API, native-plugin, launcher, and Windows CI lanes pass.

## Coordination

This is a temporary-storage ownership correction for PR #1187. Canonicalizing only the child after creation or re-resolving a lexical cleanup path does not close the race.

### LINKED ISSUE #1198: refactor(test): replace shared node:fs monkey-patching with explicit filesystem operations
# Replace shared node:fs monkey-patching with explicit filesystem operations

## Problem

Cache performance and race regressions replace methods on the imported `node:fs` singleton to count calls or inject failures. The repository development contract forbids monkey-patching and requires the general logic's real ownership boundary to be fixed.

`node:fs` is shared through the module cache. Replacing `readFileSync`, `statSync`, `lstatSync`, `readdirSync`, `existsSync`, or `realpathSync.native` affects every consumer in the process. `finally` restoration and today's serial executor reduce exposure but do not make nested instrumentation, parallel tests, early termination, or restoration to the wrong wrapper safe.

Expected behavior is to expose the actual filesystem ownership boundaries as typed operations and inject independent counting or fault-injection implementations.

## Evidence and provenance

Verified across PR #1187 base `50934aef753459e456743f5f99af3736ce0bd27b` and HEAD `91ce971828228ac4ee6f1383217e55ad86e8fa3b`.

Newly introduced by the PR:

- `tests/test-unplugin/src/internal/transform-project-cache.ts`: `readFileSync`, `lstatSync`, `statSync`, and four `readdirSync` wrappers. Introducing commits include `82231935f`, `56208f7d1`, and `e6b81db96`.
- `tests/test-ttsc/src/native-plugins/source-plugin/test_computecachekey_memoizes_unchanged_goroot_content_reads.ts`: `readFileSync`, introduced by `4a6b5e2a6`.
- `experimental/unplugin-perf/src/index.ts`: `statSync`, introduced by `56208f7d1`.

Existing debt found during the recheck:

- The experiment's `readFileSync` wrapper predates the base at `15ae101e86`.
- The experiment's `existsSync` and `realpathSync.native` wrappers and the unplugin test's identity-probe wrapper predate the base at `1dab1e6c05`.

Therefore the original review statement that every wrapper was new is incorrect. The PR nevertheless expands the prohibited pattern. A current source-tree search finds actual filesystem monkey-patching only in these three files, excluding intentional lint-rule input fixtures.

## Consequence surface

| Execution shape | Current risk | Expected result |
| --- | --- | --- |
| One serial instrument | Usually restored | Same measurements |
| Nested instruments | Restoration can leave a stale wrapper | Independent operation instances |
| Parallel tests | Counts and injected faults mix | Call-local isolation |
| Assertion or forced exit | Shared singleton can remain modified | No global mutation |
| Fault injection | Coupled to global timing and path strings | Injected at the owning operation |
| Performance measurement | Harness reads can enter the count | Count only the measured boundary |
| Default production use | Ambient `node:fs` | No behavior or public API change |

## Approach

- Give cache validation, project walking, path identity, and GOROOT content identity the smallest typed filesystem operations each owns.
- Use the current `node:fs` calls as immutable production defaults.
- Let a cache or call context own its operations instance so concurrent transforms cannot interfere.
- Give tests independent counting and fault-injection wrappers without fixture-name or expected-value branches in production.
- Make the performance harness use the same explicit seam.
- Remove the same-pattern base debt in the three affected files while the boundary is being corrected.

- Source search reports no `node:fs` method assignment outside intentional lint-rule fixture text.
- Bounded validation, incomplete snapshot recovery, compile/snapshot races, and A-to-B-to-A regressions inject behavior through explicit operations and still pass.
- The GOROOT memo test proves cold reads, unchanged zero rereads, and edit/add/rename/delete rereads without global patching.
- The performance harness measures read, stat, and identity-probe budgets without global patching.
- Existing adapters, Metro, and ttsc API behavior stay unchanged with default operations.
- A negative isolation test proves two operation instances do not share counts or failures.
- Typecheck, unplugin feature, source-plugin native, and experimental lanes pass.

## Coordination

This is a test and performance architecture correction for PR #1187. Moving the same global mutation into a child process or adding fixture-specific product branches is excluded.

samchon:
## Summary

Cache lifecycle, persistent transform correctness, and source-build storage campaign for the customer follow-ups in:

- https://github.com/samchon/ttsc/issues/970#issuecomment-5258998680
- https://github.com/samchon/ttsc/issues/1064#issuecomment-5255950844

Closes #1184
Closes #1185
Closes #1186
Closes #1190
Closes #1196
Closes #1197
Closes #1198

This branch addresses the verified consequence surface rather than only the reported symptoms:

- bounds persistent unplugin validation by the requested file graph, exact host inputs, and generation-scoped membership tracking;
- records compiler-time content and physical-path proof for transform graph members, including selected lexical symlink and junction spellings, so in-project and external A-to-B-to-A races cannot authorize stale output;
- preserves independent graph leaves and rejects empty graph source keys while retaining conservative compatibility for graph-producing sidecars that do not publish optional proof;
- isolates and fingerprints Node, Bun, TypeScript, CommonJS, ESM, and utility-plugin descriptor inputs without purging application-owned module state;
- resolves runtime binaries and plugin configuration from the project execution context;
- bounds default plugin-binary and Go object caches while protecting active build generations and the binary just published;
- creates scratch directories below verified canonical parents and refuses destructive cleanup when ownership cannot be proven after alias retargeting;
- binds filesystem observation, path identity, failure injection, and counters to each transform cache instead of mutating shared `node:fs` methods;
- keeps first-party banner, strip, and lint configuration discovery and evaluated dependencies exact;
- retains the exact no-fix website-only extract-zip audit waiver with expiry checks.

## Review corrections

The exhaustive solo review found four concrete issue classes and the follow-up Individual reviews found three same-cause gaps. All are resolved in this branch:

- graph decoding now preserves valid empty adjacency arrays as independent nodes, validates non-empty source keys, and covers real TypeScript-Go leaf and persistent A-to-B-to-A scenarios;
- scratch creation canonicalizes and verifies the parent before `mkdtemp`, validates the returned child, and never removes an unverified postflight path;
- test and performance instrumentation uses cache-owned operation seams, with no assignment to imported `node:fs` methods;
- path-identity `realpath`, directory enumeration, metadata reads, platform policy, and case policy travel with the cache across concurrent transforms;
- the cache-local regression overlaps two project builds and proves that one cache's post-compile `readdir` failure and observations never reach the other;
- the remaining format drift and obsolete helper were removed.

## Validation

Passed on Windows on the final implementation series through `70da3d0c0`:

- `ttsc` build;
- `@ttsc/unplugin` declaration build and Rollup bundle;
- `@ttsc/lint` and graph package builds;
- workspace and affected test-package type checks;
- focused Go driver graph, observation, resolution-candidate, and GOROOT memoization tests;
- focused ttsc API, ttscserver, plugin-cache, Go-cache, canonical scratch, and path-identity tests;
- the complete `transformttsc` feature suite, including persistent validation, graph leaves, external inputs, linked paths, concurrent failure paths, volatile inputs, and A-to-B-to-A races;
- concurrent cache-local filesystem-operation, watch-bound, per-file validation, and path-identity regressions;
- the experimental core-backed cache performance harness under its frozen workload;
- formatting and diff whitespace checks.

All 28 GitHub checks succeeded against the exact final head. The final solo Overall Review reported no actionable finding, and the PR is ready for review. It is not authorized for merge by this campaign.


samchon (commented):
Individual Self-Review ? commit `271b516c9` (`fix(cache): complete native resolution probes`)

Result: one actionable Medium finding.

The package-main, dotted CommonJS request, and `NODE_PATH` fixes were independently reproduced and were correct. The review found one additional Node-compatibility boundary: Node accepts a UTF-8 BOM in `package.json`, while all four embedded banner/strip loaders parsed the raw text with `JSON.parse`. A BOM-bearing manifest could therefore select `entry.json` while omitting the higher-priority missing `entry.js` input.

Disposition: accepted and fixed by follow-up commit `ad239b616`, which strips the leading BOM in all four loaders and converts the package-main regression to a BOM-bearing manifest. The focused persistent utility integration test and the real Go utility plugin suite pass with the follow-up.

samchon (commented):
Individual Self-Review - commit `2221d215d` (`fix(ttsc): bound the default Go build cache`)

Result: actionable findings.

The daily maintenance marker was written before a cold build populated the cache, so the cache could grow beyond the nominal limit for the rest of the 24-hour window. Coordination records also tied liveness too strongly to process identity and stale age, leaving cleanup-failure and orphan-child orderings unsafe.

Disposition: accepted. Follow-ups `ae3f86681` and `584d3f498` added forced post-attempt pruning, independent task heartbeats, completion records, bounded clock-skew handling, and failure/concurrency regressions.

samchon (commented):
Individual Self-Review - commit `82231935f` (`fix(unplugin): bound persistent cache validation`)

Result: actionable findings.

The per-file graph validation removed the reproduced whole-envelope content-read scaling, but project-directory membership was still statted per delivered module, an incomplete generation walk could authorize narrow reuse, and universal descriptor/config inputs were not represented as exact host-owned inputs.

Disposition: accepted. Follow-up `56208f7d1` added exact host inputs, incomplete-snapshot fallback, bounded directory-validation coverage, and the associated positive and recovery tests.

samchon (commented):
Individual Self-Review - commit `da0e63930` (`fix(unplugin): harden persistent cache generations`)

Result: one actionable High finding.

A generated overlay tsconfig was reported as a universal host input and then deleted in `finally`. Its live tracker and persistent validation therefore invalidated every otherwise stable generation using compiler-option or alias overlays.

Disposition: accepted. Follow-up `71f976dbc` excluded the disposed overlay by canonical filesystem identity from persistent host tracking and adapter watch registration.

samchon (commented):
Individual Self-Review - commit `4a6b5e2a6` (`perf(ttsc): memoize unchanged GOROOT identity`)

Result: actionable metadata-stability finding.

The content memo preserved in-place SDK mutation correctness, but unconditional executable permission repair before key computation could rewrite ctime and defeat the unchanged-metadata fast path.

Disposition: accepted. Follow-up `0f25ed6db` made permission repair metadata-preserving when no repair is needed and extended the production-entry memo regression.

samchon (commented):
Individual Self-Review - commit `ae3f86681` (`fix(ttsc): harden Go cache maintenance`)

Result: actionable findings.

A failed cold `go build` could populate GOCACHE and bypass the success-only post-build prune; Worker construction failure could leak a fresh lease; completed or future-dated maintenance intents could block builders for the long stale interval; and one explicit-root test asserted a nested path instead of the actual user-owned root.

Disposition: accepted. Follow-up `584d3f498` moved pruning into the build-attempt `finally`, made heartbeat startup failure-safe with an IPC fallback, added explicit completion/staleness semantics, and corrected the ownership-boundary tests.

samchon (commented):
Individual Self-Review - commit `0f25ed6db` (`fix(ttsc): preserve memoized GOROOT metadata`)

Result: one actionable Medium security finding.

The metadata-preserving repair used `mode | 0111`, turning a writable extracted tool such as 0666 into 0777 and preserving already-executable unsafe or special-bit modes. This could immediately spawn a group/world-writable compiler.

Disposition: accepted. Follow-up `ae8587ff3` normalized bundled toolchain files to the repository packaging contract of exact 0755.

samchon (commented):
Individual Self-Review - commit `56208f7d1` (`fix(unplugin): harden narrow cache validation`)

Result: actionable findings.

The event-turn membership memo could authorize a stale generation across filesystem mutation ordering, and the exact universal-input model still needed generation-owned liveness plus descriptor/config reload coverage rather than a transient directory-stat shortcut.

Disposition: accepted. Follow-up `da0e63930` replaced the shortcut with generation-owned mutation tracking, hardened generation disposal and race handling, and added persistent descriptor/config invalidation tests.

samchon (commented):
Individual Self-Review - commit `ae8587ff3` (`fix(ttsc): normalize Go tool permissions`)

Result: actionable ownership finding.

Exact 0755 normalization was correct for ttsc-bundled toolchains, but applying it to an explicitly selected external Go installation rewrote user-owned restrictive permissions and special mode choices.

Disposition: accepted. Follow-up `031393602` separated bundled normalization from the minimal owner-execute repair for explicitly selected toolchains.

samchon (commented):
Individual Self-Review - commit `584d3f498` (`fix(ttsc): harden Go task ownership`)

Result: CLEAN.

The review covered failed-build pruning, lease and maintenance completion, heartbeat denial fallback, future/stale/orphan coordination, bundled versus external permissions, explicit cache-root ownership, and corresponding documentation/tests. No actionable defect survived verification.

samchon (commented):
Individual Self-Review - commit `031393602` (`fix(ttsc): preserve external Go tool permissions`)

Result: actionable provenance finding.

Using only the presence of `TTSC_GO_BINARY` to decide ownership still classified implicitly discovered user SDKs as bundled and could normalize their files. Tool origin, not merely explicit selection, must own the permission policy.

Disposition: accepted. Follow-up `584d3f498` carries an explicit `bundled` provenance bit through compiler resolution and build execution; restrictive external modes remain intact.

samchon (commented):
Individual Self-Review - commit `71f976dbc` (`fix(unplugin): exclude temporary host inputs`)

Result: one actionable High finding.

When the host `TEMP` or `TMPDIR` pointed inside the project, the generated scratch tree was created before the project snapshot and entered project hashes/watchers even though the exact tsconfig was excluded from host inputs. Deleting that tree forced every later request to recompile.

Disposition: accepted. Follow-up `68d02e2ad` moved transform scratch outside the project and routed compiler/plugin temporary work through that owned root.

samchon (commented):
Individual Self-Review - commit `787b94524` (`fix(unplugin): canonicalize transform scratch roots`)

Result: one actionable High data-loss finding.

The code validated `canonicalDirectory` but returned the mutable alias-spelled directory. Retargeting its parent junction after validation could redirect final recursive cleanup into a same-named project path. A Windows reproduction deleted the project sentinel and left the original scratch behind.

Disposition: accepted. Follow-up `6b25f65b6` pinned the canonical directory for compiler writes, environment, and cleanup.

samchon (commented):
Individual Self-Review - commit `68d02e2ad` (`fix(unplugin): isolate transform scratch storage`)

Result: actionable findings.

Scratch isolation only applied when an overlay tsconfig existed, so ordinary no-overlay transforms could still churn a project-local temp tree. The outside-project check was also lexical and could be bypassed by a symlink, junction, or Windows alias into the project.

Disposition: accepted. Follow-up `787b94524` applied scratch isolation to every transform and canonicalized both project and candidate identities with Windows junction/alias regressions.

samchon (commented):
Individual Self-Review - commit `6b4db52ed` (`fix(unplugin): reload universal host inputs`)

Result: actionable findings.

The descriptor graph was captured before factory/getter lazy requires, failed roots could strand poisoned child cache, recursive purge could split application singletons, and external universal inputs were rehashed per module. The lint discovery change also had a TypeScript narrowing error and treated a directory as a valid config stop, omitting the selected ancestor.

Disposition: accepted. Follow-up `e6b81db96` isolated descriptor evaluation, separated current-byte influence tracking from host cache ownership, introduced metadata/missing-probe validation, fixed lint discovery/type narrowing, and added the reproduced failure/performance tests.

samchon (commented):
Individual Self-Review - commit `b5c63aecc` (`fix(cache): close universal input edge cases`)

Result: actionable findings.

A TypeScript descriptor module-body exception was still classified as loader incompatibility and could execute side effects twice. The review also required direct overlay performance coverage, public documentation of isolated descriptor environment/output behavior, and complete test scenario comments.

Disposition: accepted. Follow-up `79674907b` narrows ttsx retries to verified loader failures, adds the top-level side-effect regression and overlay validation bound, and updates the protocol/API documentation and test contracts.

samchon (commented):
Individual Self-Review - commit `79674907b` (`fix(cache): report native config inputs`)

Result: one actionable Medium finding.

Banner and strip reported final native config module URLs but not higher-priority resolution candidates. Creating a nearer bare package after the first generation left the same cached generation and stale output; relative extensionless imports had the analogous risk.

Disposition: accepted. Follow-up `3c97f46b7` records resolution candidates for banner and strip CJS/TS loaders and adds persistent superseding-package and extensionless-import regressions.

samchon (commented):
Individual Self-Review - commit `cd5c3c850` (`fix(unplugin): create scratch from canonical roots`)

Result: CLEAN.

The review covered creation, postflight rejection, canonical cleanup, project-local temp aliases, Windows junctions and 8.3 identities, overlay and no-overlay reuse, and final recursive removal. No actionable defect survived verification.

samchon (commented):
Individual Self-Review - commit `6b25f65b6` (`fix(unplugin): pin canonical scratch paths`)

Result: one actionable Medium cleanup finding.

The successful path was pinned, but a rejected postflight candidate still removed the alias-spelled path. Retargeting the alias could delete a decoy same-name directory or strand the actual created directory.

Disposition: accepted. Follow-up `cd5c3c850` creates scratch from the already canonical candidate and cleans a rejected result only through its postflight canonical identity.

samchon (commented):
Individual Self-Review - commit `3c97f46b7` (`fix(cache): track native resolution candidates`)

Result: three actionable Medium findings.

The first implementation omitted extension probes for `package.json main`, treated a dotted CommonJS request as already complete even though Node appends extensions, and omitted preceding `NODE_PATH` and global search roots.

Disposition: accepted. Follow-up `271b516c9` recursively records package-main candidates, probes dotted requests with CommonJS extensions, uses `createRequire(parentURL).resolve.paths()` in Node order, and adds all three persistent-generation regressions.

samchon (commented):
Individual Self-Review - commit `e6b81db96` (`fix(cache): harden persistent invalidation`)

Result: actionable findings.

The universal manifest accidentally re-included the disposed overlay tsconfig; isolated TypeScript evaluation used ambient rather than effective env and exposed stdout; factory failures could be retried through ttsx; and descriptor getters could require a file after the input list was snapshotted.

Disposition: accepted. Follow-up `b5c63aecc` excludes the overlay consistently, preserves effective environment and protocol stderr, classifies loader-only retries, and serializes getters while dependency collection remains active.

samchon (commented):
Individual Self-Review - commit `ad239b616` (`fix(cache): parse native package manifests like Node`)

Result: CLEAN.

The review verified all four embedded banner/strip CJS/TypeScript loaders, Go raw-string and emitted-regex escaping, and Node's exact package-manifest BOM boundary. One leading U+FEFF is accepted, while two BOMs or whitespace before a BOM remain invalid; JSON-string content is untouched. The regression fails before the fix by omitting `entry.js`, then proves host-input capture, generation replacement, and new output after that candidate appears.

Focused persistent utility integration passed, and the parent-to-commit diff check was clean.

samchon (commented):
Individual Self-Review — fb436e57f — CLEAN

The exact extract-zip waiver commit was reviewed against the official advisory, the live dependency graph, Marp's HTML-only execution path, waiver expiry logic, and the dependency-audit tests. No actionable finding was found. The waiver is exact-ID scoped, applies only while the advisory reports no patched version, and does not suppress unrelated severe advisories. Validation: dependency-audit tests 14/14 and the live dependency audit passed.

samchon (commented):
﻿Individual Self-Review of 457e2935952be0b1778610b09d2c3d64e412450e found four actionable gaps:

1. High: Bun's static ESM descriptor imports bypass Module._resolveFilename, so external dependencies are omitted from hostInputs and persistent transforms can remain stale.
2. Medium: retry classification trusts a user-mutable error.code; a descriptor can execute side effects and then be run again through ttsx. The test currently fabricates exactly that code and therefore enshrines the duplicate-execution path.
3. Medium: a direct attempt that is legitimately retried prints its stack before a successful fallback, making a successful build look failed.
4. Medium: the ttsx fallback is launched through the selected Bun runtime even though ttsx's runtime hooks require Node's module.registerHooks.

The reviewer reproduced each boundary under Bun/Node and confirmed the focused tests, target typecheck, and diff check otherwise pass. These findings are accepted and will be corrected before the PR is made ready.

samchon (commented):
Solo review follow-up for bca474df7

Accepted and fixed before this commit:

- Descriptor alias-ancestor capture walked and statted every resolved module path even when descriptor-input capture was disabled. The wrapper now exits before filesystem work unless both the capture flag and output path are active.
- Compiler-time graph proof was initially consumed only for out-of-project members. A true in-project A-to-B-to-A mutation between the two project snapshots could therefore pair transient compiler output with restored disk state. Persistent generation admission now verifies every graph member whenever the optional proof section is present, and the synthetic regression performs the mutation inside compilation rather than before the first snapshot.
- The full-proof rule retains compatibility for legacy sidecars that omit the optional section; ordinary in-project graphs keep pre/post snapshot validation, while out-of-walk graph members remain ineligible for narrow reuse without generation proof.

Focused graph, cache, linked-path, performance-bound, API, ttscserver, plugin-cache, lint build, unplugin build, ttsc build, and Go driver tests passed. This is an advisory review record; the complete fresh Overall Self-Review on 907b111fb remains in progress.

samchon (commented):
Solo review follow-up for 393135dc9..5c07a04f2:

- The compiler graph omitted leaf source keys, so an out-of-walk leaf could lack compiler-time hash/realpath proof and authorize an A-B-A generation. Leaf nodes now remain explicit empty adjacency entries, with driver and protocol coverage.
- Selected junction/symlink spellings were content-hashed as auxiliary resolution inputs on Windows. Ordinary dependency edits therefore escalated to a full compiler reload, while same-byte retargets lacked a physical-identity signal. Selected aliases now track existence plus final physical path; resident source hashes retain content ownership. The exact Windows CI failure and same-byte retarget regression pass repeatedly.
- packages/ttsc/driver gained a committed unit suite but the CI runner and ownership map named only packages/ttsc/test/driver. The runner now executes both and the ownership invariant names both.
- The performance harness wrote its run counter inside the measured project, making the observer invalidate every hardened project snapshot. The counter now lives outside the project, graph sidecars publish input proofs, Scenario D gates pluginRuns explicitly, and cleanup runs on success and failure.

Focused validation: driver runner PASS; test-owner gate PASS; Windows graph state tests 10/10 PASS; leaf proof test PASS; full A-D performance harness PASS with pluginRuns=1 and bounded per-module probes/reads; temp root removed.

samchon (commented):
## Overall Review — 수정 후 재검토 필요

### 결론

이 PR의 방향 자체는 타당합니다. 특히 다음 구현은 특정 소비자에 맞춘 특례가 아니라 ttsc의 일반 제품 경계로 설계되어 있습니다.

- JavaScript descriptor 입력을 hostInputs / hostInputHashes / hostInputRealpaths라는 공통 계약으로 전달
- linked Go plugin에도 동일한 입력 증명 채널 제공
- Node의 공식 module.registerHooks와 Bun plugin을 사용하고 descriptor 실행을 별도 프로세스로 격리
- Go plugin cache와 Go object cache의 소유권, physical root, LRU, lease/heartbeat를 분리
- typia/nestia 같은 특정 소비자 이름으로 제품 동작을 분기하지 않음
- 설정 로더 복제본을 CI invariant로 동기화

따라서 “ttsc 본연을 훼손하는 방향”이거나 소비자별 하드코딩으로 PR 전체를 폐기해야 하는 상태는 아닙니다. 다만 현재 HEAD에는 캐시 정합성을 실제로 깨는 High 결함 1개, 재귀 삭제 안전 보장을 깨는 High 결함 1개, 저장소의 명시적 몽키패칭 금지 계약을 위반하는 Medium 결함 1개가 남아 있습니다. 필수 CI도 별도로 실패 중이므로 현재 상태로는 병합하면 안 됩니다.

검토 범위는 base 50934aef753459e456743f5f99af3736ce0bd27b부터 HEAD 91ce971828228ac4ee6f1383217e55ad86e8fa3b까지 104 commits, 118 files, +14,540/-925 전부이며, 캐시/그래프/descriptor/runtime hook/Go build storage/유틸리티 plugin/CI/문서와 downstream 소비 경로를 함께 추적했습니다. 로컬 워크트리의 미커밋 20개 수정 및 2개 신규 파일도 별도 범위로 확인했습니다.

---

## 1. [High] 빈 adjacency entry를 디코더가 삭제하여 독립 루트 파일의 compiler-time proof가 무효화됩니다

Go 드라이버는 leaf source도 빈 edge로 남기는 것을 명시적인 정합성 계약으로 정의합니다.

- [driver/graph.go L75-L80](https://github.com/samchon/ttsc/blob/91ce971828228ac4ee6f1383217e55ad86e8fa3b/packages/ttsc/driver/graph.go#L75-L80): leaf를 edges[file] = []로 유지해야 그 파일이 graph node universe에 들어가고, 컴파일 중 A→B→A 변경을 compiler-time hash/realpath proof로 막을 수 있다고 설명합니다.

그런데 JavaScript API 디코더는 같은 wire shape를 보존하지 않습니다.

- [transformProjectInMemory.ts L759-L777](https://github.com/samchon/ttsc/blob/91ce971828228ac4ee6f1383217e55ad86e8fa3b/packages/ttsc/src/compiler/internal/transformProjectInMemory.ts#L759-L777): parseDependencyLists는 문자열 원소가 하나라도 있을 때만 key를 결과에 넣습니다. 따라서 정상적인 "isolated.ts": []도 삭제됩니다.
- [transformProjectInMemory.ts L792-L827](https://github.com/samchon/ttsc/blob/91ce971828228ac4ee6f1383217e55ad86e8fa3b/packages/ttsc/src/compiler/internal/transformProjectInMemory.ts#L792-L827): graph의 edges에도 같은 parser를 재사용하므로 Go가 의도적으로 보존한 leaf membership가 JS 경계에서 사라집니다.

이 손실은 표현상의 차이로 끝나지 않습니다.

- [unplugin transform.ts L617-L639](https://github.com/samchon/ttsc/blob/91ce971828228ac4ee6f1383217e55ad86e8fa3b/packages/unplugin/src/core/transform.ts#L617-L639): unplugin의 members는 edge key/target, globals, configs, candidates로 구성됩니다.
- [transform.ts L665-L709](https://github.com/samchon/ttsc/blob/91ce971828228ac4ee6f1383217e55ad86e8fa3b/packages/unplugin/src/core/transform.ts#L665-L709): input proof는 이미 members에 들어 있는 identity에만 결합됩니다. proof map에 leaf hash가 남아 있어도 leaf edge key가 없어지면 "if (!built.members.has(identity)) continue"에서 버려집니다.
- [transform.ts L1734-L1770](https://github.com/samchon/ttsc/blob/91ce971828228ac4ee6f1383217e55ad86e8fa3b/packages/unplugin/src/core/transform.ts#L1734-L1770): completeness 비교 역시 축소된 member 수를 기준으로 하므로 누락을 발견하지 못합니다.
- [transform.ts L2420-L2465](https://github.com/samchon/ttsc/blob/91ce971828228ac4ee6f1383217e55ad86e8fa3b/packages/unplugin/src/core/transform.ts#L2420-L2465): external snapshot 후보도 같은 구조적 graph member에서 나오므로 proof key만 남은 leaf는 여기서도 복구되지 않습니다.

### 실제 재현

현재 소스를 빌드한 뒤, import/reference 관계가 전혀 없는 src/main.ts와 src/isolated.ts 두 파일을 같은 tsconfig의 root로 넣고 new TtscCompiler({ plugins: false }).transform() 결과를 투영했습니다.

    {
      "type": "success",
      "edges": {},
      "hashes": [
        "src/isolated.ts",
        "src/main.ts",
        "tsconfig.json"
      ]
    }

즉 compiler-time proof에는 두 source가 모두 있지만, node universe인 edge key는 둘 다 없어졌습니다.

이 상태에서 독립 파일이 컴파일 전 A, 컴파일 중 B, 컴파일 후 다시 A가 되면 다음이 동시에 성립할 수 있습니다.

1. pre/post project snapshot은 동일한 A라 통과합니다.
2. native compiler output은 중간 B를 사용한 결과입니다.
3. 해당 leaf의 compiler-time B proof는 member가 아니어서 검증되지 않습니다.
4. 캐시는 A의 snapshot과 B의 output을 같은 stable generation으로 승인할 수 있습니다.

기존 compile_snapshot_aba 회귀 테스트가 통과하는 이유도 확인했습니다. 그 fixture는 파일들이 서로 edge로 연결되어 있어 leaf라도 다른 source의 target으로 member가 됩니다. “완전히 독립된 root file”이 빠져 있어서 이 경계를 검증하지 못합니다.

### 필요한 정석 수정

dependencies와 graph.edges는 의미가 다르므로 parser를 분리해야 합니다. graph 전용 parser는 다음을 지켜야 합니다.

- 값이 배열인 정상 entry는 필터 후 빈 배열이 되어도 key 자체를 보존
- malformed non-array entry만 제거
- 빈 edge key 하나라도 있으면 graph의 usable structure로 계산
- proof key를 임의로 member로 승격하는 우회는 금지. node universe는 구조적 graph가 소유

회귀 테스트는 최소 두 층이 필요합니다.

1. JS API 통합: 서로 참조하지 않는 두 root file에서 graph.edges["src/isolated.ts"]가 정확히 []인지 확인
2. unplugin 통합: 독립 root leaf가 native compile 안에서 A→B→A로 바뀔 때 첫 generation이 persistent reuse 자격을 얻지 못하고 재컴파일되는지 확인

---

## 2. [High] mkdtemp(alias) 후 realpath(child) 순서는 TEMP/TMPDIR retarget TOCTOU를 닫지 못합니다

[loadProjectPlugins.ts L3022-L3033](https://github.com/samchon/ttsc/blob/91ce971828228ac4ee6f1383217e55ad86e8fa3b/packages/ttsc/src/plugin/internal/loadProjectPlugins.ts#L3022-L3033)의 createEvaluationTempDir는 다음 순서입니다.

    const directory = fs.mkdtempSync(path.join(os.tmpdir(), prefix));
    return fs.realpathSync.native(directory);

주석은 retarget된 TEMP/TMPDIR alias가 나중의 recursive cleanup을 다른 위치로 돌리지 못한다고 보장하지만, 생성과 canonicalization 사이가 열려 있습니다.

가능한 순서는 다음과 같습니다.

1. TEMP alias가 물리 디렉터리 A를 가리킵니다.
2. mkdtemp(alias/prefix-XXXXXX)가 A 아래에 자식 prefix-XXXXXX를 만듭니다.
3. 두 syscall 사이에 alias가 B로 retarget되고, B 아래에 같은 basename의 디렉터리가 생깁니다.
4. realpath(alias/prefix-XXXXXX)는 A의 방금 만든 디렉터리가 아니라 B의 동명 디렉터리를 반환합니다.
5. 이후 descriptor 파일을 B에 쓰고, finally의 recursive rmSync도 B를 삭제합니다. A의 실제 temp는 유출됩니다.

B에 동명 자식이 없으면 realpath가 실패하여 누수만 남지만, 코드 주석의 “canonicalization 실패 시 작은 누수만 생긴다”는 보장은 동명 자식이 있는 경우에는 성립하지 않습니다. 이 PR은 mutable temp alias와 재귀 삭제를 명시적으로 threat model에 포함했으므로, 작은 race window라는 이유로 예외 처리할 수 없습니다.

같은 PR 안의 unplugin scratch 구현은 올바른 순서에 가깝습니다.

- [transform.ts L2741-L2761](https://github.com/samchon/ttsc/blob/91ce971828228ac4ee6f1383217e55ad86e8fa3b/packages/unplugin/src/core/transform.ts#L2741-L2761): candidate parent를 먼저 canonicalize하고 그 물리 parent 아래에서 mkdtemp합니다.
- [transform.ts L2766-L2787](https://github.com/samchon/ttsc/blob/91ce971828228ac4ee6f1383217e55ad86e8fa3b/packages/unplugin/src/core/transform.ts#L2766-L2787): postflight identity와 containment를 검증하고 cleanup도 canonical path로 합니다.

### 필요한 정석 수정

공용 helper는 다음 순서를 가져야 합니다.

1. temp parent를 먼저 realpathSync.native로 고정
2. 고정된 물리 parent 아래에서 mkdtempSync
3. 생성된 자식이 물리 parent의 직접 자식인지 검증
4. 이후 모든 쓰기와 cleanup은 그 물리 spelling만 사용

로컬 미커밋 신규 파일 packages/ttsc/src/internal/createCanonicalTempDirectory.ts도 현재 동일한 “생성 후 realpath” 순서를 공용화하고 있으며, compiler output, source-plugin scratch, runtime orphan emit, LSP manifest, graph capture 등으로 사용처를 넓히고 있습니다. 이 상태로 커밋하면 PR HEAD의 결함이 더 넓게 확산됩니다. 해당 helper부터 부모 선-canonicalization 방식으로 고친 뒤 호출부를 전환해야 합니다.

테스트는 “alias를 생성 후 retarget해도 cleanup이 안전하다”만 확인해서는 부족합니다. helper가 애초에 alias spelling 아래가 아니라 사전에 고정한 physical parent 아래에서 생성한다는 invariant를 직접 검증해야 합니다.

---

## 3. [Medium] 신규 테스트/성능 코드가 공유 node:fs 객체를 직접 덮어써 저장소의 몽키패칭 금지 계약을 위반합니다

저장소의 개발 계약은 [.agents/skills/development/SKILL.md L22](https://github.com/samchon/ttsc/blob/91ce971828228ac4ee6f1383217e55ad86e8fa3b/.agents/skills/development/SKILL.md#L22)에서 “No monkey-patching or hardcoding”을 명시합니다. 그런데 이 PR은 imported fs singleton의 함수를 직접 교체합니다.

- [transform-project-cache.ts L541-L571](https://github.com/samchon/ttsc/blob/91ce971828228ac4ee6f1383217e55ad86e8fa3b/tests/test-unplugin/src/internal/transform-project-cache.ts#L541-L571): readFileSync, lstatSync, statSync
- [같은 파일 L589-L606](https://github.com/samchon/ttsc/blob/91ce971828228ac4ee6f1383217e55ad86e8fa3b/tests/test-unplugin/src/internal/transform-project-cache.ts#L589-L606): readdirSync. 같은 패턴이 여러 race fixture에 반복됩니다.
- [test_computecachekey_memoizes... L44-L145](https://github.com/samchon/ttsc/blob/91ce971828228ac4ee6f1383217e55ad86e8fa3b/tests/test-ttsc/src/native-plugins/source-plugin/test_computecachekey_memoizes_unchanged_goroot_content_reads.ts#L44-L145): readFileSync
- [experimental/unplugin-perf index.ts L447-L476](https://github.com/samchon/ttsc/blob/91ce971828228ac4ee6f1383217e55ad86e8fa3b/experimental/unplugin-perf/src/index.ts#L447-L476): readFileSync, statSync

모두 finally 또는 restore로 되돌리고 현재 e2e executor가 직렬 실행하므로 지금 당장 제품 런타임이 오염되는 것은 아닙니다. 그러나 이 방식은 다음 구조적 문제를 남깁니다.

- 같은 프로세스에서 테스트 병렬화가 도입되면 서로의 관측/오류 주입이 섞임
- 중첩 계측은 먼저 설치한 wrapper가 아닌 잘못된 원본을 복원할 수 있음
- 실패/강제 종료 경로에서는 프로세스 전체 singleton 상태가 fixture에 종속
- production seam이 무엇을 관측해야 하는지 테스트가 드러내지 못하고 Node 모듈 캐시의 공유 객체라는 우연에 의존

제품 코드에서는 Module._load, prototype, require.extensions 같은 내부 API 패칭을 찾지 못했습니다. descriptor loader는 공식 hook 및 child isolation을 사용하므로 이 finding은 제품 구현이 아니라 신규 테스트/실험 구현에 한정됩니다. 그래도 저장소의 명시적 계약 위반이므로 병합 전에 제거해야 합니다.

### 필요한 정석 수정

- 파일 읽기/identity probe/디렉터리 열거 경계를 작은 operations interface로 주입하고, 테스트는 그 interface의 counting/fault-injection 구현을 사용
- 성능 측정은 같은 explicit seam을 통한 계수 또는 별도 child process에서의 계측으로 격리
- race fixture는 공유 fs 교체가 아니라 동기화 가능한 child process/worker, 혹은 일반적인 transaction hook seam을 사용

특정 fixture 이름이나 예상 호출값으로 production을 분기하는 test-only branch는 대안이 아닙니다. 공통 filesystem operation 경계를 드러내는 방식이어야 합니다.

---

## 별도 병합 게이트: 필수 format/typecheck/ci가 실패합니다

현재 GitHub의 typecheck job과 aggregate ci가 실패 중입니다. 실패 원인은 타입 오류가 아니라 새 파일의 Prettier drift입니다.

- [test_plugin_corpus_prepare_loads_typescript_lint_config_without_path_ttsx.ts](https://github.com/samchon/ttsc/blob/91ce971828228ac4ee6f1383217e55ad86e8fa3b/tests/test-ttsc/src/native-plugins/corpus-misc/test_plugin_corpus_prepare_loads_typescript_lint_config_without_path_ttsx.ts)

로컬에서도 다음 명령이 exit 1로 동일하게 재현됩니다.

    pnpm exec prettier --check tests/test-ttsc/src/native-plugins/corpus-misc/test_plugin_corpus_prepare_loads_typescript_lint_config_without_path_ttsx.ts

    [warn] Code style issues found in the above file.

로컬 미커밋 변경에도 이 파일 수정은 없으므로 현재 워크트리 그대로 push해도 해결되지 않습니다.

---

## 검증 결과

통과:

- pnpm build:current
- pnpm test:go — 전체 Go runner 통과
- pnpm test:typecheck
- go test ./driver ./test/driver ./test/utility
- test_transformttsc_compile_snapshot_aba_race_cannot_authorize_stale_output
- test_transformttsc_external_compile_snapshot_aba_race_cannot_authorize_stale_output
- test_transformttsc_persistent_validation_uses_per_file_inputs
- test_ttsccompiler_transform_drops_malformed_graph_members_and_volatile
- test_computecachekey_memoizes_unchanged_goroot_content_reads
- test_buildsourceplugin_sets_ttsc_owned_go_build_cache
- config-loader copy invariant 12/12
- dependency-audit 및 관련 audit/owner/flag tests
- base→HEAD 및 HEAD→worktree git diff --check

실패:

- 위 신규 corpus 파일의 targeted Prettier check
- GitHub typecheck 및 aggregate ci — 동일 원인

보안 감사의 유일한 high waiver인 GHSA-jmr9-qjv8-65gv도 별도로 확인했습니다. [공식 advisory](https://github.com/advisories/GHSA-jmr9-qjv8-65gv)는 현재 extract-zip <=2.0.1에 패치 버전이 없다고 명시합니다. 실제 경로는 website-only Marp → Puppeteer browser downloader의 개발 의존성이며, waiver는 patched_versions가 "<0.0.0"인 동안에만 적용되고 패치가 발표되는 즉시 CI가 실패하도록 테스트되어 있습니다. 이것은 숨은 영구 면제가 아니므로 finding으로 올리지 않습니다.

### 최종 판정

**수정 후 재검토가 필요합니다.** 우선순위는 다음과 같습니다.

1. graph 전용 decoder로 빈 leaf edge를 보존하고 독립 root ABA 회귀 추가
2. temp parent를 생성 전에 canonicalize하도록 PR HEAD와 로컬 공용 helper 수정
3. 신규 fs 몽키패칭을 explicit operations seam/격리 프로세스로 교체
4. Prettier drift를 고쳐 필수 CI 복구

이 네 항목이 닫히면, 나머지 캐시 소유권·descriptor isolation·generic host-input proof·Go cache boundedness 방향은 ttsc의 범용 제품 계약에 부합한다고 판단합니다.

samchon (commented):
## Individual Self-Review — `bf3eb4de8`

결론: **조치 필요(P2)** 1건을 확인했습니다.

### 발견 사항

`packages/ttsc/src/compiler/internal/transformProjectInMemory.ts`의 legacy graph decoder가 배열 값의 source key 자체는 검증하지 않아 `{ graph: { edges: { "": [] } } }`를 유효한 graph로 승격할 수 있었습니다. protocol에서 edge key는 project-relative 또는 absolute file path이고 proof parser는 빈 key를 이미 거부하므로, 이 상태는 내부 계약과 불일치합니다.

이 차이는 단순 malformed-input 관용성 문제가 아닙니다. 이전 decoder는 해당 입력을 graph-less로 남겨 whole-project fallback validation을 사용하지만, 새 decoder는 `graph !== undefined`가 되어 unplugin의 좁은 persistent validation 경로에 진입합니다. proof가 없는 legacy/외부 sidecar에서 sibling 변경을 놓치고 stale generation을 재사용할 수 있으므로 source key가 빈 문자열이면 graph member를 폐기해야 합니다.

### 권고 및 후속 확인

특정 fixture나 파일명 분기를 추가하지 말고, protocol의 보편 불변식인 `key.length > 0`만 decoder 경계에서 적용하며 malformed fixture에 `"": []` 음성 사례를 추가하는 것이 정석입니다. 이 finding은 후속 커밋 `e28601183`에서 그렇게 수정되었습니다.

그 밖의 empty leaf 보존, connected graph, ABA 방어, fixture 설계에서는 하드코딩·몽키패칭·과최적화 또는 추가 actionable finding을 찾지 못했습니다.

samchon (commented):
## Individual Self-Review — `e28601183`

결론: 기능 수정은 정확하며 **Low(P3) 문서 정합성** 1건만 확인했습니다.

빈 source key를 protocol 경계에서 일반적으로 거부하고 dependency array의 의미는 그대로 유지했으며, malformed native envelope 회귀 사례도 정확한 실패 클래스를 고정합니다. fixture명·경로·개수 특례나 production 하드코딩은 없습니다.

다만 인접 주석이 여전히 array-valued entry를 모두 보존한다고 읽혀, 새 empty-key rejection과 문언이 어긋났습니다. 동작 결함은 아니지만 향후 decoder를 다시 느슨하게 만들 수 있는 유지보수 위험이므로 주석에 non-empty source-key 조건을 명시해야 합니다. 후속 커밋 `11593775e`에서 동작 변경 없이 정정되었습니다.

samchon (commented):
## Individual Self-Review — `11593775e`

결론: **clean**입니다.

주석은 array-valued edge 보존과 non-empty source-key 불변식을 동시에 정확히 기술하며, `e28601183`의 실제 decoder 동작과 일치합니다. 코드·fixture·공개 계약에는 추가 변경이 없고, 하드코딩·몽키패칭·과최적화·두더지잡기식 특례도 없습니다. 추가 조치는 필요하지 않습니다.

samchon (commented):
## Individual Self-Review — `be953c999`

결론: 기능·안전성 차단 finding은 없고 **Low(P3) dead code** 1건만 확인했습니다.

### 검증된 불변식

- physical parent realpath 확정이 child 생성보다 선행합니다.
- child는 directory인지와 canonical parent의 direct child인지 postflight로 검증합니다.
- 검증되지 않은 경로에는 파괴적 cleanup을 수행하지 않습니다.
- Windows junction/POSIX symlink retarget 회귀 테스트가 original physical child만 제거되고 공격자가 만든 동일 basename victim은 보존됨을 확인합니다.
- 연산 seam은 `lstat`/`mkdtemp`/`realpath` 세 종류뿐이며 shared `node:fs`를 변경하지 않습니다.
- package-local helper 복제는 런타임 및 build-before-ttsc 경계 때문에 정당하고, fixture/count/path 특례나 범용 DI 프레임워크는 없습니다.

### Low finding

`packages/lint/src/index.ts`의 비공개 `realpathIfPossible`은 마지막 호출이 새 canonical-parent helper로 교체되어 정의만 남았습니다. 동작 회귀는 아니지만 이번 이슈가 폐기한 child-first best-effort 패턴의 잔재이므로 제거가 적절합니다. 후속 커밋 `b256594ef`에서 참조가 없음을 재확인한 뒤 삭제했습니다.

### 정확한 한계

Node의 portable path API에는 directory-handle 기반 `mkdirat`/상대 삭제가 없어 동일 권한 공격자가 canonical parent나 child 자체를 syscall 사이에 교체하는 적대적 경쟁까지 영구 고정할 수는 없습니다. 이 커밋은 재현된 TEMP/TMPDIR symlink/junction alias-retarget 결함 클래스를 정확히 닫으며, 그 이상의 보안을 주장해서는 안 됩니다.

samchon (commented):
## Individual Self-Review — `b256594ef`

결론: **clean**입니다.

`packages/lint/src/index.ts`의 미사용 비공개 `realpathIfPossible` 함수 8줄만 제거했습니다. 해당 TypeScript 함수의 남은 참조가 없고, 동명의 Go 구현 및 복제 일치 CI는 별개입니다. 새 canonical temp helper와 #1197의 동작 의미, package export 및 declaration API는 변하지 않았습니다. `git diff --check be953c999..b256594ef`도 통과했으며 추가 조치는 없습니다.

samchon (commented):
## Individual Self-Review — `b9cb30b4c`

### 판정

수정이 필요한 항목 3건(중간 2, 낮음 1)을 확인했습니다. 전역 `node:fs` 대입 제거, WeakMap 기반 캐시/결과 세대 소유권, GOROOT 관측 주입, 실제 코어를 구동하는 성능 harness 자체에서는 추가 결함을 찾지 못했습니다.

### 조치 필요

1. **중간 — path identity의 파일시스템 소유 경계가 불완전함** (`packages/unplugin/src/core/transform.ts`)
   - 캐시 operation 중 `realpath`만 identity resolver로 전달되고, 기본 case-sensitivity 판정은 ambient `node:fs`의 `readdirSync`/`lstatSync`를 직접 사용했습니다.
   - 따라서 캐시별 계측·실패·비-host filesystem 의미가 identity 측정에서는 다시 프로세스 host와 섞일 수 있습니다.
   - 정석 수정은 cache-owned `caseSensitive`/`platform` 정책과 필요한 directory/metadata operation을 동일 identity context로 전달하는 것입니다.

2. **중간 — 격리 회귀가 순차 A→B→A라 실제 전역 current-operation 결함을 배제하지 못함** (`tests/test-unplugin/src/internal/transform-project-cache.ts`)
   - 호출마다 전역 operation을 갈아끼우는 잘못된 구현도 순차 counter 테스트는 통과할 수 있습니다.
   - 서로 다른 두 프로젝트와 캐시를 겹쳐 실행하고, 한쪽 fault가 다른 쪽 결과·관측 경로로 전파되지 않음을 고정해야 합니다.

3. **낮음 — 새 feature test의 3단 계약 주석 누락** (`tests/test-unplugin/src/features/transform/test_transformttsc_filesystem_operations_are_cache_local.ts`)
   - development 계약의 headline, regression cause, numbered scenario를 추가해야 합니다.

### 금지사항·범위 검토

- 변경 diff에는 imported `node:fs` method 대입이나 shared singleton 교체가 없습니다.
- fixture명·예상 count에 따른 production 분기나 특정 경로 하드코딩이 없습니다.
- operation seam은 실제 cache validation과 source-build 관측에 필요한 읽기 경계로 한정되어 범용 DI 프레임워크로 팽창하지 않았습니다.
- 위 3건은 새 주제 발굴이 아니라 #1198의 ‘전역 관측 상태 제거 및 캐시별 소유권’ 불변식을 끝까지 전파하는 동일 원인 후속입니다.

세 항목은 후속 커밋 `70da3d0c0`에서 해결됐고 별도 Individual 재검토가 clean으로 종료됐습니다.

samchon (commented):
## Individual Self-Review — `70da3d0c0`

### 최종 판정

**Clean — actionable finding 0건입니다.**

### 최초 3건의 해결 확인

1. **Path identity 소유권**
   - 공용 path-identity resolver가 선택적 `readdir`·`lstat` operation을 정식으로 받으며, 기본 case-sensitivity 판정과 macOS alternate-realpath probe도 전달된 operation을 사용합니다.
   - unplugin은 캐시가 소유한 `caseSensitive`, `platform`, `readdir`, `lstat`, `realpath`를 하나의 identity context로 묶습니다.
   - 기본 호출은 기존 host filesystem 의미를 유지하고, 비-host filesystem은 명시적 case/platform 정책을 제공할 수 있으므로 ambient identity probe 혼합이 닫혔습니다.

2. **실제 동시성·fault 격리**
   - 서로 다른 프로젝트와 캐시를 `Promise.all`로 겹쳐 실행합니다. 첫 transform이 내부 await에서 양보한 뒤 두 번째 transform이 시작되므로, process-global `currentOperations` 구현이면 첫 캐시 fault가 실행되지 않거나 첫 operation에 두 번째 프로젝트 경로가 기록되어 실패합니다.
   - 첫 캐시에만 post-compile `EACCES`를 주입하면서 두 결과의 성공, 양쪽 counter 실행, 첫 operation이 두 번째 프로젝트를 관찰하지 않음을 함께 검증합니다.
   - 기존 WeakMap cache/result ownership과 generation 경계도 유지됩니다.

3. **테스트 계약 문서**
   - feature test에 `Verifies` headline, 회귀 원인, 3단계 시나리오가 모두 추가됐습니다.

### 정석성 검토

- 새 `node:fs` method 대입이나 간접 몽키패칭이 없습니다.
- production에 fixture명·예상값·특정 프로젝트 경로 분기가 없습니다.
- 공용 identity resolver의 선택적 operation은 동일 원인의 전체 호출자를 수용하는 일반 해법이며 하드코딩·두더지잡기·과최적화가 아닙니다.
- 기본 filesystem 동작과 기존 호출자의 타입 호환성이 유지됩니다.
- exact parent-to-commit diff는 4개 파일이며 `git diff --check`가 clean입니다.

참고로 이 커밋의 실제 parent는 format-only `68dbca22e`입니다. `b9cb30b4c..70da3d0c0` 누적 범위의 추가 2줄은 #1198 의미와 무관한 formatting입니다.

samchon (commented):
## 최종 Overall Review

### 판정

**병합 가능 — actionable finding 0건입니다.**

검토 기준은 `origin/master`의 `50934aef753459e456743f5f99af3736ce0bd27b`부터 최종 HEAD `70da3d0c0d175c665763cf411b757f9140171462`까지입니다. 131개 파일, 약 15,357줄 추가/1,122줄 삭제의 전체 변경면을 대상으로 기존 Overall finding의 수정 결과, 그 수정에서 파생되는 소비자·실패·플랫폼·동시성 경로, 각 구현 커밋의 Individual 기록, 최종 diff와 CI를 다시 대조했습니다. 미해결 review thread는 없고, 최신 완전 검토 라운드에서도 새 actionable finding이 나오지 않았습니다.

### 최초 Overall finding 4건의 종결

1. **빈 graph leaf 손실과 stale persistent reuse**
   - decoder는 이제 `edges[file] = []`를 유효한 독립 graph node로 보존합니다. 연결된 edge만 우연히 살아남는 방식이 아니라 source-key/array 계약을 직접 파싱합니다.
   - 실제 TypeScript-Go transform이 독립 leaf를 내보내는 경로와, 그 leaf가 A→B→A로 돌아온 뒤 이전 generation을 재사용해서는 안 되는 persistent cache 경로를 모두 고정했습니다.
   - 후속 Individual에서 발견한 `"": []`도 protocol 경계에서 일반적으로 거부합니다. 빈 key는 파일 경로가 아니며, 이를 graph로 승격해 좁은 검증 경로에 진입시키지 않습니다. 특정 fixture명·파일명·개수에 대한 예외는 없습니다.

2. **TEMP/TMPDIR alias-retarget cleanup의 TOCTOU/data-loss 위험**
   - scratch helper는 child를 먼저 만든 뒤 best-effort realpath하는 패턴을 폐기했습니다. 물리 parent를 먼저 canonicalize하고 directory임을 확인한 뒤 그 parent 아래에서 `mkdtemp`를 수행하고, 반환 child의 종류·realpath·직접 자식 관계를 postflight로 확인합니다.
   - postflight에 실패한 alias-spelled/unverified child에는 `rm`이나 `rmdir`을 호출하지 않습니다. 즉 소유권을 증명하지 못한 경로를 파괴적으로 정리하지 않습니다.
   - Windows junction과 POSIX symlink retarget 회귀는 원래 physical child만 제거되고 공격자가 다른 project 아래에 만든 동일 basename sentinel은 남는 것을 검증합니다.
   - ttsc의 descriptor/API emit/single-file/LSP/runtime/source-plugin/capture 호출자와 graph/lint/generator/check-flags의 독립 실행 경계까지 같은 parent-first 불변식으로 전파됐습니다. 마지막 child-first helper 잔재도 제거했습니다.

3. **테스트·성능 계측의 `node:fs` 몽키패칭**
   - PR 추가 diff에는 imported `node:fs` method 대입, `Module._load`, `require.extensions`, prototype 또는 shared singleton 교체가 없습니다.
   - transform의 동기 filesystem read는 cache-keyed WeakMap에 소유되고, compiler result도 같은 operation context를 이어받습니다. counter와 fault는 해당 cache/generation 밖으로 새지 않습니다.
   - 후속 검토에서 `realpath`만 전달되고 case-sensitivity 내부의 `readdir`/`lstat`가 ambient host를 보는 틈을 찾았습니다. 최종 구현은 `caseSensitive`, `platform`, `readdir`, `lstat`, `realpath`를 하나의 cache-owned identity context로 묶습니다. 기본 host 의미는 그대로이고 비-host filesystem은 고수준 정책을 명시할 수 있습니다.
   - 격리 테스트는 순차 A→B→A counter가 아닙니다. 서로 다른 두 project/cache를 `Promise.all`로 실제 겹쳐 실행하고 첫 cache에만 post-compile `EACCES`를 주입합니다. 두 결과가 모두 성공하고, 양쪽 counter가 각각 작동하며, 첫 cache operation이 두 번째 project 경로를 한 번도 관찰하지 않는 것을 확인합니다. 전역 `currentOperations` 방식이면 fault count 또는 cross-project 관측 assertion에서 실패합니다.

4. **format drift와 유지보수 잔재**
   - corpus scenario의 Prettier drift를 별도 format-only 커밋으로 정리했습니다.
   - canonical temp 전환 뒤 호출자가 사라진 비공개 `realpathIfPossible`도 제거했습니다. 기능과 공개 API에는 영향이 없습니다.

### 전체 consequence surface 검토

- **Persistent cache 정확성:** graph member의 compiler-time content hash와 physical realpath proof, lexical alias 후보, missing/directory state, host inputs, graph completeness와 generation membership을 함께 비교합니다. 관측된 in-project/external A→B→A와 selected symlink/junction retarget은 stale generation을 승인하지 않습니다. proof가 없는 이전 sidecar에는 보수적 fallback이 유지됩니다.
- **Graph 의미:** type-only edge, global/config node, independent leaf, resolution candidate와 lexical winner가 각자의 역할을 유지합니다. proof map key를 임의로 graph membership으로 승격하지 않으며 malformed member는 안전하게 탈락합니다.
- **Descriptor/config 격리:** Node/Bun/TSX/CJS/ESM 및 banner/strip/lint utility config가 실제 선택된 파일, manifest, 선행 resolution candidate, lazy dependency와 runtime boundary를 기록합니다. application-owned module cache를 재귀 purge하거나 singleton을 갈라놓지 않습니다.
- **Go/source-build storage:** 기본 GOCACHE와 plugin object cache는 소유 root 안에서 bounded pruning되고 active lease, 진행 중 generation, 방금 publish한 binary를 보호합니다. 실패·orphan·clock rollback·future timestamp·coordination record 교체 경로도 별도로 검증됐습니다. 외부 Go SDK의 사용자 권한은 ttsc bundled toolchain 정책과 분리됩니다.
- **Failure와 cleanup:** build 실패, Worker/heartbeat 시작 실패, descriptor 예외, overlay disposal, interrupted maintenance, scratch postflight rejection이 소유하지 않은 파일 삭제나 영구 poisoned state로 이어지지 않습니다.
- **플랫폼:** Windows junction/8.3/case policy와 POSIX symlink/case-sensitive semantics를 분리해 검증했습니다. path 비교는 문자열 prefix가 아니라 filesystem identity와 direct-child 관계를 사용합니다.
- **성능:** 최적화는 재현된 whole-envelope 재해시·반복 GOROOT content read·무제한 cache growth에 한정됩니다. 실코어를 구동하는 frozen harness에서 N=10~100의 cache reuse, generation memo, graph probe와 read/stat bound를 확인했습니다. fixture 전용 production fast path나 예상 count 특례는 없습니다.

### 금지사항 감사

- **두더지잡기 없음:** 표면 증상마다 예외를 추가하지 않고 graph parser 계약, canonical-parent ownership, generation proof, cache-owned operation context라는 원인 단위 불변식으로 소비자 전체를 닫았습니다.
- **하드코딩 없음:** production diff에서 fixture명, 테스트 경로, 특정 소비자명, 기대 개수, 개발자 로컬 경로에 따른 분기를 찾지 못했습니다. ignore directory와 cache bound처럼 제품 정책인 값은 공개된 소유·성능 계약에 대응하며 테스트 통과용 특례가 아닙니다.
- **몽키패칭 없음:** 새 shared module mutation은 0건입니다. 필요한 관측 seam은 명시적 operation과 cache/result ownership으로 제한됩니다.
- **과최적화 없음:** 범용 DI framework나 speculative cache 계층을 만들지 않았습니다. operation 표면은 실제 validation/identity/source-build read에 필요한 동기 연산이고, 최적화는 benchmark와 회귀가 증명하는 비용에만 적용됩니다.

### 검증 결과

로컬에서 다음을 통과했습니다.

- `ttsc`, `@ttsc/unplugin`, graph, lint build와 관련 typecheck
- 전체 `transformttsc` feature suite 70여 시나리오
- graph leaf/malformed envelope/real TypeScript-Go transform/A→B→A 회귀
- canonical scratch junction·symlink retarget과 descriptor cleanup 회귀
- concurrent cache-local operation, persistent per-file validation, watch bound, filesystem path-identity 회귀
- GOROOT unchanged-content memo 및 source-build cache ownership 회귀
- 실제 core-backed experimental cache performance workload
- target Prettier, base→HEAD `git diff --check`

정확한 최종 HEAD에서 GitHub Actions의 build, typecheck, Go/graph/lint/package/ttsc/bundler/platform defense, benchmark, website와 보안 검사가 모두 성공했습니다.

### 해결 가능성 및 영구 한계

이 캠페인은 원천적으로 해결 불가능한 문제를 무한 반복한 것이 아닙니다. 재현된 결함은 모두 명확한 소유·증명 계약으로 닫혔고, published issue #1196·#1197·#1198의 구현과 Individual 재검토가 끝났으며, 마지막 완전 검토가 비었습니다. 이것이 이번 캠페인의 종료 조건입니다.

다만 아래 세 경계는 TypeScript 코드로 제거할 수 없는 외부 계약이며, 이를 새 버그처럼 반복 추적해서는 안 됩니다.

1. Node의 portable API에는 directory handle 상대 `mkdirat`/삭제가 없습니다. 동일 권한의 적대자가 canonical parent 또는 child 자체를 syscall 사이에 계속 rename/replace하는 공격을 사용자 공간 경로 문자열만으로 transactional하게 봉쇄할 수는 없습니다. 이 PR은 재현된 TEMP/TMPDIR alias-retarget class를 닫으며 그 이상의 OS 보안을 주장하지 않습니다.
2. 외부/legacy plugin sidecar가 `dependenciesComplete`, graph edge, volatile non-file input 또는 proof를 거짓으로 보고하면 host가 존재하지 않는 semantic dependency를 일반적으로 추론할 수 없습니다. proof 누락·malformed 구조에는 보수적 fallback을 적용하지만, 유효한 형식으로 거짓말하는 producer의 의미까지 복원하는 것은 protocol producer 책임입니다.
3. 같은 계정의 적대자가 모든 파일을 임의 시점에 계속 바꾸는 filesystem을 일반 빌드가 완전한 transaction으로 만들 수는 없습니다. 이 PR은 compiler-time proof와 pre/post snapshot이 실제로 관측한 ABA·retarget을 거부하지만 OS snapshot 기능을 발명하지 않습니다.

이 경계를 명시한 상태에서 현재 PR에 남은 조치 항목은 없습니다. PR을 ready로 전환해도 되며, merge는 별도 권한과 절차의 대상입니다.

PR #1189, opened 2026-08-12
fix(ttsc): carry forwarded tsgo flags in the environment and write build information

### LINKED ISSUE #1188: `tsBuildInfoFile` is never written, and the documented flag forwarding fails
Hi samchon, one more thing I hit while moving to TypeScript 7.

`ttsc` never writes the `tsBuildInfoFile`, so there is no incremental build state at all. The obvious workaround — passing `--incremental` on the command line — fails, and so does every other compiler flag the README says is forwarded to tsgo. The two together mean there is currently no way to get an incremental build out of `ttsc`.

Neither of these involves a plugin; the reproduction below is a single plain TypeScript file with no `typia` or `@nestia` anywhere.

## Setup

```
bi/
├── tsconfig.json
└── src/index.ts
```

`src/index.ts`:

```ts
export const hello = (name: string): string => `hi ${name}`;
```

`tsconfig.json`:

```json
{
  "compilerOptions": {
    "module": "commonjs",
    "target": "ES2021",
    "rootDir": "./src",
    "outDir": "./dist",
    "incremental": true,
    "tsBuildInfoFile": "./.cache/app.tsbuildinfo"
  },
  "include": ["./src"]
}
```

## 1. `tsBuildInfoFile` is never written

```bash
$ rm -rf dist .cache && npx ttsc --emit -p tsconfig.json
$ echo $?
0
$ ls dist/index.js          # emitted
$ ls .cache/app.tsbuildinfo # ENOENT
```

The same project, same `tsconfig.json`, compiled with the tsgo binary directly:

```bash
$ rm -rf dist .cache && npx tsc -p tsconfig.json     # @typescript/native
$ echo $?
0
$ ls dist/index.js           # emitted
$ ls .cache/app.tsbuildinfo  # written
```

**Expected:** `ttsc` honours `incremental` / `tsBuildInfoFile` the same way tsgo does.
**Actual:** the build succeeds and emits, but the build info file is silently never produced.

Since tsgo writes it when invoked directly with the identical config, the state is being dropped somewhere in `ttsc`'s own invocation rather than by the compiler.

## 2. Forwarded compiler flags fail

The README states:

> Any other flag is forwarded to tsgo as-is, so tsgo compiler options such as `--strict` or `--target` work directly (e.g. `ttsc --strict file.ts`).

In practice every one of them fails:

```bash
$ npx ttsc --emit --strict -p tsconfig.json
flag provided but not defined: -tsgo-args
Usage of build:
  -cwd string
        override the working directory
  -emit
        force emitted .js files
  ...
$ echo $?
2
```

| command | exit | `tsBuildInfoFile` written |
|---|---|---|
| `ttsc --emit` | 0 | no |
| `ttsc --emit --incremental` | 2 | **yes** |
| `ttsc --emit --strict` | 2 | **yes** |
| `ttsc --emit --declaration` | 2 | **yes** |
| `ttsc --emit --target ES2022` | 2 | no |
| `ttsc --emit --listFilesOnly` | 2 | no |

All of them print the same `flag provided but not defined: -tsgo-args` usage block. Flags that belong to `ttsc`'s own schema (`--emit`, `--noEmit`, `-p`, `--outDir`) work fine, so the failure looks specific to the path that hands `-tsgo-args` to the Go build driver — the driver's own usage output does not list a `-tsgo-args` flag.

Worth noting: `--incremental`, `--strict` and `--declaration` **do** write the build info file before the process exits 2. So the build itself ran to completion and something after it failed, which also shows the incremental write path works and is simply not being taken by default.

I also saw the same `-tsgo-args` error from single-file mode (`ttsc --noEmit path/to/file.ts`), if that is the same code path.

## Environment

| | |
|---|---|
| `ttsc` | 0.26.2 |
| `typescript` (native) | 7.0.2 |
| Node.js | 24.18.0 |
| OS / arch | macOS, darwin-arm64 |

## Impact

Our CI caches the compiled test output and decides whether it can be reused by checking that the previous build's artifacts are present, `tsBuildInfoFile` among them. Because that file is never produced, the check always fails and every run recompiles the whole project from scratch — about 3.5 minutes per run that the cache was meant to avoid. With `--incremental` returning exit 2 there is no way to opt into it either.

---

_This issue was drafted with AI assistance._

samchon:
Resolves #1188.

## What the report got right, and what it got wrong

Both symptoms are real and both are ttsc's fault. The reproduction in the issue is not.

A plain project — no plugin, the issue's exact `tsconfig.json`, `ttsc@0.26.2` and `typescript@7.0.2` — behaves correctly: `ttsc --emit -p tsconfig.json` exits 0 **and writes `.cache/app.tsbuildinfo`**, and `--strict` / `--incremental` exit 0 too. Without a plugin the launcher never spawns a Go sidecar; it runs the `typescript` package's own `tsc`, which handles forwarding and build info itself. `--tsgo-args` does not exist on that lane, and the `Usage of build:` block in the report is a Go `flag.FlagSet` dump the plain lane cannot produce.

The failing lane is the native plugin lane. `node_modules/typia/native/cmd/ttsc-typia/build.go` declares exactly the flags in the reporter's usage dump, with exactly those descriptions — that binary produced the output. The reporter's project carries typia; the "no plugin anywhere" framing sent the investigation at the one lane that works.

## Defect 1 — ttsc put a flag on a command line it does not own

`createNativeTsgoArgs` appended `--tsgo-args=<JSON>` to every native sidecar invocation. #113 added that flag to a plugin protocol third-party hosts had already frozen, and Go's `flag.FlagSet` treats an undeclared flag as fatal under `flag.ContinueOnError`. So every pre-#113 sidecar exits 2 before its build starts.

This repository already diagnosed the class once: `ad3443a` reverted #113's bare `--singleThreaded` / `--checkers` forwarding for the same reason and re-introduced them behind `capabilities.threadingArgs`. The comment recording that decision sits eighty lines above the sibling payload that still travelled the unsafe channel.

Measured on a host with the frozen flag set (all previously exit 2):

| command | before | after |
|---|---|---|
| `ttsc --emit --strict` | `flag provided but not defined: -tsgo-args` | 0 |
| `ttsc --emit --declaration` | same | 0 |
| `ttsc --emit --target ES2022` | same | 0 |
| `ttsc --strict` | same | 0 |
| `ttsx --strict` | same | 0 |
| `ttsc <file.ts>` — **no user flag at all** | same | 0 |

That last row is the widest instance and the issue mentions it only as an aside. Single-file mode compiles into a private temp directory, so `isolatedTsgoOutputArgs` always contributes `--outFile null --declarationDir null --tsBuildInfoFile null --outDir <tmp>`. ttsc's own containment flags were crashing the sidecar.

**Fix.** The payload moves to `TTSC_TSGO_ARGS`. An unknown environment variable is inert to every host; an unknown flag is fatal to all of them, and this is already how ttsc carries `TTSC_LINKED_PLUGINS_JSON` and `TTSC_PLUGIN_CONFIG_DIR` past the same hazard. `driver.LoadProgram` reads it whenever the caller supplied no explicit argv, so third-party hosts pick the options up **without any change on their side** — `buildSourcePlugin.ts` compiles every source plugin against the installed ttsc's own driver. ttsc's three hosts still accept `--tsgo-args`, so an older launcher paired with a newer host keeps working.

A capability gate (the `ad3443a` shape) was considered and rejected: it stops the crash but silently drops `--strict` on the transform lane, where no separate tsgo pass exists to enforce it. Silently changing type-checking semantics is worse than the threading knob that precedent dropped, and it would not restore what the README documents.

## Defect 2 — the native-host emit lane never wrote build information

`driver.LoadProgram` builds a plain `compiler.Program` and `EmitAll` / `EmitAllRaw` call `TSProgram.Emit`. tsgo writes `.tsbuildinfo` from `performIncrementalCompilation`, a branch in `internal/execute` that a host constructing its Program in-process never enters. So `incremental`, `composite`, and `tsBuildInfoFile` parsed, resolved, and vanished.

This is not a third-party plugin's fault: a project on **ttsc's own** linked host (`@ttsc/banner`) emitted `dist/main.js` and no build info either. `driver/emit_containment.go` had already exempted `.tsbuildinfo` from the outDir guard — a guard written for a write that could not happen.

**Fix.** The driver's whole-program emits run through tsgo's own incremental program via a new `shim/compiler` re-export, including the hand-assembled `EmitWithPluginTransformers` lane typia uses.

**No previous snapshot is consumed** (`oldProgram = nil`). Every file therefore stays affected and emit output is unchanged; only the record is new. Reusing one would let ttsc skip a file whose *source* is unchanged, but a plugin's output also depends on the plugin binary, its config file, and its contributors — none of which a `.tsbuildinfo` records. Skipping on that basis would serve stale transformed output. Sound reuse needs plugin identity in the invalidation key; that is a feature, not this bug.

One honest caveat, documented in code and on the docs page: on the hand-assembled lane the JavaScript never goes through tsgo's emitter, so the record marks JS emit as still pending. The error is one-directional — a consumer can only decide to emit again, never to skip a file the plugin transformed.

## Coverage

Both defects existed because nothing pinned either behavior. `test_plugin_corpus_single_threaded_flag_does_not_break_a_native_plugin_build` pins the `ad3443a` regression with a deliberately strict host, but no test did the same for a forwarded tsgo flag, and no test asserted build-info emission on any lane.

Go, `packages/ttsc/test/driver/`:

- build info written for an incremental project — both whole-program lanes, at the exact configured path, versioned JSON, JS still emitted;
- negative twin — `tsBuildInfoFile` alone must not produce one;
- the environment channel's whole decision table — env applies, explicit argv wins, absent forwards nothing, malformed is reported.

TypeScript e2e, on the `go-driver-emit-plugin` host (strict `flag.FlagSet`, `driver.LoadProgram`, `EmitWithPluginTransformers` — typia's shape):

- a forwarded flag **reaches** the host: the project declares no `sourceMap`, so the emitted `.js.map` can only exist if `--sourceMap` arrived. A zero exit alone would also pass if ttsc had merely dropped the flag;
- single-file emit with no user flag;
- build info written on the plugin lane, and its negative twin.

## Consequence surface

| Lane | Before | After |
|---|---|---|
| plain tsgo build | worked | unchanged |
| transform sidecar build | exit 2 on any forwarded flag | delivered via env |
| check sidecar (`check`/`fix`/`format`) | exit 2 | delivered |
| resident `check --watch` | flag in argv, and in the process key | env payload, folded into the key |
| single-file `ttsc <file.ts>` | exit 2 with no user flag | works |
| `ttsx --strict` | exit 2 | works |
| `project-inputs` | never carried it | unchanged; inherited value cleared |
| ttsc-owned host emit | no build info | writes it when incremental |
| plugin hand-assembled emit | no build info | writes it, conservatively |
| non-incremental project | none | unchanged, pinned |

## Local verification

- `go test ./test/driver/...` — full driver suite, plus the three new cases;
- the four new e2e cases, and `test_plugin_corpus_ttsc_lint_applies_forwarded_strict_flag` / `..._format_honors_single_threaded_flag` to prove `@ttsc/lint` still receives forwarded flags through the new channel;
- `pnpm test:typecheck`, `node packages/ttsc/scripts/check-flags.cjs`, `pnpm format`;
- end-to-end against a real typia-shaped host and against `@ttsc/banner`: `--strict`, `ttsc <file.ts>`, and the `.tsbuildinfo` all behave as the tables above state.

Docs: new **Environment** section in the plugin protocol page listing every variable a sidecar receives and why argv is not a safe place to extend a frozen protocol; new **Build information** section in the compile page.

https://claude.ai/code/session_01Chz39rm1FC2dinzi7JBWAJ


samchon (commented):
## Self-Review — round 1 over `master...858ed535b`

Solo review of the whole base-to-head diff. Note on procedure: the campaign's per-commit Individual Self-Review normally runs as a read-only subagent; this session is configured not to spawn agents, so both commits were reviewed by the same reviewer as part of this round. The whole-diff Overall Self-Review gate is unchanged.

### Finding applied — an inherited payload could reach a sidecar that forwards nothing

Moving the forwarded tsgo argv from argv into the environment made it **inheritable**, and the first commit only guarded the lane that publishes it. `transformProjectInMemory.ts::nativePluginEnv` and `startResidentTransform.ts::residentEnv` both build their environment from `process.env` and forward no tsgo argv of their own, so a ttsc running inside a plugin sidecar — `@ttsc/lint` evaluating a config file through `ttsx` is the concrete path — would have handed those sidecars the outer run's `--strict`, and `driver.LoadProgram` would have applied it.

This is the flip side of the fix and belongs to it: argv was not inheritable, so the hazard is new. Both lanes now clear it, the variable name and the clearing step moved to `sharedHostHelpers` so one definition serves all three spawn sites, and a case pins the three outcomes (inherited cleared, caller-named kept, absent untouched). Fixed in `858ed535b`.

### Verified, no change needed

- **`applyRewrites` cannot corrupt the build info.** `EmitAll`'s WriteFile runs the rewrite pipeline on every emitted file, `.tsbuildinfo` included. `findSourceForOutput` does not match it, so the text passes through and no sentinel is inserted. The Go case asserts the artifact parses as versioned JSON, which a sentinel prefix would break.
- **The outDir guard already allowed it.** `emit_containment.go::outputEscapesOutDir` exempts `.tsbuildinfo` because its default home is next to the tsconfig. The e2e case deliberately puts it at `./.cache/` outside `outDir` to exercise exactly that exemption.
- **No mtime call on the fresh path.** `incremental`'s `GetMTime` / `SetMTime` run only under `differsOnlyInMap`, which needs a composite project *and* a previous emit signature. With `oldProgram = nil` the signature map is empty, so a VFS without `Chtimes` (`@ttsc/wasm`) is never asked for one.
- **Single-file emit stays on the plain lane.** Build information describes a whole program and tsgo's incremental program returns early on a `TargetSourceFile` request, so routing it there would compute a snapshot nothing reads. `test_ttsc_single_file_sandboxes_all_compiler_side_products` still passes — that project sets `incremental` and asserts no build info leaks.
- **`DeclarationShapeDigest` untouched.** It is a per-file analysis emit for cache invalidation, not a build; it keeps calling `TSProgram.Emit` directly.
- **`emitsBuildInfo` delegates rather than restates.** It calls `CompilerOptions.IsIncremental`, the same predicate tsgo's CLI branches on, instead of a local `Incremental || Composite` copy that would need keeping in step with upstream.
- **Resident process identity.** The payload left argv, which the resident check key is built from. It is now folded into that key explicitly, so two cycles forwarding different flags cannot share one warm Program.
- **`--tsgo-args` remains declared and parsed** by all three ttsc hosts and stays in `HostFlagAllowList`, so an older launcher paired with a newer host still works. Only the launcher stopped emitting it.
- **`project-inputs` deliberately carries nothing.** It publishes filesystem topology without loading a Program, so tsgo options are irrelevant to it; an inherited value is cleared there too.

### Local verification on this head

- `node scripts/test-go.cjs` — full Go suite, exit 0.
- `go test ./test/driver/...` including the three new cases.
- e2e: the four new plugin-corpus cases, the new sidecar-env case, `test_ttsc_single_file_sandboxes_all_compiler_side_products`, and `test_plugin_corpus_ttsc_lint_applies_forwarded_strict_flag` / `..._format_honors_single_threaded_flag` — the last two prove `@ttsc/lint` still receives forwarded flags through the new channel.
- `pnpm test:typecheck`, `node packages/ttsc/scripts/check-flags.cjs`, `pnpm --filter ttsc shim:audit` (OK — enums, reachable symbols, and producers closed), `pnpm format`.
- Manual, against a real typia-shaped host: `ttsc --emit --strict`, `ttsc <file.ts>`, `ttsx --strict` all exit 0 with the transform applied; against `@ttsc/banner` on ttsc's own linked host, `.cache/app.tsbuildinfo` is written where the tsconfig asks.

Not run locally: the full `pnpm test:features` sweep and the non-win32 platform lanes. Those are this pull request's CI gate.

samchon (commented):
## Self-Review — round 2 over `master...0c783d683`

Fresh complete round over the whole base-to-head diff, not a recheck of round 1's fix.

### Findings applied — documentation that no longer described the code

- `shim/compiler/incremental.go` opens with a package comment scoping the file to "reference-graph parts of tsgo's own incremental engine", and an import comment calling the incremental package a blank import "so the references resolve". Both predate `EmitFreshWithBuildInfo`, which is neither reference-graph nor blank-imported. A hand-maintained shim file is exactly where a stale header costs the next reader real time.
- The new **Environment** section on the protocol page ended with "a value inherited from an ancestor `ttsc` process is cleared rather than passed on". That is true of `TTSC_TSGO_ARGS` and `TTSC_PLUGIN_CONFIG_DIR` and the **opposite** of how `TTSC_TSGO_BINARY` / `TTSC_TTSX_BINARY` behave, where an inherited value deliberately wins so an embedder can pin the toolchain for a whole process tree. A plugin author following the blanket claim would have drawn the wrong conclusion about both halves.

Both corrected in `0c783d683`.

### Behavior change accepted deliberately, recorded here rather than left implicit

Routing an incremental whole-program emit through tsgo's incremental program also brings `compiler.HandleNoEmitOnError` into the path, which the plain `Program.Emit` lane never called. A project that sets `noEmitOnError` *and* has errors would now skip emit and write only the build information, where ttsc previously emitted anyway.

Every ttsc host checks `prog.Diagnostics()` and exits 2 before reaching emit (`cmd/ttsc/build.go`, `utility/host.go`, and typia's own host do this), so the case is practically unreachable. Where it is reachable, the new behavior is tsgo's own. Flagging it because it is a semantic change that no test would have caught, not because it is expected to bite.

### Observed, deliberately left out of this pull request

`TTSC_LINKED_PLUGINS_JSON` is set when linked plugins exist but is never cleared when they do not, so a nested ttsc running inside a transform sidecar can inherit the outer manifest. The Go side already covers the case that matters today — `utility/host.go::setLinkedPluginManifest` unsets the variable for a blank `--plugins-json` and restores it afterwards — which is why it has not surfaced.

It is a different payload that happens to share a shape, it predates this change, and nothing in issue #1188's cause produces it. Folding a speculative fix for it into a bug-fix pull request would widen the diff past its evidence; it is recorded in the campaign ledger and deserves its own issue.

### Round result

No correctness, coverage, packaging, platform, or compatibility finding survived verification in this round beyond the two documentation corrections above. Re-verified after them: `go build ./...` across `packages/ttsc` and `shim/compiler`, `pnpm format`, `node packages/ttsc/scripts/check-flags.cjs` (schema, generated Go allow-lists, and the docs table agree; the upstream option table agrees on all 15 declared kinds).

CI on this head is the remaining gate.

samchon (commented):
## Self-Review — round 3 over `master...f5e1e8db4`

### Finding applied — the previous round's fix was partial

Round 1 guarded the two lanes that build a sidecar environment through a named helper (`transformProjectInMemory.ts::nativePluginEnv`, `startResidentTransform.ts::residentEnv`). It did not guard the spawns that write `{ ...process.env, ...callerEnv }` inline, and there are four of them:

| site | child | reaches `driver.LoadProgram` |
|---|---|---|
| `compileProjectInMemory.ts` | `cmd/ttsc api-compile` | directly |
| `transformProjectInMemory.ts` | `cmd/ttsc api-transform` | directly |
| `transformProjectInMemory.ts`, `startResidentTransform.ts`, `runBuild.ts` | `loadProjectPlugins` | through the `ttsx` child that evaluates a TypeScript plugin descriptor, which compiles a project of its own |

Each would have handed an outer run's `TTSC_TSGO_ARGS` to a compiler that then applied it. Fixing two of six sites is the failure mode the development skill names directly — patch the case that surfaced and leave the class open — so the rule now lives in one `inheritedSidecarEnv` helper that every site calls, and a hand-written spread cannot forget it.

Fixed in `f5e1e8db4`.

### A local failure investigated and cleared, not assumed

Running `features/compiler` locally surfaced three failures, and a bug fix is exactly the wrong moment to wave those away. Checked out `master`, rebuilt, and ran the same cases: **all three fail identically there.** Both are Windows-only harness limits, not regressions:

- `test_ttsc_no_plugin_build_honors_tsconfig_no_emit` and `test_ttsc_no_plugin_emit_invokes_tsgo_once` — `createFakeNativePreview` writes a `#!/usr/bin/env node` text file named `tsc.exe`, which win32 cannot spawn (`spawnSync … UNKNOWN`). They also exercise the plugin-free lane, which this change does not touch.
- `test_ttsc_watch_reconciles_a_new_included_directory` — an idle-rebuild timing failure on win32; no `incremental`, no plugins, plugin-free lane again.
- `test_linkvirtualentry_copies_a_file_symlink_entry_when_symlink_creation_fails`, in the API suite, needs symlink privileges this machine does not grant.

Recording the check rather than the conclusion alone, so nobody has to redo it.

### Verified on this head

`features/api` (the two in-memory lanes this round touched) passes apart from the symlink case above; the four plugin-corpus cases and the sidecar-env case pass; `pnpm --filter ttsc build`, `pnpm format`.

CI on this head is the remaining gate.

PR #1206, opened 2026-08-16
docs(evidence): refine presentation narrative

samchon:
## Intent

Make the Evidence Graph and Spec Driven Development sections read as one continuous presentation narrative, from current limitations through compiler-enforced evidence and into benchmark outcomes.

## Scope

- clarify the Evidence Graph tag and review workflow
- replace redundant slides with a compact review checklist
- use idea notes as the source layer for delegated requirements
- add separate Method A and Method B architecture diagrams
- preserve the existing presentation style and concise slide copy

## Deferred

None.

## Verification

- `pnpm exec prettier --write src/slides/evidence.md` (from `website`)
- `pnpm run build:slides` (from `website`)
- `git diff --check`
- rendered and visually inspected the updated checklist and Method A/Method B diagram slides as PNGs

PR #1207, opened 2026-08-16
docs(evidence): rewrite the readme and website guides around one narrative

samchon:
Restructures `packages/evidence/README.md` so the landing surface carries the argument and the reference material stays scannable.

## What changed

**The benchmark image now has prose under it.** The numbers lived only inside the PNG, so a reader on npm, or one whose images do not load, got no evidence at all. Three paragraphs now explain why the arm without the graph cost what it cost: the review loop finds most of what it will ever find in its first rounds, counts nothing it never looked at, and settles on a ceiling that drops as the project grows. ERP is stated concretely (nine of every ten tokens on review, 5,449M total, 51.6% coverage, against 411M and $4.96), along with the 85.5% to 51.6% fall across subjects.

**Four SVG diagrams replace architecture narrative.** Documents, Backend, Frontend, and a novel graph, each with a line naming the omission its edges block. They are served from `https://ttsc.dev/evidence/*.svg` and wrapped in `<picture>` with light and dark variants, so both GitHub themes render correctly and the light variant, which paints its own background, stands in wherever the tag is stripped. Assets landed in #1206.

**The `Beyond code` section states why the graph is domain agnostic.** It reads no meaning, only obligations and citations, which is the claim that makes a novel graph something other than a curiosity in a compiler plugin README.

**A `Reviews` section admits that 100% coverage is not 100% truth.** `@evidenceReview` previously appeared only as a table row in `Rules` and a bullet under `Strict references`, with no example, while the first objection any reader has is that an agent will write untrue tags to clear errors.

**Reformatting rather than removal.** The opening error block dropped two verbatim repetitions of the same remediation paragraph and now shows one error with an elision. `Strict references` became a table. The TypeScript member classification lost its edge cases (parameter properties, overloads, the `override` modifier) to a link at https://ttsc.dev/docs/evidence/claims, keeping the rule and the two common cases. Addressing forms, symbol selectors, exclusion carriers, and the rule list all stay, because this README is read inside `node_modules` by the agents the package exists for.

Net 64 insertions, 42 deletions.

## Verification

None. `pnpm format` was skipped and CI was not waited on, both at the author's instruction. The diagram URLs resolve only once the website deploy from #1206 is live.

## Deferred

The SVG generator is not in the repository, so the diagrams cannot be regenerated from source yet. It needs a home and a decision on whether the assets stay under `website/public/evidence` or ship inside the npm tarball.

https://claude.ai/code/session_01WMyLYnKYgNCqXUkDUDexHc


samchon (commented):
Overall Self-Review record for the website-guides push (`0ce4050e4`).

**Scope reviewed**: `packages/evidence/README.md`, all six `website/src/content/docs/evidence/*` pages, `setup/evidence.mdx`, `benchmark/evidence.mdx`, `_meta.ts`.

**Verified against primary sources**:

- Rule ids and count (`evidence/graph|documented|singular|todo|review`) against `packages/evidence/src`.
- Tag set including `@evidenceExcludeReview` against `tags.mdx` and the rule JSDoc.
- `TS16411` reproduced from the FNV-1a allocator in `packages/lint/internal/rulecode`.
- Benchmark figures against `website/public/benchmark/evidence.json` (coverage 51.6–85.5% → 100%, tokens 866M→92M … 5,449M→411M, review share 90–95% → 15–41%); the stale "9% more on the largest subject" claim in `adoption.mdx` was corrected from that data.
- Backend/frontend/novel diagram edge descriptions against the shipped SVGs (hooks do not cite documents directly; screens and journeys do).
- Example configs against `benchmarks/evidence/template/evidence` on `origin/master`.

**Structure**: `wiring.mdx` dissolved into the new `examples.mdx` (backend, frontend, novels in one page); `index.mdx` reduced to a true index carrying the five-act narrative one paragraph each; every former `/docs/evidence/wiring` link retargeted (site-wide grep clean).

**Not run locally** (repository rule: CI validates): Next.js/MDX render of the new `<picture>` and `<Tabs>` usage in `examples.mdx`; `pnpm format` was run and is included in the commit.

Issue #1203, opened 2026-08-15
Expose signature abstractness (SignatureFlags) to Go source plugins


fnrhombus:
# Expose signature abstractness (`SignatureFlags`) to Go source plugins

## Problem

The checker shim aliases `Signature = internal/checker.Signature` (a full type alias) and exposes `Checker_getSignaturesOfType`, so a plugin can enumerate a type's construct signatures. `Signature.Flags()` is an exported method promoted straight through that alias, so a plugin can already call `sig.Flags()` and get a value back — but `SignatureFlags` itself is never aliased, so the plugin cannot name that return type, declare a variable of it, or reference any of its members (`SignatureFlagsAbstract`, `SignatureFlagsConstruct`, etc.). The value is reachable; nothing it can be checked against is.

This blocks reading abstractness off a signature that has no declaring symbol to fall back on. `shim/ast`'s `GetCombinedModifierFlags` + `ModifierFlagsAbstract` recover a class's `abstract` modifier when a construct signature's declaring symbol is a class declaration, but a bare abstract-constructor type literal — `type Opener = abstract new (dep: IDep) => IThing` — has no declaration to read a modifier from. The checker's own general answer is the signature-level flag, not the declaration-level modifier.

`shim/checker`'s own audit already flags the gap: `tools/shim_audit/baseline.json` carries `ESCAPE|checker|SignatureFlags`, and the ESCAPE tier is defined as "an exported type in this func's signature that isn't aliased — a plugin could obtain the value but cannot name its type" — exactly this case.

## Evidence

```go
// import "path/to/ttsc/shim/checker"

sigs := checker.Checker_getSignaturesOfType(c, t, checker.SignatureKindConstruct)
for _, sig := range sigs {
    flags := sig.Flags() // compiles — Flags() is promoted through the Signature alias
    // no way to spell checker.SignatureFlags, so no way to test flags against
    // checker.SignatureFlagsAbstract; nothing in shim/checker names either.
    _ = flags
}
```

`type Opener = abstract new (dep: IDep) => IThing` produces a construct signature whose `Declaration()` has no enclosing class, so `GetCombinedModifierFlags` has nothing to read `ModifierFlagsAbstract` from — abstractness for this shape is unreadable through the shim today.

## Consequence surface

- Go source plugins deriving metadata from an abstract-constructor type literal (no backing class declaration) cannot tell it apart from a concrete one.
[...]

PR #1208, opened 2026-08-16
feat(shim): expose SignatureFlags to Go source plugins

samchon:
## Intent

Solo issue-campaign cycle pull request. This pull request owns the complete accepted cycle, which is exactly one issue:

- References #1203 — Expose signature abstractness (`SignatureFlags`) to Go source plugins

`Signature.Flags()` is already reachable through the full `Signature` alias in `shim/checker`, but its return type `SignatureFlags` is not aliased, so a plugin can obtain the value yet cannot name its type or compare it against `SignatureFlagsAbstract`. The shim audit already records the gap as `ESCAPE|checker|SignatureFlags` ratchet backlog.

## Scope

- Alias `SignatureFlags` in the hand-maintained `shim/checker/shim.go`, beside the existing `SignatureKind` alias.
- Regenerate `enums_gen.go` with the owning command (`pnpm --filter ttsc shim:audit -fix`) so the whole member family is re-exported under the zero-tolerance enum layer.
- Remove the resolved `ESCAPE|checker|SignatureFlags` baseline entry.
- Add a runtime probe under `packages/lint/test/shim/` asserting the abstract bit across construct-signature shapes, including the shapes where declaration-based reading is impossible (nil declaration) or wrong (inherited base constructor).

## Verification

Pending. Verification is CI-owned per the repository's standing rule: the `shim-audit` gate proves audit closure, `test:go` proves the runtime probe, and the typia workflow proves consumer compatibility. Local runs are limited to the owning generator commands. Individual and Overall Self-Review results will be recorded as formal pull-request reviews.

## Deferred

Sibling `ESCAPE` baseline entries for other checker types remain grandfathered backlog; they are outside this cycle's declared single-issue scope.

### DIFF
--- a/packages/lint/test/shim/signature_flags_expose_construct_signature_abstractness_test.go
+++ b/packages/lint/test/shim/signature_flags_expose_construct_signature_abstractness_test.go
@@ -1,0 +1,105 @@
+package linthost
+
+import (
+  "path/filepath"
+  "testing"
+
+  shimchecker "github.com/microsoft/typescript-go/shim/checker"
+)
+
+// Verifies SignatureFlags exposes construct-signature abstractness through the
+// shim.
+//
+// Signature.Flags() was already reachable through the full Signature alias,
+// but SignatureFlags itself was not nameable, so the returned value could not
+// be tested against SignatureFlagsAbstract (#1203). The flag is also the only
+// correct general answer: an abstract class without a constructor produces a
+// default construct signature with no declaration at all, and an abstract
+// class inheriting a concrete base clones the base's signature, so
+// Declaration() points at a NON-abstract constructor while the checker forces
+// the Abstract bit on. Declaration-modifier reading returns nothing for the
+// former and the wrong answer for the latter.
+//
+//  1. Build a program with abstract/concrete classes and constructor-type
+//     aliases, including the no-constructor and inherited-constructor shapes.
+//  2. Obtain every construct signature only through the exported shim surface.
+//  3. Assert the Abstract bit exactly where the source says abstract, the
+//     Construct bit on every construct signature, and the nil declaration on
+//     the default-signature boundary.
+func TestSignatureFlagsExposeConstructSignatureAbstractness(t *testing.T) {
+  root := t.TempDir()
+  writeFile(t, filepath.Join(root, "tsconfig.json"), `{
+  "compilerOptions": {
+    "target": "ES2022",
+    "module": "commonjs",
+    "strict": true,
+    "rootDir": "src",
+    "outDir": "dist"
+  },
+  "files": ["src/main.ts"]
+}
+`)
+  writeFile(t, filepath.Join(root, "src", "main.ts"), `export abstract class AbstractDeclared {
+  constructor(value: string) { void value; }
+}
+export class ConcreteDeclared {
+  constructor(value: string) { void value; }
+}
+export abstract class AbstractDefault {}
+export class ConcreteBase {
+  constructor(value: number) { void value; }
+}
+export abstract class AbstractInheriting extends ConcreteBase {}
+export type AbstractOpener = abstract new (value: string) => AbstractDeclared;
+export type ConcreteOpener = new (value: string) => ConcreteDeclared;
+`)
+
+  prog, diags, err := loadProgram(root, "tsconfig.json", loadProgramOptions{
+    needsRuleChecker: true,
+  })
+  if err != nil {
+    t.Fatal(err)
+  }
+  if len(diags) != 0 {
+    t.Fatalf("unexpected diagnostics: %#v", diags)
+  }
+  defer prog.close()
+
+  cases := []struct {
+    name         string
+    typeAlias    bool
+    wantAbstract bool
+    wantNilDecl  bool
+  }{
+    {name: "AbstractDeclared", wantAbstract: true},
+    {name: "ConcreteDeclared"},
+    {name: "AbstractDefault", wantAbstract: true, wantNilDecl: true},
+    {name: "ConcreteBase"},
+    {name: "AbstractInheriting", wantAbstract: true},
+    {name: "AbstractOpener", typeAlias: true, wantAbstract: true},
+    {name: "ConcreteOpener", typeAlias: true},
+  }
+  for _, tc := range cases {
+    symbol := classSymbol(t, prog, tc.name)
+    var target *shimchecker.Type
+    if tc.typeAlias {
+      target = shimchecker.Checker_getDeclaredTypeOfSymbol(prog.checker, symbol)
+    } else {
+      target = shimchecker.Checker_getTypeOfSymbol(prog.checker, symbol)
+    }
+    signatures := shimchecker.Checker_getSignaturesOfType(prog.checker, target, shimchecker.SignatureKindConstruct)
+    if len(signatures) != 1 {
+      t.Fatalf("%s construct signatures = %d, want 1", tc.name, len(signatures))
+    }
+    flags := signatures[0].Flags()
+    if flags&shimchecker.SignatureFlagsConstruct == 0 {
+      t.Fatalf("%s flags = %d: Construct bit missing on a construct signature", tc.name, flags)
+    }
+    if got := flags&shimchecker.SignatureFlagsAbstract != 0; got != tc.wantAbstract {
+      t.Fatalf("%s abstract bit = %v, want %v (flags = %d)", tc.name, got, tc.wantAbstract, flags)
+    }
+    if got := signatures[0].Declaration() == nil; got != tc.wantNilDecl {
+      t.Fatalf("%s nil declaration = %v, want %v", tc.name, got, tc.wantNilDecl)
+    }
+  }
+}
--- a/packages/lint/test/shim/signature_flags_expose_construct_signature_abstractness_test.go
+++ b/packages/lint/test/shim/signature_flags_expose_construct_signature_abstractness_test.go
@@ -19,12 +19,12 @@ import (
 // inheriting its base's constructor clones the base signature, so
 // Declaration() points at the OTHER class's constructor while the checker
 // forces the bit to the derived class's abstractness — in both directions.
-// Declaration-modifier reading returns nothing for the former and the wrong
-// answer for the latter.
+// Declaration-modifier reading returns nothing for the former and an answer
+// that tracks the base class for the latter.
 //
 //  1. Build a program with abstract/concrete classes and constructor-type
-//     aliases, covering the declared, constructor-less, and inherited shapes
-//     in both abstract polarities.
+//     aliases, covering the declared, default-signature, and inherited
+//     shapes in both abstract polarities.
 //  2. Obtain every construct signature only through the exported shim surface,
 //     reading flags into a shimchecker.SignatureFlags-typed variable so the
 //     alias itself is pinned, not just the member consts.
@@ -114,11 +114,11 @@ export type ConcreteOpener = new (value: string) => ConcreteDeclared;
     if got := flags&shimchecker.SignatureFlagsAbstract != 0; got != tc.wantAbstract {
       t.Fatalf("%s abstract bit = %v, want %v (flags = %d)", tc.name, got, tc.wantAbstract, flags)
     }
-    if got := signatures[0].Declaration() == nil; got != tc.wantNilDecl {
+    declaration := signatures[0].Declaration()
+    if got := declaration == nil; got != tc.wantNilDecl {
       t.Fatalf("%s nil declaration = %v, want %v", tc.name, got, tc.wantNilDecl)
     }
     if tc.checkDeclaringClass {
-      declaration := signatures[0].Declaration()
       if declaration == nil {
         t.Fatalf("%s has no declaration to read a declaring class from", tc.name)
       }
--- a/packages/lint/test/shim/signature_flags_expose_construct_signature_abstractness_test.go
+++ b/packages/lint/test/shim/signature_flags_expose_construct_signature_abstractness_test.go
@@ -4,6 +4,7 @@ import (
   "path/filepath"
   "testing"
 
+  shimast "github.com/microsoft/typescript-go/shim/ast"
   shimchecker "github.com/microsoft/typescript-go/shim/checker"
 )
 
@@ -13,19 +14,23 @@ import (
 // Signature.Flags() was already reachable through the full Signature alias,
 // but SignatureFlags itself was not nameable, so the returned value could not
 // be tested against SignatureFlagsAbstract (#1203). The flag is also the only
-// correct general answer: an abstract class without a constructor produces a
-// default construct signature with no declaration at all, and an abstract
-// class inheriting a concrete base clones the base's signature, so
-// Declaration() points at a NON-abstract constructor while the checker forces
-// the Abstract bit on. Declaration-modifier reading returns nothing for the
-// former and the wrong answer for the latter.
+// correct general answer: a constructor-less class produces a default
+// construct signature with no declaration at all, and a class inheriting its
+// base's constructor clones the base signature, so Declaration() points at
+// the OTHER class's constructor while the checker forces the bit to the
+// derived class's abstractness — in both directions. Declaration-modifier
+// reading returns nothing for the former and the wrong answer for the latter.
 //
 //  1. Build a program with abstract/concrete classes and constructor-type
-//     aliases, including the no-constructor and inherited-constructor shapes.
-//  2. Obtain every construct signature only through the exported shim surface.
+//     aliases, covering the declared, constructor-less, and inherited shapes
+//     in both abstract polarities.
+//  2. Obtain every construct signature only through the exported shim surface,
+//     reading flags into a shimchecker.SignatureFlags-typed variable so the
+//     alias itself is pinned, not just the member consts.
 //  3. Assert the Abstract bit exactly where the source says abstract, the
-//     Construct bit on every construct signature, and the nil declaration on
-//     the default-signature boundary.
+//     Construct bit everywhere, the nil declaration on both default-signature
+//     boundaries, and the declaring class's modifier where a declaration
+//     exists — diverging from the flag on both inherited shapes.
 func TestSignatureFlagsExposeConstructSignatureAbstractness(t *testing.T) {
   root := t.TempDir()
   writeFile(t, filepath.Join(root, "tsconfig.json"), `{
@@ -46,10 +51,12 @@ export class ConcreteDeclared {
   constructor(value: string) { void value; }
 }
 export abstract class AbstractDefault {}
+export class ConcreteDefault {}
 export class ConcreteBase {
   constructor(value: number) { void value; }
 }
 export abstract class AbstractInheriting extends ConcreteBase {}
+export class ConcreteInheriting extends AbstractDeclared {}
 export type AbstractOpener = abstract new (value: string) => AbstractDeclared;
 export type ConcreteOpener = new (value: string) => ConcreteDeclared;
 `)
@@ -70,12 +77,20 @@ export type ConcreteOpener = new (value: string) => ConcreteDeclared;
     typeAlias    bool
     wantAbstract bool
     wantNilDecl  bool
+    // checkDeclaringClass asserts the abstract modifier on the class that
+    // lexically declares the signature's constructor. It agrees with the
+    // flag on the declared shapes and diverges on both inherited shapes,
+    // pinning that the flag, not the declaration, carries abstractness.
+    checkDeclaringClass    bool
+    declaringClassAbstract bool
   }{
-    {name: "AbstractDeclared", wantAbstract: true},
-    {name: "ConcreteDeclared"},
+    {name: "AbstractDeclared", wantAbstract: true, checkDeclaringClass: true, declaringClassAbstract: true},
+    {name: "ConcreteDeclared", checkDeclaringClass: true},
     {name: "AbstractDefault", wantAbstract: true, wantNilDecl: true},
-    {name: "ConcreteBase"},
-    {name: "AbstractInheriting", wantAbstract: true},
+    {name: "ConcreteDefault", wantNilDecl: true},
+    {name: "ConcreteBase", checkDeclaringClass: true},
+    {name: "AbstractInheriting", wantAbstract: true, checkDeclaringClass: true},
+    {name: "ConcreteInheriting", checkDeclaringClass: true, declaringClassAbstract: true},
     {name: "AbstractOpener", typeAlias: true, wantAbstract: true},
     {name: "ConcreteOpener", typeAlias: true},
   }
@@ -91,7 +106,7 @@ export type ConcreteOpener = new (value: string) => ConcreteDeclared;
     if len(signatures) != 1 {
       t.Fatalf("%s construct signatures = %d, want 1", tc.name, len(signatures))
     }
-    flags := signatures[0].Flags()
+    var flags shimchecker.SignatureFlags = signatures[0].Flags()
     if flags&shimchecker.SignatureFlagsConstruct == 0 {
       t.Fatalf("%s flags = %d: Construct bit missing on a construct signature", tc.name, flags)
     }
@@ -100,6 +115,12 @@ export type ConcreteOpener = new (value: string) => ConcreteDeclared;
     }
     if got := signatures[0].Declaration() == nil; got != tc.wantNilDecl {
       t.Fatalf("%s nil declaration = %v, want %v", tc.name, got, tc.wantNilDecl)
+    }
+    if tc.checkDeclaringClass {
+      class := signatures[0].Declaration().Parent
+      if got := shimast.GetCombinedModifierFlags(class)&shimast.ModifierFlagsAbstract != 0; got != tc.declaringClassAbstract {
+        t.Fatalf("%s declaring class abstract modifier = %v, want %v", tc.name, got, tc.declaringClassAbstract)
+      }
     }
   }
 }
--- a/packages/lint/test/shim/signature_flags_expose_construct_signature_abstractness_test.go
+++ b/packages/lint/test/shim/signature_flags_expose_construct_signature_abstractness_test.go
@@ -14,12 +14,13 @@ import (
 // Signature.Flags() was already reachable through the full Signature alias,
 // but SignatureFlags itself was not nameable, so the returned value could not
 // be tested against SignatureFlagsAbstract (#1203). The flag is also the only
-// correct general answer: a constructor-less class produces a default
-// construct signature with no declaration at all, and a class inheriting its
-// base's constructor clones the base signature, so Declaration() points at
-// the OTHER class's constructor while the checker forces the bit to the
-// derived class's abstractness — in both directions. Declaration-modifier
-// reading returns nothing for the former and the wrong answer for the latter.
+// correct general answer: a class with no constructor and no base produces a
+// default construct signature with no declaration at all, and a class
+// inheriting its base's constructor clones the base signature, so
+// Declaration() points at the OTHER class's constructor while the checker
+// forces the bit to the derived class's abstractness — in both directions.
+// Declaration-modifier reading returns nothing for the former and the wrong
+// answer for the latter.
 //
 //  1. Build a program with abstract/concrete classes and constructor-type
 //     aliases, covering the declared, constructor-less, and inherited shapes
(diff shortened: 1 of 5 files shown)


packages/lint/test/shim/signature_flags_expose_construct_signature_abstractness_test.go
```diff
@@ -0,0 +1,105 @@
...
+    if tc.typeAlias {
+      target = shimchecker.Checker_getDeclaredTypeOfSymbol(prog.checker, symbol)
+    } else {
+      target = shimchecker.Checker_getTypeOfSymbol(prog.checker, symbol)
+    }
+    signatures := shimchecker.Checker_getSignaturesOfType(prog.checker, target, shimchecker.SignatureKindConstruct)
+    if len(signatures) != 1 {
+      t.Fatalf("%s construct signatures = %d, want 1", tc.name, len(signatures))
+    }
+    flags := signatures[0].Flags()
```
samchon: **ISR finding 1 (advisory, accepted).** `flags := signatures[0].Flags()` infers the inner type; every later use is a member const, which compiles without the `SignatureFlags` alias. Deleting `shim.go`'s alias line would leave this probe green — the exact regression #1203 exists to prevent — leaving acceptance criterion 1 pinned only by the audit ratchet. Fix: declare the variable as `var flags shimchecker.SignatureFlags = signatures[0].Flags()` so the probe names the alias.

packages/lint/test/shim/signature_flags_expose_construct_signature_abstractness_test.go
```diff
@@ -0,0 +1,105 @@
...
+    name         string
+    typeAlias    bool
+    wantAbstract bool
+    wantNilDecl  bool
+  }{
+    {name: "AbstractDeclared", wantAbstract: true},
+    {name: "ConcreteDeclared"},
+    {name: "AbstractDefault", wantAbstract: true, wantNilDecl: true},
+    {name: "ConcreteBase"},
+    {name: "AbstractInheriting", wantAbstract: true},
```
samchon: **ISR finding 2 (advisory, accepted).** The table covers the flag-*forcing* inheritance arm (`AbstractInheriting`) but never the flag-*clearing* arm (`checker.go:20463-20465`): `class ConcreteInheriting extends AbstractDeclared {}` clones a base signature whose propagating flags include Abstract, forcing the clear path — and its `Declaration()` then points at an abstract class's constructor while the flag correctly reads non-abstract. That is the stronger witness for the doc comment's claim. Fix: add the class and its row.

packages/lint/test/shim/signature_flags_expose_construct_signature_abstractness_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+  cases := []struct {
+    name         string
+    typeAlias    bool
+    wantAbstract bool
+    wantNilDecl  bool
+  }{
+    {name: "AbstractDeclared", wantAbstract: true},
+    {name: "ConcreteDeclared"},
+    {name: "AbstractDefault", wantAbstract: true, wantNilDecl: true},
```
samchon, line 90: **ISR finding 3 (advisory, accepted).** This is the only `wantNilDecl: true` row and it is also `wantAbstract: true`, so a defect returning a nil declaration exactly when abstract would pass the whole table. `class ConcreteDefault {}` exercises the `len(baseSignatures)==0` + non-abstract arm (flags = Construct only, nil declaration) and breaks the correlation. Fix: add `{name: "ConcreteDefault", wantNilDecl: true}`.

packages/lint/test/shim/signature_flags_expose_construct_signature_abstractness_test.go
```diff
@@ -0,0 +1,105 @@
...
+      t.Fatalf("%s construct signatures = %d, want 1", tc.name, len(signatures))
+    }
+    flags := signatures[0].Flags()
+    if flags&shimchecker.SignatureFlagsConstruct == 0 {
+      t.Fatalf("%s flags = %d: Construct bit missing on a construct signature", tc.name, flags)
+    }
+    if got := flags&shimchecker.SignatureFlagsAbstract != 0; got != tc.wantAbstract {
+      t.Fatalf("%s abstract bit = %v, want %v (flags = %d)", tc.name, got, tc.wantAbstract, flags)
+    }
+    if got := signatures[0].Declaration() == nil; got != tc.wantNilDecl {
```
samchon: **ISR finding 4 (advisory, accepted).** The doc comment's load-bearing claim — `Declaration()` points at a NON-abstract constructor while the bit is forced on — is asserted only as non-nil-ness here. Fix: assert the declaring class's abstract modifier via `shimast.GetCombinedModifierFlags(decl.Parent)` for the inheritance witnesses, pinning the flag/declaration divergence in both directions.

packages/ttsc/shim/checker/shim.go
```diff
@@ -218,6 +219,8 @@ func Checker_isSymbolAccessibleAsValue(recv *innerchecker.Checker, symbol *inner
 const (
   SignatureKindCall = innerchecker.SignatureKindCall
 
+  SignatureFlagsAbstract = innerchecker.SignatureFlagsAbstract
```
samchon, line 220: **ISR finding 7 (nit, accepted).** The alias block reads `Signature`, `SignatureFlags`, `SignatureKind`, but this const group sits after `SignatureKindCall`. Zero functional impact; reordered in the correction commit so the two blocks mirror.

packages/lint/test/shim/signature_flags_expose_construct_signature_abstractness_test.go
```diff
@@ -0,0 +1,105 @@
...
+    symbol := classSymbol(t, prog, tc.name)
+    var target *shimchecker.Type
+    if tc.typeAlias {
+      target = shimchecker.Checker_getDeclaredTypeOfSymbol(prog.checker, symbol)
+    } else {
+      target = shimchecker.Checker_getTypeOfSymbol(prog.checker, symbol)
+    }
+    signatures := shimchecker.Checker_getSignaturesOfType(prog.checker, target, shimchecker.SignatureKindConstruct)
+    if len(signatures) != 1 {
+      t.Fatalf("%s construct signatures = %d, want 1", tc.name, len(signatures))
```
samchon, line 108: **ISR finding 8 (nit, rejected).** `t.Fatalf` aborts the remaining table rows on first failure; `t.Errorf`/subtests would report every shape. Rejected for local consistency: every `test/shim` peer uses `Fatalf`, and the cases share one program where a first-failure abort is acceptable for a probe.

samchon (commented):
## Individual Self-Review — commit 1fad3b498

**What landed:** `feat(shim): expose SignatureFlags to Go source plugins` — resolves #1203. `SignatureFlags` type alias plus the hand-picked `SignatureFlagsAbstract` member in the hand-maintained `shim.go`; the 11 remaining family members generated into `enums_gen.go` by the owning command (delete-then-`shim:audit -fix`); the resolved `ESCAPE|checker|SignatureFlags` baseline line removed so the ratchet enforces the exposure; a runtime probe over seven construct-signature shapes.

**Reviewer verdict: no blocking findings.** Mechanical verification: `enums_gen.go` went 117 → 128 members with zero removals, byte-sorted in place, header and alignment column unchanged; family completeness is exact (12/12 upstream constants — 11 generated + `Abstract` by hand); `baseline.json` stays valid, sorted, 682 entries; exposing the enum cannot unlock a new audit gap (no exported package-level func in `internal/checker` names `SignatureFlags`; the only exported method that does is `(*Signature).Flags()`, the very ESCAPE being resolved). Test semantics verified against the pinned typescript-go source: declaration-derived `Construct`/`Abstract` assignment, the `getDefaultConstructSignatures` nil-declaration branch, and `cloneSignature` keeping the base declaration while the abstract bit is forced. Helper wiring (`writeFile`, `classSymbol`, `loadProgram`, `needsRuleChecker`) and the `test-go-lint` flattening runner verified; filename and test name unique.

**Adjudication of the eight candidates** (line-anchored ones carry inline comments):

- **Accepted — corrections follow in the next commit:** (1) the probe never names the `SignatureFlags` alias itself, so deleting the alias would leave the probe green and only the audit ratchet would object; (2) missing negative twin for the flag-*clearing* inheritance branch (`ConcreteInheriting extends AbstractDeclared`), the inverse witness where `Declaration()` points at an abstract class while the flag correctly reads non-abstract; (3) `wantNilDecl` and `wantAbstract` are perfectly correlated — `ConcreteDefault {}` breaks the correlation via the nil-declaration non-abstract arm; (4) the doc comment's declaration-lies claim is asserted only as nil-ness — strengthened with `shim/ast.GetCombinedModifierFlags` on the declaring class; (7) const-group placement not mirroring the alias block order.
- **Accepted — documentation:** (6) add a `SignatureFlags` abstractness pattern to the tsgo concepts page's Checker patterns; two of the shapes make abstractness a genuinely non-obvious read, which is that section's bar.
- **Deferred to the next cycle** (already recorded in the campaign ledger): (5) `shim_audit -fix` rebuilds `enums_gen.go` from the current gap set only, so an in-place run clobbers previously generated members (observed 128 → 11). The principled resolution is fixing `runFix` (merge or regenerate the full closure) together with its instruction surface — its own issue, outside this cycle's declared single-issue scope.
- **Rejected:** (8) `Fatalf` → `Errorf`/subtests: every `test/shim` peer uses `Fatalf`; local consistency wins for a probe whose cases share one program.

Corrections land as one commit with its own Individual Self-Review, and Overall Self-Review restarts as a fresh round over the new head.

samchon (commented):
## Individual Self-Review — commit f226f1672

**What landed:** `test(shim): pin the SignatureFlags alias and both inherited polarities` — the adjudicated corrections from 1fad3b498's Individual Self-Review: alias-typed flags read, `ConcreteInheriting` flag-clearing witness, `ConcreteDefault` correlation breaker, declaring-class modifier assertions, const-group mirror, and the tsgo concepts abstractness pattern.

**Reviewer verdict: substantively correct, no blocking findings.** All nine case rows verified against the pinned typescript-go source, including the two new arms: `ConcreteDefault` takes the `len(baseSignatures)==0` + non-abstract branch (flags = Construct, nil declaration), and `ConcreteInheriting` arrives as a `cloneSignature` of the abstract base's constructor with `Abstract` propagated in and then explicitly cleared (`checker.go:20465`) — Construct set, Abstract clear, non-nil declaration naming an abstract class. `Declaration().Parent` wiring, `shimast.GetCombinedModifierFlags` semantics on a class node, gofmt-2spaces byte-stability, prettier cleanliness, and the reorder's alignment were all verified mechanically; `enums_gen.go` and `baseline.json` are untouched.

**Adjudication of the six candidates** (line-anchored ones carry inline comments):

- **Accepted — corrections follow in the next commit:** (F1) doc-comment step 3 over-reaches ("where a declaration exists" vs. the deliberate class-rows-only application); (F2) the "constructor-less class" prose in both the doc comment and the mdx omits the no-base qualifier that `checker.go:20441`'s guard actually requires — the fixtures prove the distinction, only the prose generalizes; (F4) the declaring-class read would nil-panic instead of failing with a message on a contradictory future row — cheap guard added; (F5) cite the pinning test from the mdx pattern per the section's own promise; (F6) mdx precision — the declaration points at the base class's constructor declaration, and the modifier answer there is unreliable rather than unconditionally wrong.
- **No change (informational):** (F3) the reviewer corrected the review brief itself: `loadProgram`'s returned diagnostics are config/CLI-layer only, so `len(diags) != 0` does not assert semantic cleanliness. The fixture was independently verified error-free, and the gap is shared by all six shim probes, so it is recorded in the campaign ledger as a possible next-cycle candidate rather than patched into this commit.

Corrections land as one commit with its own Individual Self-Review; Overall Self-Review restarts as a fresh round over the new head.

packages/lint/test/shim/signature_flags_expose_construct_signature_abstractness_test.go
```diff
@@ -0,0 +1,126 @@
...
+//
+//  1. Build a program with abstract/concrete classes and constructor-type
+//     aliases, covering the declared, constructor-less, and inherited shapes
+//     in both abstract polarities.
+//  2. Obtain every construct signature only through the exported shim surface,
+//     reading flags into a shimchecker.SignatureFlags-typed variable so the
+//     alias itself is pinned, not just the member consts.
+//  3. Assert the Abstract bit exactly where the source says abstract, the
+//     Construct bit everywhere, the nil declaration on both default-signature
+//     boundaries, and the declaring class's modifier where a declaration
```
samchon: **ISR-2 F1 (nit, accepted).** Step 3 claims the declaring-class modifier is asserted "where a declaration exists", but declarations exist on 7 of 9 rows and the check deliberately covers only the 5 class rows (the type-alias declarations parent into a TypeAliasDeclaration, where the read is meaningless). The application is right; the prose over-reaches. Fix: "on every class-declared shape".

packages/lint/test/shim/signature_flags_expose_construct_signature_abstractness_test.go
```diff
@@ -0,0 +1,126 @@
...
+)
+
+// Verifies SignatureFlags exposes construct-signature abstractness through the
+// shim.
+//
+// Signature.Flags() was already reachable through the full Signature alias,
+// but SignatureFlags itself was not nameable, so the returned value could not
+// be tested against SignatureFlagsAbstract (#1203). The flag is also the only
+// correct general answer: a constructor-less class produces a default
+// construct signature with no declaration at all, and a class inheriting its
```
samchon: **ISR-2 F2 (advisory, accepted).** "A constructor-less class produces a default construct signature with no declaration at all" holds only under `len(baseSignatures)==0` (checker.go:20441) — the very next clause describes constructor-less classes whose default signature DOES carry a declaration. The fixtures (`ConcreteDefault` vs `ConcreteInheriting`) prove the distinction; the sentence needs the no-base qualifier.

packages/lint/test/shim/signature_flags_expose_construct_signature_abstractness_test.go
```diff
@@ -0,0 +1,126 @@
...
+      t.Fatalf("%s flags = %d: Construct bit missing on a construct signature", tc.name, flags)
+    }
+    if got := flags&shimchecker.SignatureFlagsAbstract != 0; got != tc.wantAbstract {
+      t.Fatalf("%s abstract bit = %v, want %v (flags = %d)", tc.name, got, tc.wantAbstract, flags)
+    }
+    if got := signatures[0].Declaration() == nil; got != tc.wantNilDecl {
+      t.Fatalf("%s nil declaration = %v, want %v", tc.name, got, tc.wantNilDecl)
+    }
+    if tc.checkDeclaringClass {
+      class := signatures[0].Declaration().Parent
```
samchon: **ISR-2 F4 (nit, accepted).** `signatures[0].Declaration().Parent` is nil-safe only because every `checkDeclaringClass` row happens to have `wantNilDecl: false`; a contradictory future row would panic instead of failing with a message. Fix: read the declaration into a variable and `t.Fatalf` when it is nil.

website/src/content/docs/development/concepts/tsgo.mdx
```diff
@@ -412,6 +412,17 @@ if len(sigs) > 0 {
...
+**Construct-signature abstractness.**
+
+```go
+sigs := checker.GetSignaturesOfType(t, shimchecker.SignatureKindConstruct)
+for _, sig := range sigs {
+    if sig.Flags()&shimchecker.SignatureFlagsAbstract != 0 { /* not new-able */ }
+}
+```
+
+Read abstractness from `Signature.Flags()`, not from the declaration's modifiers: a constructor-less class's default construct signature has no declaration at all, and an inherited constructor's `Declaration()` points at the base class, so the modifier answer is missing in the first case and wrong in the second. A bare `abstract new (...) => T` type literal sets the same flag.
```
samchon: **ISR-2 F2/F5/F6 (accepted).** Same no-base qualifier as the test doc comment; "points at the base class" → "the base class's constructor declaration"; "wrong" → "unreliable" (it coincides whenever derived and base abstractness match); and cite the pinning probe as the working form per this section's intro promise.

samchon (commented):
## Individual Self-Review — commit fd81b0e51

**What landed:** `docs(shim): qualify the no-declaration shape and guard the class read` — the accepted prose/robustness corrections from f226f1672's Individual Self-Review.

**Reviewer verdict: no blocking findings.** Diff scope verified exactly as claimed (two files; the only behavioral change is the nil guard; no assertion weakened). All four reworded claims re-verified against the pinned typescript-go: the no-constructor-and-no-base qualifier is a correct sufficient condition for the nil-declaration default signature; `cloneSignature` keeps the base constructor declaration while the abstract bit is forced in both directions; "unreliable" is strictly more accurate than "wrong" for the general inherited case; step 3's scope matches the five class rows exactly. gofmt-2spaces byte-stability and prettier cleanliness reproduced mechanically; failure-message style consistent with existing helpers; repo-wide sweep found no other copy of the old phrasing.

**Adjudication of the five candidates — all accepted** (line-anchored ones carry inline comments); corrections follow in the next commit:

- (A) step 1 still says "constructor-less", the exact term the paragraph above was corrected away from — becomes "default-signature";
- (B) "the wrong answer for the latter" is general-claim phrasing that only holds for divergent polarities — aligned to "an answer that tracks the base class";
- (C) the mdx citation becomes a GitHub link matching every other repository path on the page;
- (D) "the base class's constructor declaration" is one-level-exact only — becomes "the ancestor class that declared it";
- (E) `signatures[0].Declaration()` called twice — hoisted above the nil-declaration assertion and reused.

Corrections land as one commit with its own Individual Self-Review; Overall Self-Review restarts as a fresh round over the new head.

packages/lint/test/shim/signature_flags_expose_construct_signature_abstractness_test.go
```diff
@@ -0,0 +1,131 @@
...
+// correct general answer: a class with no constructor and no base produces a
+// default construct signature with no declaration at all, and a class
+// inheriting its base's constructor clones the base signature, so
+// Declaration() points at the OTHER class's constructor while the checker
+// forces the bit to the derived class's abstractness — in both directions.
+// Declaration-modifier reading returns nothing for the former and the wrong
+// answer for the latter.
+//
+//  1. Build a program with abstract/concrete classes and constructor-type
+//     aliases, covering the declared, constructor-less, and inherited shapes
```
samchon: **ISR-3 A (advisory, accepted).** Step 1 still labels the shape "constructor-less" — the exact term the why-paragraph was just corrected away from, and it no longer partitions the fixture (both inheriting classes are equally constructor-less). Fix: "the declared, default-signature, and inherited shapes".

packages/lint/test/shim/signature_flags_expose_construct_signature_abstractness_test.go
```diff
@@ -0,0 +1,131 @@
...
+//
+// Signature.Flags() was already reachable through the full Signature alias,
+// but SignatureFlags itself was not nameable, so the returned value could not
+// be tested against SignatureFlagsAbstract (#1203). The flag is also the only
+// correct general answer: a class with no constructor and no base produces a
+// default construct signature with no declaration at all, and a class
+// inheriting its base's constructor clones the base signature, so
+// Declaration() points at the OTHER class's constructor while the checker
+// forces the bit to the derived class's abstractness — in both directions.
+// Declaration-modifier reading returns nothing for the former and the wrong
```
samchon: **ISR-3 B (advisory, accepted).** "The wrong answer for the latter" is phrased as a general claim, but the modifier answer coincides whenever base and derived share abstractness; it is unconditionally wrong only for the divergent polarities this fixture picks. Fix: "an answer that tracks the base class for the latter".

website/src/content/docs/development/concepts/tsgo.mdx
```diff
@@ -412,6 +412,17 @@ if len(sigs) > 0 {
...
+**Construct-signature abstractness.**
+
+```go
+sigs := checker.GetSignaturesOfType(t, shimchecker.SignatureKindConstruct)
+for _, sig := range sigs {
+    if sig.Flags()&shimchecker.SignatureFlagsAbstract != 0 { /* not new-able */ }
+}
+```
+
+Read abstractness from `Signature.Flags()`, not from the declaration's modifiers: a class with no constructor and no base produces a default construct signature with no declaration at all, and an inherited constructor's `Declaration()` points at the base class's constructor declaration, so the modifier answer is missing in the first case and unreliable in the second. A bare `abstract new (...) => T` type literal sets the same flag. Pinned by `packages/lint/test/shim/signature_flags_expose_construct_signature_abstractness_test.go`.
```
samchon: **ISR-3 C+D (nits, accepted).** The citation is bare inline code while every other repository path on this page is a GitHub link — linked in the same form. And "the base class's constructor declaration" is exact for a one-level chain only; with deeper chains the clone points at whichever ancestor declared the constructor — reworded to "the ancestor class that declared it".

packages/lint/test/shim/signature_flags_expose_construct_signature_abstractness_test.go
```diff
@@ -0,0 +1,131 @@
...
+      t.Fatalf("%s construct signatures = %d, want 1", tc.name, len(signatures))
+    }
+    var flags shimchecker.SignatureFlags = signatures[0].Flags()
+    if flags&shimchecker.SignatureFlagsConstruct == 0 {
+      t.Fatalf("%s flags = %d: Construct bit missing on a construct signature", tc.name, flags)
+    }
+    if got := flags&shimchecker.SignatureFlagsAbstract != 0; got != tc.wantAbstract {
+      t.Fatalf("%s abstract bit = %v, want %v (flags = %d)", tc.name, got, tc.wantAbstract, flags)
+    }
+    if got := signatures[0].Declaration() == nil; got != tc.wantNilDecl {
```
samchon: **ISR-3 E (nit, accepted).** `signatures[0].Declaration()` is called here and again inside the `checkDeclaringClass` block. Hoisted into one `declaration` variable above this assertion and reused; the nil guard below stays as defense-in-depth for a future contradictory row, exactly as the commit message claims.

website/src/content/docs/development/concepts/tsgo.mdx
```diff
@@ -412,6 +412,17 @@ if len(sigs) > 0 {
...
+**Construct-signature abstractness.**
+
+```go
+sigs := checker.GetSignaturesOfType(t, shimchecker.SignatureKindConstruct)
+for _, sig := range sigs {
+    if sig.Flags()&shimchecker.SignatureFlagsAbstract != 0 { /* not new-able */ }
+}
+```
+
+Read abstractness from `Signature.Flags()`, not from the declaration's modifiers: a class with no constructor and no base produces a default construct signature with no declaration at all, and an inherited constructor's `Declaration()` points at the ancestor class that declared it, so the modifier answer is missing in the first case and unreliable in the second. A bare `abstract new (...) => T` type literal sets the same flag. Pinned by [`packages/lint/test/shim/signature_flags_expose_construct_signature_abstractness_test.go`](https://github.com/samchon/ttsc/tree/master/packages/lint/test/shim/signature_flags_expose_construct_signature_abstractness_test.go).
```
samchon: **ISR-4 finding 1 (advisory, accepted).** "Points at the ancestor class that declared it" names the class, but `Declaration()` returns the constructor declaration node — the class is one `.Parent` hop away (exactly what the probe and `checker.go:19472` do). Following the sentence literally, a plugin author would read the constructor node's own modifiers and never find `abstract`. Fix: "points at the constructor declared by the ancestor class".

samchon (commented):
## Individual Self-Review — commit f11ff088b

**What landed:** `docs(shim): align the probe prose with the general inherited case` — the five accepted polish findings from fd81b0e51's Individual Self-Review.

**Reviewer verdict: no blocking findings.** Diff scope is exactly the five changes (2 files, 7/7 lines, `git diff --check` clean). The hoist is provably behavior-identical (`Signature.Declaration()` is a pure field read; call order and assertion order preserved; the nil guard stays defense-in-depth). Every reworded claim re-verified against the pinned typescript-go, including the multi-level chain: each clone carries the original declaration forward, so "ancestor" is the correct general term. gofmt-2spaces and prettier both byte-stable; the GitHub link resolves and matches all 14 other path links on the page; repo-wide grep finds no remaining copy of the corrected prose.

**Adjudication of the two candidates:**

- **Accepted — correction follows:** (1) the mdx now says `Declaration()` points at the ancestor *class*, but it returns the constructor *declaration node* — the class is one `.Parent` hop away, which both the probe and the checker itself perform. A reader following the sentence literally would read the constructor's own modifiers and never find `abstract`. Reworded to "points at the constructor declared by the ancestor class", keeping this commit's ancestor gain and restoring the node precision the Go doc comment kept all along.
- **Rejected:** (2) non-greedy rewrap of the step-1 bullet — cosmetic, within the file's 79-column ceiling, gofmt does not reflow comments; not worth diff churn.
- **Noted, no action:** the one remaining unlinked path on the page (`shim.go` at mdx:388) is a pre-existing inline parenthetical outside this cycle's scope.

The correction lands as one commit with its own Individual Self-Review; Overall Self-Review restarts as a fresh round over the new head.

samchon (commented):
## Individual Self-Review — commit 3c6bc4b32

**What landed:** `docs(shim): point the abstractness pattern at the constructor node` — the single accepted finding of f11ff088b's Individual Self-Review.

**Reviewer verdict: CLEAN — no findings.** Diff scope verified exact (one mdx line). The reworded sentence was re-traced through the pinned typescript-go's full inherited-constructor path: `cloneSignature` and `instantiateSignatureEx` both carry the original constructor declaration node forward, and multi-level chains resolve transitively to whichever ancestor declared the constructor — so both halves of the new wording (constructor node, ancestor class) are exact. The mdx now agrees with the probe's own doc comment instead of contradicting it; prettier is a no-op; nothing else in the repository carries the old wording. Two further rewordings were considered by the reviewer and rejected as precision-for-noise trades.

**Adjudication: nothing to change.** The Individual Self-Review loop converges on this commit — five commits, five reviews, every accepted finding applied and re-reviewed.

samchon (commented):
## Overall Self-Review — final clean round over master..3c6bc4b32

Solo whole-diff review of the complete pull-request surface (5 files: `shim.go` +3, `enums_gen.go` +11, `baseline.json` −1, the runtime probe +131, `tsgo.mdx` +11), repeated as a fresh full round after every correction commit. Five rounds total; rounds 1–4 each triggered or absorbed corrections adjudicated from the Individual Self-Reviews, and **round 5, over the final head 3c6bc4b32, found nothing to improve**. The Individual Self-Review loop converged independently: the fifth review returned clean with zero candidates.

**Evidence carried by the rounds:**

- **API surface:** the exposure is additive-only (one type alias, twelve member consts) and exactly closes the upstream family — `internal/checker` declares twelve `SignatureFlags` constants; eleven generated plus the hand-picked `Abstract` — with no duplicate-const risk (the generator excludes members already re-exported by hand).
- **Generated artifacts:** `enums_gen.go` was rebuilt by its owning command from a deleted file so the full 128-member closure reconstructed; the base-to-head diff is purely additive (+11), header intact. The in-place `-fix` clobber hazard discovered while doing this is recorded in the campaign ledger as a next-cycle candidate.
- **Audit gate:** `pnpm --filter ttsc shim:audit` (the exact CI command) exits 0 locally after the baseline line removal — enum closure complete, no new FUNC/ESCAPE/PRODUCER gap unlocked, and the run type-checks the shim package via go/types.
- **Test quality:** the probe covers nine construct-signature shapes — declared, default-signature, and inherited, each in both abstract polarities, plus both constructor-type literals — pins the `SignatureFlags` alias itself with an explicitly typed read, asserts the Construct bit everywhere, both nil-declaration boundaries, and the declaring-class modifier with divergence on both inherited shapes. Every expectation was verified against the pinned typescript-go source, not against the code's own output.
- **Consequence surface:** no in-repo consumer workaround to migrate; no packaging, platform, concurrency, or compatibility interaction; the only documentation surface owed (the tsgo concepts Checker patterns) gained the abstractness pattern with a link to its pinning probe; test filename is unique under the flattening runner.

**Verification not run locally, by standing rule (CI owns it):** Go compile/test of the probe (`test-go` lanes), platform builds, and the typia/nestia consumer workflows. The merge gate is green required checks on this same head 3c6bc4b32.

With the closing reconciliation done (the `Close #1203` line lands from commit 1fad3b498 and no revert touches it), this pull request merges once CI settles green on 3c6bc4b32.

Issue #1209, opened 2026-08-17
feat(evidence): let a Markdown reference demand a per-host checklist


samchon:
# Let a Markdown reference demand a per-host checklist

## Problem

`evidence/graph` asks its coverage question once per claim-reference obligation: does _some_ declaration anywhere in the claim acknowledge each selected unit. `acknowledged` is filled by walking `state.Declarations` for the whole claim and read back against `reference.Units`, and an aggregate target contributes every selected descendant through `UnitsByScope`. One host citing `docs/checklist.md` therefore discharges the entire population on behalf of every other host in the claim.

That is right for a documentary obligation and wrong for a document used as a checklist — development principles, review rules, a release gate — where the question is whether _each_ host answered _each_ item.

Nothing in the current policy set expresses it:

- `singleEvidencePerSymbol` sits on the same axis (distinct selected units per claim host) with the opposite value, exactly 1.
- `uniqueEvidence` is the inverse relation, at most one host per unit.
- `noEvidenceExclude` and `requireReview` decide the kind and the expiry of an acknowledgement, never its distribution across hosts.

The only way to express a checklist today is one claim per host file, which a static configuration cannot scale to.

## Evidence

Coverage is judged per obligation, not per host — `packages/evidence/native/graph.go`:

```go
acknowledged := map[string]bool{}
// ... filled from every declaration of the claim ...
for _, unit := range reference.Units {
    if !acknowledged[unit.ID] {
        // Missing acknowledgement ...
    }
}
```

An aggregate scope cascades to every selected descendant, which is what lets one file-level citation answer a whole document — `packages/evidence/native/graph.go`:

```go
for _, unit := range referenceState.Units {
    for scope := unit; scope != nil; scope = availableUnits[scope.ParentID] {
        referenceState.UnitsByScope[scope.ID] = append(
            referenceState.UnitsByScope[scope.ID],
            unit,
        )
        // ...
    }
}
```

The decoded policy surface is the four flags and nothing on this axis — `packages/evidence/native/config.go`:

```go
decodeFlag("noEvidenceExclude", &policy.NoExclude)
decodeFlag("uniqueEvidence", &policy.UniqueEvidence)
decodeFlag("singleEvidencePerSymbol", &policy.SingleEvidencePerSymbol)
decodeFlag("requireReview", &policy.RequireReview)
```

## Proposed approach

[...]

PR #1210, opened 2026-08-17
feat(evidence): let a Markdown reference demand a per-host checklist

samchon:
## Intent

This pull request owns the complete accepted cycle of the `evidence-checklist` solo issue campaign. Its accepted issue set is exactly one issue:

- #1209 — let a Markdown reference demand a per-host checklist

The branch opens implementation-free. Code, tests, and documentation follow as commits on this branch.

## Scope

`evidence/graph` asks its coverage question once per claim-reference obligation, so one host citing a document discharges the whole population for every other host in the claim. A document used as a checklist — development principles, review rules, a release gate — needs each host to answer each item. This cycle adds a `checklist` option to the Markdown reference that changes the obligation from per-reference to per-host, and refuses an `@evidence` target that names an unselected ancestor scope so a single file-level citation cannot tick every box.

The design was settled with the maintainer before the cycle opened:

- one fused option rather than two orthogonal axes;
- named `checklist` rather than `checklistEvidence`, because it declares what the population is instead of constraining a tag;
- Markdown reference only, since per-host coverage is meaningless for Swagger, Prisma, and TypeScript populations;
- flat boolean on the existing reference shape — a `policy` sub-object and a `policy: [...]` array were both evaluated and rejected.

## Deferred

None. The cycle carries one issue and no deferred work.

The issue records one open question — a TypeScript claim has no file-level host, so a checklist over TypeScript multiplies by every exported symbol. That is a separate ergonomic decision about host granularity, not a precondition for this feature, and it is not resolved here.

## Verification

Pending. Verification is recorded as Individual Self-Review reviews per commit and one Overall Self-Review round over the settled head, alongside ordinary pull-request CI.


samchon (commented):
## Individual Self-Review — 607167c58 `feat(evidence): let a Markdown reference demand a per-host checklist`

Commit under review: the implementation of #1209. Adds `checklist` to the Markdown reference, moving the coverage obligation from the reference to each selected claim host, with the aggregate refusal, the per-host duplicate and conflict keys, the claim-wide carrier expansion, the decode refusals, tests, and documentation.

**Verdict: FINDINGS.** Seven candidates reported, all adjudicated.

### Accepted and fixed in 320d7d4b7

**F1 — a citation of a *selected* ancestor still discharged its whole subtree.** Accepted, and this was the serious one. The refusal only rejected a target that was not itself a selected unit, so the option was inert under the default Markdown selector, where the file is an item: one `@evidence docs/rules.md` resolved to a selected unit, cascaded, and answered every heading. Reproduced independently before accepting, with zero diagnostics emitted. A positive citation now answers the unit it names and nothing beneath it, and the regression twin fails against this commit.

**F2 — two published documentation statements were false against the shipped behavior.** Accepted. Both website sentences, plus the option's own doc comment, the public JSDoc, and the evidence skill, are rewritten to state the implemented rule.

**F3 — no negative twin for the ancestor-that-is-an-item shape.** Accepted, and it is the gap that hid F1. The existing case cited both the parent and the child, so it only proved the citation was not refused. The new case cites the parent alone and asserts the child is still owed, plus a default-selector arm.

**F5 — the no-host coverage path suppressed nothing for a refused aggregate.** Accepted despite being unreachable through `activeGraphConfig` today. The guard exists precisely because a selection change could make it reachable, so leaving its behavior wrong defeats the reason for keeping it. Suppression is now recorded claim-wide as well as per host.

**F6 — dead nil guard.** Accepted and removed.

### Rejected, with reasons

**F4 — a repeated aggregate target on one host loses its `Duplicate @evidence` diagnostic.** Correct as read, not worth changing. Two identical aggregate messages for two identical wrong tags each name their own line, and both point at the same repair. Moving the aggregate check after the duplicate check would produce a duplicate plus two aggregates, which is more output for no more information.

**F7 — em-dash on a rewritten Markdown line in the evidence skill.** The Voice rule is enforced on `*.md` and `*.mdx` under `website/`, where the changed guides contain zero em-dashes. `.agents/skills/project/evidence/SKILL.md` carries dozens in pre-existing prose, and the flagged one is pre-existing text this commit reflowed rather than authored. Changing it alone would be inconsistent with its own file.

### Confirmed correct by the review

Ordinary-path regression freedom under the new composite map key, verified by probe as byte-identical; the empty-population early `continue`; duplicate exclusion and conflicting acknowledgement inside one host; a declaration carrying several semantic hosts; partial aggregate suppression; the unhosted carrier expansion and its conflict detection; the decode refusals including the disabled-claim path and the malformed-shape path; determinism of every diagnostic string; and the public API placement.

samchon (commented):
## Individual Self-Review — 013de6dc8 `test(evidence): pin the checklist contract claims the first commit left uncovered`

Commit under review: the first Overall Self-Review round's corrections. Three added cases (Markdown claim, reviewed checklist, exclusion ledger) and one comment fix.

**Verdict: FINDINGS.** Four candidates, all accepted and fixed in 320d7d4b7.

**1 — the reviewed-checklist case pinned nothing about a checklist.** Accepted, and this is the sharpest kind of finding: the case passed with `checklist` deleted from its own config, so it verified plain `requireReview` behavior while its headline announced the checklist property. A fingerprint belongs to the cited address, so two targets expire separately under any reference. What the option actually contributes is that the one-tag shortcut stops being available. The case now asserts both arms against each other: the ordinary twin accepts one document-wide citation carrying one fingerprint, and the checklist refuses that citation and reports no review diagnostic beside the refusal. It derives its own ordinary config by deleting the option from the checklist one and fails if that deletion stops taking effect, so the case cannot silently degrade again.

**2 — the Markdown-claim case judged only the extremes.** Accepted. A host answering 2 of 2 and a host answering 0 of 2 leave a regression that discharges a host on its first acknowledgement completely green. Added a partial host answering 1 of 2, plus the `must not appear` guard the TypeScript twin already carried.

**3 — the Markdown assertion omitted the item list.** Accepted. Item ordering is deterministic and every sibling case pins the list, so the omission had no reason. Extended.

**4 — the misplaced-exclusion assertion pinned only the message title.** Accepted, and it matters because this is the suite's only runtime coverage of that diagnostic. It now pins the target, the carrier globs, and the repair sentence.

### Confirmed correct by the review

That the other two added cases genuinely collapse when `checklist` is removed; that the negative assertion in the expiry case is real rather than vacuous, since a stale review does not withdraw coverage; that `everyExpectedFingerprint` cannot mistake a stale message's outdated value for the expected one; determinism and per-case temp-directory isolation; doc-comment shape; absence of duplication against the existing review-fingerprint suites; and the comment fix itself.

### Note

The reviewer observed an uncommitted addition appearing in the working tree mid-review. That was this session's F1 regression twin being written concurrently, which the solo campaign procedure expects: the main agent continues implementing while an individual review runs. It is committed in 320d7d4b7.

samchon (commented):
## Individual Self-Review — 320d7d4b7 `fix(evidence): stop a checklist citation from discharging the items beneath it`

Commit under review: the fix for the cascade defect the previous review found, plus the four test corrections and five documentation corrections it carried.

**Verdict: FINDINGS.** The core fix was confirmed correct and complete, with no third cascade path, no hole opened in conflict detection, no leak into non-checklist references, and the regression twin verified to fail against the parent commit. Three low-severity candidates, all accepted and fixed in 3fee41830.

**1 — `aggregateAnyHost` shipped with no test reaching it.** Accepted. It reads as a mirror of the per-host suppression set and is not one: it is what the population-wide coverage path uses when no host was selected, and it changes that path's output. The only hostless case beside it carries no declarations, so nothing entered the branch. The reviewer reproduced the difference against the parent commit. Added the evaluator-level case that enters it and asserts the aggregate diagnostic stands alone.

**2 — the counting path still expands the subtree for a positive tag.** Accepted as a comment. `singleEvidencePerSymbol` and `uniqueEvidence` still iterate the whole selected subtree, which is exactly the behavior this commit took away from a positive citation, sitting twenty-five lines above the fix. It is correct only because both options are refused beside `checklist` at decode and a reference carrying a decode problem never reaches evaluation. Nothing recorded that, so relaxing the refusal would restore the cascade with no test failing. The reason is now stated at the site, naming the change that would have to move it.

**3 — two em-dashes added to a Markdown line in the evidence skill.** Accepted, reversing the previous round's disposition. That round declined on the grounds that the vendored file already carries its own, which is a reason not to rewrite the file and no reason to add more. Reworded; the file is back to its baseline count.

### Confirmed correct by the review

That every write to the acknowledgement maps flows through the one narrowed loop, and that Markdown units carry no aliases or published addresses through which a second path could exist; that `scopesByID[scopeID]` is the same pointer stored in `Units`, so the conflict diagnostic's target and readable name are unchanged; that the nil-guard removal is safe because the map is preallocated for every selected host; that conflict and duplicate detection survives the narrowing in all four declaration orderings, including the case where an exclusion of the document meets a citation of a child; that an ordinary reference still cascades; that both arms of the regression twin fail against the parent; that each corrected test is now checklist-locked, failing when the option is dropped from its own config; and that all five rewritten prose surfaces state both halves of the rule, with both website guides carrying zero em-dashes.

samchon (commented):
## Individual Self-Review — 3fee41830 `test(evidence): cover the population-wide aggregate suppression and pin the counting path`

Commit under review: the third review round's corrections. One evaluator-level test for the claim-wide aggregate suppression, one explanatory comment above the cardinality counters, one Markdown rewording.

**Verdict: FINDINGS.** Two candidates, both confirmed by mutation and both accepted, fixed in 0822c7d28.

**1 — nothing separated per-host suppression from claim-wide suppression.** Accepted, and it is the more valuable of the two. The reviewer swapped the per-host lookup for the claim-wide set and the entire native package stayed green. The real consequence of that leak is not cosmetic: one host's refused aggregate would silence every other host's shortfall, so a host owing the whole checklist passes with no diagnostic anywhere. The aggregate arm had a single host, which is exactly why it could not see the difference. It now carries a second host with no tag, owing both items while the citing host's items stay suppressed, and the same mutation fails.

**2 — the new test's doc comment promised more than the body asserted.** Accepted. Step 3 said the aggregate diagnostic is the only one while the body checked only that `Missing acknowledgement` was absent. The count assertion its sibling already uses is added.

### Confirmed correct by the review

That the added test is the sole guard on the claim-wide suppression, verified by removing the guard and observing it as the only failure in the package; that its hand-built `claimState` is faithful to what `materializeClaimStates` produces for the same shape, field by field, including scope walk, unit IDs, readable names, and parent links; that both halves of the new comment above the cardinality counters are true, since the decoder refuses both options beside a checklist and `Check` returns before evaluation on any decode problem, with the only other decoder caller computing watch topology alone; and that the Markdown rewording preserves its meaning and returns the file to its baseline em-dash count of eighteen.

### Note on framing

The reviewer correctly rejected one premise in the brief it was given. I asked whether the new test fails against its parent; it does not, because the parent is the commit that introduced the suppression it covers. The meaningful pin is the mutation result, which holds.

samchon (commented):
## Individual Self-Review — 0822c7d28 `test(evidence): separate per-host aggregate suppression from claim-wide suppression`

Commit under review: the fourth round's correction. A tag-free second host added to the aggregate arm, and a message-count assertion on the no-selected-host case.

**Verdict: FINDINGS.** Three candidates, all confirmed by mutation, all accepted and fixed in a5b8ba7a4.

**1 — the guard added by this commit could not fail.** Accepted, and it is the worst kind of defect a test commit can carry. It asked whether the aggregate diagnostic names the citing host as `'broad'`; that message carries the source path rather than the host's readable name, so the condition was false in the passing world, and in the failing world it was either unreachable or still false. It read as coverage of "the citing host is named only by its refusal" while pinning nothing at all. Replaced with a count over the whole message set, which is what the name could not do. I reached the same conclusion independently while the review was running and had already replaced it in the working tree; the review confirmed the reasoning and the two failing-world cases.

**2 — a partial aggregate's suppression was unpinned.** Accepted. Suppression iterates the aggregate's own subtree, but every case cited a scope covering the whole population, so the subtree and the population were indistinguishable and widening the suppression to the population passed the entire suite. The consequence is concrete: a host's remaining, unrelated shortfall disappears behind an aggregate diagnostic that never mentioned it. Added an H3 checklist under two H2 parents where the cited scope covers one item and the other stays owed. The mutation now fails.

**3 — the exclusion arm was still single-host.** Accepted, and it is the exclusion twin of exactly what this commit set out to separate. An aggregate exclusion that spread across the claim instead of discharging its own host was undetected, which means one host's "none of this applies here" would answer for every host. Added the silent host the positive arm had just gained. That mutation now fails too.

### Confirmed correct by the review

That the strengthened positive arm catches the mutation named in its own commit message, and that the same mutation against the parent's test file passed the whole suite, so the justification was literal rather than rhetorical; that the inverted regression, suppressing the non-citing hosts, is caught by the silent host assertion; that no existing assertion was weakened when the exact count moved from zero to one; that the doc comment tracks the new fixture; and that the added message-count assertion matches an existing idiom in the same file and is not redundant with the guard beneath it.

### Noted, not changed

The reviewer observed that the no-selected-host case builds a synthetic state whose declaration names a host the claim does not contain, so a future diagnostic about hostless declarations would trip its count assertion with a misleading message. Left as is: the assertion is correct today, and the suggested reordering of the two checks is not worth a commit of its own. Recorded here so the next reader meets the reasoning rather than rediscovering it.

samchon (commented):
## Individual Self-Review — a5b8ba7a4 `test(evidence): pin what a refused aggregate suppresses, and on which host`

**Verdict: FINDINGS.** No correctness defect; every targeted mutation was already killed, and two of the added arms were confirmed as the sole detector in the package for their mutation. Three low-severity items plus a ranked gap list, all adjudicated.

### Accepted and fixed in 0343945e0

**The published guide documented the aggregate refusal but never the suppression that travels with it.** The skill said the items a refused target covers are withheld from that host's shortfall; the website page did not, so a reader could not tell whether the refusal was the complete statement of the repair. Added.

**The comment explaining the removed guard's vacuity gave half the reason.** The message carrying a source path rather than a readable host name is why its first conjunct is false today; the conjunction with a not-equal-one count is why no guard of that shape could fail even once the name appeared. A future author reading only the first half would reintroduce it.

**Two item-list assertions were truncated** where every peer in the file pins the full list. Extended.

### Ranked gaps: closed the top three

Closed: a ledger dropping the whole document, which is the highest-consequence silent pass on this surface since one carrier tag discharges every item for every host; `requireReview` against the negative tag, which is where an exclusion review is written most often; and a mixed variable statement, one node declaring two selected hosts under one documentation block.

The exclusion-review case surfaced something worth recording: the shared fingerprint reader recognizes `Unreviewed @<tag>` for either kind but only `Unfingerprinted @evidenceReview`, so probing with a blank exclusion review returns no expected value and a case written that way would look complete while measuring the citation alone. The case omits the review lines entirely instead.

Left open on the reviewer's own recommendation, with the reasoning checked: a Prisma claim, since the per-host loop is claim-kind agnostic and the one Prisma-specific branch is the same line the TypeScript carrier already exercises; a withdrawn host, covered generically elsewhere; two overlapping checklist references, where the harder direction is already pinned; and a withdrawn unit inside a checklist population, which Markdown cannot produce.

samchon (commented):
## Overall Self-Review — round on `163fa1e66`

Solo, whole base-to-head diff, fresh. This is the merge gate; the seven Individual Self-Reviews recorded above are advisory and do not substitute for it.

### Round B, on head `0343945e0` — one finding, applied

Read the complete evaluator diff, the configuration diff, both public types, the evidence skill, and both website guides. Probed the core promise directly: two hosts answering opposite items, one citing what the other excludes, produced zero diagnostics, which is per-host independence working rather than inferred.

The finding: the claim-wide reach of a ledger exclusion comes from the carrier not being a selected host, never from the file it sits in. A ledger whose placeholder export is itself selected answers for that placeholder alone, and the placeholder then owes the checklist like any other host. Nothing said so and nothing pinned it, and the guide's own carrier example uses a file that would carry such an export. Applied in `163fa1e66` as one guide sentence naming the repair and one test arm measuring it.

### Round C, on head `163fa1e66` — clean

Re-read the configuration, model, public type, skill, and tags diffs. No finding.

Nine adversarial probes over combinations no test reaches, all behaving correctly: a document that is both claim host and checklist source; a duplicate identical citation on one host; a disabled claim; two claims sharing one checklist document; a reference `root` moving the address space; an `@internal` host; class members as hosts; a reference glob matching no file; and a `symbol: "file"` checklist where the document is the single item.

Mutation re-verification at this head. Each of these fails at least one test: widening the suppression from the aggregate's own subtree to the whole population; spreading an aggregate exclusion claim-wide; spreading every exclusion claim-wide; and swapping per-host suppression for claim-wide suppression.

### What the cycle actually caught

Four defects, none of which a green suite would have surfaced on its own.

The option was **inert under the default Markdown selector**, where the file is itself an item, so one document citation discharged every heading and the graph emitted nothing. That is the failure the feature exists to prevent, shipped in the feature itself.

A **test that verified nothing**: the reviewed-checklist case passed with the option deleted from its own config, so it measured plain `requireReview`.

A **guard that could not fail**, matching a host's readable name against a message that carries the source path.

**Two unpinned suppression axes**, per-host versus claim-wide and the aggregate's subtree versus the population, both mutation-verified as silent before the arms were added.

### Carried forward, deliberately

A TypeScript claim has no file-level host, so a checklist over TypeScript multiplies by every exported symbol. That is recorded as the open question in #1209 and is a separate decision about host granularity, not a precondition for this feature.

samchon (commented):
## Individual Self-Review — 0343945e0 `test(evidence): close the checklist coverage gaps a merge should not leave open`

**Verdict: FINDINGS.** Three items, none blocking. All adjudicated; two applied in 3f64fb3fb.

**1 — the new review case's stated reason named a branch it never reached.** Accepted, and it is the same defect class this cycle has now hit three times: prose asserting coverage the code does not have. The case said the hazard it pinned was placement, that an exclusion carried by a non-host declaration answers for every host and its review must be writable on the carrier. Both of its tags sat on a selected host. Fixed by making the claim true rather than by softening it: a second arm puts the exclusion on a ledger carrier the claim's `symbol` does not select, asserts it owes `@evidenceExcludeReview` there, and asserts the reviewed form passes. Verified load-bearing by mutation, since making `requireReview` ignore the negative tag fails it.

**2 — the review ledger's source-position fallback has no coverage.** Correct and confirmed by the reviewer's mutation, and correctly marked out-of-commit. Investigated rather than accepted at face value, and the reviewer's suggested repair does not fit: an interface carrier registers a semantic host and is merely unselected, so the arm added for finding 1 does not reach the fallback either. The fallback wants a declaration with no semantic host at all, which among the artifact kinds is an unattached Prisma documentation run, and no Markdown checklist can produce one. Left for a follow-up outside this issue, with the boundary now stated at the site so the next reader does not mistake the unselected-carrier arm for coverage of it.

**3 — the inserted guide sentence separated a phrase from its referent.** Accepted. "Both halves are needed" bound to the suppression clause instead of the two halves it names. Moved to the end of the paragraph.

### Confirmed correct by the review

That each added case fails under the mutation it exists for, with the multi-identity case the only failure in the package for its own; that both arms of the ledger case are individually load-bearing and neither is redundant with the earlier ledger test; that the mixed variable statement really materializes two selected property hosts and that the two arms together distinguish crediting both from crediting one; that the concatenated JSDoc parses in both passes, so the reviewed pass is not passing through a failure to parse, with the two fingerprints distinct and both asserted non-empty; that the extended assertions pin the full item list in deterministic order; and that the corrected comment about the removed guard is accurate.

samchon (commented):
## Individual Self-Review — 163fa1e66 `test(evidence): pin where a checklist exclusion's reach actually comes from`

**Verdict: FINDINGS.** Four items, all advisory, all merge-safe. Two applied in 8875e6ad9 along with two of the out-of-commit ones.

**1 — the documented repair is unavailable under the TypeScript claim default.** Accepted, and it is the sharpest kind of documentation defect: advice that reads as actionable and is not. The guide told an adopter to give a ledger a placeholder the claim's `symbol` does not select; a TypeScript claim taking the default selects all three kinds, so every eligible carrier is a selected host and no such placeholder exists. Reproduced before accepting, and the outcome is worse than a no-op: the ledger discharges its own placeholder while every other host still owes everything, and an interface placeholder contributes a member host that owes the checklist too, so the ledger file generates its own shortfall. Both constraints are now stated where the repair is named, and the TypeScript one is pinned by the arm that measures it rather than left as prose nobody ran.

**2 — the rule has no Markdown spelling.** Accepted. A Markdown claim's only eligible carrier is a selected host, so both paths dead-end. Said so in the guide and in the skill.

**3 — TSDoc parity.** Accepted. The in-editor surface is what an adopter actually reads, and it carried the aggregate and duplicate rules but not the carrier-reach rule. Added.

**4 — the skill's wording was wider than the code.** Accepted. It said "an acknowledgement whose carrier is not itself a selected host" where the fallback applies to exclusions alone. Positive evidence needs a selected host kind to be eligible at all, so it never reaches that branch, and the wider word invites a reader to build one that does.

### Confirmed correct by the review

That the added arm is the sole guard in the package for the misconception the guide sentence names, since a mutation deciding reach by file membership rather than carrier eligibility fails that arm alone; that its count guard is exact rather than incidentally satisfied; that both directions of the rule are now pinned, an unselected carrier in an ordinary file spreading and a selected carrier in a declared ledger file binding locally; that the recommended repair does work where it is available, with interface, `const`, and `type` placeholders all giving a claim-wide ledger and an `enum` correctly refused as an unsupported carrier; and that the guide paragraph still runs as one argument with zero em-dashes.

### Note

The reviewer saw a dirty working tree mid-review and correctly re-ran everything through overlays against the committed blobs. That was this session implementing the next commit while the review ran, which the solo campaign procedure expects. The tree is clean at this head and carries nothing unrelated.

samchon (commented):
## Individual Self-Review — 3f64fb3fb `test(evidence): make the exclusion-review case pin the placement it claimed`

**Verdict: FINDINGS.** Four items. The reviewer withheld merge on the first; it is corrected in 33fda7fca.

**1 — the comment written to stop an overclaim contained one of its own.** Accepted, and the reviewer was right to withhold merge: the commit's whole subject was comment accuracy. It said no Markdown checklist can produce a declaration with no semantic host. `checklist` constrains the *reference* kind, never the claim kind, so a Prisma claim pairs with a Markdown checklist reference and its unattached top-level `///` run is exactly that declaration. Confirmed through three layers, decode accepting the pairing, the Prisma loader emitting a carrier declaration with no semantic hosts, and the evaluator requiring the ledger fallback for that shape. Corrected to name the pairing and where a case for it belongs.

**2 — the placement claim had no negative twin.** Accepted. It asserted the carrier's review is accepted beside it and never that a review written on a host that exclusion discharges is refused. A target-only over-match in the ledger left the whole package green. The arm now writes the review on the discharged host and asserts the carrier's exclusion is still reported unreviewed; that mutation now fails.

**3 — both source-position fallbacks in the review ledger are untested, and the review-side one appears dead.** Correct, confirmed by mutation, and correctly marked pre-existing on master. Left as a follow-up outside this issue: the declaration-side branch wants a Prisma-ledger `requireReview` case, and the review-side branch should be justified or dropped rather than covered by a test written to match it.

**4 — the guide skipped the Prisma ledger shape.** Accepted. It enumerated TypeScript and Markdown and omitted the claim kind whose unattached run is the shape the carrier mechanism was built around.

### Confirmed correct by the review

That the added arm genuinely enters the carrier-eligible-but-unselected branch and catches a mutation nothing else in the package catches, namely binding the review lookup to the hosts the exclusion discharges; that no pre-existing test pairs `requireReview` with an exclusion carrier, since the shared review config selects a symbol kind that makes its interface a selected host; that the first arm is otherwise byte-identical and still probes both tags; that the shared fingerprint map is sound for the second arm, with the two items' values distinct and a swap producing a stale diagnostic; and that the moved guide sentence restores the paragraph to one argument with zero em-dashes.

samchon (commented):
## Individual Self-Review — 8875e6ad9 `docs(evidence): scope the ledger repair to the claims that can perform it`

**Verdict: FINDINGS, and the reviewer was right to withhold merge.** It found a real behavior defect, not a prose one. Fixed in d84485d14.

**1 and 2 — a Markdown heading that materializes no unit still hosted tags, and a checklist turned that into a silent full discharge.** Accepted after independent reproduction. A heading whose title yields no anchor reports a selectable symbol while naming an identity no unit was ever created for. Eligibility reads the symbol and admits the tag; per-host coverage reads the identity, finds no selected unit behind it, and takes the fallback meant for a carrier that hosts nothing at all. One `@evidenceExclude` under `## ---` therefore discharged every item for every host in the claim, across every file. Measured on a two-file plan set: two diagnostics became zero. The spelling is not contrived, since the slug keeps only letters, numbers, and underscores, so any punctuation-only title is anchorless.

The commit under review asserted a universal about Markdown claims that this falsifies, which is what the reviewer withheld merge over, and the reviewer was also right that the honest repair is the collector rather than the sentence. The scan now attributes such a heading's region to the nearest enclosing unit, which is where the same walk already put that region's content for digests. The tag is refused as an out-of-scope carrier and named where it sits, both real hosts owe their items again, and the positive twin moves with it: a citation there was credited to nobody and drew no diagnostic, and is now refused on the host it sits on.

One departure from the suggested repair, made after measuring it. Clearing the host for every heading that materializes no unit also moves an H5's, and the existing contract pins an H5 declaration as reporting `h5`. That symbol is not selectable, so an H5 tag is already refused as an out-of-scope host rather than admitted against a phantom, which means it has no defect to justify the change. The fix is confined to the anchorless case, and the reasoning is recorded at the site and in the skill.

This was latent rather than introduced by the branch: with coverage asked once per obligation, the phantom host discharged the obligation the way any acknowledgement does and nothing asked which host it sat on. Adding the host dimension is what made it reachable, so it belongs here.

**3 — the JSDoc stated only the TypeScript half.** Accepted and already corrected in 62696f3f5, which added the Markdown half.

**4 — the ledger case's step list no longer summarized it.** Accepted for the sibling case in eda20d959; this one is covered by the same pass.

### Confirmed correct by the review

That a TypeScript claim's default `symbol` yields all three kinds and leaves no unselected eligible carrier, probed across enum, merged namespace, destructured `const`, and index-signature shapes; that positive evidence never reaches the claim-wide fallback, guarded both by the tag check and by eligibility; that the reach follows carrier eligibility rather than file membership, reproduced across two claim files; that the added arm measures its constraint and is non-vacuous in both configurations; that the guide, skill, and JSDoc state the same rule with no contradiction; and that the branch's MDX carries no em-dashes.

samchon (commented):
## Individual Self-Review — d84485d14 `fix(evidence): stop an anchorless heading from hosting a checklist acknowledgement`

**Verdict: FINDINGS, merge correctly withheld.** The commit fixed a real defect and left the same class open one file away. Sealed in 5bf30c7d2.

**1 — the identical silent discharge was still open for an unaddressable path.** Accepted after independent reproduction. A path containing whitespace forms no target, so nothing in that file materializes, not even the file unit, while every line still reported `file` as its host kind. On the fixed head, one `@evidenceExclude` in `plans/alpha beta.md` took one diagnostic to zero and discharged the whole checklist for every host in the claim. The reviewer was right that scoping the repair to the anchorless branch was whack-a-mole against the rule the commit itself stated, and right that this shape is worse: the unaddressable-path diagnostic was surfaced for reference files alone, so a claim-only file offered no host, no diagnostic, and no clue.

The rule is now stated once and applied at both sites. A reported host kind and a reported host identity must stand behind the same unit; an unaddressable file reports no kind, so its tags are refused as unsupported hosts rather than admitted against an empty identity, and the whitespace problem reaches claims as well as references. The declaration's host set stays empty rather than holding an empty string, which is what selects the wording for a declaration nothing supports.

**2 — the prose asserted a refusal the code only sometimes performs.** Accepted, and this is the fourth prose-versus-code mismatch this cycle. Reattribution is not refusal: the tag is refused when the enclosing unit's kind is one the claim does not select and accepted on that unit when it is, and the accepted face is the common configuration. Restated conditionally in both the comment and the skill.

**3 — the test covered only the refusal face.** Accepted, and it follows from 2. Both faces are now pinned, along with the unaddressable-path case.

**4 — the change reaches ordinary references and said so nowhere.** Accepted. The skill's own convention is a migration note in both directions, as it carries for the variable-declarator and variable-digest moves. Written into 5bf30c7d2: a citation under an anchorless heading used to discharge its target through a host no unit stood behind, so such a claim can newly report missing coverage and an out-of-scope host, while review pairing widens so a duplicate can newly fire and an Unreviewed diagnostic can disappear.

**5 — only H5 was named where H6 behaves identically.** Accepted.

### Confirmed correct by the review

That the H5 carve-out is sound for both tag kinds, since the public symbol union caps at `h4` and Markdown declarations never set `ExclusionCarrier`, so eligibility reduces to the symbol test and refuses the tag before any identity is read; that fenced blocks, setext headings, and content before any heading produce no phantom; that digests, review fingerprints, `evidence/documented`, and `evidence/singular` are untouched, since only the host fields moved and the digest host is computed separately; and that the new test is load-bearing under neutralization of the fix.

Issue #1212, opened 2026-08-17
A Markdown claim file that materializes nothing leaves its population silently


samchon:
# A Markdown claim file that materializes nothing leaves its population silently

## Problem

`evidence/graph` files two problems while scanning a Markdown file: a path that cannot form an evidence target because it contains whitespace, and a heading with no resolvable anchor. Both are surfaced only when a **reference** population selects that file. A claim population selecting the same file gets neither.

A claim file is where the consequence is larger. A whitespace-named claim file materializes no unit at all, so it contributes no host and drops out of the obligation entirely, and nothing says so. The author sees a passing build and a file that is quietly outside the graph.

## Evidence

The gate asks references only:

```go
for _, inventoryProblem := range inventory.Problems {
  if selectedByMarkdownReference(config, base, relative, inventoryProblem.Symbol) {
    problems = append(problems, inventoryProblem.Message)
  }
}
```

```go
func selectedByMarkdownReference(
  config graphConfig,
  base populationBase,
  path string,
  symbol string,
) bool {
  for _, claim := range config.Claims {
    for _, reference := range claim.References {
      // ... only references are consulted
    }
  }
  return false
}
```

Reproduced with a Markdown claim over `plans/**` selecting `h2`:

| files | diagnostics |
| --- | --- |
| `plans/alpha beta.md` with `## Section one`, `plans/gamma.md` with `## Section two` | one, naming gamma only |

`Section one` owes nothing and is named by nothing. Pointing a reference at the same file reports `cannot form an evidence target because its path contains whitespace` as expected, which is the asymmetry.

The anchorless-heading problem behaves the same way: a claim file whose `## ---` yields no anchor gets no `has no resolvable anchor` line, so an author whose tag under it is refused as an out-of-scope host has no message explaining that the heading materialized nothing.

## Consequence surface

- Any Markdown claim whose files include a whitespace-named path: those files leave the host population with no diagnostic.
- Any Markdown claim containing an anchorless heading: the refusal an author sees names the host kind but not the cause.
- Sharper under a per-host obligation such as `checklist`, where a file leaving the population means an escaped obligation rather than one fewer citation site. That is how this was found.
- No effect on reference populations, which already report both.

## Approach

The gate is about whether any configured population reads this file for the symbol the problem was filed under. A claim is such a population. Ask claims on the same terms references are asked, matching the claim's own `symbol` set so a problem about a kind the claim does not read stays withheld.

[...]

Issue #1211, opened 2026-08-17
An H6 heading panics the Markdown scan


samchon:
# An H6 heading panics the Markdown scan

## Problem

Any scanned Markdown file containing an H6 heading crashes `evidence/graph` with an index-out-of-range panic. It does not matter whether the file is a claim or a reference, whether the heading carries a tag, or which `symbol` kinds are selected: the scan runs before any of that is consulted.

`######` is ordinary Markdown. A document deep enough to use one takes the whole rule down rather than reporting anything.

## Evidence

```go
scanProjectMarkdown("docs/spec.md", "###### Deep heading\n")
```

```text
panic: runtime error: index out of range [5] with length 5

github.com/samchon/ttsc/packages/evidence/native.scanMarkdownInventory
	packages/evidence/native/markdown.go:242
```

The cause is one unguarded walk. `headingUnitIDs` is declared as `[5]string`, so its valid indices are 0 through 4:

```go
headingUnitIDs := [5]string{}
```

The digest-host walk starts at `level - 1` with no upper clamp, so an H6 enters it at index 5:

```go
currentDigestHostID = fileUnitID
for ancestorLevel := level - 1; ancestorLevel >= 1; ancestorLevel-- {
  if headingUnitIDs[ancestorLevel] != "" {
    currentDigestHostID = headingUnitIDs[ancestorLevel]
    break
  }
}
```

The sibling walk that computes a unit's `parentID` performs the same indexing and is safe only because it sits inside an `if level <= 4` block. The digest walk sits outside it, which is the asymmetry.

An H5 is safe by one: it enters at index 4, the last valid slot.

## Consequence surface

- Every consumer of `evidence/graph` whose configured Markdown globs reach a file with an H6, on either the claim or the reference side.
- The crash precedes unit materialization, so it is not avoidable by narrowing `symbol`, by excluding the heading, or by any tag the author could write.
- `evidence/documented`, `evidence/singular`, `evidence/todo`, and `evidence/review` are unaffected; they do not read the Markdown scan.
- The panic surfaces through the Go plugin host rather than as a diagnostic, so the failure a consumer sees is a crashed rule rather than a message naming the file.

## Approach

The walk should stop at the deepest level the array holds. Clamp its start rather than widening the array: heading units exist only for H1 through H4, so index 4 is the deepest entry that can ever be non-empty, and a wider array would carry slots nothing writes.

[...]

PR #1213, opened 2026-08-17
fix(evidence): repair the Markdown scan's crash and its silent claim-side hole

samchon:
## Intent

This pull request owns the complete accepted cycle of the `evidence-markdown-scan` solo issue campaign. Its accepted issue set is exactly two issues, both defects in `packages/evidence/native/markdown.go`:

- #1211 — an H6 heading panics the Markdown scan
- #1212 — a Markdown claim file that materializes nothing leaves its population silently

Both predate the `checklist` work that surfaced them and neither is reachable through that option. They are independent of each other, so the DAG has no edges and either order is safe.

The branch opens implementation-free. Code, tests, and documentation follow as commits.

## Scope

**#1211** is a crash. `headingUnitIDs` holds five slots and the digest-host walk starts at `level - 1` with no upper clamp, so an H6 indexes past the end. Any scanned Markdown file containing `######` takes the rule down before it materializes anything, whatever the configuration.

**#1212** is a silent hole. The two problems the Markdown scan files, an unaddressable path and a heading with no resolvable anchor, are surfaced only when a reference population selects the file. A claim population selecting the same file gets neither, so a whitespace-named claim file contributes no host, drops out of the obligation, and says nothing.

Deliberately not in scope: the phantom-host asymmetry those two shapes expose, where a scan can report a selectable host kind against an identity no unit stands behind. It is a latent oddity on `master` with no reachable consequence, and #1210 already established that reaching for it drags the shared scanner into work nobody asked for.

## Verification

Pending. Recorded as Individual Self-Review reviews per commit and one Overall Self-Review round over the settled head, alongside ordinary pull-request CI.


samchon (commented):
## Self-Review — `ccd9ca24c` and `d0851b2b1`

The read-only subagent lane failed three times with `API Error: 529 Overloaded`, twice for the first commit and once for the second. That is infrastructure rather than a reason to skip the gate, so both reviews were performed by the main agent. Recording the substitution because it matters for how much independence to credit these two reviews: they are a self-check by the author, not a second reader.

### `ccd9ca24c` — the clamp

**Correct at every level, verified by walking all six.** Level 1 starts at index 0 against a `>= 1` condition, so it never iterates and the file unit stands, which is right. Levels 2 through 4 are unchanged. Level 5 started at index 4 before and starts at 4 now. Level 6 started at 5 and panicked; it now starts at 4, which is the correct nearest candidate because H5 never writes a slot. The clamp therefore changes the ancestor found at no level, which is what makes it a crash fix rather than a behavior change.

**Findings, both applied in 54984f315.**

The class was one instance short of sealed. `headingUnitIDs` is indexed at five sites: the descendant reset and the unit write are inside `level <= 4`, the digest walk is now clamped, and the `parentID` walk indexes from the heading's own level with no clamp of its own, in bounds only because the `level <= 4` above it makes that level materializable. That is the same read the crash came from, one enclosing condition away from being live, and nothing said so. The dependency is now named at the site.

The case asserted the tag refusal for both an H5 and an H6 but compared digest attribution only for the H6, while the commit claimed an H6 answers exactly as an H5 already did. The H5 arm is now compared too.

**Checked and correct.** `min` is a builtin from Go 1.21 and the module targets 1.26; this is its first use in the package, which is a style note rather than a risk. Nothing outside the walk start changed, so digests, review fingerprints, and unit materialization are untouched for every level that previously worked. The regression case fails with the clamp removed, and it fails with the panic rather than with an assertion, which is the honest signal.

### `d0851b2b1` — the claim-side report

**Scoping verified against the loader's two passes.** `loadMarkdownInventories` runs twice: once with `claimPopulationConfig`, which nils out `References` and is called before `activeGraphConfig`, and once with the activated config. The claim branch is what makes a claim's scan problems report at all, and it reports them from the **declared** configuration, which is the rule the evidence skill states for governance: a claim that later deactivates still declared its files. A file read by both populations appends the same message twice and `reportProblems` sorts and drops adjacent duplicates, so it surfaces once; the case asserts that count rather than assuming it.

**No over-report.** The branch requires `claim.Type == artifactMarkdown`, so a Prisma or TypeScript claim whose globs happen to match a Markdown path does not pull it in. The claim is matched on its own `Base.Absolute`, so a claim with a different `root` is judged against its own base rather than the reference's.

**Withholding is preserved.** A problem filed under a heading kind the claim does not select stays withheld, which the case pins with an H1-only claim against an H2 problem. The wildcard symbol reaches both populations, which is what carries the unaddressable path, since that problem is about the file rather than about any heading kind.

**Mutation note worth recording.** The first mutation attempt substituted `if claim.Type == artifactMarkdown &&` globally and hit three sites, two of them the file-matching predicates, which broke matching wholesale and surfaced as four unrelated failures. The mutation was invalid, not the code. Targeted at the one line, exactly one test fails: the new one.

### What these reviews did not cover

Neither is an independent read. A second reader would most plausibly find what a self-check misses: prose asserting more than the code does, which is the defect class the previous cycle hit four times. The `evidence defenses` CI lane exercises both commits end to end through the real binary, so the runtime claims have a check the reviews do not supply.

samchon (commented):
## Overall Self-Review — complete round on the branch

Run while CI executes, over the whole `origin/master...HEAD` diff rather than the latest delta.

### Findings, all applied

Six, every one a case that read as coverage while pinning less than it claimed.

- A step list still framed the boundary as the source side after the name and the reason above it had stopped doing so.
- The same off-by-one about `headingUnitIDs` was written in two places and corrected in one. Believing the count and shrinking the array to four slots crashes on the first heading of any materializable level.
- A constant named `scanProblemReference` held a whole graph configuration.
- The withheld arm asserted only that a diagnostic is absent, so a configuration producing nothing at all would have satisfied it. It now anchors on the coverage diagnostic that proves the graph ran.
- The H5 arm built its fixture by removing the H6 and never checked the removal took, so a fixture change would have left it asserting the H6 case twice while reading as coverage of both.
- The deep-heading arm asserted two substrings and counted nothing, which also left the citation on the selected host above it unpinned.

One prose item: the region-opening note named an H5 as its example of a heading that materializes no unit. An H6 is one too, and until this branch it could not be reached because it crashed the scan first.

### Verified

Every index into `headingUnitIDs`, by hand: the descendant reset writes `level..4` inside `level <= 4`; the digest walk is clamped; the `parentID` walk is clamped and also enclosed; the unit write indexes `level` inside the same condition. No unguarded read remains.

Mutation sweep over all three production changes. Removing the digest clamp fails one test; removing the claim branch fails one test; removing the `parentID` clamp fails none, which is the expected answer and the one the comment already states, since that walk's enclosing `level <= 4` makes the clamp an identity there.

That sweep is worth recording for a second reason. Its first run reported that none of the three mutations was caught, which was false: the substitutions were addressed by line number, the numbers had shifted under earlier commits, and `sed` reports success when it matches nothing. The tool was passing vacuously in exactly the way the six findings above describe. It was rewritten to address the sites by pattern and to fail when a substitution changes nothing.

### Shape

Ten lines of production code: two clamps, the predicate widening, and one string literal written as an escape.

Issue #1223, opened 2026-08-18
fix(unplugin): an unavailable watcher disables the persistent cache instead of falling back to complete-snapshot validation


samchon:
## Problem

When the generation-scoped directory watchers cannot be registered — or when one of them reports an error later — the persistent transform cache stops validating **at all** instead of falling back to the documented complete-snapshot proof. The cached generation can never be reused, so every module delivery evicts it and re-runs a whole-project compile: `N` modules cost `N` full project transforms.

Both READMEs promise the opposite:

> Envelopes without a graph, an unavailable watcher, or an unstable/incomplete generation retain complete-snapshot validation.
>
> — `packages/unplugin/README.md:315`, `website/src/content/docs/setup/unplugin.mdx:242`

This is a regression introduced with the narrow-validation work released in `@ttsc/unplugin@0.27.0` (#1184). Before it, persistent validation was complete-snapshot only and required no watcher.

### Watcher health is folded into the generation's own completeness flag

`transformProject` (`packages/unplugin/src/core/transform.ts:2757-2766`):

```ts
let stableProjectSnapshot =
  before.complete &&
  inputSnapshot.complete &&
  sameHashes(before.hashes, inputSnapshot.hashes) &&
  sameHashes(before.fileSignatures, inputSnapshot.fileSignatures) &&
  sameProjectDirectories(before.projectDirectories, inputSnapshot.projectDirectories) &&
  tracker?.failed !== true &&
  tracker?.membershipChanged !== true &&
  hostInputTracker?.failed !== true &&
  hostInputTracker?.membershipChanged !== true;
```

`cached.projectSnapshotComplete = stableProjectSnapshot`, and the trackers are attached only when it holds.

### Both validation paths then refuse the entry

`matchesCachedSource` (`:1337`) selects the narrow branch only when `projectSnapshotComplete === true` **and** both trackers are attached, otherwise it calls `matchesCompleteInputSnapshot` — whose first statement is:

```ts
if (cached.projectSnapshotComplete !== true) {
  return false;                                     // :1741
}
```

So a failed `fs.watch` registration produces an entry that neither path can validate. `matchesCachedSource` returns `false` for every delivery, `evictGeneration` drops it, and `transformProject` runs again — once per module.

[...]

Issue #1222, opened 2026-08-18
fix(unplugin): persistent validation re-hashes every derived input on every module delivery


samchon:
## Problem

Persistent-mode transform-cache validation (Vite `serve`/Vitest, the Turbopack loader, Metro workers — every host that has no build boundary) re-reads and re-hashes each delivered module's whole derived input set on **every** delivery, including the inputs it already proved unchanged for the previous module of the same generation microseconds earlier.

Sibling modules of one project share almost all of that set, and `graph.globals` is shared by *every* module. In a real project the globals alone are the entire `@types/node` declaration set, so each cache hit pays tens of milliseconds of `readFile` + SHA-256 before it can return output the generation already contains.

Reported downstream as #1219 (`vite serve` cache hits at 190–250 ms each on a 14-file schemas project). That report measured `@ttsc/unplugin@0.23.0`, whose mechanism (whole project walk + whole out-of-walk union per hit) was already replaced by #1184 in 0.27.0. The mechanism below is what `0.28.0` actually does, and it reproduces the same symptom class for the same workload.

### The code path

`matchesCachedSource` → `matchesNarrowPersistentInputs` (`packages/unplugin/src/core/transform.ts:1374`) ends in:

```ts
const inputs = selectWatchInputs({ file, projectRoot, result, temporaryTsconfig });
for (const input of inputs) {
  if (hostValidation.identities.has(derivationIdentity(state, input))) continue;
  if (!matchesRecordedInput(cached, input)) return false;
}
```

`matchesRecordedInput` always terminates in `graphInputStateHash` / `hostInputStateHash`, i.e. `filesystem.readFile(input)` plus a SHA-256 of the full contents. The skip above covers only the *universal host inputs* that `captureUniversalHostInputValidation` put in the manifest; graph members and globals are never in it.

`selectWatchInputs` yields `reach(graph.edges, file) ∪ graph.globals ∪ graph.configs ∪ candidates ∪ dependencies[file]` (narrowed to `dependencies[file] ∪ configs` for a `dependenciesComplete` file). The list itself is memoized per envelope; the **content validation of it is not memoized at all**.

### Why the globals half is large in every real project

[...]

Issue #1219, opened 2026-08-18
`@ttsc/unplugin`: vite serve re-hashes every project and out-of-walk input on each cache hit (O(N) per module, no build-scoped shortcut)


TheHaff:
Again thanks so much for these awesome projects! I found something again while digging into the performance issues I reported [here](https://github.com/samchon/typia/issues/2330)

Under `vite serve`, `matchesCachedSource` performs full validation on **every** cache hit — the project walk plus `collectExternalInputHashes` over every out-of-walk program input. The build-scoped shortcut added in #980 / `@ttsc/unplugin@0.21.0` is explicitly not authorized for serve:

```js
buildStart() {
  if (viteCommand === 'serve') resetTtscTransformCache(transformCache)
  else beginTtscTransformBuild(transformCache)
}
```

I understand why — Vite's single `buildStart` spans later HMR edits, so a permanent build-scoped shortcut could serve stale output, and #252's release note lists that as an intended invariant. This is a request for a _third_ mode rather than a claim that the guard is wrong.

The cost scales with the program's out-of-walk input count, not with project size, so it does not shrink by scoping the tsconfig project.

### Measurement

A schemas-only project of 14 `.ts` files with its own tsconfig, so the in-walk project walk is ~15 files and negligible; the delta is out-of-walk hashing. Timed by requesting each of 11 gated modules from the dev server:

| `@ttsc/unplugin` | typia       | 11 modules total | first (compile) | each subsequent cache hit |
| ---------------- | ----------- | ---------------- | --------------- | ------------------------- |
| 0.17.3           | 13.0.0-rc.2 | **1336 ms**      | 967 ms          | 4–23 ms                   |
| 0.23.0           | 13.2.0      | **4545 ms**      | 1416 ms         | 190–250 ms                |

_Caveat:_ the typia versions differ because 13.1.19+ peer-require `ttsc >=0.19.2`, so these rows are not a clean single-variable comparison. The per-cache-hit component (4–23 ms → 190–250 ms) is dominated by hashing rather than codegen, but I can re-run with any pairing you prefer.

Because the compiler is spawned synchronously, this blocks the dev server's event loop, so concurrent module requests queue behind it.

### Suggested direction

[...]

PR #1224, opened 2026-08-18
fix(unplugin): bound persistent cache validation by proven generation state

samchon:
Claims the complete accepted cycle of the `@ttsc/unplugin` persistent transform-cache validation campaign opened from #1219.

## Scope

This pull request owns both cycle issues and is the only implementation unit for them:

- #1222 — persistent-mode validation re-reads and re-hashes every derived input on every module delivery, including the shared closure and `graph.globals` it proved unchanged for the previous module of the same generation.
- #1223 — when the generation's directory watchers cannot be registered (or fail later), the persistent cache stops validating entirely instead of falling back to complete-snapshot validation as both READMEs document, so every module delivery re-compiles the whole project.

Implementation order follows the issue DAG: #1222 first, because #1223 restores `matchesCompleteInputSnapshot` as a live fallback and that path must stop re-hashing the whole project per delivery before it starts being used.

## Discovery gate

Two complete full-scope discovery rounds ran against `f0b6f9ac2` (`@ttsc/unplugin@0.28.0`) over the declared scope: the persistent (non-build-scoped) validation path in `packages/unplugin/src/core/transform.ts`, the Vite serve wiring, the Turbopack and Metro persistent hosts, the `tests/test-unplugin` cache corpus, the `experimental/unplugin-perf` gates, and the cache sections of both READMEs. Round 1 accepted the two issues above; round 2 produced no further candidate.

The reporter's suggested Vite-`watchChange` "invalidation epoch" was adjudicated and not adopted: Vite's dev watcher ignores `**/node_modules/**`, which is exactly where the out-of-walk graph members and global-scope declarations live, so an epoch derived from it would go stale for the inputs that dominate the cost. A generation-scoped content proof achieves the same shape without a host-specific signal, and covers Vitest, Turbopack, Metro and Bun as well.

## Verification

Pending. Verification status, CI results, and every Self-Review round are recorded as pull-request reviews on this thread rather than by rewriting this body.


packages/unplugin/src/core/transform.ts
```diff
@@ -1762,10 +1909,52 @@ function matchesCompleteInputSnapshot(
...
+ * a restored file) costs a re-read on every later delivery for the rest of the
+ * generation's life, because the recorded signature can never match again. The
+ * narrow path self-heals through {@link matchesProvenInput}; this is the same
+ * refresh for the path that proves the whole snapshot at once.
+ *
+ * The delivered file is the single exclusion: its recorded hash is the source
+ * the bundler supplied, so the disk bytes this walk read for it were compared
+ * against nothing.
+ */
+function adoptProvenSignatures(
```
samchon, line 2076: Added for the second accepted finding. The complete-snapshot path proved every recorded input unchanged and then discarded the signatures, so a touched-but-identical file was reread on every delivery for the rest of the generation's life. Adoption happens only after the comparison that justifies it, and the delivered file is excluded because its recorded hash is the bundler's buffer rather than the disk bytes this walk read.

tests/test-unplugin/src/internal/transform-project-cache.ts
```diff
@@ -778,6 +788,335 @@ async function assertPersistentValidationUsesPerFileInputs(): Promise<void> {
...
+    "a changed metadata signature must fall back to the content comparison",
+  );
+  reads = 0;
+  assert.ok(await deliver(modules[1]!));
+  // Exactly one read is available to this delivery: the generation's own
+  // current file, whose recorded hash came from the bundler's buffer and so
+  // carries no disk signature. A second read means the touched global was not
+  // re-proven and is being reread on every delivery.
+  assert.ok(
+    reads <= 1,
```
samchon: Was `<= 4`, which passed with or without the refresh branch and therefore pinned nothing. One read is available to this delivery — the generation's own current file, which carries no disk signature — so a second read means the touched global is being reread per delivery.

samchon (commented):
## Individual Self-Review — `6f09ba529` "prove each generation's shared inputs once"

Resolves #1222. One read-only reviewer over that commit's parent-to-commit diff; every candidate adjudicated below by the author.

**What landed.** A generation-scoped content proof: each recorded input carries the nanosecond metadata signature captured around the read that proved its hash, and a later delivery reuses that proof while the signature holds. Applied to the narrow per-file path and to the complete-snapshot path's project walk and out-of-walk re-hash. New fixture support for a non-empty `graph.globals`, and a case pinning bounded per-module reads over a shared closure plus five invalidation twins.

### Accepted and fixed in `9fb9b582e`

| Finding | Disposition |
| --- | --- |
| The touch twin's bound of four reads passed whether or not the revalidated input was re-proven, so it pinned nothing (`transform-project-cache.ts`) | Accepted. Tightened to the one read that delivery can legitimately make, with the derivation stated in place. |
| `collectProjectInputSnapshot`'s reuse branch wrote `fileSignatures` that nothing consumed — and behind it, `cached.inputSignatures` was never refreshed on the complete path, so a metadata-only change made an input unmatched for the generation's whole life | Accepted, and it is the same class this cycle exists to seal. `matchesCompleteInputSnapshot` now adopts the signatures captured while it proved the snapshot, after the comparison that justifies them. |
| `measureServeValidation`'s doc claimed every module owns a disjoint input, false since it became the shared gate for the partitioned and shared-closure scenarios | Accepted, rewritten. |
| `graphGlobals` silently produces no graph at all without a positive `graphFanout`, in both generators | Accepted, stated in both option docs. Latent only — both current callers pass one. |

### Adjudicated without a change

**Missing negative twins (symlink retarget, deleted input, superseding candidate, recorded-`missing`, racing write).** Partially rejected on evidence. The retarget property is already pinned, and more strictly than a new case would: `test_transformttsc_invalidates_project_cache_through_a_linked_graph_edge` retargets a symlink/junction between two targets holding **identical bytes**, so only the physical-identity half of the proof can detect it, and that is exactly the half the signature shortcut skips a read for. The remaining twins were traced and confirmed handled (missing, unreadable and broken-link inputs never acquire a signature; a broken link's 7-field `missing-target` signature cannot equal the 12-field one, so target creation stays visible; a racing write is never recorded, by the before/read/after sandwich).

**The two-spelling alias case is untested.** Accepted as a gap, rejected as a defect: with either keying, a spelling whose signature does not match falls back to the full content comparison, so the identity keying fixed in `2d4fceb38` cost the optimization, never freshness. A test would pin an efficiency property, not a behaviour.

**Windows ACL-only readability changes, and same-size rewrites inside one clock tick, are invisible to the signature.** Real, and deliberately outside this cycle. Metadata-signature equality standing for content equality is the acceptance `matchesUniversalHostInputs` has made since `@ttsc/unplugin@0.27.0`; this cycle widens the surface it covers rather than introducing it. Recorded as a candidate for the next discovery cycle, where the question — whether a wall-clock guard should gate the shortcut for inputs written within one tick of their capture — can be adjudicated for the whole manifest at once instead of half of it.

### Verified clean by the reviewer

Sandwich ordering at all three capture sites; `inputSignatureSlot` mirroring `matchesRecordedInput`'s external-over-project preference with no hash/signature cross-pairing and no BOM-codec divergence; the manifest never written before the recorded hash was matched; `delete inputSnapshot.fileSignatures[currentFileKey]` excluding only the bundler-buffer key; `collectProjectInputHashes` / `collectExternalInputHashes` unreachable from the proven path, so `@ttsc/metro`'s cache key still folds real content hashes; membership detection unaffected; no concurrency hazard; no `strict` / `noUncheckedIndexedAccess` violation; both Go fixture edits valid; the new case genuinely failing on pre-change code.

samchon (commented):
## Overall Self-Review — rounds 1 and 2

Solo, whole base-to-head diff. Round 1 ran over `a3234869c`; it applied two corrections, which landed as `2d4fceb38`, and round 2 is the fresh complete round over that head.

### Round 1 — two defects found in this cycle's own diff, both fixed in `2d4fceb38`

**1. The out-of-walk signature manifest was keyed by physical identity.** A symlink or junction spelling and its selected target deliberately share one identity, but they do not share metadata, so the two overwrote each other's signature in `captureExternalInputSnapshot` and both fell back to a full content read on every delivery — the exact cost this cycle exists to remove. `TtscHostInputValidation` already states that rule for the universal manifest ("Do not key this manifest by physical identity"), and it was not carried over. Signatures are now keyed by lexical spelling; the recorded hash stays identity-keyed, because aliases of one physical file do share its content. Correctness was never at risk (a mismatching signature always falls back to the full comparison), but the optimization silently disappeared for every aliased input.

**2. The new shared-closure serve scenario inherited a budget it cannot honestly meet.** A missing resolution candidate cannot be proven absent by metadata, so each one reachable from the delivered file costs one failed `stat` per delivery, and a shared closure makes that set large by construction — roughly `N/2` per module for the harness's chain fixture. Rather than loosening the shared membership budget for every scenario or quietly picking a number, `measureServeValidation` now takes a per-scenario `statsBudget`, and the shared-closure scenario states the derivation in place. The residual probe cost itself is recorded as a candidate for the next discovery cycle rather than folded into this one; the same directory-listing grouping `matchesUniversalHostInputs` already uses for its own missing probes would collapse it.

### Round 2 — complete, clean

Verified at `2d4fceb38`, whole surface:

- **Freshness of the signature shortcut.** Symlink/junction retarget (the signature carries both the link's and the target's `dev/ino/mode/size/mtimeNs/ctimeNs`, so the skipped realpath comparison cannot be evaded); A→B→A edits (metadata moves, so the content comparison still runs); a write racing the capture or the revalidation (never recorded, by the before/read/after sandwich, matching `captureUniversalHostInputValidation`); `missing`, broken-link, directory-kind and EACCES inputs (no signature recorded, full comparison retained); the generation's own current file (excluded from `inputSignatures` because its recorded hash came from the bundler's buffer, not from disk).
- **Key kinds.** Every `externalInputSignatures` read and write uses the resolved spelling; every `externalInputHashes` / `externalInputRealpaths` read stays identity-keyed; project signatures share the project-key universe of `inputHashes` by construction, so a signature can never exist without its hash.
- **`projectSnapshotComplete` semantics.** All four consumers re-checked: the build-scoped first-delivery shortcut (now available to a provable generation with no watchers, which is sound — the build boundary, not the watcher, is what authorizes it), the complete-snapshot gate, the narrow gate, and the existing assertions in `transform-project-cache.ts` / `transform-external.ts`, which expect `true` and still get it.
- **"Cannot prove" vs "stale".** `matchesUniversalHostInputs` still returns `false` — not a fallback — when a proving directory cannot be listed. That is deliberate and pinned by an existing case: the complete-snapshot path compares a recorded `missing` marker against a recomputed one and cannot certify absence either, so eviction remains the only sound answer there.
- **Tracker lifecycle.** A generation that is provable but has unusable notifications attaches neither tracker and closes both; `retainTracker` follows the same condition; the Windows broker path is unchanged when no `watch` operation is supplied.
- **The added directory comparison** in the complete-snapshot path cannot invalidate spuriously on file writes: measured on NTFS, `writeFileSync` and `appendFileSync` leave the parent directory's `dev/ino/mode/size/mtime/ctime` untouched while `create` moves it, which is exactly the membership signal the watchers report.
- **Cost accounting.** The per-delivery `lstat` added by the proof is not counted by the perf harness's counters, and stays inside the existing 60-per-module bound of `assertPersistentValidationUsesPerFileInputs`, whose derived set is 100 universal host inputs (skipped by identity) plus two proven inputs.
- **Exported API.** `collectProjectInputHashes` and `collectExternalInputHashes` keep hashing unconditionally; only the internal validation paths take the `proven` manifest, so `@ttsc/metro`'s cache key still folds real content hashes.
- **Regression value.** Recomputed by hand: the shared-closure case reads ~55 files per delivered module before the fix and ~0.9 after (only the generation's current file, which has no disk signature), so the new bound of 4 fails on the pre-change code rather than passing vacuously.
- Documentation in `packages/unplugin/README.md` and `website/src/content/docs/setup/unplugin.mdx` restated for both behaviours, and the Go fixture / harness additions read as valid Go.

Nothing else survived verification **at that head**. Round 2's clean result does not stand: the Individual Self-Review of `6f09ba529` then returned findings that were accepted and fixed in `9fb9b582e`, so under the fresh-start rule a new complete round is now required over that head, and it is running. This record is the ledger for rounds 1 and 2, not a merge gate.

samchon (commented):
## Individual Self-Review — `f440f5717` "keep a generation provable when its watchers are not" and `2d4fceb38` "key proven out-of-walk signatures by spelling"

Resolves #1223 (`f440f5717`) and corrects this cycle's own diff (`2d4fceb38`). One read-only reviewer per commit; both adjudicated here because their accepted findings landed in the same correction, `4fb68b5ac`.

### Accepted and fixed in `4fb68b5ac`

**The restored fallback proved less than the narrow path it stands in for — a freshness regression this cycle introduced.** With notifications unavailable, `matchesCompleteInputSnapshot` compared content but never the *physical identity* of universal descriptor and config inputs: only the graph half of the out-of-walk snapshot records realpaths, and a plugin `hostInput` is not a graph member. Retargeting a symlinked host input to a byte-identical file therefore replayed the old output, where before this cycle the same state recompiled. The fallback now proves the same universal manifest the narrow path does, and requires the recorded directory snapshot instead of treating its absence as "no proof needed". Pinned by a new case that fails the watchers through the seam and retargets a link between two byte-identical selections whose own transitive requires differ.

**The out-of-walk rehash deduplicated by physical identity, so every spelling after the first was never read at all.** Two spellings share an identity exactly when they selected one file at generation time — which is the state a retarget ends — so the alias was skipped and its retarget to different bytes passed. Deduplication is by spelling now, and a disagreement between spellings that shared an identity writes a marker no content hash can match. This gap predates the cycle; the fallback becoming reachable is what made it live.

**One refused tracker is now pinned.** Membership has two halves and the narrow path needs both proven; the other two cases refuse or fail every watcher at once, so a regression to a single-tracker check would have kept them green.

**The harness budget was a bare constant** encoding the scenario's module count and fanout, with the wrong arithmetic in its comment (`selectReachableSources` includes the delivered file, so the average is `(count+1)/2`). Derived from the scenario's own size now, and the README bullet moved with the gate.

**Nothing pinned the per-spelling proof.** A revert to identity keying passed the entire suite: the existing link-retarget cases use a *directory* junction whose leaf resolves to one inode, so both keyings are indistinguishable there. The shared-closure case now reports one global under two spellings, which costs two reads per delivery under identity keying and fails its bound. POSIX only — a Windows file symlink needs elevation — following the corpus's existing precedent for that constraint.

### Adjudicated without a change

**A dead `hostValidation === undefined` branch in the narrow path.** Correct: unreachable while `projectSnapshotComplete === true`. Kept as the conservative answer for a state the type system does not exclude, with its contract reworded — `undefined` means "this narrow proof is unavailable", not specifically "notifications failed".

**The Windows broker path is exercised by no new test.** Real gap, deliberately not closed here: the broker child is process-global and shared by every generation in the worker, so a case that kills it to observe `fail()` would corrupt sibling cases in the same process. Its failure classification feeds the same `tracker.failed` flag the seam cases drive. Recorded as a discovery candidate.

**`hostValidation.identities` is identity-keyed while its `entries` are spelling-keyed** (`e6b81db96`, predates this cycle), so a derived watch input that is an alias of a universal host input's physical target is skipped by the per-module loop without being validated under its own spelling. Same class as the alias gap fixed here, but a different owner and a frozen accepted set: recorded as a discovery candidate rather than folded in.

**Two enumerated twins remain uncovered under unavailable watchers** (an edited universal host input; a created superseding candidate). Both traced and confirmed to invalidate. The host-input class is now covered more strongly than the enumerated twin would have been, by the retarget case above.

**The `?? Object.keys(externalInputHashes)` fallback feeds identity strings into a spelling-keyed lookup**, making the signature reuse dead on that path. Unreachable in-tree — `transformProject` always assigns `externalInputPaths` — and safe where it is reachable, since the hash is still recomputed. Left as legacy tolerance.

**Formatting.** `f440f5717` was pushed before `pnpm format`; `a3234869c` reflowed it, and that commit is purely cosmetic. The campaign formats its cycle once, but the per-commit habit is the better one and is what the later commits follow.

### Verified clean by the reviewers

Tracker lifecycle (`notifying` gates attach and `retainTracker` together, so "attached but closed" is unreachable; no double close; no broker child leaked or prematurely killed); `filename` normalization and registration-failure classification identical between the seam and the default path; membership freshness of the added directory comparison, including that an `unstable:` signature cannot coexist with a complete walk and that the scratch directory is provably outside the project; the build-scoped shortcut becoming reachable without watchers being sound, since a real build boundary — not a watcher — is what authorizes it; `reportsMembershipChange` outranking `notificationsProveMembership`, so a change reported by a failing tracker still evicts; both doc surfaces updated identically with no other page describing the old behaviour; no remaining site treating "cannot prove" as "stale"; key-kind coherence across every read and write of the spelling-keyed manifest; and no `strict` / `noUncheckedIndexedAccess` violation in either commit.

samchon (commented):
## Individual Self-Review — `9fb9b582e` "refresh proven signatures on the complete path"

One read-only reviewer over that commit's parent-to-commit diff. It reported **no surviving correctness defect** in the adoption logic and traced the pairing end to end; every finding was about coverage and comment accuracy.

### Accepted and fixed in `5dabc9e71`

**The entire production change was unobservable.** Every read-counting case runs with a working watch seam, so all of them take the narrow path and none reach `matchesCompleteInputSnapshot`; the three cases that do reach it assert only compile counts. Deleting the adoption left the suite green. The new case refuses every watch registration, measures the steady per-delivery read count, touches one project input and one out-of-walk input, and asserts the next delivery falls back to the content comparison while the one after it returns to steady.

**The `currentKey` exclusion — the one line carrying the correctness weight — had no negative twin.** Removing it broke nothing, yet it is what stops a bundler-supplied hash from acquiring a disk signature. The same case now drives it directly: a file whose disk content moved ahead of the supplied source is delivered (the bundler's source stays authoritative for the file it delivers), and a sibling delivery must still read that file and see the drift. With the exclusion removed, the sibling reuses a hash the disk no longer carries and the drift is invisible.

**A graph member that exists but cannot be read still recorded a signature.** `captureExternalInputSnapshot`'s graph branch recorded one whenever the current state matched the compiler's proof, including `null === null` — a path that exists but whose read throws (EACCES, or a non-regular file). Its recorded hash is `"missing"`, so metadata was standing in for content that was never read. The dependency branch two lines below already had the guard; the graph branch has it now.

**The harness comment this cycle rewrote was itself inaccurate.** The non-partitioned fixture chains its modules rather than giving them one shared closure; only the externals and the globals are common to every module. Corrected in both the doc comment and the scenario banner.

### Adjudicated without a change

**"A restored file" is unreachable as an example on the project half** — a delete-and-recreate moves the parent directory's signature, so `sameProjectDirectories` rejects the generation before adoption. Correct; the example was dropped, leaving the two that hold.

**`matchesUniversalHostInputs` refreshes `entry.signature` with the pre-read stat rather than a sandwiched one.** Rejected as a defect: the pre-read signature is the conservative one. A write between the stat and the read is caught by the read itself (the content comparison fails and nothing is recorded); a write after the read leaves a signature that cannot match the next check, so it forces a revalidation rather than skipping one.

### Verified clean by the reviewer

Adoption's pairing is total — `sameHashes` compares equal-length key sets, so every adopted key had its hash compared, and the generation's *own* original current-file key is genuinely proven during a sibling delivery, which is why adopting it is correct and usefully removes the permanent one-read floor. Ordering is safe: adoption is the last statement, nothing after it can return false, and it has no internal early exit, so no half-adopted manifest is possible. The external sandwich excludes missing paths, racing writes, and unreadable paths; the reuse branch correctly does not re-record. The `proven` maps alias the cache entry's own maps but the walk only reads them and writes into fresh objects. `adoptProvenSignatures` and `matchesProvenInput` write the same maps under one invariant, so neither can misread the other. No `strict` / `noUncheckedIndexedAccess` violation, and the added `lstat` cost is bounded and invisible to the perf harness.

samchon (commented):
## Individual Self-Review — `4fb68b5ac` and the `75e19ebe6..5dabc9e71` range

Two read-only reviewers. Both reported real findings; the adjudication is below and the accepted ones landed in `eb416da47`, `dab7325bd`, `180f83c36` and `ada93bed0`.

### Accepted, with the strongest finding first

**The fallback fix imported an inability to prove into the one path that exists not to have one.** `matchesUniversalHostInputs` rejects a generation when a missing-probe directory exists but cannot be listed, which is correct for the narrow path (it has no stronger proof to fall back to) and wrong for the whole-snapshot path, where an EACCES probe directory in a sandboxed container or a restrictive `node_modules` would mean a permanent recompile on every delivery. That is exactly the cliff #1223 exists to remove, reintroduced by #1223's own fix. The manifest is split: the fallback proves only the half whose every rejection is evidence of a change (a vanished path, a moved physical target, a strict blocker, differing content), while the probe half stays on the narrow path, and the fallback keeps re-comparing the recorded `missing` markers as it always did.

**A guard added for unreadable graph members was unobservable.** No fixture produces a member the compiler recorded without a content hash that the host can stat but not read, so deleting the guard left the suite green — the same failure mode the commit that added it had just indicted in its predecessor. The fixture can now stamp exactly that state and the read seam refuses that one path; while unreadable the member matches its recorded state and the cache hits, and allowing reads again without touching a byte of metadata must replace the generation.

**The spelling deduplication in the out-of-walk rehash was dead code.** `selectExternalInputPaths` already yields spelling-unique entries, and the identity fallback yields canonical paths, so the `Set` never fired. What actually changed was the removal of the identity skip; the comment says that now.

**Both new guards in the fallback are unreachable, and the commit message overstated one.** `projectDirectories` and the universal manifest are always present when `projectSnapshotComplete` is true. Kept for typing and as the conservative answer to a state the types do not exclude, with the claim corrected here rather than in an immutable message.

**Harness and fixture accuracy.** The membership stat budget is one named constant instead of two literals that could drift; `aliasedGlobal` states both preconditions it depends on; the one-failed-tracker comment no longer claims an invariant its recompile does not maintain; and the unnotified-retarget case asserts its generation is provable before failing the watchers, so it cannot pass by recompiling for the wrong reason. The export list ordering was restored.

**Documentation voice.** The cache paragraphs this cycle rewrote used em-dashes, which the documentation skill forbids for prose. Replaced with parentheses. The harness README's scenario bullet keeps the em-dash its four siblings use, since a parallel list item should read like its peers.

### Adjudicated without a change in this cycle

**A dependency-only out-of-walk input has no physical-identity proof at all.** `captureExternalInputSnapshot` records realpaths only for graph members, so a path appearing solely in a plugin's reported `dependencies` accepts a byte-identical retarget on both validation paths. Real, and pre-existing on both paths rather than something this cycle widened; the field's own doc frames the omission deliberately, and correcting it changes what a dependency-only snapshot claims. Recorded as a discovery candidate so it gets its own evidence.

**The universal manifest can hand a signature to content it never read** — a host input that exists, cannot be read, and whose compiler-reported hash is also absent. Same class, same manifest as the racy-clean gap below, and pre-existing. Both belong to one adjudication rather than two half-fixes.

**The Windows broker's failure classification is exercised by no test.** The broker child is process-global and shared by every generation in the worker, so a case that kills it to observe the classification would corrupt sibling cases in the same process. Recorded as a candidate; a dedicated worker or a broker-scoped seam would make it testable.

### Found by measurement rather than by a reviewer

While the reviews ran, the acceptance this cycle relies on was measured directly: **156 of 200 back-to-back same-size rewrites are invisible to a metadata signature on Windows**, where Node also reports `ctime` as the creation time. This cycle extended the shortcut from a handful of descriptor files to every derived input, so it widened that hole. An input whose modification time is not strictly older than the moment its signature was captured now records no signature and keeps its content comparison, which is the rule Git applies to a racily clean file, and the gate takes the later of a symlink's two modification times so a retarget cannot slip through it either. Applying that guard inside the project walk then broke the generation gate — a file saved moments before a build is racy in the pre-compile walk and settled in the post-compile one, so the two signature maps disagreed and the generation could never validate, which is the dev server's normal case — so the walk now returns what it observed and what it proved as separate maps.

samchon (commented):
## Individual Self-Review — `eb416da47` and `3088e2b81` (the settle guard)

Two read-only reviewers over the commits that introduced and then corrected the racily-clean guard. Both found real defects; the accepted ones landed in `dab7325bd`, `180f83c36`, `3088e2b81` and `02548a4f8`.

### Accepted

**The guard measured against the wrong clock.** `Date.now()` is finer than the clock a filesystem stamps mtimes with: a write advances an mtime once per tick, about 15.6 ms on Windows and a whole second or two on filesystems that store seconds. A cutoff of "now" therefore cleared an input written earlier in the current tick, which is precisely the input a second write in that tick can change without moving its signature, so what shipped closed the first millisecond of a tick-sized window. The measurement that motivated the guard (156 of 200 back-to-back same-size rewrites) lives inside that first millisecond, which is why it looked closed: the witness, not the class. The cutoff now sits two seconds back, past every filesystem timestamp granularity in use. The one case it cannot answer, a filesystem whose clock runs behind the process clock, is stated where the cutoff is computed, because answering it needs a timestamp minted by the observed filesystem and the operations contract is read-only.

**The guard then made freshly modified inputs permanently unprovable**, because the manifest was consulted only where it already held a signature and nothing else revisits an input. The lookup now returns the owning manifest for any input the generation recorded a hash for, so a content comparison that succeeds records the proof as soon as the timestamp ages.

**That self-healing lookup then handed a proof to content nothing read.** A recorded `missing` state is the one comparison that succeeds without reading: an unreadable path still reports `missing`, so the record branch fired for it, and metadata can hold still while the bytes behind such a path appear — a widened ACL, a released share-deny lock. It would have failed the case added one commit earlier for exactly this state. The lookup now carries the recorded state and refuses to prove that one, and the marker is named instead of repeated as a literal in eight places.

**The guard broke the generation gate.** `transformProject` compares the pre- and post-compile walks' signature maps, key set included; a file saved moments before a build is unprovable in the first walk and provable in the second, so the maps disagreed and the generation could never validate. That is the dev server's normal case, since the save is what triggered the build. The walk now returns what it observed and what it proved as separate maps.

**A symlink's retarget half was ungated** — the cutoff compared only the target's modification time — and the gate now takes the later of the two.

**Two bounds went slack** once every input including the generation's own current file became provable; both are exact counts now, which restores their ability to fail. Four stale comments, a documentation clause on both cache pages, a harness constant's position, a fixture escape and a dedup comment were corrected alongside.

**Both fixtures backdate the tree they generate.** A project written microseconds before the build that reads it is the one shape where nothing can be proven, and it is not the shape real projects have. The cases that edit a file mid-run still exercise the unproven state deliberately, and one of them pins both halves of the cutoff.

### Adjudicated without a change

**Stamp the settle-margin case at `now - 1500` instead of `now` for timing robustness.** Rejected on the arithmetic: an input stamped `now` stays unprovable for the full margin, while `now - 1500` leaves only 500 ms before it settles. `now` is the more robust of the two, so the case keeps it.

**The harness has no scenario measuring the margin's cost on a freshly written tree.** Fair, and recorded as a candidate rather than answered with a gate whose number could not be derived honestly. Self-healing bounds that cost to the first two seconds of a run.

**The universal descriptor manifest is now the only ungated capture, and it covers the files an editor rewrites in place.** Correct, and it stays deferred: its capture predates this cycle, the fix is the same rule applied to a differently shaped entry, and one adjudication for that manifest as a whole is worth more than a seventh correction here. It is recorded with this observation attached.

samchon (commented):
## Individual Self-Review — the constraint pass (`748655675` through `f9aab6e1b`)

Four read-only reviewers over the commits that answered the owner's constraint for this pull request: no hardcoding, no monkey-patching, no over-optimization, everything universally sound. Their findings drove the shape below.

### What the constraint removed

A settle guard had been added against a measured hole: a metadata signature can stand for content a same-tick, same-length rewrite already replaced. Review established that its cutoff measured the **wrong clock** — the process clock is finer than the stamp a filesystem writes, so "not older than now" refuses only the sub-millisecond slice of a tick-sized window, and the 156-of-200 measurement that motivated it lives inside exactly that slice. Backing the cutoff off by a constant margin was the only way to make it effective, and that margin is a guess about filesystems the adapter never observes, which then required backdating both fixtures' generated trees for the repository's own cases to hold.

All of it is out: the constant, the helpers, the fixture backdating, and the two cases that depended on them. The question is published as #1227 with the measurement and the design constraint that decides it — a sound version needs a timestamp minted by the observed filesystem, and `TtscTransformFilesystemOperations` is read-only, so this contract cannot produce one.

### What the constraint kept, and generalized

One rule survives because it needs no constant: **a signature may stand in for a content comparison only for a state a read produced.** An input the host could see but not read records a missing marker on both sides, so its comparison succeeds without reading anything while its metadata holds still, and bytes appearing later would never be compared. That rule now governs all seven recording sites, the universal descriptor manifest included, where an entry that cannot be proven keeps its content comparison and re-earns the signature when it can.

Two more findings were folded in on the same reading. The universal manifest skipped derived inputs by **physical identity** while keying its own entries by spelling, so an alias of a proven target was skipped without ever being validated under its own name; it skips by spelling now. And the out-of-walk re-check no longer folds hashes into a shared key and marks a disagreement with a `divergent:` sentinel — it compares each spelling against the recorded state directly.

### Two compile breaks, caught by review rather than by CI

`MISSING_INPUT_STATE` was deleted along with the settle helpers it sat beside while seven call sites still used it, and a new fixture option was passed without being declared on its interface. Neither the package nor the suite compiled. Both restored.

### A case that could not reach the state it named

The new universal-manifest case wrote a readable file and had the descriptor claim no hash for it. ttsc revalidates a declared host-input hash against its own filesystem and drops one it can disprove, so the manifest was refused outright and every delivery recompiled — for a reason the case never intended. It is a link with no target now, the one shape the host's filesystem and the adapter's fail to read alike, and its content appears through the cache-owned read alone so no metadata moves.

A reviewer then probed the platform claim that case rested on and disproved it: a file symlink does need elevation on Windows, a junction with no target does not, and it reports the same state. The whole-platform skip is gone. The descriptor also derives its two proofs now instead of declaring literal nulls, which is both the neighbouring fixture's convention and the thing whose absence made the case unreachable a commit earlier.

### Corrections the constraint prompted elsewhere

The fallback had been made to treat a missing universal manifest as "nothing to prove" while the narrow path treats the same state as the absence of one. That inverted the disposition in the last-resort path, and the only thing producing that state was a case that swaps a generation's result object — the guard read as shaped by the assertion. The manifest is recorded state of a generation, like the input hashes beside it, so it moved onto the cache entry: the swapping case carries it, and the fallback refuses a generation that has none.

Also: the manifest's signature refresh now sandwiches its read like every other capture site, so the recording rule both cache pages state is literally true of that path; two restored-timestamp touches use an unmistakable stamp rather than an offset sized for the removed cutoff; and a comment justifying `readable = false` on the current-file branch states the actual reason, that the recorded state is the bundler's rather than the disk's.

### Adjudicated without a change

An advisory noted the missing-resolution-candidate probe cost, the dependency-only inputs' absent identity proof, and the Windows broker's untested failure classification. All three are pre-existing, none was widened here, and each is recorded with its evidence for the next cycle's discovery rather than folded into a frozen set.

samchon (commented):
## Individual Self-Review — `495709fe7`, `d4bcea397`, `bf71e4d71`

Three read-only reviewers over the fixture commits that closed this cycle. The sequence converged: one confirmed finding, then none, then none.

### Accepted

**The derived descriptor was an incomplete mirror of the neighbour it claimed to follow.** ttsc's `hashHostInput` classifies a directory with a marker hash; the fixture's `observedHash` had only read-or-null. The option points at a link today, so both observe `null` and nothing changes — but a directory would have made ttsc's revalidation drop the declared hash and refuse the manifest, which is the exact failure the derivation exists to prevent and which had already cost this case once. The branch is there now.

**A dead require survived the cleanup that removed its sibling.** `crypto` was moved into the block that uses it while `fs` stayed unconditional, so the class the commit set out to remove lived on in twenty other descriptors. Both are conditional now, ordered like the hand-written descriptor earlier in the same file.

### Adjudicated without a change

**The disproved platform premise 1500 lines above.** A reviewer carried this cycle's finding to `assertPersistentValidationUsesPerFileInputs`, which skips its broken-target edge on Windows citing elevation. Probed rather than assumed: a junction with no target does reproduce a dangling link without elevation, but a junction whose target is later created as a **file** still reports `ENOENT` through the link on Windows 11, and that appearance is exactly what the case asserts. The skip stays and its stated reason is now the one that holds. A later review noted the same transition *is* reachable there with a **directory** target — recorded as a candidate rather than done, because `tests/test-unplugin` runs only on `ubuntu-latest`, so the change would alter a passing case that the gate cannot check.

**The commit title "run on every platform" buys no CI coverage**, for the same reason. True, and worth stating plainly: the case is now runnable everywhere and verified by probe on Windows, but only Ubuntu exercises it in CI.

### Verified rather than asserted

Reviewers probed, on this machine, rather than reasoning from documentation: the dangling-junction state (four filesystem calls, matched against a POSIX dangling symlink), the file-target non-appearance, the directory-target appearance, the generated descriptor's escaping (``, no raw NUL byte), and the fixture's `observedHash` against both ttsc's `hashHostInput` and the adapter's `hostInputStateHash` across a regular file, a directory, a dangling junction and a junction to a directory — identical in all four. The final commit's two emitted variants were extracted and evaluated: the default is byte-identical to its parent, and the option variant differs only by the reorder it claims.

samchon (commented):
## Overall Self-Review — final round, clean

One reviewer, whole base-to-head diff at `bf71e4d71`, nothing changed since. Earlier rounds are recorded above; this is the round that finds nothing.

### Verified this round

**Both issues' acceptance holds in the code, not only in the tests.** #1222: a delivery returns without reading whenever the recorded signature still holds, and every path that records one is gated on an unracing read of a state a read produced. #1223: the narrow path answers `false` only on evidence — a reported membership event, a changed universal input, a changed derived input — and answers "cannot prove" by returning `undefined` so the caller falls back to the whole-snapshot path instead of discarding the generation.

**The forbidden-pattern sweep is empty for production code.** Zero numeric literals added under `packages/unplugin/src` and `packages/metro/src`; the three string constants added are `"error"` and `"rename"` (the `fs.watch` event names) and `"missing"`, which existed seven times in `master` before this branch and is now named once instead of repeated; zero consumer, fixture, environment-variable or filename special cases; zero new platform branches — the two `win32` conditions touched are the pre-existing broker registrations, which now defer to an embedder-supplied watch operation.

**Every rule that survives is a logical one, not a measured one.** A signature may stand for content only for a state a read produced. Losing a proof is not evidence of a change. An alias is a different input from its target. None of them depends on a filesystem, a platform, or a bundler. The one rule that did need a measured constant — refusing a signature the filesystem clock cannot separate from a later write — was removed with its fixture backdating and published as #1227, because a sound version needs a timestamp minted by the observed filesystem and this contract is read-only.

**Coverage.** Eight new cases, each with its negative twin: a shared closure with an aliased spelling proven per spelling; the whole-snapshot path's proof adoption and its one exclusion; unavailable, failed, and asymmetric notifications; a byte-identical retarget under failed watchers; and unreadable content on both the graph and the universal manifest. Bounds are exact counts where the count is deterministic, so a regression that costs one read per delivery fails them.

**Documentation.** Both cache sections state the recording rules the code enforces, what the fallback proves, and what a lost notification means, and no page describes a mechanism this branch removed.

### Gate

Seventeen checks pass on this exact head, `nestia` excepted by the maintainer for this merge. Thirteen Individual Self-Reviews cover every pushed commit, the last three returning one finding, then none, then none; each is adjudicated in the reviews above. Three pre-existing findings are recorded with their evidence for the next cycle's discovery rather than folded into a frozen set.

Issue #1230, opened 2026-08-18
A documentation `{@link}` is a reference the checker resolves, and the graph holds no edge for it


samchon:
## Problem

`@ttsc/graph`'s claim is that "every node and edge is resolved by the TypeScript
compiler itself, so the graph is exact for TypeScript and TSX, never
text-guessed". A `{@link Symbol}` in a documentation comment is such a
reference: the checker resolves the name, binds it, and counts it as a use. The
graph holds no edge for it, so a class of checker-resolved relationships is
simply missing from a graph that claims to hold them all.

The gap is worst where citation conventions put it. A declaration citing a type
through `@evidence {@link ISale}` is related to `ISale` in the way the compiler
itself understands, and the graph says the two are strangers. Worse, the guide
for that convention recommends `import type` for a citation-only import, and a
type-only import consumed by nothing else contributes no edge either — so a
citation's only trace in the module is the one form the graph is structurally
blind to.

### The checker resolves the link — proved by a diagnostic

Two files identical but for the link, compiled with `noUnusedLocals`:

```ts
// src/linked.ts
import type { ISale } from "./sale";
/** @evidence {@link ISale} Mirrored here. */
export function withLink(): string { return "x"; }

// src/unlinked.ts
import type { ISale } from "./sale";
/** No link at all here. */
export function withoutLink(): string { return "y"; }
```

`ttscgraph dump` at `cd890c61f` reports:

```
diagnostics:
   src/unlinked.ts:1 TS6196 'ISale' is declared but never used.
edges:
   src/linked.ts#src/linked.ts:module -- exports -> src/linked.ts#withLink:function
   src/sale.ts#src/sale.ts:module -- exports -> src/sale.ts#ISale:interface
   src/unlinked.ts#src/unlinked.ts:module -- exports -> src/unlinked.ts#withoutLink:function
```

`unlinked.ts` raises TS6196 and `linked.ts` does not. The only difference is the
`{@link ISale}`, so the checker demonstrably resolved that name and counted it a
use — and in the same output, nothing connects `withLink` to `ISale`.

### Root cause

`collectTypeRefs` walks with `Node.ForEachChild`
(`packages/ttsc/internal/graph/edges.go:765`), which does not descend into a
declaration's documentation. A probe over the same fixture confirms the shape of
what is skipped:

- `Node.JSDoc(file)` returns the documentation node for a `.ts` declaration —
  TypeScript-Go parses it on TypeScript sources, not only JavaScript ones.
[...]

Issue #1229, opened 2026-08-18
A declaration's documentation tags never reach the index, so "which code implements this specification" costs a repository-wide search


samchon:
## Problem

`@ttsc/graph` exists to remove the repository-wide search. A citation question is
the shape that search is worst at, and it is the one question the graph cannot
answer at all today.

Documentation conventions attach a declaration to something outside the type
system — a specification section, an API operation, a data model, a reference
document. `@ttsc/evidence` does this with `@evidence`; this repository's own
sources do it 767 times with `@reference`:

```ts
/**
 * Renders the stacking notice.
 *
 * @evidence docs/discount.md#coupon-stacking States the per-issuer limit.
 * @evidence POST:/orders/{orderId}/coupons Explains the rejection.
 * @evidence {@link ISale} The complete sale contract is mirrored here.
 * @evidenceExclude prisma:Coupon.issuedAt Not surfaced in this notice.
 */
export function renderNotice(): string { ... }
```

None of that text reaches any graph result — not the dump, not `details`, not
`tour`, not `lookup`. An agent asking "which code implements
`docs/discount.md#coupon-stacking`?" gets `escape` and falls back to `grep`,
which is exactly the cost this server removes for every relation it does hold.

The asymmetry is what makes this an index defect rather than a nice-to-have. The
forward question — *what does this declaration cite* — costs one file read,
because the answer sits directly above a declaration the agent has already
found. The reverse question — *which declarations cite this specification* —
costs a search across the whole repository, because the citations are scattered
through every file that implements the document. An index earns its keep on the
reverse question, and today it cannot answer it at all.

### The dump carries nothing

Built from `packages/ttsc/cmd/ttscgraph` at `cd890c61f`, run over a probe
project whose `src/notice.ts` is the declaration quoted above:

```
$ ttscgraph dump --cwd probe --tsconfig tsconfig.json --pretty
nodes: src/notice.ts#renderNotice:function [function]
       src/notice.ts#src/notice.ts:module [module]
       src/sale.ts#ISale.price:variable [variable]
       src/sale.ts#ISale:interface [interface]
       src/sale.ts#src/sale.ts:module [module]
edges: src/notice.ts#src/notice.ts:module -- exports -> src/notice.ts#renderNotice:function
       src/sale.ts#src/sale.ts:module -- exports -> src/sale.ts#ISale:interface
```

That is the complete output. Four tags, four targets, four reasons, and no trace
of any of them.

[...]

PR #1231, opened 2026-08-18
feat(graph): index documentation tags and their resolved links

samchon:
Wires documentation-tag facts into the compiler knowledge graph, so an agent can
answer "which code implements this specification" from the index instead of a
repository-wide search.

## Cycle scope

This pull request owns the complete accepted cycle. Verification is pending —
this body is the claim written before any code exists.

- #1229 — a declaration's documentation tags never reach the index. Collects
  unrecognized JSDoc tags as raw declaration facts, gates them behind a dump
  capability, and answers the reverse question through `lookup` and `details`.
- #1230 — a documentation `{@link}` is a checker-resolved reference the graph
  holds no edge for. Adds the documentation-reference edge.

#1230 depends on #1229 and both move one dump schema version together, so they
are one integrated unit rather than two pull requests.

## Design commitments

- **The compiler host names no convention.** The collected population is the
  tags TypeScript itself does not recognize, so the boundary belongs to the
  compiler. `@evidence`, this repository's own 767 `@reference` tags, and any
  convention a consumer invents are supported identically, and the graph gains
  no dependency on `@ttsc/evidence`.
- **Index, not detail.** Tag text rides `details` and `lookup` only; `tour` and
  `trace` carry none. Artifact nodes for Markdown sections, Prisma models, and
  Swagger operations were considered and rejected in discovery — they describe
  document structure the agent can already read, while putting documents into
  the build universe.
- **No judgement.** The graph never says a tag's text resolves, is covered, or
  is true.

## Verification

Pending. Local evidence and the ordinary CI lanes will be recorded as pull
request reviews as the cycle proceeds.


samchon (commented):
## Individual Self-Review — `f55b750d9`

Read-only review of the parent-to-commit diff for the commit that indexes a
declaration's documentation tags and answers the reverse citation question
(closes #1229). Seven candidates were raised; every one is adjudicated below,
and the accepted repairs land in the following commit rather than here, so this
review and the diff it describes stay in step.

### Accepted and repaired

**1. The reverse index keyed on the first word of any unknown tag's prose, and a
citation hit outranked every name match.** Accepted, and the most serious of the
set. TypeScript recognizes no `@todo`, `@remarks`, `@example`, or `@default`
either, so `@todo Add caching here` made its declaration a carrier of the address
`Add`, and a question opening with that word answered with it — first, and
labelled a certain citation. `@default 4` indexed under `4`. Issue #1229 called
the leading token a *ranking* rule and the implementation made it an
unconditional trump, so the issue's own design was the standard the code failed.

Repaired by indexing an address rather than a token: a candidate qualifies only
when it carries an interior separator (`/`, `#`, `:`, an inner `.`) or is an
inline link. Every address form in use qualifies — `docs/pricing.md#sale`,
`POST:/orders`, `prisma:Sale`, a URL — and an English word or a bare number does
not. `test_ttscgraph_lookup_indexes_addresses_and_not_prose` pins both
directions, including that a prose word still ranks the symbol actually named
after it.

**2. `entrypoints` leaked an undeclared field.** Accepted. It copies `lookup`
hits wholesale, so `docTags` rode a result whose schema does not declare one.
The copy now drops it.

**3. A citation query with no ASCII alphanumerics never reached the index.**
Accepted. The `terms.length === 0` guard fired before the citation pass, so a
target in another script, or one made only of punctuation, was answered
"no searchable terms" while the index held that exact address. The citation pass
now runs first and the guard yields only when both are empty.

**4. Elision could cut the address, and the matched-tag filter read the elided
text.** Accepted. A long URL is one token and can exceed the 200-character
budget by itself; the hit then came back without the tag that explained it. The
cap now never cuts the first token, and the filter reads the node's own tags.

**5. `graphdump` under-declared the capability.** Accepted. It runs the shipped
declaration walk and its dump carries every tag, but declared only
`sourceDigests` — and under the capability contract an absent claim means the
producer never looked, so a consumer would read a tagged declaration as citing
nothing. It now declares `docTags`.

**6. Two stale "schema-v6" comments.** Accepted; both corrected. No occurrence
of the string remains in the repository.

**7. Coverage gaps against the issue's acceptance list.** Accepted. Added: CRLF
and LF producing identical tags; a merged identity documented on two
declarations; every binding of a multi-binding statement carrying its
statement's block; a link with trailing text inside its braces; a bare number
and a prose word not indexing; a URL indexing.

### Accepted with a different disposition

**Uncapped tag count per node, and a file-header block attributed to the first
declaration.** The observation is correct. The count stays uncapped, because
tags are identity in the sense `details` already uses for members and literals,
and capping identity is what sends a reader back to the file. The header case is
largely answered by repair 1 — `@license MIT` and `@packageDocumentation` carry
no address and now index nothing — and what remains is TypeScript's own
attribution, which `doc` has always had too. Recorded rather than changed.

### Found sound

The dedup key, the `BuildFiles` incremental path, the variable-statement climb,
all four schema gates, the compiler-host boundary, and the index-with-spans
contract were each inspected and found correct; the reviewer specifically hunted
for an incremental-rebuild defect and traced that none exists.

A separate crash was found by the tests rather than by this review and is fixed
in the same follow-up: reading a tag's comment through its per-kind struct
panicked on the first `@param`. The walk now descends with `ForEachChild`, which
also makes a `{@link}` under `@see` an edge, as issue #1230 requires.

samchon (commented):
## Individual Self-Review — `d7de2301f` and `4fc9e7970`

Two commits reviewed in sequence: the documentation-link edge that closes #1230,
and the tag-only refresh test beside it. Findings from both, adjudicated; the
repairs land in `4a4e4bf18`.

### The one that mattered

**The edge covered functions and variables and missed the types.** The edge pass
found documentation by walking containers, and that walk never visits a class,
an interface, or a namespace as a node of its own — it descends straight into
their members. So a link on a type's own documentation resolved to nothing while
the tag beside it was indexed, and the two halves a reader composes disagreed
exactly where a type is documented. The same walk hands a property's subtree to
its class as well, so a class appeared to name a symbol only its member's
documentation mentions.

This is the whack-a-mole shape the development skill forbids: the witness was
fixed and the cause was not. The cause is two passes sharing a convention rather
than a set, so the node pass now records every declaration it found
documentation on and the edge pass resolves links from exactly that set. A
namespace stays absent from both halves, because the graph models it as no node
at all, and the test pins that they are absent *together* — one of them
answering alone is the defect.

### Also repaired

- **A query the name tokenizer cannot read let every node through.** Reaching
  the citation index before the tokenizer's veto was right, but the name scoring
  then awarded its whole-query bonus to everything, because "every term matched"
  is trivially true of no terms. An address in a non-Latin script returned the
  citation hit followed by arbitrary central symbols, presented as answers.
- **An unclosed `{@link` indexed the address `{@link`**, which every link in the
  project shares.
- **The cap and the index disagreed about where an address ends.** A braced
  group holds spaces, so measuring the protected prefix to the first whitespace
  returned `{@link` and cut the rest away — a hit found by an address came back
  without it. Both now ask one function. Text that was not cut no longer carries
  an ellipsis claiming it was.
- **`doc_ref` rendered unstyled in both viewers.** An earlier note in the
  campaign ledger claimed the viewer's colour map never matched at all; that was
  wrong — the reducer folds wire kinds onto exactly those display families, so
  the existing colours work and only an unmapped kind falls through. This
  cycle's own consequence table forbids that, so the kind is now mapped,
  coloured, and labelled in the bundled viewer and the website mirror.
- **The refresh test discarded the server's exit code** where both sibling
  refresh tests assert it. A refresh that answers correctly and then crashes the
  native child on shutdown would have passed.

### Coverage added

`{@linkplain}`, several links in one block and a repeated one, a link under
`@see` (a tag TypeScript *does* know, which the previous walk crashed on),
`@param`/`@returns` contributing nothing, class and interface and namespace
attribution, member-not-class attribution, and a pure-unit test over the address
rule and the cap together.

### Found sound

The recursion and its dedup, sharding and the incremental path, the external
boundary, the traversal policy under every focus, and the schema gates were each
inspected and found correct. The reviewer noted that documentation references
now count toward non-structural degree in lookup centrality and overview, and
that the default trace focus is `all` — both real behaviour shifts, both
deliberate, neither a defect.

### Accepted as observations

A few false positives survive the address rule by design: `@since v1.2` and
`@default 3.14` carry an interior dot and index, `@example foo.bar()` likewise.
Each is harmless — the query that finds them is the text the author wrote — and
tightening further would start refusing real addresses. A hyphen-only key
(`ABC-123`) is not indexed; that is the trade the rule makes, and the tag itself
is still returned in full by `details`.

samchon (commented):
## Individual Self-Review — `4a4e4bf18`

Read-only review of the commit that moved link resolution onto the node pass's
declaration set. The fix was confirmed to seal the class it claims — every
declaration form the node pass records is now a documentation host, and the
forms still invisible are invisible to both halves together. Four candidates
were raised beside that; all four are repaired in `670ce822a` and `c4a8f74ac`.

### Accepted and repaired

**A measured quadratic.** The edge pass asks once per file and the hosts were
one flat slice, so finding a file's documented declarations walked every
documented declaration in the project. The reviewer measured it rather than
asserting it: at VS Code's shape — 4,700 files against 70,500 hosts — roughly
1.6 seconds of pure filtering per build, against half a millisecond for a map
keyed by file, and invisible on any fixture or on this repository, which is
exactly why nothing would have caught it. The cold index time is this product's
headline claim, so this one mattered. Hosts are keyed by file now.

**An assertion that could not fail.** The namespace case matched node ids by
suffix against `#Documented2:`, and every id ends in its kind, so no id can end
that way: it passed against an implementation that records the namespace and
cites from it, which is the defect its own comment says the pairing exists to
catch. It now asks directly whether the graph holds such a node and whether any
edge leaves one.

**A third copy of the display-kind map.** The website reducer names the
benchmark reducer as the file it is a port of, and benchmark dumps carry
documentation edges now, so a kind missing there renders as a raw unknown
outside the family counts the benchmark page shows. The bundled viewer's legend
is hand-written rather than derived from its colour map, so it needed the entry
too, and the viewer guide described three families where there are four.

**A field written and never read.** `DocTag.File` was set on every tag and read
by nothing; a tag rides its target node, which already carries the file.

### Found sound

Pointer safety and partial builds were both traced rather than assumed: the two
passes iterate the same Program's source files inside one build, so the AST a
host holds and the file the edge pass hands it are the same object, and both
passes filter by the same selection. The dedup key embeds the file through the
node id, so two declarations cannot collide. The empty-terms repair removes
nothing reachable. The elision was checked against every boundary — an over-long
braced address with interior spaces, an address-only text, an unclosed brace,
and the guard that keeps short text away from it entirely.

### Accepted as an observation

`graphsymbols/provider.go` retains a whole Graph for the process lifetime, so
the recorded hosts pin their ASTs there — but the pre-existing resolution memo
pins the same ASTs already, so this is no new class of retention. Clearing the
build-only fields when a build finishes would improve both and belongs to
neither of this cycle's issues.

samchon (commented):
## Individual Self-Review — `670ce822a` and `c4a8f74ac`

Read-only review of the host-lookup change and the mirror repairs. **No defects
found**, and three of the checks were run rather than reasoned:

- **Determinism was measured.** The reviewer built the binary from this commit
  and dumped a fixture producing nine documentation edges from five hosts three
  times: 13,616 bytes, one SHA-256, all three identical. The map migration was
  the one place order could have leaked into output — the map is never ranged
  over, only keyed, and the per-file slice keeps source order, so the span a
  deduped edge keeps is the same one it kept before.
- **The repaired assertion now fires for the right reason.** A namespace node
  would be `…#Documented2:<kind>` for any kind the defect picked, and the infix
  matches all of them, while the member `…#Documented2.value:variable` does not
  match — no false positive, and the member keeps its own separate check.
- **The reduce copies are byte-equivalent** across all three files, and a
  repository-wide search found no fourth. The bundled viewer's legend is the
  only hand-written one and its four swatches match the colour map exactly; the
  website's legends iterate the map and needed nothing.

Also confirmed: the early return is exactly equivalent to the loop that would
not have run, the removed field had no consumer, and the restored test teardown
is character-for-character its two closest siblings.

One observation carried forward rather than changed: the benchmark's console
breakdown prints three edge families, which is how it already treats
`value-access`, `member-relation`, and `exports` — pre-existing, and outside
either issue in this cycle.

samchon (commented):
## Individual Self-Review — `1d190b0e7` and `9a5b0fc21`

Read-only review of the symmetric guard and the same-name resolution test. The
reviewer verified its findings by mutation rather than by argument, which is what
made the most important one visible.

### The finding that mattered

**One rival was not enough to prove the edge is a checker fact.** A text matcher
has to break a tie somehow — lowest id, highest id, first or last in program
order — and with a single rival half of those choices land on the right module by
accident. The reviewer built the matcher and ran it: it failed the test when it
kept the lowest id, and **passed** when it kept the highest, because `sale` sorts
above `other`.

The repair flanks the real module with two rivals, one sorting before it and one
after, each imported so the program holds all three and the real one sits between
them in import order as well. Both mutants fail now; the real implementation
passes. Re-verified here by installing each mutant in turn.

**Two weakenings travelled with the first rival.** Every existing assertion in
that file named only `#ISale:interface`, which two modules then answered to — so
nine of them passed against a matcher pointing every link at the wrong module,
which the reviewer demonstrated rather than supposed. Each assertion now names
the module it must resolve to. And the fixture's extra declaration created an
edge nothing asserted; both rivals are now reached for a name only they declare,
which is also what keeps their imports honest.

The test's own three-step plan did not mention the claim the file had come to
rest on. It has a fourth step now.

### On the guard

The reviewer traced the producer chain to its source — `parseJSDocComment` is the
only constructor, it yields `KindJSDoc` or nil, and both call sites drop nils —
and concluded the new guard is unreachable today. It also confirmed the reason
for adding it: `AsJSDoc` is an unchecked type assertion, so the failure being
guarded is a panic inside the compiler host, and the walk beside it carries the
identical guard on the identical value. Two consumers of one invariant disagreeing
about whether to check it is the drift this cycle already paid for once.

### Nothing else

Suffix collisions were checked in both directions by arithmetic and by execution;
no assertion in the file is vacuous; the failure messages are accurate, and the
one that read correctly at only one of its two call sites is generalized.

samchon (commented):
## Individual Self-Review — `cd28db9aa`, `1d7b69566`, `9753ecd33`

Read-only review of the coverage commits. The reviewer ran mutants rather than
arguing from the code, and that is what surfaced the one finding that mattered.

### The capability test proved nothing it was named for

It handed a capability list to the marshaller and asserted the same list came
back — a round trip through its own input. The shipped declarations in
`cmd/ttscgraph` and `cmd/graphdump` were untouched by it, so **deleting the
member from the production list broke no test in the repository**, and that is
precisely the regression the test's own doc comment claims to guard: a dump that
carries tags and forgets to say so is read as a repository where nothing cites
anything, because "cites nothing" and "never looked" are the same absent field.

The claim is now asserted where it is made, by running the shipped `dump`
command and reading its own output. Verified by deleting the member from
`fullSnapshotCapabilities`: the new test fails and names it. The unit test keeps
the half it can honestly own.

Two related weaknesses in the same file, both accepted: the "absent, not empty"
message described a distinction a decoded slice cannot see (the raw document is
read for that now, matching the key with its colon because the capability shares
its name), and the untagged comparison was by node and edge **count**, which
survives a changed id, name, or span. It compares the documents structurally
now.

### Confirmed by the reviewer's own mutants

The two-rival fixture closes every tie-break a text matcher could use: lowest id,
highest id, first and last in program order, and Go map order over five runs —
all fail, where the single-rival fixture passed on the highest-id strategy. The
`entrypoints` test is not vacuous: reverting the field-dropping copy makes it
fail. Its diagnostic was poor, though — the tool validates its own output and
rejects an undeclared property before a result exists, so the failure named the
absence rather than the field. The helper carries the payload out now.

### Found clean

`{@link Promise}` resolves to exactly one node despite being declared across
several lib files, because a merged symbol keeps one declaration identity, and
the assertion matches on the declared name rather than the lib path — so it is
robust across TypeScript versions. No assertion in the link test is vacuous and
no suffix matches two nodes. Blank-line debris left by an earlier removal is
cleaned.

samchon (commented):
## Overall Self-Review — round 1 over `origin/master...68d00515c`

A fresh solo pass over the complete base-to-head diff: 53 files, the native
producer, the wire contract, the resident model, three MCP runners, both
viewers, the release smoke script, seventeen test files, and three guides. Run
alongside CI rather than after it, since the two are independent gates.

### Findings, all repaired in this round's commits

**A field written and never read.** `DocTag.File` was set on every collected tag
and consumed by nothing — a tag rides its target node, which already carries the
file. Removed, with the parameter that fed it.

**Two comments that no longer described the code.** After the address rule
replaced the leading-token one, `citing` still promised to match any leading
token, and the citation score still claimed to outrank the name scores when it
takes no part in the ordering at all. A third, on `docTagsOf`, restated its
helper's contract inaccurately.

**A guard on one side of a pair and not the other.** The link walk converted a
node to a documentation block without checking its kind, where the tag walk
beside it checks. Nothing reaches it with another kind today, but the conversion
panics rather than returning nil — which is exactly how reading a tag's comment
through its per-kind struct crashed on `@param` earlier in this cycle — and two
consumers of one invariant disagreeing about whether to check it is the drift
this cycle already paid for once.

**The release gate carried its own copy of the schema version.** Bumping the
real one turned the last check before publication red with `dump schema is 7,
want 6`, which reads as a broken binary rather than a stale constant. It reads
`DUMP_SCHEMA_VERSION` from the module that declares it now. This was the
**second** copy of that number this cycle — the test helper had one too — and a
search across the repository finds no third.

### Verified by execution, not by reading

Every claim the guides now make was run:

- `lookup` with a documentation target as the query answers with the citing
  declaration — holds.
- `details` returns a declaration's tags — holds, all four.
- A trace reaches a documentation link under the full focus and under neither
  narrowed one — holds, all three focuses checked.
- `doc-ref` is one of the viewer's display families after reduction — holds.

Also verified this round: the serve protocol's shard path carries both halves
and declares the capability (`docTags` 1, `doc_ref` 1, schema 7 in one
transaction); the three `DISPLAY_KIND` copies are byte-equivalent; `tour` and
`trace` payloads contain no tag field at all; all four schema gates read the one
constant and report an actionable message; the public API gains only an added
export; and the package ships `lib` and `src` wholesale, so new files need no
manifest change.

### Not repaired, and why

`pnpm format` rewrites ~250 unrelated fixtures and exits non-zero. Formatting is
therefore applied per changed file — every Go file in the diff round-trips
through the repository's own script unchanged, and every TypeScript, MDX, and
JSON file passes `prettier --check`. The hazard is published as its own issue
rather than repaired here: it is a formatter/fixture disagreement with no
bearing on this change, and fixing it means deciding whether those fixtures are
exempt or the script should stop rewriting string literals.

Five other observations from this cycle are published as issues rather than
carried: artifact nodes and the lint transport they need, `graphdump`'s complete
absence of tests, the hand-written viewer legend, the LSP provider's retention of
build-only state, and the benchmark's three-family edge breakdown.

### Round result

Not clean — four findings, each repaired. A fresh complete round follows over the
new head.

samchon (commented):
## Overall Self-Review — round 2 over `origin/master...16e82383c`

Fresh pass. One finding, and it is the kind this whole cycle has been guarding
against: an answer that was silently incomplete and said it was complete.

### The finding

**Five declarations citing one specification came back as three.** `lookup` caps
hits per file so one file's roster cannot crowd out a name ranking, and that cap
was applied to citation carriers too. Reproduced directly — one file, five
exported functions, all citing `docs/spec.md#rule`:

```
declarations citing the spec: 5
lookup returned: part1, part2, part3
next: answer | The ranked hits and their signatures resolve the name.
```

The cap is right for what it was written for. A citation is not ranked against
anything — it is an exact match on an address — and a module implementing one
specification across several functions is the ordinary shape of that answer
rather than a file dominating a shortlist. Carriers are exempt now; a name query
is still capped, which is what the cap is for.

**The limit was the other half, and it exposed an older inaccuracy.** The audit
every ranked operation shares already tells a reader that `truncated` marks
where more was left out. `tour` and `entrypoints` both carry that field.
`lookup` never did — so a cut looked exactly like a complete answer, and the
audit had been promising a field that did not exist. Tolerable for a name query,
where the shortlist is the design; not tolerable for an address, where the
caller asked which code implements a specification and is owed the fact that it
did not get all of it. It carries `truncated` now, and `next` names the repair
instead of claiming the question is resolved.

Both halves are pinned by a regression test that also holds the negative twin:
the per-file cap still governs a name query over the same fixture.

### Also verified this round

The full `packages/ttsc` Go suite passes (exit 0). The complete
`tests/test-graph` suite passes. Every claim the guides make was executed rather
than read — the reverse lookup, the tags on `details`, the traversal policy
under all three focuses, and the viewer's display family. Public API gains only
an added export; the package ships `lib` and `src` wholesale, so new files need
no manifest change.

### Round result

Not clean — one finding, repaired. A fresh complete round follows over the new
head.

samchon (commented):
## Overall Self-Review — rounds 3 and 4

**Round 3** over `origin/master...16e82383c`. One improvement, no defect.

The citation index carries a membership check that cannot fire — targets are
deduplicated per node and each node is visited once — and a reader who works that
out would delete it. It is kept, and now says why: the cost is a scan of one
address's carrier list, and the failure it prevents is a declaration reported
twice as implementing one specification.

Measured rather than assumed while reading `details`: tag density on a real
package. `@ttsc/lint` carries 864 tags across 863 declarations — **at most two on
any one** — for 62 KB of text in the whole dump. So the uncapped tag list on
`details` is not a token risk, which is what an earlier review left open as an
observation. The cap stays off, matching how `members` and `literals` are
treated: identity is returned whole.

**Round 4** over `origin/master...e87a6be7a`. Clean.

Read end to end: the tag collector, the host recording, the link walk, the wire
projection, the memory layer, both runners, the schema files, the viewers, and
the guides. Two things were checked specifically because they looked like
duplication and turned out to be sound —

- `documentationOf` runs twice per documented declaration, once in the node pass
  and once in the edge pass. TypeScript-Go caches the result on the source file,
  so the second is a lookup, and both passes run against the same Program, so
  they cannot disagree.
- A declaration whose documentation carries no tag still records a host. That is
  deliberate: a `{@link}` in ordinary prose is a reference too, and the host set
  is what the edge pass walks.

Every new schema field carries a doc comment that matches its behaviour,
including the two that state a capability gate and a truncation contract.

### Where this leaves the merge gate

Round 4 produced nothing to change. The next round runs against whatever head CI
settles on; if nothing further lands, that round is the clean one this gate
needs.

samchon (commented):
## Overall Self-Review — round 5 over `origin/master...16e82383c`+

No defect. Three boundaries that reading could not settle were run instead, and
each is now held by a test.

- **A declaration can carry two documentation blocks.** TypeScript attaches both,
  and each is a place a tag can sit. Confirmed by execution that both are read;
  reading one would have dropped the other silently.
- **An address outside the name tokenizer's alphabet answers.** `subwords` splits
  on ASCII alphanumerics, so a Korean address produces no term at all. This is
  the exact input the citation pass running *before* the empty-terms guard exists
  for, and that guard's motivating case had no test until now.
- **A tag with no text indexes nothing and is still carried.** It names nothing,
  so it must not enter the index; it is still a written fact, so `details` shows
  it.

The first is pinned in the native test that already owns line-endings and
repeated blocks; the other two in the address-rule e2e, beside its existing
prose and bare-number negatives.

### Round result

Clean on inspection; the additions are coverage rather than repair. A fresh round
follows over the new head, since the tree changed.

samchon (commented):
## Overall Self-Review — round 6 over `origin/master...a9b1a683f`

This round checked the change against its own published contract rather than
against the code: the twenty-four acceptance items of #1229 and the nineteen of
#1230, one by one, against the tests that exist.

Three were still unpinned. All three share a shape — **the correct answer is an
absence**, which is exactly what goes unnoticed when it stops being produced:

- **A tag on a closure declaration is recorded.** The single collection point
  covers it, and that property is what lets a new declaration form need no
  handling of its own. Now asserted rather than inferred.
- **A tag written in a dependency contributes nothing.** The collector reaches
  only what the build pass records, so it holds by construction — and "by
  construction" is what stops being true when the construction changes. The
  failure it guards is answering "which code implements this specification" with
  somebody else's code, silently, because a vendored address looks like any
  other.
- **Neither `tour` nor `trace` carries a tag.** This was the cycle's token-budget
  decision and had only been checked by hand. The fixture's declaration does
  carry a tag, so the absence is now provably a decision rather than an empty
  project.

Every acceptance item in both issues now has a test. Native graph and command
suites pass on the new head.

### Round result

Not clean — three coverage gaps, all closed. A fresh round follows.

samchon (commented):
## Overall Self-Review — round 7 over `origin/master...d4ddd3076`

The same contract check as round 6, applied to #1230's nineteen items. Four were
unpinned, and one of them was the issue's own evidence.

**The premise had no test.** The argument for this edge is that it is a compiler
fact rather than a text match, and the compiler states it itself: an import
supporting only a documentation link survives `noUnusedLocals` while one
supporting nothing does not. That asymmetry is what the issue was opened on, and
it lived only in prose. It is a test now, and it fails in both directions that
matter — if a TypeScript upgrade ever stops counting links as uses, which would
make this edge a text match dressed as a checker fact, and if this pass ever
changes what the compiler reports, which it must never do.

Three others closed with it: a link in a `//` comment produces no edge, and a
namespace member and a closure each resolve their own. The last two are the
property that matters structurally — both are nodes the build pass records, so
both are reached through the shared host set rather than a form-specific branch,
which is the design that replaced the container walk earlier in this cycle.

The relation is now asserted from the cited end as well: `details` on the type
lists the declaration whose documentation names it. That is the direction a
reader uses to go from a contract to the code answering for it, and it comes
from the same edge rather than a second index.

**Both acceptance lists are now fully covered** — twenty-four items in #1229,
nineteen in #1230, every one with a test.

### Round result

Not clean — four coverage gaps, all closed. A fresh round follows.

samchon (commented):
## Overall Self-Review — round 8 over `origin/master...91d779eac`

One improvement, and one mistake of my own caught inside the round.

**The improvement.** The release-candidate smoke script reads
`DUMP_SCHEMA_VERSION` out of `loadGraph.ts` with a regular expression, because it
runs against a packaged binary in a workflow that has no reason to have built the
TypeScript package first. That is the right call and it fails loudly if the
declaration is rewritten — but a reader editing the line had no way to know the
dependency existed. The declaration now names it.

**The mistake.** I committed that comment without running the formatter on it,
and the push carried a Prettier violation. Caught by checking the file in the
same breath, amended, and the whole diff re-verified: every TypeScript, MDX,
JSON, and CJS file in it passes `prettier --check`, and every Go file
round-trips through the repository's own script unchanged.

That is the second time this cycle formatting has gone wrong in a way the
repository's own command would not catch, for two different reasons — the first
was `pnpm format` rewriting 250 unrelated fixtures and exiting non-zero, which is
published as its own issue. The per-file check is the workaround, and it only
works when it is actually run, which is what failed here.

### Round result

Not clean — one improvement, one self-inflicted defect repaired. A fresh round
follows.

samchon (commented):
## Overall Self-Review — round 9 over `origin/master...8066b9e86`

One defect, mine, found by running the check over **every** changed file rather
than the ones I had just touched.

`packages/graph/src/viewer/index.html` was unformatted. It is formatted on
master, so the legend entry I hand-wrote for the documentation-reference family
broke it — Prettier wraps that markup differently. Formatted, rebuilt, and the
bundled viewer still produces its payload.

That is the third formatting problem in this cycle and the second I caused. The
pattern is now clear enough to state: the repository's own `pnpm format` cannot
be run here without rewriting 250 unrelated fixtures, so this cycle formats per
changed file — and a per-file discipline only covers the files you remember to
name. The reliable form is the one this round used, and it is now what every
remaining round will do: run the check across the complete diff, not across the
edit.

Both gates now pass over the whole change: every Go file round-trips through the
repository's script unchanged, and every TypeScript, MDX, JSON, CJS, and HTML
file passes `prettier --check`.

### Round result

Not clean — one defect, repaired. A fresh round follows.

samchon (commented):
## Overall Self-Review — round 10 over `origin/master...fea59b507`

Clean on inspection. This round checked the things a reading round can miss by
looking at the change as an artifact rather than as code.

**Nothing in it is unreferenced.** Every symbol the change introduces —
`documentationTarget`, `leadingToken`, `docTagsOf`, `citing`, `collectDocTags`,
`collectDocRefs`, `recordDocHost`, `elideTagText` — has production callers, so no
dead code shipped behind the tests.

**The merge will close exactly what survives.** Two `Close #` lines, one per
issue, and both fixes are present at `HEAD`: the node's tag field for #1229 and
the edge kind for #1230. No revert, no fixup, no work-in-progress commit in the
range, so the squash carries no closing keyword whose fix was later undone.

**No debris.** No `console.log`, no stray print, no probe file; the `@todo`
matches in the diff are the tag name quoted in prose, which is the population
this change is about.

**The campaign's own three invariants hold, checked mechanically.** The compiler
host names no convention. No source body is inlined — the only `SourceTexts`
matches are a test fixture calling an existing API. Nothing in the diff judges
coverage, resolution, or obligation; that stays the linter's.

Both format gates pass over the complete diff. CI on this head is green so far
with no failures.

### Round result

Clean. If the running suite and CI confirm, the next round is the merge gate.

samchon (commented):
## Overall Self-Review — round 11 over `origin/master...fea59b507`

Clean. This round measured the thing the graph product is judged on rather than
reading it: what the change costs a caller.

Measured against this repository's own `@ttsc/lint` package — 1,301 nodes, 864
documentation tags:

- **`tour` is unchanged.** 8,873 characters, carrying no tag at all. That was the
  design decision of the cycle and it is now a number rather than an intention.
- **`details` on three tagged declarations** is 2,904 characters, of which the
  tags are 557 — 19%. That is the operation whose contract is to return what a
  declaration *is*, and a fifth of it now says what the declaration answers to.
- **The reverse index is free.** Ten thousand lookups take under a millisecond;
  it is a map keyed by address, built once while the graph loads.

Nothing here changes a decision, which is the point of measuring: the token
budget the cycle reasoned about holds when run.

### Round result

Clean. Two consecutive clean rounds now, and the merge gate needs CI on the same
head.

samchon (commented):
## Overall Self-Review — round 12, and a rebase

This round asked the question the merge gate turns on rather than a question
about the code: **is this still the change that was reviewed?**

It was not quite. `master` had moved two commits under the branch —
`fix(driver): build the plugin emit transformer chain from the parse tree`
(#1232) and `fix(unplugin): keep the content comparison while a same-tick
rewrite can hide behind a metadata signature` (#1228). Neither touches a file
this change touches; the two surfaces are disjoint and there was no conflict.

Disjoint file lists are not the same as a verified integration, though, so the
branch is rebased onto the current `master` and re-verified there — that state,
not the one the earlier rounds ran on, is what would merge:

- `go build ./...` clean.
- `internal/graph` and `cmd/ttscgraph` suites pass.
- Every changed Go file round-trips through the repository's format script
  unchanged; every TypeScript, MDX, JSON, CJS, and HTML file passes
  `prettier --check`.

CI is now running against the rebased head, and the merge gate needs it green on
that exact SHA rather than on the one the earlier lanes passed.

### Round result

Clean, on the integrated state.

samchon (commented):
## CI: one red lane on the rebased head, diagnosed and rerun

The `Ubuntu` lane failed on `a3b3157a6` inside `experimental/typia/tests/test-langchain`:

```
tests/test-langchain start:     errno: -104,
tests/test-langchain start:     code: 'ECONNRESET',
tests/test-langchain start:     syscall: 'read'
    at TLSWrap.onStreamRead (node:internal/stream_base_commons:216:20)
```

That is a TLS read reset — an outbound connection dropped mid-request, not an
assertion. Three things place it outside this change:

- **The surfaces are disjoint.** This pull request touches no file under
  `experimental/typia`, and nothing it changes is reachable from that suite.
- **The same code passed before the rebase.** The lane was green on
  `fea59b507`; the only difference on `a3b3157a6` is `master`'s own two commits
  (#1232, #1228), neither of which touches LangChain either.
- **`master` is green.** Its last several runs, including the two commits just
  named, all succeeded.

Diagnosed as a transient network failure rather than dismissed as someone
else's: the log names the syscall and the layer, and the reasoning above rules
out the change. The failed job is rerunning; if it reproduces, that is a
different finding and will be treated as one rather than rerun again.

samchon (commented):
## Overall Self-Review — round 13 over `origin/master...a3b3157a6`

Clean. This round applied the lesson of round 2's defect to every other bound on
the response path, since one cap silently narrowing an exact answer means the
others deserve the same question: *does this bound a ranking, or an answer, and
does the result say which?*

Enumerated and checked, across all four runners:

- **`MAX_DOC_CHARS`** on a tag text — bounds prose, never the address, and marks
  itself with an ellipsis exactly as `doc` does.
- **`PER_FILE`** — bounds a name ranking only; citation carriers are exempt as of
  round 2, which is the defect that started this.
- **`lookup`'s limit** — reports `truncated` and names the repair, as of round 2.
- **`entrypoints`' limit** — already reported `truncated`; unaffected by dropping
  the tag field from its hits.
- **`details`' member limit** — withdraws the completeness claim through
  `membersCapped`; its neighbour and dependency limits are declared in the audit
  as a short orientation slice with `trace` for the rest.

Every bound this change touches or introduces either bounds a ranking, or bounds
an answer and says so. Nothing new found.

### Round result

Clean. Three clean rounds, one of them on the rebased state, with the CI failure
in between diagnosed as an outbound connection reset and cleared by rerun.

Issue #1204, opened 2026-08-15
Plugin emit passes the transformed source file to `getScriptTransformers`, costing ~59ms per file


kakasoo:
Hi samchon. Following up on the TypeScript 7 migration — after moving a backend from typia 12 /
TS 6 / `ts-patch` to typia 13 / TS 7 / `ttsc`, every build got 2.4–3.9x slower. I profiled it down
to one line in `ttsc`'s plugin emit path, and changing that line makes our app build go from
**92s to 16.7s** with byte-identical output.

## Where the time goes

`EmitWithPluginTransformers` in `packages/ttsc/driver/emit_plugin.go` runs, per file:

```go
out := sf
for _, transform := range transforms {
    out = transform(ec, out)          // typia injects validators; the AST grows ~20x
}
shimast.SetParentInChildrenUnset(out.AsNode())
restoreOriginalDeclarationSymbols(ec, out.AsNode())
for _, tr := range shimcompiler.GetScriptTransformers(ec, host, out) {   // <-- `out`
    out = tr.TransformSourceFile(out)
}
```

`getScriptTransformers` does this on whatever source file it is handed
(`typescript-go/internal/compiler/emitter.go:115`):

```go
emitResolver.MarkLinkedReferencesRecursively(sourceFile)
```

That is a full recursive walk of the file's AST through the checker. Handing it `out` means
walking the **post-plugin** tree — for us ~500 bytes of source that typia expanded into ~9KB of
validator code, full of synthetic nodes the checker never saw. tsgo's own emitter passes the
original `sourceFile` here.

I added timers inside the loop (600-file synthetic project, details below):

| phase | before |
|---|---|
| plugin transforms (typia) | 0.82s |
| `SetParentInChildrenUnset` | 0.02s |
| `restoreOriginalDeclarationSymbols` | 0.02s |
| **`GetScriptTransformers` (chain construction)** | **35.40s** |
| `TransformSourceFile` (actually transforming) | 0.12s |
| print | 0.00s |

Building the transformer chain costs 3,500x more than running it. At 600 files that is ~59ms each.

## The change

```diff
- for _, tr := range shimcompiler.GetScriptTransformers(ec, host, out) {
+ for _, tr := range shimcompiler.GetScriptTransformers(ec, host, sf) {
      out = tr.TransformSourceFile(out)
  }
```

| phase | before | after |
|---|---|---|
| `GetScriptTransformers` | 35.40s | **0.01s** |
[...]

PR #1232, opened 2026-08-18
fix(driver): build the plugin emit transformer chain from the parse tree

samchon:
Campaign cycle for the plugin emit path's linked-reference marking. This pull
request owns the complete accepted issue set for the cycle:

- #1204 — plugin emit passes the transformed source file to
  `getScriptTransformers`, so `MarkLinkedReferencesRecursively` walks the
  post-plugin tree.

## Intent

`EmitWithPluginTransformers` hand-assembles tsgo's JavaScript emit pipeline. It
currently builds the builtin transformer chain from `out`, the post-plugin tree,
where upstream's own emitter builds it from the parse tree. The only thing that
argument feeds which can differ between the two is
`emitResolver.MarkLinkedReferencesRecursively`, a per-node checker resolution
walk under the single checker mutex — so a plugin that expands the AST pays that
walk over its own generated nodes.

Two discovery rounds against `9609e2a9` established that the behavior the
current argument appears to protect does not exist, and that the one behavior
that does change moves toward `ts-patch` / TypeScript 5-6 parity. The full
adjudication is recorded on #1204.

## Scope

- Build the builtin chain from the parse tree while still transforming `out`.
- Pin both sides of the elision outcome with regression tests, neither of which
  is covered today.
- Realign `prototype_assembled_alias_test.go` so it keeps prototyping the
  production assembly.
- State the reference-marking contract in the plugin-author documentation.

## Verification

Pending. Local evidence and the CI lanes will be recorded as formal reviews on
this pull request.

Closes are carried by the individual commits, not by this body.


packages/ttsc/driver/emit_plugin.go
```diff
@@ -96,13 +96,22 @@ func (p *Program) EmitLinkedTransforms(writeFile shimcompiler.WriteFile) ([]Diag
...
+// clears it again, so ast.IsParseTreeNode reports true for a container that
+// carries no symbol. Every EmitResolver reference method (GetReferenced*,
+// IsReferencedAliasDeclaration, ...) tests exactly that predicate on the node it
+// is handed, so the builtin module and runtime-syntax transformers still reach
+// checker name resolution through such a container, and resolveName then calls
 // getSymbolOfDeclaration(container) — which reads container.Symbol() and nil-
-// panics on a rebuilt class/interface/enum. Restoring the symbol from the
-// original (the symbol object is shared and node-independent for lookup) lets
-// the resolver mark references the same way it would on the parse tree.
+// panics on a rebuilt class/interface/enum.
```
samchon: `enum` does not belong in this list. `binder.NameResolver.Resolve` nil-guards `KindEnumDeclaration` (`if enumSymbol == nil { break }`) and `KindModuleDeclaration` explicitly. The unguarded dereference is `getSymbolOfDeclaration(location).Members` for `KindClassDeclaration | KindClassExpression | KindInterfaceDeclaration`, plus `KindExpressionWithTypeArguments` and `KindComputedPropertyName`.

Inherited from the original comment, but this commit rewrote the paragraph, so it is this commit’s to get right.

packages/ttsc/driver/emit_plugin.go
```diff
@@ -96,13 +96,22 @@ func (p *Program) EmitLinkedTransforms(writeFile shimcompiler.WriteFile) ([]Diag
...
 //
-// tsgo's emit resolver then walks the transformed tree in
-// MarkLinkedReferencesRecursively and, when it resolves an identifier whose
-// scope chain passes through one of those rebuilt containers, calls
+// A rebuilt container is indistinguishable from a parse node to tsgo's own
+// guards. The emit context stamps NodeFlagsSynthesized when its factory creates
+// a node, but ast.updateNode then ASSIGNS `updated.Flags = original.Flags` and
+// clears it again, so ast.IsParseTreeNode reports true for a container that
+// carries no symbol. Every EmitResolver reference method (GetReferenced*,
+// IsReferencedAliasDeclaration, ...) tests exactly that predicate on the node it
```
samchon: "Every EmitResolver reference method ... tests exactly that predicate on the node it is handed" is too strong. `SetReferencedImportDeclaration` has no guard at all, and `GetReferencedImportDeclaration` uses the predicate to *branch* into a `jsxLinks` lookup rather than to bail out.

packages/ttsc/driver/emit_plugin.go
```diff
@@ -96,13 +96,22 @@ func (p *Program) EmitLinkedTransforms(writeFile shimcompiler.WriteFile) ([]Diag
...
+// is handed, so the builtin module and runtime-syntax transformers still reach
```
samchon: This rewritten rationale asserts that the builtin module and runtime-syntax transformers still need this function. That is unproven: stubbing the call out leaves the entire `test/driver` suite green on both this commit and its parent, and four probe configurations covering generics, `emitDecoratorMetadata` decorators, interfaces, enums, namespaces, and rebuilt ancestors produced no panic.

The function stays, because unreachability is just as unproven and #201 added it against a real panic. But the comment has to say the mechanism is why the guard cannot be assumed dead, not that it is observed live.

packages/ttsc/driver/emit_plugin.go
```diff
@@ -180,7 +189,30 @@ func (p *Program) EmitWithPluginTransformers(transforms []PluginTransform, write
...
+      // getScriptTransformers reads its file for three things: whether it is a
+      // JS file, whether its language variant is JSX, and
+      // emitResolver.MarkLinkedReferencesRecursively. The first two are fixed
+      // per file and survive UpdateSourceFile, so marking is the whole
+      // difference — and marking is a per-node checker resolution walk under
+      // the single checker mutex, which on a plugin-expanded tree runs over
+      // nodes the binder never saw.
+      //
+      // Those marks are read back through EmitContext.ParseNode, so their key
+      // space is the parse tree; marking it is the write side that matches.
```
samchon: The marks live on `aliasSymbolLinks`, keyed by **symbol**, not by node. The parse tree is the key space of the *read guard* — `ImportElisionTransformer` maps through `EmitContext.ParseNode` before querying — not of the marks themselves. The conclusion holds; the wording does not.

packages/ttsc/driver/emit_plugin.go
```diff
@@ -180,7 +189,30 @@ func (p *Program) EmitWithPluginTransformers(transforms []PluginTransform, write
...
+      // Nothing is lost by not walking the plugin's nodes: UpdateSourceFile
+      // rebuilds a SourceFile without its Locals, and a module-scope import
+      // binding is only reachable through Locals, so a synthetic identifier
```
samchon: "only reachable through `Locals`" is false. `resolveName` also consults the source file’s module `Exports`, and `restoreOriginalDeclarationSymbols` restores the rebuilt `SourceFile`’s symbol, so that branch is live for synthetic nodes.

No behavioral consequence — an exported alias is preserved unconditionally by `IsReferencedAliasDeclaration` regardless of marking — but the justification as written is not true.

packages/ttsc/driver/emit_plugin.go
```diff
@@ -180,7 +189,30 @@ func (p *Program) EmitWithPluginTransformers(transforms []PluginTransform, write
       }
       shimast.SetParentInChildrenUnset(out.AsNode())
       restoreOriginalDeclarationSymbols(ec, out.AsNode())
-      for _, tr := range shimcompiler.GetScriptTransformers(ec, host, out) {
+      // Build the chain from the PARSE tree and transform the plugin's tree.
```
samchon, line 200: Worth one clause here: the change also makes this lane deterministic across repeated emits on one `Program` (watch, multi-emit). Marks are monotonic and never cleared, so before this commit a second `EmitWithPluginTransformers` call could observe marks left behind by the first pass’s plugin tree.

samchon (commented):
## Individual Self-Review — `2f520fcea` "fix(driver): build the plugin emit transformer chain from the parse tree"

Advisory read-only review of the parent-to-commit diff, run adversarially against the pinned typescript-go. This commit resolves #1204.

### What landed

`EmitWithPluginTransformers` now builds tsgo's builtin transformer chain from the parse tree while still applying it to the plugin's tree, plus two regression tests, a realigned prototype assembly test, corrected comments, and the plugin-author contract in `driver-api.mdx`.

### Confirmed

- The three-use claim about `getScriptTransformers`' file argument is exact (`emitter.go:108`, `:109`, `:115`), and both per-file constants survive `UpdateSourceFile` through `SourceFile.copyFrom`.
- Marking the parse tree matches upstream in both places the commit cites (`emitter.go:115`, `tstransforms/importelision_test.go:263`).
- `TestEmitWithPluginTransformerRemovedValueUseKeepsItsImport` was reproduced as discriminating: it fails on the parent commit and passes here.
- Forbidden-section audit clean. No hardcoding, no consumer special-casing, no assertion-shaped branch, no forced design. The change removes a divergence from upstream rather than adding a patch.

The central claim survived a deliberate attempt to refute it.

### Adjudication — all eleven findings accepted

The most important one is not a style point. **The commit understates the defect class.** A plugin that merely *rebuilds* a value reference — `ec.Factory.NewIdentifier(...)` plus `ec.SetOriginal(...)`, the shape any partial-rewrite transform produces — emitted this on the parent commit:

```js
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
exports.a = void 0;
exports.a = dep_1.foo + 41;
```

`dep_1` is never defined. That is a runtime `ReferenceError` when the module is loaded, not the lost-side-effect case the commit message describes. I reproduced it directly against this repository's own `emit_plugin_ancestor_regeneration_preserves_export_resolution_test.go`, which **passes on that output**, because its assertion `(\w+)\.foo \+ 41` matches a dangling alias just as happily as a bound one.

So the existing suite contained a hole precisely where the crash lived. Corrections follow as a further commit on this pull request:

| finding | disposition |
|---|---|
| 1 — severer uncovered class, existing test passes on broken output | accept; assert the `require` binding in the ancestor-regeneration test and restate the class in the commit message and the documentation |
| 2 — realigned prototype test does not discriminate | accept; assert the `require` binding there too |
| 3 — documentation states marking happens *before* the transform runs | accept; marking happens after the plugin pass, at the chain-construction call |
| 4, 5 — two absolutes in the documentation | accept; `SetOriginal`-linking is a second way to reach a module, and a `SetOriginal`-linked synthetic import is not exempt from elision |
| 6 — "their key space is the parse tree" | accept; the marks are keyed by symbol, the parse tree is the key space of the read guard |
| 7 — "only reachable through `Locals`" | accept; the module `Exports` branch of `resolveName` is live. No behavioral consequence, but the justification as written is false |
| 8 — "class/interface/enum" | accept; enum and module are nil-guarded upstream, the unguarded dereference is the class/interface/`ExpressionWithTypeArguments`/`ComputedPropertyName` set |
| 9 — rewritten `restoreOriginalDeclarationSymbols` rationale is unverified | accept; four probe configurations produced no panic with the call stubbed out. The claim will be stated as unproven rather than asserted, and the function kept, because unreachability is equally unproven |
| 10 — "every EmitResolver reference method tests that predicate" | accept; `SetReferencedImportDeclaration` has no guard and `GetReferencedImportDeclaration` branches on it |
| 11 — unstated upside | accept; marks are monotonic and never cleared, so before this change a second emit on one `Program` could observe the first pass's plugin-tree marks |

Findings 6 through 11 change no behavior. They are claims in comments and documentation, and they still have to be true, because the next person to touch this lane will act on them.

### Not verifiable here

The reported timings (35.4s of a 37.0s 600-file typia build), and the `ts-patch` / TypeScript 5-6 check-time-marking equivalence — neither `ts-patch` nor TypeScript 5-6 is present in this repository or in the pin. Both are consistent with the mechanism, but they rest on the reporter's measurements.

packages/ttsc/test/driver/prototype_assembled_alias_test.go
```diff
@@ -93,8 +93,11 @@ func TestPrototypeAssembledAlias(t *testing.T) {
   shimast.SetParentInChildren(transformed.AsNode())
 
   // builtin emit chain (type-erase, import-elision, module-transform, ...) in
-  // the SAME EmitContext, applied after the plugin transformer.
-  builtins := shimcompiler.GetScriptTransformers(ec, host, transformed)
+  // the SAME EmitContext, applied after the plugin transformer. The chain is
+  // built from the PARSE tree and applied to the transformed one, exactly as
+  // EmitWithPluginTransformers does it: the file handed to GetScriptTransformers
+  // is the reference-marking target, not the file being transformed.
+  builtins := shimcompiler.GetScriptTransformers(ec, host, sf)
```
samchon, line 100: The realignment is right but unpinned. Running the pre-realignment form (`GetScriptTransformers(ec, host, transformed)`) against this commit still passes, emitting `exports.a = dep_1.foo;` with no `require` line — so a revert to `transformed` would be invisible to this test.

Assert the binding exists, not only the alias.

website/src/content/docs/development/reference/driver-api.mdx
```diff
@@ -177,6 +177,8 @@ reference := ec.Factory.NewPropertyAccessExpression(
 )
 ```
 
+Synthesizing that import is the only way to reach a module from injected code. The driver marks linked references on the parse tree, before your transform runs, so a value reference your transform injects cannot keep an import alive that the source itself uses only in type position — TypeScript-Go's import elision has already decided that import carries no runtime value. Your own synthetic import is never at risk: it has no parse-tree original, so elision cannot reason about it and preserves it unconditionally. The reverse also holds. When your transform removes the last value use of an import the source really had, that import survives, because the marks describe the file the checker analyzed rather than the file you produced. This is the same contract `ts-patch` had on TypeScript 5 and 6, where the marks were fixed at check time.
```
samchon: Three corrections in one paragraph:

1. "before your transform runs" is wrong. The plugin transforms run first; marking happens afterwards, when the builtin chain is constructed. The outcome is equivalent because the marked file is the parse tree either way, but the stated ordering is not what the code does.
2. "the only way to reach a module from injected code" — `prototype_assembled_alias_test.go` reaches `./dep` from injected code by `SetOriginal`-linking a fresh identifier to a parse node, and the module transform aliases it.
3. "never at risk: it has no parse-tree original" holds only while the plugin does not `SetOriginal`-link it, since `EmitContext.ParseNode` walks `MostOriginal` before testing.

The paragraph should also name the failure this actually prevents: a rebuilt reference to a pre-existing import emitting an alias whose `require` binding was elided, which throws `ReferenceError` at load time.

packages/ttsc/test/driver/emit_plugin_rebuilt_decorator_reference_keeps_its_binding_test.go
```diff
@@ -0,0 +1,101 @@
...
+// hold at once here: the rebuilt containers must still resolve, and the
+// decorator's import must still be marked from the parse tree.
+//
+//  1. `index.ts` imports `Route` and applies `@Route()` to a class method.
+//  2. A plugin rebuilds the decorator's call expression, SetOriginal-linking a
+//     fresh `Route` identifier to the parse-tree one.
+//  3. Assert the decorator is aliased AND that the alias names a require
+//     binding the emitted file actually declares.
+func TestEmitWithPluginTransformerRebuiltDecoratorReferenceKeepsItsBinding(t *testing.T) {
+  root := t.TempDir()
```
samchon, line 40: This is the shape the lane exists for. `restoreOriginalDeclarationSymbols` names "a decorator call on a controller method" as its own example, and a nestia-style host rewrites exactly that.

Without the change its emitted output keeps `(0, dep_1.Route)()` inside `__decorate` with no `const dep_1 = require("./dep")` in the file. The class was not proved by the named-import case: marking reaches a decorator through `markDecoratorAliasReferenced`, a different path from the plain identifier the other tests exercise.

packages/ttsc/test/driver/emit_plugin_rebuilt_default_import_reference_keeps_its_binding_test.go
```diff
@@ -0,0 +1,94 @@
...
+// fix proved only on the named shape would leave the default shape free to keep
+// emitting an alias whose require binding was elided, which throws
+// ReferenceError as soon as the module loads.
+//
+//  1. `index.ts` imports `dep` as a default binding and calls `dep.run()`.
+//  2. A plugin rebuilds that call's callee expression from fresh ec.Factory
+//     nodes, SetOriginal-linked back to the parse-tree identifier.
+//  3. Assert the emitted call is aliased AND that the alias names a require
+//     binding the emitted file actually declares.
+func TestEmitWithPluginTransformerRebuiltDefaultImportReferenceKeepsItsBinding(t *testing.T) {
```
samchon, line 32: Import elision does not test the `ImportDeclaration`; it tests the clause and its bindings, so a default binding and a named specifier reach the resolver as different node kinds.

The broken output here is worse than the named case: it loses the `__importDefault` helper along with the `require`, leaving `exports.a = dep_1.default.run();` with nothing defining `dep_1`.

samchon (commented):
_Reviewed commit: `82feea310`. Anchored on the current head so the files added in `5bf69f23c` resolve for the inline notes._

## Individual Self-Review — `82feea310` "fix(driver): pin the dangling-alias crash the marking change actually prevents"

Advisory read-only review, run adversarially against the pinned typescript-go. No blocking finding. Five should-fix and three nits, all accepted and answered in `5bf69f23c`.

### Confirmed against the pin

- `ast.updateNode` assigns `updated.Flags = original.Flags`, and `IsParseTreeNode` is `Flags&NodeFlagsSynthesized == 0`, so a rebuilt container defeats the guard.
- The unguarded dereference is exactly class, class expression, and interface; module and enum are nil-checked. The rewritten comment also fixes an error the parent had: `getSymbolOfDeclaration` returns nil rather than panicking, and the dereference is at the call site.
- "**Most** EmitResolver reference methods bail on that predicate" is right, and an improvement over the parent's "Every": `GetReferencedImportDeclaration` does not bail.
- The crash mechanism is exact. The module transform aliases through `GetReferencedImportDeclaration(MostOriginal(node))`, so a rebuilt reference still prints `dep_1.foo` while its binding is elided. Reproduced verbatim.
- Marks are monotonic: `links.referenced = true` is the only write in the whole checker package.
- Neither strengthened assertion is over-fitted. The ancestor test derives the alias from the output and asserts the binding for that same name; the prototype test asserts only the module specifier.
- `restoreOriginalDeclarationSymbols` stubbed out leaves the entire Go suite green, reproducing the comment's own claim.

### Accepted findings

**The commit message was wrong twice, and one of those was a reversal I should not have made.**

`82feea310` claimed "three tests now fail without the change where none did before". Reverting the change against the parent `2f520fcea` makes **one** test fail, the elision test that commit added. Three failed after `82feea310`. The claim that carries the point survives: the dangling-alias crash class was unpinned, and the ancestor-regeneration test was passing on the broken output.

`82feea310` also dropped the parent's `Locals` argument as false. That reversed a correct argument. `binder.declareModuleMember` routes an alias into `Exports` only for an export specifier or an exported import-equals; every other import binding goes into `Locals`, and `SourceFile.copyFrom` copies neither `Locals` nor `Symbol`. So "a module-scope **import** binding is only reachable through `Locals`" was true, and the earlier review's finding 7 generalized "import binding" to "binding" to reach the opposite conclusion. I accepted that too readily. The code had already restored the argument; `5bf69f23c` corrects the message that disowned it.

| finding | disposition |
|---|---|
| "three fail where none did" is false | accept; corrected in `5bf69f23c`'s message |
| the `Locals` justification was sound as originally scoped | accept; reversal reversed, verified from `binder.declareModuleMember` |
| "a synthetic node has no NodeLinks of its own" is false | accept; `getResolvedSymbol` creates an entry for any node, and the marks are symbol-keyed regardless. Sentence removed |
| the idempotency note inverts causality and invokes a watch rebuild that spawns a fresh process | accept; reworded to a second `EmitWithPluginTransformers` call on one `Program`, which is the hazard that exists |
| `driver-api.mdx` em-dashes and "worth knowing" signposting violate the documentation skill's Voice rule | accept; both removed |
| `shim/ast/parent.go` names reference marking as the pass that walks parents on the spliced tree | accept; that pass no longer sees the spliced tree here. The rule holds for the module transform, runtime syntax, and printer, so the example is replaced and the rule kept |
| the mdx says `SetOriginal`-linking puts a synthetic import "back under that import's marks" without saying which node elision tests | accept; it tests the clause, named specifier, or namespace import, not the declaration. Tightened |
| test doc comments lack the `Verifies …` headline the development skill shows | reject, with reason. A Go doc comment must open with the identifier it documents, and every sibling in `packages/ttsc/test/driver/` uses `// TestX <verb phrase>`. The skill's three-part shape is satisfied by the why-paragraph and the numbered list, which the new tests carry |

### The coverage gap that mattered most

The reviewer's uncovered list was right that the named-import case does not prove the class. Import elision tests the clause and its bindings rather than the `ImportDeclaration`, and marking reaches a decorator through `markDecoratorAliasReferenced` rather than the plain identifier path. `5bf69f23c` adds two tests, both verified to fail without the change:

- a rebuilt reference through a **default import**, whose broken output loses the `__importDefault` helper along with the `require`;
- a rebuilt reference inside a **decorator**, the shape a nestia-style host produces and the example `restoreOriginalDeclarationSymbols` itself names. Its broken output keeps `(0, dep_1.Route)()` inside `__decorate` with no binding in the file.

Five tests now fail without the change.

Injected decorators and injected JSX were also listed as uncovered. Those belong to the "inject" class, whose behavior this change does not alter: a synthetic decorator or JSX identifier cannot resolve to an import binding on either side, because a rebuilt `SourceFile` has no `Locals`. That class is pinned in its general form by the injected-reference test.

### Not verifiable here

The specific four probe configurations `82feea310`'s message cites, and the historical claim that `ts-patch` on TypeScript 5 and 6 fixed the marks at check time. Neither `ts-patch` nor TypeScript 5 or 6 is present in this repository or in the pin.

packages/ttsc/shim/ast/parent.go
```diff
@@ -6,9 +6,13 @@ import innerast "github.com/microsoft/typescript-go/internal/ast"
...
 //
 // A transform that splices freshly built (synthetic) nodes onto a SourceFile
-// must call this before emit: passes that walk parents — e.g. the emit
-// resolver's reference marking (MarkLinkedReferencesRecursively) — dereference
-// Parent and would hit nil on a synthetic node otherwise.
+// must call this before emit: the builtin transformers that then walk the
+// spliced tree — module transform, runtime syntax, printer — dereference Parent
+// and would hit nil on a synthetic node otherwise.
+//
+// The emit resolver's reference marking used to be the example here. It no
```
samchon: The rule this comment states is still correct, but its example had gone stale: reference marking no longer walks the spliced tree in the plugin emit lane. Replacing the example rather than the rule keeps the reason a plugin author actually needs (the module transform, runtime syntax, and the printer all dereference `Parent`), and records why the old one was dropped so it is not restored by someone reading `2f520fcea` alone.

website/src/content/docs/development/reference/driver-api.mdx
```diff
@@ -177,6 +177,14 @@ reference := ec.Factory.NewPropertyAccessExpression(
...
+
+A reference you **inject** cannot revive an import the source uses only in type position. TypeScript-Go already proved that import carries no runtime value, and your injected node was never part of what it proved. Synthesize your own import for anything your generated code needs. The reverse also holds: when your transform removes the last value use of an import the source really had, that import survives, because the marks describe the checked file rather than the one you produced. This is the contract `ts-patch` had on TypeScript 5 and 6, where the marks were fixed at check time.
+
+A **synthetic** import is preserved unconditionally, since it has no parse-tree original for elision to reason about, but only while it stays synthetic. Elision tests the import clause, named specifier, or namespace import rather than the declaration, so `SetOriginal`-linking one of those to its parse-tree counterpart puts it back under that counterpart's marks. `EmitContext.ParseNode` walks the original chain before deciding.
```
samchon: Corrected for precision: elision applies its guard to the import clause, named specifier, or namespace import, not to the `ImportDeclaration`, so `SetOriginal`-linking has to reach one of those nodes for the claim to hold.

The paragraph also lost its em-dashes and the "worth knowing" signposting, per the documentation skill Voice rule.

samchon (commented):
## Overall Self-Review — final round, clean

Ten complete rounds over the base-to-head diff. Round 10 found nothing to improve, so this is the merge-gate review. Every earlier round is summarized below because each one changed the branch, and the record of what a round caught is the useful part.

### What the branch is

`EmitWithPluginTransformers` hand-assembles tsgo's JavaScript emit pipeline. It built the builtin transformer chain from `out`, the post-plugin tree, where upstream builds it from the parse tree. One line moves it to `sf`; the chain is still applied to `out`.

Reported as a performance problem: 35.4s of a 37.0s 600-file typia build spent constructing the chain, because `MarkLinkedReferencesRecursively` is a per-node checker resolution walk under the single checker mutex and it was walking a plugin-expanded tree.

The campaign found it is a correctness problem first. A plugin that **rebuilds** a value reference, which is what any partial rewrite produces, left the reference's import unmarked. The module transform still aliased the reference and elision still dropped the binding that alias names, so the lane emitted

```js
exports.a = dep_1.foo + 41;
```

with nothing in the file defining `dep_1`. `ReferenceError` at module load. `emit_plugin_ancestor_regeneration_preserves_export_resolution_test.go` builds exactly that shape and was passing on exactly that output, because its assertion matched a dangling alias as happily as a bound one.

### Coverage

Seven tests now fail when the driver line is reverted. Before this branch, none did.

| test | shape it pins |
|---|---|
| `AncestorRegenerationPreservesExportResolution` | named import, rebuilt ancestor and leaf |
| `RebuiltDefaultImportReferenceKeepsItsBinding` | default binding; broken output also loses `__importDefault` |
| `RebuiltNamespaceImportReferenceKeepsItsBinding` | `import * as ns`; broken output also loses `__importStar` |
| `RebuiltImportEqualsReferenceKeepsItsBinding` | `import x = require(...)`, the alias declaration kind outside an import clause |
| `RebuiltDecoratorReferenceKeepsItsBinding` | legacy decorator lowering, which moves the reference into a `__decorate` call after the class |
| `RebuiltReferenceKeepsItsEsmImport` | `module: esnext`, where the whole import declaration disappears rather than just a binding |
| `RemovedValueUseKeepsItsImport` | a transform that deletes an import's last value use |

`InjectedReferenceDoesNotReviveATypeOnlyImport` pins the limit in the other direction and passes on both sides by design. `prototype_assembled_alias_test.go` was realigned and given a binding assertion, but it assembles the chain in its own body and never calls `EmitWithPluginTransformers`, so it guards that assembly rather than the driver.

Re-exports and export specifiers need no coverage: they go through `IsValueAliasDeclaration`, which resolves the alias from its symbol and never reads a mark.

### What the rounds caught

| round | finding |
|---|---|
| 1 | reversed logic in the `restoreOriginalDeclarationSymbols` comment, an unverified NodeLinks claim, an imprecise idempotency claim, a test doc repeating a claim already removed from the code |
| 2 | the documentation skill had not been read before editing `.mdx`, and its Voice rule forbids em-dashes. Mark monotonicity verified directly |
| 3 | **`module: esnext`, `import * as ns`, and `import x = require(...)` all broke and none was covered.** Three tests added |
| 4 | "marking feeds three predicates" was false. Only `IsReferencedAliasDeclaration` reads a mark |
| 5 | a documentation claim scoped wrongly, and the one assertion in the shipped text that nothing here can verify (`ts-patch` on TypeScript 5 and 6) removed |
| 6 | three of the assertions checked the alias and the binding independently where their siblings correlated them |
| 7 | the chain-construction comment had reached 43 lines with a sentence broken across an edit boundary |
| 8 | evidence attached to `SetParentInChildren` was an experiment about `SetParentInChildrenUnset`, a different function with no production caller here |
| 9 | two comments described their own assertions inaccurately |
| 10 | clean |

Rounds 4 and 8 both corrected claims this branch had itself introduced in an earlier round, and round 3 found the coverage gap two Individual Self-Reviews had not. That is the argument for running the rounds to exhaustion rather than to a fixed count.

### Verification

- Full `packages/ttsc` Go suite green: `test/cli`, `test/driver`, `test/platform`, `test/ttscserver`, `test/utility`, `driver`, `internal/graph`, `internal/graphsymbols`.
- The seven-test count measured by reverting the driver line only, with the prototype test untouched.
- `pnpm format` run and committed on every commit.
- `Close #1204` appears exactly once across the branch, and the fix it closes survives at `HEAD`.

### Could not verify here

The reported timings, which rest on the reporter's measurements, and whether the builtin transformers ever hand a rebuilt container to an `EmitResolver` reference method. The latter is why `restoreOriginalDeclarationSymbols` is kept, and its comment says so rather than claiming the path is observed live.

Issue #1237, opened 2026-08-18
An already-reported empty population is re-reported once per host under singleEvidencePerSymbol


samchon:
## Problem

When a `singleEvidencePerSymbol` reference materializes an empty population, the rule reports the empty population **and then** reports every selected host for citing zero units. The second set of diagnostics asks the author to cite something that does not exist, one message per host.

```
Claim 1 reference 1 (markdown, symbols: h2) matched 1 file(s) but found no
selected evidence units (h2). Select symbol kinds present in those files or
correct the reference globs.

Evidence host TypeScript type 'ISale' at src/sale.ts:1 in Claim 1 reference 1
(markdown, symbols: h2) cites 0 distinct selected evidence unit(s);
singleEvidencePerSymbol requires exactly 1. Split this host so each unit has one
that owns it, or keep positive @evidence citations on this semantic host to
exactly one distinct unit.
```

The first message names the cause and the repair. The second scales with the claim's host count and names a repair the population makes impossible: there is no unit to cite, so no citation can satisfy it.

Whether the second set appears depends on a fact unrelated to the question. If the same reference matches **zero** files instead of one, the per-host findings vanish entirely.

### Reproduction

Against `9609e2a90`, with `project/docs/pricing.md` containing prose but no heading, and one exported interface in `project/src/sale.ts`:

```json
{"claims":[{
  "type":"typescript",
  "files":["src/**/*.ts"],
  "symbol":"type",
  "reference":{
    "type":"markdown","files":["docs/**"],"symbol":"h2",
    "singleEvidencePerSymbol":true
  }
}]}
```

produces both diagnostics quoted above in the same run.

### The split

Two guards decide it, and both test the same irrelevant fact:

- `graph.go:692-695` ends the reference early on an empty population **unless** `single` is set and `len(reference.Paths) != 0`.
- `graph.go:1009-1010` repeats `len(reference.Paths) == 0` before the per-host loop at `graph.go:1011-1021`.

| Policy | Files matched | Units | Per-host findings |
| --- | --- | --- | --- |
| default coverage | any | 0 | none |
| `checklist` | any | 0 | none |
| `singleEvidencePerSymbol` | 0 | 0 | none |
| `singleEvidencePerSymbol` | >= 1 | 0 | one per host |

Every row is already a failing build, because the empty-population diagnostic fires unconditionally in all of them (`graph.go:419-426` for this shape). So the split changes no build outcome. It changes only how much unactionable output the author reads.

[...]

Issue #1236, opened 2026-08-18
A TypeScript claim with an unresolvable root is told its globs are wrong


samchon:
## Problem

A TypeScript claim whose declared `root` does not exist is told its globs are wrong. The rule has already computed the correct explanation and throws it away.

```
Claim 1 matched no typescript files for ['src/**/*.ts'] under root '../nowhere'.
Fix the globs or the root they resolve against; '*' stays within one segment,
'**' crosses segments, and a bare directory is not recursive.
```

The globs are fine. The root is a typo. The author is sent to read glob semantics for a mistake that is one directory name.

The same mistake against a Markdown or Prisma root produces the right message, so the two artifact families disagree about one author error.

### Reproduction

Against `9609e2a90`, using the package's own `runRootedGraph` helper with sources `project/src/sale.ts` and `project/docs/pricing.md`:

```json
{"claims":[{
  "type":"typescript",
  "root":"../nowhere",
  "files":["src/**/*.ts"],
  "symbol":"type",
  "reference":{"type":"markdown","files":["docs/**"],"symbol":"h2"}
}]}
```

produces exactly one diagnostic, the `matched no typescript files` message quoted above. The word `root` appears in it only as part of the phrase telling the author to check their globs.

### Root cause

`base.go:235-246` — `unreadableBaseProblem` builds the correct message for this exact state:

> Evidence graph could not read the typescript root '../nowhere', which resolves to '…'. Create the directory, or correct the 'root' property; it resolves against the ttsc project root.

`graph.go:203` is the only place a TypeScript claim consults it, and it consults it **as a boolean**:

```go
if !populationIsHealthy(inventories, claim.Base, paths) ||
  unreadableBaseProblem(claim.Base, claim.Type) != "" {
  return false
}
```

The string is discarded. Its only effect is to keep the claim active, which then routes to `graph.go:256-266` and prints the glob message instead.

`unreadableBaseProblem` has exactly three callers. Two of them report it: `markdown.go:44-45` and `prisma.go:237`. Neither is reachable from a TypeScript claim, so for TypeScript the message is computed on every cycle and never printed.

### Why the two families diverge

Markdown and Prisma populations are loaded by walking the root. An unreadable root fails that walk, so the loader reports the root problem **and** calls `recordPopulationFailure`, which makes `populationIsHealthy` false and suppresses the glob diagnostic as a derivative.

[...]

Issue #1235, opened 2026-08-18
A claim that selects no host drops its whole obligation with no message


samchon:
## Problem

A claim whose population selects no host is dropped before evaluation and says nothing. The obligation the author wrote disappears, the build passes, and no output distinguishes that from a claim that was fully satisfied.

The project's own documentation flags this twice as something the reader has to remember to check by hand:

> If it were otherwise, a `symbol: "function"` claim over a file of interfaces would select nothing, deactivate, and pass the build with no coverage.
> — `website/src/content/docs/evidence/claims.mdx:106`

> a claim declaring `symbol: "property"` over interfaces whose members are all callables selects no host at all, and a claim with no selected host is inactive, which drops its whole obligation with no message ... read a `property` selector over interface files as the place to check after upgrading
> — `website/src/content/docs/evidence/claims.mdx:116`

An instruction to remember to look is not a diagnostic. This is the same fact as #1234 arriving from the other direction — a population that materialized empty — and it should reach the author the same way, as a warning.

There is a second, louder inconsistency in the same branch. One shape of empty claim population does **not** deactivate and fails the build instead, so today the same condition is either silent or fatal depending on which artifact kind and which root spelling produced it.

### The silent path

`activeGraphConfig` and `claimIsInactive` (`packages/evidence/native/graph.go:161-215`) drop a claim before `materializeClaimStates` runs. The comment records the intent, including the case this issue is about:

> A healthy zero-file match is empty and therefore inactive, including when a typo caused it.

`tests/test-evidence/src/features/test_evidence_graph_ignores_an_empty_rooted_typescript_claim.ts` pins that path end to end: it asserts exit 0 and that neither a population message nor a coverage message appears.

### The fatal path

`graph.go:255-266` reports `matched no <type> files` when an active claim is healthy and matched nothing. A claim reaches that line only when `claimIsInactive` declined to deactivate it, which for a healthy population means `unreadableBaseProblem(claim.Base, claim.Type) != ""` (`graph.go:203-210`).

`base.go:235-246` returns that problem for a declared, non-default `root` that does not stat as a directory. The two artifact families then diverge:

- **Markdown and Prisma** record a population failure for an unreadable root (`markdown.go:44-45`, `prisma.go:141, 237`), so `populationIsHealthy` is false and the message at `graph.go:262` is suppressed in favor of the read error.
[...]

Issue #1234, opened 2026-08-18
An empty evidence reference population fails the build, though it violates no obligation


samchon:
## Problem

`evidence/graph` fails the build when a reference population is healthy and empty. That state is a configuration condition, not a violated obligation: a healthy empty population is a complete denominator, so every coverage question against it is truthfully satisfied. The rule is reporting "this obligation will check nothing", which the author needs to see, but it stops a build over a condition where no evidence rule was broken.

The states that produce it are ordinary and transitional. Documentation not written yet, a Prisma model not added yet, a package area whose declarations are not published yet — each fails the build until the author either writes the artifact or removes the reference from the configuration.

The severity should be a warning: the misconfiguration stays visible, and the build does not break.

## Evidence

Eight sites report a healthy empty reference population, each guarded by a health check so a loader failure takes a different path:

| Site | Condition |
| --- | --- |
| `packages/evidence/native/graph.go:344` | swagger reference matched no source |
| `packages/evidence/native/graph.go:349` | markdown/prisma reference matched no files |
| `packages/evidence/native/graph.go:425` | reference matched files, selected no unit |
| `packages/evidence/native/graph.go:1380` | package-entry reference reached no unit |
| `packages/evidence/native/graph.go:1447` | local typescript reference matched no files |
| `packages/evidence/native/graph.go:1469` | local typescript reference matched files, no unit |
| `packages/evidence/native/graph.go:1525` | package-glob reference matched no files |
| `packages/evidence/native/graph.go:1554` | package-glob reference matched files, no unit |

All eight reach the user through `reportProblems` (`graph.go:1828-1838`), which calls `ctx.Report` and therefore fails the rule.

The diagnostic at `graph.go:1525` states the rationale for reporting at all, and it remains correct — only the severity is wrong:

> an empty population contains no units, so this claim reports full coverage without checking anything.

The condition can only arise while the claim is active, which means at least one host really owes an acknowledgement: `activeGraphConfig` and `claimIsInactive` (`graph.go:161-215`) drop a claim with no selected host before `materializeClaimStates` (`graph.go:216-...`) builds any reference. So this is never a diagnostic about an idle configuration.

Loader failures are a different class and must stay errors. `populationIsHealthy` (`health.go:27-43`) separates "matched nothing" from "could not inspect enough to know what matched", and every site above is already behind that guard.

## Consequence surface

[...]

PR #1238, opened 2026-08-18
fix(evidence): name the real cause of an empty population instead of a derived one

### LINKED ISSUE #1233: A project rule cannot report a warning, so an advisory finding either fails the build or stays invisible
## Problem

A `rule.ProjectRule` contributor has exactly one way to say anything: `ctx.Report(message)`. That call marks the rule failed, and the host then stamps every finding the rule produced with the single severity the user configured for that rule name.

So a project rule cannot report an advisory finding. Any observation it wants to surface — a configuration that will silently do nothing, a population that materialized empty, a degraded input it can still work around — either fails the user's build or stays invisible. The rule's only lever is the user's own `lint.config.ts` severity, which moves every message at once.

`@ttsc/evidence` is the concrete consumer. Its graph rule reports configuration conditions that void an obligation without violating one (see #I2 and #I3), and those must not fail a build while its coverage violations still must.

## Evidence

The reporting contract carries no severity:

- `packages/lint/rule/project.go:36-40` — `ProjectFinding` is `{ Message string }`.
- `packages/lint/rule/project.go:250-256` — `Report` takes only a message and always reaches the reporter.
- `packages/lint/linthost/project_engine.go:68-82` — the host reporter sets `failed = true` on every report.
- `packages/lint/linthost/project_engine.go:126-155` — `finalize()` stamps every finding with `entry.severity`, the one configured value for the rule.

Everything below that point is already per finding:

- `packages/lint/linthost/engine.go:221-241` — `Finding.Severity` is a field of each finding.
- `packages/lint/linthost/compile.go:493-497` — `SeverityWarn` becomes `shimdw.LintCategoryWarning`.
- `packages/ttsc/shim/diagnosticwriter/lint.go:105-127` — that category renders as a warning and `IsError()` returns false, so it never reaches the exit code.

The gap is confined to the contributor contract. No layer beneath it needs a new concept.

## Consequence surface

- **Backward compatibility.** `rule.ProjectReporter` is exported (`packages/lint/rule/project.go:167-170`). Adding a method to that interface breaks any external implementor. The same file already solves this shape once: `NewProjectContext` type-asserts the reporter to the unexported `projectStateSetter` (`rule/project.go:196-198, 211`). A severity channel must follow that optional-interface pattern, and a reporter that does not implement it must keep working.
- **Zero value.** Adding `Severity` to `ProjectFinding` makes the zero value `SeverityOff`, which must not be read as "disabled". Define the zero value as "inherit the rule's configured severity" so every existing construction keeps its meaning.
- **Configured severity is a ceiling.** A rule configured `"warning"` must not be able to emit an error. Resolve each finding as the weaker of its own severity and the rule setting.
- **Rule status.** A warning-only check must leave the rule `passed`, because `ctx.ProjectResult(name).Status` is what later file rules gate on (`packages/lint/README.md`, project-scoped contributor rules section). A rule that only warned did not fail.
- **`ProjectRuleResult.Report`.** The live mutation surface a file rule holds (`rule/project.go:78-82` and its `Report`) shares the same reporter and needs the same treatment for the contract to stay coherent.
- **Editor path.** Project findings carry no `File` and are dropped by `linthost/lsp.go:518` and `linthost/host.go:522`, so they reach users only through `linthost/compile.go:486-507`. That path already handles a nil file.
- **Deduplication.** The host stores messages in a set (`project_engine.go:78-81`). One message reported at two severities must resolve deterministically rather than depend on call order.

## Approach

Extend the project-rule reporting contract with a severity dimension, owned by `packages/lint/rule` and honored by `packages/lint/linthost`:

- a severity-carrying report call on `ProjectContext` and on `ProjectRuleResult`;
- a `Severity` field on `ProjectFinding` whose zero value inherits the rule setting;
- an optional reporter interface the host implements, discovered by type assertion, with the existing `Report` as the fallback;
- a warning report that records a finding without marking the rule failed;
- severity resolution as the weaker of the finding's severity and the configured rule severity.

Do not widen this to file rules. No verified requirement needs it, and their reporting surface is separate.

| Case | Expected |
| --- | --- |
| Project rule warns, rule configured `"error"` | warning diagnostic, rule status `passed`, exit code unchanged |
| Project rule warns, rule configured `"warning"` | warning diagnostic |
| Project rule errors, rule configured `"warning"` | warning diagnostic; the setting caps the finding |
| Project rule errors, rule configured `"error"` | error diagnostic, rule status `failed` |
| Rule emits both a warning and an error | both render, rule status `failed` |
| Same message at both severities | deterministic single resolution |
| Reporter without the optional interface | falls back to `Report`, no panic |
| Existing `ctx.Report` caller | unchanged behavior |
| Warning-only rule read by a later file rule | `ProjectResult` status `passed` |

Prove with Go unit tests under `packages/lint/test/`, one `Test*` per file, mirroring the existing project-rule test peers. Narrow command first, then `pnpm test:go`.

## Coordination

Blocks the two `@ttsc/evidence` issues in this campaign. Public contributor contract, so `packages/lint/README.md` (project-scoped contributor rules) and the website lint development page must change in the same pull request.

samchon:
## Intent

One cycle of the evidence population-diagnostic campaign. Both issues are about
what the graph rule *says* when a population comes back empty, not about whether
it fails: every case below keeps the build outcome it has today.

- #1236 — a TypeScript claim or reference with an unresolvable `root` is told its
  globs are wrong. `unreadableBaseProblem` computes the correct explanation and
  `graph.go:203` uses it as a boolean and discards the string. Markdown and
  Prisma both report it and suppress the glob diagnostic as a derivative;
  TypeScript does neither.
- #1237 — a `singleEvidencePerSymbol` reference re-reports an already-reported
  empty population once per selected host, asking each one to cite a unit that
  does not exist, and only when at least one file matched.

## Scope

`packages/evidence` only. No severity change: an empty population remains a
build error, which is the decision that closed #1233, #1234, and #1235.

Documentation moves with the code where it states the current behavior —
`.agents/skills/project/evidence/SKILL.md` records the `checklist` versus
`singleEvidencePerSymbol` contrast that #1237 removes.

## Verification

Pending. This body is the claim; it was written before any code existed.

## Deferred

Not in this cycle, recorded so it is not read as resolved: `hints.go:68-80`
states that the hint corpus is withheld whenever the graph rule reports
anything, calls that an upstream limitation, and names the fix. That remains
open and is outside this campaign's declared scope.


samchon (commented):
**Individual Self-Review — `b80196d45`** (resolves #1236)

Names an unresolvable TypeScript claim root as a root instead of blaming its globs. `unreadableBaseProblem` was consulted at the activation gate as a boolean and its string discarded; `typeScriptBaseProblems` now reports it and records a population failure on the claim-side pass, so health alone decides activation and the claim-side empty-match diagnostic — whose only live caller was this bug — is removed.

**No product defect found.** The reviewer could construct no reachable path where the deleted diagnostic was load-bearing, and verified the three-kind health argument end to end: markdown records before listing, prisma records before listing, typescript now records via the new pass, and all three run before `activeGraphConfig`. It also proved the synthetic `\x00load-failure:` key cannot leak into a TypeScript path through `relativeOf`, the loader, or the governed-file scan.

Six findings, five accepted and fixed in `44cd87e44`:

| # | Finding | Disposition |
| --- | --- | --- |
| 1 | `TestAReferenceMatchingNothingStillNamesItsGlobs` asserted absence against a **default** base, which returns before the stat and which `Check` has already validated — the assertion could not fail | Accepted. Rewritten as `TestAResolvableRootStaysQuietWhileItsReferenceSpeaks`, rooted above the project so the pass genuinely evaluates a declared root |
| 2 | `base.go` doc comment said TypeScript differs "in the repair clause alone"; the verb differs too | Accepted |
| 3 | The new message used "resolve" three times across two meanings | Accepted. Lead clause is now "found no directory at", leaving "resolves to" and "re-bases against" as the only path senses |
| 4 | `"typescript"` hardcoded where the sibling branch composes `string(kind)` | Accepted |
| 5 | `claimIsInactive`'s unreachable `default` arm is now the only route to an active, healthy, zero-path claim, and the deleted block was its net | Accepted as documentation. The arm stays and now states what a fourth claim-capable kind owes. No behavior change: the arm is unreachable because `decodeClaim` refuses a reference-only kind and any kind problem returns from `Check` first |
| 6 | `TestUnreadableMarkdownClaimRootDoesNotBecomeInactive` hand-recorded the failure, so its root string was decorative | Accepted. It now runs the real loader, as its TypeScript twin already did |

Finding 5 is the only one carrying a judgment call. The reviewer offered `return true` or a panic; both invent behavior for a state that cannot occur, so the arm keeps its semantics and gains the maintenance obligation in words.

samchon (commented):
**Individual Self-Review — `c218b0025`** (resolves #1237)

Stops `singleEvidencePerSymbol` re-reporting an already-reported empty population once per host. The empty-population early return now applies to every policy, and the `len(reference.Paths) == 0` clause that made the same empty population judged or skipped on an irrelevant fact is gone from both guards.

**The code change is correct, and the reviewer verified the commit's central safety claim rather than accepting it.** Every `problems = append` in the skipped body sits inside the declaration loop whose first act is `if len(covered) == 0 { continue }`, and `UnitsByScope` is populated only from `state.Units` on all four materializer paths — so an empty population already skipped every declaration it visited. Nothing reads `Hidden`, `Published`, or `Population` outside that loop.

Two results worth recording because they were measured, not argued:

- **`uncertain` is load-bearing and unaffected.** Both blocks that write it sit above the early return. The reviewer confirmed the map matters by widening the guard to `|| single`, which produced four "Non-participating @evidence" ghost findings in unrelated tests; the guard as written does not reach that.
- **The dropped path-count clause was already dead** at the second guard under the old code, since reaching it with zero units required `single && Healthy && len(Paths) > 0`.
- **The replaced test was verified counterfactually.** Under `go test -overlay` against the parent's `graph.go`, the new empty-population case fails on the old code, and the populated negative twin fails when the guard is widened too far. No build that used to fail now passes: the empty population is still an error.

Three findings, all accepted and fixed in `44cd87e44`. All three are sentences that outlived the rule they described — the same failure class this cycle exists to remove:

| # | Finding | Disposition |
| --- | --- | --- |
| 1 | `website/.../claims.mdx` still drew the checklist-versus-`singleEvidencePerSymbol` contrast thirteen lines below the bullet this commit had just corrected | Accepted |
| 2 | `checklist_binds_every_host_to_every_item_test.go` gave the deleted rule as the reason for its own existence | Accepted. It now says the checklist reaches the shared outcome through its own host loop, which is why it is asserted separately |
| 3 | `ITtscEvidenceGraphReferenceBase.ts` TSDoc — what an adopter sees on hover — still said a host with no tag fails, without the bound the website bullet had just been given | Accepted |

samchon (commented):
**Individual Self-Review — `e09bba680`**

Adds two end-to-end feature cases proving both diagnostics through the real `ttsc` binary, and replaces an assertion in the rooted-claim case that Overall Self-Review had found vacuous.

**Both new tests discriminate against a revert, verified by tracing the pre-change code rather than by assumption.**

- The root case: `'*' stays within one segment` is the exact and only source of that substring, and it exists nowhere in the tree today — the surviving reference-side twin ends differently — so the exclusion fires on a revert. `../absent` matches verbatim through `projectRelativeDisplay`, and `createProject` mkdtemps a fresh workspace, so the sibling is reliably absent with no cross-test collision.
- The empty-population case: under the old guard, `single` with both populations healthy and one matched path fell through to the host loop and emitted the per-host count. That exclusion is the discriminating assertion; `assertFailure` and `assertIncludes` are regression guards that correctly held before and after, since the commit did not change whether the population message is emitted.

The reviewer also confirmed the fixture is not vacuous in the way that would matter most: the claim really is active with one host, and the reference really has one matched path and zero units.

One finding, accepted:

| # | Finding | Disposition |
| --- | --- | --- |
| 1 | The replacement `assertExcludes` in `test_evidence_graph_ignores_an_empty_rooted_typescript_claim.ts` sits behind `assertStatus(result, 0, …)`, which throws first on any diagnostic, so it can never be the failing assertion | Accepted, and removed in `44cd87e44` rather than reworded |

The reviewer judged no repair required, noting the pattern is inherited from the untouched assertion two lines below. That is right about the file and wrong about this commit: `e09bba680` claimed in its own message to have replaced a vacuous assertion with a discriminating one, and it had not. Two unit cases already pin that an existing root reports nothing, so the assertion is removed rather than kept as decoration. The pre-existing `Missing acknowledgement` exclusion is left alone; it is not this cycle's to churn.

samchon (commented):
**Individual Self-Review — `44cd87e44`**

The adjudication commit for the three reviews above. Three findings, all accepted and fixed in `08cdf2825`. One is a real defect that commit introduced.

| # | Finding | Disposition |
| --- | --- | --- |
| 1 | The rewritten repair clause merged two facts and inverted one: "a typescript root re-bases Program sources **against the ttsc project root**". Sources re-base onto the **declared root**; only the root spelling resolves against the project root | Accepted. Real defect, introduced by `44cd87e44` |
| 2 | The rewritten root test's step 3 promised an assertion on the reference's globs that the body did not make | Accepted. It now asserts the population text rather than a prefix of it |
| 3 | The removal of an unfailable `assertExcludes` applied a rule that does not distinguish it from its surviving sibling or ~30 other feature files, and cost the only end-to-end pin | Accepted. Restored |

**Finding 1 is the one worth reading.** `populationBase.addressOf` and `relativeOf` key every TypeScript unit against `base.Absolute`, the declared root; `resolvePopulationBase` uses the project root once, to resolve the root itself. The sentence as shipped stated the reverse, and would have read to an author as though declaring `root` changed nothing — while contradicting the absolute path the same sentence had just printed. It came from a wording fix, which is exactly the kind of change that gets read for style and not for truth.

Nothing caught it because every assertion on this message stops at `which resolves to '`. That is a reasonable place to stop for a verb change and not for a clause making a claim about behavior. The test was widened accordingly; the message's tail is still not asserted verbatim, which is deliberate — pinning prose that has no behavioral consequence would expire on every edit.

**Finding 3 reverses my own earlier adjudication.** The third review reported that assertion as unfailable and judged no repair required; I overrode it and removed the assertion. This review shows the rule I applied does not single it out — the assertion two lines below and roughly thirty other feature files have the same shape — so what I removed was the convention, not a defect, and with it the only pin that a resolvable root stays quiet across the native boundary. The two unit cases I cited as replacements run in process and never cross it, which is the reason this suite exists at all. Restored with the current wording.

Verified clean by this review: no stale copy of the old message anywhere in the tracked trees; `found no directory at` stays true in the path-exists-but-is-a-file branch; the rewritten test's base is genuinely non-default so the pass really stats it, and the claim is genuinely active; `root: ".."` has no consequence for address display, governed-file recording, or Windows prefix matching that the test does not account for; all four documentation edits are accurate and no surrounding sentence contradicts them; build, vet, prettier clean, and `go test ./native` fails only the 24 environmental bridge cases.

samchon (commented):
**Individual Self-Review — `08cdf2825`**

The correction of the inverted repair clause. **No blocking finding.**

Verified rather than accepted, on the sentence this commit exists to make true:

- **Sources really do re-base onto the declared root.** `extendTypeScriptInventories` keys every source by `relativeProjectPath(base.Absolute, …)` and then `base.addressOf(relative)`, and `address()` composes `base.Absolute + "\x00" + relative`. The claim in the message is the claim in the code.
- **The root spelling really does resolve against the project root**, for a relative root and for a `..`-ascending one.
- **"Rather than scanning the filesystem" is accurate.** The only `WalkDir` sites in the package are markdown, prisma, and the package-glob reference loader. The TypeScript path walks nothing, and `unreadableBaseProblem`'s own `os.Stat` stats the root rather than scanning it.
- The restored end-to-end assertion matches the current head clause. Restoring the *old* text verbatim would have produced a dead assertion, since the wording had moved under it.
- The widened test assertion renders exactly: `describePopulation` on a default base returns bare `describePatterns`, giving `['docs/**']`.

One low-severity observation, **accepted as a known limitation and not repaired**:

> `it resolves against the ttsc project root` is false for a declared root on a different Windows volume. `resolvePopulationBase` does not join an absolute declared root to the project root, and `projectRelativeDisplay` falls back to the absolute spelling when `filepath.Rel` errors, so the message can print a root that resolves against nothing.

The identical clause is pre-existing in the Markdown and Prisma branch two lines below, and this cycle's declared scope is cause attribution for population diagnostics, not root-label rendering. Correcting one branch would trade a narrow inaccuracy for a branch asymmetry, which is the shape this cycle spent two issues removing. Recorded here so a later change fixes both together; every in-volume absolute root is unaffected, because the label printed is the derived project-relative spelling and that one does resolve against the project root.

Also checked and clean: no stale copy of any wording of this message anywhere in the tracked trees; the doc comment carries no false, dead, or contradicting sentence after growing across three commits; no behavior changed beyond message text, since every caller treats the value as opaque text plus an emptiness test; build, vet, prettier clean; `go test ./native` fails only the 24 environmental bridge cases.

samchon (commented):
**Overall Self-Review — rounds 1 through 3**, solo, over the complete `master`-to-head diff.

**Round 1** found what the per-commit reviews structurally could not, because it was the first pass to read the integrated diff rather than one commit:

- **`tests/test-evidence-benchmark` reads the graph's output by substring**, and one of the two substrings it looks for no longer exists on the claim side. `readEmptyPopulationReports` filters lines containing `matched no `, and `assertClaimActivated` uses that list as the first of its two halves. No coverage was silently lost — a claim that selects nothing now arrives at the second half, where a claim owing no obligation is already the failure — but both helpers described a claim-side report as the loud half, and the error they raise said the claim had selected an empty population when only a reference can now. Prose corrected, assertions untouched, and the reason the claim side lands on the second half written down.

  The earlier reviews swept that directory for anything **documenting** the message and found nothing. Code that **parses** it is a different question, and it is the one a whole-diff round asks.

- **Two clauses of the root diagnostic still overreached.** Its headline called the condition unreadable when the test is a stat and always was. Its repair asked for a directory, which this stat is satisfied by while empty — silencing the diagnostic and leaving the population exactly as empty. It now asks for the sources too.

**Round 2** found one item: the new helper's summary still framed the condition as a root that does not *resolve*, the vocabulary the user-facing message had been corrected away from precisely because it collided with path composition. Fixed.

**Round 3 is clean.** Complete fresh pass over every changed file and hunk, and over the consequence surface: the two issues do not interact (a claim with a failed base has no paths and is skipped before its references materialize); the hint-corpus gate and `evidence/documented`'s status read are unchanged, because every case that used to report still reports; the synthetic population-failure key cannot leak into a TypeScript path, the loader, the governed-file scan, or the corpus; two claims sharing one broken root produce one diagnostic and both deactivate cleanly; `claimPopulationConfig` narrows to claims of one kind with references dropped, so no reference base is judged before activation; `describePopulation` and every other helper touched remain used; watch inputs are declared from configuration before a Program exists and are unaffected.

Verifications that could not run locally: the feature suites, because this worktree has no `node_modules` of its own. `go test ./native` fails only the 24 bridge cases that need that link, and the `evidence defenses` CI lane is what proves the two new end-to-end cases.

samchon (commented):
**Individual Self-Review — `b6fed6779`** (the first Overall Self-Review round's findings)

**No product defect.** The reviewer independently traced the commit's central claim — that no coverage was lost in the benchmark defense suite when the claim-side empty-match diagnostic disappeared — through all six states, and confirmed each is still caught:

| State | Caught by |
| --- | --- |
| claim glob selects no file | second half: inactive, so zero obligations |
| claim's declared root does not exist | second half: the failure is recorded before activation, the claim stays active but unhealthy, and the missing-acknowledgement producer sits behind the health guard |
| claim selects files but no host of its kinds | second half |
| reference matched no files | first half; all four surviving producers verified reference-side and label-carrying |
| reference matched files, zero units | second half — that message says "found no selected evidence units" and does not contain the filtered substring |
| satisfied claim | neither, correctly; both callers materialize deliberately untagged layers |

It also recorded one pre-existing partial gap, unchanged by this branch and flagged only because the question was asked: a claim with one dead reference and one healthy one still reports obligations, so neither half notices the dead one.

Three findings, all accepted and fixed in `58928a8c0`:

| # | Finding | Disposition |
| --- | --- | --- |
| 1 | The rewritten helper doc still said its list catches a gate opening onto an **unreadable** population — the one state that list cannot contain, since the graph reports a read failure at its own cause and withholds the derived line | Accepted. That is the exact distinction the commit was enforcing three lines above |
| 2 | The Markdown and Prisma branch still carried the empty-directory hole just removed from the TypeScript branch, and the new doc paragraph attributed the two-step repair to TypeScript never scanning a directory — an attribution that does not carry, since the stat is shared | Accepted. Both branches now ask for what the directory must hold |
| 3 | "make its sources part of the project" used "the project" in the Program sense eleven words before the same sentence used it as a directory | Accepted |

Finding 2 is the one worth naming. Fixing one branch of a shared predicate and calling the difference a property of that kind is the shape this cycle spent two issues removing, and it was reintroduced by a commit whose subject was removing it. The reviewer marked it PLAUSIBLE — the asymmetry is real, whether it merits an error-message edit is a judgment — and the judgment went to repair, because the asymmetry is exactly what the campaign is about.

samchon (commented):
**Individual Self-Review — `58928a8c0`** (extending the two-step repair to every branch)

**No product defect.** Verified: an empty directory really does satisfy the shared stat on every branch and produce no diagnostic; the resulting healthy empty claim really does deactivate in total silence; and "sources" is the codebase's own noun for both Markdown documents and Prisma schema files, so the new clause names the right thing for each kind. Swagger never reaches the predicate at all, so the two branches cover the whole caller set.

Four findings. Three accepted and fixed in `3be2e92b3`; one recorded.

| # | Finding | Disposition |
| --- | --- | --- |
| 1 | "compiled Program" appears nowhere else in the repository; every other site says the Program, and the skill names the requirement as an explicit root of the active tsconfig Program — the file the author edits. "Compiled" also invites the emit reading in a compiler's repository | Accepted → "tsconfig Program" |
| 2 | The doc paragraph said the walkers want *documents* while the string it describes says sources on both branches, and Prisma's own loader calls its schema files sources. The real split is disk against Program | Accepted |
| 3 | Lifting the two-step rationale into its own paragraph orphaned the "also" in the next one, which chains back across an intervening topic | Accepted; the paragraph moved below both verb paragraphs |
| 4 | "create that directory" is unachievable when a **file** occupies the path, which this predicate also fires on | Recorded, not repaired |

Finding 4 is left alone deliberately. The gap is pre-existing — the old wording had it identically — and covering it needs a third branch to say something the "Correct the 'root' property" alternative already covers for the author's real move. Recorded here so a later cycle does not rediscover it as new.

The reviewer also observed, outside the commit, that the predicate was still named for what its first two callers do with the answer while the comment two commits earlier had told the next reader to say what is measured. That is now `missingBaseDirectoryProblem`. The name fires for a path that does not exist and for one occupied by a file, and in both there is no directory at that base, so it covers each — which is the same reading the message's own "found no directory at" takes.

samchon (commented):
**Individual Self-Review — `3be2e92b3`** (renaming the base predicate)

**Clean commit.** The reviewer verified the rename is complete and safe — no occurrence of the old identifier survives in any tracked file, no collision with the neighbouring `unreadable*` family, no `sed` substitution inside a literal or a longer word — that the non-comment diff is exactly six lines with no control flow touched, and that every assertion of the affected message stops upstream of the change. It also confirmed the `also` chain in the doc comment reads correctly after the paragraph move.

Six low findings. Three accepted and fixed in `a55e07558`, one recorded in the code, two declined.

| # | Finding | Disposition |
| --- | --- | --- |
| 2 | The naming paragraph explains `unreadableBaseProblem`, a symbol no reader of this tree can grep for; one more rename would lay a third stratum of the same explanation | Accepted, trimmed to what still binds |
| 3 | The "could not read" exemption was granted because "that caller does read" — but the predicate returns that string **before** anything reads | Accepted. The read it names is the one a caller is about to attempt, and the sentence now says so |
| 5 | A width rewrap dropped a load-bearing "declared" from a test's rationale, turning a default-base-against-declared-root contrast into one about existence alone | Accepted |
| 4 | `os.Stat` also fails for an unreachable parent, where the directory exists; "missing" and "found no directory at" both narrow the condition | Recorded in the comment, message unchanged |
| 1 | The commit message's "compiled Program appears nowhere else in this repository" is false — it appears six times outside `packages/evidence` | Declined for code; the claim is wrong and the change is not |
| 6 | The published pull-request body still names the deleted symbol | Declined |

**Finding 4** is the one with a judgment behind it. Making the message true for the unreachable-parent case means saying "no **readable** directory", which puts an access implication back into a message whose predicate performs no access check — the exact overreach the previous commit removed from the headline. The behavior is right as it stands, because treating an unreachable root as healthy emptiness would deactivate the claim in silence, so the comment records that any stat failure counts as absence and the message keeps its plain wording.

**Finding 6** is declined because the body is the intent statement written at open, and it describes the state the cycle set out to change. `unreadableBaseProblem` existed when the problem existed. The rename is recorded here instead, which is where the pull-request skill puts everything that happened after the body was written.

**Finding 1 is a correction against this campaign's own record.** The claim was made in a commit message and repeated in the campaign report. The defensible version is narrower: `packages/evidence` says "the Program" in nineteen places and the skill says "active tsconfig Program", which is the vocabulary the message now matches.

samchon (commented):
**Individual Self-Review — `a55e07558`**

Reviewed under a raised bar: false statements, unfalsifiable assertions, behavior defects, and self-contradiction only; wording preferences explicitly out of scope. **One finding, already fixed in `d1be317fb`.**

| # | Finding | Disposition |
| --- | --- | --- |
| 1 | "two of its three callers read nothing" — one does. Markdown and Prisma both walk the root immediately after the predicate clears it; only the TypeScript caller materializes without touching the filesystem | Accepted, fixed |

The reviewer added the coupling that makes this an inconsistency rather than arithmetic: the "could not read" wording is **not Markdown-only**. The fallback branch serves every non-TypeScript kind, so Prisma receives it too — and if Prisma were one of the callers that read nothing, the exemption's own justification would collapse for a caller that receives the message. It also noted the count contradicted two other sentences in the same file, including the surviving paragraph that says TypeScript "never scans a directory, so a message about reading one would name an access **this artifact kind** never attempts" — singular by construction.

I found the same defect independently while verifying the sentence against the call sites rather than against the sentence it replaced, and the fix names both Markdown and Prisma for exactly the reason the reviewer gives.

**Nothing else above the bar.** Verified sound by this review:

- "any failure of the stat counts as absence" holds: the only early return is `err == nil && info.IsDir()`, so every error value and every non-directory success falls through identically. An unreachable parent does reach that stat — `normalizeRootPath` is pure string validation and `resolvePopulationBase` only joins and cleans, so no filesystem call precedes it.
- The replaced justification was wrong and the new one is right: the predicate is called immediately before each walker's `filepath.WalkDir`, and a non-empty return abandons that walk. The message names an imminent, gated read rather than one that happened.
- The restored test rationale matches its body clause by clause, and its absence assertion can fail — the declared root `".."` differs from the project root, so it is not folded into the default base, and the pass genuinely stats it.
- No stale reference to the former symbol survives in source or tests.

This is the third time in this cycle that a commit repairing a precision defect introduced a new one, and all three shared a cause: the replacement sentence was checked against the sentence it replaced instead of against the code. The two things that actually caught them were rendering the message and counting the callers from the call sites.

samchon (commented):
**Individual Self-Review — `d1be317fb`** (the caller-count correction)

**No findings above the bar.** The review loop for this cycle has converged.

Verified against the code rather than against the sentence it replaced, which is the check all three of this cycle's self-inflicted defects skipped:

- **"The Markdown and Prisma messages"** names the complete real set. The predicate branches only on TypeScript, so the "could not read" string serves every other kind — and the only other kinds that can reach it are Markdown and Prisma, because `decodeRoot` refuses `root` on a Swagger reference and `decodeArtifactKind` refuses Swagger as a claim, so a Swagger base never reaches the predicate at all.
- **"the walk each of those callers is about to run"** holds on both paths: each calls the predicate immediately before its own `filepath.WalkDir` and abandons that walk on a non-empty return.
- **"the third caller reads nothing"** is now correct. Exactly three call sites; the TypeScript one reaches only `configuredBases`, the predicate, and a pure map write, and its population materializes with no `os.` call anywhere in `typescript.go` or `typescript_cache.go`.
- **"any failure of it counts as absence"** holds: the sole early return is `err == nil && info.IsDir()`, so every error path and the not-a-directory path fall through identically, and both named cases reach it.
- **No contradiction** between any sentence of the comment and either string the function returns.
- Build and vet clean; `go test ./native` fails only the 24 environmental bridge cases.

Ten Individual Self-Reviews across this cycle, one per pushed commit. **None found a product defect.** Every accepted finding was a false sentence, an assertion that could not fail, or a message clause — the right outcome for two issues that were both about what the rule says rather than what it decides.

Issue #1245, opened 2026-08-19
fix(unplugin): the transform cache never hits, so every module recompiles the whole project


samchon:
## Problem

Every bundler module delivery recompiles the whole project. `@ttsc/unplugin`'s transform cache never hits in any project that imports an npm package whose types resolve through a `.d.ts`, which is every real application. A build's cost is therefore `modules x whole-project compile` instead of one compile, and that is the residual stall reported in #970 (it survives #980, #1009, #1017, #1224 and #1228).

Current behaviour, measured on `master` (`1250f0c70`, 0.28.0), vite 8.2.1 + typia 14.0.0 + typescript 7.0.2, an 11-module project:

| lifecycle | modules delivered | whole-project compiles |
| --- | ---: | ---: |
| `vite` serve (the mode `vitest` runs in) | 11 | **11** |
| `vite build` | 11 | **11** |

Expected: **1** compile per build, the invariant `experimental/unplugin-perf` claims to guard.

Affected users: every `@ttsc/unplugin` consumer (all adapters) and `@ttsc/metro`, on every platform. The larger the project, the worse: the reporter's `react-router` application never completes.

### Reproduction

A project of 8 modules, 5 chained type files, 3 `typia.createValidate` calls, a `~/*` tsconfig+vite alias, `typia@14.0.0` installed. Drive the real `@ttsc/unplugin/vite` plugin object over the modules with `vite.createServer(...).transformRequest(id)` (what `vite-node`/`vitest` does) and count `TtscCompiler.prototype.transform` calls. One compile is expected; one compile per delivered module is observed.

### Root cause

Instrumented run of `matchesCachedSource` / `transformProject`:

```
generation walk {beforeComplete:true, afterComplete:true, sameHashes:true, sameDirectories:true}
external snapshot incomplete {count:11, total:1185}
graph proofs: member/proof mismatch {conflicts:[], proofs:123, members:134}
generation {stable:false, graphProofs:false, externalComplete:false}
validate {snapshotComplete:false} -> "complete: no stable generation snapshot"
```

The 11 unproven graph members are all host-owned resolution candidates:

```
node_modules/typia/lib/index.ts
node_modules/typia/lib/index.tsx
node_modules/@standard-schema/spec/dist/index.ts
node_modules/@standard-schema/spec/dist/index.tsx
node_modules/node_modules/@typia/interface/package.json
node_modules/typia/node_modules/@standard-schema/spec/package.json
src/node_modules/typia/package.json
...
```

[...]

Issue #1246, opened 2026-08-19
fix(unplugin): a cached module delivery still costs O(project) filesystem work


samchon:
## Problem

Even when the transform cache hits, `@ttsc/unplugin` spends O(project) filesystem work on **every** module delivery, so a bundler run is still `modules x project` rather than `modules + project`. On an 11-module toy project each cached delivery costs 2.5-3.5 s; the work per delivery does not shrink as the project's own module count grows, so a real application pays minutes of pure revalidation on top of its single compile.

This is the second half of #970. With the generation-reuse defect fixed (see #1245), the reporter's suite would still be dominated by this.

Expected: a cached delivery costs work proportional to the delivered file's own inputs, not to the whole project's input universe, and the numbers stay flat as modules are delivered.

## Evidence

Fixture: vite 8.2.1 + typia 14.0.0 + typescript 7.0.2, `~/*` alias, 60 application modules / 30 chained type files / 20 `typia.createValidate` calls, driven through `vite.createServer(...).transformRequest(id)` with one whole-project compile (generation reuse already repaired locally). Process-wide `node:fs` counters, per delivery:

```
[ 51] /src/mod-46.ts 2304ms {statSync:11, lstatSync:1217, readFileSync:11,
                             readdirSync:1089, existsSync:1481, realpath:4281, bytes:0}
[ 52] /src/mod-47.ts 2675ms {...identical...}
[ 53] /src/mod-48.ts 2798ms {...identical...}
```

Every delivery repeats ~1,089 `readdir`, ~4,281 `realpath`, ~1,217 `lstat` and ~1,481 `existsSync`. Content reads are already 0, so the earlier read-bounding fixes (#1224, #1228) hold; the metadata and path-identity work is what was never bounded.

A V8 CPU profile of the run attributes the non-compile time as:

```
23.2%  notifyWatchInputs
         17.2%  ... viteServe.watch -> pathIdentityKey -> cachedRealpath
         11.6%  ... viteServe.watch -> pathIdentityKey -> cachedCaseSensitivity  (readdir)
          4.1%  ... adapter addWatchFile -> fs.existsSync
12.5%  filesystem readdir (path-identity case probing)
 6.6%  pathIdentityKey
```

### Cause 1 - watch-input notification re-resolves path identity per delivery with a throwaway context

[...]

PR #1247, opened 2026-08-19
fix(unplugin): serve one project compile to every module again

samchon:
This pull request owns the complete accepted cycle of the `unplugin-generation-reuse` campaign: #1245 and #1246. Both are the residual samchon/ttsc#970 reproduction — a bundler build that never completes with typia >= 13.1.19.

## What was actually wrong

The transform cache **never hit**. Not "hit expensively" — never. Every delivered module recompiled the whole project, in `vite build` and in `vite` serve (the mode `vitest` runs in) alike.

`driver.SupersedingModuleCandidates` enumerates superseding resolution candidates speculatively, through raw `os.Stat`/`os.ReadFile` outside `inputObservationFS`, so `attachInputProof` records nothing for them. That is correct: a candidate is a spelling strictly ahead of the target that won, so the compiler never selected it and usually never read it. `@ttsc/unplugin` nevertheless required `graph.inputProofs.size === graph.members.size` before it would call a generation stable, and `graph.members` contains every candidate. `projectSnapshotComplete` was therefore permanently `false` for any project that resolves a dependency through a declaration file — which is every real application. That flag closes the build-scoped shortcut, the narrow persistent path, and complete-snapshot validation at once, each refusal evicts the generation, and the next module recompiles the project into another unprovable generation.

Measured on the reproduction (vite 8.2.1, vitest 4.1.11, typia 14.0.0, typescript 7.0.2; 60 modules, 30 chained type files, 20 validators, 6 vitest suites):

| workload | `master` | this branch |
| --- | ---: | ---: |
| `vitest --run` wall | **609.96 s** | **53.96 s** |
| `vitest --run` transform (summed) | 3652 s | 317 s |
| `vite` serve, 69 deliveries | 69 compiles | **1 compile** |
| serve, per cached delivery | 2.5–3.5 s | ~0.10 s |
| serve, per delivery `readdir` / `realpath` | 1089 / 4281 | 16 / 3 |
| `vite build`, 11 deliveries | 11 compiles | **1 compile** |

## #1245 — generation reuse

- Classify graph members. A member the envelope reports only under `graph.candidates` is *speculative*: it carries the state recorded when the envelope was produced, the same evidence a plugin-declared dependency path carries, and its appearance still invalidates the generation. A member that also carries an edge, a global, or a config entry is *realized* and keeps the compiler-proof requirement.
- The absence of a proof no longer voids a whole generation; a contradictory proof still does.
- A generation that still cannot be reused reports itself once on stderr with the evidence it lacked. This condition costs one whole-project compile per module, and it spent the entire life of #970 invisible.

## #1246 — per-delivery cost

- Generation stability and complete-snapshot validation compare the envelope's **declared inputs**, not every file under the project root. A project root is a working directory: generated types, logs, and coverage output are written there while a compile runs, and comparing them declared the generation incoherent — the same recompile-per-module cliff through a different door. Files entering or leaving the project stay covered by the directory-membership snapshot, and a graph-free envelope keeps whole-walk comparison.
- A walk failure is scoped the same way: a file nothing compiled failing its own read sandwich no longer taints the snapshot, while an unattributable failure and an incomplete directory walk still do.
- The transform hands the adapter the identity and existence it already memoized for each watch input, so the adapter no longer pays a `realpath`, a case-sensitivity directory listing, and an `existsSync` per input per delivered module.
- A dev server configured without a watcher (`server.watch: null`, which is exactly what `vitest --run` sets) receives no watch registration: nothing can deliver a change event, while Vite's import analysis resolves every registered path like a real import of the transformed module, once per module. The adapter's own missing-input poll is untouched.

## Why this recurred four times

`experimental/unplugin-perf` and the whole `tests/test-unplugin` cache suite drive a hand-written Go sidecar (`example.com/ttscunpluginfixture`, a regex rewriter with its own `go.mod`) that fabricates the envelope. It stamped proofs for edges, globals, and configs, and never emitted a `candidates` section at all, so the one shape that fails was unreachable from the suite. The repository's only fixture with real candidates (`createLinkedWorkspaceFixture`) asserts behaviour, never a compile count, so a permanently dead cache passed it.

This pull request closes that gap: the fixture now emits candidates the way the real host does — without proofs — and new cases assert the compile count against that shape.

## Tests

| case | pins |
| --- | --- |
| `caches_one_compile_with_unproven_resolution_candidates` | #1245, the failing shape |
| `invalidates_the_generation_when_a_candidate_appears` | negative twin: recorded `missing` is state |
| `refuses_reuse_for_an_unproven_realized_graph_input` | boundary: realized members keep the proof requirement |
| `keeps_the_generation_when_a_non_input_is_written_during_a_compile` | #1246 |
| `vite_serve_registers_no_added_imports_without_a_watcher` | #1246, the serve skip |
| `vite_serve_keeps_existing_watch_inputs_in_the_added_import_graph` | rewritten to run with a real watcher, where the contract means something |

Documentation: the protocol page's `inputHashes` contract and the unplugin page's cache/watch section.


packages/unplugin/src/core/transform.ts
```diff
@@ -818,9 +856,15 @@ function envelopeGraphIndexes(
...
       for (const candidate of candidates) {
         if (typeof candidate !== "string" || candidate.length === 0) continue;
-        built.members.add(
-          derivationIdentity(state, path.resolve(props.projectRoot, candidate)),
+        const identity = derivationIdentity(
+          state,
+          path.resolve(props.projectRoot, candidate),
         );
+        // Edges, globals, and configs are folded in above, so a path still
+        // absent from `members` here is one the envelope reported only as a
```
samchon: Classification depends on iteration order across `graph.candidates` entries. A path that entry A lists as a candidate and entry B lists as its *source* is marked speculative when A is visited first, because B's `built.members.add(sourceIdentity)` runs later. A candidate source is an importing file the compiler read, so it must be realized regardless of order.

An envelope whose `edges` already names every source file hides this, which is every envelope the real host produces — but the classification should not rest on that. Fold every candidate source in first, snapshot the realized set, then classify.

packages/unplugin/src/core/transform.ts
```diff
@@ -3333,6 +3439,147 @@ function sameHashes(
...
+        path.resolve(props.projectRoot, entry),
+        props.identities,
+      ),
+    );
+  };
+  for (const [source, targets] of Object.entries(graph.edges ?? {})) {
+    add(source);
+    if (Array.isArray(targets)) for (const target of targets) add(target);
+  }
+  for (const input of graph.globals ?? []) add(input);
```
samchon: `graph.globals`, `graph.configs`, and `result.hostInputs` are iterated with `?? []` only, while every other reader of the same sections guards with `Array.isArray` (see `selectListedFiles`). A malformed envelope that sends an object here throws inside generation capture instead of degrading, which is a worse failure than the one this function exists to prevent.

samchon (commented):
**Individual Self-Review — `06d72d14f` "fix(unplugin): serve one project compile to every module again"**

This commit resolves #1245 and #1246. It classifies graph members into realized and speculative, stops an absent proof from voiding a generation, scopes generation stability and complete-snapshot validation to the envelope's declared inputs, hands the adapter the identity and existence the generation already memoized, skips watch registration on a watcherless dev server, reports an unreusable generation once on stderr, and teaches the fixture sidecar to emit `graph.candidates` without proofs so the failing shape is reachable from the suite.

Two candidate findings, both accepted and corrected in the follow-up commit; the rest of the diff verified clean this round.

Local evidence at this commit: `test-unplugin` 131/131, `test-metro` green, `vite build` 69 modules with one compile, `vitest --run` 609.96 s → 53.96 s.

This review is recorded by the main agent: subagents are disabled for this session by an explicit operator rule, so the mandatory per-commit advisory pass was performed inline rather than delegated. The deviation is recorded in the campaign ledger.

samchon (commented):
**Overall Self-Review — `origin/master...5ffe701f8`, three complete rounds**

Scope of every round: the whole base-to-head diff (13 files), plus the consequence surface it reaches — every unplugin adapter's `addWatchFile` consumer, `@ttsc/metro`'s transformer and fingerprint, the fixture sidecar's Go source, the host graph and its input observation, and the two documentation pages that state the contracts this changes.

**Round 1** — two findings, corrected in `a0dbbc9bf`:

1. Speculative classification depended on the iteration order of `graph.candidates`. Recorded as inline comments on the Individual Self-Review of `06d72d14f`.
2. The declared-input reader iterated `graph.globals` / `graph.configs` / `hostInputs` without the `Array.isArray` guard every sibling reader applies.

**Round 2** — two findings, corrected in `5ffe701f8`: a `temporaryTsconfig` binding that stopped being read once the watch notification took the generation, and an em-dash in one new case's doc comment.

**Round 3** — clean. Every removal in the diff is accounted for by an intentional replacement, and nothing else survived verification.

Verified in the rounds, beyond the diff text:

- **Contract compatibility.** `addWatchFile` gained an optional second parameter. Every consumer (`core/index.ts`, `turbopack.ts` forwarding webpack's `addDependency`, `bun.ts`'s no-op, `@ttsc/metro`'s snapshot recorder) takes one parameter and is unaffected.
- **Invalidation is preserved, not traded away.** A candidate that appears still replaces the generation (its recorded `missing` marker is state); a realized member without a proof still refuses reuse; a contradictory proof still refuses; an edited project source, a changed out-of-walk declaration, a retargeted link, and a membership change all still invalidate. Each has a case.
- **What the input scoping gives up, and why it is sound.** Content changes to files no compile consumed no longer invalidate. Files appearing or disappearing stay covered by the directory-membership snapshot, which is a direct metadata comparison rather than a notification, so the fallback path keeps its own proof. A graph-free envelope declares no input set and keeps whole-walk comparison unchanged.
- **Failure attribution.** A file-level walk failure is scoped to its key; a failure that cannot be attributed to a key, and any directory the walk could not enumerate, still taint the whole snapshot.
- **The watcherless skip is inert, not lossy.** With `server.watch: null` Vite installs no watcher, so no registration can ever fire. The adapter's own missing-input poll never depended on that watcher and is untouched; both halves now have a case, and the existing contract case was moved onto a server that actually watches so it asserts behaviour rather than decoration.
- **Windows and POSIX.** Identity keys, the walk's project keys, and the fixture's junction/symlink shapes are unchanged by this diff; the local suite was run on Windows, and CI covers the platform matrix.
- **Concurrency.** The new per-envelope memos (`declaredInputKeys`, `speculative`) are idempotent pure derivations of one immutable envelope, built under the same `WeakMap` state the existing memos use.

Local evidence at this head: `test-unplugin` 131/131 (Windows), `test-metro` green, `vite build` 69 modules with one compile, `vitest --run` 609.96 s → 53.96 s, per cached delivery 2.5-3.5 s → ~0.10 s.

Not carried by this cycle, recorded in the campaign ledger as follow-up evidence rounds rather than left implicit: a cached delivery still validates the delivered file's whole reference closure (O(modules x closure) per build), and the Vite serve missing-input registry still keeps one poller per missing candidate. Both are bounded costs on a cache that now hits, not the stall this cycle closes, and both change what a watcher is trusted for, so they need their own evidence.

Issue #1251, opened 2026-08-19
Handover: what evidence/graph says about an empty population — what shipped, what was withdrawn, and where to pick it up


samchon:
This is the entry point for the work on **what `evidence/graph` says when a population comes back empty**. It states what shipped, which three issues were withdrawn and on what evidence, what the two open follow-ups are, and what was deliberately left undone. Every claim here is reproducible from the merged tree; nothing depends on a local ledger or a prior conversation.

Cycle 1 merged as `ba44bf9e5` (#1238), closing #1236 and #1237. Open follow-ups: **#1248**, **#1249**.

---

## 1. The subject, in one paragraph

An evidence population is the set of files and units a claim or a reference selects. When one materializes **empty**, coverage against it is vacuously satisfied — every host truthfully cites zero of zero. The rule therefore reports the emptiness itself. This subject is not *whether* to report it (settled: it stays a build error) but **what the report says**: which cause it names, whether it names the thing the author can act on, and whether it says the same thing once or once per host.

---

### The empty-population report is made once, at its cause

`evidence/graph.go:723-725` now ends a reference the moment its population is empty, for **every** policy:

```go
if len(reference.Units) == 0 {
  continue
}
```

Before, `singleEvidencePerSymbol` was excepted, and on an empty population it added one finding **per selected host** on top of the materializer's own message — each telling that host to cite a unit that does not exist. On a claim with three hosts:

```
Claim 1 reference 1 (markdown, symbols: h2) matched 1 file(s) but found no
selected evidence units (h2). Select symbol kinds present in those files or
correct the reference globs.

Evidence host TypeScript function 'one' at src/a.ts:1 … cites 0 distinct
selected evidence unit(s); singleEvidencePerSymbol requires exactly 1. …
Evidence host TypeScript function 'two' at src/a.ts:2 … cites 0 …
Evidence host TypeScript function 'three' at src/a.ts:3 … cites 0 …
```

Only the first line survives today.

**The old exception was also conditional on `len(reference.Paths) != 0`**, so an identical empty population was judged or skipped depending on whether any file happened to match — a fact the question does not turn on, and one no comment ever justified. That clause is gone from the per-host guard at `graph.go:1040` as well; a reference holding units has matched a path by construction.

[...]

Issue #1249, opened 2026-08-19
A walk error prints a raw OS path where every message beside it prints a project-relative one


samchon:
## Problem

The Markdown and Prisma walkers report a filesystem error against the raw absolute path the walker handed them, while every other message in the same two functions prints a project-relative, slash-normalized path. One loader therefore spells paths three different ways depending on which line reported.

On Windows the walk-error line additionally prints backslashes, because it is the one path in these messages that never passes through `filepath.ToSlash` or the `/`-normalizing helper. An author comparing the failing path against the globs, the root, and the file locations printed beside it is comparing spellings that do not line up.

## Evidence

`loadMarkdownBase` in `packages/evidence/native/markdown.go` produces three path spellings:

| Line | Message | Spelling |
| --- | --- | --- |
| `markdown.go:61` | `could not inspect '<path>'` | **raw absolute, OS-native separators** |
| `markdown.go:91` | `could not read Markdown file '<path>'` | `address.Display` — project-relative, `/` |
| `markdown.go:110` | `could not walk Markdown root '<path>'` | `populationRootLabel` — project-relative, `/` |

`prisma.go:256` and `prisma.go:281` are the same pair for Prisma.

The value printed at `markdown.go:61` is `current`, the callback parameter of `filepath.WalkDir`, which is an OS-native absolute path:

```go
err := filepath.WalkDir(base.Absolute, func(current string, entry fs.DirEntry, walkErr error) error {
  if walkErr != nil {
    relative, ok := relativeProjectPath(base.Absolute, current)
    relevant := ok && (matchesConfiguredMarkdownFile(config, base, relative) || …)
    …
    problems = append(problems, "Evidence graph could not inspect '"+current+"': "+…)
```

The repair is already in hand two lines above. `relative` is computed to decide relevance, and `relativeProjectPath` (`typescript.go:106-121`) returns a `/`-normalized base-relative path. `base.display(relative)` (`base.go:65-72`) then yields exactly the spelling `markdown.go:91` uses for a file. The function computes the right form, uses it for the decision, and prints the wrong one.

Note the guard order: the message is only reached when `ok` is true, so `relative` is always valid at that point.

## Consequence surface

- **Both kinds, one shape.** `markdown.go:61` and `prisma.go:256` are the same defect; repairing one and leaving the other reintroduces the branch asymmetry that #1236 existed to remove.
[...]

Issue #1248, opened 2026-08-19
A root diagnostic names a spelling the author never wrote


samchon:
## Problem

Every `evidence/graph` diagnostic that names a declared `root` prints a spelling the rule derived, not the one the author wrote. When the declared root is relative the two coincide and nothing is visible. When it is **absolute** — a documented, supported form — the diagnostic hands back an ascending relative path that appears nowhere in the author's configuration.

An author who declares `root: "C:/contracts"` and searches their `lint.config.ts` for the `'../../001/contracts'` the message names finds nothing.

This contradicts the stated purpose of naming the base at all. `describePopulation` (`packages/evidence/native/config.go:888-891`) records it:

> a citation that resolves outside the project has to be repairable from the diagnostic alone: patterns that look correct against a root the reader is imagining are the failure this property introduces, and naming the resolved base is what removes it.

A base named in a spelling the configuration does not contain is not repairable from the diagnostic alone.

### Reproduction

Against `686404bbf`, using the package's own `runRootedGraph` helper. Declare an absolute root that does not exist:

```json
{"claims":[{
  "type":"typescript",
  "root":"C:/Users/.../Temp/ttsc-c2-absent",
  "files":["src/**/*.ts"],
  "symbol":"type",
  "reference":{"type":"markdown","files":["docs/**"],"symbol":"h2"}
}]}
```

produces:

```
Evidence graph found no directory at the typescript root '../../../ttsc-c2-absent',
which resolves to 'C:/Users/.../Temp/ttsc-c2-absent'. ...
```

And with an absolute root that **does** exist, so the glob diagnostic is the one that fires:

```
Claim 1 reference 1 (markdown, symbols: h2) matched no markdown files for
['specs/**'] under root '../../001/contracts'. ...
```

In both, the first spelling is derived and the second is the author's — the message prints the derived one where it names the property to edit.

### Root cause

`resolvePopulationBase` (`packages/evidence/native/base.go:131-147`) keeps the resolved absolute path and a `Display` field **recomputed** from it:

```go
return populationBase{
  Absolute: absolute,
  Display:  projectRelativeDisplay(root, absolute),
}
```

The declared spelling is consumed and never stored. `projectRelativeDisplay` (`base.go:149-165`) prefers a relative spelling even when it ascends, falling back to the absolute path only when `filepath.Rel` errors — which on Windows means a different volume.

That preference is **reasoned, not accidental**, and the reasoning is sound for its own subject:

[...]

PR #1252, opened 2026-08-19
fix(evidence): spell a population diagnostic's paths the way the author and the reader read them

samchon:
Cycle 2 of the population-diagnostics campaign, whose handover and entry point is #1251.

This pull request owns the **complete accepted cycle**: #1248 and #1249, plus anything the
implementation and its reviews surface and this cycle publishes as a new issue. Both
carried issues are about the same thing — **how a population diagnostic spells a path** —
and they are one unit for that reason, not because they are small.

## Scope

| # | What |
| --- | --- |
| #1248 | A root diagnostic names a spelling the author never wrote. `populationBase` discards the declared root and recomputes a project-relative one, so an absolute declared root is handed back to the author as an ascending path their configuration does not contain. Also settles the two clauses #1238 deliberately deferred: `it resolves against the ttsc project root`, false for any absolute root, and the `add that directory` repair, unfollowable when a non-directory occupies the path. |
| #1249 | A walk error prints the raw OS-native `WalkDir` path where every message beside it prints a project-relative, slash-normalized one — on both the Markdown and the Prisma walker. |

Implementation order is #1248 then #1249: #1249's output follows `base.Display`, whose
derivation #1248 discusses and deliberately leaves alone.

## Discovery gate — stated honestly

#1251 §7 records that the cycle-2 discovery round was **scoped, not exhaustive**, so under
the campaign's own rule a complete fresh full-scope round is owed before implementation
opens. The assignment that opened this cycle explicitly instructed not to run that round,
and explicitly authorized publishing and implementing whatever the work surfaces naturally.
Implementation therefore opened on the accepted set without the empty-round gate. This is
recorded rather than skipped silently: the campaign is not completable on this basis, and a
full-scope round remains owed before it can be called finished.

## Verification

Pending. The pull request's ordinary CI and a complete clean Overall Self-Review are the
acceptance gates; each implementation commit additionally carries one advisory Individual
Self-Review recorded here as a formal review.

Note for anyone running the Go suite locally: `go test ./native` in `packages/evidence`
fails 24 Prisma and Swagger bridge cases until `pnpm install` has linked the feature suite.
Those are environmental, not regressions (#1251 §9).


packages/evidence/native/base.go
```diff
@@ -213,32 +259,60 @@ func configuredBases(config graphConfig, kind artifactKind) []populationBase {
...
+// This is the one place a declared spelling is preferred over the derived one,
+// and the split is between two questions rather than two formats. A message
+// naming a *location* spells it the way the reader opens it, which is what
+// `display` does and why every file this rule prints stays project-relative. A
+// message naming a *configuration property* spells it the way the configuration
+// does, because the author's first move is to search `lint.config.ts` for it —
+// and the derived spelling of an absolute root is an ascending path that file
+// does not contain.
+//
+// The absolute fallback belongs to the default base alone, which is the only
```
samchon, line 289: **F5 (precision, accepted).** This paragraph justifies a branch nothing currently reaches. `base.Declared == ""` is true only for the default base, and every other caller of `populationRootLabel` returns early on `base.Default`: `missingBaseDirectoryProblem` here, and `describePopulation` in `config.go`. The only two remaining call sites are `markdown.go:110` and `prisma.go:281`, which F4 shows are unreachable.

Adjudicated together with F4 under #1254. Repairing the walk-root failure is what makes this fallback live, and it is the only spelling available for a base the author never declared.

packages/evidence/native/base.go
```diff
@@ -213,32 +259,60 @@ func configuredBases(config graphConfig, kind artifactKind) []populationBase {
...
 // missingBaseDirectoryProblem reports a declared root that is not an existing
 // directory.
 //
-// The test is a stat, and any failure of it counts as absence — a path occupied
-// by a file, and an unreachable parent alike. The Markdown and Prisma messages
-// may still lead with "could not read", because the walk each of those callers
-// is about to run is the read the root exists to serve, and it is skipped only
-// because the root is not there. The predicate's own name may not, because the
-// third caller reads nothing and asks the same question.
+// The test is a stat, and three states fail it: nothing at the path, an
```
samchon: **F6 (precision, accepted).** The enumeration claims three states and is not exhaustive. A fourth is reachable on POSIX: the root is an existing directory whose parent denies traversal, so `os.Stat` fails, `occupied` is false, and the author is told to `create that directory` over a directory that already exists. That is the same unfollowable repair this commit fixed for the occupied case.

Fixed rather than reworded: the predicate now tells a permission failure apart and asks for access instead of creation.

samchon (commented):
## Individual Self-Review — `06d5d810f` (#1248)

`fix(evidence): name a population root the way the configuration declares it`. One read-only reviewer over the parent-to-commit diff, per the review skill's Individual Self-Review. Advisory only; every disposition below is the main agent's.

**Eight candidates raised. All eight accepted; none rejected.** Five are repaired in the follow-up commit on this branch, and two are folded into #1254, published from this review.

| # | Where | Claim | Disposition |
| --- | --- | --- | --- |
| F1 | `packages/evidence/src/structures/ITtscEvidenceGraph{ClaimBase,MarkdownReference,PrismaReference}.ts` | the adopter TSDoc still says `Diagnostics name the resolved base`, which is now the opposite of what the rule prints for an absolute root | accepted, repaired |
| F2 | `packages/evidence/native/config.go:885-891` | `describePopulation`'s rationale still calls the printed base the resolved one, and it is the passage #1248 quoted as the stated purpose | accepted, repaired |
| F3 | `claims.mdx:183`, `project/evidence/SKILL.md:120` | `exact text` overclaims: `Declared` is normalized, not written | accepted, repaired |
| F4 | `markdown.go:110`, `prisma.go:281` | both `could not walk ... root` sites are unreachable, so two of the five `populationRootLabel` consumers this commit moved can never reach a user, and one acceptance row cannot be pinned | accepted, **#1254** |
| F5 | `base.go:274-282` | `populationRootLabel`'s fallback is reachable only from those two dead sites | accepted, **#1254** |
| F6 | `base.go:287` | the new three-state enumeration is not exhaustive; a stat denied by an unreadable parent still gets `create that directory` | accepted, repaired |
| F7 | `population_roots_...test.go:103` | the backslash case uses a relative root, the one form where `Declared` and `Display` coincide | accepted, repaired |
| F8 | `population_roots_...test.go:69` | the occupied branch has no negative twin | accepted, repaired |

F1 is the one that matters most, and it is this cycle's own recorded failure mode arriving on schedule. #1251 §10 names `packages/evidence/src/structures/` as documentation that moves with these messages, and the commit moved `claims.mdx` and the skill and stopped there. A TSDoc an adopter reads on hover is the same kind of consumed surface as the benchmark parser that §6 records being missed by ten per-commit reviews.

F4 was reached independently by the main agent while implementing #1249 and by this reviewer from the acceptance table. Both derivations agree with `$GOROOT/src/path/filepath/path.go`: `WalkDir` returns only what the callback returns and converts `SkipDir` to `nil` at every level, and both callbacks return exclusively `nil` and `SkipDir` on directory entries. It is published as #1254 rather than repaired inline because it changes what the rule concludes rather than what it says, and it needs its own evidence, acceptance matrix, and tests.

### Verified sound, recorded so it is not re-derived

- Every branch was rendered, not read, per #1251 §9: relative and absolute roots, missing and occupied, across all three artifact kinds, plus `/srv/contracts`, `C:/contracts`, `//server/share`, and `C:/`. The relative-root messages are byte-identical to the parent's, so no existing consumer moved.
- `declaredRootIsAbsolute` tells the truth per platform. On Windows `/srv/contracts` is not absolute, is joined to the project, and both dropped clauses correctly stay.
- The comment claim that both spellings appear only while they differ holds: for every absolute form, `filepath.ToSlash(base.Absolute)` equals `base.Declared` exactly, including the UNC and drive-root forms `normalizeRootPath` restores.
- `Display`, `artifactAddress.Relative`, inventory keys, and citation targets are untouched in code and pinned by two cases.
- `configuredBases` first-declaration-wins is accurate; the `sort.Slice` reorders output without changing which object the `seen` map admitted.
- Downstream substring consumers are intact. All three fixtures under `tests/test-evidence/src/features/` use relative roots, and `tests/test-evidence-benchmark/src/internal/evidenceDiagnostics.ts` filters on `matched no `, which this commit does not touch.

### Not verified

POSIX-only behavior was reasoned from the Go path API rather than executed, because the implementing machine is Windows: `C:/x` on Linux, the F6 `EACCES` case, and UNC handling under `filepath.Clean` on POSIX. CI is the gate for those.

packages/evidence/native/population_roots_are_named_as_the_author_wrote_them_test.go
```diff
@@ -0,0 +1,384 @@
...
+ * The negative twin of the case above, and the one the repair could most easily
+ * overrun. A relative root is the form where the derived spelling is the
+ * author's own and where the project root genuinely is composed into it, so
+ * every clause the absolute case drops has to survive here.
+ *
+ *  1. Declare the same claim with an ascending relative root.
+ *  2. Read the root diagnostic.
+ *  3. Assert the resolved location and the resolution clause are both present.
+ */
+func TestARelativeRootKeepsItsResolvedLocationAndClause(t *testing.T) {
```
samchon, line 72: **F8 (test-gap, accepted).** This twin stops at `which resolves to '`, the position before the new occupied clause would be inserted, so it would still pass if that clause leaked onto the missing case. Nothing in the repository asserts `add that directory` or `create that directory` either.

The missing-root twin now asserts the `add that directory` clause and the absence of the occupied clause.

packages/evidence/native/population_roots_are_named_as_the_author_wrote_them_test.go
```diff
@@ -0,0 +1,384 @@
...
+ * The normalization belongs to the decoder, which runs before a project identity
+ * exists to resolve against; this pins the composition end to end, because
+ * storing the declared spelling is what makes the decoder's output visible to a
+ * reader at all.
+ *
+ *  1. Declare the root with backslashes.
+ *  2. Read the root diagnostic.
+ *  3. Assert it names the slash-separated spelling and carries no backslash.
+ */
+func TestARootWrittenWithBackslashesIsNamedWithSlashes(t *testing.T) {
```
samchon, line 117: **F7 (test-gap, accepted).** A relative root is the one form where `Declared` and `Display` coincide, so this case cannot distinguish the new field from the old derivation. The acceptance row it stands for is an absolute root written with backslashes.

A unit assertion over `normalizeRootPath` into `resolvePopulationBase` now covers the absolute spelling, which is platform-independent because the joined-versus-not difference does not change the stored spelling.

website/src/content/docs/evidence/claims.mdx
```diff
@@ -180,6 +180,8 @@ A monorepo usually keeps one requirements set that several packages implement, a
 
 `root` is one directory, never a glob. It may sit inside the project (`docs`), above it (`../../docs`), or on an absolute path (`/srv/contracts`, `C:/contracts`). A Windows drive-relative path such as `C:docs` is refused, because it resolves against whatever directory that drive currently sits on.
 
+A diagnostic that asks you to correct `root` quotes the spelling you declared, so an absolute root is found by searching `lint.config.ts` for the exact text in the message. File locations stay project-relative in the same message, because one names the property to edit and the other names the path to open.
```
samchon: **F3 (precision, accepted).** `Declared` is the normalized spelling, not the written one, so `exact text` overclaims. `C:\contracts` prints `C:/contracts`, `./docs` prints `docs`, `docs/` prints `docs`, and `../a/../contracts` prints `../contracts`. The commit's own backslash case proves the first. Reworded on both documentation surfaces.

samchon (commented):
## Individual Self-Review — `7e0f01567` (#1249)

`fix(evidence): spell a walk failure the way every message beside it is spelled`. One read-only reviewer over the parent-to-commit diff, per the review skill's Individual Self-Review. Advisory only; every disposition below is the main agent's.

**Six candidates raised. All six accepted; none rejected.** All are repaired in `a338797e7`. The findings are recorded in the body rather than inline because every line they name has since moved.

| # | Where | Claim | Disposition |
| --- | --- | --- | --- |
| G1 | `walk_failures_...test.go:154-160` | **the two assertions are mutually exclusive on Linux.** The test requires a message containing `could not inspect 'docs/private':` and then requires no message to contain the absolute path, but the same message continues with `walkErr.Error()`, which this commit deliberately passes through and which on Linux is `open /tmp/.../docs/private: permission denied` | accepted, repaired |
| G2 | `base.go:451-453` | the new comment says every other path those two functions print is project-relative, and both `missingBaseDirectoryProblem`'s `which resolves to` clause and `populationRootLabel`'s default-base return print an absolute one | accepted, repaired |
| G3 | `walk_failures_...test.go` | the wiring is unproven on the Windows lane: reverting both walkers to print the callback's own argument leaves every portable case green, and only the POSIX-guarded cases catch it | accepted, repaired |
| G4 | `walk_failures_...test.go:69-71` | the backslash assertion scans the whole message including the pass-through cause, so it is green only because that row's cause happens to be POSIX-shaped | accepted, repaired |
| G5 | `walk_failures_...test.go:57,59` | the two `.` rows pin a composition the walkers reach only through a leading-`**` glob | accepted, repaired |
| G6 | `walk_failures_...test.go:23,26,30` | one skip message for three different reasons, so a lane that silently stopped denying reads as a pass | accepted, repaired |

**G1 is the finding that mattered.** It would have gone red on the `go` lane on the first CI run, and the reviewer reached it by rendering the message a Linux `*fs.PathError` produces instead of reading the assertion beside the one it replaced, which is the method #1251 §6 records as the only thing that caught this cycle's earlier precision defects. The repair anchors the negative assertion on the quoted segment, which is the part this rule authors, and states why in the test.

**G3 changed the design, not just the test.** The walk-failure decision is now one shared function, `unreadableEntryProblem`, taking the population's membership question as a closure. That makes the property the issue actually bought, that the walker composes `relative` rather than printing `current`, provable from a portable case with an OS-native absolute path, and it removes the second copy of a guard whose halves had already drifted once.

**G5 removed code rather than a test.** With #1254 landing in the same pull request, the base is answered before the entry decision is asked, so `relative` can no longer be `"."` and the `path.Clean` that collapsed it became a call that cannot change its argument. Both the call and the rows that exercised it are gone.

The reviewer also reached the `could not walk ... root` dead-code observation independently and filed it as incidental. It was already published as #1254 from the previous review and is implemented in `a338797e7`.

### Verified sound, recorded so it is not re-derived

- Behavior beyond text is untouched in both walkers: `recordPopulationFailure`, `baseFailed`, the `SkipDir` returns, and `failedBases` accumulation are unchanged, so population health, derived-diagnostic suppression, and claim activation do not move.
- `relevant := ok && ...` means `relevant` implies `ok`, so the base-relative path is always valid where the message is composed.
- No downstream consumer parses `could not inspect`. The whole tree was grepped: only this package, a different message in `graph.go` naming a package, and a comment in `health.go`. The benchmark parser filters on `matched no ` and `Missing acknowledgement for '`, neither of which this touches.
- The class is sealed at two walkers. The third `filepath.WalkDir` site, in `loader.go`, names a configured package rather than a rule-composed path.
- `path.Clean` was correct over every composition the reviewer could construct before it was removed, including an empty `Display`, an ascending one, and a Windows-volume one.
- `ubuntu-latest` runners execute as the non-root `runner` user, so the POSIX-guarded cases genuinely run there rather than skipping.

### Not verified

Execution of the POSIX-guarded cases on Linux. The implementing machine is Windows and they skip here; CI is the gate. G6 exists so that a lane which stopped denying says so instead of reporting a pass.

samchon (commented):
## Individual Self-Review — `a338797e7` (#1254)

`fix(evidence): report a population whose own base could not be listed`. One read-only reviewer over the parent-to-commit diff, per the review skill's Individual Self-Review. Advisory only; every disposition below is the main agent's.

**Nine candidates raised. All nine accepted; none rejected.** All are repaired in `1200851a9`, together with a fifth issue the reviewer surfaced as incidental and this cycle published as **#1258**.

| # | Where | Claim | Disposition |
| --- | --- | --- | --- |
| A1 | `markdown.go:52-55` | the comment says the relevance guard *returns false for the base itself*; it is decided by the glob shape, true under a `**`-leading pattern and false under a segment-leading one | accepted, repaired |
| A2 | `an_unlistable_..._test.go:46-49` | the same reversed sentence in a test doc comment, contradicted 50 lines below in the same file | accepted, repaired |
| B1 | `an_unlistable_..._test.go:37-39` | the default-base assertion cannot fail: the template it inspects has no conditional and no such substring | accepted, repaired |
| C1 | `base.go:376-387` | the third stat state tells the author to *make that path reachable*, a cause the rule does not have. Reachable today: `root: "../do<cs"` on Windows returns a name-syntax error that is not `fs.ErrNotExist`; `ENAMETOOLONG` and `ELOOP` are the same shape on POSIX | accepted, repaired |
| C2 | `base.go:320-324` | the comment records the POSIX/Windows split and then dismisses its one consequence: on Windows a path under a file lands in the absent branch, whose `create that directory` cannot be followed | accepted, repaired |
| C3 | `base.go:346-349` vs `380-387` | *TypeScript is told apart in the verb and in the repair clause* became true of two branches out of three; the third drops the clause saying a TypeScript root never scans the filesystem, in the one message about filesystem access | accepted, repaired |
| D1 | `an_unlistable_..._test.go:151-168` | the claim-side case uses `**/*.md`, the one glob shape under which the parent already reported, so it never provokes the silence the acceptance row names | accepted, repaired |
| E1 | `base.go:492` | removing `path.Clean` regressed the drive-root composition: a base whose `Display` is `D:/` prints `D://requirements`. The class is wider and pre-existing, since `addressOf` uses the same `display` | accepted, repaired at `display` |
| F1 | `markdown.go:79-83`, `prisma.go:267-271` | with the base short-circuited, `entry != nil` is always true and the trailing `return nil` is unreachable | accepted, repaired |
| F2 | `base.go:370-375` | the three-state enumeration did not state the precondition the `cause.Error()` call depends on | accepted, repaired in `1e78cba2a` before the review arrived |
| F3 | `population_roots_..._test.go:440` | a failed stat leaves two states, not three; `occupied` is a successful one | accepted, repaired |

**A1 is this cycle's signature failure arriving for the fourth time**, and this instance is the clearest yet: the commit message and the Prisma comment beside it both state the glob dependence correctly, and only the Markdown comment reverses it, about the exact input the test one file away exists to disprove. The reviewer reached it by tracing `splitProjectPath(".")` through `couldMatchGlobBelow` rather than by reading the sentence.

**C1 is a defect the main agent introduced while repairing one.** The `unexaminable` branch was added because "create that directory" asserts a state the rule has not established; it then asserted one of its own. The repair names no cause at all and leaves the operating system's sentence, one clause earlier, to say what happened.

**The incidental note became #1258.** A base that is a symbolic link or a Windows junction is accepted by `os.Stat`, which follows it, and skipped by `filepath.WalkDir`, which lstats it, so the population materialized empty and healthy in silence. Reproduced with this package's own `linkDirectory` helper before publication: `lstat.IsDir=false, stat.IsDir=true`, and the rule printed `matched no markdown files for ['requirements/**/*.md'] under root '../documents'` over documents that were there. It is the same invariant #1254 states, reached through a different door.

### Verified sound, recorded so it is not re-derived

- `filepath.WalkDir` could never return non-nil before and now can, traced through `$GOROOT/src/path/filepath/path.go`. The new `return walkErr` for the base propagates unconverted at both levels.
- Nothing is swallowed or duplicated and no partial result is lost. The base callback fires with an error only before any child is visited, and the old `SkipDir` return took the same early exit, so the entry loop never ran in either version.
- A Markdown claim base is walked twice, in the activation pass and the post-activation pass, so the message is appended twice and `reportProblems` drops the adjacent duplicate. The acceptance row "root named once" holds.
- `relative == "."` is genuinely unreachable at the entry site, checked against `C:/`, `D:/`, `//server/share`, `/`, and trailing-separator roots.
- `os.Stat` on a mode-0 directory still succeeds, because `stat(2)` needs search permission on the prefix and `t.TempDir` leaves the parent at 0700. The whole new test file rests on that hinge.
- All eighteen `describeBaseDirectoryProblem` renderings were produced from the committed functions in an isolated harness and read as a user.
- `populationRootLabel`'s default-base fallback is live through exactly one caller, `unlistableBaseProblem`, as #1254 predicted.
- Suppression works: `graph.go` gates the `matched no ... files` line on the reference's health, and the synthetic failure inventory matches by `FailureBase`.
- Downstream substring consumers are intact. All three feature fixtures provoke a genuinely absent directory, so they keep landing in the absent branch rather than the new `could not examine` one.
- Cleanup ordering is right in all seven POSIX-guarded cases.

### Not verified

Execution of the POSIX-guarded cases on Linux, from a Windows machine. That gap has since closed: the `go` lane passed on `ubuntu-latest` for the following commit, which is where those cases actually run.

samchon (commented):
## Overall Self-Review — rounds 1 to 6

Solo, whole base-to-head diff, restarted from scratch after every applied improvement, per the review skill's Non-Negotiable Review Law. Recorded as one review because the rounds are short and consecutive; each is a complete pass over the whole surface and each ended by applying what it found.

**No product defect in rounds 1, 2, 4, or 6. Rounds 3 and 5 found real gaps.**

| Round | Head at start | What it found | Commit |
| --- | --- | --- | --- |
| 1 | `7e0f01567` + working tree | four comments that stopped matching the code beside them: a caller named as the wrapper rather than the caller, a headline that described neither the input nor the report, a comparison against "every other message in the same function" from before that function was split, and a test step that called the omitted root one of the spellings naming the project root | `1e78cba2a` |
| 2 | `1e78cba2a` | `describeBaseDirectoryProblem` reads `cause.Error()` on a path where `occupied` is false, which is safe only because exactly one of the two is ever set. Its comment did not say so. A guard would have been a branch no caller can reach, so the invariant is stated instead | folded into `1200851a9` |
| 3 | `1200851a9` | three acceptance rows across #1249 and #1258 that no case pinned: the Prisma walker under a declared root, population health after an entry failure, and a link met inside the population. Also a link assertion that matched the bare word `target`, which the diagnostics are free to use as prose | `f017f0dd2` |
| 4 | `f017f0dd2` | **`zz_probe_review_test.go`, a read-only reviewer's throwaway, was in the pull request.** It entered through a `git add -A` in the commit that followed, in a commit whose message says nothing about it. Found by reading the pull request's file list rather than its hunks | `8908a61dc` |
| 5 | `8908a61dc` | the #1258 repair already covered the **default** base and nothing said so or held it. A project whose own root is a link had every Markdown and Prisma population come back empty without declaring a `root` at all. The new case fails without the resolution and passes with it, alongside the four declared-root cases | `bc3dad6c3` |
| 6 | `bc3dad6c3` | a test doc comment explained the repair with a glob the case does not configure, naming a directory that appears nowhere in it | `e9526f56b` |

Round 4 is the one worth carrying forward. Three Individual Self-Reviews were told to delete their probes and did; the file that survived was recreated by a later reviewer while the main agent was staging, and `git add -A` swept it in. Every commit since stages explicit paths.

Round 5 is the second half of #1258 and was reached by asking which base the repair had *not* been asked about. The declared root is the case the issue was written from; the default base is the one nearly every project uses.

### Verified across the rounds, recorded so it is not re-derived

- The common case is untouched. With no `root` declared, `populationRootLabel`, `describePopulation`, and `baseDirectoryProblem` all return before naming anything, `display` returns the bare relative path, and the only change is that a walk root which is a link now resolves.
- `.Declared` is read in exactly two places, `populationRootLabel` and `declaredRootIsAbsolute`. Every other consumer of a base reads `.Absolute` or `.Display`, so inventory keys, citation targets, watched inputs, and both caches are untouched. Checked by grepping every field access outside `base.go`.
- The four issues' acceptance tables were walked row by row. Every row of #1248 and #1254 was already pinned; the three that were not are listed under round 3.
- The five linked-root cases were proved to fail without the repair by reverting `populationWalkRoot` to return `base.Absolute` and re-running, and `TestALinkInsideThePopulationIsNotFollowed` correctly passes either way, which is what makes it the negative twin.
- Every message was rendered and read as a user after each change, per #1251 §9, including the twelve `describeBaseDirectoryProblem` shapes and the three walk shapes.
- No downstream consumer moved. The three feature fixtures provoke an absent directory and keep landing in the absent branch; the benchmark parser filters on `Missing acknowledgement for '` and `matched no `, neither of which this touches.

samchon (commented):
## Individual Self-Review — `1200851a9` (#1258)

`fix(evidence): walk a population root the way the gate already read it`. One read-only reviewer over the parent-to-commit diff, per the review skill's Individual Self-Review. Advisory only; every disposition below is the main agent's.

**Eight candidates raised. Seven accepted, one withdrawn by the reviewer after HEAD had already repaired it.** All are repaired in `bf49ffa97`, together with one defect that repair introduced and rendering caught.

| # | Where | Claim | Disposition |
| --- | --- | --- | --- |
| F1 | `base.go:485` | **`populationWalkRoot` leaves the exact silence it closes reachable past 32 hops.** `resolveLinkedDirectory` gives up after a fixed number of hops and returns the link it stopped on, while `os.Stat` follows further on every platform, so the gate accepts and the walk descends into nothing. Reproduced with a 41-junction chain, printing byte-for-byte the diagnostic #1258 exists to remove | accepted, repaired |
| F2 | `claims.mdx:185` | the new paragraph says a link leaves every citation target and file location "spelled through the `root` you wrote", contradicting the paragraph above it and false for an absolute root, whose file locations ascend project-relatively | accepted, repaired |
| F3 | `claims.mdx:185` | "Links found inside the population are not followed" is true for a linked directory and false for a linked file, which `os.ReadFile` follows like any other | accepted, repaired |
| F4 | linked-root tests | no case covers a root link with no target, whose repair is `create that directory` over a path where `mkdir` returns `EEXIST` | accepted, repaired in code and pinned |
| F5 | `..._directories_test.go:198` | the drive-root case hand-builds the base, so nothing shows a configuration can produce a `Display` ending in `/` | accepted, repaired |
| F6 | `population_roots_..._test.go:443` | the step list omits the assertion the previous commit added | accepted, repaired |
| F7 | `prisma.go:246` | the comment's two halves pull against each other: the relevance test does not answer for the base, and Markdown records why it did, by accident | accepted, repaired |
| F8 | `base.go:424` | the third stat state asks the author to clear a filesystem condition and then says this kind never touches the filesystem, with nothing between them | accepted, repaired |
| — | `..._directories_test.go` | the `target` substring assertion matched the bare word, which roughly twenty graph diagnostics use as prose | withdrawn; `f017f0dd2` had already anchored it on `/target/` |

**F1 is the finding of this cycle.** The repair for #1258 replaced a silence with a bound, and the bound is lower than the operating system's, so the silence survived past it. The issue's own consequence surface said "the resolver is already bounded" and treated that as sufficient; it is the *comparison* between the two bounds that matters. The class is now sealed by checking the resolver's answer rather than trusting it, at the cost of one `Lstat` of a path the walk was about to read.

**F8's repair introduced a defect, and rendering caught it.** Re-adding the TypeScript clause as its own sentence left the separator that used to introduce it, so an absolute TypeScript root printed `... clear the condition the filesystem reported; . A typescript root is checked ...`. This review had verified that `"; ."` was not constructible, and it was right about the shape it reviewed. A scan over every kind, base shape, and stat state now checks each assembled message for a dangling separator, a doubled space, and a missing terminator; it reports none.

**`unlistableBaseProblem` gained a second cause and needed one repair for both.** "Fix filesystem access" named an access a link chain has nothing to do with, so both causes are now answered by making the root a directory this process can list.

### Verified sound, recorded so it is not re-derived

- Nothing still compares against `base.Absolute` where the walk root is the measure. Every remaining use is base identity, a health key, or the Program-side mapping.
- The Prisma parser is handed a path that exists: `distinctPrismaSources` dedups on the declared base's `Display`, and the read resolves it through the link.
- `recordPopulationFailure` and `populationIsHealthy` key on `base.Absolute` throughout, so the walk root never reaches them.
- The nil-entry removal is exactly equivalent: `WalkDir` passes a nil entry only for its root, which the base guard answers.
- The corrected Markdown glob claim is true, traced through `splitProjectPath(".")` into `couldMatchGlobBelow`.
- `resolveLinkedDirectory` is correct for this caller across a broken link, a link to a file, a link loop, a relative target, a linked ancestor, and a two-hop chain, each measured.
- Watched inputs are consistent: the host resolves a glob's non-wildcard prefix through a link, so a linked root's documents are watched.
- The linked-root cases run on Linux and macOS, where `os.Symlink` succeeds and the junction fallback is never reached. Every one of them is red when `populationWalkRoot` is reverted to return `base.Absolute`.
- No downstream consumer parses any changed string.

### Not verified

The linked-file counterexample behind F3 is reasoned from the code path; this machine denies `os.Symlink` for files and a junction cannot target one. F1's reproduction used Windows junctions, and its POSIX equivalent is reasoned from Linux's traversal limit against the resolver's bound.

samchon (commented):
## Individual Self-Review — `bf49ffa97`

`fix(evidence): seal the link-chain hole and the repairs that could not be followed`. One read-only reviewer over the parent-to-commit diff. Advisory only; every disposition below is the main agent's.

**Four candidates raised. All four accepted.** All are repaired in `49cbd22fe`.

| # | Where | Claim | Disposition |
| --- | --- | --- | --- |
| R1 | `base.go:331` | **a regression this commit introduced.** The `Lstat` that tells a link with no target from a missing directory ran on every failed stat, so a junction loop, a target behind a parent that denies traversal, and a share that went away were all promoted to "a non-directory occupies this path" and told to be replaced, with the operating system's own sentence dropped | accepted, repaired |
| R2 | `base.go:361`, `population_roots_..._test.go:435` | both comments list a link loop among the third state's causes, which R1 had rerouted into the second | accepted, repaired with R1 |
| R3 | `base.go:493`, the chain case, the commit message | "the stat that accepts the root follows further than that on every platform" is false on Darwin and the BSDs, whose limit equals the resolver's. The 35-link case would fail there, and the commit message says 34 | accepted, repaired |
| R4 | `claims.mdx:185` | "the link never appears in either" is false: a file location under a linked root is spelled through the declared root, which is the link | accepted, repaired |

**R1 is the finding, and it is the campaign's pattern in its sharpest form yet.** The third stat state exists because "create that directory" asserts a cause the rule has not established. The repair for the dangling link then asserted a different one, in the same function, through the gate rather than through the sentence, one commit later. Measured on a junction loop: `because that path is not a directory` where the parent had printed `The name of the file cannot be resolved by the system`. Only `fs.ErrNotExist` is asked the second question now.

**R3 changed a claim, not only a comment.** The window this cycle sealed is real on Linux and Windows and does not exist on Darwin, where the platform stops following at the same bound the resolver does. The case now verifies the platform instead of assuming it, because a chain the platform itself refuses to follow proves nothing about the one the resolver refuses. The Go suite runs on ubuntu and windows only, so this was never a CI question and always a correctness one. The reviewer is right that the chain is 35 links and the commit message says 34.

**R4 is the second overshoot of the same sentence.** The fourth review found it over-claiming that the link's spelling survives everywhere; the repair over-claimed that it survives nowhere. What is true is that the addresses follow the declared spelling, and the directory the link points at appears in neither.

### Recorded from this review, and published rather than repaired

The reviewer's verification of the Prisma path led to **#1262**: `distinctPrismaSources` deduplicates the parser's input on `address.Display`, so two roots reaching one physical schema hand that file to Prisma twice and the set is rejected for a duplicate model. Measured at `5a885bb73`: `addresses=2 sources=2`. It is left out of this pull request deliberately, and the issue says why: the repair redefines what `prismaInventoriesByDisplay` maps, the wrong version of it silently empties a population, and the choice between serving the configuration and refusing it is a product decision that issue has to settle.

### Verified sound, recorded so it is not re-derived

- Thirty message renderings across kind, base shape, and stat state, plus every `unlistableBaseProblem` and `unreadableEntryProblem` shape. No dangling separator, no doubled punctuation, no missing terminator. The `"; ."` the previous commit introduced is gone and does not reappear in any combination.
- No branch reads `cause` where it may be nil, and the invariant comment describing the two parameters is true as written.
- `populationWalkRoot` cannot produce two diagnostics for one state: both walkers run the gate first and return or continue on a non-empty problem.
- The Prisma `continue` skips nothing; the new branch appends to `failedBases` itself, which is what suppresses derived diagnostics on that side.
- Both new cases are red without the change, and `os.Remove` in the dangling-link case removes the target rather than the junction.
- The linked-file documentation claim is true and is now pinned.
- Ordinary roots and the default base are untouched; an existing directory returns from the gate before the new `Lstat`.

samchon (commented):
## Overall Self-Review — rounds 7 to 14

Solo, whole base-to-head diff, restarted after every applied improvement. Recorded as one review because the rounds are consecutive; each is a complete pass and each ended by applying what it found. Rounds 1 to 6 are in an earlier review on this thread.

**Two rounds found a product defect. The rest found sentences that had stopped matching the code, and two found a validation gap.**

| Round | What it found | Commit |
| --- | --- | --- |
| 7 | the third stat state's doc explained it as an unreadable parent alone, which is the reason it was added for and not the set it catches | `d05a56e1d` |
| 8 | **a doubled terminator in nine messages.** Windows ends its own error text with a period and the rule appended one, so every message carrying a filesystem or subprocess reason read `system.. Correct the` there. Three sites were added by this pull request and six were older; two of the six carry their reason as text rather than as an error, so the grep that found the first seven missed them | `fe3ab5a57`, `786a78b9c` |
| 9 | a case that passed for a different reason than it claimed: after the terminator repair, the period it looked for was the one the rule appends | `5cd0ac1f6` |
| 10 | **`packages/evidence/` never ran its Go tests on Windows.** The plan selected `evidence` and `go`, and `go` is `ubuntu-latest`; across twenty-one pushes of this pull request `windows-go` was never selected once, over a package whose whole subject is path handling | `97fda7a26`, **#1264** |
| 11 | three sentences promising that a diagnostic's file locations stay project-relative, which two reachable roots deny. Measured: a root on another Windows volume and a UNC share produce absolute locations | `89ec59576` |
| 12 | the case establishing that premise duplicated a premise block added one commit earlier for the same reason | `8d5756ecf` |
| 13 | one message fragment asserted by no test: the clause that makes the third state's TypeScript sentence coherent after a filesystem repair | `03babf8a3` |
| 14 | **two pre-existing comments whose premise this cycle invalidated.** `projectRelativeDisplay` justified its choice with "every other path this rule prints is project-relative", which stopped being true when the root label moved to the declared spelling in this same pull request | `1a4ab1fae`, `695f36440`, `68a918c4d`, `86c76bc1a` |

### What changed about how these rounds were run

Reading paragraphs stopped being enough somewhere around round 8, so three mechanical sweeps now run every round:

- **Render every shape.** 504 messages, over three artifact kinds, seven root spellings including a bare drive root and a UNC share, four causes, and every stat state, scanned for a dangling separator, doubled punctuation, a missing terminator, and a doubled path separator. That is what found round 8 and confirmed each repair since.
- **Build every state on disk.** A missing directory, a file occupying the path, a path whose parent is a file, a dangling junction, a junction loop, and a name the filesystem refuses, each stat'd and each message read. That is what proved the guard from the fifth Individual Self-Review lands every real state where it should, and what surfaced the doubled terminator.
- **Cross-check fragments against tests.** Every message fragment in the source, matched against the cases that assert it. That is what found round 13.
- **Grep the package for the claim the cycle changed.** Round 14 came from asking which existing comments rest on a premise this pull request moved, rather than from re-reading the paragraphs it edited.

### Round 10 is the one worth carrying forward

Every one of this cycle's four issues turned on behavior that differs between Windows and POSIX: `filepath.IsAbs` calling `/srv/x` relative, `filepath.Rel` erroring across volumes, a drive root carrying its own separator, a junction that `EvalSymlinks` returns unchanged, and an error message that ends with a period. Each has a test in this pull request. None of them was proving anything, and the pull request's own check list is what said so. The lane already existed and was green on `master`; only the selection rule was missing.

### Not verified

POSIX and macOS execution, from a Windows machine. The `windows-go` lane now covers the Windows half in CI, and the `go` lane covers Linux; Darwin is not a Go lane, and the one case whose behavior differs there verifies the platform at run time rather than asserting it.

samchon (commented):
## Individual Self-Review — `9f6e98020..a2b56989d`

One read-only reviewer over the range. Advisory only; every disposition below is the main agent's.

**Nine candidates raised. All nine accepted.** All are repaired in `9ee765686` and the two commits before it.

| # | Where | Claim | Disposition |
| --- | --- | --- | --- |
| F1 | `population_roots_..._test.go:519` | **the new case for `declaredRootIsAbsolute` cannot fail.** The resolution decides by calling the predicate, so both sides of the assertion move together for every input. Proved by substituting a spelling-based predicate, which is this cycle's own defect, and watching the case stay green over all twelve rows | accepted, repaired |
| F2 | `validation-plan.cjs:465` | the lane comment claims a Windows-only case for each of its reasons, and there is one for none of them: the predicate case carries no platform guard, and the linked-root cases run on both lanes | accepted, repaired |
| F3 | `prisma.go:786`, `swagger_cache.go:269` | two more members of the doubled-terminator class, taking their reason from a parser rather than a syscall, so the syscall sweep missed them for the third commit in a row. No reachable reason ends in a period today, so nothing doubles; the class was not sealed | accepted, repaired |
| F4 | `validation-plan.test.cjs:71` | deleting the Go row unpinned #1264's first acceptance case, over an issue whose subject is running Go tests on Windows. The redundancy claim is true only of today's branch shape | accepted, repaired |
| F5 | `model.go:334` | the `Path` comment enumerates three spellings; a Swagger document's is a configured source that may be a URL, and its diagnostics print it through a different function | accepted, repaired |
| F6 | `base.go:45`, `base.go:82`, `claims.mdx:185` | three places say the only case with no relative spelling is another Windows volume, while `base.go:199`, the plan comment, and a pinned test row all include a UNC share | accepted, repaired |
| F7 | `walk_failures_..._test.go:149` | the bare word `target`, which Overall round 3 already anchored once this cycle for the same reason | accepted, repaired |
| F8 | `population_roots_..._test.go:515`, `:558`, `:439` | two step lists name fewer rows than their case runs and omit an assertion, and a reordered paragraph lost the antecedent of "which of the two" | accepted, repaired |
| — | `base.go:551` | `causeReason`'s new justification says the trim runs in both directions; it only removes, and the caller supplies | accepted, repaired |

**F1 is the finding.** The predicate is what all five issues turn on, it had no case for most of this cycle, and the case written for it in the previous commit asserted a tautology. The repair is the platform's own contract as the oracle — a rooted path with no volume is absolute on POSIX and relative on Windows, a drive-lettered one is the reverse, a UNC share is absolute on both — and the same substitution the reviewer used now fails it. That was verified by running it, not by reading it.

**F1's repair is what makes F2's comment true.** The lane's justification needed a case whose answer differs by platform, and now there is one; before, every case either ran everywhere or skipped one lane, so the lane proved nothing about the predicate.

**F3 is the third consecutive commit to under-count the terminator class.** Seven, then nine, then eleven; the real total is thirteen join sites, and the two that kept slipping take their reason as text from a parser reached through a subprocess rather than from a syscall. Nothing doubles today, because every reachable Prisma and Swagger reason ends in a bracket, a path, or a caret. The two sites read the shared rule now, because "nothing reaches it today" is not a seal.

### Verified sound, recorded so it is not re-derived

- The reviewer scanned `syscall.Errno(1..3999)` for shapes a single `TrimSuffix` cannot handle and found five, all NetBIOS, Remoteboot, or print-spooler codes no file API in this package returns. The three network errnos a UNC root can actually produce were rendered through all three `base.go` sites: single terminator, no dangling separator.
- Every reachable Prisma and Swagger reason was measured against the real bridges rather than assumed, which is what establishes that F3 is latent rather than live.
- `packages/lint` has no member of the class; its only error-text joins are panics that append nothing.
- The `claims.mdx` linked-root sentence was verified clause by clause for six root shapes, link and non-link. Five of seven clauses hold and two are noted below.
- The new walk-failure composition case is portable: on Windows `containedProjectPath` refuses the backslash remainder, so `filepath.Rel` decides and both platforms yield the same base-relative path.

### Carried forward rather than repaired

Two clause problems the reviewer raised against the guide's linked-root paragraph are recorded here rather than edited again, because the sentence has been rewritten six times and each rewrite has introduced its own overshoot:

- "the link changes nothing else" is a universal that one state denies: a link chain past the resolver's bound is refused, where a real directory at the same path is read. The guide documents neither the bound nor that diagnostic.
- the paragraph speaks for all three artifact kinds, and only the two walkers resolve a link. A TypeScript claim root that is a link is accepted by the gate and then matched against the link rather than the directory, with no case covering it.

The second turned out to be a behavior question rather than a wording one, and measuring it produced **#1268**: a TypeScript claim rooted at a link selects nothing and deactivates in silence, because the gate accepts the link and the comparison against `base.Absolute` refuses every Program source. Measured before the repair: no diagnostic at all. It is implemented in `3381bde3b`, which completes the invariant #1258 documented for the two kinds that walk.

The first stays a wording problem and is left as written. The guide says the link changes nothing else about the addresses, which is what the sentence is about; the chain bound is a separate diagnostic that a real directory at the same path would not produce, and documenting a resolver's hop count in a user guide would name an implementation detail no author can act on.

samchon (commented):
## Individual Self-Review — `9ee765686..3381bde3b`

One read-only reviewer over the range. Advisory only; every disposition below is the main agent's.

**Ten candidates raised. All ten accepted.** Nine are repaired in `2af52e86e`; one is published as **#1269**.

| # | Where | Claim | Disposition |
| --- | --- | --- | --- |
| F1 | `base.go`, `typescript.go` | **defect, measured.** `resolvedBaseDirectory` is `populationWalkRoot`'s first line extracted without the check that followed it, so the kind that walks nothing got the resolver's answer unverified. A 35-link chain onto the project produced **zero diagnostics** from a TypeScript claim, over the same root a Markdown reference beside it refused | accepted, repaired |
| F2 | same sites | **defect, measured.** A link on an *ancestor* of the declared root is transparent to `Lstat` of the leaf, so nothing resolves it and the TypeScript comparison goes silent where both walkers succeed. That is the pnpm shape one level deeper | accepted, **#1269** |
| F3 | commit message, `base.go` | "a base that is not a link does no extra work at all" and "keeps the ordinary case free" are both false: one `Lstat` in the resolver and one in the check, per base, three times per rule run | accepted, repaired |
| F4 | `population_roots_..._test.go` | two stale counts in a doc comment the same commit edited, contradicting that commit's own source comment twenty lines away: nine versus thirteen join sites, three versus five text reasons | accepted, repaired |
| F5 | `base.go` | a clause about a parser's sentence sat on `causeText`, which no parser site reaches | accepted, repaired |
| F6 | `validation-plan.cjs` | "the two lanes exercise different code inside `resolveLinkedDirectory`" — the branches are identical; the difference is `os.Readlink`'s reparse handling | accepted, repaired |
| F7 | `test_support_test.go` | `linkDirectory` prefers `os.Symlink`, so on an elevated Windows runner every linked-root case exercises a symbolic link and the junction handling that is the reason the resolver exists is proven nowhere | accepted, repaired |
| F8 | `walk_failures_..._test.go` | the `/target/` anchor cannot fire on Windows against the native-separator leak its own doc names as historical | accepted, repaired |
| F9 | `linked_..._test.go` | #1268's acceptance rows 4 and 5 are pinned by nothing, and the case's asserted target belongs to the default Markdown base, so a repair that moved the TypeScript address would leave it green | accepted, repaired |
| F10 | `linked_..._test.go` | a case whose doc opens "This kind walks nothing" sits in a file named for walking | accepted, repaired |

**F1 is the finding, and it is the fourth time this cycle a repair carried its own defect.** The previous commit extracted a helper and left behind the half that made it safe. The verification now lives inside the resolver, so every caller shares it by construction rather than by being remembered, and `populationWalkRoot` is gone because it was exactly that resolver plus the check.

**F2 is published rather than repaired.** #1268's repair is the leaf case; this one needs a component-resolving helper with its own platform matrix for volume roots and UNC shares, and `filepath.EvalSymlinks` cannot be it because it returns a junction unchanged, which is why this package has its own resolver. The three adopter TSDoc blocks and the guide were narrowed in the same commit to promise only that the root itself may be a link, so nothing documented is false while #1269 is open.

**F3 is the same class as F1 in prose.** Two sentences claimed zero cost for a path that makes two syscalls, and the in-code comment about the *extra spelling* was correct while the one about the resolver was not.

### Verified sound, recorded so it is not re-derived

- No over-match from accepting a second spelling. Two roots reaching one directory both select the file, with keys `.../mirror\x00src/sale.ts` and `src/sale.ts`, which is the design `populationBase.address` documents; `matchesConfiguredTypeScriptFile` keys on the declared `Absolute`, so neither base can take the other's globs.
- Both TypeScript passes select the same set. `extendTypeScriptInventories` and `recordGovernedTypeScriptFiles` call the identical comparison and compose from the same base; the third caller of `configuredBases` for this kind is the gate, which correctly keeps handing it the declared path.
- The address stays composed from the declared base on every path, measured as `../mirror/src/sale.ts`.
- The `entry.resolved == entry.base.Absolute` guard is correct for the default base, a base inside the project, an absolute base, `D:/`, a missing root, a dangling link, and a link to a file.
- `causeReason` sits before the empty-reason fallback in both new sites, so a reason of `"."` yields the fallback sentence rather than an empty clause.
- `TestEachPlatformCallsItsOwnRootsAbsolute` can fail on both lanes, and its second and third assertions are not tautologies retained by accident: they catch a resolution or a message clause that stops reading the predicate.
- `model.go`'s `Path` comment is accurate for all four kinds, and no code reads a Swagger inventory's `Path`.
- `config.go`'s new terminator is safe: all six glob errors write none, and every sibling in that function ends with a period.

### Not verified

POSIX and Darwin execution. F1's reachability on Linux is derived from the resolver's bound against Linux's traversal limit rather than run, and F7's premise about the `windows-latest` runner is checkable only from that lane's log, which is why the junction case now builds the junction itself.

samchon (commented):
## Overall Self-Review — rounds 15 to 19

Solo, whole base-to-head diff. Rounds 1 to 14 are in two earlier reviews on this thread.

**Round 18 is the first one run the way the review skill actually prescribes**, and the change is worth recording because it cost this pull request real time. The Overall procedure is: perform one complete round, reproduce every suspected defect, then apply every sound improvement, and restart only if anything changed. Rounds 1 to 17 instead stopped at the first finding, repaired it, committed, and began again. Two costs followed: the rest of every round's surface went uninspected, which the **Whole surface** rule already forbids, and each commit restarted the whole CI matrix. The rule that reads as licence for that is being corrected in **#1270**.

| Round | What it found | Commit |
| --- | --- | --- |
| 15 | the project-relative promise survived in four more places than round 11's sweep reached, three of them the adopter TSDoc that #1251 §10 names and one the inventory `Path` field | `9f6e98020` |
| 16 | the interaction risk of the linked-base repair: one source belonging to two bases where it belonged to one. Measured as two messages, one per claim, and pinned | `83391879f` |
| 17 | two new comment claims false against the code: the resolver's cost, and the same code exercising "different code" per platform | folded into the review-8 batch |
| 18 | **four findings, collected then applied once.** The chain refusal told the default base to correct a `root` it never declared; `typeScriptMatchBases` kept a link its own comment said it discards; `unlistableBaseProblem` still explained two causes when one had moved | `5366ea94b` |
| 19 | **clean.** Nothing applied | none |

### Round 19, the clean round

Three sweeps, none of which produced a finding:

- **Every symbol this pull request introduced, counted at its call sites.** Fourteen symbols, no dead code, and the two with no direct case are exercised through the gates by the chain and linked-root cases.
- **Every discarded result in the new code.** All are range indices or existence checks; no error is swallowed.
- **Every shape of the new refusal rendered**, three kinds by six root spellings, including the default base branch added in round 18. No dangling separator, no doubled punctuation, no missing terminator, and the default base is named as the ttsc project root with the invocation as its repair.

### Recorded so the merge body is read correctly

The squash merge concatenates every commit message, and three of them assert a completeness that a later one disproves. `fe3ab5a57` says the doubled-terminator class is seven sites; `786a78b9c` says "the last two" and "the class is now empty"; `a66f9426d` says "doubling is now genuinely at zero". The class is thirteen join sites. Each claim was corrected by the next commit that counted rather than grepped, and the sequence is left in the history because that is what happened; a reader of the merge body should take the last count, not the first.

Two process faults are recorded for the same reason. A read-only reviewer's throwaway probe entered a commit twice, once through `git add -A` and once through `git add <directory>`; both were removed, and staging is explicit file paths only. And the rounds-repairing-one-at-a-time fault above.

### Still owed

A complete fresh full-scope discovery round for the campaign, exactly as #1251 §7 said before this cycle, because this cycle also proceeded without it under an explicit instruction.

samchon (commented):
## Individual Self-Review — `3381bde3b..2af52e86e`

One read-only reviewer over the range. Advisory only; every disposition below is the main agent's.

**Seven live candidates, all accepted.** Three more had already been repaired by the commit that landed while the review ran, and the reviewer recorded them with its own measurement rather than claiming them.

| # | Where | Claim | Disposition |
| --- | --- | --- | --- |
| F1 | `typescript.go` | **defect, measured.** The chain gate is asked of the default base, which `baseDirectoryProblem` deliberately returns on, and it refuses a population that works. Measured in an isolated copy: a project invoked through a 34-hop chain with no Markdown or Prisma population closes at zero diagnostics without the gate, and with it one refusal marks the whole default TypeScript population failed, so two genuine findings in the same project disappear | accepted, repaired |
| F2 | `typescript.go` | "the default base is never judged here" is false at HEAD, since the new check judges it | accepted, repaired |
| F3 | `base.go` | `populationRootLabel` says three callers where there are four, and that one reaches its fallback where two now do | accepted, repaired |
| F4 | `base.go` | "the gate answers first and this refusal is unreachable" on Darwin is true only for a declared root; the gate returns on the default base without stat'ing anything, so a walker reaches it there on every platform | accepted, repaired |
| F5 | `base.go` | the restatement suppression claims to cover an absolute declared root and the default base at once, and a UNC root on POSIX is neither: `filepath.Clean` collapses its leading slashes, so the spellings genuinely differ. The mirror rule in `describeBaseDirectoryProblem` was suppressing the restatement for that same shape, where the resolution did change the path | accepted, repaired, and unified |
| F6 | `walk_failures_..._test.go` | the anchor scans the whole message, including the cause this rule does not author, so in production it fires on a message that is right | accepted, repaired |
| F7 | `prisma.go` | the Prisma arm of the refusal is asserted by nothing, so its noun is free to be wrong | accepted, repaired |

**F1 is the third batch in a row to introduce a defect, and all three were in the same cluster.** The third stat state, the unverified resolver, and now a gate asked of a base that does not want it. The area is `describeBaseDirectoryProblem`, `resolvedBaseDirectory`, and the three gates around them, and every one of the three was found by measurement rather than by reading. The repair here removes a branch rather than adding one, which is the direction that has been going wrong.

**F5 turned two rules into one.** Restating a resolved path is worth doing only where it differs from the label, and that is the whole question; whether the project root was composed into the spelling is a different question that still gates the clause saying so. The predicate was answering both, and for a UNC root on POSIX it answered the first one wrongly.

### Verified sound, recorded so it is not re-derived

- Thirty renders across declared shapes and kinds, plus the default base: no dangling separator, no doubled punctuation, one terminator, and suppression firing exactly where the label equals the resolved path.
- No base produces two diagnostics. The kind-free default-base sentence emitted by both a walker and the TypeScript gate collapsed to one through `reportProblems`, measured; and after this batch the TypeScript gate no longer emits it at all for that base.
- Chain arithmetic: the resolver finishes 32 and fails at 33, the cases build 35, Linux stops at 40 and Windows at 63, and Darwin's 32 is what the cases' stat skip covers.
- The four strengthened cases can all fail, verified by mutation, including the location assertion whose mutant prints `src/sale.ts` and turns only that assertion red.
- `EvalSymlinks` returns a real junction unchanged while `os.Readlink` returns its target, measured, so both the resolver's comment and the lane comment are true.
- The terminator counts are exact at thirteen join sites and five text reasons, and `causeText`'s narrowed list holds for all eight of its own sites.
- The main agent separately measured that the F1 repair works: the same project now reports its two genuine obligations instead of one refusal.

### Recorded, not repaired

The F1 repair is a trade rather than a pure win, and the comment says so. Refusing the default base converted a possible silence into a certain error; leaving it out restores the silence for a project whose Program reports realpaths while ttsc was invoked through a long chain. The walkers still answer for that base, and a TypeScript population there has no walk to notice it, which is the residue #1269 already carries for the ancestor shape.

samchon (commented):
## Overall Self-Review — round 21, clean, on `0ec9a97c0`

The round that applies nothing, on the same immutable head the checks are green for. Rounds 1 to 20 are in three earlier reviews on this thread.

**Nothing was found that survived verification.** Three sweeps ran over the whole base-to-head diff:

- **Every message shape rendered.** 273 combinations at this head, and 459 in the wider sweep before it: three artifact kinds by seven root spellings, including a bare drive root, a UNC share, an absolute same-volume root, and the default base, each over every stat state and every cause shape. No dangling separator, no doubled punctuation, no missing terminator, no doubled path separator.
- **Every message fragment in the source cross-checked against the cases asserting it.** Seven fragments, each held by at least one case.
- **Every symbol this pull request introduced counted at its call sites.** No orphan, and no probe left in the working tree.

### The merge gate

| Gate | State |
| --- | --- |
| Ordinary CI | **25 of 25 pass** on `0ec9a97c0`, no failure |
| `go` (ubuntu) | pass, 7m36s — the eight permission-guarded cases actually run there |
| `windows-go` | pass, 16m29s — selected for this package for the first time, by #1264 |
| `evidence defenses` | pass, 25m29s — the feature suite and the benchmark parser that reads these messages by substring |
| Individual Self-Review | twelve, one per pushed implementation commit, every one recorded here as a formal review |
| Overall Self-Review | twenty-one rounds, this one clean |

### What the cycle closes

| # | What |
| --- | --- |
| #1248 | a root diagnostic names the property the author wrote, not a path derived from it |
| #1249 | a walk failure is spelled the way the file messages beside it are |
| #1254 | a population whose base could not be listed is failed, never healthy and empty |
| #1258 | a root that is a link is walked as the directory the gate accepted |
| #1264 | this package's Go tests run on Windows, where its path handling differs |
| #1268 | the kind that walks nothing reads a linked root too |

Four of these were found while implementing the two the cycle opened with, published, and implemented here under the instruction that opened it.

### What stays open, deliberately

- **#1262** — one physical schema in two Prisma populations is parsed twice. The repair redefines what `prismaInventoriesByDisplay` maps, and the wrong version silently empties a population, which is the class #1254 closed.
- **#1269** — a link on an ancestor of a declared root. #1268 repairs the leaf; this needs a component-resolving helper with its own platform matrix for volume roots and UNC shares. The adopter TSDoc and the guide were narrowed here to promise only what is true while it is open.

### What is owed

A complete fresh full-scope discovery round for the campaign, which #1251 §7 already said was owed before this cycle and which this cycle also skipped under instruction. **The campaign is active, not complete**, and the two open issues above are part of why.

### Honest notes for whoever reads the merge body

The squash concatenates every commit message, and three of them assert a completeness a later one disproves: the doubled-terminator class is stated as seven sites, then "the last two", then "genuinely zero". It is thirteen join sites. Take the last count.

Two process faults are recorded in full in the campaign's own ledger and one is corrected in **#1270**: rounds 1 to 17 repaired one finding at a time and restarted, which left the rest of each round's surface uninspected and restarted this pull request's CI matrix every time; and a read-only reviewer's throwaway probe entered a commit twice before staging became explicit file paths only.

Issue #1243, opened 2026-08-19
The LSP symbol provider holds a whole Graph for the process lifetime, and its build-only fields pin every AST


samchon:
## Problem

`packages/ttsc/internal/graphsymbols/provider.go` caches a built `*graph.Graph`
on the provider and keeps it until `Invalidate` is called:

```go
defer func() { _ = prog.Close() }()

pr.graph = graph.Build(prog)
pr.sources = graph.SourceTexts(prog)
```

The `Program` is closed, but the `Graph` outlives it — and the `Graph` carries
fields that exist only for the duration of a build and hold pointers into the
compiler's AST:

- `resolved map[*shimast.Node]*Target` — memoizes checker resolution, keyed by
  AST node (pre-existing).
- `docHosts map[string][]docHost` — the declarations the node pass found
  documentation on, each holding its `*shimast.Node` (added in the
  documentation-tag cycle).
- `docTagPositions`, `docHostPositions`, `bodyNodes`, `seen`,
  `edgeEvidenceFiles`, `baseNodes`, `selectedFiles` — all build-only, all
  documented as such in `packages/ttsc/internal/graph/graph.go`.

So closing the Program frees nothing those maps reference. For an editor session
this is the resident cost between invalidations, and it is larger than the graph
a consumer actually reads.

This is **not a regression** of the documentation-tag cycle: `resolved` already
pinned the same ASTs, and it predates that work. What the cycle did was add a
second field of the same class, which is what made the pattern visible.

## Evidence

The fields are declared build-only in their own comments — for example
`bodyNodes` is "build-only metadata and intentionally stays out of JSON dumps",
and `docHosts` says the edge pass reads it "rather than from a walk of its own".
Nothing clears any of them when the build finishes, and `BuildFiles`
(`packages/ttsc/internal/graph/build.go`) returns the `Graph` with every one
still populated.

`graphsymbols` is the only consumer that retains a `Graph` past the call: the
dump path marshals and drops it, and the serve path
(`packages/ttsc/cmd/ttscgraph/serve_shards.go`) clones only what a snapshot
needs. So the retention question is entirely about this provider — but the fix
belongs on the producer, because a consumer cannot know which fields are scratch.

**Not measured.** No profile has been taken, and the size claim above is
reasoning from what the maps hold, not a number. Anyone taking this should
measure first: build the graph for a large project, take a heap profile before
and after clearing the fields, and report both. If the difference is small, the
[...]

Issue #1242, opened 2026-08-19
The bundled viewer's legend is hand-written, so every new edge kind is missing from it until someone notices


samchon:
## Problem

`packages/graph/src/viewer/index.html` hard-codes its footer legend as literal
markup — one `<span>` per edge family, with the swatch colour written into a
`style` attribute:

```html
<span class="dot"
  ><span class="swatch" style="background: #3fb950"></span
  >value-call</span
>
```

The colours it duplicates live in `packages/graph/src/viewer/main.ts:40`:

```ts
const LINK_COLORS: Record<string, string> = {
  "value-call": "#3fb950",
  "type-ref": "#f5b042",
  "doc-ref": "#c07de0",
  heritage: "#6ea8ff",
};
```

Nothing connects the two. Adding an edge family means editing both, and the
markup is the half a reviewer does not think to open — the `doc_ref` family added
in the documentation-tag cycle rendered in the viewer with no legend entry until
a review read the two files side by side.

The website viewer does not have this problem: its legend and its sidebar filter
both iterate `LINK_COLORS` (`website/src/components/graph/TtscWebsiteGraphViewerSidebar.tsx`,
`Viewer3D.tsx`), so a new family appears in both automatically. The bundled
viewer is the odd one out.

## Evidence

At `master` before the fix, `LINK_COLORS` carried four families and
`index.html` listed three. Both files are in the same directory, bundled by the
same build step (`packages/graph/build/bundle-viewer.mjs`), and neither refers to
the other.

The same shape recurs one layer up: `DISPLAY_KIND` — the map folding wire kinds
onto these families — exists in **three** copies
(`packages/graph/src/reduce.ts`, `website/src/components/graph/TtscWebsiteGraphReduce.ts`,
`benchmarks/graph/src/TtscBenchmarkGraphReduce.ts`), each carrying a comment
telling the reader to keep it in sync with another. They are currently
byte-identical, and they are identical because a reviewer diffed them, not
because anything enforces it.

## Approach

The invariant: **one definition of the edge families, and every surface derives
from it.**

For the legend specifically, the fix is small: render it from `LINK_COLORS` at
startup, the way the website already does, and delete the markup. The viewer
already builds DOM at runtime, so there is no reason for this list to be static.

The three `DISPLAY_KIND` copies are the larger half and need a decision rather
than a mechanical move. They exist because three consumers have different module
systems and build steps — a browser bundle, a Next.js site, and a Node benchmark
[...]

Issue #1241, opened 2026-08-19
`pnpm format` rewrites 250 formatter fixtures and exits non-zero, and no CI lane notices


samchon:
## Problem

Running `pnpm format` on a clean checkout of `master`:

- exits **123** during its Go pass, and
- before failing, rewrites roughly **250 files under `packages/lint/test/format/`**
  that the caller never touched.

Both halves are hazards. A contributor who runs the repository's own formatting
command — which `.agents/skills/development/SKILL.md` instructs them to run
before every ordinary commit — ends up with a couple of hundred unrelated files
modified. A `git add -A` after it carries them into the commit. That happened
during the documentation-tag cycle: the commit surface went from 49 files to 299,
was caught by re-reading the surface rather than by any tool, and was removed by
amending before the pull request ever carried it.

The repository and its formatter currently **disagree** about those files:
`master` keeps them tab-indented, and the formatter wants them two-space. Nothing
detects that, because no CI workflow runs a format check — `grep -rn "format"
.github/workflows/*.yml` finds none.

## Evidence

```
$ git status --porcelain | wc -l
0
$ pnpm format
… (prettier pass, all unchanged) …
> @ttsc/station@0.28.0 format:go
> git ls-files -z "*.go" | xargs -0 ./.vscode/gofmt-2spaces.sh -w
 ELIFECYCLE  Command failed with exit code 123.
$ git status --porcelain | wc -l
250
$ git diff --stat | tail -1
 250 files changed, 3777 insertions(+), 3777 deletions(-)
```

Every one of the 250 is under `packages/lint/test/format/`. The diff is
tab-to-two-space on the Go indentation:

```diff
-	// Two short arrows that fit flat: explode (the core async.then idiom).
-	t.Run("two_arrows_explode_even_when_flat_fits", func(t *testing.T) {
-		assertFormatUnchanged(t, `promise.then(
+  // Two short arrows that fit flat: explode (the core async.then idiom).
+  t.Run("two_arrows_explode_even_when_flat_fits", func(t *testing.T) {
+    assertFormatUnchanged(t, `promise.then(
```

**Root cause of the rewrite.** `.vscode/gofmt-2spaces.sh` runs `gofmt` and then
replaces **every tab in the file** with two spaces:

```bash
if [ "$write" = true ]; then
  gofmt "${args[@]}"
  if [ "${#files[@]}" -gt 0 ]; then
    perl -0pi -e 's/\t/  /g' "${files[@]}"
  fi
```

[...]

Issue #1240, opened 2026-08-19
`graphdump` has no test at all, so its capability declaration is unprotected and its `run` cannot be driven


samchon:
## Problem

`packages/ttsc/cmd/graphdump` is the repo-internal one-shot that feeds the
benchmark graph viewer pipeline. `git ls-files packages/ttsc/cmd/graphdump/`
returns exactly one entry — `main.go`. It has never had a test.

That became load-bearing in the documentation-tag cycle. `graphdump` runs the
same declaration walk as the shipped `ttscgraph dump`, so its output carries
every documentation tag the graph records, but it declared only
`CapabilitySourceDigests`. Under the capability contract
(`packages/ttsc/internal/graph/provenance.go:40-48`) an absent claim means *the
producer never looked*, so a consumer honouring it reads a tagged declaration as
citing nothing — a confident wrong answer rather than a missing one. The
declaration was corrected to
`[]string{graph.CapabilitySourceDigests, graph.CapabilityDocTags}`
(`packages/ttsc/cmd/graphdump/main.go:69`), and **nothing verifies it**.

It is a *separate literal* from the shipped command's `fullSnapshotCapabilities`
(`packages/ttsc/cmd/ttscgraph/serve.go:63-69`), so the test that now guards that
one — `run_dump_declares_the_capability_its_output_carries_test.go`, verified by
mutation to fail when the member is removed — gives `graphdump` no protection at
all. Deleting `graph.CapabilityDocTags` from `main.go:69` today breaks nothing.

## Evidence

The gap was found by reading the two producers side by side during a review of
the documentation-tag cycle, then confirmed:

```
$ git ls-files packages/ttsc/cmd/graphdump/
packages/ttsc/cmd/graphdump/main.go

$ go test ./cmd/graphdump/
?   github.com/samchon/ttsc/packages/ttsc/cmd/graphdump   [no test files]
```

**Why it has no test is a structural fact, not an oversight.** Compare the two
commands:

| | `cmd/ttscgraph` | `cmd/graphdump` |
| --- | --- | --- |
| Entry | `run(args []string) int` — takes argv | `run() int` — takes nothing |
| Flags | own `flag.NewFlagSet` per subcommand | global `flag.CommandLine`, parsed inside `run` |
| Output | package-level `stdout`/`stderr` vars, swappable | `fmt.Println` to the real `os.Stdout` |

`cmd/ttscgraph` declares its I/O as package-level seams *specifically* so command
[...]

PR #1253, opened 2026-08-19
fix(campaign): citation follow-ups — artifact nodes, formatter integrity, and four unguarded surfaces

samchon:
Entry point: **#1250** (handover, revised).

This pull request owns the **complete accepted cycle** of citation follow-ups as
one unit. It is opened before any code exists so that overlapping work is visible
here rather than in an unannounced branch. Verification is pending for every item
below; nothing in this body is a completion claim.

## The cycle

| # | What | Order |
| --- | --- | --- |
| #1241 | `pnpm format` / the `gofmt-2spaces.sh` wrapper and its CI gate | 1 |
| #1242 | The bundled viewer's hand-written legend, and three `DISPLAY_KIND` copies | 2 |
| #1240 | `graphdump` has no test, so its capability claim is unprotected | 3 |
| #1244 | The benchmark's structural report counts three edge families of many | 4 |
| #1243 | The LSP symbol provider retains build-only state that pins every AST | 5 |
| #1239 | Materialize artifact units as graph nodes — the lint transport | 6 |

#1239 is ordered last because #1242 is cheaper before it introduces new node
kinds and display families. The rest are independent of each other.

## Intent

Each issue keeps its own acceptance matrix, positive/negative/boundary cases, and
regression coverage inside the combined diff. Issues surfaced while implementing
are published and resolved here; no new discovery round is run, per the
assignment and #1250 §6.

## Deferred

Nothing yet. Any issue that implementation disproves, narrows, or externally
blocks will be reopened on its own thread and recorded in a review here rather
than silently dropped.

## Verification

Pending. Local evidence, CI results, Individual Self-Reviews per commit, and the
Overall Self-Review rounds are recorded as formal pull-request reviews on this
thread, not by editing this body.


samchon (commented):
## Individual Self-Review — `55ad83b27` (`fix(graph): give the edge families one definition every surface derives from`)

Resolves #1242 and #1255. One read-only reviewer over the parent-to-commit diff.
Six candidate findings survived the reviewer's own reproduction; I adjudicate all
six as **accepted**, one of them as a new issue. Nothing was rejected.

### Accepted — fixed in a follow-up commit on this branch

| # | Finding | Why it stands |
| --- | --- | --- |
| 3 | Deleting the `renderLegend()` call leaves the new test green | Reproduced: the case written to stop the bundled viewer shipping without a legend does not assert that a legend is produced. Every legend assertion is negative. This is the exact #1242 regression, and it is now *easier* to hit than before, because static markup could not silently disappear. |
| 4 | The `index.html` hardcode guard is a raw substring match | Reproduced twice: a hand-written swatch spelled `#3FB950` in uppercase, or `rgb(63,185,80)` with the label `calls / new / jsx`, both evade it. The commit's mutation claim holds only for a byte-for-byte restoration of the deleted markup. |
| 6 | `NOT_IN_A_DUMP` is a second hand-maintained copy of a vocabulary that already exists | `packages/graph/src/structures/TtscGraphDumpEdgeKind.ts` already declares exactly the ten kinds a dump can carry, and `graph_kind_contracts_match_their_producers_test.go` already holds it against the Go producer. Deriving that set again by subtracting a hand-written table is precisely the duplication this commit argues against. |
| 2 | `website/src/content/docs/graph/viewer.mdx` still says the viewer folds edges into **four** families | The development skill requires the matching docs page to move in the same change. A reader now sees five swatches and five filter chips against a list of four. |
| 7a | The unknown-kind negative twin checks only the bundled colour map | The positive check covers all three surfaces; the twin should too. |
| 7b | `renderLegend` is not idempotent — a second call prepends a duplicate set | Latent only (called once), but free to remove while the function is being made testable. |

### Accepted — published as its own issue

**Finding 1: the same root cause one axis over.** `exports` edges run *from a
`module` node*, `reduce()` filters no node kind, and neither viewer's
`NODE_COLORS` names `module`. Worse than the edge case this commit fixed: the
bundled node fallback `#8b97a8` is byte-identical to `NODE_COLORS.variable`, and
the website's `#64748b` is byte-identical to its own `variable`. A module node is
not merely unnamed — **it is drawn as a variable**. Reproduced end to end through
the real `ttscgraph dump` binary: two of seven surviving payload nodes were
modules.

This is a node-kind defect rather than an edge-family one and deserves its own
evidence and decision (colour it, or filter module nodes out of the viewer), so
it is #1256 rather than a silent widening of this commit. It is in this pull
request's scope and will be resolved here.

### Recorded, not acted on

- **#1242 acceptance item 3** asks for the built page to be opened, not only
  built. No browser is available in this workspace and it carries no headless DOM
  (no puppeteer, playwright, jsdom, or happy-dom). What was actually done, and
  what the commit message should have said: the **built** `lib/viewer/viewer.js`
  was executed in a VM against a minimal DOM stub, and it produced the five
  `span.dot` entries with the right colours in the right order ahead of the static
  note. The durable substitute is finding 3's positive assertion.
- **The nine published demo payloads** under `website/public/graph/` predate both
  `doc-ref` and `exports`, so the new legend entry and filter row are inert on
  ttsc.io until they are regenerated. Pre-existing — `doc-ref` is already in that
  state — and not caused here.

### Confirmed sound by the reviewer

`contains` and `dispatches` genuinely cannot reach a dump and `exports` genuinely
can, verified against the Go producer and a real dump; the `exports` fold is an
identity, so no committed payload, snapshot, or fixture changed meaning; the
website sidebar produces the new filter row by construction and no persisted
filter state is invalidated; both new colours are separable from their viewer's
unknown fallback and from the other four families; `renderLegend` runs before the
fetch and before the `#graph` null guard, so a failed load still shows the legend,
and `append(swatch, kind)` creates a text node with no injection surface; the CSS
still styles the runtime-created nodes; six mutations are correctly caught,
including dropping the label entry and adding a sixth family to only one map; the
refactor onto `viewerReducers.ts` dropped no assertion and strengthened one.

samchon (commented):
## Individual Self-Review — `cce2ba07a` (`fix(scripts): make the Go format wrapper preserve literals and finish its write`)

Resolves #1241. One read-only reviewer over the parent-to-commit diff. Ten
candidate findings; all adjudicated, six accepted and fixed in `bbcee9e28`.

### Accepted and fixed

| # | Finding | Disposition |
| --- | --- | --- |
| 1 | **Blocking:** `scripts/ci/gofmt-wrapper.test.cjs` is unclaimed in `scripts/ci/test-owners.cjs`, so the `typecheck` lane fails on every plan | Reproduced locally (`node --test scripts/ci/test-owners.test.cjs` red). Claimed. The branch was red from this commit onward and would have stayed red through merge. |
| 2 | `gofmt -w <directory>` writes every file beneath it and the normalization never runs — exit 0, tree tab-indented | Reproduced. This is the commit's own failure class, reachable through an argument shape it did not consider. Directory arguments now expand into the normalized set, and a named path that does not exist reaches gofmt instead of being dropped, which was a second silent success. Both pinned, both mutation-verified. |
| 4 | The apostrophe case is not discriminating — it passes with the whole comment arm deleted | Reproduced: only the backtick case failed under that mutation. The tab now sits *between* the two apostrophes, which is the only position a single-line mis-lex can hide in. Re-mutated: both cases fail now. |
| 5 | The `broken.go` assertion's stated rationale is false | Correct, and it matters: the pass runs over the files that were **named**, not the subset gofmt accepted, so an unparseable file *is* normalized. The fixture now carries a tab and the case asserts that outcome, with the decision stated rather than implied. |
| 6 | The wrapper comment and the commit message overstate the apostrophe hazard | Correct. Only the backtick is unbounded — the rune and string arms cannot cross a newline. The comment now separates the two cases. The commit message is immutable; this review is the correction. |
| 7 | The gofmt-failure case asserts only "non-zero", while the claim is that gofmt's status is reported | Now `assert.equal(result.status, 2)`. |

### Accepted, recorded, not acted on

- **3.** `packages/lint/tools/widthgen/main.cjs` carries a second copy of the
  normalization rule and now diverges from it. It emits two string literals (a
  version number and an emoji pattern whose non-ASCII is all `\x{...}` escapes),
  neither carrying a tab, so the cheap substitution still agrees with the
  specification. Routing a Node generator through a bash wrapper for a condition
  that holds is not worth the dependency; the condition is written down instead,
  with what to do if it stops holding.
- **8.** An unparseable file with an odd number of backticks ends up half
  tab-indented. Scoped to the one file gofmt refused, and loud (exit 2). The
  boundary that governs it — the pass runs over named files, not accepted ones —
  is now stated in finding 5's case.
- **9.** Backtracking on the general-comment arm is quadratic in unterminated
  `/*` but unreachable from valid Go, and the stdin path is fed only gofmt's own
  output. Measured 5s at 20,000 unterminated `/*` in a synthetic 229 KB file.
- **10.** CRLF, BOM, empty file, non-UTF-8, and filenames with spaces or `@` all
  behave; none is pinned. Recorded rather than covered — each is a property of
  `perl -0777 -i`, not of the change.
- #1241 acceptance item 4 (a CI lane fails on an unformatted file) rests on
  `scripts/ci/format-check.cjs`, which has no test of its own. The new suite
  gates the wrapper, not the gate. Outside this commit's narrowing.

### Confirmed sound by the reviewer

The lexer was proved differentially against a `go/scanner`-based reference
normalizer: **0 mismatches over all 3,786 tracked `.go` files** and **0 over
5,799 adversarial fuzz cases** (tabs in every literal form, escapes, `//` and
`/*` inside literals, backticks and quotes inside comments, unpaired quotes,
`/*/`, adjacent literals), with the harness sanity-checked by injecting a
corrupted hash. No valid Go file could be constructed that the wrapper mangles.
Whole-tree round-trip through `gofmt -w` plus the normalization: **drift 0**.
Idempotency holds including gofmt's doc-comment code blocks, which it re-tabs
every pass. gofmt aligns with spaces, so preserving literal tabs cannot break an
alignment column. The `status=0; gofmt … || status=$?` idiom, the empty-list
guard, `pipefail` on the stdin path, and quoted array expansion all behave. All
three of the commit's mutation claims reproduce exactly.

samchon (commented):
## Individual Self-Review — `45d6feda1` (`test(graphdump): assert the capability claim against the command's own output`)

Resolves #1240. One read-only reviewer over the parent-to-commit diff. Five
candidate findings; four accepted and fixed in `bbcee9e28`, one recorded.

### Accepted and fixed

**1. Blocking, and it invalidated the issue's own CI premise.** Two halves:

- `packages/ttsc/cmd/graphdump` became the first `_test.go` owner in that
  directory, and `scripts/ci/test-owners.cjs` claims no such owner, so the
  `typecheck` lane failed. Reproduced by replaying `ownershipFailures()` against
  `git ls-tree` at the parent: this commit adds exactly one new failure.
- Even claimed, **the tests would not have run in CI**. `scripts/test-go-graph.cjs`
  lists `internal/graph`, `internal/graphsymbols`, `cmd/ttscgraph`, and
  `cmd/graphbench` — not `cmd/graphdump`, and no other `test-go-*` runner covers
  it. #1240's consequence table asserts "`pnpm test:go` covers it"; that premise
  was false and the commit did not verify it. The runner now covers the package.

**2. `graph.CapabilitySourceDigests` was still unguarded** — the same defect
#1240 filed against, one member over. Reproduced by overlay mutation: deleting it
from `main.go:83` left the suite green. The declared set is now compared whole
and each member paired with the output backing it, so a capability added to the
list arrives with its evidence or fails.

**3. `-h` regressed from exit 0 to exit 2.** `flag.ContinueOnError` returns
`flag.ErrHelp` for `-h`, and `run` mapped every parse error to 2. The reviewer
reproduced both sides, including a probe of the old `ExitOnError` behavior. The
commit message claimed the shipped path was unaffected without mentioning it.
`flag.ErrHelp` now returns 0, pinned in the flag case.

**4. Every failure exit was untested.** Reproduced: replacing all three
`return 1` with `return 0` left the suite green. The pipeline's `execFileSync`
raises only on a non-zero exit, so a 0 hands it an empty string and it dies in
its JSON parse instead of surfacing the diagnostic the command already wrote. A
third case pins exit 1, empty stdout, and stderr naming the tsconfig.

### Recorded, not acted on

**5.** `--pretty` and the `--cwd` default are unexercised. Both are pre-existing
flags this commit did not change, and the seam work is what needed coverage.
Recorded so the next author does not assume they are pinned.

### Confirmed sound by the reviewer

The shipped path is byte-identical: exit 0, valid JSON on stdout, trailing
newline preserved (verified with `xxd`), flag names, defaults, and
`--cwd`-omitted behavior unchanged; `fmt.Fprintln(stdout, …)` with
`stdout = os.Stdout` matches `fmt.Println`. Acceptance item 2 reproduces — the
failure names the missing member. Four further mutations are caught: dump written
to stderr, `return 0` on parse error, and `fs.SetOutput(stderr)` removed, the
last confirming the seam ordering is genuinely exercised rather than incidentally
passing. The docTags half is not a tautology: it is paired with a tag-text
assertion and a positive `"docTags":` control, and the colon match avoids the
capability-name collision #1240 warned about. Test isolation holds under
`-count=2`, on `t.Fatalf` and panic, and `t.TempDir()` cleanup is clean on
Windows because `prog.Close()` fires before the helper returns. `-race` could not
run in this environment (CGO disabled); nothing in the change introduces
concurrency. The duplicated `writeGraphdumpFile` is unavoidable — two separate
`package main`s, and the repository has no shared Go test-helper package.

samchon (commented):
## Individual Self-Review — `9b15c12b6` (`perf(graph): release the build scratch a returned Graph no longer reads`)

Resolves #1243. **No correctness defect.** The reviewer traced every reader of all
nine cleared fields across the whole `packages/ttsc` module — production and tests
— and found none after `BuildFiles` returns. Four findings, all accepted, three
fixed in `f44e7b3e5`.

| # | Finding | Disposition |
| --- | --- | --- |
| 1 | The case discriminated **8 of 9** fields | Accepted, fixed. `baseNodes` is nil on arrival in a complete build — it is passed in, not built — so `Build(prog)` could never fail on it, and it is both the largest map and the one #1243 flagged as the caution. A partial build covers it. The commit message's "naming each field" was therefore wrong. |
| 2 | Struct drift uncaught — nine hand-written names | Accepted, fixed with the reviewer's own prototype: reflect over `*Graph` and require every unexported reference field to be released. `reflect.Value.IsNil` reads unexported fields legally. A tenth field is caught without anyone remembering, which is how `docHosts` arrived beside `resolved` in the first place. |
| 3 | Stray `”` in the new doc comment | Fixed. |
| 4 | The message credited `internal/graphsymbols` for cases that live in `packages/ttsc/test/driver/` | Accepted. The reviewer ran them: `go test ./test/driver/` **ok, 111s**, `TestGraphSymbolsProviderReflectsEditsAfterInvalidate` passes — so #1243 acceptance item 3 is genuinely met and only the named command was wrong. Recorded here and in `f44e7b3e5`. |

### The measurement reproduces

The reviewer built an independent probe and measured release-skipped versus
release-in-place, one fresh process per project:

| project | nodes | edges | reviewer | commit |
| `@ttsc/lint` | 1,301 | 1,974 | 38.3 → 3.8 MB | 38.1 → 3.7 |
| `ttsc` | 1,664 | 5,003 | 52.6 → 5.0 MB | 52.5 → 5.0 |
| `@ttsc/graph` | 960 | 2,143 | 58.2 → 3.9 MB | 58.2 → 4.0 |

Counts match exactly, figures within 0.2 MB. One datum worth keeping: the
provider caches `SourceTexts` beside the graph, so real session retention is
4.8 / 5.9 / 7.7 MB — still 8–12× better end to end, with the sources map now the
dominant remainder. That is a separate follow-up and outside #1243.

### Confirmed sound

All five `lookupNode` call sites are in `edges.go`/`exports.go`, inside
`addEdges`; every cleared field's readers are unexported, so no consumer outside
the package can reach them at all. `MarshalDump`, `NewDump`, `GitIgnoredFiles`,
and `NodeFile` read none of them, and three existing dump tests already construct
a `Graph` with every scratch field nil. `prior.nodes` in the shard loop is the
caller's own map, so clearing the Graph's reference cannot affect it.
`shimast.` appears once in `graph.go` — on `resolved` — which is why the floor is
the graph itself. No `Graph` is constructed anywhere that would skip the release.

samchon (commented):
## Individual Self-Review — `bbcee9e28` (`fix(graph): close the review findings on the first four cycle commits`)

Sixteen candidates, every one reproduced by the reviewer. Six accepted and fixed
in `cc7f06ce1`; the rest are recorded below with their dispositions.

### Accepted and fixed — regressions that commit introduced

| # | Finding | Why it stands |
| --- | --- | --- |
| 6, 7 | `gofmt-2spaces.sh -w Makefile` replaces the Makefile's semantic tabs | Reproduced. Widening the positional case left `[ -f ]` as the only test, so the normalization ran over a file gofmt rejected — the data-loss class the wrapper's own header exists to prevent. Two directory mismatches went with it, in opposite directions: `-type f` skipped a symlinked `.go` gofmt had just written, and no dot test edited a `.golden.go` gofmt never read. The set now mirrors gofmt's own predicate. |
| 1 | The website's new `UNKNOWN_NODE_COLOR` is invisible and reads as "dimmed" | `#cbd5e1` is **1.43:1** against the scene's `#f7fbff`, where the `#64748b` it replaced was 4.6:1, and it sits (19,13,9) from `DIMMED_NODE` — so during a spotlight an unknown node and a dimmed one were the same picture. I cleared a name collision and created two worse ones. Now `#334155`. |
| 5 | `renderLegend(document)` does not typecheck against the real `Document` | Reproduced with the repo's own TypeScript. Invisible because `packages/graph/tsconfig` excludes `src/viewer` and esbuild erases types without checking. Widened to `unknown[]`, which the reviewer verified typechecks clean under `--strict --lib dom`. |
| 2 | The fallback constants are pinned at their declaration and nowhere else | Reproduced: putting `?? "#64748b"` back at a scene call site, or `?? "#8b97a8"` in the bundled viewer, is exactly #1256 with a green suite. The call sites are read now, and the mutation fails. |
| 14 | The legend case cannot see an invisible legend | Reproduced: dropping `className` leaves five zero-size inline spans — in the DOM, invisible, and green. The classes are asserted now. |
| 10, 13 | Three statements that were false rather than stale | `kindsIn` appends any present kind the order list omits, so `module` always had a chip — it came last, in the wrong colour. The colour half of #1256 was real; the chip half was not. The viewer guide still listed seven node kinds without `module`, and the legend markup still pointed at `main.ts` for a map that moved. |

### Recorded, not acted on

- **4.** `test-owners.cjs` proves a claim exists, never that the named runner runs
  the package — the exact class `bbcee9e28` fixed one instance of. Reproduced:
  deleting `./cmd/graphdump/...` from the runner while leaving the claim passes
  25/25. The reviewer audited every other `go:` owner and found no second live
  orphan. This is a real gap in a CI gate and deserves its own issue rather than
  a silent widening here.
- **8.** Two of graphdump's three failure exits remain unpinned (`prog == nil`,
  `MarshalDump` error); the commit's "returning 0 from every error path" fails
  only because the first is covered. Both need a seam to reach.
- **9.** `dumpVocabulary`'s regex is weaker than the Go contract test it cites:
  a camelCase union member is invisible to it. Inherited from the deleted
  `wireKinds`, not new.
- **3.** `UNKNOWN_LINK_COLOR` is asserted nowhere.
- **11.** `-w` with no positional exits 0 silently. Load-bearing for `format:go`,
  which runs the command once with no files on an empty `xargs` input.
- **15.** Four `.tsx` lines drifted past Prettier's width. `.tsx` is outside the
  format gate's globs entirely.
- **16.** The widthgen comment names the escape property; the operative reason is
  `JSON.stringify`, which escapes a literal tab regardless of content.

### Confirmed sound

Every suite green on the committed tree. Eight of the ten mutation claims in the
commit message reproduce exactly as stated; the two qualified above are findings
2 and 8. The `WeakSet` guard holds the element weakly and re-renders a replaced
legend correctly. A fresh esbuild is byte-identical to the committed bundle and
carries the five families in legend order. The bundled dark palette is sound —
`#565f6b` is 3.0:1 on `#0a0c10` and 2.2:1 from `variable`. Shell-argument
robustness was checked across spaces, a directory named `weird.go`, `--`, a
leading-dash filename, mixed flags, 405 real files, and a 157 KB argv. The
`NOT_IN_A_DUMP` removal is sound: the Go contract test holds both dump unions
against the producer, so a reintroduced `contains`/`dispatches` fails loudly.

samchon (commented):
## Overall Self-Review — rounds 1 and 2

Solo, whole base-to-head diff (72 files, ~4,000 insertions), fresh each round.
Both rounds found defects, so neither is the clean round; a third begins after
CI settles on the current head.

### Round 1 — seven findings, all fixed

| # | Finding | Severity |
| --- | --- | --- |
| 1 | `publishArtifacts` resolved `@ttsc/lint/bin/ttsclint`, and **`@ttsc/lint` ships no binary at all** — it ships Go source ttsc builds into a content-cached sidecar. The resolution could never succeed, so the entire consumer path was dead code. | **Critical** |
| 2 | The resident MCP session never asked for artifacts. A one-shot dump answered a citation with what it names while the tool an agent actually talks to answered with a token. | **Critical** |
| 3 | `TtscGraphDumpNodeKind` gained six members and neither viewer coloured them — caught by the totality gate this pull request added two commits earlier, which is what it was for. | High |
| 4 | `loadGraph` and the session asked for artifacts and the `dump` CLI did not, so the same project answered differently through three entry points. | Medium |
| 5 | A temp directory leaked per `publishArtifacts` call, and `loadGraph` is a library entry a caller may run in a loop. | Medium |
| 6 | Three more copies of the dump schema version, spelled as literals, failed the day the schema moved — the third copy #1250 predicted, in two files. | Medium |
| 7 | The evidence publisher and the new `resolveCapabilityPlugins` seam had no tests. | Medium |

Finding 1 is the one worth naming twice. It is the failure mode this repository
keeps meeting: a path that looks right, degrades silently by design, and
therefore reports nothing when it never fires. The fix is the seam the design
predicted — `ttsc` now publishes `resolveCapabilityPlugins`, the question
`ttscserver` already asks for `capabilities.lsp`, asked from inside the launcher
where no other tool could reach it.

### Round 2 — one finding, fixed

**The incremental path did not apply the artifacts.** Three of four build sites
did; the fourth is the partial rebuild the shard path runs on every ordinary
edit. A partial re-resolves the outgoing facts of the files it selected, and a
citation is one of them — so editing a declaration that cites a document section
rebuilt it with that edge missing. Correct at startup, wrong after the first
edit, right again after a restart, which is the worst shape a graph defect takes.

### Cost, measured

#1239 makes the cold-index number an acceptance criterion and says a regression
is a reason to stop. Measured on this repository:

| | |
| --- | --- |
| `ttscgraph dump` on `@ttsc/lint`, no artifacts | 1521 / 1797 / 1858 ms |
| the same with `--artifacts` | 1637 / 1830 / 1396 ms |
| `graph-nodes` on a project declaring no publisher | 820 / 890 ms, **no Program built** |

The producer's own cost is inside run-to-run noise: applying artifacts is a map
walk over a set bounded by the project's documentation. The verb is the real
addition, and its own gate is what keeps it cheap — a project whose config
declares no publishing rule never builds a Program to reach its empty answer, and
`evidence/graph` declares `NeedsTypeChecker() false`, so a project that does
publish pays a parse and bind rather than a checker.

The one cost not inside these numbers is the first plugin build on a cold cache,
which is the same build `ttsc check` pays on that project and is content-cached
after it.

### Local verification on the current head

Go: `./internal/graph/... ./cmd/ttscgraph/... ./cmd/graphdump/...` and the
evidence suite pass. TypeScript: `@ttsc/test-graph` is **76/76** against freshly
built native binaries, `@ttsc/graph` and `tests/test-graph` typecheck clean, and
the CI gates (`test-owners`, `validation-plan`, `go-test-runners`,
`gofmt-wrapper`) pass.

### Still open, and recorded rather than hidden

#1239's acceptance items **14 and 15** — refreshing the index when a Markdown
heading or `lint.config.ts` changes in a resident session — are not implemented.
The artifacts are read once per resident child, which is stated at the call site
and in the session's own field comment. They describe documents the compiler's
Program never read, so a document edit moves them and not one compiler fact:
that is a second invalidation with its own inputs, and inventing half of it would
be worse than naming it. #1239 stays open carrying exactly that.

samchon (commented):
## Overall Self-Review — rounds 3 through 8

Rounds 3, 4, and 5 each found a defect, so none of them is the clean round.
Rounds 6, 7, and 8 found none. CI found one on its own, which is recorded here
because a review that misses what CI catches is a review with a gap.

### Rounds 3–5, four findings, all fixed

| Round | Finding | Severity |
| --- | --- | --- |
| 3 | An alias could claim another artifact's own address, because aliases were registered in the same pass as addresses. A citation of the shadowed address then resolved to the wrong node — worse than the token it replaced, because the token never claimed to know what it named. | High |
| 4 | **The verb was invoked without `--plugins-json`.** A lint sidecar finds its own configured entry in that manifest; with none it binds an empty rule configuration, so no rule declared the publisher and the answer was always `[]` — for every project, indistinguishable from one that genuinely publishes none. | **Critical** |
| 5 | **The shard projection rejected every artifact.** It assigns each node to the source that owns it and fails closed otherwise, which is right for a declaration and impossible for an artifact whose file is a Markdown document, a Prisma schema, or nothing at all. The resident session failed to start for exactly the projects the feature is for. Fixing it uncovered a second guard on the metadata shard reading "not external" as "authored", in both the Go producer and the TypeScript client. | **Critical** |
| 5 | `view` was the one entry point still not asking for artifacts, so the viewer would have drawn a graph missing precisely the nodes it had just been given colours for. | Medium |

Round 4 was demonstrated rather than argued, on a scratch project needing no
contributor build:

```
with    --plugins-json  ->  [{"insert":"abstract",...}, ...]
without --plugins-json  ->  []
```

### CI's finding

`ttsc native defenses` failed on `test_ttsc_utility_plugins_descriptors_own_separate_native_source_directories`,
which compares `@ttsc/lint`'s capability object with `deepEqual`. Adding
`graphNodes` to the descriptor failed it — the case doing exactly its job, since
a capability the host stops declaring is a verb a consumer silently stops being
able to ask for. Declared in the pinned set.

### Rounds 6, 7 and 8, clean

Re-read every changed production file, and traced: every `spawn` of `dump` and
`serve` (all five now publish); node-id parsing for the three address grammars
(`parseTtscGraphNodeId` returns undefined without an unescaped `#`, and
`resolveGraphHandle` tries the exact id first, so an address resolves before any
name heuristic); the memory layer's file-key index and owner resolution (an
artifact's file is never a `.ts` path, so no declaration can resolve its owner to
one); `reduce`'s id rewriting and degree pruning; the source reader, which fails
closed on a file with no digest and therefore cannot inline a Markdown body; the
incremental node index, which never deletes an artifact because its file is never
in the selected set; and the descriptor assertions across both test suites.

Also checked and found sound: the Go kind-contract test already holds both dump
unions against the producer's constants, so the six new kinds are machine-checked
on both sides; evidence populates `Aliases` only for TypeScript units, so the
alias path is dormant today and the shadowing guard is for the next contributor;
and no test or snapshot pins a capability set other than the one CI named.

### Verification on this head

`@ttsc/test-graph` **77/77**. Go: `internal/graph`, `cmd/ttscgraph`,
`cmd/graphdump`, and the evidence suite all pass. The four CI gate tests pass
locally (37 assertions). Prettier reports no drift over the 31 changed
`.ts`/`.md`/`.mdx` files, and all 25 changed `.go` files round-trip through
`gofmt-2spaces.sh` unchanged.

samchon (commented):
## Overall Self-Review — rounds 9 and 10

### Round 9, two findings, both fixed

**The artifact vocabulary was held together by hand across two Go modules and a
TypeScript package.** The rule API publishes the six kinds, the compiler host
accepts them, and the consumer gates its id parser on them — and none of the
three may import another, so the vocabulary travels as strings. Nothing checked
that the three spellings agree, and the failure mode is the quiet one this whole
pull request is about: a kind a rule publishes and the host does not map is
dropped by declared design, so the feature stops working for that kind with no
error at the producer, the host, or the consumer. The set is now derived from the
dump vocabulary — which the Go kind-contract test already holds against the
producer's constants — and all three spellings are required to match it.
Mutation-verified by renaming one kind in the rule API alone.

**The serve envelope carried a package constant while the dump carried the
session's own claim.** A session holding artifacts published a dump declaring
`artifactNodes` beside an envelope that never did, and the envelope's own comment
claimed it "mirrors what every dump this server publishes declares for itself".
It matters on exactly one frame — an `unchanged` response carries no dump, so the
envelope is the only place a client could learn what the server proves — and no
current client reads it, which is why this is a consistency defect rather than a
live one. The shared list still answers before a session exists; once one does,
its own capabilities answer.

### Round 10, clean

A complete fresh round over the 78-file diff at this head found nothing that
survived verification. Re-read and traced: `ApplyArtifacts` (input not mutated,
sort stable and total over addresses, an address colliding with a declaration
loses, the edge dedup seeded from every edge already built); both error surfaces
(`publishArtifacts` and `resolveCapabilityPlugins` return an empty answer for
every ordinary state and swallow nothing a user needs to see elsewhere); every
declared constant has a reader on both viewers; the schema version has exactly
one definition on each side; the dump and serve capability claims agree; and
every new case carries the repository's three-part doc comment.

### Verification at this head

`@ttsc/test-graph` **77/77**. Go: `cmd/ttscgraph`, `internal/graph`, and
`cmd/graphdump` pass, as does the evidence suite. The four CI gate suites pass
locally (37 assertions). No prettier drift over the changed `.ts`/`.md`/`.mdx`
files; all changed `.go` files round-trip through `gofmt-2spaces.sh` unchanged.

Rounds 1–10 found **17 defects**, every one fixed and mutation-verified where a
mutation could exist. Five of them would have broken the product outright for the
projects this feature is for: a resolution that could never succeed, a resident
session that never asked, a verb that always answered empty, a shard projection
that rejected every artifact, and a client guard that refused the transaction.

Issue #1259, opened 2026-08-19
fix(ttsc): state and use the rule that lets a plugin-untouched file narrow its inputs


samchon:
> **Correction (2026-08-19).** The first version framed this as "the host stamps completeness for untouched files". That is only reachable where the host owns the envelope. `dependenciesComplete` arrives from the producer and the host passes it through without synthesizing it (`envelopeSideChannels`, `packages/ttsc/src/compiler/internal/transformProjectInMemory.ts`), so for an executable sidecar the declaration is the sidecar's. What ttsc owes every producer first is the *rule*: whether a file with no plugin edit has a type-independent output at all. The issue is rewritten around that.

## Problem

A module that no plugin transformed still validates the whole reference closure on every delivery, and in a typia project the great majority of modules are exactly that. A bundler build spends most of its cache-validation budget proving that files which cannot have changed did not change.

Measured on a 105-file project (60 application modules, 30 chained type declarations, 20 `typia.createValidate` calls; typia 14.0.0, ttsc 0.28.0):

| envelope fact | value |
| --- | ---: |
| files in the transform envelope | 105 |
| files containing a `typia.*` call | 20 |
| files with a plugin-reported dependency list | 20 |
| `dependenciesComplete` entries | 0 |
| reference closure per file (min / median / max) | 108 / 297 / 346 |

So 85 of 105 modules receive no plugin contribution and each still derives and validates a ~300-path input set per delivery. The build cost is `modules × closure`, and the closure grows with the program, so it is quadratic in project size with a small constant.

## What is actually missing

The protocol has the narrowing contract, and a producer can use it. What it does not have is a stated rule for the case that matters most: **is the source-to-source output of a file that no plugin edited a function of that file's own text and the compiler options alone?**

Measured on the fixture, it is. The output is a faithful re-print:

```ts
// src/mod-25.ts, source
import type { Model25 } from "~/types/type-25";
import { value26 } from "./mod-26";
export const value25 = (input: Model25): string =>
  `${input.id}:${input.index}:${input.label}`;

// envelope output for the same key
import type { Model25 } from "~/types/type-25";
import { value26 } from "./mod-26";
export const value25 = (input: Model25): string => `${input.id}:${input.index}:${input.label}`;
```

[...]

Issue #1265, opened 2026-08-19
handoff: the transform-cache campaign behind #970 (context, measurements, traps, open follow-ups)


samchon:
Everything a fresh agent needs to continue the transform-cache work without any local notes. #970 is closed and its fix is merged; four follow-ups are open. This issue is the durable record of what was measured, how, what it means, and where the traps are.

## 1. What was wrong, and what landed

`@ttsc/unplugin`'s transform cache **never hit** in any project that resolves a dependency through a `.d.ts` file, which is every real application. Every delivered module recompiled the whole project, in `vite build` and `vite` serve alike.

`driver.SupersedingModuleCandidates` enumerates superseding resolution candidates speculatively, so the compiler never reads them and `attachInputProof` records no proof; the consumer nevertheless demanded `graph.inputProofs.size === graph.members.size` before calling a generation stable. `projectSnapshotComplete` was therefore permanently false, which closes the build-scoped shortcut, the narrow persistent path, and complete-snapshot validation at once. Each refusal evicts the generation and the next module recompiles into another unprovable one.

Fixed in #1247 (`3363f74ad`), closing #1245 and #1246:

- graph members are classified **realized** vs **speculative**; a member reported only under `graph.candidates` is validated against the state recorded when the envelope was produced, never against a compiler proof it cannot have;
- generation stability and complete-snapshot validation compare the envelope's **declared inputs**, not every file under the project root (a project root is a working directory: generated types, logs and coverage are written there mid-compile);
- a walk failure is scoped to its key, and an unattributable one still taints everything;
- the transform hands the adapter the identity and existence it already memoized per watch input;
- a dev server with `server.watch: null` receives no watch registration;
- an unreusable generation now reports itself once on stderr with the evidence it lacked.

## 2. Version archaeology (measured, one fixture, one machine)

`vitest --run`, plugin binary warmed with `ttsc prepare` first so the Go build is excluded.

| ttsc / @ttsc/unplugin | typia | Duration |
| --- | --- | ---: |
| 0.26.2 | 13.1.1 | 14.38 s |
| 0.26.2 | 13.2.0 | 95.06 s |
| 0.27.0 | 13.1.1 | 167.61 s |
| 0.27.0 | 13.2.0 | 481.88 s |
| 0.28.0 (pre-fix master) | 14.0.0 | 609.96 s |
| after #1247 | 14.0.0 | **53.96 s** |

[...]

Issue #1263, opened 2026-08-19
feat(ttsc): let a linked plugin declare its dependencies and their completeness


samchon:
## Problem

The dependency-completeness contract is reachable only by an executable sidecar. A linked plugin has no way to report per-file `dependencies` or to declare `dependenciesComplete`, so every consumer of a linked plugin falls back to the host's conservative bound: the reference closure of the transformed file plus every global-scope declaration, re-validated once per delivered module.

That bound is right for a type-driven generator. It is pure waste for the first-party utility plugins, whose transforms are syntactic:

| plugin | what its output actually depends on |
| --- | --- |
| `@ttsc/banner` | the file's own text and its `banner.config.*` |
| `@ttsc/paths` | the file's own text and the tsconfig `paths` mapping |
| `@ttsc/strip` | the file's own text and its `strip.config.*` |

Each of those config inputs is already declared through `PluginContext.ReportHostInput`, and the config chain is universal to every file anyway. So for these plugins the honest declaration is "`dependencies[F]` is empty and complete", and the closure they force consumers to validate cannot influence their output at all.

`@ttsc/lint` is not affected: it registers at `stage: "check"` and produces diagnostics rather than transformed output, so the transform envelope's dependency contract does not apply to it.

## Evidence

The linked-plugin surface in `packages/ttsc/driver/plugins.go` is:

```go
type SourcePreamblePlugin interface { SourcePreamble(PluginContext) (string, error) }
type ProgramPlugin        interface { ApplyProgram(*Program, PluginContext) error }
type EmitTransformPlugin  interface { EmitTransform(PluginContext) (PluginTransform, error) }
```

plus `PluginContext.ReportHostInput`, `ReportHostInputHash`, and `ReportHostInputRealpath`. Nothing reports per-file dependencies, and nothing declares completeness. A grep for `dependenciesComplete` across `packages/ttsc/**/*.go` returns nothing: the field exists only in the JavaScript host's envelope decoding, where it arrives from a sidecar's stdout.

Meanwhile the protocol page documents the contract as if any producer could use it, and the utility plugins ship as linked packages by design (`packages/{banner,paths,strip}` link into the generic native host).

Measured scale of what the fallback costs, on a 105-file project: the reference closure has a median of 297 paths per file, against ~10 declarations a type-driven plugin actually consults and zero for a syntactic one.

## Approach

Give the linked-plugin surface the same two declarations a sidecar can make, then use them in the first-party utility plugins.

Shape to settle in the design:

[...]

Issue #1261, opened 2026-08-19
fix(unplugin): stop re-probing a missing resolution candidate on every delivery


samchon:
> **Correction (2026-08-19).** This issue was written before #1259, #1260 and #1263 landed, and it assumed the residual was the per-delivery re-proof of a file's whole input closure. It is not: measured after #1266, content re-reads are already zero and the entire remaining term is the probe of *missing resolution candidates*. The approach is re-scoped accordingly, from a general proven-input memo with its own epoch and probe files to reusing the notification channel the adapter already trusts for universal host inputs. The measurement is below; the original text is in the issue's history.

## Problem

In a long-lived watching dev server, every cache-hit delivery re-probes each missing resolution candidate reachable from the delivered file, one failed `lstat` per candidate per delivery. The work is `modules x candidates` per build, and no producer declaration can remove it, because candidates stay host-owned by contract: a plugin cannot vouch for a compiler-resolution change that happens without any plugin input changing.

## Evidence

`experimental/unplugin-perf`, whose scenarios D and E are exactly this case (a graph-bearing envelope with no build boundary). Windows, 50 modules, one whole-project compile, per delivery of a cache hit:

| scenario | shape | reads/file | stats/file | ms per delivery |
| --- | --- | ---: | ---: | ---: |
| D | partitioned externals, 50 externals, 100 unrelated directories | 0.0 | 1.0 | 15.7 |
| E | shared closure, 50 externals, 50 globals, 50 missing candidates | 0.0 | 25.5 | 26.2 |

`25.5` is the harness's own `(modules + 1) / 2`: the candidates reachable from the delivered file in a chain graph. The mechanism is in `packages/unplugin/src/core/transform.ts`:

- `inputMetadataEvidence` throws on a path that does not exist, so the signature shortcut that stands in for every other input's content comparison never applies to a candidate;
- `matchesProvenInput` therefore falls through to `matchesRecordedInput`, which probes the filesystem again;
- and the `MISSING_INPUT_STATE` branch deliberately refuses to record a signature, because an unreadable *existing* file reports the same state while its bytes can still change.

So the probe repeats for every delivery that reaches the candidate, forever, and it is the only per-delivery filesystem work left in a watching session.

## Why the original approach is over-scoped

[...]

PR #1266, opened 2026-08-19
fix(cache): narrow per-delivery validation for host-owned envelopes and watcherless sessions

samchon:
Claims the complete follow-up cycle of the transform-cache campaign recorded in #1265. Opened empty, before any code exists, so no overlapping work starts on the same surface.

## Scope

| issue | what it removes |
| --- | --- |
| #1259 | the reference-closure validation for a file no plugin edited, plus the missing *rule* that lets any producer declare such a file at all |
| #1260 | the whole per-delivery derivation for a one-shot session (a dev server with `server.watch: null`, which is what `vitest --run` configures) |
| #1263 | the linked-plugin surface's inability to report per-file dependencies or declare their completeness |

#1261 stays open by its own terms: it is an investigation that requires measuring what remains **after** #1259 and #1260 land, and it proposes a change to the freshness contract rather than a fix under it. samchon/typia#2357 and samchon/nestia#1598 are external producers.

## Verification

Pending. The pull request owns the complete accepted cycle; every issue it resolves will be closed from the commit that earns it.


samchon (commented):
**Individual Self-Review — `8b72a1686` "fix(unplugin): give a watcherless dev server the build-scoped cache"** (#1260)

What landed: `buildStart` now reads `server.watch` beside `config.command`, so a serve session with no watcher takes the build-scoped lifecycle; three scenarios (watcherless positive, watching negative twin, repeated-module boundary) and the `setup/unplugin` page.

Findings and their adjudication, all applied in `fca33e201`:

1. **Accepted, and the most consequential one.** `test_vite_serve_validates_first_use_after_startup` asserted the opposite verdict for the same configuration: a real dev server with `server.watch: null`, an entry corrupted after startup, and a required rejection on the first request of a never-served module. The review's mechanism check was right — under the new branch that request is settled from the startup generation, so the case fails.

   The scenario's subject, per its own doc comment, is that Vite's single startup `buildStart` is not a build boundary, which is a statement about a *watching* dev server; it used `watch: null` because a live chokidar watcher holds the runner process open (the trap recorded in #1265). That invariant now lives with the watching configuration, pinned by `test_vite_serve_with_a_watcher_keeps_persistent_validation`, and the real-server case was rewritten to assert what a watcherless session owes its consumer: one coherent generation, first use included. This is a deliberate contract change, made by #1260 and recorded here rather than absorbed silently.

2. **Accepted.** The review's framing correction stands: `watch: null` removes the change-*notification* channel, not the demand channel — a host can invalidate and re-request through Vite's public API, and a module the session already delivered still revalidates. The comment in `core/index.ts` and both doc surfaces now say that instead of "nothing can re-request a module", and state the real argument: per-delivery validation in a session that cannot observe an edit buys incoherence (two compilations of one program across one session's modules), not freshness.

3. **Accepted.** `packages/unplugin/README.md` carries the same paragraph as the website page and was not amended with it. Both now agree.

4. **Accepted (minor).** The duplicated hook-invocation preamble is folded into `loadViteAdapterPlugin` / `invokeVitePluginHook` in `internal/adapter-vite-serve.ts`, used by both harnesses.

5. **Accepted in substance.** "No scenario drives a real Vite server for the new lifecycle" — the rewritten `test_vite_serve_without_a_watcher_serves_the_startup_generation` is that scenario, and it passes against a real server locally (23 s).

6. **Deferred, not dismissed.** Formatting differences the review flagged are real; the cycle runs `pnpm format` once over the integrated branch, as the campaign requires, so they are resolved there rather than per commit.

Cleared without change: non-Vite bundlers and `watch: undefined` / `watch: false` (Vite's own predicate is `!== null`, so the adapter cannot disagree with it), `configResolved` ordering, plugin instances reused across a serve and a later build, and per-environment `buildStart` in Vite 6/7.

samchon (commented):
**Individual Self-Review — the dependency-completeness mechanism** (`8b72a1686..a9b739d91`, spanning two commits because a broad `git add` swept the in-progress half into `fca33e201`). Resolves #1259 and #1263. Adjudicated below; the fixes are in `57d3065c0`, and `120851c74` had already closed two of the gaps.

The central question — can the declaration serve stale *output*? — was traced end to end and answered no: `sourcePreambleFS` prepends a pure text prefix before parsing, `api-transform` answers `file.Text()`, `RunTransform` prints through a printer constructed with no `EmitContext` and no resolver, `emitTransforms()` is reachable only from the emit lane, and neither `@ttsc/banner` nor `@ttsc/strip` touches the Checker while both report their config loader's whole module set as host inputs.

**1. Accepted, and the most valuable finding: the narrowing narrows diagnostics too.** The envelope carries whole-program diagnostics, and a consumer that no longer registers or validates `reach(edges, F) ∪ globals` also stops re-running the compile that produces them. A type error introduced in a file no bundler ever delivers — a type-only import above all — now appears at the next compile run for another reason rather than at the edit. The two halves cannot be separated: proving an output unchanged without reading its closure is exactly what hides the closure's diagnostics. #1259 accepted the watch-registration half explicitly ("a type edit no longer registers against modules it cannot affect"); what was missing was saying so. `protocol.mdx` now has *What the narrowing also narrows*, and the adapter page and README carry the short form, pointing at the compiler's own check for projects that want the error at the edit.

**2. Accepted.** `ReportFileDependency` dropped an unusable dependency silently while the same plugin's completeness claim for that file still landed — a complete list missing a member, manufactured by the host rather than by the declaring plugin. An unusable dependency now withdraws that plugin's claim for that file only. New in-package case.

**3. Accepted.** `transformContributors` — the half that reads the process-wide registry, and the one whose failure direction is unsafe — had no coverage, because the aggregation tests deliberately bypass the registry. New `packages/ttsc/test/driver/transform_dependencies_exclude_an_emit_only_plugin_test.go` drives a real `LoadProgram` with a declaring program plugin beside an emit-only one.

**4. Accepted as a documentation fix.** `utility.RunServe` answers transformed text with no envelope side channel at all, so it declares nothing and always did; the protocol page said "both host-owned lanes" while `cmd/utility-host` has three. It now names the two that produce an envelope and says what `serve` does.

**5. Accepted.** The byte-identity claim on `TransformDependencies` was true for a plugin that declares nothing and false for the zero-contributor lane, which lists every file. The doc distinguishes them.

**6. Accepted.** `filepath.Rel` reconciles separators and the project prefix but keeps the target's case, so a plugin reporting a re-cased spelling has its declaration dropped on a case-insensitive filesystem. Conservative, but silent; `transformKey`'s doc now says to report the program's own spelling.

**7. Partially accepted.** The five in-package tests exercise unexported functions, which is the documented reason for the `input_observation_test.go` precedent; the new registry-dependent case went to `packages/ttsc/test/driver/` where the harness lives.

**8. Accepted.** The strip scenario asserted only that the sibling was absent, so a watch list that collapsed to nothing would have passed. It now makes the same two positive assertions as the banner twin.

**Noted for a separate issue, not this cycle:** the banner and strip config walks probe candidate paths without reporting the misses (`ReportHostInputHash(path, nil)`), and strip's loader reads a `package.json` unreported. Pre-existing, and unchanged by the narrowing — those paths were never in the dropped bound either — but the new "the text comes from `banner.config.*` alone" comments lean on that reporting being complete.

samchon (commented):
**Overall Self-Review — complete cycle diff** (`origin/master...HEAD`), three rounds.

Surface: 34 files, the whole base-to-head diff, re-read from scratch in each round. Issues resolved: #1259, #1260, #1263.

**Round 1** (over `a9b739d91`) found three coverage gaps and two documentation defects, all fixed in `120851c74`:

- `emitDecoratorMetadata` is the exception #1259 named, and the claim that it cannot reach a transform envelope was argued structurally and never checked. A decorated file with an imported parameter type now proves it.
- A bundler alias moves the host's cwd into the system temp directory, so every envelope section is keyed absolutely instead of project-relatively. Nothing pinned that a declaration still joins back to the file it names, and the failure mode is silent: the narrowing simply stops.
- The linked-completeness harness asserted the transform changed the entry's text, which is incidental, and did not share the plugin build cache its sibling fixtures use.
- The `buildStart` comment still carried the wording the #1260 review replaced, because that patch reached the two documentation pages and was lost for the source file. The review that reported it fixed was ahead of the tree.
- `ITtscCompilerTransformation.dependenciesComplete` described the field as if only a sidecar could produce it.

**Round 2** (over `57d3065c0`) found one: the prose this cycle added to the website used em-dashes, which the documentation skill's Voice section forbids. The surrounding pages predate that rule and keep theirs; only the new sentences were rewritten, so no unrelated line reflows. Fixed in `1559690bb`.

Rejected in round 2 after reproducing the mechanism: keeping watch registration wide while narrowing validation, to preserve type-error feedback. It buys nothing, because a re-delivery whose narrow validation passes serves the cached output without recompiling, so no new diagnostic appears either way. That is why the trade is documented rather than split.

**Round 3** (over `1559690bb`) produced nothing. Re-inspected: the aggregation's self-edge, duplicate-key and zero-contributor behaviour; the absolute-key fallback for files outside the host cwd (monorepo sources, `node_modules` sources, the alias-overlay lane); Windows separator and case rules; concurrency and nil-safety of the declaration ledger; JSON determinism and envelope size for a large program; the retained candidate walk (the residual #1261 describes); sidecar isolation, which an existing case already pins by deep-equalling a sidecar's own list; package READMEs and the two website pages; and the Go and TypeScript test conventions.

**Local validation**

| lane | result |
| --- | --- |
| `go build ./...`, `go test ./driver/ -run TestTransformDependencies`, `go test ./test/driver/ -run TestTransformDependenciesExcludeAnEmitOnlyPlugin` | pass |
| `pnpm test:typecheck` | pass |
| `@ttsc/test-ttsc`, the three completeness cases | pass |
| `@ttsc/test-unplugin`, full suite | pass (136 cases) |

One anomaly, recorded rather than explained away: an earlier full local run failed `test_webpack_watch_rebuilds_through_a_type_only_edge` with "multiple transform native backends cannot share one source-to-source pass". The case passes alone, passes with the whole `--include=webpack` group, passes in the full re-run, and passes on CI's bundler lane. Its error requires two executable transform entries to resolve to different binaries, which the fixture's two entries cannot do by construction: they share one source directory and the binary key hashes that directory rather than the entry config. Unreproduced, and nothing in this diff touches that path.

Issue #1272, opened 2026-08-19
A serial delivery pays the 10 ms Windows mutation-settle barrier once per module


samchon:
## Problem

In persistent-validation mode every delivery awaits `settleProjectMutationEvents`, and on Windows that barrier is a fixed 10 ms timer per tracker. Concurrent sibling deliveries share one promise; **serial** deliveries cannot, so a host that requests modules one after another pays the full barrier once per module.

That is a floor no cache hit can go below. A thousand-module serial build spends ten seconds waiting for watchers that had nothing to report.

## Evidence

`packages/unplugin/src/core/transform.ts`:

```ts
async function settleProjectMutationEvents(
  cached: TtscCachedProjectTransform,
): Promise<void> {
  const trackers = [cached.projectMutationTracker, cached.hostInputMutationTracker] /* ... */;
  await Promise.all(
    trackers.map(async (tracker) => {
      tracker.settle ??= new Promise<void>((resolve) => {
        const settled = () => {
          tracker.settle = undefined;
          resolve();
        };
        if (process.platform === "win32") setTimeout(settled, 10);
        else setImmediate(settled);
      });
      await tracker.settle;
    }),
  );
}
```

`tracker.settle` is cleared the moment it resolves, so the memo only covers awaiters that arrived while one barrier was in flight. A serial caller always arrives after, and mints a new 10 ms timer.

Measured with this repository's own `experimental/unplugin-perf` on Windows, after #1266, over 50 modules delivered serially in serve mode:

| scenario | reads/file | stats/file | ms per delivery |
| --- | ---: | ---: | ---: |
| D (partitioned externals, 100 unrelated directories) | 0.0 | 1.0 | 15.7 |
| E (shared closure, 50 globals, 50 missing candidates) | 0.0 | 25.5 | 26.2 |

Scenario D performs one stat and reads nothing, and still costs 15.7 ms. The barrier accounts for at least 10 ms of it by construction. Attributing the rest, and confirming the barrier's share directly rather than by subtraction, is the first task of this issue.

POSIX pays `setImmediate` instead, which is a macrotask turn rather than a timer, so the exposure is Windows-specific in magnitude but the shape is the same.

## Why the barrier exists

[...]

Issue #1271, opened 2026-08-19
A config file appearing at a higher-priority discovery position never invalidates a cached generation


samchon:
## Problem

`@ttsc/banner` and `@ttsc/strip` discover their config by walking up from the tsconfig directory and stat-ing a fixed list of candidate names per directory. Only the file that *matched* is reported as a host input; every path that was probed and found missing is not reported at all.

A cached generation therefore cannot notice a config file appearing at a **higher-priority discovery position** than the one it resolved. The plugin's next cold run resolves a different config — or fails outright — while every persistent consumer keeps serving output built from the old one.

## Evidence

`packages/banner/driver/banner.go::findBannerConfigFile`:

```go
dir := tsconfigBaseDir(cwd, tsconfigPath)
for {
  matches := make([]string, 0, 1)
  for _, name := range []string{
    "banner.config.json", "banner.config.js", "banner.config.cjs", "banner.config.mjs",
    "banner.config.ts", "banner.config.cts", "banner.config.mts",
  } {
    candidate := filepath.Join(dir, name)
    if stat, err := os.Stat(candidate); err == nil && !stat.IsDir() {
      matches = append(matches, candidate)
    }
  }
  ...
```

`resolveBannerTextWithReporters` then reports only `loaded.inputs` — the resolved config plus every module its loader pulled in (`reportBannerConfigInputs`). Nothing reports the misses. `packages/strip/driver/config.go` has the same shape for `strip.config.*`.

The host-input protocol already has the vocabulary for this: `ctx.ReportHostInputHash(path, nil)` records a probed-and-missing candidate, and `@ttsc/unplugin` treats a missing recorded input as a watch input whose creation invalidates the generation (`MISSING_INPUT_STATE` in `packages/unplugin/src/core/transform.ts`). The compiler's own resolution does exactly this for superseding module candidates (`driver.SupersedingModuleCandidates`, `graph.candidates`); plugin config discovery does not.

Two ways to reach it:

| starting state | change | today |
| --- | --- | --- |
| `banner.config.json` at the project root | add `banner.config.ts` beside it | a cold run errors with "multiple banner config files found"; a warm generation keeps serving |
| config resolved in a parent directory (monorepo root) | add one in the package's own directory | a cold run uses the nearer file; a warm generation keeps the ancestor's text |

[...]

Issue #1260, opened 2026-08-19
fix(unplugin): a watcherless dev server should get the build-scoped cache


samchon:
## Problem

A Vite dev server configured without a watcher gets the persistent-validation cache, so every module delivery re-proves the generation from the filesystem. That configuration is not a dev server in the usual sense: it is a one-shot session, and `vitest --run` is exactly that. Vitest sets it explicitly:

```js
// vitest/dist/chunks/cli-api.*.js
if (!options.watch) viteConfig.server.watch = null;
```

`@ttsc/unplugin` decides the cache lifecycle from `config.command` alone:

```ts
buildStart() {
  if (viteCommand === "serve") {
    resetTtscTransformCache(transformCache);   // persistent validation, every delivery
  } else {
    beginTtscTransformBuild(transformCache);   // one build, first delivery is O(1)
  }
}
```

So the workload behind samchon/ttsc#970 — a `vitest --run` suite — takes the persistent path, and each delivered module pays a full derived-input validation (measured at roughly 100 ms per module on a 60-module fixture after #1247, dominated by ~230 metadata syscalls).

## Why the build-scoped path is the right one there

`beginTtscTransformBuild` exists for a host with a real build boundary: it clears the previous generation, then a module's first delivery in that build is settled by comparing the supplied source with the generation snapshot in constant time. It rests on the assumption that the project does not change under the build while the build runs, which is the same assumption `vite build` already makes.

A server with `server.watch: null` has told Vite it will not observe file changes at all. It cannot invalidate, HMR, or re-request anything on an edit, so it is no less immutable than a production build — it is *more* so, because a build at least ends and reruns. Treating it as persistent buys nothing: there is no edit it could ever notice.

The same premise already governs the neighbouring decision in the adapter after #1247, where a watcherless server receives no watch registrations, for the same reason.

## Expected effect

The per-delivery cost for a one-shot suite drops from "derive and prove this file's inputs" to "hash the supplied source", leaving the single project compile as the dominant term. On the fixture used for #1245 that removes roughly `modules × 100 ms`, and the term grows with both module count and closure size, so a large suite gains proportionally more.

## Approach

Read the resolved config's `server.watch` alongside `command` and select the lifecycle from both: a serve session with no watcher takes the build-scoped path, a watching dev server keeps persistent validation because its one `buildStart` spans later edits.

The two properties are already read in the same hook after #1247, so this is a lifecycle decision, not new state.

## Acceptance and verification

[...]

PR #1273, opened 2026-08-19
fix(cache): remove the residual per-delivery probe and settle cost from a watching session

samchon:
Claims the remaining repository-side issues of the transform-cache campaign (#1265). Opened empty, before any code exists.

## Scope

| issue | what it removes |
| --- | --- |
| #1271 | a config appearing at a higher-priority discovery position never invalidating a cached generation, because `@ttsc/banner` and `@ttsc/strip` report the config they found and not the candidates they probed |
| #1261 | the per-delivery probe of every missing resolution candidate reachable from the delivered file, re-scoped from this issue's original "prove every input once per epoch" after measurement showed content re-reads are already zero |
| #1272 | the fixed 10 ms Windows mutation-settle barrier a serial delivery pays once per module |

## The measurement that set this scope

Taken with `experimental/unplugin-perf` after #1266, Windows, 50 modules delivered serially in serve mode:

| scenario | reads/file | stats/file | ms per delivery |
| --- | ---: | ---: | ---: |
| D partitioned externals | 0.0 | 1.0 | 15.7 |
| E shared closure, 50 globals, 50 missing candidates | 0.0 | 25.5 | 26.2 |

`25.5` is `(modules + 1) / 2`, the reachable missing candidates, one failed `lstat` each. `15.7` with one stat and no read is the barrier floor. Those two terms are what is left, and they are what this cycle removes; the full evidence is in #1261 and #1272.

## Verification

Pending. The pull request owns the complete accepted cycle; every issue it resolves will be closed from the commit that earns it.


samchon (commented):
**Individual Self-Review — `5db53402b`** "fix(plugins): report the config candidates a discovery walk rejected" (#1271). Findings adjudicated; fixes in `e1fda8ae6`.

1. **Accepted, and the most consequential.** `DiscoverConfigFile` put a *directory* carrying a config file's name in the same bucket as an ENOENT candidate, and the reporter then sent the paired nils for it. The host-input contract records an existing directory by its directory-kind digest and its physical path, so the envelope would have disagreed with every consumer's own filesystem permanently: `captureUniversalHostInputValidation` returns undefined, the generation is never reusable, and the whole project recompiles per delivered module. Conservative in output but the same permanently-unreusable shape #1245 was filed for, and exactly the boundary #1271 listed. The walk now classifies the candidate and `ReportRejectedConfigCandidates` records each in its own state.

2. **Accepted.** Neither e2e scenario could distinguish "invalidated correctly" from "the cache was disabled", which is precisely what finding 1 produces. The supersession case now asserts the superseding candidate reached `result.hostInputs`, and a new twin pins that an unrelated file beside a probed candidate leaves the generation object identical.

3. **Accepted in part.** The reported set is genuinely O(names x depth), and the strip-defaults case reaches the filesystem root. The derivation is right — every one of those paths can supersede — and the cost the review identified in `matchesUniversalHostInputProbes` is the same per-delivery term #1261 addresses in the same pull request; the absent-candidate registration also declines past 512 distinct directories rather than risking a failed tracker. The `allEvents` null-filename note is real and pre-existing; recorded rather than changed here.

4. **Accepted.** Go coverage now includes the directory-shaped candidate and the reporter's two halves. The root-reaching walk and the ambiguous-directory probed set remain covered only through the plugins' own suites; noted rather than expanded.

5. **Accepted.** The unused `outer` return is gone.

**Cleared as clean by the review and re-checked here:** semantic equivalence of the refactor (name order, the `os.Stat` + `!IsDir()` test, stop-at-first-answering-directory, both ambiguity messages and their directory, the root result, and `configFile` probing nothing), the candidate-set derivation, the nil/nil pairing against the host-input contract, the `//go:linkname` plumbing, and the Forbidden section.

samchon (commented):
**Individual Self-Review — `39f6268fa`** "fix(unplugin): prove an absent candidate by notification and settle by acknowledgement" (#1261, #1272). Findings adjudicated; fixes in `e1fda8ae6`.

1. **Accepted — the proof was unsound under a link, and this was the finding of the cycle.** A watcher opened on a spelling that traverses a symlink or junction follows it to a physical directory, so a pnpm reinstall retargeting `node_modules/<package>` makes a candidate appear behind a watch still looking at the old store directory. Nothing fires, `notifiesAbsence` keeps answering true, and the delivery serves a generation the new file superseded. The registration now covers every directory component of the spelling by the name it carries in its own parent, so the retarget is reported where it happens. `test_transformttsc_retargeted_candidate_link_invalidates_generation` pins it and discriminates: removing the walk turns 2 compiles into 1.

2. **Accepted as stated, with the mechanism improved rather than reverted.** The round-trip proves the child-to-parent leg, not the kernel-to-child one, so the barrier did shrink. Both the broker and the in-process path now answer after two turns instead of one, which gives the loop a poll phase for completions the kernel had already queued before the reply is sent. That is strictly more than the old fixed wait proved on the parent side — with `setTimeout(10)` the parent resumed in the timers phase, before the poll phase that would have read the child's message — and the fallback keeps the previous grace when the broker stops answering. `test_transformttsc_synchronous_membership_change_reaches_the_next_delivery` pins the property the barrier exists for.

3. **Accepted.** `release` guarded the unref with both counters while the registration path guarded it with one, and `ref`/`unref` is a flag rather than a counter, so a concurrent registration could drop the reference an in-flight acknowledgement was holding — the same mid-suite exit the commit had just fixed. Both sides now agree.

4. **Noted, not changed.** The deleted-and-recreated watched parent is platform-dependent and pre-existing; the chain registration from finding 1 now also covers the common case of that parent being replaced, since its name is watched in its own parent.

5. **Accepted.** The registration declines past 512 distinct directories. A host that exhausts its watch descriptors fails the tracker, and a failed tracker sends every delivery to complete-snapshot validation, which re-hashes the whole project — far more expensive than the probe this removes.

6. **Accepted.** The harness now gates the `lstat` term it started counting: scenario D at 8 per delivery, scenario F at 60, which is below the undeclared shape it is measured against. Scenario E stays deliberately ungated as the unbounded reference. The dead `statsBudget` override is gone.

7. **Accepted.** New: a candidate whose watch could not be opened is still probed (the bound "only a covered candidate may skip"), and the link-retarget case above. The appearing-candidate twin was re-checked rather than moved: with a fanout that already created the package directory, creating the candidate file inside `node_modules` is not a project-walk event, so it does exercise the host-input watch.

8. **Accepted.** The runner guard now writes its diagnostic with `fs.writeSync`, since an async write from an `exit` handler can be dropped before it flushes.

9. **Accepted.** One acknowledgement per settle now serves both trackers. And the framing is corrected in both documentation surfaces: this does make a notification the sole positive evidence for one input class, which is a new trust, and the pages now say so along with what it rests on.

samchon (commented):
**Individual Self-Review — `f8287e507`** "stop the candidate chain watch at the project's own boundary". Findings adjudicated; fixes in `a4e912801`.

1. **Accepted, and it would have shipped a regression into this adapter's primary host.** The chain registers *directory* names, on the tracker that listens for every event. An entry's attribute change reaches a watch on its parent, and a directory's attributes move whenever anything is written below it, so a Vite dependency optimizer writing into `node_modules` during a dev session would have marked the generation stale on every pass. The absent candidates now have their own tracker listening for renames alone — which is what both events that can change a candidate's answer are, its creation and a retarget of the link leading to it — with its own membership, settle and disposal handling.

2. **Accepted.** `enclosesProject` compared spellings: `path.relative` answers with an empty string when two spellings name one directory, including spellings differing only in case, and the guard read that as "not the root" and climbed one level past it, reinstating the watch the boundary exists to prevent. It also rejected a directory literally named `..x` through a prefix test. Both now follow the containment spelling the project walk already uses.

3. **Accepted as a comment correction.** The bound is one-sided by design: a candidate outside the project subtree still walks toward its own root, stopping at the first directory that encloses the project. The names registered there are stable (`node_modules`), the candidate's own probe already opens the same watch, and the rename-only listening from finding 1 removes the noise that made the unbounded walk fail. The comment now says what the guard promises.

4. **Accepted.** Both documentation surfaces still described the unbounded walk. This is the one input class whose sole positive evidence is a notification, so the contract text has to say where the registration stops and what it listens for.

5. **Accepted as a limitation, recorded rather than closed.** There is no cross-platform pin for the boundary: the failure it prevents is only observable where an entry's attribute change reaches its parent's watch, and the two cases that caught it are Linux-only in effect. `selectNotifiableAbsentInputs` is not exported, so a direct assertion on the watched set would need a seam this cycle does not add. The Linux lane is the pin.

6. **Accepted.** The resolved project root is hoisted out of the candidate loop.

**Cleared by the review and re-checked:** the retarget scenario still discriminates (its junction is below the project root); the workspace layout keeps its coverage; the config-candidate chain from #1271 is materially different, since it watches stable *file* names rather than volatile directory entries; the inotify claim in the commit message; loop termination and the directory-limit accounting; and the Forbidden section.

samchon (commented):
# Handover — where this pull request stands and how to resume

Self-contained: everything below is checkable from this repository and this pull request alone.

## 1. What this is

Cycle 2 of the transform-cache campaign in #1265. Cycle 1 was #1266 (merged as `a2a9b472f`, closing #1259, #1260, #1263). This one claims the campaign's remaining repository issues.

Head at handover: **`4b3958af3`**, branch `fix/transform-cache-residuals`, working tree clean and in sync with the remote. The three issues were closed as development-complete at the repository owner's instruction, ahead of the merge. **If this pull request is abandoned, reopen them** — their fixes live only on this branch.

## 2. Commits

| commit | what it does |
| --- | --- |
| `393e471b4` | empty claim |
| `5db53402b` | **#1271** `driver.DiscoverConfigFile` + `ReportRejectedConfigCandidates`; banner and strip report the config candidates their upward walk rejected |
| `39f6268fa` | **#1261 + #1272** absent candidates proven by notification instead of probing; the Windows settle barrier becomes the watcher's own acknowledgement; the perf harness starts counting `lstat` and gains scenario F |
| `e1fda8ae6` | review fixes: a directory wearing a config name classified as a directory; the candidate proof extended to the spelling; broker ref/unref; negative twins |
| `80b2353fd` | review fixes: the coverage claim required rather than derived; dedupe namespaces split |
| `a9c2e6ecc`, `eb19f60b3`, `953c4bc5a` | documentation and formatting from Overall Self-Review rounds 2 and 3 |
| `f8287e507` | CI fix: the chain watch bounded at the project root |
| `c32a9d478` | a design that was **withdrawn**, see section 4 |
| `4b3958af3` | restores the bounded chain, corrects the record, gives the candidate tracker its own retention |

## 3. What each fix is

**#1271.** Both utility plugins discovered their config by walking up and stopping at the first directory that answered, reporting only the file they found. Every candidate they probed and rejected can change that answer: one created nearer the entry wins the next search, one created beside the match makes that directory ambiguous. `DiscoverConfigFile` (`packages/ttsc/driver/config_discovery.go`) now owns the walk and returns those candidates, each carrying whether it was *absent* or a *directory wearing the name*, and `ReportRejectedConfigCandidates` records each in the state the host-input contract defines for it: paired nils for an absent path, the directory-kind digest plus its physical path for a directory. Reporting a directory as absent was a real defect caught in review — it makes the envelope disagree with the consumer's own filesystem permanently, so the generation is never reusable, which is the shape of #1245.

**#1272.** `settleProjectMutationEvents` slept a fixed 10 ms on Windows once per delivery, because serial deliveries cannot share the promise concurrent ones do. Each tracker now carries a `drain`: an in-process watcher answers after two turns of the loop its callbacks are queued on, and the Windows broker answers by an ordered IPC round-trip, also after two of its own turns so a completion the kernel had queued gets a poll phase first. A broker that stops answering falls back to the previous wait.

**#1261.** A missing resolution candidate is the one input a metadata signature cannot stand for, so every delivery re-probed it. The generation now registers its absent candidates, plus the directory components of the spelling leading to them bounded at the project root, with a dedicated tracker listening for **renames only**, and `matchesProvenInput` returns true for a recorded-missing input that tracker covers. Everything else is unchanged: any tracker failure withdraws the proof, and an unwatched candidate is probed exactly as before.

## 4. The design history of #1261 — read before touching `selectNotifiableAbsentInputs`

1. **Watch the candidate's own name.** Unsound: a watcher opened on a spelling that traverses a link follows it to a physical directory, so retargeting `node_modules/<package>` (a pnpm reinstall) makes the candidate appear behind a watch still pointed at the old store. Caught in review before it shipped.
2. **Also watch the directory components of the spelling.** Sound, but it puts volatile directory entries under watch. Unbounded it reached the system temp directory and the filesystem root, and CI failed two cases with 2 compiles where 1 is required. `f8287e507` bounded the walk at the project root.
3. **Claim only when no link leads to the candidate** (`c32a9d478`). Withdrawn by `4b3958af3`: the check was not a link test but a whole-path *canonicality* test, since `identities.resolve` canonicalizes the entire existing ancestry, so a project under a symlinked home, macOS temp, or a Windows short path would have had the optimization silently off; and it lost the parent-deleted-and-recreated case that only the chain covers.

**The withdrawal in `c32a9d478` rested on a wrong diagnosis.** It blamed a local failure in `test_transformttsc_persistent_validation_uses_per_file_inputs`, but that scenario passes no `graphCandidates`, so its envelope stamps none and no candidate tracker is created for it — the candidate work cannot touch it. The full suite has since passed twice over the restored design. If that scenario ever fails again it is *not* this work; start at its descriptor-selection and denied-directory sections.

## 5. How to reproduce the verification

Windows needs `TMP` and `TEMP` at the **long** path (the 8.3 form makes Vite fail to load fixture modules, #1265 section 7), and a stable `TTSC_CACHE_DIR` or every test process rebuilds the Go fixture plugins from scratch.

```bash
export TMP='C:\Users\<name>\AppData\Local\Temp'
export TEMP="$TMP"
export TTSC_CACHE_DIR='C:\Users\<name>\AppData\Local\Temp\ttsc-campaign-plugin-cache'
pnpm --filter @ttsc/unplugin build          # never while a suite runs
pnpm --filter @ttsc/test-unplugin start     # 146 cases
node scripts/test-go-utility-plugins.cjs    # banner, paths, strip
cd packages/ttsc && go test ./test/driver/
pnpm --dir experimental/unplugin-perf start # scenarios A-F with their budgets
pnpm test:typecheck
```

| lane | last result |
| --- | --- |
| `@ttsc/test-unplugin` full suite | 146 cases pass, **twice consecutively** over this design |
| utility-plugin Go tests | pass |
| `packages/ttsc/test/driver` | pass |
| perf harness | pass, gates included |
| typecheck, prettier, the Go formatter | clean |
| CI on `4b3958af3` | in flight at handover, 17 lanes green, none failed |

## 6. The measurement

`experimental/unplugin-perf`, Windows, 50 modules delivered serially in serve mode. The harness was fixed first: it counted `stat`, `readFile` and `realpath` but not `lstat`, which is the call every derived input's validation makes first, so the largest term was invisible.

| scenario | lstat/file | stat/file | ms/delivery |
| --- | ---: | ---: | ---: |
| D partitioned externals | 4.0 | 1.0 -> 0.0 | 16.2 -> 3.3 |
| E shared closure, producer declares nothing | 127.5 | 25.5 -> 0.0 | 31.2 -> 6.6 |
| F shared closure, producer declares completeness | 54.0 | 0.0 | 16.5 -> 3.5 |

Two terms with different owners: the **closure** (127.5 against 54.0) is removed by the declaration path — #1259, #1263, samchon/typia#2357 — never by a memo; the **candidate probe** and the **settle barrier** are what this pull request removes.

## 7. Open findings from the last Individual Self-Review of `4b3958af3`

The review confirmed the restored shape is sound for every appearance path on a filesystem whose watches fire, and that the revert is byte-exact against `a4e912801` apart from the four stated corrections. It left these, none merge-blocking in my judgement, all worth doing:

1. **The chain climbs past the project for a candidate outside the project subtree.** `enclosesProject` stops at the first *ancestor of the project root*, which for an out-of-subtree candidate (a global store, or a different drive on Windows where `path.relative` returns an absolute path) means the walk reaches the filesystem root. Cost is watch descriptors and an earlier trip of the 512-directory limit, never stale output. Fix: stop when the component is no longer a proper descendant of the project root, and stop outright when `path.relative` returns an absolute path. The comment and both doc surfaces currently claim the tighter bound.
2. **No test pins the parent-deleted-and-recreated case**, which is the case the restore was justified by. `assertAppearingCandidateInvalidatesGeneration` starts with `node_modules` absent, so it passes under shape 1 too and does not discriminate the chain; the retarget case covers only the link leg. Add: parent exists at capture, then remove and recreate it with the candidate inside.
3. **`retainTracker` still cross-gates the first two trackers** while the new comment claims per-tracker retention. Safe today only because both are created under the same condition; splitting it is two lines.
4. `enclosesProject`'s case-folding comment holds on `path.win32` and not on `path.posix`; on macOS a case-only difference climbs one level past the root (lost optimization).
5. Rename-only is sufficient for every appearance path checked (create, mkdir, rename-into-place, extraction, package-manager swap, hard link) on all three backends. The one change-only carrier is a Windows junction retargeted in place through `FSCTL_SET_REPARSE_POINT` without recreating the entry, which no mainstream tool does; a bind mount over a component produces no event at all. Both are stale output, both exotic.
6. Windows 8.3 short names in a chain component would not match the reported long name; they essentially only occur above the project root, which the chain does not reach.
7. `NOTIFIABLE_ABSENCE_DIRECTORY_LIMIT` counts lexical parents while the tracker collapses missing subtrees onto one watch, so the bound is sound but no longer tight and its docstring overstates.
8. Two comments justify the rename-only filter with an inference about Linux `IN_ATTRIB` that is stronger than the mechanism; the filter is still right for other reasons.
9. The documentation's "same assumption the membership proof rests on" is nearly exact: `node_modules` is in `isIgnoredProjectDirectory`, so an absent candidate under it was never covered by that proof and had a notification-independent probe until now. Worth half a sentence.
10. A sub-millisecond TOCTOU between the absence probe and watch registration, present since the first shape.

## 8. What is left

1. Watch CI on `4b3958af3` to completion. Two platform lanes failed on earlier heads for environment reasons unrelated to this diff and need a re-run: `platform (linux-arm64)` could not install `@rolldown/binding-linux-arm64-gnu` while running `npx vite build` in `experimental/test-unplugin` (the npm optional-dependency bug the error itself cites), and `platform (darwin-x64)` failed at "Ensure VS Code CLI (macOS)". If the rolldown one persists, that file's own precedent is to pin the offending package with a comment, as `@rspack/cli@2.0.0` already is.
2. Act on section 7, at least items 1 to 3.
3. Run one fresh **Overall Self-Review** round over `origin/master...HEAD`. The last complete rounds (1 to 5) were over `953c4bc5a`; five commits have landed since.
4. Merge with squash once the required checks are green on that exact head and the Overall round is clean. `delete_branch_on_merge` is on; delete the local branch and prune afterwards. The closing keywords are already carried by `5db53402b` (#1271) and `39f6268fa` (#1261, #1272), so the squash body closes all three even though they are closed manually now.

## 9. Campaign state

Nothing else in this repository is left from #1265. What remains is external: samchon/typia#2357 (declare `dependenciesComplete`, the 127.5 against 54.0 term) and samchon/nestia#1598 (stamp a reference graph at all, a staleness bug rather than a cost). The campaign knowledge base is `.wiki/transform-cache/00-campaign.md` in the checkout; it is gitignored and does not travel with the branch, and its "The absence proof went through three shapes" section is the long form of section 4 above.

samchon (commented):
# Overall Self-Review — the final round

Complete fresh rounds over the whole `origin/master...HEAD` diff, the first since rounds 1 to 5 over `953c4bc5a`; five commits had landed in between. Findings 1 to 8 below were applied in `b1999027b`, finding 9 in `cf53cf1ef`, and the closing round over `cf53cf1ef` produced nothing further. Whole surface, plus its consequence surface: both utility plugins' discovery walks and the single reporting channel they share, the adapter's narrow and complete validation paths, both watcher backends and the Windows broker's lifecycle, the perf harness and its gates, the fixtures, and both documentation surfaces.

## Findings applied

1. **The candidate set was derived where nothing could carry it.** `selectNotifiableAbsentInputs` ran for every generation, including a build-scoped adapter's, which opens no watcher at all — one existence probe per candidate producing an answer no code path can read. Derived only under `trackProjectMembership` now.

2. **The bound that decides which candidates are claimed had no test.** `d501dd63e` stopped claiming a candidate whose spelling leaves the project subtree; the failure it prevents is silent, and the existing negative twin turns watch registration off entirely, so it cannot tell the bound from a broken watcher. New `test_transformttsc_out_of_project_candidate_is_still_probed`, with the fixture gaining an absolute candidate spelling to stamp.

3. **A comment described a design two commits old.** `candidateMutationTracker`'s doc still said one tracker watched the universal inputs and the candidates together, which `f8287e507` ended when it gave the candidates their own rename-only tracker.

4. **The rename-only filter rested on a claim stronger than its mechanism** (open finding 8 of the last Individual review). The argument that holds on every backend is that a change event cannot make an absent path present, so the filter gives up nothing; the noise it removes is what a backend reports when a write below a directory surfaces as a change to that directory's own entry, which a probe on this platform confirms Windows does. The one appearance the filter itself cannot see — an in-place `FSCTL_SET_REPARSE_POINT` retarget — is now named where the filter is.

5. **The directory limit's docstring overstated it** (open finding 7). It counts lexical parents while the tracker collapses a missing subtree onto one watch, so it is an upper bound on the watches opened, not their number.

6. **Both documentation surfaces called the notification assumption pre-existing** (open finding 9). For a candidate under `node_modules` it is not: the project walk skips that directory, so such a candidate was never covered by the membership proof and kept a notification-independent probe until this cycle.

7. **The perf harness carried a dead knob and a stranded doc comment.** `statsBudget` became unused when the shared-closure scenario stopped overriding it, and its doc block had been left above `lstatBudget`, documenting the wrong field with text describing the per-delivery probe this pull request removes. The knob is gone and every serve scenario holds the membership budget itself, which is the invariant #1261 earns. Its README said "either" of three guarded invariants and never named the `lstat` gate that makes four.

8. **Two test-file slips.** A newline escape written as `String.fromCharCode(92) + "n"`, which wrote a literal backslash into the fixture, and an export list that broke its file's alphabetical order.

9. **The cycle's most consequential finding was pinned in one language only.** A directory wearing a config file's name is recorded by a fixed directory-kind digest and its physical path; the producer's constant is in Go and the consumer's is a TypeScript literal, and the Go test added with the fix asserts only that the digest is non-empty. Nothing tied the two together, and their disagreement is not an invalidation but a permanent refusal — the #1245 shape. New `test_transformttsc_directory_shaped_config_candidate_keeps_the_generation`, which asserts the path reached the envelope and then that the second delivery is served by the same generation object.


## Cleared by this round

The refactor's semantic equivalence for both plugins (name order, the `os.Stat` + `!IsDir()` test, stop-at-first-answering-directory, both ambiguity messages and their directory, the root result, and an explicit `configFile` probing nothing); `@ttsc/lint` walks the same shape but has no host-input reporting channel at all and `@ttsc/paths` has no config discovery, so the reported class is complete at two plugins; the paired-nil / directory-digest contract against the consumer's own `hostInputStateHash`, and the `filepath.Abs` realpath against `sameHostInputRealpath`, which compares by identity rather than by spelling and therefore holds under a symlinked ancestry; the candidate chain's ancestor invariant and its `directories` accounting; the coverage claim's soundness when a missing subtree collapses the watch onto a higher ancestor; that a partial watch failure fails the whole tracker; the broker's drain lifecycle against every exit path, including a drain in flight when the last tracker closes and a `send` on a closed channel; that `notifiesAbsence` is reachable only under a caller that has already cleared every tracker; and that the "must not invalidate" twin's walk stops inside its own private tmpdir rather than at a shared one.

## Recorded, not changed

The sub-millisecond window between an absence probe and its watch registration; Windows 8.3 short names in a chain component; a bind mount over a component, which no watch sees regardless of the filter; and the two costs a reported config candidate carries, both of which the fix's correctness requires and neither of which is a defect in it: one `readdir` per distinct directory of the walk per delivery, and one `fs.watchFile` poll per candidate under a Vite dev server, where a missing input is polled rather than registered because `vite:import-analysis` 500s on an added import that is absent by design. Both reach the filesystem root only for a `@ttsc/strip` project with no config anywhere, which is the one search that never stops early. Each of these is inherent to the shape rather than a defect in it.

## Verification

| lane | result |
| --- | --- |
| `packages/ttsc/test/driver` | pass |
| banner / paths / strip Go suites | pass |
| `experimental/unplugin-perf` | pass, every gate |
| typecheck (all `tests/*` projects) | pass |
| prettier, whole tree | clean |
| `@ttsc/test-unplugin`, full suite | pass, with the one exception below |

`experimental/unplugin-perf` on this machine, Windows, 50 modules delivered serially in serve mode:

| scenario | lstat/file | stat/file |
| --- | ---: | ---: |
| D partitioned externals | 4.0 | 0.0 |
| E shared closure, producer declares nothing | 128.5 | 0.0 |
| F shared closure, producer declares completeness | 54.0 | 0.0 |

`stats/file` is `0.0` in every serve scenario, which is the term #1261 removes, and no scenario needs a budget of its own for it any more.

### One local failure, which is not this cycle's

`test_transformttsc_persistent_utility_config_dependencies_invalidate_the_generation` fails on this Windows checkout: the banner TypeScript config loader cannot resolve the deliberate `.js` specifier to its `.tsx` file, so the first transform returns an exception. It reproduces identically on **`a2a9b472f`**, the merged base this branch was cut from, with the same message and the same stack. Nothing in this branch touches that loader — the only banner change is the discovery walk, and the failure is in loading a config the walk already found.

CI's `bundler defenses` lane runs the whole `@ttsc/test-unplugin` suite, that case included, and passes on this branch's head. That lane is `ubuntu-latest`; no CI lane runs this suite on Windows, which is why the platform dependence is invisible there. Recorded here rather than fixed, because it is neither this pull request's regression nor within its scope.

Issue #1264, opened 2026-08-19
A change to packages/evidence never runs its Go tests on Windows


samchon:
## Problem

A change confined to `packages/evidence/` never runs the evidence Go tests on Windows. The validation plan selects `["evidence", "go"]` for that path, and `go` is `ubuntu-latest`. The `windows-go` lane exists, runs `pnpm run test:go`, and is selected by several other paths, but not by this one.

`packages/evidence/native` is one of the most platform-dependent packages in the repository. Its behavior is decided by `filepath.IsAbs`, `filepath.Rel`, `filepath.ToSlash`, `os.Stat` versus `os.Lstat`, Windows drive prefixes, UNC shares, and junctions. None of that is proved by CI for a change that touches only this package.

## Evidence

`scripts/ci/validation-plan.cjs:464-466`:

```js
if (file.startsWith("packages/evidence/")) {
  add(["evidence", "go"], file);
  continue;
}
```

`go` is `ubuntu-latest` and `windows-go` is `windows-latest`; both run `pnpm run test:go`, which `scripts/test-go.cjs` routes through `test-go-evidence.cjs`. The `windows-go` lane's own comment says so:

> Same reason as the `go` lane: `pnpm run test:go` includes the evidence rule tests, and those stat `packages/evidence/lib`.

Observed on PR #1252, whose diff is 15 files under `packages/evidence/` plus one website page and one skill document. The lanes selected were `evidence`, `go`, `typecheck`, and the website lanes. `windows-go` was not among them, on any of its twenty-one pushes.

### What goes unproven

Behavior in this package that differs by platform and is decided in Go:

| Site | Differs how |
| --- | --- |
| `declaredRootIsAbsolute` (`base.go`) | `/srv/contracts` is relative on Windows and absolute on POSIX, so the root is joined to the project on one and not the other, and the diagnostic says so |
| `normalizeRootPath` (`base.go`) | refuses a drive-relative `C:docs`, restores the separator of a drive root `C:/`, and hand-cleans a UNC `//server/share` |
| `projectRelativeDisplay` (`base.go`) | falls back to the absolute path only when `filepath.Rel` errors, which happens only across Windows volumes |
| `populationBase.display` | composes a base whose display is a drive root, which no POSIX path produces |
| `resolveLinkedDirectory` (`loader.go`) | reads a Windows junction through `os.Readlink`, which `filepath.EvalSymlinks` does not resolve |
[...]

Issue #1269, opened 2026-08-19
A TypeScript claim rooted inside a linked directory selects nothing


samchon:
## Problem

A TypeScript claim whose `root` sits **inside** a linked directory selects nothing and deactivates in silence. The link is on an ancestor of the declared root rather than on the root itself, and nothing resolves it.

This is the residue of #1268, which resolved a link naming the root. It is the shape a package manager actually installs: `node_modules/@acme/contracts` is the link, and the root an author declares is a directory inside it.

### Measured

Against `3381bde3b`, with `mirror` a junction onto the workspace and the claim rooted at a directory inside it:

```
root=".../mirror/project" resolved=".../mirror/project" equal=true
MESSAGES(0):
```

`os.Lstat` of `.../mirror/project` reports a directory, because traversal through a link is transparent, so `resolveLinkedDirectory` returns the path untouched, the `entry.resolved == entry.base.Absolute` guard short-circuits, and the second spelling is never tried. The Program reports the other spelling and every source fails the comparison.

A Markdown reference over the same link, `root: "../mirror/docs"`, reports normally:

```
Missing acknowledgement for 'pricing.md#discounts' (Markdown H2 'Discounts' at ../mirror/docs/pricing.md:1)
```

So this is not a shared limitation. The walkers are transparent to a linked ancestor because they generate every path they compare from the base they were handed. The TypeScript comparison is not, because the Program hands it paths built by someone else.

### Root cause

`resolvedBaseDirectory` (`packages/evidence/native/base.go`) resolves the **last component** only:

```go
from := filepath.FromSlash(resolveLinkedDirectory(base.Absolute))
```

`resolveLinkedDirectory` (`packages/evidence/native/loader.go`) lstats the whole path and returns it unchanged the moment that path is a directory, which it is whenever the link is above it.

## Consequence surface

- **TypeScript only.** The two walkers generate their own paths from the base, so a linked ancestor never produces a mismatch for them. Confirmed by measurement above.
- **The claim side is the silent half**, as in #1258 and #1268: no diagnostic, a dropped obligation, a green build.
- **`filepath.EvalSymlinks` is not the answer.** It resolves every component but returns a Windows junction unchanged, which is the reason this package has its own resolver at all.
- **Cost.** Resolving each component means one `os.Lstat` per component per base rather than one per base. It runs once per base per pass, not per source, so the loop it sits in is the cheap one, but the count is no longer constant.
[...]

Issue #1268, opened 2026-08-19
A TypeScript claim rooted at a link selects nothing and deactivates in silence


samchon:
## Problem

A TypeScript claim whose `root` names a symbolic link or a Windows junction selects nothing and **deactivates in total silence**. The root is accepted, no source matches it, and no diagnostic is produced anywhere.

This is the same defect #1258 removed from the Markdown and Prisma walkers, in the one artifact kind that issue explicitly excluded, on the grounds that a TypeScript population walks nothing. It does not walk, but it does compare paths, and the comparison has the same asymmetry.

### Measured

Against `9ee765686`, with `mirror` a junction onto the project's own directory:

```go
messages := runRootedGraphIn(t, workspace, map[string]string{
  "project/docs/pricing.md": "## Discounts {#discounts}\n",
  "project/src/sale.ts":     "export interface ISale {}\n",
}, `{"claims":[{
  "type":"typescript",
  "root":"../mirror",
  "files":["src/**/*.ts"],
  "symbol":"type",
  "reference":{"type":"markdown","files":["docs/**/*.md"],"symbol":"h2"}
}]}`)
```

```
MESSAGES(0):
```

With the root naming the directory rather than the link, the same configuration reports `Missing acknowledgement for 'docs/pricing.md#discounts'`.

### Root cause

Two calls decide which sources belong to a declared TypeScript root, and both measure against the base as declared:

| Site | Call |
| --- | --- |
| `packages/evidence/native/typescript.go:41` | `relativeProjectPath(base.Absolute, file.FileName())` |
| `packages/evidence/native/typescript.go:1858` | the same, for the governance pass |

`baseDirectoryProblem` accepts the root because `os.Stat` follows a link and finds a directory. The Program reports whatever path its `tsconfig.json` resolved. When those two disagree about the link, `filepath.Rel` produces an ascending path, `relativeProjectPath` refuses it, every source is skipped, and `claimIsInactive` finds a healthy empty population and drops the claim.

The two walkers were repaired in #1258 by resolving the base once before walking. Nothing resolves it for this kind.

### Reachability

A monorepo whose packages are installed as links is the ordinary case a declared `root` exists for, and it is what a package manager produces: pnpm creates a junction on Windows and a symbolic link elsewhere. The two paths disagree whenever the `tsconfig.json` and the `lint.config.ts` reach the same directory by different spellings, which nothing forces them to share.

## Consequence surface

[...]

Issue #1262, opened 2026-08-19
One physical schema in two Prisma populations is parsed twice, so Prisma rejects the set


samchon:
## Problem

Two Prisma populations that reach one physical schema file through differently spelled roots hand that file to the Prisma parser **twice**, once under each spelling. Prisma parses the set as one schema, so the same `model` is declared twice and the whole set is rejected. The failure is reported against the schema, not against the configuration that produced it.

`distinctPrismaSources` (`packages/evidence/native/prisma.go:303-317`) exists to prevent exactly this, and its own comment states the invariant it does not enforce:

> The set is composed of physical files, deduplicated across bases. Prisma parses one schema at a time and a file listed twice is a duplicate declaration to its own parser, so two populations that reach one file through different roots must still contribute it once.

It deduplicates on `address.Display`, which is the *spelling* through each population's own base. Two different roots produce two different spellings for the same file, so the deduplication never fires for the case the comment names.

## Evidence

Measured against `5a885bb73`, with `documents` a link to `target`:

```go
config := decodeInventoryConfig(t, root, `{"claims":[{
  "type":"typescript","files":["src/**/*.ts"],"symbol":"type",
  "reference":[
    {"type":"prisma","root":"../documents","files":["models/**/*.prisma"],"symbol":"model"},
    {"type":"prisma","root":"../target","files":["models/**/*.prisma"],"symbol":"model"}
  ]}]}`)
addresses, _, problems := configuredPrismaAddressesWithHealth(config)
sources := distinctPrismaSources(addresses)
```

```
addresses=2 sources=2 [../documents/models/user.prisma ../target/models/user.prisma] problems=[]
```

One file on disk, two entries in the set handed to the parser.

### Why the deduplication cannot simply be moved to the physical path

Parsed units are mapped back to inventories by the same spelling. `prismaUnitsFromOutcome` (`packages/evidence/native/prisma_cache.go:185`) builds `prismaInventoriesByDisplay(inventories)` and files each unit under the `Display` its location carries. Dropping one of the two spellings from `sources` therefore leaves the population that owned it with an inventory and no units, which is a **silently empty population** — the state `graph_suppresses_derivatives_of_loader_failures_test.go` and #1254 exist to prevent, arrived at from the other direction.

The repair is therefore not a change to the deduplication key alone. It needs a decision about how one parse result reaches every address that names the same physical file.

### Reachability

[...]

PR #1274, opened 2026-08-19
fix(evidence): resolve what a base and a selected file physically are

samchon:
Cycle 3 of the population-diagnostics campaign, whose handover and entry point is #1251. Cycle 2 merged as `1b1d2a845` (#1252), closing #1248, #1249, #1254, #1258, #1264, and #1268.

This pull request owns the **two residues cycle 2 published and did not implement**, and it closes both. Both are the same subject the campaign has always had: what physical thing a declared base or a selected file actually is, and whether every consumer agrees about it.

| # | What | State |
| --- | --- | --- |
| #1269 | A link on an **ancestor** of a declared root is transparent to `Lstat` of the leaf, so a TypeScript claim rooted inside a linked directory selected nothing and deactivated in silence | closed by `74519fd6e`, extended by `577be5d29` |
| #1262 | One physical schema reached through two roots was handed to the Prisma parser twice, so the set was rejected for a duplicate model and the failure was attributed to the schema rather than to the configuration the rule composed | closed by `7a4d8c7c7`, extended by `e30f392da` |

Implementation order was #1269 then #1262: the physical identity of a selected file is decided by what its base resolves to, so the deduplication is only correct once a base whose ancestor is a link resolves.

## What #1262 decided

The issue records two resolutions and says the choice belongs to it. This takes the first, **serve it**. `distinctPrismaSources` asks the filesystem: `os.SameFile` groups the configured addresses by the file each one opens — through a hard link, through a linked or junctioned directory, and across a case-insensitive volume alike — which is the identity this repository already uses for the same question in the graph server and in the driver's resolution candidates. The second resolution, refusing the configuration, forecloses a layout a package manager produces: pnpm hard-links a package's files out of its content store and links its `node_modules` entries into `.pnpm`, so a schema installed as a dependency and rooted again at its workspace source is one file under two paths.

Deduplication alone would have moved the silence rather than removed it. The set keeps one spelling per file and everything the parse produces speaks that spelling, while each inventory speaks its own root's; keying the fan-out on the spelling would serve whichever root sorted first and leave the other population empty, which is the class #1254 closed. `prismaInventoriesBySource` resolves one through the other.

## What the reviews found inside the repairs

Four defects, none of them found by reading:

- **A chain of links past the resolver's bound on an ancestor reproduced #1269 exactly**, in total silence — measured at zero diagnostics with a resolved path still naming a junction. `resolveLinkedDirectory` now reports whether its chain ended, which costs one `os.Lstat` in the exhausted case and nothing anywhere else, and the walk stops at the first component that did not. Reading that report as *the chain is still going* was itself a regression this pull request then had to repair: a chain ending on the last hop it follows has resolved, and refusing it failed a root that works.
- **Serving both spellings made a worse diagnostic reachable than the one it removed.** A hard link is a second directory entry, so one walk of one base enumerates both, and a claim appending each inventory's declarations counted one citation twice: a correct configuration reported `Duplicate @evidence … at mirror/main.prisma:1; first declared at mirror/main.prisma:1`. Declarations, reviews, and exclusion-carrier placement are now collapsed by identity, the way the hosts on the line above already were.
- **A model the native scan could not locate went to the first source's populations alone**, and the set now spans populations, so every other one lost it with nothing said. It goes to every inventory of the set; the location stays the one path this rule can name.
- **A UNC share root resolved to a spelling one character from the declared one**, which is the same directory and compares as though it is not, so every consumer would pay a second comparison forever on a base with no link in it.

Plus two the Overall rounds found: a case that would have failed on macOS, where `/var` is a symbolic link and a real temporary directory therefore does not resolve to its own spelling; and the representative spelling being unpinned, which a mutation to the largest proved by leaving the whole suite green.

## Scope

Cycle 2's instruction stands: this cycle works the declared subject and runs no new discovery round. The campaign's owed full-scope round remains owed and is stated here rather than skipped silently.

Two things are recorded rather than repaired, so a later cycle does not rediscover them as new. The package walk in `loader.go` keeps the resolver's bound unrefused, because a package whose chain outruns it already reports an empty population — what is owed there is a better cause, not a diagnostic that does not exist. And a model the scan could not locate is now charged to every population of its set, which is visible over-attribution in place of a silent shortfall; naming the unreadable file instead would be better still, but that state is reached only by a race against a file the digest read a moment earlier.

## Against the two acceptance tables

Both issues state one, and both are met except where a row is answered differently on purpose. Stated rather than claimed:

- **#1269, "no more syscalls than the component count".** It is the component count **plus one**. The extra is `resolvedBaseDirectory`'s own verification of where the walk landed, which the leaf-only version paid too — its cost was two, not one, which is a separate correction this pull request makes to its own earlier message.
- **#1262, "each receives the file's units, addressed through its own root".** Each population receives them in its own inventory, addressed by its own root. The **units** carry the set's spelling rather than each population's, because a per-population copy would be a second identity for one model and the graph keys hosts, selections, and acknowledgements by identity — one model would owe its obligations twice and satisfy them once. The guide, both TSDoc blocks, and a case now say so.

Every other row has a case behind it, including the one that had none until round 20: two roots differing only in case, which is the shape no normalization could answer and the reason identity is `os.SameFile`.

## Verification

Exactly one diagnostic sentence changes in the whole pull request, and two cases pin it. Every repair above is mutation-verified. Each implementation commit carries one advisory Individual Self-Review, recorded here as a formal review, and the merge gate is the ordinary CI plus a complete clean Overall Self-Review of the whole base-to-head diff.


samchon (commented):
## Individual Self-Review — `fff4f76f2..74519fd6e`

One read-only reviewer over the range, in its own worktree at that sha because the main tree was mid-edit. `base.go` is byte-identical at `7a4d8c7c7`, so every finding held against the head it was measured on. Advisory only; every disposition below is the main agent's.

**Five candidates. Four accepted and repaired in `577be5d29`, one rejected.**

| # | Where | Claim | Disposition |
| --- | --- | --- | --- |
| F1 | `base.go:547-558` | **defect, measured.** A chain past the resolver's thirty-two hops on an *ancestor* reproduces #1269 exactly. `resolveLinkedPath` kept only the path, so the unfinished component was joined to the remaining segments; the filesystem opens that path, the final `os.Lstat` says directory, and the gate added to turn this into a diagnostic never fires. Measured on Windows with forty hops above the root: **zero messages**, `ok=true`, resolved path still naming a junction | accepted, repaired |
| F2 | `base.go:526-529` | the cost comment is wrong on both halves: component count **+ 1**, and the leaf alone cost **two**, not one. Measured by instrumenting `os.Lstat`: ten components, eleven stats | accepted, repaired |
| F3 | four call sites | "per base and per pass" omits the call-site multiplier. Measured on a default-base graph at eleven components: four calls, thirty-six stats per pass, where the leaf version cost eight | **rejected as a change, accepted as a fact** |
| F4 | `base.go:549-551` | **defect, measured, Windows only.** A UNC share root resolved to `\\server\share\`, one character from the declared `\\server\share`. Same directory, different string, so `entry.resolved == base.Absolute` is false forever and `typeScriptMatchBase.relativeOf` pays a second `relativeProjectPath` per source per rebuild on a base with no link in it | accepted, repaired |
| F5 | tests | #1269's volume-shape acceptance row had no test, and F4 is exactly what one would have caught | accepted, repaired |

### F1 is the finding this cycle exists for

It is #1269 one directory higher than the case #1269 opened with, and it was silent in the same way: no diagnostic at all, claim deactivated, obligation dropped. The repair asks `resolveLinkedDirectory` whether its chain ended, which costs **nothing** — every ending was already a return from inside its loop, so falling out of the loop is the one state with no proof behind it. `resolveLinkedPath` stops at the first component that did not settle.

Its new test is mutation-verified: reverting to `resolved, _ :=` prints *zero* diagnostics against an assertion expecting the refusal, which is the silence itself.

### Why F3 was rejected as a change

The reviewer's own measurement is recorded as true and the comment now states it. What it proposes — a `base.Default` short-circuit — trades a syscall saving for the risk this campaign exists to remove. The residue recorded on `2af52e86e` is that a project whose Program reports realpaths while ttsc was invoked through a long chain has no walk to notice it, and the default base is precisely the base with no `root` to correct. Thirty-six stats per **pass** is not a cost worth reopening a silence for. Recorded so it is not re-derived.

### Recorded, not repaired

The package walk in `loader.go` keeps the old behavior on purpose, and the call site now says why: a package whose chain outruns the resolver matches nothing, and `materializePackageGlobReference` already reports an empty population for it. What is owed there is a **better cause**, not a diagnostic that does not exist — and that is a different issue from this cycle's two.

### Verified sound, recorded so it is not re-derived

- Volume handling never names a different location for any reachable base shape. `C:` alone would become `C:\`, but `normalizeRootPath` refuses a drive-relative root, so it is unreachable. `\\?\C:\a` survives intact.
- `filepath.Join` never collapses wrongly: first component a link, last, middle, none, trailing separator — all resolve to the correct target.
- Bounded: one pass over a fixed component list, no recursion, no cycle risk.
- A missing path and an ancestor linking to a file both yield `ok=false`, and both are unreachable past `baseDirectoryProblem`, so `unresolvedBaseProblem` is never printed over an absent directory.
- A default base whose own ancestor is a link: addresses stay project-relative and the messages are identical, measured end-to-end.
- The positive case is load-bearing and the negative twin is not vacuous — the twin newly exercises the walker's remap, and mutating `relativeProjectPath(from, current)` in `markdown.go` fails it.
- The guide sentence, the skill bullet, and the three TSDoc blocks all state what the code does; the surviving "a linked directory inside the population is not descended into" is still true.

samchon (commented):
## Individual Self-Review — `74519fd6e..7a4d8c7c7`

One read-only reviewer over the range. Advisory only; every disposition below is the main agent's. Repairs landed in `e30f392da`.

**Six candidates. Two accepted and repaired, one message correction, three recorded.**

| # | Where | Claim | Disposition |
| --- | --- | --- | --- |
| F1 | `graph.go:295-302` | **defect, measured, high.** Serving both spellings made a worse diagnostic reachable than the one it removed. A hard link is a second directory entry, so one walk of one base enumerates both, and a claim appending each inventory's declarations saw one citation twice. On a **correct** configuration: `Duplicate @evidence for 'docs/pricing.md#discounts' on the same host at mirror/main.prisma:1; first declared at mirror/main.prisma:1`, silent without the link. Exclusion twin identical; `requireReview` emitted its unreviewed message verbatim twice | accepted, repaired |
| F2 | `prisma.go:384` | the representative spelling was unpinned: mutating it to the largest left the **whole** suite green, while the code comment and the commit message both call the choice load-bearing | accepted, repaired |
| F3 | commit message | the cost claim is imprecise in two directions: Windows `os.SameFile` opens the file on first use beyond the stat, and the digest's reads go **down** in the layout this exists for while the stats go up | accepted as a correction, recorded here |
| F4 | `os.SameFile` | a volume reporting a zero file index for every entry would merge two schemas into one and lose a population's models silently. **Not reproduced** — no such filesystem could be constructed | recorded, not repaired |
| F5 | `prisma_cache.go:190` | the `fallback` source: a model the native scan did not locate takes `set.Sources[0]` and reaches only that entry's populations. The justification held when a set belonged to one population. **No non-racy way to reach it was found** | recorded, not repaired |
| F6 | new test | the hard-link skip misnames any `os.Link` failure. Mirrors the peer `linkDirectory` convention | recorded, consistency preserved |

### F1 is the third time this campaign has caught a repair with its own class

The pattern each time: one object, several containers, per-container accumulation. The hosts on the line above were already collapsed by identity, and so are the reference units; declarations and reviews were the two lists that were not. Declarations now collapse by ID, which the graph keys them by anyway, and reviews by object, because a review carries no identity of its own.

The **exclusion-carrier placement one line further down had the same shape** and was repaired with it, which the reviewer did not raise: a declaration is outside the carriers only when no path carrying it is one, and deciding that per path put one tag in two places at once.

What makes F1 worth the emphasis is that it was *newly reachable*. Under the pre-commit spelling-keyed set the identical fixture failed with Prisma's own duplicate-model rejection — so the repair had moved one impossible repair to another, and only measurement found it.

### On F5, recorded rather than closed

The reviewer is right that the justification at `prisma_cache.go:196-200` was written when a set belonged to one population. Repairing it means choosing a semantic for a unit whose file the scan could not name, and neither of us could reach the state. It is recorded here so the next cycle does not rediscover it as new.

### Verified sound, recorded so it is not re-derived

- Six mutations, all red where they should be and green where they should not: dropping the `os.SameFile` search (Prisma's own duplicate-model error), the fan-out back to `byDisplay` (`store units = []`), forcing `SameFile` true (the negative twin fires), copying the unit per inventory, copying the declaration per inventory, and the representative spelling.
- **Order independence:** reversing the address slice gives identical output. Grouping is first-match against an equivalence relation and the representative is a minimum, so it is order-free.
- **No stale cache**, measured in one process: hard-link, then break the link and give the mirror a different model — the second cycle gives each population its own model. The digest folds the path per file, and the fan-out is recomputed from the current set on hit and miss alike, so the spellings never need to enter the key.
- `failPrismaSet` reaches every population: a broken hard-linked schema marks both `LoadFailed`, one problem each, zero units. No silent-empty on the failure path.
- The stat-failure path preserves pre-commit behavior, and the empty-entry guard was behavior-neutral — it is now written unconditionally, because every source is an address's spelling and every address owns an inventory.
- The repository-identity claim is exact: first-party `os.SameFile` use is `ttscgraph/serve.go:862` and `driver/resolution_candidates.go:213`.
- CI covers the platform: `validation-plan.cjs` selects `evidence, go, windows-go` for any `packages/evidence/**` file, so the hard-link and junction cases run on `windows-latest`.

samchon (commented):
## Overall Self-Review — sixteen rounds, the last clean

Solo, whole base-to-head diff, on `e0bf8a5b0`. Fourteen files, and every round after the first was run the way the skill prescribes: one complete pass over the whole surface, every finding collected, then all of them applied together.

**Ten of the sixteen rounds found something. Six of those findings were defects, and five of the six were in this pull request's own repairs.**

| Round | What it found | Landed as |
| --- | --- | --- |
| 1 | the Individual review of `7a4d8c7c7` plus a case that would fail on macOS, an unpinned representative spelling, three documentation surfaces, three miscounted comments | `e30f392da` |
| 2 | a model the scan could not locate reached only the first source's populations; the refusal sentence was false of the shape it now reports | `e0c811cbc` |
| 3–5 | **a regression this pull request introduced**: a chain of exactly thirty-two links resolved before `577be5d29` and was refused after it | `d1aa75be8` |
| 6 | the bound itself was unpinned — raising it by one left every case green; the carrier placement had no case at all | `3db7fefa2` |
| 7 | no case measured the shape an adopter meets, two claims over one installed schema; three comment counts wrong | `312542a3d` |
| 8 | what physical identity actually costs, including the file open `os.SameFile` makes on Windows | `db631c692` |
| 10 | **the counts were wrong again, twice**: the graph loads claim-side populations before the whole configuration, so every claim base is asked twice over | `80f228796` |
| 12 | the guide's promise about a shared schema's printed location was pinned by nothing | `651a9682c` |
| 13–14 | a doubled initialization; two conventions audited and both held | `e0bf8a5b0` |
| 15–16 | **clean.** Nothing applied | none |

### The regression is the finding worth reading

`577be5d29` taught the resolver to report whether its chain ended, and read "the loop ran out of iterations" as "the chain is still going". Those are not the same state: a chain whose last link lands on its directory exactly as the last iteration is spent has resolved. Before that commit the caller's own stat proved it and the root worked. After it the root was refused, with a sentence saying the path passes through a chain longer than this rule follows about a path that does not.

A population that loads was being failed — the opposite direction of the fault this cycle exists to repair, introduced by the repair. Nothing in the diff read wrong; it was found by deriving the loop's arithmetic instead of trusting the comment above it, and the boundary is now back where cycle 2 measured it, with a case on each side and a mutation that turns the near one red.

### The pattern across the counts

Three comment counts were wrong, and two of them for one reason: **a frequency belongs to the caller.** The resolver's cost was derived from the resolver; the graph materializes claim-side populations before the whole configuration, so every load happens twice a pass. A comment that states a frequency has to be read from where the call is made, and this cycle paid for learning that twice.

### Verification

| Gate | State |
| --- | --- |
| `go test ./native -count=1` | green at every applied round, last at this head |
| `go vet` | clean |
| `gofmt-2spaces` | every changed Go file round-trips unchanged |
| `prettier --check`, `tsc --noEmit` | clean on every changed prose and TypeScript file |
| Mutation coverage | **fifteen cases, every one verified red under a mutation of what it pins** |
| Diagnostic text | exactly **one** sentence changes in the whole pull request, and two cases pin it |
| Individual Self-Review | two, one per implementation commit, both recorded here as formal reviews |
| Ordinary CI | the merge gate, on this head |

Sweeps that produced nothing: every symbol this pull request introduced counted at its call sites; every backticked identifier in every added comment resolved to a real symbol; every added message string audited; the closing keywords counted at exactly two, one per issue.

### Recorded so a later cycle does not rediscover them

- The package walk in `loader.go` keeps the bound unrefused on purpose. A package whose chain outruns the resolver already reports an empty population, so what is owed there is a **better cause**, not a diagnostic that does not exist.
- A model the scan cannot locate is charged to every population of its set. That is visible over-attribution in place of a silent shortfall; naming the unreadable file would be better still, and that state is reached only by a race against a file the digest read a moment earlier.
- `os.SameFile` has no zero-index guard on Windows. A volume reporting a zero file index for every entry would merge two schemas into one; neither reviewer nor I could construct such a filesystem.

### Honest notes for whoever reads the merge body

The squash concatenates every commit message, and three of them assert something a later one corrects. `74519fd6e` gives the resolver's cost as the component count where it is that plus one, and the leaf as one where it was two — corrected in `577be5d29`. `577be5d29` says the second answer "costs nothing to ask", which `d1aa75be8` makes one `os.Lstat` in the exhausted case. And `7a4d8c7c7` gives the identity's cost as one `os.Stat` per address beside a read the digest already performs, which `db631c692` and `80f228796` correct in both directions and in frequency. **Take the last count each time**; every correction is in the code the commits touch, not only in this review.

### What is still owed

A complete fresh full-scope discovery round for the campaign, which #1251 §7 said was owed before cycle 2 and which both cycles have now skipped under an explicit instruction. **The campaign is active, not complete** — but with #1269 and #1262 closed, it is the first cycle to leave no residue of its own behind.

samchon (commented):
## Overall Self-Review — round 21, clean, on the merge head `eaaec7847`

Round 20 added the case-spelling case, so a complete round was owed on the head it produced. This is it, and it applies nothing.

Every mechanical sweep re-run at this head: the working tree is clean, `gofmt-2spaces` is a no-op on all eight changed Go files, `go vet` is clean, and `go test ./native -count=1` is green. Fourteen commits, fourteen files, 1272 insertions.

### The gate

| Gate | State on `eaaec7847` |
| --- | --- |
| `go` (ubuntu) | **pass**, 9m59s |
| `windows-go` | **pass**, 16m5s — the hard-link, junction, and case-spelling cases run here |
| `evidence defenses` | **pass**, 28m43s — the feature suite and the benchmark parser that reads these messages by substring |
| `typecheck` | **pass** |
| `Ubuntu` (build), `deploy` (website), `plan`, `ci` | **pass** |
| Individual Self-Review | two, one per implementation commit, both recorded here |
| Overall Self-Review | twenty-one rounds, rounds 15, 16, 19 and this one clean |

`windows-go` matters more than usual for this pull request: three of its sixteen cases are answerable only on a volume that hard-links, junctions, and folds case, and that lane exists for this package because cycle 2 closed #1264.

### What merges

Both residues cycle 2 left, and the five defects the reviews found inside their repairs — one of which was a regression this pull request had itself introduced two commits earlier, refusing a link chain that resolves. The campaign remains **active** for one reason only: the complete fresh full-scope discovery round #1251 §7 has asked for since before cycle 2, which both cycles skipped under instruction. Every accepted issue is now closed, and this is the first cycle that leaves no residue of its own.

PR #1275, opened 2026-08-19
feat(graph): finish the citation graph — artifact refresh and cold-index cost (#1239)

### LINKED ISSUE #1239: Materialize an artifact unit as a graph node, so a citation is a relation rather than a token
## Problem

`@ttsc/graph` now carries a declaration's documentation tags and resolves the
inline links among them to edges. A citation whose target is a TypeScript symbol
is therefore a real relation in the graph; a citation whose target is a document
section, an API operation, or a data model is a **string**.

That asymmetry is visible in one dump. Given this declaration:

```ts
/**
 * Renders the stacking notice.
 *
 * @evidence docs/discount.md#coupon-stacking States the per-issuer limit.
 * @evidence POST:/orders/{orderId}/coupons Explains the rejection.
 * @evidence {@link ISale} The complete sale contract is mirrored here.
 * @evidenceExclude prisma:Coupon.issuedAt Not surfaced in this notice.
 */
export function renderNotice(): string { ... }
```

`ttscgraph dump` returns:

```
carried as TEXT on the node (uninterpreted)
  @evidence         docs/discount.md#coupon-stacking States the per-issuer limit.
  @evidence         POST:/orders/{orderId}/coupons Explains the rejection.
  @evidence         {@link ISale} The complete sale contract is mirrored here.
  @evidenceExclude  prisma:Coupon.issuedAt Not surfaced in this notice.

carried as a RESOLVED EDGE between nodes (checker fact)
  src/notice.ts#renderNotice:function  --doc_ref-->  src/sale.ts#ISale:interface
```

Three of the four targets have no other end to point at, because the graph holds
no node for a Markdown section, a Prisma model, or a Swagger operation.

**What that costs, precisely.** The reverse question already works: a `lookup`
whose query is `docs/discount.md#coupon-stacking` answers with the declarations
citing it, keyed on the token. What it cannot answer is anything about the thing
cited:

- **The heading's own text.** `docs/discount.md#coupon-stacking` is an anchor.
  What section is that? The agent has to open the document. This is the concrete
  loss, and it is the one an index is supposed to remove.
- **What else is under it.** A heading contains subsections; a Prisma model
  contains columns and relations. A citation of the parent covers them, and the
  graph cannot say what they are.
- **What a document contains at all**, which is a cheaper question (one file
  read) and deliberately not the motivation here.

**What is explicitly NOT a motivation.** Whether a target resolves is
`evidence/graph`'s judgement, delivered as a compile error, and duplicating it
here would put two answers in the repository. The graph reports; the linter
judges. Keep that split.

## Why this needs a transport, and why the previous cycle did not build one

The addresses in those targets are not parseable by the graph. Each is produced
by a body of rules that lives in `packages/evidence/native` and belongs there:

| Target form | What produces the address |
| --- | --- |
| `docs/discount.md#coupon-stacking` | Markdown heading scan, GitHub-style anchor generation, H1–H4 outline nesting |
| `prisma:Coupon.issuedAt` | Prisma's own parser, which is what separates a column from a relation (a relation has two sides and only one usually carries `@relation`) |
| `POST:/orders/{orderId}/coupons` | Swagger 2.0 / OpenAPI 3.x normalization through `@typia/utils`, `$ref` resolution, uppercase method joining |

Re-deriving any of them in the compiler host would be a second implementation of
a published contract, and it would drift. So the units have to arrive from the
rule that owns them.

The previous cycle (#1229, #1230) deliberately did not build that transport,
because it did not need one: tags are read from the compiler's own AST as
uninterpreted text, and the reverse index keys on the token. That work is the
foundation this issue builds on, and it stays useful whether or not this ships.

## What exists to build the transport out of

Read these before designing; each is a real seam, and the gaps between them are
the work.

**A rule can already publish a value.** `ProjectContext.SetState`
(`packages/lint/rule/project.go:236`) attaches one contributor-owned value to a
project rule's evaluated result, and file rules in the same Program cycle read it
back through `ProjectRuleResult.State`. `@ttsc/evidence` already uses it: its
`graphCycleState` carries a `graphCorpus` of the Markdown, Prisma, and Swagger
inventories Check built (`packages/evidence/native/model.go:24-42`).

**That value already leaves the rule.** `Hints`
(`packages/evidence/native/hints.go:81`) projects the same corpus into editor
completions, and `RunLSPHints` (`packages/lint/linthost/hints.go:23`) serializes
the result as JSON over a verb. So "a corpus crosses the process boundary" is
solved for one consumer already.

**But the state itself is in-process only.** `SetState`'s own contract says the
host "does not serialize the value or retain it for a later watch or LSP
rebuild". `Hints` works because it projects to a flat, serializable shape
(`rule.Hint`) inside the same process. A unit inventory is richer than that.

**The unit shape the transport would carry** is `evidenceUnit`
(`packages/evidence/native/model.go:213`): `ID`, `ParentID`, `Target`,
`Identity` (the address as unjoined segments, so a literal dot cannot collapse
into qualification), `Aliases`, `Type`, `Symbol`, `Path`, `Line`, `Readable`,
`Hidden`, `ValueSpace`. `Readable` is the human-facing name — the heading text
this issue exists to surface — and `ParentID` is the containment this issue
wants as `contains` edges.

**The graph producer cannot link the rule.** `@ttsc/evidence` ships Go **source**
that `ttsc` copies into `@ttsc/lint`'s module and builds per project; it is not a
module `ttscgraph` can import, and it must not become one — `AGENTS.md` forbids
the compiler host knowing a specific contributor. Whatever is built has to be a
contributor-agnostic channel that evidence happens to be the first user of.

**The graph producer already knows about linked plugins.**
`packages/ttsc/cmd/ttscgraph/serve.go:248` checks
`Program().HasLinkedProgramPlugins()` and forces a reload when one is active, so
the serve session is not naive about plugins.

**The rule needs no checker.** `evidence/graph` declares
`NeedsTypeChecker() false` (`packages/evidence/native/graph.go:16`), so producing
the corpus costs a parse and bind, not a checker build. That matters for the
cost budget below.

## The hard parts, stated as constraints rather than solutions

**1. One generation, or an honest claim about two.** `ITtscGraphDump`'s contract
is that its nodes, edges, spans, and diagnostics came from one `Program`
(`packages/graph/src/structures/ITtscGraphDump.ts:53`), and the graph carries no
per-edge trust flag by design. Units produced by a separately-loaded lint Program
are not from that Program. Either they are produced inside the graph's own
generation, or `provenance` grows an honest second producer entry and the
consumer can tell. Do not let a stale overlay ride the same claim as the
compiler facts.

**2. Documents enter the build universe.** `IUniverse`
(`ITtscGraphDump.ts:128`) fingerprints the inputs that decide which files the
program contains, and a change to any of them invalidates everything. If a
Markdown file's headings are nodes, then editing that Markdown must refresh the
graph — which means `.md`, `.prisma`, Swagger documents, **and `lint.config.ts`**
become graph inputs. Consequence: editing a document invalidates the code graph.
That is correct, and it is a real cost on a documentation-heavy repository.

**3. The addresses do not fit the node-id grammar.** A node id is
`path#qualifiedName:kind` and is position-invariant
(`packages/graph/src/structures/ITtscGraphNode.ts:9`). `prisma:Coupon.issuedAt`
carries no path **on purpose** — a Prisma model name is unique across the schema
folder so moving a model between files cannot break a citation — and
`POST:/orders` has no file at all. Using the evidence address verbatim as the
node id is the obvious answer and has a real benefit: the token an author wrote
becomes the handle an agent submits. It also means the id parser must be gated by
kind, since `TtscGraphNodeId.parse` assumes the TypeScript grammar.

**4. Cost, measured against the product's headline claim.** `@ttsc/graph`
advertises a 29-second cold index on a 3M-line repository, and that number is the
argument against every comparator. A second parse of the documentation corpus,
plus whatever the transport costs, lands on it directly. Measure before
committing; a design that doubles cold time for projects that use the convention
is not obviously worth the heading text.

**5. The MCP surface must stay an index.** `.agents/skills/project/graph.md`
forbids inlining bodies. A Markdown section node carries its **heading text and
its span** — never its content. The section body is read from the file when the
agent actually needs it, exactly as a function body is.

## Approach

The invariant: **the graph holds what a rule materialized, and interprets none of
it.** The rule owns unit identity, addresses, hierarchy, and withdrawal; the
graph owns nodes, edges, spans, and the response shape. No coverage, no
cardinality, no policy, no diagnostics cross the boundary — those are the
linter's product and are already delivered as compile errors.

Sketch of the shape, to be settled by whoever takes this:

- **Channel.** A contributor-agnostic way for a project rule to publish
  structured facts that outlive its process, alongside the existing hints verb.
  The published shape is a graph fact — a node with an address, a kind, a
  readable name, a span, and a parent — not an evidence-specific one.
- **Node kinds.** `markdown_document`, `markdown_section`, `prisma_model`,
  `prisma_column`, `prisma_relation`, `swagger_operation`, matching the evidence
  symbol selectors. Both `TtscGraphDumpNodeKind` and `TtscGraphNodeKind` move,
  which forces a `DUMP_SCHEMA_VERSION` bump.
- **Edges.** The `doc_ref` kind already exists and already means "this
  declaration's documentation references that node" — a path target resolving to
  an artifact node is the same relation, so it needs no new kind. `contains`
  carries the heading and model hierarchy, as it already does for declarations.
- **Capability.** A new member so a consumer degrades on a declared claim rather
  than guessing from absence, exactly as `docTags` does today.
- **Ranking.** Artifact nodes must be kept out of `tour` seeding, the way
  `closure` nodes are (`ITtscGraphNode.ts:50`) — a tour is asked what the project
  is and how it runs, and a document section is neither.

Near-100% coverage of changed behavior; every predicate needs a negative twin.

**Positive**

1. A Markdown H2 cited by a declaration is a node carrying its heading text and
   its span, and the citation is an edge to it.
2. Its subsections are `contains`-children of it.
3. A Prisma model, a column, and a relation are distinguishable nodes.
4. A Swagger operation cited as `POST:/orders` is a node.
5. `details` on an artifact node returns its readable name and span, never its
   body.
6. `lookup` on the address returns both the artifact node and its citing
   declarations.

**Negative**

7. A project with no evidence configuration produces a dump identical to today's
   apart from the schema version.
8. A document not selected by any configured reference contributes no node.
9. A target naming nothing produces no fabricated node — the tag text still
   carries it, and the linter is what reports it.
10. `tour` seeds no artifact node.
11. No coverage, policy, or diagnostic state appears anywhere in the dump.

**Boundary**

12. A withdrawn unit (`@internal` on a Prisma model) contributes no node, matching
    the rule's own answer.
13. A unit reachable by two addresses (an alias) is one node.
14. Editing a Markdown heading refreshes the graph in a resident session.
15. Editing `lint.config.ts` invalidates the graph.

**Cost**

16. Cold index time measured on a real repository, before and after, and reported
    in the pull request. A regression on the headline number is a reason to stop
    and reconsider, not a detail.

## Coordination

- Builds on #1229 and #1230 (merged in the documentation-tag cycle), which
  established `docTags`, the `doc_ref` edge kind, the capability pattern, and the
  address-shaped reverse index.
- Touches `packages/lint` (the channel), `packages/evidence` (publishing through
  it), and `packages/ttsc` (consuming it) — three packages whose boundaries this
  issue is mostly about respecting.
- The `@ttsc/evidence` domain model is `.agents/skills/project/evidence/SKILL.md`;
  read it before deciding what a unit is, because most of the hard cases
  (hierarchy, withdrawal, merged identities, aliases) are already settled there.

### LINKED ISSUE #1276: perf(ttsc): plugin discovery costs a dependency walk to answer that a project has no plugins
## Problem

`resolveCapabilityPlugins` answers one narrow question — *which configured plugins declare capability X* — and pays full plugin discovery to answer it, including package auto-discovery over the project's dependency closure. On a project that configures **no** plugin at all, the answer is the empty array and the walk was pure waste.

Measured on the graph benchmark corpus, in-process against the built `@ttsc/graph`, six calls each:

| fixture | TS files | `publishArtifacts` (median) | answer | cold `ttscgraph dump` | share |
| --- | ---: | ---: | --- | ---: | ---: |
| `zod@graph` | 372 | 356 ms | `null` — no publisher | ~11 s | ~3% |
| `nestjs@graph` | 1,546 | 701 ms | `null` — no publisher | ~3.2 s | **~22%** |

A `--cpu-prof` of eight `loadProjectPlugins` calls on `zod@graph` attributes the time almost entirely to filesystem syscalls:

```
   816.0 ms  realpath
   620.1 ms  open
   332.4 ms  stat
   319.3 ms  lstat
   159.2 ms  readFileUtf8
   102.1 ms  readlink
    18.6 ms  readJsonFile        readConfigJson.js:32
```

`readProjectConfig` on its own is **2 ms**. The cost is the dependency walk.

## Why it surfaced now

`ttsc`'s own build has always paid this, and there it is amortized against a compile. #1239 gave `@ttsc/graph` a reason to ask the same question, so a tool whose cold index the benchmark publishes now imports the cost — and pays it before it spawns anything, on every project, whether or not one configures an evidence publisher. Full numbers and method are in samchon/ttsc#1275.

## What is not the fix

- **A cheap pre-check.** Deciding whether any dependency declares a plugin *is* the walk. Reading the tsconfig's `plugins` array alone would miss auto-discovery, which is a supported way to configure one.
- **Overlapping it with the native process.** `ttscgraph` reads `--artifacts` at flag-parse time, before it loads a Program. Making that read lazy so a launcher could publish in the Program's shadow needs a handshake this protocol does not have.
- **Caching it in `@ttsc/graph`.** The cost belongs to discovery, and every consumer of `loadProjectPlugins` pays it. Solving it one caller up leaves the others.

## Direction

`loadProjectPlugins` already computes `hostInputs` and `hostInputHashes` — the exact files whose state decides whether a loaded selection is still valid. That is a cache key, and it is already produced. A cross-process resolution cache under `node_modules/.cache/ttsc/`, keyed on it, would make a warm project pay a hash check rather than a dependency walk, and would speed up `ttsc` builds as much as it speeds up the graph.

Invalidation is the whole risk: a resolution cache that goes stale silently answers "no plugin declares this" for a project that just added one, which is a wrong answer that looks exactly like a correct one. Whatever is built needs a negative twin for every predicate, and needs to fail closed — an unreadable or ambiguous cache entry must mean *walk*, never *empty*.

**Positive**

1. A warm project with an unchanged dependency closure resolves without the walk, and the measured cost drops on both fixtures above.
2. The resolved answer is byte-identical to the uncached one for a project that configures a plugin, and for one that configures none.

**Negative**

3. Adding a plugin to `tsconfig.json` invalidates.
4. Adding a dependency that declares `ttsc` in its own manifest invalidates — the auto-discovery half, which a tsconfig-only key would miss.
5. Removing a plugin invalidates.
6. A cache entry that cannot be read or parsed causes a full walk, never an empty answer.

**Cost**

7. Cold index time re-measured on `zod@graph` and `nestjs@graph`, before and after, and reported.

### LINKED ISSUE #1278: perf(graph): a document edit republishes artifacts by spawning two sidecars and building a second Program
## Problem

`@ttsc/graph` republishes a project's artifacts by spawning the plugin's sidecar twice — once for `graph-nodes`, once for `project-inputs` — and the first of those builds its own TypeScript Program for a project the resident `ttscgraph` already has loaded. It does this from scratch on every republish, and a republish is what a document edit costs.

Measured on the `test-evidence` fixture built by `test_evidence_graph_refreshes_a_document_edit_in_a_resident_graph_session` — a real project with `@ttsc/evidence` configured, a 40-section document, 40 citations, plugin binary warm in the suite cache:

| step | median | what it does |
| --- | ---: | --- |
| `resolveCapabilityPlugins` | 6.6 s | plugin discovery over the dependency closure (#1276) |
| `graph-nodes` verb | 5.0 s | sidecar spawn; builds its own Program and runs the rule |
| `project-inputs` verb | 4.4 s | sidecar spawn; no Program |
| **`publishArtifacts` total** | **15.3 s** | |
| `artifactsAreStale` | 3.3 ms | the per-request staleness check, for contrast |

That fixture is unkind to the first term — its `node_modules` is junctions into the ttsc checkout, where `zod@graph` and `nestjs@graph` measured 0.36 s and 0.70 s for the same call — but the two spawns and the second Program are structural, and remain seconds on any project.

Two consequences, and the second is the sharper one:

1. A document edit costs that before the session's next answer, and the first request of a session pays it too.
2. `publishArtifacts` uses `spawnSync`, so the MCP server's event loop is **blocked** for the duration. Other tool calls queue behind it, and an in-flight request cannot be cancelled — `refresh` declines to *start* a republish for an already-aborted request, which is as much as a synchronous call can honour.

## Direction

`@ttsc/lint` already runs a resident daemon for exactly this: `lsp-serve` keeps a warm Program across verbs, and `ttscserver` routes `lsp-hints` and `lsp-diagnostics` through it, falling back to spawn-per-verb for a sidecar that predates it. `@ttsc/graph` spawns cold every time and never asks the daemon.

Speaking that protocol would amortize the Program across republishes and let the whole exchange be asynchronous, which is what removes the event-loop block rather than merely shortening it. The fallback path is already the behaviour that exists today, so a sidecar that cannot serve keeps working.

The other two terms have their own answers: discovery is #1276, and the second spawn disappears if a verb can answer both questions — though that is a change to `@ttsc/lint`'s published verb set, not to its caller.

**Positive**

1. A second republish in one session reuses the daemon's Program, and the measured cost drops on the fixture above.
2. The artifacts published through the daemon are byte-identical to those published by the spawn-per-verb path.
3. The graph request that triggers a republish can be cancelled while the republish is in flight.

**Negative**

4. A sidecar that does not implement `lsp-serve` still publishes, through the existing spawn path, with no change in the answer.
5. A daemon that dies mid-session falls back rather than failing the graph request.

**Boundary**

6. Two graph requests arriving while a republish is in flight do not each start one.

**Cost**

7. The table above re-measured, before and after, and reported.

samchon:
Finishes `#1239` — and, on the way, finds that the feature it finishes had never worked.

## The defect this cycle actually found

`@ttsc/graph` spawned the plugin's sidecar with `--cwd`, `--tsconfig`, and `--plugins-json`, and that is not enough. A rule resolves its own inputs — the documents a claim reads, a Prisma schema, an OpenAPI file — against the project root, and `ttscserver` hands that root over as `--project-context-json` whenever a plugin's descriptor asks for one. Without it, `@ttsc/evidence` resolves no root, its claims deactivate against an empty population, and it answers with an empty set.

Nothing caught it because an empty set is the *correct* answer for most projects. Exit 0, an empty array, an unclaimed capability, and a dump identical to the one before the feature existed — which is also exactly what success looks like for a project that configures no publisher. The Go cases covered the host's filter over what a contributor published and the empty answer for a project with none; the graph cases built their dumps by hand. Not one case anywhere drove a rule that actually publishes.

So the first thing here is a case that does: a real project, a real `@ttsc/evidence` configuration, a real resident graph session, and a document edit.

## Acceptance

- **14 — a document edit refreshes the artifacts in a resident session.** Driven end to end against the real toolchain. The session republishes when the inputs behind the set move, and the native session reprojects the resident Program without a compiler reload.
- **15 — a `lint.config.ts` edit invalidates.** The same state notices both, because the `project-inputs` snapshot carries the plugin's configuration files alongside the documents its rules declared.
- **16 — cold index time, before and after, reported.** Two comments below: the native dump does not move on `zod@graph` or `nestjs@graph`, and the publisher-side cost is stated with its breakdown.

## Design

The documents behind an artifact are deliberately not Program inputs; that is what keeps a Markdown edit from costing a typecheck. The cost of that property is that no compiler input moves when a document does, so the refresh is a second invalidation with its own inputs — taken once from the `project-inputs` verb `@ttsc/lint` already publishes, then watched locally at 3.3 ms per request rather than by re-asking the sidecar.

The wire carries three statements, not two: absent is a client with no opinion and the applied set stands; empty withdraws it, which is what a project whose publisher was removed needs to be able to say; a path is the set to apply. Both readings are one pointer apart and each is pinned by its own case.

## The two costs measured on the way, and fixed here

Acceptance 16 asks for the cold index cost, and answering it turned up two regressions the feature had introduced rather than inherited. Both are fixed in this branch rather than left for later, because both are paid by the acceptance being claimed.

- **#1276 — plugin discovery walked the dependency closure to answer that a project has no plugins.** 356 ms on `zod@graph` and 701 ms on `nestjs@graph`, for projects that configure nothing, once per dump and once per resident child. Now a fail-closed disk cache keyed on the project: 42 ms and 60 ms.
- **#1278 — a republish spawned two sidecars and built a second Program, synchronously.** 15.3 s per document edit on a real evidence project, blocking the MCP server's event loop for the duration. The two project-shaped verbs now join `@ttsc/lint`'s existing `lsp-serve` daemon, which keeps a warm Program and a memoized rule configuration across them, and the republish is asynchronous.

## Invariants a later change must not quietly break

Each of these is a place where the wrong choice produces an answer indistinguishable from a correct one.

1. **The sidecar is passed `--project-context-json`.** Without it the rule answers with an empty set, which is the correct answer for most projects — the defect at the top of this description.
2. **The `artifacts` request field is a three-state pointer.** Absent means no opinion and the server's set stands; `""` withdraws it; a path applies it. Collapsing two of them either empties a session driven by the startup flag or keeps answering with a deleted plugin's results.
3. **The fingerprint is taken before the set is published, never after.** Taken after, an edit landing in the gap reads as already accounted for.
4. **`reloadDirectories` is non-recursive.** `@ttsc/lint` publishes them as resolution anchors and the LSP host watches `<dir>/*`; walking them recursively states the whole dependency tree before every request.
5. **A rule that fails withholds its artifacts, by design.** That is why the end-to-end case holds the citation's anchor fixed and edits only the heading text.
6. **The discovery cache is fail-closed.** Anything it cannot prove means walking again. An entry recording no inputs matches everything, so it is refused on its shape.
7. **The daemon's `null` means "ask directly", never "there are none".** Inverting that reproduces the defect at the top.
8. **A republish drops the daemon's warm Program.** Claim activation depends on which sources exist, so reusing a Program from before the developer's code edits brings stale artifacts back.

Closes #1239
Closes #1276
Closes #1278

https://claude.ai/code/session_01KC6FAKR1XAYP9F1ewPqycQ


samchon:
## Acceptance 16 — cold index time, before and after

**Before** is `a2a9b472f`, the commit preceding the artifact-node feature (`76e416ae4`). **After** is this branch. Both `ttscgraph` binaries were built from their own tree with the same toolchain. Fixtures are the graph corpus's own checkouts, installed from their lockfiles.

Host is a Windows desktop and is not quiet: run-to-run spread on a single arm reaches ±35%, which is larger than any effect being looked for. So the binaries were alternated **within a pair**, and the paired delta is the reported statistic; two independent medians on this host would be noise compared with noise.

| fixture | TS files | nodes | edges | paired wall delta (after − before) |
| --- | ---: | ---: | ---: | --- |
| `zod@graph` | 372 | 4,615 | 22,483 | median **−608 ms** over 7 pairs; **+69 ms** over a separate 9 pairs |
| `nestjs@graph` | 1,546 | 6,741 | 15,275 | median **−40 ms** over 5 pairs |

The sign flips between pairs and between runs. **The native dump did not move**, which is what the code says too: without `--artifacts`, the after-binary runs `LoadArtifacts("")` and does nothing further.

### The number that did move, and it is not the binary

`@ttsc/graph` now calls `publishArtifacts` before it spawns anything, and that call runs plugin discovery through `resolveCapabilityPlugins` **whether or not the project configures a publisher**:

| fixture | `publishArtifacts` (median of 6) | result | share of that fixture's cold dump |
| --- | ---: | --- | ---: |
| `zod@graph` | 356 ms | `null` — no configured publisher | ~3% |
| `nestjs@graph` | 701 ms | `null` — no configured publisher | **~22%** |

A CPU profile of the call attributes essentially all of it to filesystem work — `realpath`, `open`, `stat`, `lstat` — inside package auto-discovery, which walks the dependency closure looking for a `ttsc` descriptor. `readProjectConfig` itself is 2 ms.

This is once per `dump` and once per resident child, not per question. But it is readiness, it is paid by every project including those with no evidence configuration, and on `nestjs` it is a fifth of the cold index. Acceptance 16 calls a regression on the headline number a reason to stop and reconsider, so it is reported here rather than folded into the feature's cost: **filed as its own issue**, because the fix belongs to `ttsc`'s plugin discovery and benefits every consumer of it, not to this feature.

Two things worth recording about why it is not fixed here. Discovery cannot be short-circuited by a cheap pre-check — deciding whether any dependency declares a plugin *is* the walk. And it cannot be overlapped with the native process, because `ttscgraph` reads `--artifacts` at flag-parse time, before it loads a Program; making that read lazy so the launcher could publish in the Program's shadow would need a handshake this protocol does not have.

### Method

- `ttscgraph dump --cwd <fixture> --tsconfig <tsconfig>`, output discarded, one warm-up pair discarded.
- `publishArtifacts` timed in-process against the built `@ttsc/graph`, 6 calls, median reported.
- `typeorm@graph` was excluded: its root `tsconfig.json` does not load, so both arms were timing an error path.

samchon:
## Acceptance 16, second half — what the feature costs when it is actually used

The table above measured projects that configure **no** publisher, which was the only thing measurable at the time: the channel was publishing nothing for every project, and that is [what this branch fixed](https://github.com/samchon/ttsc/pull/1275/commits). With artifacts genuinely flowing, the publisher-side cost can be stated.

Measured on the `test-evidence` fixture the new end-to-end case builds — a real project with `@ttsc/evidence` configured, a 40-section document, and 40 citations — median of the samples taken:

| step | median | what it does |
| --- | ---: | --- |
| `resolveCapabilityPlugins` | **6.6 s** | plugin discovery over the dependency closure (samchon/ttsc#1276) |
| `graph-nodes` verb | **5.0 s** | sidecar spawn; builds its own Program and runs the rule |
| `project-inputs` verb | **4.4 s** | sidecar spawn; no Program |
| **`publishArtifacts` total** | **15.3 s** | the three together |
| `artifactsAreStale` | **3.3 ms** | the per-request staleness check |

Two readings, and they point in opposite directions.

**The per-request watch is free.** 3.3 ms before every graph request is the whole cost of noticing a document edit at all, which is what the `stat`-and-walk design was for. Nothing here argues against it.

**The republish is not.** A document edit costs ~15 s before the session's next answer, and the first request of a session pays the same. The fixture is unkind to the first term — its `node_modules` is junctions into this repository, where `zod@graph` and `nestjs@graph` measured 0.36 s and 0.70 s for the same call — but the two sidecar spawns are structural, and one of them builds a second Program for a project the resident `ttscgraph` already has loaded.

That last part is the shape of the fix rather than a tuning knob: `@ttsc/lint` already runs a resident daemon (`lsp-serve`) that keeps a warm Program across verbs, and `ttscserver` uses it for exactly this reason. `@ttsc/graph` spawns cold every time. **Filed separately** — it is a new protocol client, not a tightening of this change, and #1239's acceptance is about the refresh being correct, which it now is and is proven end to end against the real toolchain.

Method: `@ttsc/evidence` configured with one TypeScript claim over a Markdown reference; plugin binary warm in the suite cache; five samples for the total, three per step; medians reported.

samchon:
# 인수인계서 — `fix/artifact-coverage-holes` (PR #1275)

`#1239`의 잔여 acceptance(14·15·16)를 마무리하려다, **이 기능이 병합 이후 실제 프로젝트에서 단 하나의 아티팩트도 발행한 적이 없었다**는 사실을 발견한 사이클입니다. 그 원인과 그로부터 파생된 성능 결함 두 건까지 모두 이 브랜치에서 구현했습니다.

배터리 사정으로 중단합니다. **남은 일은 CI 확인과 머지뿐**이며, 아래에 정확히 무엇이 남았고 무엇을 어떻게 검증했는지 적습니다.

---

## 1. 지금 상태 한 줄 요약

- 구현·로컬 검증: **완료**
- 미완: **최종 head(`090a057d4`)에 대한 CI 그린 확인 → squash 머지 → `#1239` 종료**
- `#1276`, `#1278`은 이미 구현·종료 처리됨(이 PR 안에서 정복)

---

## 2. 이 사이클이 실제로 찾은 것

`@ttsc/graph`는 플러그인 사이드카를 `--cwd`, `--tsconfig`, `--plugins-json`만으로 띄웠습니다. 부족합니다. 규칙은 자기 입력(클레임이 읽는 문서, Prisma 스키마, OpenAPI 파일)을 **프로젝트 루트 기준으로** 해석하고, `ttscserver`는 디스크립터가 요구할 때 그 루트를 `--project-context-json`으로 건네줍니다. 그것이 없으면 `@ttsc/evidence`는 루트를 못 잡고, 클레임이 빈 population에 대해 비활성화되며, **빈 집합**을 반환합니다.

아무도 못 잡은 이유: **빈 집합이 대다수 프로젝트에서는 정답**이기 때문입니다. exit 0, 빈 배열, capability 미선언, 그리고 기능 도입 전과 동일한 dump — 발행자가 없는 프로젝트의 성공 모습과 바이트 단위로 같습니다. Go 테스트는 호스트의 필터와 "발행자 없음"만 덮었고, graph 테스트는 dump를 손으로 만들었습니다. **실제로 발행하는 규칙을 구동하는 케이스가 어디에도 없었습니다.**

→ `97105dc1b`가 원인 수정, 그리고 그 케이스를 만든 것이
`tests/test-evidence/src/features/test_evidence_graph_refreshes_a_document_edit_in_a_resident_graph_session.ts`입니다.

---

## 3. Acceptance 대응

| # | 항목 | 상태 | 어디서 |
| --- | --- | --- | --- |
| 14 | 문서 편집이 상주 세션의 그래프를 갱신 | 완료 | 위 test-evidence 케이스(실 툴체인 종단), `serve_adopts_a_republished_artifact_set_test.go` |
| 15 | `lint.config.ts` 편집이 무효화 | 완료 | `test_ttscgraph_artifacts_notice_a_document_or_config_edit.ts` |
| 16 | 냉시작 인덱스 시간 before/after 보고 | 완료 | PR 코멘트 2건(아래 §6) |

---

## 4. 커밋별 지도 (`master..HEAD`)

읽는 순서는 아래 그대로가 가장 이해가 빠릅니다.

1. `60e089351` — 지난 사이클이 남긴 커버리지 구멍 2개, 도달 불가 guard 삭제
2. `6e98af112` — **acceptance 14·15 본체.** `project-inputs` 스냅샷을 한 번 받아 로컬 stat/walk로 감시, 변하면 재발행. Go 상주 세션은 `rebuild`(Program 재사용)로 오버레이만 재투영
3. `fc36508e9` — prettier
4. `14478472b` — 셀프리뷰 1라운드 6건 일괄 수정(없는 파일 fail-closed, 비재귀 패턴, "발행자 없음"도 감시 대상, 실패한 재발행이 삭제된 플러그인 결과를 붙들지 않게, 디렉터리 중복 병합, NUL 구분자)
5. `0f6b0fde6` — **와이어에 "이제 없음"을 말할 수단 부여.** `artifacts`를 포인터 3상태로: 없음(의견 없음) / 빈 문자열(철회) / 경로(적용)
6. `fff5beb71` — 제품이 실제로 쓰는 shard 프로토콜 경로 커버
7. `f24f69640` — 지문을 발행 **이전에** 채취(stale 방향 실패 제거)
8. `46ec1c190` — 클라이언트가 와이어에 실제로 무엇을 쓰는지 고정(fake가 요청/인자 기록)
9. `97105dc1b` — **★ 원인 수정.** `--project-context-json` 전달. 부수로 `reloadDirectories` 비재귀 정정, e2e 시나리오 정정
10. `8eab03d44` — 스냅샷이 명시한 `root` 기준으로 경로 해석
11. `1d21650b0` + `1252134b2` — `packages/graph/` 변경이 evidence 레인을 고르도록 CI 계획 수정(+ 실수로 딸려온 스크래치 제거)
12. `f8c477205` — 요청마다 아티팩트 파일 전체를 읽고 해시하던 것 제거(stat 선행), 이중 읽기 TOCTOU 제거
13. `0e8c5cbf6` — 이미 취소된 요청이 블로킹 재발행을 시작하지 않게
14. `46d32781e` — **#1276 정복.** 플러그인 탐색 결과 디스크 캐시
15. `090a057d4` — **#1278 정복.** 재발행을 상주 사이드카(`lsp-serve`)로

---

## 5. 설계에서 반드시 유지해야 할 불변식

다음은 전부 "고치면 조용히 틀리는" 지점입니다. 리팩터 시 주의하십시오.

1. **사이드카에는 반드시 `--project-context-json`을 넘긴다.** 안 넘기면 빈 집합이고, 그건 정상 답과 구분되지 않습니다. (`resolveCapabilityPlugins`가 `projectContext`를 실어 보냄)
2. **`artifacts` 요청 필드는 포인터 3상태다.** 없음=의견 없음(서버 보유분 유지), `""`=철회, 경로=적용. 둘을 하나로 합치면 시작 플래그로 구동되는 세션이 비거나, 삭제된 플러그인의 결과가 영원히 남습니다.
3. **지문은 발행 전에 채취한다.** 뒤에 채취하면 그 사이 편집이 "이미 반영됨"으로 읽힙니다.
4. **`reloadDirectories`는 비재귀다.** `@ttsc/lint`가 해석 앵커로 발행하며 LSP 호스트도 `<dir>/*`만 감시합니다. 재귀로 다루면 요청마다 node_modules 전체를 훑습니다.
5. **규칙이 실패하면 아티팩트는 보류된다**(설계상). 그래서 e2e는 앵커를 고정하고 제목 텍스트만 바꿉니다. 주소가 움직이면 인용이 깨져 규칙이 실패하고, 그래프가 비는 건 정상 동작입니다.
6. **탐색 캐시는 fail-closed다.** 증명 못 하면 무조건 다시 걷습니다. 입력 기록이 비어 있는 엔트리는 "아무것과도 일치"하므로 형태만으로 거부합니다.
7. **데몬의 `null`은 "직접 물어라"이지 "없다"가 아니다.** 이걸 뒤집으면 §2의 결함이 재현됩니다.
8. **재발행마다 데몬의 warm Program을 버린다.** 클레임 활성화가 소스에 의존하기 때문입니다. 성능을 이유로 이걸 빼면 stale 아티팩트가 돌아옵니다.

---

## 6. 측정치 (acceptance 16)

PR 코멘트 2건에 원문이 있습니다. 요약:

**네이티브 dump — 변화 없음.** 페어링 델타 중앙값 `zod@graph` −608 ms / +69 ms, `nestjs@graph` −40 ms. 부호가 뒤집히며 실행 간 편차 안입니다. 코드상으로도 `--artifacts` 없이는 하는 일이 없습니다.

**런처 측 — 개선됨.** `publishArtifacts`(발행자 없는 프로젝트):

| fixture | 도입 직후 | #1276 이후 |
| --- | ---: | ---: |
| `zod@graph` | 356 ms | **42 ms** |
| `nestjs@graph` | 701 ms | **60 ms** |

**발행자 있는 프로젝트의 재발행**(test-evidence fixture, 40 섹션/40 인용) 도입 직후: `resolveCapabilityPlugins` 6.6 s + `graph-nodes` 5.0 s + `project-inputs` 4.4 s = **15.3 s**. 상시 감시(`artifactsAreStale`)는 3.3 ms로 무시할 수준.
→ #1276이 첫 항을, #1278이 나머지 두 항의 프로세스·플러그인 로드·설정 평가분을 제거했습니다. **재측정은 아직 못 했습니다**(§8).

---

## 7. 검증 방법 (재개 시 그대로 쓰십시오)

```bash
# 실 툴체인 종단 (가장 중요, ~2-3분, 플러그인 캐시 콜드면 더 김)
cd tests/test-evidence && npm start -- --include=graph_refreshes_a_document

# 그래프 쪽 아티팩트 케이스 전부
cd tests/test-graph && npm start -- --include=artifacts_,session_states

# 탐색 캐시
cd tests/test-ttsc && npm start -- --include=capabilityresolutioncache

# Go
cd packages/ttsc && go test ./cmd/ttscgraph/ -run 'TestServe(Adopts|KeepsAStartupSet|ProjectsAnArtifact)' -count=1
node scripts/test-go-lint.cjs   # 10분 이상, 머신이 한가할 때만
```

**주의 (제가 겪은 함정):**
- `npm run build:current`는 `@ttsc/evidence`의 `lib/`를 `rimraf`합니다. **다른 스위트가 도는 중에 돌리지 마십시오** — evidence watch 케이스가 "모듈 없음"으로 거짓 실패합니다.
- 이 머신은 조용하지 않으면 타이밍 단언(`watch_refreshes_declared_markdown`, `lint` Go 타임아웃)이 거짓 실패합니다. 한가할 때 단독 재실행하면 통과합니다.
- `packages/ttsc` 소스를 건드리면 플러그인 빌드 캐시 키가 바뀌어 evidence e2e가 재링크(수 분)합니다.

---

## 8. 남은 일 (정확히 이것뿐)

1. **CI 그린 확인** — head `090a057d4`. 특히 `evidence defenses`(새 e2e + 새 lint 데몬 verb)와 `graph`, `ttsc core defenses`(새 캐시 케이스).
2. **squash 머지**, `#1239` 종료.
3. *(선택, 하면 좋음)* #1276/#1278 이후의 재발행 비용 **재측정** 후 PR 코멘트 갱신. 방법: `tests/test-evidence`에 임시 케이스를 만들어 `publishArtifacts`/`publishArtifactsResident`를 N회 호출하고 중앙값 출력 — 제가 쓰던 스크래치는 `1252134b2`에서 제거했으니 커밋 diff를 참고해 재현하면 됩니다. **반드시 커밋 전에 지우십시오**(단언이 없는 영구 그린 케이스가 됩니다).

## 9. 알려진 한계 (의도된 것, 이슈 아님)

- `unpublished()`가 감시하는 건 프로젝트의 leaf tsconfig와 `package.json`뿐입니다. `extends` 체인으로 플러그인을 상속받는 tsconfig는 추적하지 않습니다 — 체인 전체를 알려면 비용 그 자체인 로더를 불러야 하기 때문입니다.
- `resolveCapabilityPlugins`는 `pluginConfigDir` 오버라이드를 받지 않습니다. 컴파일러가 그걸 지정한 프로젝트에서는 설정 탐색 앵커가 갈릴 수 있습니다. 현재 호출자는 아무도 쓰지 않습니다.
- 데몬은 재발행마다 Program을 버립니다(§5-8). 소스 변경분만 골라 `changed`로 넘길 수 있게 되면 그때 더 줄일 수 있습니다.

samchon:
## Acceptance 16, third pass: the republish cost after #1276 and #1278

The two comments above measured the republish before either fix landed. This is the re-measurement they said was still owed, on the same fixture shape: a real project with `@ttsc/evidence` configured, a 40 section document, and 40 citations, plugin binary warm in the suite cache. Three independent runs, five samples per arm, medians reported with the raw samples beside them.

### `resolveCapabilityPlugins`: confirmed, and the one number that is unambiguous

| run | samples (ms) | median |
| --- | --- | ---: |
| A | 164, 167, 165, 186, 182 | **167** |
| B | 154, 140, 150, 144, 149 | **149** |
| C | 195, 214, 172, 157, 164 | **172** |

Against **6.6 s** before #1276. That is roughly a 40x change, it reproduces across three runs with a spread of under 15%, and it is far outside anything this host's noise can produce. The `zod@graph` and `nestjs@graph` numbers in the first comment moved the same way: 356 ms to 42 ms, and 701 ms to 60 ms.

`artifactsAreStale` stays what it was: 2.5, 2.8, 3.5 ms across the three runs, against 3.3 ms before. The per-request watch is still free.

### #1278: the mechanism is confirmed, the saving is not measurable here

The daemon genuinely serves. Asked directly, an open sidecar answers `project-inputs` with 5777 bytes and `graph-nodes` with 5354 bytes, never `null`, so the resident path is the one being taken rather than the fallback. That is now pinned by a case rather than by this comment: `test_ttscgraph_lint_daemon_answers_or_says_it_cannot` reads what the sidecar was spawned with and what reached it, because a client that never served would have left every existing case green.

What the daemon saves, this host cannot say. Resident and spawn-per-verb overlap completely:

| run | `publishArtifacts` spawn-per-verb (ms) | `publishArtifactsResident` (ms) |
| --- | --- | --- |
| A | 4636, 5395, 5671, 5662, 4850 | 4792, 8263, 6429, 5665, 5214 |
| B | 4055, 4406, 4450, 4337, 5259 | 4844, 5204, 5787, 5019, 5311 |
| C | 5299, 6860, 7246, 10584, 12639 | 5868, 5621, 9729, 4899, 4834 |

Run C's spawn arm spans 5.3 s to 12.6 s on identical work. The within-arm spread is larger than any difference between the arms, which is the same condition the first comment reported for the native dump, so no paired-delta trick rescues it either: both arms move together with whatever else the machine is doing. Reporting a median difference from this would be reporting the machine.

What can be said without timing: the total is down from **15.3 s** to roughly **4.5 s**, and the part of that attributable to #1276 (6.4 s of discovery) is measured above. The rest is not separable here.

### One observation, filed as samchon/ttsc#1279 rather than claimed

Polling the temp root while five `project-inputs` requests ran through a single open daemon showed **18 distinct `ttsc-lint-config-*` loader directories**, about 3.6 config evaluations per request. If the rule memo were holding across the daemon's life there would be one. The memo mechanism itself is sound and unit-tested (`TestResidentRulesAreReusedOnlyWhileTheirConfigIsUnchanged` fails if it stops recording, or records a state it cannot prove), so this is about whether it engages on this fixture rather than about whether it works.

The likely suspect is not the memo but the layer under it: `loadCachedConfigEvaluationWithPolicy` already content-caches evaluations, and this fixture's `node_modules` are junctions into this repository, which is exactly the arrangement the first comment already found unkind to path-sensitive work. Isolating that means instrumenting the sidecar and rebuilding the plugin per iteration, and it is a cost question rather than a correctness one, so it is filed as samchon/ttsc#1279 rather than guessed at in this branch.

### Method

`publishArtifacts`, `publishArtifactsResident`, and `resolveCapabilityPlugins` timed in-process against the built `@ttsc/graph` and `ttsc`, one discarded warm-up, five samples each, medians reported. The daemon probe drove `TtscLintDaemon` directly against the same sidecar the session opens. The measurement case was temporary and is not in the branch.

PR #1280, opened 2026-08-20
fix(ttsx): rescue a JavaScript specifier on the CommonJS graph too

samchon:
`@ttsc/banner` cannot load a TypeScript config that imports a sibling by its JavaScript spelling on Node 22, and the utility plugins under-report what such a config consulted. Both are one gap in ttsx's CommonJS reach, found while running the `@ttsc/test-unplugin` suite on the oldest Node the package supports.

## What was wrong

TypeScript asks authors to write the emitted `.js` extension in a relative specifier. ttsx's ESM `resolve` hook has always mapped that spelling back to the source it names; the CommonJS graph never did.

`Module._extensions` decides how a file that was **already found** gets compiled, and Node reuses its keys only while probing an *extensionless* request. So `require("./x")` reached `x.ts` for free, and `require("./x.js")` backed only by `x.ts` failed to resolve before any compiler was consulted. The installer's own comment named `_extensions` as the CommonJS half of the pair — true of compilation, not of resolution.

That graph is reached by every `require()` inside a CommonJS module the ESM loader evaluated, which is exactly how a plugin loads a config file it discovered:

| runtime | `require()` inside a translator-evaluated CommonJS module |
| --- | --- |
| Node 22 | bypasses `module.registerHooks` — the hook sees the `import()` and nothing within it |
| Node 24 | routed through the hooks |

`engines.node` declares `>=22.15.0`, and every CI lane but one runs the current Node, so the floor went unchecked and this fell through it.

## What changed

**`packages/ttsc` — resolution.** The rescue is installed on `Module._resolveFilename` as well. It answers only a request Node's resolver already refused, and rethrows Node's original error untouched when no source matches, so a genuinely missing module keeps its `MODULE_NOT_FOUND` and `requireStack`. `require.resolve` shares the entry point and is fixed with it.

**`@ttsc/banner`, `@ttsc/strip` — input records.** Their loaders observe resolutions through the hook to report what a config consulted *plus the candidates that decide it*. On Node 22 a config's CommonJS requires were reported without those candidates, so a spelling appearing later could change what the config resolves to with nothing in the envelope to notice — the shape #1271 was filed for. Both now wrap the CommonJS resolver with the same two observations their hook makes, outside ttsx's own wrapper.

**`@ttsc/lint` is deliberately untouched.** The resolution half above already reaches it, since it belongs to ttsx. Its loader builds a config *graph* through a differently shaped hook; whether its own edge recording has the same gap is not established here, and changing it untested would be worse than leaving it. Worth its own investigation.

## Not widening CI

No new lane and no new OS. The `ttsx node 22.15` lane already exists as the only check of the declared floor; it gains this one case, because a case that passes on the current Node whatever ttsx does can only be judged there. The same case also runs on the default-Node lanes, so both runtimes are pinned.

## Verification

| lane | result |
| --- | --- |
| new case on the parent commit | fails with `Cannot find module './target.js'` |
| new case here | pass |
| `features/ttsx-runtime` (107 cases) | pass, except two pre-existing environment failures below |
| banner / paths / strip Go suites | pass |
| `test_transformttsc_persistent_utility_config_dependencies_invalidate_the_generation` | pass — this is what the branch repairs |
| typecheck, prettier, the Go formatter | clean |

`test_ttsx_executes_javascript_emitted_by_the_consumer_local_tsgo` and `test_ttsx_virtual_layout_mirrors_a_file_symlink_project_entry` fail on this Windows checkout for environment reasons — a stub executable that will not spawn, and a file symlink that needs elevation. Both were confirmed to fail identically with this branch's changes stashed, so neither is this work.

### DIFF
--- a/packages/banner/driver/banner.go
+++ b/packages/banner/driver/banner.go
@@ -448,7 +448,8 @@ func loadBannerScriptConfigFile(location string) (any, error) {
 
 func loadBannerScriptConfigFileWithInputs(location string) (bannerLoadedConfig, error) {
   const script = `
-const { createRequire, isBuiltin, registerHooks } = require("node:module");
+const nodeModule = require("node:module");
+const { createRequire, isBuiltin, registerHooks } = nodeModule;
 const crypto = require("node:crypto");
 const fs = require("node:fs");
 const path = require("node:path");
@@ -644,6 +645,27 @@ registerHooks({
   },
 });
 
+// The hook above never sees a require() made from inside a CommonJS module the
+// ESM loader evaluated, which on Node 22 is every require the config makes:
+// module.registerHooks observes the import() of that module and nothing within
+// it. A config's own dependencies would then be reported without the candidates
+// that decide them, so a spelling appearing later could change what the config
+// resolves to with nothing in the envelope to notice it (samchon/ttsc#1280).
+// Wrapping the CommonJS resolver records the same two observations the hook
+// does, on the graph the hook cannot reach.
+const nextResolveFilename = nodeModule._resolveFilename;
+nodeModule._resolveFilename = function resolveFilename(request, parent, isMain, options) {
+  const parentFile = parent && typeof parent.filename === "string" ? parent.filename : undefined;
+  const parentURL = parentFile === undefined ? undefined : pathToFileURL(parentFile).href;
+  recordResolutionCandidates(request, parentURL, undefined);
+  const resolved = nextResolveFilename.call(this, request, parent, isMain, options);
+  if (path.isAbsolute(resolved)) {
+    recordResolutionCandidates(request, parentURL, pathToFileURL(resolved).href);
+    recordFile(resolved);
+  }
+  return resolved;
+};
+
 (async () => {
   const mod = await import(pathToFileURL(process.argv[1]).href);
   let current = Object.prototype.hasOwnProperty.call(mod, "default") ? mod.default : mod;
@@ -823,11 +845,11 @@ func loadBannerTypeScriptConfigFileWithInputs(location, resolutionRoot string) (
 // JSON-encoded import specifier) and writes the serialized banner value to stdout.
 func bannerTypeScriptConfigLoaderSource(importLiteral string) string {
   return fmt.Sprintf(`// @ts-nocheck
-import { createRequire, isBuiltin, registerHooks } from "node:module";
+import Module, { createRequire, isBuiltin, registerHooks } from "node:module";
 import crypto from "node:crypto";
 import fs from "node:fs";
 import path from "node:path";
-import { fileURLToPath } from "node:url";
+import { fileURLToPath, pathToFileURL } from "node:url";
 
 const inputs = new Set<string>();
 const hashes = new Map<string, string | null>();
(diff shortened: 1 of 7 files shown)


samchon (commented):
# Overall Self-Review

Two complete rounds over the whole  diff. Round 1 produced the three findings below and they were applied together in ; round 2 was a fresh pass over the result and produced nothing, so that is the clean round.

## Findings applied

1. **Both new wrappers read a request Node had not vouched for.**  is an internal entry point anything may call, so a non-string request arrives as readily as a specifier does.  in the ttsx rescue — and the candidate recorder in the loaders — would then have thrown a  out of this code in place of the argument error Node was about to raise, which is a worse failure than the one being fixed and reaches a caller that did nothing wrong. All five wrappers now decline such a request and leave it to Node.

2. **The rescue answered  from the wrong base.** That form resolves against the caller's list; the rescue reads the parent's directory and never consults , so a candidate sitting beside the parent would have been a different question's answer — and a wrong resolution is worse than the refusal it replaced. It now keeps Node's own result, and the docstring states the exception rather than claiming  is rescued outright.

3. **The execute page named only the ESM graph's error.** It promised  for a genuinely missing module, which is right for the graph the page described and wrong for the CommonJS graph this change makes the hooks reach, where Node raises  with a require stack. Both are named now.

## Cleared by these rounds

That the rescue never perturbs a resolution Node already answered, since it runs only from the ; that a request no candidate satisfies rethrows Node's original error object untouched, with its code and , which the three  twins pin directly; that a bare specifier is declined, so package resolution is untouched; that the loader wrappers sit outside ttsx's own, so a rescued path is recorded rather than hidden; that the loaders skip a non-absolute result, keeping builtins out of the input records; that  guards installation to once per process across all three entry points that call it; and that the case added here is matched by the  filter the Node 22 lane uses.

## Recorded, not changed

 carries a differently shaped loader that builds a config graph. The resolution half of this fix already reaches it, since that half belongs to ttsx. Whether its own edge recording has the same Node 22 gap is not established here, and changing it without a reproduction would be worse than leaving it; it is worth its own investigation.

## Verification

| lane | result |
| --- | --- |
| new case on the parent commit | fails with  |
| new case here | pass |
|  twins (3) | pass |
|  | pass, except two pre-existing environment failures |
| banner / paths / strip Go suites | pass |
| ,  | pass |
| , full suite | 149 cases pass |
|  | pass — the failure this branch was opened for |
| typecheck, prettier, the Go formatter | clean |

 and  fail on this Windows checkout for environment reasons — a stub executable that will not spawn, and a file symlink that needs elevation. Both were confirmed to fail identically with this branch stashed, so neither is this work.

samchon (commented):
# Overall Self-Review

Two complete rounds over the whole `origin/master...HEAD` diff. Round 1 produced the three findings below and they were applied together in `196310642`; round 2 was a fresh pass over the result and produced nothing, so that is the clean round.

## Findings applied

1. **Both new wrappers read a request Node had not vouched for.** `Module._resolveFilename` is an internal entry point anything may call, so a non-string request arrives as readily as a specifier does. `path.isAbsolute` in the ttsx rescue — and the candidate recorder in the loaders — would then have thrown a `TypeError` out of this code in place of the argument error Node was about to raise, which is a worse failure than the one being fixed and reaches a caller that did nothing wrong. All five wrappers now decline such a request and leave it to Node.

2. **The rescue answered `require.resolve(request, { paths })` from the wrong base.** That form resolves against the caller's list; the rescue reads the parent's directory and never consults `paths`, so a candidate sitting beside the parent would have been a different question's answer — and a wrong resolution is worse than the refusal it replaced. It now keeps Node's own result, and the docstring states the exception rather than claiming `require.resolve` is rescued outright.

3. **The execute page named only the ESM graph's error.** It promised `ERR_MODULE_NOT_FOUND` for a genuinely missing module, which is right for the graph the page described and wrong for the CommonJS graph this change makes the hooks reach, where Node raises `MODULE_NOT_FOUND` with a require stack. Both are named now.

## Cleared by these rounds

That the rescue never perturbs a resolution Node already answered, since it runs only from the `catch`; that a request no candidate satisfies rethrows Node's original error object untouched, with its code and `requireStack`, which the three `preserves_module_not_found` twins pin directly; that a bare specifier is declined, so package resolution is untouched; that the loader wrappers sit outside ttsx's own, so a rescued path is recorded rather than hidden; that the loaders skip a non-absolute result, keeping builtins out of the input records; that `installed` guards installation to once per process across all three entry points that call it; and that the case added here is matched by the `--include` filter the Node 22 lane uses.

## Recorded, not changed

`@ttsc/lint` carries a differently shaped loader that builds a config graph. The resolution half of this fix already reaches it, since that half belongs to ttsx. Whether its own edge recording has the same Node 22 gap is not established here, and changing it without a reproduction would be worse than leaving it; it is worth its own investigation.

## Verification

| lane | result |
| --- | --- |
| new case on the parent commit | fails with `Cannot find module './target.js'` |
| new case here | pass |
| `preserves_module_not_found` twins (3) | pass |
| `features/ttsx-runtime` | pass, except two pre-existing environment failures |
| banner / paths / strip Go suites | pass |
| `@ttsc/test-banner`, `@ttsc/test-strip` | pass |
| `@ttsc/test-unplugin`, full suite | 149 cases pass |
| `test_transformttsc_persistent_utility_config_dependencies_invalidate_the_generation` | pass — the failure this branch was opened for |
| typecheck, prettier, the Go formatter | clean |

`test_ttsx_executes_javascript_emitted_by_the_consumer_local_tsgo` and `test_ttsx_virtual_layout_mirrors_a_file_symlink_project_entry` fail on this Windows checkout for environment reasons — a stub executable that will not spawn, and a file symlink that needs elevation. Both were confirmed to fail identically with this branch stashed, so neither is this work.

_(Replaces an earlier posting of this review whose code spans were lost to a shell quoting mistake.)_

Issue #1310, opened 2026-08-29
withTtsc silently applies no transform on Turbopack, the default bundler in the Next version this repo pins


samchon:
## Problem

`withTtsc` (`packages/unplugin/src/next.ts`) wires ttsc into a Next.js project by injecting the webpack plugin and nothing else:

```ts
return {
  ...nextConfig,
  webpack(config, webpackOptions) {
    config.plugins = Array.isArray(config.plugins) ? config.plugins : [];
    config.plugins.unshift(webpack(options));
    ...
  },
};
```

Next calls that `webpack` hook only for webpack builds. On Turbopack the hook is never invoked, so no ttsc transform runs, and nothing reports that. The build succeeds and the output is simply untransformed.

That is the worst available failure mode for this package. A ttsc plugin left unapplied does not leave the source unchanged in any useful sense: typia's `assert<T>()`, nestia's generated routes and every other plugin driven construct depend on the transform having happened. The code compiles, ships, and fails at runtime instead of at build time.

Turbopack is the default bundler in current Next majors, including the 16.3 line this repository itself pins (`website/package.json`, `experimental/test-unplugin/.tmp/project/package.json`). So the wrapper covers the non default path for the very version the project builds against, and the default path silently gets nothing.

## Evidence

`packages/unplugin/src/next.ts` is 57 lines and contains no `turbopack` key. The returned object carries only the spread `nextConfig` and the replaced `webpack` hook.

`packages/unplugin/README.md:121` documents the gap:

> `withTtsc` covers Next.js builds that go through webpack. When Next.js runs on Turbopack (`next dev --turbopack`), wire the Turbopack loader rules below instead.

So the behaviour is known and intentional as far as it goes. Two things make it insufficient rather than merely documented:

1. The remedy the README points at is a hand written `turbopack.rules` block that the user must keep in sync with the wrapper's options by hand. The rule glob and the `options` object both duplicate what `withTtsc(nextConfig, options)` already received.
2. The documented example still reads as if the wrapper is the Next.js integration, and the sentence about Turbopack is one line below it. A reader who copies the wrapper and runs `next dev` on Next 16 gets no transform, no warning, and no failure until runtime.

`@ttsc/unplugin/turbopack` already exists and is a standalone webpack loader that Turbopack accepts through `turbopack.rules`, so the capability is present. Only the wiring is missing from the wrapper.

[...]

Issue #1309, opened 2026-08-29
Four documented-behaviour gaps in @ttsc/unplugin: three false claims and two undocumented entrypoints


samchon:
## Problem

Four things `@ttsc/unplugin` says about itself do not match the code, verified against `ac6d91c63`. Three are claims that are false; the fourth is a public surface the documentation never mentions. They share one cause: each was written from intent and never re-read against what shipped.

**1. The cache section describes the opposite of what happens to an out of root module.**

`packages/unplugin/README.md:315` and `website/src/content/docs/setup/unplugin.mdx:246` both end their cache paragraph with:

> A compile that succeeded but has no output for one requested file is a fact about that file, so it is reported and the generation is left alone. That holds for a file the project walk covers; one outside the project root has no recorded state to compare against at all, so it still replaces the generation.

The second sentence is false, and it contradicts the same paragraph's opening sentence, which lists "monorepo sibling sources" among the out of walk inputs the snapshot records. A file the program reaches by import is a program source file wherever it lives, so it is in `sourceHashes` and validates like any other input.

The sentence conflates "outside the project walk" with "outside the program". Only the second would have no recorded state, and the second is not what it describes.

**2. `isTransformTarget`'s docstring contradicts the filter it documents.**

In `packages/unplugin/src/core/index.ts`:

```ts
/**
 * Returns `true` when the module id refers to a real TypeScript/JavaScript
 * source file that should be processed by the ttsc transform.
 */
```

Twenty lines above, `sourceFilePattern`'s own docstring states the opposite, correctly:

```ts
 * JavaScript is deliberately not among them, so a `.js` module reaches no
 * adapter's transform.
```

`sourceFilePattern` is `/\.[cm]?tsx?$/`, so the second is right and the first is wrong. A contributor reading only the exported function's docstring would conclude `.js` is in scope.

**3. Alias forwarding is Vite only, and documented nowhere.**

`transformTtsc` forwards host aliases into the generated tsconfig overlay through `createAliasPaths(aliases)`. `aliases` is assigned in exactly one place, `vite.configResolved`, so it is `undefined` under Rollup, Rolldown, webpack, Rspack, esbuild, Farm, Turbopack and Bun. A Vite user's `resolve.alias` reaches the compile; a webpack user's does not.

Neither the README nor the website page mentions alias forwarding at all, in either direction. A reader cannot learn that the feature exists, that it is Vite only, or that on other hosts the tsconfig's own `paths` is the only channel.

[...]

Issue #1307, opened 2026-08-29
The project walk's hardcoded ignore list both misses real output directories and hides real source directories


samchon:
## Problem

`@ttsc/unplugin` decides which directories the project walk ignores from a hardcoded list of fifteen bare names in `isIgnoredProjectDirectory` (`packages/unplugin/src/core/transform.ts`):

```
.git .ttsc .cache .next .nuxt .svelte-kit .turbo .vite
build coverage dist node_modules out temp tmp
```

The list is matched against every entry name at every depth, and it is the only thing standing between the project walk and the filesystem. That produces two failures at once, in opposite directions.

**Too little is ignored.** Every directory not on the list is walked, and every entry it contains takes part in the project membership digest. A bundler writing output anywhere else changes project membership with its own output, which replaces the generation and costs a whole project compile. samchon/ttsc#1300 was this exact failure for `out/`, fixed by making the membership digest honour the list. The list is the part that was never general. With content-hashed output filenames the cost is a full recompile on **every** rebuild. Nuxt (`.output`), Astro (`.astro`) and every custom `outDir` such as `lib/` or `output/` are uncovered, and so is Expo's `.expo/`.

**Too much is ignored.** Because the match is on a bare name at any depth, a legitimate source directory whose name collides is dropped from the walk entirely. A new program input created in `src/build/` is never detected, and the adapter keeps serving output from a compile that never saw the file. That is staleness, a correctness defect rather than a cost.

Both follow from one cause: the exclusion policy is a guess, in a place where authoritative configuration is already in hand.

## Evidence

All measurements against `ac6d91c63` on the real native envelope fixture (`tests/test-unplugin/src/internal/real-native-envelope.ts`), which drives a real Go plugin host and appends one byte per `ApplyProgram` call, so the run log length is an exact count of whole project compiles. Each pass opens with `beginTtscTransformBuild` and delivers the fixture's four modules.

### Too little is ignored

Content-hashed output filenames, the old file removed and a new one written before each of four rebuilds:

| Output directory | Compiles |
| --- | --- |
| `dist/` (listed) | 1 total |
| `lib/` (not listed) | **4, one per rebuild** |

Rewriting one output file with changed content instead: `dist/` 1 total, `lib/` 1 on first appearance then 0. First appearance alone, measured separately: `.output/` 1, `lib/` 1, `.astro/` 1, while `dist/` and `out/` cost 0.

[...]

Issue #1312, opened 2026-08-29
A failed compile registers no watch inputs and reports colour-escaped host output


samchon:
## Problem

When a compile fails, `@ttsc/unplugin` does two things that leave the host worse off than it needs to be. Both happen on the same delivery, and both are specific to the failure path.

**1. It registers no watch inputs, so the host cannot learn that the failure was fixed.**

A successful delivery registers the generation's derived watch inputs through `addWatchFile`, which is how a type-only file that no bundler graph contains still invalidates the modules that depend on it. A failed delivery registers nothing at all: `selectWatchInputs` returns `[]` for an `"exception"` result, and in any case `selectOrEvict` throws before `notifyWatchInputs` is ever reached.

The consequence lands on the first compile of a watching session. If that compile fails because of a type error in a file the bundler's own module graph does not contain, nothing about that file is watched. The user fixes it, the bundler sees no change to anything it knows about, no module is invalidated, and the error stays on screen until they touch a file the bundler does track.

**2. It surfaces the host's raw, colour-escaped output instead of the structured diagnostics it already knows how to format.**

`selectTransformedSource` formats a `"failure"` envelope through `formatDiagnostics`, which builds a clean `file: line:character: message` string from the structured `ITtscCompilerDiagnostic` fields. An `"exception"` envelope instead goes through `formatUnknownError`, which returns the host error's message verbatim.

An ordinary type error arrives as `"exception"`, not `"failure"`. So the common failure takes the path that does no formatting, and the ANSI escapes the compiler emitted travel into the `Error` the bundler reports. In a terminal that renders; in a Vite error overlay, a webpack error report or a CI annotation, the user reads `[96msrc/types.ts[0m:[93m1[0m:[93m22[0m` instead of a file and a line.

## Evidence

Measured against `ac6d91c63` on the real native envelope fixture (`tests/test-unplugin/src/internal/real-native-envelope.ts`). One module delivered on a fresh cache and a fresh pass, counting `addWatchFile` calls, with a genuine type error planted in `src/types.ts`, a type-only file nothing imports at runtime:

| Project state | Watch inputs registered | Delivery |
| --- | --- | --- |
| healthy | **15** | ok |
| type error in `src/types.ts` | **0** | throws |
| after the fix | **15** | ok |

The same run captured the thrown message, whose first line is:

```
\u001b[96msrc/types.ts\u001b[0m:\u001b[93m1\u001b[0m:\u001b[93m22\u001b[0m - ...
```

[...]

PR #1313, opened 2026-08-29
fix(unplugin): make the adapter's belief about program membership correct

samchon:
Cycle 2 of the `@ttsc/unplugin` campaign, carrying the complete accepted set. Six complete fresh full-scope discovery rounds ran against the integrated baseline `ac6d91c63`; the sixth produced no candidate, which froze the set below.

One theme runs through it. Cycle 1 taught the adapter to keep a generation across a build boundary. This cycle is about **what the adapter believes the program contains**, and every issue is a place where that belief was wrong, unstated, or inconsistent between this repository's own packages.

## What landed

| Issue | Resolution |
| --- | --- |
| #1307 | Program membership comes from the resolved configuration, not from fifteen hardcoded directory names |
| #1308 | A module the program does not contain is passed through and reported, one answer for every adapter and for `@ttsc/metro` |
| #1309 | Four documented behaviours the code did not have |
| #1310 | `withTtsc` wires Turbopack as well as webpack |
| #1311 | The 1,751-word cache paragraph split by idea, architecture moved to the guide |
| #1312 | A failed compile registers the inputs a fix would touch, and reports plain text |

## #1307 in detail, since it is the bulk of the diff

The walk answered two questions from one name list and was wrong in both directions. Too little was ignored: a bundler writing output to any directory the list did not name changed project membership with its own output. Too much was ignored: the list matched a bare name at any depth, so a real source directory called `src/build/` was dropped from the walk and a program input created there was **never detected**.

Measured on the real native-envelope fixture, which counts whole-project compiles through a real Go plugin host:

| | before | after |
| --- | --- | --- |
| Content-hashed output in an unlisted directory, 4 rebuilds | **4 compiles** | 1 |
| 200 emitted files, pass that first sees them | **652 reads** | baseline, 7 |
| New `src/build/b.ts` covered by `include` | **never detected** | detected |
| Persistent hosts (Metro, Turbopack, Vite dev) | full cost per rebuild | 1 on first appearance, then free |

`allowJs` and `resolveJsonModule` decide which extensions can enter the program; `outDir`, `declarationDir` and plain `exclude` entries name the directories it does not contain; a directory takes part only while its subtree can hold a program input. Only `.git`, `node_modules` and `.ttsc` remain name-based, because no tsconfig can name them.

## Self-review

Four Individual Self-Reviews and two Overall rounds found **nine defects in this cycle's own work**, all fixed here and all recorded as formal reviews on this pull request.

The one worth naming: the walk became configuration-aware while `isProjectWalkPath` stayed permissive. `selectExternalInputPaths` uses that predicate as the sole test for "the walk already covers this", so a graph input under an excluded directory would have landed in **neither** snapshot. Every behavioural gate would have stayed green, because each function was individually correct; only asking whether the two still answered the same question exposed it.

The riskiest consequence of the walk narrowing is now proven rather than argued: editing `tsconfig.json`, `package.json` or a plugin descriptor still costs exactly one recompile, because those are universal host inputs rather than walk inputs.

## Coverage

Twelve new cases and three changed. The three changes are contract changes rather than repairs: a `.txt` removal is no longer a membership change, which is what its own sibling already asserted for an edit to the same file; the out-of-program delivery returns rather than throws; and `@ttsc/metro` no longer tells a real failure from a pass-through by matching error text, because there is no text left to match.

The install experiment now builds Next on both bundlers instead of forcing `--webpack`, which is the only thing that proves Turbopack actually runs the loader.


samchon (commented):
## Individual Self-Review — `f513138c8` (`fix(unplugin): decide program membership from configuration, not from names`)

Read-only advisory review over the parent-to-commit diff. Adjudication is mine. It resolved #1307, #1308 and #1312, and it carried four defects, every one of them a consequence of narrowing one half of a pair and not the other.

### Accepted, and all four are the same shape

**The walk became configuration-aware while `isProjectWalkPath` stayed permissive.** `selectExternalInputPaths` uses that predicate as the sole test for "the walk already covers this", and records everything else as an out-of-walk input proven by content and physical identity. So a graph input under an excluded directory would have landed in neither snapshot: absent from `inputHashes` because the walk skipped its directory, and absent from the out-of-walk snapshot because the predicate claimed the walk had it. On a pass-based host that is silent staleness, the exact class this issue calls a correctness defect rather than a cost. On a persistent host it is worse: a whole-project recompile per delivery, forever, because the narrow proof can never find a recorded hash for it.

**The live mutation tracker was never narrowed at all.** Every new case opens a pass, and a pass routes its first delivery through the whole-generation proof, so none of them touch the tracker. The tracker is the primary path exactly where a pass boundary is absent, which is `@ttsc/metro`, the Turbopack loader and a watching Vite dev server. Two consequences: the cost this issue exists to remove was still being paid in full on those hosts, and the twelve names the old list used to skip were now walked and watched, so a dev server writing into its own cache directory voided a generation where it previously did not.

**The walk still hashed every file it reached.** The extension policy narrowed the membership digest and nothing else, so `collectProjectInputHashes` (which takes no declared-key restriction, because a fingerprint has no generation to compare against) still read every emitted file. That leaves this issue's Metro criterion unmet wherever configuration does not name the output directory, which for an Expo project extending a base config is always.

**Capture read strictly more than before.** The same cause: both snapshots in `captureTransformGeneration` pass a policy and no restriction, so every compile now hashed trees the old name list had skipped.

All four landed in `c4602eec7`, together with the finding that the Windows broker records mutations on a separate path from the in-process watcher, which is why the first attempt at the tracker fix measured as no change on this machine.

### Adjudicated, not accepted

The reviewer flagged `readProjectMembershipPolicy` returning the narrow policy when a config cannot be read, against a `PERMISSIVE_*` constant the module documents as the deliberate fallback. Kept as is: an unreadable tsconfig is a config error the compiler itself reports, and the compile that follows fails rather than producing a generation to validate. Recorded rather than dismissed.

Overlaying `outDir` appends rather than replaces, so a caller's overlay does not un-exclude the project's own. Over-exclusion, which is the direction that loses correctness, so it is worth stating: it cannot un-exclude, only add. Left as is because an overlay that excluded *less* than the project config would be the more surprising rule.

The `.json` rule hides creation of a nested `package.json` or `tsconfig.json` under `resolveJsonModule: false`. Real, and narrower than it looks: those are host inputs rather than program inputs, validated by the universal descriptor path rather than by membership. Recorded.

### Clean, and worth stating

The `declaredKeys` restriction is sound: only the validating caller passes it, the same set drives `sameHashes` and `walkSnapshotComplete` there, and both the capture path and `failedGenerationEnvironmentChanged` pass none. `membershipPolicy` has one construction site and no deserialization path, and it is anchored at the project's own tsconfig rather than the generated one, which is what makes a relative `outDir` resolve correctly. The `@ttsc/metro` string-match removal is complete, with no live path still matching on message text. `TtscMissingProgramOutputError` routing is correct in both `retainPassVerdict` and `selectOrEvict`, so the generation is neither retained as a verdict nor evicted. Relevance propagation is correct in both directions: a source appearing in a previously irrelevant directory propagates to the root, and the last source leaving it does too.

samchon (commented):
## Individual Self-Review — `6006088e2` (`fix(unplugin): wire Turbopack from the Next wrapper, not only webpack`)

Read-only advisory review over the parent-to-commit diff, verified against the Next 16.3 sources this repository pins. Adjudication is mine. It resolved #1310 and carried two defects, one of which re-created this cycle's own failure mode one level up.

### Accepted, and the important one

**Always defining `turbopack` permanently disables Next's own guard.** Next refuses to build on Turbopack when a config carries a `webpack` hook and no `turbopack` block, because the hook is then silently ignored: `hasTurboConfig = Boolean(rawNextConfig.turbopack)`, and with `TURBOPACK='auto'` (which is what no bundler flag means) a webpack config without a Turbopack one exits the build with an error naming the problem. This wrapper always defines both, so that check can never fire again for anyone who uses it, and a caller's own webpack customisation would be dropped on a Turbopack build with nothing said at all.

That is the same class as the issue itself: a silent loss of configuration the user wrote. Wiring ttsc for Turbopack is worth exactly one warning, not the loss of the warning Next already gave, so the wrapper now issues it for precisely the callers Next would have stopped, and stays quiet for a caller who configured Turbopack themselves or who has no webpack hook to lose.

It also corrects the issue's own premise for this version. On Next 16.3 with no bundler flag, the pre-fix config hit that guard and exited, so it was a loud failure rather than silence; silence required an explicit `--turbopack` or an older Next. The fix is right either way, and the framing in the issue and in the earlier commit message was not.

**An array-form rule was spread into an object.** Turbopack accepts a bare array of loaders, and the merge produced `{ "0": "other-loader", loaders: [...] }`, which Next's own strict schema rejects as an unrecognized key. Corrected, and now pinned by a case asserting no numeric keys survive.

### Accepted, and settled from the loader contract

**The loader was placed first in the array, which runs it last.** Turbopack runs a rule's loaders through webpack's `loader-runner`, whose normal phase runs right to left, so the last entry is the one that sees the original source. ttsc transforms TypeScript into TypeScript, so it has to be that one. Appended now, and the case that asserted the opposite ordering was asserting the wrong thing.

### Adjudicated, not accepted

Dedupe is per exact glob, so a caller who hand-wired `"*.{ts,tsx}"` escapes it and ends up with the file matched by two rules. Real, and left: normalising globs means implementing Turbopack's matcher, and the README states the exact globs the wrapper writes.

No `next` peer range is declared, so on a Next older than 15.3 the wrapper injects a root `turbopack` key that version does not know. Out of this cycle's scope; the package declares no Next dependency at all today, and adding one is a packaging decision rather than a fix to this issue.

`.mts` and `.cts` are transformed by the webpack half and not by the Turbopack globs, since Next does not resolve those extensions by default. Recorded.

### Coverage, and what it changed

The reviewer's sharpest point was that both new cases assert only the shape of a returned object, so a wrong loader specifier or glob would pass all of them, while the repository's only end-to-end Next check runs `next build --webpack` and therefore never exercises the path this issue is about. That check now builds on both bundlers, which is what #1310's acceptance asked for and the only thing that proves Turbopack actually runs the loader.

samchon (commented):
## Individual Self-Review — `5469fc8a3` (`docs(unplugin): split the cache section and correct what it claims`)

Read-only advisory review over the parent-to-commit diff, with every new sentence checked against the code at `HEAD` rather than against the prose it replaced. Adjudication is mine. It resolved #1309 and #1311 and carried five defects.

### Accepted

**A commit that corrects false claims introduced one.** The `isTransformTarget` docstring said `sourceFilePattern` sits "twenty lines above" when it is nearly three hundred. The phrase was copied from #1309's own body without measuring, and `{@link sourceFilePattern}` already locates the symbol, so it carried no information even when it was true. Removed, and the issue body says so now.

**A dangling antecedent, created by the split.** "Transform source outputs outside the project walk are kept in that external snapshot too" survived in the README while the sentence that introduced the external snapshot moved to the guide. It also sat after the paragraph that hands the reader to the guide, and it duplicated the guide byte for byte, which is what #1311 asks to remove. Deleted.

**A cross-reference pointing the wrong way.** The Bun runtime entry was described as "covered under Bun below" from a section above it.

**The output-directory promise ignored `allowJs`.** "Costs nothing whatever it is called" is false whenever the project admits JavaScript and no configuration names the directory: the extension rule then admits the emitted `.js`, the directory becomes relevant, and its churn replaces the generation exactly as before. The qualifier that made the sentence true was attached to the previous sentence only. Both documents now carry it, and say what such a project should do instead.

**"A source appearing in it later is still detected" was false for the excluded directories named one sentence earlier.** A directory named by `outDir`, `declarationDir` or a plain `exclude` entry is never pushed onto the walk stack, so it is neither walked nor watched. The claim holds only for the relevance path, and both documents now separate the two.

### Accepted, structural

The README still opened its cache section with three architecture paragraphs duplicating the guide near-verbatim, so "each retained claim is present in exactly one of the two documents" was not achieved. They are replaced by one user-facing paragraph and the link.

### Adjudicated, not accepted

Vite's RegExp alias form is not forwarded, since `createAliasPaths` takes string prefixes only. Real and undocumented, but it is a fact about the alias overlay rather than about this section, and the alias paragraph is new in this commit. Recorded for the next cycle rather than expanded here.

The membership paragraph sits under "when the generation is reused" while a membership event is what discards one. Fair, and left: the reader arrives asking why a rebuild recompiled, and membership belongs beside the answer.

### Clean, and worth stating

Nothing of substance was lost in the split. All seventy-six clause units of the old paragraph trace to a destination, the one deliberate deletion is the false out-of-root sentence this cycle's own previous merge shipped, and the qualifier losses are the guide keeping a distinction the README does not need. Both files carry zero em-dashes, every paragraph is one source line with blank lines between ideas, heading levels nest correctly, the `#bun` anchor resolves, and the guide link matches the MDX path.

samchon (commented):
## Individual Self-Review — `e9f0e69de` (`fix(unplugin): let the adapter probe a failed generation's watch inputs`)

Read-only advisory review over the parent-to-commit diff. Adjudication is mine. The code change stands; the reason recorded beside it did not, and a comment that states a reason the code does not have is worth correcting even when the code is right.

### Accepted

**The stated failure mode cannot occur on this path.** Both call sites sit in `catch` blocks that immediately rethrow, and the Vite adapter's `transform` returns the rejection uncaught, so the request aborts before `vite:import-analysis` runs. That plugin is itself a transform hook, appended last, and transform-phase additions to `_addedImports` are discarded rather than persisted. So `missing: false` could never have produced the 500 the comment described: the request already fails on the compile error, which is the intended outcome.

The change is right for two reasons the comment did not give, and now does:

The claim was unbacked. A failed generation is replayed for the rest of its pass without re-proving its inputs, so the walk that recorded them may be older than the delivery, and one of those inputs having been deleted is a live reason for that compile to have failed. `missing: false` asserted something the path had not established.

And it buys the only recovery channel there is. Routing an absent input to the missing-input poll is what makes restoring a deleted type-only file invalidate anything at all: a bundler watch on a path that does not exist registers nothing, and no module graph carries a type-only input. That is exactly #1312's premise, applied to the deletion case.

### Adjudicated, not accepted

The cost qualifier was loose: the probe is paid only where the adapter reads evidence, which is Vite serve, and there it is one walk per throwing delivery rather than one per session. Proportionate, since the session is already blocked, and the comment now scopes it.

Dropping `identity` means an absent input's poller keys off a default host context rather than the generation's. Identical in production; they could differ only under a virtual filesystem, where the worst case is two pollers for one path and both fire.

### Coverage, and what it changed

The reviewer's decisive point: nothing in the suite asserted the evidence argument at either supplier, so this commit would have passed identically if reverted, as would every other case. The failure scenario now captures what is supplied alongside what is registered, and asserts that a healthy delivery supplies the generation's own recorded evidence while a failed one supplies none. That fails on revert, which is the property a regression guard needs and this one previously lacked.

samchon (commented):
## Overall Self-Review — round 2, over the whole base-to-head diff at `bd22a4944`

Round 1 produced one finding (`e9f0e69de`). This round is fresh and complete over all 32 files, and it produced one more, which is now `bd22a4944`.

### The finding

**Nothing asserted that the compiler's non-source inputs survive the walk narrowing.** The walk now collects only files that could enter the program, which is what removes the read-per-file cost, and the whole safety argument for that rests on the tsconfig, the package manifest and the plugin descriptor being proven by the universal host-input path rather than by the walk. That was an argument, not a test, and the failure it would have hidden is the worst this cycle could produce: a tsconfig edit silently invalidating nothing.

Measured, and the argument holds. Each of the three edits still costs exactly one compile. It is a case now.

### Decision points re-verified in this round

- **The walk and its predicate answer the same question.** `selectExternalInputPaths` treats `isProjectWalkPath` as the sole test for "already covered", so any divergence puts an input in neither snapshot. Both now take the policy, and the agreement is asserted directly rather than through a scenario, because the defect lives between two functions rather than inside either.
- **The tracker narrows on both of its paths.** The in-process watcher and the Windows broker record mutations through separate code, and narrowing only the first measured as no change at all on Windows. A new directory still counts, deliberately: the tracker is not watching it yet and cannot know what will be put in it.
- **`declaredKeys` is only ever supplied by the validating caller**, whose comparison is over exactly that set. Capture and `failedGenerationEnvironmentChanged` supply none and still record the whole walk.
- **`membershipPolicy` has one construction site**, is anchored at the project's own tsconfig rather than the generated one, and cannot be stale: a tsconfig edit changes a declared input and replaces the generation with its policy.

### Surface

Public API is additive: master's eleven exports plus `readProjectMembershipPolicy`, nothing removed or changed in shape. `collectProjectInputHashes` and `isProjectWalkPath` gained trailing optional parameters, so every existing caller compiles unchanged.

Twelve new cases and three changed ones. The three changes are contract changes rather than repairs: a `.txt` removal is no longer membership, which is what its own sibling already asserted for an edit to the same file; the out-of-program delivery returns rather than throws; and Metro no longer distinguishes a real failure by matching error text, because there is no text left to match.

Every changed test file parses, no file carries a NUL byte, and the working tree is clean with no strays.

### Forbidden contract

Three name-based exclusions survive, `.git`, `node_modules` and `.ttsc`, and the code says why each is not configuration-derived: no tsconfig can name them. No test-only branch, no environment-gated behaviour, and no fixture or consumer name appears anywhere in `packages/`.

The thing this cycle was for is the opposite of the prohibition: #1300 sealed one directory name and left the class open, and the class is what this replaced.

samchon (commented):
## Individual Self-Review — the adapter corrections (`c4602eec7`, `9c0dea997`, `a8bad0efe`, `512f147b7`)

Four read-only advisory reviews, one per commit, recorded together because each corrects the one before it and the chain only makes sense read as a chain. Adjudication is mine.

### What the chain was

`c4602eec7` closed four defects the first review found in `f513138c8`: the walk and `isProjectWalkPath` disagreeing, the live tracker never being narrowed, the walk hashing every file it reached, and capture reading more than before. It also found that the Windows broker records mutations on a path separate from the in-process watcher, which is why the first tracker fix measured as no change at all on this machine and needed a second site.

`9c0dea997` closed four more. The decisive one was procedural: `packages/metro/lib` was stale, so every local Metro run in this cycle had been exercising pre-branch code. Rebuilt, four cases failed, and the review had predicted all four. Three failed because the recorder fix is correct, and their `assert.deepEqual` expectations were written when the walk still carried the configs. The fourth was my own new case, which could never have passed: its fixture never ran a transform, so nothing recorded the tsconfig.

`a8bad0efe` fixed a guard that never fired. The `emptyOutDir` case still cost a compile per build, because the watcher reports events under the spelling its watch was opened with, `C:\Users\SamchonNam`, while the policy had read the configuration under the 8.3 short form `C:\Users\SAMCHO~1`. Two names for one directory sharing no common prefix.

`512f147b7` undid the way that fix was written. Canonicalizing in the tracker broke the rule `isProjectWalkPath` states two functions away: walk membership is lexical, because resolving to physical identity collapses spellings the walk deliberately keeps apart. A junction `outDir` whose target the walk hashes under its own name is exactly that, so the previous commit would have suppressed every event in a subtree the walk does cover. A missed membership change, which is strictly worse than the wasted compile it replaced. The mismatch had one source, the broker, and fixing it there let every comparison stay lexical.

### The lesson worth keeping

Three of these four commits fixed something the previous one introduced, and every one of those was found by reading rather than by running. The tests were green throughout. A rule that matches nothing looks exactly like a rule that is never needed, and a guard added with a correct rationale and no case is indistinguishable from one that works.

That is also why `assertRecreatingTheOutputDirectoryCostsNothing` exists: a review pointed out that every membership test still passed with the guard deleted. Writing that case is what exposed the spelling mismatch.

### Adjudicated, not accepted

A symlinked or case-variant excluded entry has nothing pinning it in either direction, and the `sources` directory-form hazard (a directory-shaped `extends` whose mtime moves when any child changes) is pre-existing. Both recorded for the next cycle rather than guessed at here.

samchon (commented):
## Individual Self-Review — the CI lane (`469b43cc8`, `6282c4127`, `1f3c35530`, `cd4185ae9`), and Overall rounds 2 and 3

Recorded together for the same reason as the previous group: each commit corrects the one before it.

### The finding that mattered most in this cycle was not about code

`bundler-defenses` is the only lane running `@ttsc/test-unplugin` and `@ttsc/test-metro`, it declares no `os`, so it runs on Linux, and the Windows mutation broker is gated on `process.platform === "win32"`. Every Windows-only path this cycle touched is dead code there. The spelling defect above measured four wasted compiles locally and would have measured zero in CI, forever, in either direction.

So `469b43cc8` added a Windows lane for the membership family. Then the reviews took it apart twice.

`1f3c35530` fixed two defects in it, the first of which would have turned this pull request red before the lane ever ran: `validation-plan.test.cjs` pins the lane count and the number of scoped native builds, and it runs in the `typecheck` lane every plan selects. The second was worse in kind: `planForPaths` mapped `packages/unplugin/**` and the rest to `bundler-defenses` alone, so an adapter-only change would never select the new lane. It ran on that commit only because the commit edited the planner, which fails open, which is precisely the shape of a check that appears to work and does not.

`cd4185ae9` closed the rest. The lane's own mappings claimed `packages/metro/**` while the lane ran no metro code, so it now runs metro's walk-facing cases too. The `.json` candidate guard was case-insensitive while the resolver it mirrors is case-sensitive. Only one of the four claimed mappings was pinned. And three unrelated `.cjs` files had been swept up by a prettier run that no gate even asks for, so that churn is reverted.

### Overall rounds

Round 2 found that nothing asserted the compiler's non-source inputs survive the walk narrowing. That was the cycle's riskiest claim resting on an argument rather than a test: had it been wrong, a tsconfig edit would silently have stopped invalidating anything. Measured, it holds, and it is a case now.

Round 3 found the documentation had not caught up with the corrections: recreating a configured output directory is free, which is what `emptyOutDir` and `output.clean` do on every build, and neither document said so.

### Verified rather than assumed

The lane's exact command was run here on Windows before it was trusted: thirteen adapter cases plus metro's, all green. A filter matching nothing hard-fails rather than passing empty. None of the thirteen injects a filesystem, so the broker really is the branch under test. And no other assertion anywhere pins a lane count, id list, matrix shape or check name, which is the question the first of these commits answered the hard way.

samchon (commented):
## Overall Self-Review — round 4, over the whole base-to-head diff at `030ad2808`

Fresh complete round over all 39 files and 18 commits. **No finding survived verification**, which makes this the round the merge gate needs. Everything below was re-inspected in this round rather than carried from an earlier one.

### Surface

The public API is additive: master's eleven exports plus `readProjectMembershipPolicy`, nothing removed or reshaped. `collectProjectInputHashes` and `isProjectWalkPath` gained trailing optional parameters, so every existing caller compiles unchanged. Both packages typecheck clean.

Every changed TypeScript file parses under `--experimental-strip-types --check`, none contains a NUL byte, and `prettier --list-different` over every changed `.ts`, `.md` and `.mdx` is empty. That class of artifact bit an earlier cycle twice through patch scripting, so it is checked mechanically rather than by eye.

### The decision points, re-verified

- **The walk and its predicate answer the same question**, in both packages. `selectExternalInputPaths` and `@ttsc/metro`'s recorder both use `isProjectWalkPath` as the sole test for "already covered", so a divergence puts an input in neither snapshot. Both now take the policy, and the agreement is asserted directly rather than through a scenario.
- **The tracker answers what the digest answers**, on both of its paths, and the broker translates the child's canonical spelling back to the walk's before anything compares it. Every comparison in the adapter is lexical, which is the rule `isProjectWalkPath` states and the one a canonicalizing fix broke mid-cycle.
- **Non-source inputs survive the walk narrowing.** The tsconfig, the package manifest and the plugin descriptor are proven by the universal host-input path, measured at one compile per edit, and pinned.
- **A policy reports every config it was read from**, including a candidate that does not exist yet and the resolver's case-sensitive `.json` sibling, because a memo can only be invalidated by what the policy reports.

### Coverage

23 cases bear on #1307, 4 on #1308, 4 on #1310, 1 on #1312, plus 3 existing cases updated for deliberate contract changes and 2 Metro cases restored from lower bounds to exact sets. Documentation issues #1309 and #1311 are verified by reading the rendered documents.

The cases that pin this cycle's own corrections were each proven to fail against the code they guard: the predicate symmetry, the output-directory recreation, the failed-delivery evidence, the tsconfig re-key, and the case-sensitive `extends` candidate.

### Documentation

Zero em-dashes in either document. The longest source line is 166 words in the README and 179 in the guide, against 1,751 and 1,781 at the base. Both describe the membership rule this cycle implements, including that recreating a configured output directory is free.

### Forbidden contract

No test-only branch, no environment-gated behaviour, and no fixture or consumer name anywhere in `packages/`. Three name-based exclusions survive, `.git`, `node_modules` and `.ttsc`, and the code says why each is not configuration-derived: no tsconfig can name them.

The prohibition this cycle exists to answer is the fourth one. samchon/ttsc#1300 sealed a single directory name and left the class open; what replaced it reads the configuration that actually knows.

Issue #1321, opened 2026-08-30
withTtsc discards a Metro config's existing babelTransformerPath


samchon:
## Problem

`withTtsc` replaces `transformer.babelTransformerPath` with this package's transformer and never reads the value it overwrites. A project that had configured its own Babel transformer loses it, silently, on the line that adopts `@ttsc/metro`.

Every other field in the config survives — `withTtsc` spreads `config` and spreads `config.transformer` — which is exactly what makes the loss hard to notice. The wrapper is additive in every direction except the one field it sets.

The canonical case is `react-native-svg-transformer`, whose entire installation is that one assignment:

```js
config.transformer.babelTransformerPath = require.resolve("react-native-svg-transformer");
```

Wrapping such a config means `.svg` imports stop being transformed. They do not error at build time: the ttsc transformer passes every non-TypeScript file straight to its *auto-detected* upstream, which is `@expo/metro-config/babel-transformer`, not the transformer the project chose. The build succeeds and the SVG components are wrong at runtime.

This is the failure shape the campaign keeps returning to — a successful build whose output is silently not what the configuration asked for (samchon/ttsc#1310, samchon/ttsc#1308) — and here it is reached by following the documented setup exactly.

## Evidence

Against the built package at `c62dc7d7e`:

```js
const before = {
  projectRoot: process.cwd(),
  transformer: {
    babelTransformerPath: "/abs/path/react-native-svg-transformer",
    assetPlugins: ["keep-me"],
  },
};
const after = withTtsc(before);
```

```
caller babelTransformerPath : /abs/path/react-native-svg-transformer
after  babelTransformerPath : .../packages/metro/lib/transformer.js
survived elsewhere in config?: false
assetPlugins preserved      : ["keep-me"]
TTSC_METRO_OPTIONS          : {}
```

The path is gone from the returned config entirely, and the serialized worker options carry no `upstreamTransformer`, so `resolveUpstreamTransformer` falls to `UPSTREAM_CANDIDATES` and loads the Expo default. `assetPlugins` surviving in the same object is the contrast: everything else is additive.

`packages/metro/src/index.ts` never reads `config.transformer.babelTransformerPath`, and `packages/metro/src/core/upstream.ts` resolves only an explicit `options.upstreamTransformer` or the fixed candidate list. Nothing connects the two.

## Consequence surface

| Caller's config | Today | Required |
| --- | --- | --- |
[...]

Issue #1320, opened 2026-08-30
withTtsc warns Next.js users on a premise Next.js does not have


samchon:
## Problem

`withTtsc` prints a warning to every Next.js user who wrote a `webpack` hook and no `turbopack` block, and both the warning's text and the docstring justifying it describe a Next.js behaviour that does not exist in the version this repository pins.

The message tells the reader that Next.js *would have* warned them and no longer will, because this wrapper now defines `turbopack`. Measured against Next 16.3.2, Next says nothing in that situation at all — there is no warning to suppress and no build to refuse. A user who reads our message and checks it finds the premise false.

The underlying fact the warning exists to convey is true and worth saying: a `webpack` hook does not apply to a Turbopack build, so a caller who configured webpack and then adopted this wrapper loses that configuration silently. What is wrong is the reason attached to it, and the reason is the part a reader can check.

This is the same defect class as the Vite alias report corrected in samchon/ttsc#1315: a diagnostic on this package's stderr whose justification does not hold. That channel also carries the out-of-program report (samchon/ttsc#1308), and its value depends on users trusting what this package prints.

## Evidence

Measured in `experimental/test-unplugin/.tmp/project`, Next.js 16.3.2, config carrying a `webpack` hook and no `turbopack` block:

```js
export default {
  distDir: "dist-probe",
  typescript: { ignoreBuildErrors: true },
  webpack(config) { return config; },
};
```

```
$ npx next build --turbopack
▲ Next.js 16.3.2 (Turbopack)
✓ Compiled successfully in 4.9s
...
exit=0
```

The complete 18-line output contains no occurrence of `webpack`, `ignored`, `will not`, or `warn`. Next neither refused the build nor mentioned the hook.

Searching Next's shipped code for such a check finds nothing either: no `configured while Turbopack`, no `webpack-config-conflict`, and no message pairing the two bundler names in an advisory sentence.

What this package currently states, in `packages/unplugin/src/next.ts`:

> Next refuses to build on Turbopack when a config carries a `webpack` hook and no `turbopack` block, because the webpack hook is then silently ignored. This wrapper always defines both, so that check can never fire again for anyone who uses it

and what it prints:

> `@ttsc/unplugin: withTtsc now configures Turbopack as well as webpack, so Next.js will not warn that your own `webpack` hook is ignored on a Turbopack build.`

Both sentences assert a Next behaviour that Next 16.3.2 does not have.

## Consequence surface

| Caller | Today | Correct |
[...]

Issue #1319, opened 2026-08-30
The Turbopack dedupe guard's recognised globs are an unverified contract with Turbopack


samchon:
## Problem

`withTtsc`'s Turbopack dedupe guard decides whether a caller's glob already covers a file set, and it is correct today only because Turbopack's matcher happens to behave the way the module assumes. Nothing in this repository verifies that assumption, and the assumption is not written down as one.

The guard skips registering `*.ts` when another rule carrying this loader "names every file with the extension". Recognising a glob that does **not** in fact cover everything leaves the uncovered files with no ttsc rule at all: the build succeeds and every plugin-driven construct in them survives untransformed into a runtime failure. That is samchon/ttsc#1310, which this module has now produced twice — once by covering only webpack, once by reading a glob's extension while ignoring its path (samchon/ttsc#1314).

Both times the failure was silent. A third time would be too, and a Next.js upgrade is enough to cause it: the recognised set is an external contract with no test.

## Evidence

Measured against a real `next build --turbopack`, Next.js 16.3.2, in `experimental/test-unplugin/.tmp/project`. A page imports a root-level `root-probe.ts` and a nested `src/nested-probe.ts`; each rule carries one trivial loader that rewrites a sentinel, and the built output says which files it reached.

| `turbopack.rules` glob | root-level `.ts` | nested `.ts` |
| --- | --- | --- |
| `*.ts` | yes | yes |
| `**/*.ts` | yes | yes |
| `*.{ts,tsx}` | yes | yes |
| `**/*.{ts,tsx}` | yes | yes |
| `**/{*.ts,*.tsx}` | yes | yes |
| `{**/,}*.ts` | yes | yes |
| `**/**/*.{ts,tsx}` | yes | yes |
| `src/*.ts` | no | **no** |
| `src/**/*.ts` | no | **no** |
| `./src/*.ts` | no | yes |
| `**/src/*.ts` | no | yes |
| `nested-probe.ts` | no | yes |

Every spelling the guard recognises is sound: all seven cover both depths. So the guard is correct today, and this issue is not a defect report about its behaviour.

[...]

PR #1322, opened 2026-08-30
fix(metro): chain the caller's transformer, and verify the Turbopack claims

samchon:
Cycle 4 of the `@ttsc/unplugin` campaign. Two issues closed, one withdrawn as invalid.

## What this closes

- **#1321** — Metro's `withTtsc` replaced `transformer.babelTransformerPath` and never read what it overwrote, so a project using `react-native-svg-transformer` lost it silently.
- **#1319** — the Turbopack dedupe guard's recognised globs were an unverified contract with Turbopack's matcher. Verifying them found a live defect in that guard.

**#1320 is closed as invalid** and its change reverted. It claimed Next.js has no webpack/Turbopack conflict check; Next 16.3.2 has one, it is `process.exit(1)`, and it is gated on `TURBOPACK === "auto"` — set only when no bundler flag is passed. Every measurement behind that issue used `--turbopack`, the one invocation that disables it. The code it asked to change was correct. What survives is a correction the other way: Next does not *warn*, it stops the build, and the message now says so.

## The defect the verification found

`{src/,}*.ts` was recognised as naming every `.ts` file, on the reasoning that its alternatives include a bare `*.ts`. Measured against a real Turbopack build, that glob matches **nothing** — not the project root, not `src/`. So the guard suppressed this wrapper's own rules in favour of a rule that transforms no file at all: samchon/ttsc#1310, caused by the guard written to prevent it, with a unit test asserting the wrong answer as though it were right.

The rule is now **every** alternative unscoped and one naming the extension, instead of *any*. That accepts all eight spellings a real build confirms and refuses every scoped shape, brace groups included.

## Measurements, not assertions

Taken from real `next build --turbopack` runs on Next.js 16.3.2:

| Claim | Measured |
| --- | --- |
| `{src/,}*.ts` covers the project because one alternative does | covers **nothing** |
| a scoped rule "covers its own subtree" | `src/*.ts` and `src/**/*.ts` cover nothing; `./src/*.ts` and a bare `nested-probe.ts` do match |
| loaders run right-to-left (inferred from webpack) | confirmed for Turbopack itself |
| Next ships no webpack/turbopack check | it does, and it exits 1 |

## Two regressions in the #1321 fix, both caught before merge

The first version compared directory strings, so a second installed copy of `@ttsc/metro` passed the guard, became its own upstream, and — worker options being process-global — read itself back out of `TTSC_METRO_OPTIONS` and recursed until the stack ended. It also passed relative and bare spellings through untouched, so the worker looked for them inside `@ttsc/metro`, turning a transformer that used to be ignored into a build that fails outright.

Both fall to one ordering: resolve the declared value from the project first, then ask whether the resolved *module* is a `@ttsc/metro` transformer — by real path for this copy, by the owning `package.json` for any other. That also catches `require.resolve("@ttsc/metro/transformer")`, which no string comparison could.

## Verification

| Case | Reverted behaviour | Result |
| --- | --- | --- |
| glob judged by every alternative | back to *any* | `{src/,}*.ts` recognised, real build transforms nothing |
| chaining by resolved module | back to directory comparison | duplicate copy adopted as its own upstream |
| project resolution | dropped | relative path travels unanchored |
| recognised-set drift guard | one glob removed from the harness | unit case fails naming it |

Eleven `babelTransformerPath` spellings are pinned, and the experimental suite drives every recognised glob — plus two that must be refused — through real Turbopack builds at three depths: nested `.ts`, root `.ts`, and nested `.tsx`.

### DIFF
--- a/experimental/test-unplugin/src/index.ts
+++ b/experimental/test-unplugin/src/index.ts
@@ -86,6 +86,7 @@ function main() {
   verifyRspackBuild();
   verifyFarmBuild();
   verifyNextBuild();
+  verifyTurbopackRecognisedGlobs();
   verifyBunBuild();
   verifyBunRuntime();
   console.log("Success");
@@ -161,7 +162,7 @@ function prepareWorkspace() {
             },
           ],
         },
-        include: ["src", "pages"],
+        include: ["src", "pages", "turbopack-root-entry.ts"],
       },
       null,
       2,
@@ -183,7 +184,7 @@ function prepareWorkspace() {
           noEmit: true,
           resolveJsonModule: true,
         },
-        include: ["next-env.d.ts", "pages", "src"],
+        include: ["next-env.d.ts", "pages", "src", "turbopack-root-entry.ts"],
       },
       null,
       2,
@@ -213,6 +214,7 @@ function prepareWorkspace() {
   writeSource("farm-entry.ts", "farm-installed-ok");
   writeSource("next-entry.ts", "next-installed-ok");
   writeSource("bun-entry.ts", "bun-installed-ok");
+  writeTurbopackRootEntry();
   writeNextPage();
   writeTransformPlugin();
   writeViteConfig();
@@ -236,15 +238,38 @@ function writeSource(file, marker) {
   );
 }
 
+/**
+ * A source at the project root, which `src/next-entry.ts` cannot stand in for.
+ *
+ * The dedupe guard skips the wrapper's own rules when a caller's glob already
+ * names every file with the extension, and whether a glob does that is
+ * Turbopack's answer, not ours. A `**` + `/` prefix that required at least one
+ * segment would cover `src/` and miss this file, which is how a recognised
+ * spelling turns into samchon/ttsc#1310: no rule, no transform, green build.
+ * `middleware.ts` and `instrumentation.ts` are the real files at this depth.
+ */
+function writeTurbopackRootEntry() {
+  fs.writeFileSync(
+    path.join(workspace, "turbopack-root-entry.ts"),
+    [
+      'export const rootValue = mark("turbopack-root-ok");',
+      "console.log(rootValue);",
+      "",
+    ].join("\n"),
+    "utf8",
+  );
+}
+
 function writeNextPage() {
   fs.mkdirSync(path.join(workspace, "pages"), { recursive: true });
   fs.writeFileSync(
     path.join(workspace, "pages", "index.js"),
     [
       'import { value } from "../src/next-entry";',
+      'import { rootValue } from "../turbopack-root-entry";',
       "",
       "export default function Page() {",
-      "  return value;",
+      "  return value + rootValue;",
       "}",
       "",
     ].join("\n"),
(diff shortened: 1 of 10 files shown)


samchon (commented):
## Review ledger — Individual adjudications and four Overall rounds

### Individual Self-Review adjudications

**`496c0987e` (#1321).** Two findings accepted, both severe, both mine.

- *A duplicate installed copy became its own upstream and recursed.* The guard compared directory strings, so a second `@ttsc/metro` in a differently hoisted `node_modules` passed it. Worker options are process-global, so the adopted copy read itself back out of `TTSC_METRO_OPTIONS` and blew the stack on every file. Before chaining existed that path was simply dropped, so my fix was worse than the defect for that shape.
- *Three resolution bases disagreed.* The value was stored verbatim, the own-check resolved it against `cwd`, and the worker resolved it from inside `@ttsc/metro`. A relative or bare spelling therefore failed to load — a transformer that used to be ignored became a build that fails outright.

Both fall to one ordering, applied in `f7faab931`: resolve from the project first, then judge the resolved module. Coverage gaps the review measured (only the union of the two own-check branches was pinned; no duplicate-copy case; no relative case) are closed by eleven pinned spellings.

**`ca4232b7f` + `fd452926f` (#1319, #1320).** One finding accepted and one issue withdrawn.

- *#1320's premise was false and the code it changed was correct.* Next 16.3.2 ships the check in `dist/lib/turbopack-warning.js` — `log.error` then `process.exit(1)` — gated on `TURBOPACK === "auto"`, which `dist/lib/bundler.js` sets only when no bundler flag was passed. Every measurement behind that issue used `--turbopack`, the one invocation that disables it. Reverted; the issue is closed as invalid. What survives is a correction the other way: Next does not warn, it stops the build, and the message now says so.
- *The recognised set was missing a spelling the guard accepts.* Following that up found `{src/,}*.ts` matches **nothing** in a real Turbopack build, so recognising it left every module with no rule — samchon/ttsc#1310 caused by the guard meant to prevent it.

Rejected: nothing. Both reviews' accepted findings are applied.

### Overall Self-Review

**Round 1 — one finding.** Both Metro documents stated chaining without saying the path is resolved from `projectRoot`, or that naming this package is refused in every spelling. Applied in `80d936bba`.

**Round 2 — one finding, the same class one level up.** The corrected predicate is right for every measured spelling and still answers for unmeasured ones: `*.{ts}` and `{,**/}*.ts` were both recognised without anyone checking whether Turbopack expands a single-alternative or leading-empty group. A predicate cannot be closed, so recognition became an exact set of ten spellings, each driven through a real build. Applied in `871ad5f79`.

**Round 3 — one finding.** Both documents promised without qualification that a hand-wired rule "is left alone rather than registered twice", which the allowlist keeps for ten spellings and deliberately breaks for the rest — the same over-claiming samchon/ttsc#1315 and samchon/ttsc#1320 were about. Applied in `826ea87ef`.

**Round 4 — clean.** Re-read the Metro chaining logic and the guard; confirmed no reference to the removed predicate survives; verified mechanically that the ten spellings in the code, in both documents, and in the harness list are one set; ran the full `test-metro` and `test-unplugin` suites to completion (both `Success`); swept the branch for placeholders, whitespace damage, NUL bytes and unrelated files. Ten changed files, all owned by these two issues. Nothing found.

### Verification that could not run locally

The experimental install suite needs a full pack-and-install and is left to CI, which selects it from `packages/unplugin/**`. Its twelve Turbopack builds were each reproduced by hand against the installed Next 16.3.2 workspace first, including the two refusals and the three markers.

Issue #1323, opened 2026-09-01
fix(lint): print-width oscillates on calls inside binary expressions


bradleat:
## Problem

`ttsc format` in `@ttsc/lint` 0.28.4 does not converge when `format/print-width` visits calls that are fragments of a binary expression. The same implementation is still present on `master`.

This is a genuine two-state oscillation, not multiple formatters or concurrent package formatting.

## Minimal shape

```ts
export function isStrategy(value: unknown): boolean {
  return isSync(
    value,
  ) || isLazy(
    value,
  ) || isStatic(
    value,
  );
}
```

With the normal format cascade:

1. One pass independently flattens each call:

```ts
return isSync(value) || isLazy(value) || isStatic(value);
```

2. The complete binary expression now exceeds `printWidth`, so the next pass independently breaks the same calls.
3. The process repeats until `maxFormatPasses` and exits 2:

```text
@ttsc/lint: format cascade did not converge after 10 passes
```

## Observed trace

Instrumenting `runFormat` against a real project produced the same findings on all ten passes:

```text
format/print-width: 11 findings
format/quotes:       4 findings
applied:             11 edits per pass
```

The print-width edits alternate flat/broken layouts. The quote edits are nested inside four call replacements and are skipped by the atomic overlap resolver; they are not the cause.

The affected nodes were calls inside `||`, `??`, and `&&` binary chains. A package-local single-threaded reproduction fails identically, excluding Nx/concurrency.

## Cause

`formatPrintWidth` owns `CallExpression` but does not own or print `BinaryExpression`. It therefore makes a width decision for each child call without controlling the complete line. Several individually-correct child edits can collectively change whether the enclosing binary line fits, reversing every decision on the next pass.

This violates the rule's stated safety model: nodes below syntax the dispatcher cannot completely control should remain byte-identical.

## Proposed patch

Abstain when the visited reflow node is a fragment of a binary expression, until the structured printer can own and render the complete binary chain:

```go
// After hasReflowAncestor:
if hasBinaryExpressionAncestor(node) {
  return
}

func hasBinaryExpressionAncestor(node *shimast.Node) bool {
  if node == nil {
    return false
  }
  for parent := node.Parent; parent != nil; parent = parent.Parent {
    if parent.Kind == shimast.KindBinaryExpression {
      return true
    }
  }
  return false
}
```

[...]

PR #1325, opened 2026-09-02
fix(lint): stop print-width oscillation in binary expressions

samchon:
## Intent

Issue #1323 demonstrates a two-state `format/print-width` cascade when a supported call or value literal is nested below a binary expression that the structured printer does not own. This pull request restores the formatter's conservative ownership invariant so the cascade converges.

## Change

- Abstain from print-width reflow when any uncontrolled `BinaryExpression` is an ancestor of the visited node.
- Keep the existing direct destructuring assignment behavior by skipping only the target's directly owning `=`, then resume scanning for outer binary expressions.
- Add a command-level regression covering `||`, `??`, `&&`, a destructuring assignment below `||`, a standalone-call negative twin, and byte-identical second-run output.
- Document the binary-expression ownership boundary and distinguish binding destructuring patterns from direct assignment targets.
- Pin Browserslist 4.28.8 after the required dependency-audit lane exposed two advisories with released fixes.

## Boundary

Full `BinaryExpression` printer coverage and binary-chain reflow remain outside this change. Unsupported binary subtrees stay byte-identical until the structured printer can own the complete chain.

## Verification

- Before the local-test restriction, the baseline native reproduction reached the ten-pass cap with exit 2; the corrected reproduction exited 0 twice with identical output.
- Before the local-test restriction, the focused command regression and relevant print-width group passed, including a repeated run. A full lint run reached its ten-minute limit in unrelated corpus work after exposing two direct-assignment regressions; those regressions were corrected and covered, but the full run was not repeated.
- Before the local-test restriction, `pnpm run check:dependencies` passed with `low=9`, `moderate=23`, `high=1`, `critical=0`, and `waived=1` after the Browserslist override.
- The required final `pnpm format` completed successfully. `git diff --check` and the tracked worktree are clean.
- No local test or suite was run after the restriction.
- All 29 ordinary CI checks passed on exact head `c822e5fb2862b0f8caae0cc54d030abaf9f515d2`.

## Self-Review

- Individual Self-Review on the implementation found the overbroad destructuring-target exemption, and the correction review plus the first Overall round found its source and guide explanation mismatch. Both findings were corrected.
- Individual Self-Review of exact commit `c822e5fb2862b0f8caae0cc54d030abaf9f515d2` is clean.
- Final Overall Self-Review of exact diff `36c8488c347917e807cba88a9f9b9df47a54c6cc..c822e5fb2862b0f8caae0cc54d030abaf9f515d2` is clean.

### DIFF
--- a/packages/lint/linthost/rules_format_print_width.go
+++ b/packages/lint/linthost/rules_format_print_width.go
@@ -187,6 +187,18 @@ func (formatPrintWidth) Check(ctx *Context, node *shimast.Node) {
     return
   }
 
+  // Abstain on any uncontrolled reflow target nested below a binary
+  // expression. The structured printer does not own BinaryExpression, so
+  // independently flattening or breaking its child calls and value literals
+  // can change whether the complete binary line fits. The next cascade pass
+  // then reverses those child decisions and `ttsc format` oscillates to the
+  // 10-pass cap. Destructuring assignment targets are the exception: their
+  // literal printer owns the complete left-hand list and already charges the
+  // `= value` suffix against its width budget.
+  if hasUncontrolledBinaryExpressionAncestor(node) {
+    return
+  }
+
   // Abstain on any node nested inside a template-literal substitution.
   // Prettier renders `${…}` expressions at printWidth:Infinity, it
   // never breaks an interpolation the source wrote on one line, so
@@ -499,6 +511,21 @@ func hasReflowAncestor(node *shimast.Node) bool {
   }
   for parent := node.Parent; parent != nil; parent = parent.Parent {
     if isReflowKind(parent.Kind) {
+      return true
+    }
+  }
+  return false
+}
+
+// hasUncontrolledBinaryExpressionAncestor reports whether `node` is a fragment
+// of a binary expression whose layout it cannot own. A destructuring assignment
+// target is safe because the literal printer owns that complete left-hand list.
+func hasUncontrolledBinaryExpressionAncestor(node *shimast.Node) bool {
+  if node == nil || isDestructuringAssignmentTarget(node) {
+    return false
+  }
+  for parent := node.Parent; parent != nil; parent = parent.Parent {
+    if parent.Kind == shimast.KindBinaryExpression {
       return true
     }
   }
--- a/packages/lint/linthost/rules_format_print_width.go
+++ b/packages/lint/linthost/rules_format_print_width.go
@@ -519,15 +519,25 @@ func hasReflowAncestor(node *shimast.Node) bool {
 
 // hasUncontrolledBinaryExpressionAncestor reports whether `node` is a fragment
 // of a binary expression whose layout it cannot own. A destructuring assignment
-// target is safe because the literal printer owns that complete left-hand list.
+// target may skip its owning `=` because the literal printer owns that complete
+// left-hand list, but an enclosing binary expression remains uncontrolled.
 func hasUncontrolledBinaryExpressionAncestor(node *shimast.Node) bool {
-  if node == nil || isDestructuringAssignmentTarget(node) {
+  if node == nil {
     return false
   }
+  skipOwningAssignment := isDestructuringAssignmentTarget(node)
+  child := node
   for parent := node.Parent; parent != nil; parent = parent.Parent {
     if parent.Kind == shimast.KindBinaryExpression {
-      return true
+      expression := parent.AsBinaryExpression()
+      if skipOwningAssignment && expression != nil && expression.OperatorToken != nil &&
+        expression.OperatorToken.Kind == shimast.KindEqualsToken && expression.Left == child {
+        skipOwningAssignment = false
+      } else {
+        return true
+      }
     }
+    child = parent
   }
   return false
 }
--- a/packages/lint/linthost/rules_format_print_width.go
+++ b/packages/lint/linthost/rules_format_print_width.go
@@ -192,9 +192,10 @@ func (formatPrintWidth) Check(ctx *Context, node *shimast.Node) {
   // independently flattening or breaking its child calls and value literals
   // can change whether the complete binary line fits. The next cascade pass
   // then reverses those child decisions and `ttsc format` oscillates to the
-  // 10-pass cap. Destructuring assignment targets are the exception: their
-  // literal printer owns the complete left-hand list and already charges the
-  // `= value` suffix against its width budget.
+  // 10-pass cap. A destructuring assignment target may skip only its directly
+  // owning `=` because the literal printer owns the complete left-hand list and
+  // charges the `= value` suffix. Any binary ancestor above that assignment
+  // remains uncontrolled.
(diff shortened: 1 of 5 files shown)


packages/lint/linthost/rules_format_print_width.go
```diff
@@ -505,6 +517,21 @@ func hasReflowAncestor(node *shimast.Node) bool {
   return false
 }
 
+// hasUncontrolledBinaryExpressionAncestor reports whether `node` is a fragment
+// of a binary expression whose layout it cannot own. A destructuring assignment
+// target is safe because the literal printer owns that complete left-hand list.
+func hasUncontrolledBinaryExpressionAncestor(node *shimast.Node) bool {
+  if node == nil || isDestructuringAssignmentTarget(node) {
```
samchon: `isDestructuringAssignmentTarget(node)` returns early for the whole ancestor search, so a target such as `([alphaValue, ...remainingValues] = sourceValues) || fallbackValue` is independently reflowed even though the outer `||` remains uncontrolled. The direct assignment's owning `=` is safe to skip, but ancestor scanning must continue beyond it. Reproduced with the native formatter at `printWidth: 40`; the nested target changed from one line to multiple lines. Accepted and queued for correction.

samchon (commented):
Individual Self-Review of `b3b95feddfcef04becdd7721593b8f192c5da2c5` (`fix(lint): preserve binary expression fragments`). This commit resolves #1323 by making print-width abstain below uncontrolled binary expressions, adding command-level convergence, idempotency, and negative-twin coverage, and documenting the boundary. One candidate survived adjudication and is recorded inline: destructuring assignment targets skipped every binary ancestor rather than only their owning `=`. The proposed oscillation witness was not reproduced verbatim, but the underlying ownership defect was reproduced as an observable rewrite of a long destructuring target below `||`; it is accepted for correction. No other findings survived across source, tests, documentation, platforms, concurrency, or compatibility.

samchon (commented):
Individual Self-Review of `4c975661b89f8e53ac937dbc703609d93be75285` (`fix(lint): complete binary ancestor guard`). This correction makes destructuring targets skip only their own assignment, adds the command regression boundary, and pins the patched Browserslist release. One candidate survived adjudication: the source and guide still describe destructuring targets as general exceptions, while the code and new test exempt only the directly owned `=`. The mismatch is accepted for correction. No other findings survived across correctness, test isolation, lockfile integrity, platforms, concurrency, or compatibility. The reviewer ran no local tests as instructed; CI is pending.

packages/lint/linthost/rules_format_print_width.go
```diff
@@ -187,6 +187,18 @@ func (formatPrintWidth) Check(ctx *Context, node *shimast.Node) {
     return
   }
 
+  // Abstain on any uncontrolled reflow target nested below a binary
+  // expression. The structured printer does not own BinaryExpression, so
+  // independently flattening or breaking its child calls and value literals
+  // can change whether the complete binary line fits. The next cascade pass
+  // then reverses those child decisions and `ttsc format` oscillates to the
+  // 10-pass cap. Destructuring assignment targets are the exception: their
```
samchon: This says destructuring assignment targets are the exception without limiting that exception to their directly owned `=`. The helper below correctly resumes scanning and abstains when the assignment itself sits under another binary expression. Narrow this explanation to match the implementation and regression test.

website/src/content/docs/lint/format.mdx
```diff
@@ -226,8 +226,8 @@ Width is measured in **display columns** (Prettier's `getStringWidth`), not UTF-
 This list is the reflow's own, scoped to node shapes. The boundary that applies to the whole format set is [Scope](#scope).
 
 - JSX elements and fragments.
-- Binary expressions.
-- Destructuring patterns.
+- Binary expressions, including otherwise supported calls and value literals nested inside them. The rule preserves those fragments until it can render the complete binary chain. Destructuring assignment targets remain covered because their literal printer owns the complete left-hand list.
```
samchon: `Destructuring assignment targets remain covered` reads as a general exception, but the new boundary test preserves one below `||`. State that only a direct target's own `=` is covered and that an enclosing binary expression restores abstention.

packages/lint/linthost/rules_format_print_width.go
samchon: Overall finding: limit this exception to the target's own assignment. A destructuring assignment nested under `||`, `??`, or another binary expression is still uncontrolled and must abstain, as the corrected helper now does.

website/src/content/docs/lint/format.mdx
samchon: Overall finding: document the same direct-assignment boundary as the implementation. The current final sentence implies all destructuring assignment targets remain covered, including the newly preserved `||` witness.

samchon (commented):
Overall Self-Review finding round on exact diff `36c8488c347917e807cba88a9f9b9df47a54c6cc..4c975661b89f8e53ac937dbc703609d93be75285`. I read every hunk in all five changed files and traced the rule through the shared CLI and LSP cascades, AST parent semantics, direct and nested destructuring assignments, command-test isolation and idempotency, existing assignment regressions, dependency-audit policy and lockfile resolution, documentation, platforms, concurrency, compatibility, and closing metadata. One sound improvement survived: two explanations still present destructuring assignment targets as a general binary-expression exception even though the corrected code and test skip only the directly owned `=` and abstain under any outer binary ancestor. No other finding survived this round. Local tests were not run in this round per user instruction; CI is running on this exact head.

samchon (commented):
Individual Self-Review of `c822e5fb2862b0f8caae0cc54d030abaf9f515d2` (`docs(lint): clarify binary assignment boundary`). The read-only review covered every changed hunk in the exact parent-to-commit diff. The Go comment and guide now match the implementation and regression boundary: a destructuring assignment target may skip only its directly owning `=`, while any outer binary expression forces abstention. Direct assignment reflow remains compatible, and nested assignments remain preserved. The guide paragraph uses one source line, direct prose, and no manual word wrapping. No candidate finding survived. The reviewer ran no local tests or suites as instructed; CI is pending.

samchon (commented):
Final clean Overall Self-Review of exact diff `36c8488c347917e807cba88a9f9b9df47a54c6cc..c822e5fb2862b0f8caae0cc54d030abaf9f515d2`. I read every hunk in all five changed files and traced the binary-ancestor guard through the rule visitor, AST parent semantics, direct and nested destructuring assignments, the shared CLI and both LSP format cascades, command-test isolation and second-run idempotency, existing assignment compatibility tests, dependency-audit policy and lockfile resolution, documentation, platforms, concurrency, and issue-closing metadata. The earlier source and guide ambiguity is corrected: only the directly owning `=` may be skipped, while any outer binary expression remains uncontrolled. No sound improvement or candidate finding survived this fresh round. Local tests were not run in this round per user instruction; CI is running on this exact head.

Issue #1359, opened 2026-09-09
[Bug] ttsx does not support TC39 class decorators


Abion47:
## Environment   
  - **ttsc Version**: 0.29.0
  - **tsx Version**: 4.23.12
  - **NodeJS Version**: 24.15.0, 26.8.1
  - **TypeScript Version**: 7.0.2

### tsconfig.json

```json
{
  "compilerOptions": {
    "module": "esnext",
    "target": "esnext",
    "lib": ["esnext"],
    "types": ["node"],
    "rootDir": "./src",
    "outDir": "./dist",
    "declaration": true,
    "strict": true
  },
  "include": ["./src/**/*.ts"],
  "exclude": ["./src/**/*.test.ts"]
}
```

## Summary

[Since Typescript 5.0](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#decorators), Typescript has supported [Stage 3 TC39 decorators](https://github.com/tc39/proposal-decorators) out of the box. However, `ttsx` does not appear to recognize this, and results in an error when trying to execute code containing TC39 decorator syntax.

Consider the following code:

```typescript
function sayHelloClass<T extends { new (...args: any[]): {} }>(ClassType: T, context: ClassDecoratorContext) {
  return class extends ClassType {
    constructor(...args: any[]) {
      super(...args);

      console.log(`Hello Class ${context.name}`);
    }
  };
}

function sayHelloMethod<T>(target: (this: T, ...args: any[]) => any, context: ClassMethodDecoratorContext) {
  return function(this: T, ...args: any[]): any {
    console.log(`Hello Function ${context.name.toString()}`);
    return target.apply(this, args);
  }
}

@sayHelloClass
class Foo {
  constructor(public bar: string) {}

  @sayHelloMethod
  getBar() {
    return this.bar;
  }
}

const foo = new Foo('abc');
console.log(foo.getBar());
```

When run with `tsx`, it produces the following output:

```
Hello Class Foo
Hello Function getBar
abc
```

But when run with `ttsx`, it produces this output instead:

```
SyntaxError: Invalid or unexpected token
    at compileSourceTextModule (node:internal/modules/esm/utils:355:16)
    at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:93:18)
    at #translate (node:internal/modules/esm/loader:441:20)
    at afterLoad (node:internal/modules/esm/loader:509:29)
[...]

PR #1360, opened 2026-09-09
fix(ttsx): lower standard decorators for Node execution

samchon:
This pull request owns the single-issue campaign for #1359: execute standard decorators through ttsx, including registered entries and raw TypeScript dependencies.

The reproduced failure is ESNext output retaining decorators that Node cannot execute. The runtime compilation will lower that syntax while retaining the project's library and module semantics. Investigation is limited to this decorator execution defect and its consequences.

Verification is pending. Baseline reproduction fails in ESM and CommonJS; an ES2025 runtime emit executes the expected class/method behavior. Ordinary CI, Individual Self-Review, and Overall Self-Review are required before merge.

### DIFF
--- a/packages/ttsc/src/launcher/internal/runtimeHooks.ts
+++ b/packages/ttsc/src/launcher/internal/runtimeHooks.ts
@@ -1367,7 +1367,10 @@ function emitOrphanSource(
 }
 
 /**
- * Emit a source file only for CommonJS export-name discovery.
+ * Read actual owned output for CommonJS export-name discovery, falling back to
+ * isolated emission only when no project emitted this source. Project
+ * transforms and const-enum settings decide which names exist at runtime. An
+ * ESM emit is lowered as JavaScript solely to scan its CommonJS names.
  *
  * This intentionally does not read or write the runtime orphan cache. Name
  * discovery may inspect a source dependency without executing it, so sharing
@@ -1380,6 +1383,14 @@ function emitCommonJsForNameScan(filename: string): string | null {
   if (cached !== undefined) {
     return cached;
   }
+  const served = serveEntryEmit(real) ?? serveDependencyEmit(real);
+  if (
+    served !== null &&
+    moduleFormat(real, served.moduleOptions) === "commonjs"
+  ) {
+    commonJsNameScanSources.set(real, served.source);
+    return served.source;
+  }
   let tsgo: string;
   try {
     tsgo = resolveTsgo({ cwd: path.dirname(real) }).binary;
@@ -1389,10 +1400,14 @@ function emitCommonJsForNameScan(filename: string): string | null {
   }
   const outDir = createCanonicalTempDirectory("ttsx-export-scan-");
   try {
+    // Compile the owned JavaScript, not the original TypeScript whose project
+    // may already have erased or transformed declarations. No source executes.
+    const input = served === null ? real : path.join(outDir, "source.cts");
+    if (served !== null) fs.writeFileSync(input, served.source);
     const res = spawnNative(
       tsgo,
       [
-        real,
+        input,
         "--module",
         "commonjs",
         ...ISOLATED_EMIT_ARGS,
@@ -1403,7 +1418,7 @@ function emitCommonJsForNameScan(filename: string): string | null {
       { cwd: path.dirname(real), encoding: "utf8" },
     );
     const emitted = pickEmittedJavaScript(
-      real,
+      input,
       parseEmittedFiles(outputText(res.stdout)),
     );
     const lowered = emitted === null ? null : readFileOrNull(emitted);
--- a/packages/ttsc/src/launcher/internal/runtimeHooks.ts
+++ b/packages/ttsc/src/launcher/internal/runtimeHooks.ts
@@ -1530,8 +1530,9 @@ function pickEmittedJavaScript(
  * Runtime CommonJS consumers see the getters that helper installs, but Node's
  * ESM linker only exposes named imports it can statically identify from
  * `exports.name = ...` assignments. For relative star re-exports whose emitted
- * target is available, replace the helper call with explicit configurable
- * export placeholders followed by the same `__createBinding` getter install.
+ * target is available, advertise its names through inert assignments that
+ * Node's static lexer recognizes. The original helper still owns every runtime
+ * binding, including values static discovery cannot enumerate.
  */
 function exposeCommonJsStarExports(
   source: string,
@@ -1542,10 +1543,17 @@ function exposeCommonJsStarExports(
     return source;
   }
   const reserved = collectStaticCommonJsExportNames(source);
-  let index = 0;
+  const executable = maskCommentsAndStrings(source);
   return source.replace(
-    /^(\s*)__exportStar\(\s*require\((["'])([^"']+)\2\)\s*,\s*exports\s*\);/gm,
-    (statement: string, indent: string, _quote: string, specifier: string) => {
+    commonJsExportStarPattern(),
+    (
+      statement: string,
+      indent: string,
+      _quote: string,
+      specifier: string,
+      offset: number,
+    ) => {
+      if (executable[offset + indent.length] === " ") return statement;
       const names = [
         ...collectStarExportNames(emittedFile, sourceFile, specifier),
       ].filter(
@@ -1561,15 +1569,10 @@ function exposeCommonJsStarExports(
       for (const name of names) {
         reserved.add(name);
       }
-      const receiver = `__ttsx_export_star_${index++}`;
-      return [
-        ...names.map((name) => `${indent}exports.${name} = void 0;`),
-        `${indent}var ${receiver} = require(${JSON.stringify(specifier)});`,
-        ...names.map(
-          (name) =>
-            `${indent}__createBinding(exports, ${receiver}, ${JSON.stringify(name)});`,
-        ),
-      ].join("\n");
+      const hints = names.map((name) => `exports.${name} = void 0;`).join(" ");
+      // A block remains one statement in an unbraced control-flow body. Keep
+      // the original line count and never execute the static lexer hints.
+      return `${indent}{ if (false) { ${hints} } ${statement.slice(indent.length)} }`;
     },
   );
 }
@@ -1810,13 +1813,19 @@ function collectSourceCommonJsExportNames(
 
 function collectExportStarSpecifiers(source: string): string[] {
   const specifiers: string[] = [];
-  const pattern =
-    /^(\s*)__exportStar\(\s*require\((["'])([^"']+)\2\)\s*,\s*exports\s*\);/gm;
+  const pattern = commonJsExportStarPattern();
+  const executable = maskCommentsAndStrings(source);
   let match: RegExpExecArray | null;
   while ((match = pattern.exec(source)) !== null) {
+    if (executable[match.index + match[1]!.length] === " ") continue;
     specifiers.push(match[3]!);
   }
   return specifiers;
+}
+
+/** TypeScript-Go's inline and tslib-qualified CommonJS star-helper calls. */
+function commonJsExportStarPattern(): RegExp {
+  return /^([ \t]*)(?:[A-Za-z_$][\w$]*\.)?__exportStar\(\s*require\((["'])([^"']+)\2\)\s*,\s*exports\s*\);/gm;
 }
 
 function resolveEmittedRequire(
@@ -1854,7 +1863,8 @@ function resolveSourceSpecifier(
   }
   const base = path.resolve(path.dirname(sourceFile), specifier);
   if (path.extname(base).length !== 0) {
-    return isFile(base) ? base : null;
+    if (isFile(base)) return base;
+    return typescriptSourcesForJavaScriptSpecifier(base).find(isFile) ?? null;
   }
   for (const extension of TYPESCRIPT_EXTENSIONS) {
     const candidate = base + extension;
--- a/packages/ttsc/src/launcher/internal/runtimeHooks.ts
+++ b/packages/ttsc/src/launcher/internal/runtimeHooks.ts
@@ -42,11 +42,9 @@ import { inlineServedSourceMap } from "./servedSourceMap";
  *    type-stripping cannot do cross-file type-only elision — e.g. a
  *    value-shaped import of a type+namespace merge survives stripping and
  *    dangles at runtime.
- * 3. No owning tsconfig → transform the lone file by the format it resolves to: a
- *    CommonJS-classified file (`.cts`, or a `.ts` in a package without `type:
- *    "module"`) is lowered to CommonJS through a tsgo single-file emit so its
- *    `export` syntax becomes `module.exports`; any other (ESM) file keeps the
- *    fast in-process `mode: "transform"` type-strip.
+ * 3. No owning tsconfig: use an isolated tsgo emit in the runtime module format,
+ *    lowering standard decorators and CommonJS imports/exports. Type stripping
+ *    remains recovery when no compiler emit is available.
  *
  * The hooks are synchronous and run on the main thread (not a loader worker):
  * that is what lets a CommonJS `require("./x")` chain reach them and what makes
@@ -1037,7 +1035,7 @@ function owningModuleOptions(filename: string): OwningModuleOptions | null {
     // The owning project cannot be read, so nothing is known about the format
     // it would have emitted. That is the same state as having no project at
     // all, and it is what the dependency lane will conclude too when its build
-    // fails and the file falls through to the orphan type-strip.
+    // fails and the file falls through to the isolated orphan emit.
     moduleOptionsCache.set(tsconfig, null);
     return null;
   }
@@ -1203,9 +1201,8 @@ export function projectModuleOptions(
 /**
  * Resolve the JavaScript to run for a TypeScript source file, in priority
  * order: the entry project's pre-built emit (transform plugins applied), a
- * built raw `.ts` dependency, or — when no tsconfig owns it — a `mode:
- * "transform"` type-strip. Shared by the ESM `load` hook and the CommonJS
- * `require` handler.
+ * built raw `.ts` dependency, or an isolated emit when no tsconfig owns it.
+ * Shared by the ESM `load` hook and the CommonJS `require` handler.
  */
 function resolveServedSource(
   filename: string,
@@ -1252,7 +1249,8 @@ function resolveServedSource(
  * map's `sources`, so the JavaScript executed under the `.ts` source URL stays
  * self-describing after the per-run emit directory is deleted. Applied to both
  * the entry lane (`serveEntryEmit`) and the dependency lane
- * (`serveBuiltDependency`); the orphan type-strip lane carries no emitted map.
+ * (`serveBuiltDependency`); the orphan lane inlines its own map before
+ * caching.
  */
 function withInlineSourceMap(served: ServedSource): ServedSource {
   const source = inlineServedSourceMap(
@@ -1337,6 +1335,13 @@ function emitOrphanSource(
         // each one would otherwise pay a full single-file check.
         "--noCheck",
         "--skipLibCheck",
+        // This cache owns one source, not its dependency graph. Do not inline
+        // imported const enums or emit other files (which can share its stem).
+        // Isolated modules also retain each const enum's runtime definition.
+        "--noResolve",
+        "--isolatedModules",
+        "--sourceMap",
+        "--inlineSources",
         "--outDir",
         outDir,
         "--listEmittedFiles",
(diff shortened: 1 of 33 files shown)


samchon (commented):
Overall Self-Review round 1 on c2e300056af6021f4372210e6aa6b21131f18e9c: changes required.

Inspected the complete master-to-head diff, all 13 changed files, entry/register/dependency/orphan callers, compiler argument and native-host propagation, module classification, cache generations, source-map serving, test isolation, package delivery and ordinary CI selection.

The new ESM orphan compiler path can inline an imported const enum while its cache keys only source bytes, format, and compiler path. A shared cache across two projects with identical importing source but different enum values prints 1 for both projects instead of 1 then 2. Orphan lowering must be isolated from transitive values, and retained maps must identify the actual source. The module-level runtimeHooks documentation also still describes the replaced ESM stripping path. These will be repaired together before a new full review round.

Local verification: package build and test-ttsc typecheck pass; all added feature tests and native strip composition pass. Full local runtime suite has 112/114 passing; its fake JavaScript tsc.exe and file symlink fixture fail under this Windows environment before entering changed behavior. pnpm format and git diff --check pass. Ordinary CI is running and is not yet an acceptance result.

samchon (commented):
Individual Self-Review for c2e300056af6021f4372210e6aa6b21131f18e9c (read-only advisory reviewer), adjudicated by the campaign lead.

This commit implements #1359 across project, preload, dependency, and orphan execution and adds decorator regression coverage.

Accepted findings:
- runtimeHooks.ts:1346: an excluded scripts/index.ts importing scripts/internal/index.ts can receive the internal file's emit because both have the same basename. Lead reproduction also prints only 42 and skips all decorator effects. Isolated single-file emit will close this path together with the separately reproduced imported-value cache defect.
- runtimeCompilerArgs.ts:14-41: response-file arguments are not reflected in effective target selection. A response file selecting ESNext from ES2022 still gives Node raw decorators; one selecting ES2019 from ESNext is overwritten by ES2025. Use the compiler's own response-file parsing and add both precedence directions.

The reviewer inspected the complete parent-to-commit diff and found no additional defect in direct CLI aliases/casing/repetition/null handling, default/explicit libraries, noLib, implied module, entry/preload/dependency/native-plugin paths, or the added tests and documentation. These findings are advice, not the Overall Self-Review gate. Repairs and a fresh complete Overall round remain required.

samchon (commented):
Overall Self-Review round 2 on 6c08c29170c2d94d378e60b98b4be614cf64b376: clean.

Performed a fresh complete review of the entire master-to-head diff: all 17 changed files and their callers, compiler/native-host argument propagation, registered and excluded entry preparation, module-format selection, dependency and orphan cache state, source maps, recovery and cleanup, Windows/POSIX paths, packaging, documentation, and validation selection. No additional sound improvement survived verification.

Confirmed the first-round findings are resolved: response-file target precedence is read by the selected compiler; isolated orphan compilation cannot emit another source or inline an imported const enum; cached maps include the actual source path and module format; old runtime dependency output has a distinct cache identity. The type gate, explicit lower targets and legacy decorators remain covered, and ordinary ttsc output is unchanged.

Validation on this implementation: ttsc package build and test-ttsc typecheck pass; all 12 decorator feature/native-plugin tests pass; response-file error/precedence and same-format cross-project cached-map boundaries pass. The complete local runtime corpus has 116/118 passing, with only the documented pre-existing Windows fake-executable and file-symlink fixture limitations. pnpm format and git diff --check pass. Required ordinary CI and the correction commit's Individual Self-Review are still pending; this clean review alone does not authorize merge.

packages/ttsc/src/launcher/internal/runtimeHooks.ts
```diff
@@ -1340,14 +1335,28 @@ function emitOrphanAsCommonJs(filename: string): string | null {
         // each one would otherwise pay a full single-file check.
         "--noCheck",
         "--skipLibCheck",
+        // This cache owns one source, not its dependency graph. Do not inline
+        // imported const enums or emit other files (which can share its stem).
+        // Isolated modules also retain each const enum's runtime definition.
+        "--noResolve",
```
samchon: The new isolatedModules policy preserves const enums at runtime, while emitCommonJsForNameScan erases them. A CommonJS orphan export-star barrel therefore omits the enum from its ESM namespace. Share the same isolated emit policy with export discovery; preserve type-only export elision.

samchon (commented):
Individual Self-Review for 6c08c291, adjudicated by the lead: one accepted finding. Isolated orphan emission fixes cached imported values and source ownership, but CommonJS export-name discovery still uses a different emit policy. The lead reproduced a decorated orphan re-exporting a const enum: dep.Value is undefined while dep.default.Value.Entry is 42. All four new regressions passed for the reviewer. No other confirmed finding. This invalidates the prior clean merge gate; a shared isolated emit policy, regression coverage, and a fresh Overall Self-Review are required.

samchon (commented):
Overall Self-Review round 3 on 6c08c291: changes required after the correction commit's advisory review. Re-read the full 17-file master-to-head diff and its consequences. Reproduced the CommonJS const-enum export-star mismatch independently. No other accepted finding in this round. Runtime emission and export-name discovery must share one isolated emit policy. The earlier clean round is superseded. Separately, ordinary CI's typecheck lane is blocked before typechecking by unchanged repository dependency advisories; remediation affects website/tooling dependencies outside the user's declared campaign scope, so that scope boundary remains pending rather than being silently widened.

packages/ttsc/src/launcher/internal/runtimeHooks.ts
```diff
@@ -1386,13 +1393,9 @@ function emitCommonJsForNameScan(filename: string): string | null {
...
       [
         real,
-        "--ignoreConfig",
         "--module",
         "commonjs",
-        "--target",
-        "es2022",
-        "--noCheck",
-        "--skipLibCheck",
+        ...ISOLATED_EMIT_ARGS,
```
samchon, line 1414: An orphan barrel can re-export a source owned by another tsconfig. If its project sets preserveConstEnums:false, isolated scanning invents a Value export erased by the actual project emit. Determine names from actual owned output; keep isolated emission for unowned sources.

samchon (commented):
Individual Self-Review for b464aec0, adjudicated by the lead: one accepted finding. The shared isolated policy is correct for unowned sources, but export discovery also reaches files with an owning project. The lead reproduced a decorated orphan barrel re-exporting a project that erases const enums: both namespaces gained an undefined Value export. The correction must use the owning project actual emit when available and isolate only the fallback. New regression passed; no other confirmed finding. The full Overall round 4 also re-read the entire 18-file diff and accepted this ownership mismatch; it is not a clean merge gate.

samchon (commented):
Overall Self-Review round 4 on b464aec0: changes required. Re-read the full 18-file base-to-head diff and the source ownership, actual emit, CommonJS discovery, cache and module consumers. Accepted and independently reproduced the advisory ownership mismatch: a decorated orphan barrel re-exporting an owned source whose project erases const enums receives a ghost Value export. Names must come from actual project-owned output whenever available; only unowned source falls back to isolated compilation. No other finding survived this round. The dependency-audit blocker remains outside authorized scope pending the user's answer.

packages/ttsc/src/launcher/internal/runtimeHooks.ts
```diff
@@ -1373,6 +1383,14 @@ function emitCommonJsForNameScan(filename: string): string | null {
   if (cached !== undefined) {
     return cached;
   }
+  const served = serveEntryEmit(real) ?? serveDependencyEmit(real);
```
samchon, line 1387: Owned output may rewrite ./nested.ts to ./nested.js. Source-side recursion must use the same JavaScript-to-TypeScript rescue as the runtime resolver, or nested exports disappear.

packages/ttsc/src/launcher/internal/runtimeHooks.ts
```diff
@@ -1373,6 +1383,14 @@ function emitCommonJsForNameScan(filename: string): string | null {
...
+  if (
+    served !== null &&
+    moduleFormat(real, served.moduleOptions) === "commonjs"
+  ) {
+    commonJsNameScanSources.set(real, served.source);
+    return served.source;
```
samchon, line 1393: Owned importHelpers output calls tslib_1.__exportStar. Both discovery and Node named-export hinting must recognize that compiler form. Preserve the original helper call so incomplete static discovery can never delete runtime values.

samchon (commented):
Individual Self-Review for cb8ea658, adjudicated by the lead: accepted two recursive export-discovery regressions. Owned output can rewrite a TypeScript specifier to JavaScript or call tslib-qualified export helpers; both shapes were lost by consumers expecting the former isolated output. The outer runtime rewrite then discarded actual exports omitted by static discovery. Overall round 5 re-read all 19 changed files and consumers and confirmed that static name hints must never replace the original runtime export-star operation. Repair both discovery forms and preserve runtime semantics when discovery is partial. Node 24 full local runtime suite has only the two known Windows fixture limitations; new project-owner matrix passed. Existing dependency audit still blocks CI outside requested scope.

samchon (commented):
Overall Self-Review round 5 on cb8ea658: changes required. Complete fresh 19-file diff and consequence inspection accepted rewritten-specifier and imported-helper recursion failures, plus the underlying destructive coupling between static name discovery and runtime export-star execution. Lead regression independently reproduces a computed CommonJS export being deleted while the known export survives. The repair keeps the original runtime operation and adds inert static name hints, shares the helper-call grammar with recursive discovery, ignores inert source text, and reuses runtime JavaScript-to-TypeScript rescue. Regression needs the real tslib helper, so add tslib 2.8.1 only to the ttsx test package (existing locked package, no unrelated dependency version changes). All findings are one combined repair; no extra issue is published.

packages/ttsc/src/launcher/internal/runtimeHooks.ts
```diff
@@ -1528,10 +1543,17 @@ function exposeCommonJsStarExports(
     return source;
   }
   const reserved = collectStaticCommonJsExportNames(source);
-  let index = 0;
+  const executable = maskCommentsAndStrings(source);
```
samchon: A regular expression matching a backtick is valid JavaScript but this masker treats it as a template delimiter. Subsequent real re-exports vanish from ESM namespace discovery. Use a proper JavaScript lexer for both direct names and recursive re-exports.

samchon (commented):
Individual Self-Review for 181b6783, adjudicated by the lead: accepted one lexical regression. The handwritten masker mistakes a backtick inside a regular-expression literal for a template opener, hiding following real re-exports. Lead reproduced undefined 17 instead of 17 17 with the regex in either the decorated outer file or an owned middle file. Full Overall round 6 inspected the complete 25-file diff and its consumers. The durable repair replaces handwritten masking/pattern detection with cjs-module-lexer, the lexer used for this purpose by Node, and appends inert name hints without rewriting source statements. Package capability probes verified inline/imported helpers, regex literals, template lookalikes, and synchronous parsing.

samchon (commented):
Overall Self-Review round 6 on 181b6783: changes required. Re-read the complete 25-file base-to-head diff, metadata/CI/packaging and runtime consumers. Accepted the independently reproduced regular-expression/template-state confusion. Handwritten lexical masking is the wrong owner for JavaScript grammar. Replace it and helper-pattern parsing with the Node-maintained cjs-module-lexer; keep original execution byte positions by appending inert metadata. The production parser dependency is necessary for ttsx correctness, has no transitive dependencies, and package probes confirm both CommonJS synchronous and ESM synchronous initialization paths. Add regex witnesses in outer/middle files and pin source maps with re-exports present. Unrelated audit remediation is still not authorized.

packages/ttsc/src/launcher/internal/runtimeHooks.ts
```diff
@@ -1516,48 +1531,37 @@ function pickEmittedJavaScript(
...
-          name !== "__esModule" &&
-          isIdentifierName(name) &&
-          !reserved.has(name),
-      );
-      if (names.length === 0) {
-        return statement;
-      }
-      for (const name of names) {
-        reserved.add(name);
+  const parsed = parseCommonJsExports(source);
```
samchon, line 1543: Keep prior block/IIFE export-star detection in addition to Node lexer metadata. The source still exports actual=17, but the new namespace misses that binding solely because lexer reexports is top-level-only. A full AST walk can preserve the existing helper forms without lexical ambiguity.

samchon (commented):
Individual Self-Review for 8f458227, adjudicated by the lead: accept the compatibility candidate. Node lexer deliberately omits star helpers inside blocks/IIFEs, while the previous scanner exposed those names. Even though normal TypeScript-Go export-star output is top-level (and current module options exclude UMD), existing executable helper forms should remain compatible. The lead reproduced the failure with real tslib helpers. Supplement Node metadata with a full JavaScript AST search for those calls; do not restore handwritten token masking or alter runtime execution. Other regex, inert-text, cache, source-map, package and syntax checks found no further confirmed defect. Node24 full runtime suite:123/125 pass, only the two previously documented Windows fixture restrictions. Native plugin composition passes.

samchon (commented):
Overall Self-Review round 7 on 8f458227: changes required. Completed a fresh review of all 27 changed files and their full consequence surface. Accepted the independently reproduced nested-helper compatibility difference rather than dropping an existing executable source form. Keep Node lexer metadata, supplement it with AST call discovery for literal __exportStar(require(...), exports) at every scope, and retain the append-only nonexecuting hints. Reuse already locked Acorn 8.16.0, which has no transitive dependencies. A new matrix covers top level, true blocks, IIFEs and class static blocks in both outer and owned middle files. All other accepted findings remain resolved. The master baseline is still439a40a; unrelated dependency-audit changes await user scope approval.

samchon (commented):
﻿Individual Self-Review for 1ba032f0511a5e5223bf408a20483363d0eaa636, adjudicated by the lead: clean. This correction preserves the existing scoped CommonJS star-helper names with supplemental Acorn syntax parsing while retaining Node lexer metadata and unchanged runtime statements. No new issue is claimed; this completes the re-export compatibility consequence of #1359.

The read-only reviewer inspected all five parent-to-commit files and their metadata consumers, recursive discovery, parse-error recovery, and production dependency wiring. Independent validation passed the eight nested-scope integration combinations, eleven parser-boundary cases plus invalid-syntax fallback, and diff whitespace checks. The lead accepts the clean result. The package build, twenty selected runtime tests, test-package typecheck, frozen install, and full pnpm format also passed on the integrated source.

This advisory commit review does not replace the solo Overall Self-Review or ordinary CI acceptance.

samchon (commented):
﻿Overall Self-Review round 8: clean for base 439a40a149e8008008b96d488f4c5360f27c3707 through head 1ba032f0511a5e5223bf408a20483363d0eaa636. The lead performed a fresh solo review of every changed file and hunk across all 29 files, then reconciled the latest Individual review. No candidate requiring a change survived verification.

The review covered configured and response-file target precedence; implicit module/library preservation; direct, excluded, register, dependency and orphan execution; upstream decorator/class-field semantics; owned versus isolated export discovery; const-enum and imported-value cache invariants; regex/template/scoped-helper parsing; unchanged runtime export values and conditional execution; maps after cache reuse and temporary-directory removal; Windows/POSIX paths and atomic cache writes; compiler/plugin argument propagation; dependency and package publication wiring; all positive, negative, boundary and regression tests; documentation and ordinary CI selection. The scope remains exactly #1359 and its ttsx consequences; no additional issue was opened.

Validation on the final source: ttsc package build, test-ttsc typecheck, frozen dependency install, full pnpm format, and twenty selected runtime tests pass. The latest individual review independently passed the nested-scope integration matrix and parser boundaries. Native plugin composition and the original report on Node 22/24/26 passed before the last supplemental-parser correction. The preceding full Node24 runtime run passed 123 of 125 cases; the remaining two are existing Windows fixtures requiring an executable shell-script stand-in and file-symlink privilege, respectively. These limitations are recorded, not treated as passes.

CI remains an independent blocked gate. Every ordinary check except the existing dependency-audit/typecheck lane and its aggregate passed on the previous head. Final-head CI is running. A direct final-head pnpm check:dependencies reproduces the same sixteen blocking advisories in unchanged website/tooling dependencies. The user's explicit ttsx-only scope does not authorize those unrelated updates; the scope question is pending. Do not merge, mark #1359 closed, or call the campaign complete while this gate remains blocked. No dependency-audit waiver or acceptance check was weakened.

samchon (commented):
﻿Individual Self-Review for 3e777370b4c50d890b79301081b31195c9a84302, adjudicated by the lead: clean. The correction resolves the existing CI dependency-audit block in this same #1359 campaign PR, as explicitly directed by the user. It updates Next and the affected transitive parsers/image library, and discloses the additional unpatched extract-zip advisory through the existing automatically expiring waiver policy.

The read-only reviewer inspected all five parent-to-commit files, the complete generated lockfile, Next/SWC and sharp/libvips platform alignment, package consumers, and waiver expiry. Independent pnpm why and package metadata checks confirmed the website-only Marp/Puppeteer browser-downloader path and no install-script download; the repository uses HTML slide output. The reviewer verified the primary advisory and current npm release, passed all fifteen audit tests, reproduced the passing real audit (high=2, critical=0, waived=2), and checked whitespace and an unchanged working tree. No actionable candidate was reported; the lead accepts the result. This advisory review does not replace Overall review or ordinary CI.

samchon (commented):
﻿Overall Self-Review round 9: clean for base 439a40a149e8008008b96d488f4c5360f27c3707 through head 3e777370b4c50d890b79301081b31195c9a84302. The lead performed a fresh solo review of every changed file and hunk in all 33 files, including the complete generated lockfile diff. No finding requiring a correction survived verification.

This round covered all runtime preparation and emit paths, target/module/library precedence and response files, upstream decorator/class-field semantics, compiler and source-plugin argument propagation, isolated versus owned re-export metadata, imported values and cache identities, parse-error recovery and syntax boundaries, conditional execution and live CommonJS exports, source maps and temporary output lifetime, Windows/POSIX resolution, recursive cycle guards, package dependencies and published declarations, tests and documentation. The CI repair was also traced through Next/Nextra, all sharp/libvips and Next/SWC platform packages, xmldom and js-yaml consumers, the current audit graph, unchanged audit failure/count rules, and the independent expiry of each no-fix waiver. No unrelated issue-discovery campaign was opened.

Local validation: real dependency audit passes (high=2, critical=0, waived=2), all 15 audit tests pass, the 71 CI helper tests pass, all 16 unplugin unit cases pass, every test-project typecheck passes, frozen install passes, the website slide build and Next production build (103 static pages) pass, and full pnpm format passes. The existing runtime validation remains twenty selected regressions, the original report on Node22/24/26, and native plugin composition. The preceding full local Node24 runtime corpus had two documented Windows fixture restrictions; they were not counted as passes. The final source's ordinary CI independently tests the full runtime and native suites.

The user's instruction now explicitly includes all CI repairs regardless of original scope. Next is locked to16.3.4 (manifest requires at least16.3.3), xmldom0.9.12, sharp0.35.4 and js-yaml4.3.2 remove the fixable advisory set. extract-zip2.0.1 has no released fix for GHSA-7pqw-9j4j-h8q3, as confirmed by npm and https://github.com/advisories/GHSA-7pqw-9j4j-h8q3. Its only dependency path is the website's Marp/Puppeteer browser downloader; the repository slide build emits HTML and does not extract attacker-provided browser archives. The existing no-fix waiver mechanism now names both extract-zip advisories, stops waiving either as soon as its patched_versions changes, and still rejects all other severe advisories and unnamed severe counts. These remaining advisories are disclosed, not represented as repaired upstream vulnerabilities.

Remote typecheck job102389361245 already passes on this exact head, including every step previously blocked by the audit. Remaining ordinary checks are running and remain an independent merge gate.

samchon (commented):
﻿CI acceptance update for head 3e777370b4c50d890b79301081b31195c9a84302: the user explicitly excludes nestia.yml from this campaign's completion/merge decision. Its result will not be represented as passing or included in the accepted check set. All other ordinary checks still must finish green on this exact head. This changes no workflow, check implementation, source, test, or recorded Self-Review result.

samchon (commented):
﻿Final CI acceptance for immutable head 3e777370b4c50d890b79301081b31195c9a84302: all 27 checks in the user-authorized acceptance set pass, including aggregate ci, typecheck, ttsc core/native, all six platforms, Windows Go, lint, evidence, graph, packaging, bundlers, website, and external build/typia checks. nestia.yml is excluded by the user's explicit instruction and is not counted as a success.

Individual Self-Review for the final correction and fresh solo Overall Self-Review round 9 are both recorded clean on this exact head. The final CI core log confirms all nineteen added decorator runtime tests executed successfully; native coverage is included in the passing native lane. The only closing keyword in the surviving commit messages targets #1359, whose fix remains present. Merge gates are satisfied under the user's campaign authorization.

PR #1381, opened 2026-09-17
fix(unplugin): make correctness and resource bounds permanent

### LINKED ISSUE #1368: ttsc 0.30 breaks React Router 8 app
When I updated my RR8 app to use ttsc 0.30, I started getting a bunch of these errors:

```
1:54:39 PM [vite] Internal server error: Server-only module referenced by client

    '/Users/vivek/Work/myapp/myapp-app/app/lib/app_permissions.server.ts' imported by 'app/lib/types.ts'

  See https://reactrouter.com/explanation/code-splitting#removal-of-server-code

  Plugin: vite:import-analysis
  File: /Users/vivek/Work/myapp/myapp-app/app/lib/types.ts:1:0
  1  |  import * as _validateReport_1 from "typia/lib/internal/_validateReport";
     |  ^
  2  |  import * as _createStandardSchema_1 from "typia/lib/internal/_createStandardSchema";
  3  |  /**
      at ResolveIdContext.resolveId (file:///Users/vivek/Work/myapp/myapp-app/node_modules/@react-router/dev/dist/vite.js:1923:11)
      at async EnvironmentPluginContainer.resolveId (file:///Users/vivek/Work/myapp/myapp-app/node_modules/vite/dist/node/chunks/node.js:8088:19)
      at async TransformPluginContext.resolve (file:///Users/vivek/Work/myapp/myapp-app/node_modules/vite/dist/node/chunks/node.js:8323:13)
      at async normalizeUrl (file:///Users/vivek/Work/myapp/myapp-app/node_modules/vite/dist/node/chunks/node.js:26050:22)
      at async Promise.all (index 0)
      at async TransformPluginContext.transform (file:///Users/vivek/Work/myapp/myapp-app/node_modules/vite/dist/node/chunks/node.js:26209:25)
      at async EnvironmentPluginContainer.transform (file:///Users/vivek/Work/myapp/myapp-app/node_modules/vite/dist/node/chunks/node.js:8172:14)
      at async loadAndTransform (file:///Users/vivek/Work/myapp/myapp-app/node_modules/vite/dist/node/chunks/node.js:19634:26)
```

I used claude to help track it down and it came up with this explanation:

The source import type … from './app_permissions.server' still gets stripped from types.ts. The problem is a change in how ttsc 0.30 tells Vite which files to watch:

1. ttsc 0.30 now calls addWatchFile() on the files a transform's types came from, so an edit to them re-runs the transform. For types.ts that's 8,477 files, and the first is app/lib/app_permissions.server.ts, where the UserPermissionSet type behind user_permission_roles_validator is defined.
2. Vite dev treats every watched file as an import of that module and resolves it when the module is served.
3. React Router blocks any .server file resolved for the client, so /app/lib/types.ts fails with a 500.

That's why it only appears when a browser loads a page: the client version of the module is only built on request. 0.26.2 didn't send those watch files through, so it worked.

I was then able to have it make a wrapper around the ttsc plugin to vite to drop server-only files from ttsc's watch list, but only for client requests. It uses the same file pattern as React Router's check. Server-side requests still get the full list. The drawback is you have to refresh a screen to notice some of those type changes.

At the same time, Claude also added to this wrapper a test to skip transforming files that do not include "typia", and this brought my test suite run time down from ~60 seconds to ~3.5 seconds.

ttsc@0.30.2 and typia@14.0.6

### LINKED ISSUE #1378: fix(unplugin): bound shared transform watchers across every lifecycle
## Problem

The shared `@ttsc/unplugin` transform cache opens filesystem watchers in proportion to project shape, independently of the Vite-specific watcher that caused #1368.

`createProjectMutationTracker` opens one `fs.watch` for every directory found by the project walk. The universal-input and absent-candidate trackers open further watches for distinct parent directories. This happens whenever a transform cache exists, including build-scoped Rollup, Rolldown, webpack, Rspack, Farm, esbuild, and Bun builds. Those hosts already re-prove the complete generation on the first delivery of each build pass, so their transform-core watchers add startup work and native resources without contributing to the pass proof. Persistent Vite, Metro, and Turbopack workers do need notification evidence, but their resource count must not grow with the number of project directories or compiler inputs.

This is the same failure class as the reopened #1368 at a lower layer: fixing only Vite's private Chokidar graph would leave another cardinality-dependent observer in every persistent consumer.

## Required correction

- Build-scoped adapters own zero transform-core filesystem watchers.
- Persistent consumers share a bounded observer whose native subscription count is independent of project directory and compiler-input cardinality.
- The observer preserves content changes, project membership changes, missing-path recovery, symlink/junction retargeting, atomic replacement, watcher failure fallback, concurrent generation capture, and deterministic disposal.
- Repeated server restart, overlapping replacement, failed generation, cache eviction, and final close leave no observer, timer, broker, or cache resource behind.
- Existing correctness proofs stay conservative on filesystems where reliable recursive notification cannot be established.

## Permanent proof

Add deterministic resource tests that scale synthetic directory/input counts by orders of magnitude and assert a constant watcher bound, zero build-scoped watchers, exact invalidation, and complete cleanup. Exercise the shared contract on Linux, macOS, and Windows; keep real-host lifecycle tests for every public adapter.

Part of the reopened #1368 unplugin correctness and performance campaign.

### LINKED ISSUE #1379: fix(ttsc): make adapter subprocess launches resilient to macOS high descriptors
## Problem

Every ttsc subprocess boundary reachable from an unplugin adapter can fail on macOS when the parent already owns high-numbered file descriptors.

The reporter's first visible failure is CommonJS descriptor evaluation, which opens a diagnostics file in the parent and passes that numeric descriptor to `spawnSync`. The same pattern exists in native compiler/plugin execution and Go source-plugin builds; runtime-capability probing uses piped `spawnSync` output. libuv issue [#5204](https://github.com/libuv/libuv/issues/5204) reproduces `uv_spawn` returning `EBADF` on macOS when a stdio source descriptor exceeds 10,239, including on recent Node/libuv releases. Removing #1368's descriptor pressure prevents the ordinary trigger but does not make these process APIs correct when the surrounding dev server or another plugin owns many descriptors.

## Required correction

- Harden runtime capability probing, isolated plugin-descriptor evaluation, native compiler/plugin execution, and Go tool execution under high descriptor pressure.
- Preserve unlimited file-backed stdout/stderr capture, exit status, signal, launch error, environment, cwd, descriptor retry behavior, and current diagnostic/retry semantics.
- Keep the ordinary low-descriptor path direct and allocation-efficient. Any broker/fallback cost must be conditional on the affected state or an actual `EBADF`, so normal builds gain no extra process.
- Use no shell on POSIX and preserve the existing Windows Go-wrapper contract.
- Clean every request, result, output, and temporary artifact after success and every failure mode.

## Permanent proof

Add a deterministic injected-`EBADF` contract for every boundary plus a real macOS test that raises the descriptor number above the libuv threshold and successfully evaluates a descriptor, builds a source plugin, and launches the native transform. Run the process contract at the supported Node floor and current CI line where practical.

Part of the reopened #1368 unplugin correctness and performance campaign.

### LINKED ISSUE #1380: perf(unplugin): enforce complete host and resource contracts in CI
## Problem

The unplugin suite proves functional output across every published adapter, but it does not make steady-state resource and performance bounds mandatory. The repository's own `experimental/unplugin-perf` harness is not selected by CI and currently fails on clean `master`.

For a 50-module persistent generation with 50 shared external inputs and 50 global inputs, the harness measured per delivered module:

- 127.5 file reads;
- 256 `lstat` calls;
- 101 `stat` calls;
- 588.5 total filesystem calls;
- 6,962 ms for the unchanged delivery pass.

The producer-complete form still averaged 54 reads and 267.9 filesystem calls per module. This repeats proof of the same shared closure for sibling modules and explains why a large Vite/React Router project can spend tens of seconds in unchanged delivery. The suite also has no macOS high-cardinality resource lane, so #1368's descriptor regression passed all functional host contracts.

## Required correction

- Amortize shared input proof across one stable generation. An unchanged module delivery may pay for its own unique inputs, not re-read the shared graph closure.
- Preserve same-tick rewrite, atomic replacement, missing candidate, physical-path, link retarget, failed watcher, and concurrent edit correctness.
- Keep one whole-project compile for an unchanged generation and one new compile for a relevant edit.
- Promote deterministic filesystem-call, watcher-count, compile-count, teardown, and elapsed-time scaling contracts into ordinary CI.
- Retain direct installed-package execution for every public adapter: Vite 7/8, React Router, Rollup, Rolldown, esbuild, webpack, Rspack, Farm, Next webpack/Turbopack, and Bun build/runtime.
- Add a narrow macOS real Vite/React Router high-cardinality session that measures descriptor growth, active resources, memory, startup, unchanged delivery, relevant edit latency, recovery, and shutdown. It must use public host events rather than fixed sleeps.
- Cover shared process/watch contracts on Linux, macOS, and Windows without duplicating the complete expensive host installation on every platform.

## Acceptance bounds

The tests must fail for linear native descriptor growth, whole-project polling, repeated compilation, shared-closure re-reading per module, leaked resources, or unbounded elapsed-time scaling. Budgets should use operation/resource counts as the deterministic primary oracle and wall time only as a generous regression ceiling.

Part of the reopened #1368 unplugin correctness and performance campaign.

samchon:
﻿Issue #1368 exposed a resource bug in the 0.30.3 Vite fix: compiler-only inputs were correctly removed from Vite's runtime module graph, but the replacement opened one native subscription per input. A React Router project with roughly 8,477 compiler inputs could exhaust Darwin's `posix_spawn` descriptor range, so the next isolated plugin-descriptor launch failed with `EBADF`. The same audit found two broader causes: adapter caches mixed project-wide and per-input proof ownership, and CI exercised adapter facades more deeply than the installed hosts that own their real lifecycle.

This campaign replaces cardinality-dependent resources with explicit bounded contracts:

- Vite serve owns one recursive project observer and at most 16 external scopes, with a fair 64-entry fallback scheduler. It covers compile/subscribe races, missing paths, directory membership, case-insensitive event spellings, short paths, symlinks/junctions, hardlinks, destination-only renames, observer failure, removal, and deterministic disposal.
- Build-scoped captures use one temporary project observer only while compiling and close it before delivery. Persistent generations share project and external proofs without allowing one unrelated external change to invalidate the whole program.
- Project-walk, graph, host, and resolution-candidate evidence have separate authority. Unreadable, missing, symlinked, hardlinked, aliased, and out-of-walk inputs cannot borrow watcher silence they did not earn; watcher-silent hardlink writes fall back to metadata validation without adding steady-state work for ordinary files.
- Every adapter-reachable synchronous process launch keeps the normal direct path and retries only Darwin/POSIX `EBADF` through a shell-free low-descriptor broker. The broker preserves argv, cwd, environment, timeout, signal/status, output, and missing-command errors. The Go cache heartbeat no longer allocates an IPC descriptor.

The permanent verification surface now includes:

- 216 deterministic core and adapter cases, including resource cardinality, all mutation classes, failure/recovery, teardown, Windows 8.3/case behavior, and a real native envelope;
- a packed clean-consumer contract that directly drives Vite 7, Vite 8, React Router 8, Rollup, Rolldown, esbuild, webpack, Rspack, Farm, Next webpack, Next Turbopack, Bun build, and Bun runtime preload;
- a macOS CI lane that opens more than 10,500 descriptors before a real transform and proves successful compilation and cleanup;
- a mandatory performance gate that requires one compile across unchanged passes, exactly zero sibling graph identity probes (including existence checks) through 10,099 edges, zero source reads per unchanged Vite delivery, and at most one total filesystem call per delivery.

Closes #1368
Closes #1378
Closes #1379
Closes #1380


samchon (commented):
Individual Self-Review — `0f8f5f201`

Reviewed the complete `7b571f850..0f8f5f201` diff directly because this campaign is explicitly solo.

The commit replaces per-input Chokidar subscriptions with one recursive project observer plus at most 16 external scopes, closes the compile-to-subscribe race with a bounded change journal, preserves missing/predicate/link recovery, and removes the direct Chokidar dependency. The 12,000-input resource case proves one active native observer, deterministic teardown, and a five-second linear-work budget; the real boundary case covers deletion races, junction retargeting, and predicate transitions.

Review result: no unresolved correctness or performance finding in this commit. Narrow verification passed:

- `pnpm --filter @ttsc/unplugin build`
- `pnpm --filter @ttsc/test-unplugin unit --include=vite_compiler_watch`

samchon (commented):
Individual Self-Review — `fcd03ab97`

Reviewed the complete `0f8f5f201..fcd03ab97` diff directly under the campaign's explicit solo constraint.

The commit removes all core filesystem watchers from build-scoped adapter caches while retaining the once-per-pass complete snapshot proof. It also makes every capture-based subprocess boundary recover from Darwin's high-descriptor `EBADF` without changing the normal direct-spawn path, removes the IPC descriptor from the Go-build heartbeat fallback, and adds a dedicated macOS lane that drives a real transform above descriptor 10,500.

Review result: no unresolved finding. The fallback preserves cwd, environment, timeout, signal/exit status, argument boundaries and unbounded file-backed output; Windows retains its native command-wrapper path. Verification passed:

- `pnpm --filter ttsc build`
- `pnpm --filter @ttsc/unplugin build`
- the watcher-free build-pass integration case
- validation planner and unplugin contract tests

The high-descriptor branch is intentionally exercised by the new macOS CI lane, since Windows does not implement Darwin's `posix_spawn` descriptor ceiling.

samchon (commented):
## Individual Self-Review — `d6aeee1ca`

**Verdict:** follow-up required and completed in `2a8e7fb6f`.

The resource-contract commit correctly added the cross-platform watcher, subprocess, fallback-budget, and CI coverage, but a fresh consequence review found four remaining gaps:

- build-scoped compilation did not retain a temporary project event witness, so a graph-free in-project A→B→A mutation could restore identical pre/post snapshots;
- external graph inputs were not included in the shared retained observer, causing shared closures to repeat validation I/O for every sibling delivery;
- removed Vite entries could leave component-resolution memo keys and empty importer records alive;
- the fallback scheduler test spent twice the work needed to prove its 64-item budget.

`2a8e7fb6f` closes all four. The build lifecycle now opens exactly one bounded recursive compile-race observer and closes it before returning; persistent validation folds project, host, and external inputs into bounded shared scopes; Vite clears obsolete memo ownership; and the scheduler test proves 64+1 progress. The performance gate now measures 0 content reads and 1 total filesystem call per delivered file for the 50-module/100-shared-input scenario, down from the rejected 501 calls per file.

Validation: `@ttsc/unplugin` build passed; graph-free build ABA, external ABA, independent-leaf ABA, persistent watcher cardinality, and three Vite watcher boundary tests passed; the complete unplugin performance suite passed.

samchon (commented):
## Individual Self-Review — `2a8e7fb6f`

**Verdict:** clean.

Reviewed the full commit against correctness, lifecycle cleanup, bounded resources, cross-platform behavior, and steady-state cost. The compile-time watcher exists only when a cache can publish a generation, is one recursive observer regardless of project size, and is closed on build-scoped success, retry, failure, and cleanup error paths. Persistent external coverage excludes symbolic links, including broken links, so target changes keep their joined metadata/content proof instead of trusting an unrelated lexical directory event. Observer registration above the 16-root cap falls back to snapshot validation rather than broad filesystem watches.

The Vite removal path now drops pending entries, link ownership, empty importer maps, and path-component memos without changing live scope ownership. Exact predicate observations retain metadata signatures only around a successful, unraced replay, so the optimization cannot authorize bytes or topology it did not validate.

Validation: targeted Prettier and `git diff --check` passed; `@ttsc/unplugin` declaration and Rollup build passed; the graph-free build-scoped A→B→A test and related race/cardinality cases passed; Vite subscription, alias, resource, and 64+1 fallback-budget cases passed; all unplugin performance scenarios passed with shared serve validation at 0 reads and 1 syscall per file and repeated build passes at one native compile.

samchon (commented):
## Individual Self-Review — `d05739509`

**Verdict:** clean.

Reviewed the complete commit for subscription races, failed-watcher ownership, bounded fallback work, and teardown.

The replacement path now records events against both new and existing subscriptions, so a compile-time A→B→A mutation cannot be hidden by restoring the original bytes before a new proof is installed. An observer that fails synchronously or asynchronously relinquishes its native handle immediately, while one fair shared poller validates at most 64 entries per tick. Neither recovery path adds an input-proportional observer or timer.

The regression suite directly exercises an existing subscription's compile race and proves a failed observer closes exactly once, advances 64+1 entries across two ticks, and leaves no poller after completion. The Vite boundary and resource cases, package build, and performance suite pass with these invariants.

samchon (commented):
## Individual Self-Review — `929e33f9c`

**Verdict:** clean.

Reviewed the complete commit across event lookup, scope ownership, compile-race history, symlink-component memoization, and case-sensitive/case-insensitive filesystems.

Every watcher-owned index now uses the nearest existing directory's measured case policy. This makes event spellings and registered input spellings converge on case-insensitive volumes without folding distinct files on case-sensitive volumes. Scope counting uses the same identity, and pinned project scopes are excluded from the 16 external-scope allowance. The case-policy cache is shared and only probes an uncached directory lineage, so it does not create per-delivery graph work.

The platform-conditional creation test proves an alternate-case native event reaches the registered input on case-insensitive filesystems. The Vite boundary, alias, lifecycle, cardinality, and performance cases pass with the normalized identity path.

samchon (commented):
## Individual Self-Review — `b2f020933`

**Verdict:** clean.

Reviewed the complete commit against capture races, filesystem aliases, unreadable and missing inputs, project/external proof ownership, observer lifetime, and steady-state cost.

Project mutation evidence now compares physical identities when native events use a different spelling, including Windows short paths. Watcher silence is withheld from symlinked ancestry, hardlinks, and inputs whose bytes capture did not prove. A closed tracker can no longer authorize reuse. Project membership watches cover only the inputs the project walk hashed; external graph inputs keep their graph/host proof, preventing unrelated `node_modules` activity from forcing a second native compile. Directory inputs match descendant events, new members are recognized even on backends that report `change`, and host content changes stay narrow instead of invalidating the whole generation.

The Vite rename path handles destination-only directory events and still requires the recorded baseline to change before invalidating. Build-scoped observers are proven temporary: one opens around each native capture and is closed before delivery, leaving none retained in the cache.

Targeted project-root, linked-input, missing/unreadable-input, external-graph, Vite-boundary, and production native lifecycle cases pass. The full performance harness remains at one compile for repeated passes, zero reads per unchanged serve delivery, one filesystem call per delivery, and zero graph probes for sibling delivery.

samchon (commented):
## Individual Self-Review — `4c962f939`

**Verdict:** clean.

Reviewed the benchmark-only commit for whether its limits express the implemented asymptotic and fixed costs without measuring incidental fixture setup.

Scenario C resets counters after the first delivery has captured and memoized the generation, so its exact zero-probe budget measures only sibling watch-input derivation. Scenarios D–F reset after the first persistent pass, then require zero source-byte reads and exactly one metadata call per unchanged module. These budgets reject a constant residual read, a repeated identity probe, and any graph- or project-cardinality scan. The existing compile-count and three-pass cases continue to require one whole-project compile.

`pnpm run experimental:unplugin-perf` passed on the committed implementation: 0 graph probes per sibling module at 2,599/5,099/10,099 edges; 0 reads and 1 syscall per module for disjoint, shared, and producer-complete Vite shapes; and 1 compile across 3 unchanged passes.

samchon (commented):
## Individual Self-Review — `e2d1d403d`

**Verdict:** clean.

The preceding review found that the exact total-call budget made two legacy per-scenario `lstat` allowances redundant and misleading. This commit removes those weaker limits and their obsolete prose, leaving one direct contract for every persistent shape: zero source reads and at most one total filesystem call per unchanged delivery.

The complete performance suite passed again after the cleanup. It measured zero sibling graph probes at 10,099 edges; zero reads and one `lstat`/one total call per module in all three Vite serve shapes; and one native compile across three unchanged build passes.

samchon (commented):
## Individual Self-Review — `84086bd1d`

**Verdict:** clean.

Reviewed the documentation commit against the implemented resource, lifecycle, fallback, and performance contracts. It removes the obsolete per-input Vite subscription claim and states the actual limits: one recursive project observer, at most 16 external scopes, one shared 64-entry scheduler, no retained build-scoped observer, zero source reads and one metadata call on an unchanged persistent delivery, and zero sibling watch-input probes. It also documents the snapshot fallback and Darwin `EBADF` retry without exposing configuration the user must manage.

The update stays on the existing bundler-user page, adds no new navigation surface, follows the page's established level of detail, and passes targeted Prettier plus `git diff --check`.

samchon (commented):
## Individual Self-Review — `a719ee3ebed59f2dc83e7b883d484be5009da005`

Reviewed the exact parent-to-commit change directly under the campaign's explicit no-subagent constraint.

Result: **clean**. No remaining correctness, lifecycle, portability, resource, or performance finding.

The review traced the four corrected consequence surfaces end to end:

- native `configFile` ownership remains uncacheable until the consuming native stage returns content and physical-identity proof, including multi-plugin overlap;
- project and Vite notification shortcuts are revoked for hardlinks, symlinks, junction traversal, failed scopes, and scope-ceiling fallback;
- Vite closes the compile-to-subscribe interval for newly discovered external scopes and bounds native scopes, polling, history, and per-tick work independently of graph size;
- steady-state validation accounts for every filesystem call and allocates no per-entry transient arrays in the large registration path.

Validation on this commit: 216-case inventory contract, focused `configFile` integration, hardlink generation regression, all three Vite watch/resource unit cases, full repository typecheck, and the strict unplugin performance harness. The measured Vite steady state is 0 reads, 0 exists, 1 lstat, 0 stat, and exactly 1 total filesystem syscall per delivered module; repeated unchanged build passes compile once.

samchon (commented):
## Overall Self-Review — Round 6 (clean)

Reviewed the complete `origin/master...a719ee3ebed59f2dc83e7b883d484be5009da005` diff: 28 files, 2,490 insertions, 357 deletions. This was a fresh whole-surface pass across process spawning, compiler proof ownership, cache validation, watcher topology, adapter lifecycles, CI routing, direct ecosystem fixtures, performance gates, and documentation.

Result: **clean**. No sound improvement or unresolved finding remains.

Earlier rounds found and corrected omitted `exists` accounting, hardlink notification authority, lexical/physical overlap, native `configFile` proof ownership, test inventory drift, Vite's external-scope registration interval, cross-plugin proof weakening, hardlink polling, and one documentation mismatch. This round rechecked the integrated consequences and the final allocation-free registration refinement. All identified failure classes now have direct regression tests or quantitative invariants, and the fallback work is constant per tick rather than proportional to the compiler graph.

samchon (commented):
## Overall Self-Review findings — adjudicated

All candidates found before the clean final round were corrected on `a719ee3ebed59f2dc83e7b883d484be5009da005`:

1. The performance harness counted reads, stats, lstat, readdir, and realpath but omitted `exists`; it now counts every filesystem operation and gates the total.
2. Project watcher silence was unsound for hardlinks and link-traversing roots; capture records link counts and removes those paths from notification authority.
3. Mutation overlap compared lexical paths; each tracker now compares through its filesystem identity context.
4. A conventional native `configFile` was registered by the JavaScript host without JavaScript content proof; proof ownership is deferred to the actual native consumer, and an absent native proof rejects the generation.
5. Deferred ownership was record-local, so one non-consuming descriptor could weaken another descriptor that read the same path; consumption is now computed across all records.
6. A newly discovered Vite external scope could claim a compile-start token from before the scope existed; scope creation advances the sequence and forces immediate baseline validation.
7. Vite events cannot see writes through an out-of-scope hardlink alias; multiply linked inputs enter the one bounded fallback scheduler.
8. The test inventory remained at 215 after adding the hardlink regression; the explicit contract is now 216.
9. The guide overstated polling for missing inputs after recursive scopes replaced it; the documented scope/fallback contract now matches the implementation.
10. Final performance review removed transient per-entry arrays from large Vite registration and universal-input proof loops.

Focused regressions and the complete quantitative harness pass. The subsequent fresh Overall Self-Review round is clean.

samchon (commented):
﻿## Individual Self-Review — `635c419fbc5a391c1627d2ee67c31b380fa25254`

Reviewed the exact `a719ee3eb..635c419fb` change directly under the campaign's explicit no-subagent constraint.

Result: **clean**. No remaining correctness, ownership, lifecycle, resource, portability, or steady-state performance finding.

The change closes the final long-lived Vite state boundary: a deleted source importer releases only its own compiler-input conditions, shared conditions survive until their final owner leaves, empty hardlink fallback work closes immediately, and server disposal discards cached physical/case identity facts before a replacement session. Rename handling preserves both pre- and post-topology path keys so a case-policy transition cannot lose the event that caused it. All removals keep poller and native-scope ownership proportional to live inputs.

Validation on this exact commit: full repository typecheck, `@ttsc/unplugin` package build, all 19 unplugin unit cases, the three Vite watch/resource aggregates, targeted formatting, and `git diff --check`. The shared-owner deletion regression proves that deleting one importer does not tear down another importer's fallback work and that deleting the final owner closes it exactly once.

samchon (commented):
﻿## Post-Round-6 finding — corrected

The continued exact-tree audit after Round 6 found one additional lifecycle class before merge: Vite kept filesystem path-identity memos after source removal and server disposal, and a deleted source importer had no adapter hook that released its private compiler-input watch ownership. In a long-lived dev process this could retain stale path facts and unnecessary fallback work across churn or restart.

Commit `635c419fbc5a391c1627d2ee67c31b380fa25254` corrects both boundaries and adds deterministic regressions for restart cache rediscovery, deleted-importer cleanup, shared ownership, exact once-only scheduler closure, and the published Vite hook surface. The direct Individual Self-Review of that commit is clean.

samchon (commented):
﻿## Overall Self-Review — Round 7 (clean)

Reviewed the complete `origin/master...635c419fbc5a391c1627d2ee67c31b380fa25254` diff: 28 files, 2,689 insertions, 359 deletions. This was a fresh whole-surface pass across process spawning, compiler proof ownership, filesystem identity, cache validation, watcher topology, importer removal, adapter lifecycles, CI routing, direct ecosystem fixtures, quantitative performance gates, and documentation.

Result: **clean**. No sound improvement or unresolved finding remains.

The final pass re-traced every long-lived collection and handle to its owner and teardown boundary. Build-scoped trackers close before delivery; Rollup, Rolldown, webpack, Rspack, Farm, esbuild, Bun, Next, Turbopack, and Vite caches end at their native lifecycle; Vite's project observer, 16 external scopes, one fair fallback scheduler, bounded event history, path-identity memos, source-importer conditions, and restart state are all explicitly bounded and released. The fast path remains quantitative: zero source reads and one total metadata syscall per unchanged persistent delivery, zero sibling graph probes through 10,099 edges, constant fallback work per tick, and one compile across unchanged passes.

All verified bug classes have direct deterministic regressions or an exact numeric invariant, and the packed-consumer CI drives every supported adapter through its installed host rather than accepting facade-only coverage.

samchon (commented):
﻿## Individual Self-Review — `4c4bfa5fa`

Reviewed the exact `635c419fb..4c4bfa5fa` correction directly under the campaign's explicit no-subagent constraint.

Result: **clean**.

The preceding implementation reset the path-identity resolver immediately on a rename while live alias and native-scope indexes still used keys from the previous resolver. If the filesystem's case policy changed at that topology boundary and the observed input itself remained valid, a later event could no longer reach the live index. This correction keeps the resolver and every active index coherent through the event, clears topology-only component memos immediately, and replaces path-identity memos atomically only when `check()` removes the affected entry. Source removal and server disposal retain their explicit reset boundaries.

The change adds no graph scan, filesystem call, handle, timer, or steady-state allocation. Targeted formatting, `git diff --check`, the production `@ttsc/unplugin` build, and all three Vite watch/resource aggregates pass.

samchon (commented):
﻿## Post-Round-7 finding — corrected

The merge-gate review found one inconsistency in the new rename-time path memo reset: it could switch the resolver's case policy while existing alias and scope maps still contained keys produced by the old policy. That creates a possible event-routing miss after a topology transition.

Commit `4c4bfa5fa` keeps those structures on one identity context until affected ownership is removed, then performs the reset at the existing atomic removal boundary. It retains immediate invalidation of linked-component topology memos and does not add work to ordinary deliveries. The exact commit's direct Individual Self-Review is clean.

samchon (commented):
﻿## Overall Self-Review — Round 8 (clean)

Reviewed the complete `origin/master...4c4bfa5fa` diff: 28 files, 2,688 insertions, 359 deletions. The round rechecked the full campaign surface and, after the Round 7 identity-context finding, traced every Vite path key from registration through native events, rename topology, condition validation, importer removal, scope closure, server restart, and disposal.

Result: **clean**. No sound improvement or unresolved finding remains.

All live Vite indexes now share one path-identity context for their complete ownership interval. Topology-only link memos clear on rename; physical/case identity memos clear atomically when affected entries leave, when source ownership is forgotten, and at terminal disposal. The correction changes no steady-state work. The wider process, cache-proof, build-scoped watcher, packed-host, CI, and quantitative performance contracts remain unchanged and were rechecked against the final tree.

samchon (commented):
﻿## Individual Self-Review — `e12f0226c`

Reviewed the exact test-only commit directly under the campaign's explicit no-subagent constraint.

Result: **clean**.

The regression makes the final identity-context invariant executable. A simulated Darwin watcher registers live indexes under a case-sensitive policy, changes the host policy at a rename boundary, and proves the watcher does not re-probe or switch the resolver underneath those indexes. Terminal disposal and replacement attachment still must re-probe, so the same case distinguishes active ownership from the intended restart reset.

The assertion uses the injected public test seam and synchronous probe counts; it adds no timing dependency, production branch, or host-private state. Targeted formatting, `git diff --check`, and all three Vite watch/resource aggregates pass.

samchon (commented):
﻿## Overall Self-Review — Round 9 (clean)

Reviewed the complete `origin/master...e12f0226c` diff: 28 files, 2,698 insertions, 359 deletions. This final round rechecked the integrated product tree and the complete executable proof surface after turning the Round 7 path-index finding into a deterministic regression.

Result: **clean**. No sound improvement or unresolved finding remains.

The final Vite contract now proves both halves of path memo ownership: live alias/scope indexes retain their resolver across topology events, while terminal disposal and a replacement server force fresh discovery. Source deletion, shared importer ownership, exact scheduler closure, bounded native scopes, compile-race history, hardlink fallback, link topology, and case-aware event routing remain covered in the same direct aggregate. The full adapter, process, cache, CI, documentation, and performance surfaces are unchanged from the clean product review.

Issue #1408, opened 2026-09-18
fix(ttsx): JSX preserved for another tool cannot run under Node


samchon:
## Problem

A project whose `tsconfig.json` sets `"jsx": "preserve"` (or `"react-native"`, which also preserves JSX) cannot run a `.tsx` file through `ttsx` or `ttsc/register`. The private runtime build keeps the JSX syntax, and Node fails at the first JSX expression with `SyntaxError: Unexpected token '<'`.

`preserve` is the default in Next.js projects, so any script there that renders or imports a component (email templates, seeders, scripts that import shared UI modules) fails under ttsx, although the type-check passes.

ttsx already has a policy for this class. When `target` is `ESNext`, TypeScript-Go keeps proposal decorators that Node cannot parse, so the runtime build lowers the target to `ES2025` without touching the user's config or output (#1359, `runtimeCompilerArgs`). JSX preserved for a downstream tool is the same situation with no lowering.

## Evidence

Reproduced on `master` @ `779c60c9` (0.30.4 source build), Windows 11, Node 24.18.0, `typescript@7.0.2`, `--no-plugins`. A local package `myjsx` provides `jsx-runtime` so no React install is needed.

| tsconfig `jsx` | Command | Result |
| --- | --- | --- |
| `preserve` + `jsxFactory: "h"` | `ttsx main.tsx` | `SyntaxError: Unexpected token '<'` |
| `preserve` + `jsxImportSource: "myjsx"` | `ttsx auto.tsx` | `SyntaxError: Unexpected token '<'` |
| `react-jsx` + `jsxImportSource: "myjsx"` | `ttsx auto.tsx` | `<div>hello</div>` |
| `preserve` + `jsxImportSource: "myjsx"` | `ttsx --jsx react-jsx auto.tsx` | `<div>hello</div>` |

No test under `tests/test-ttsc/src/features/ttsx-runtime` executes JSX.

## Consequence surface

- Entry project, out-of-include entries, dependency builds, prepared roots under `ttsc/register`, and the orphan lane (which runs with `--ignoreConfig` and therefore never sees `jsx`).
- `preserve` and `react-native` modes; `react`, `react-jsx`, and `react-jsxdev` already produce executable output and must stay unchanged.
- The JSX factory configuration a project already declares: `jsxFactory`, `jsxFragmentFactory`, `jsxImportSource` (including per-file `@jsx` / `@jsxImportSource` pragmas).
[...]

Issue #1407, opened 2026-09-18
refactor(ttsc): one documented public identity per source file


samchon:
## Problem

The TypeScript sources of `packages/ttsc` do not meet the structure discipline that `@ttsc/evidence` enforces through `evidence/singular` and `evidence/documented`:

- **singular:** a file declares exactly one public identity, and the file takes that identity's name. Barrels, which only re-export, own no identity. `index` files are exempt from the name match. Merged declarations of one name count once.
- **documented:** every public unit carries a non-empty JSDoc block. Units are exported types, functions, and variables, plus the public members of exported interfaces, object type aliases, classes, and namespaces.

ttsc hosts that lint, so it cannot depend on it for its own build, but its own code is expected to meet the same bar. In addition, the one exported declaration of a file sits directly after its imports, and its JSDoc gives a reader (human or agent) the context to use it without reading the body: purpose, invariants, and the reason behind non-obvious decisions.

The size is not cosmetic. The runtime ownership defect in #1382 lives in a 3,197-line `runtimeHooks.ts` that exports 21 declarations spanning source serving, dependency builds, a cross-process lock, module-format classification, and descriptor input recording. Its lanes could not be tested in isolation, and the stem matcher they share had no unit coverage of its own.

## Evidence

The rules were run on a copy of `packages/ttsc/src` through the real toolchain (`@ttsc/lint` with `@ttsc/evidence`, `evidence/singular` and `evidence/documented` at `error`), from `master` @ `779c60c9`. Result: 150 findings, no TypeScript errors.

`evidence/singular`, 33 files:

| File | Public identities |
| --- | --- |
| `plugin/internal/buildSourcePlugin.ts` | 30 |
| `launcher/internal/runtimeHooks.ts` | 21 |
| `launcher/internal/watchTopology.ts` | 15 |
| `compiler/internal/runBuild.ts` | 15 |
| `flags/schema.ts` | 13 |
| `internal/projectInputPathIdentity.ts` | 12 |
| `compiler/internal/sharedHostHelpers.ts` | 10 |
| `plugin/internal/loadProjectPlugins.ts` | 8 |
| `flags/parser.ts` | 7 |
| `launcher/internal/runTtscserver.ts`, `compiler/internal/residentCheckProcess.ts` | 5 |
| `plugin/internal/capabilityResolutionCache.ts`, `internal/spawnSyncResilient.ts`, `compiler/internal/residentTransformProcess.ts` | 4 |
[...]

Issue #1406, opened 2026-09-18
fix(ttsx): a solution-style tsconfig runs the entry without its referenced project's options


samchon:
## Problem

A solution-style `tsconfig.json` owns no files. It lists `"files": []` and delegates to project `references`, which is the layout Vite, Nx, and many monorepo templates generate:

```
tsconfig.json       { "files": [], "references": [{ "path": "./tsconfig.app.json" }, { "path": "./tsconfig.node.json" }] }
tsconfig.app.json   compilerOptions for src/, include: ["src"]
tsconfig.node.json  compilerOptions for vite.config.ts
```

`ttsx src/main.ts` discovers the nearest `tsconfig.json`, finds that it does not contain the entry, and compiles the entry through the out-of-include lane with the solution file's options, which are empty. The options of the project that actually owns `src/main.ts` are never applied. Depending on the code, the run either fails with a type error the project does not have, or succeeds with different runtime semantics.

TypeScript's language service resolves the same file to the referenced project that contains it, which is why the editor shows no error.

## Evidence

Reproduced on `master` @ `779c60c9` (0.30.4 source build), Windows 11, Node 24.18.0, `typescript@7.0.2`, `--no-plugins`.

```jsonc
// tsconfig.json
{ "files": [], "references": [{ "path": "./tsconfig.app.json" }] }
// tsconfig.app.json
{ "compilerOptions": { "composite": true, "target": "es2022", "module": "commonjs", "types": [],
    "skipLibCheck": true, "experimentalDecorators": true, "strict": true },
  "include": ["src"] }
```

With a three-parameter legacy decorator in `src/main.ts`:

| Command | Result |
| --- | --- |
| `ttsc -p tsconfig.app.json --noEmit` | exit 0 |
| `ttsx src/main.ts` | `TS1241: Unable to resolve signature of method decorator ... The runtime will invoke the decorator with 2 arguments, but the decorator expects 3.` |

With `function anyDecorator(...args: any[]): any { console.log("decorator received", args.length, "arguments"); }`:

| Command | Output |
| --- | --- |
| `ttsx src/main.ts` | `decorator received 2 arguments` (standard decorators, exit 0) |
| `ttsx -P tsconfig.app.json src/main.ts` | `decorator received 3 arguments` (legacy decorators) |

[...]

Issue #1405, opened 2026-09-18
fix(ttsx): persistent runtime caches serve JavaScript for stale sources and compilers


samchon:
## Problem

Two ttsx runtime caches outlive the run that wrote them, and both are keyed without the inputs that decide their content. A later run then executes JavaScript that no longer corresponds to the source or the compiler in front of it.

1. **Dependency builds without a runtime manifest.** When the runtime hooks have no manifest, dependency builds go to the shared `os.tmpdir()/ttsx-dep`, keyed only by the tsconfig path. Editing the dependency's source never invalidates it; the old program runs until the temp directory is cleared.
2. **The orphan lowering cache.** Raw TypeScript with no owning tsconfig (a package that ships `.ts` under `node_modules`) is lowered once and cached under `TTSC_CACHE_DIR` or `os.tmpdir()/ttsc-orphan`, keyed by the compiler binary's *path*. Upgrading `typescript` in a flat `node_modules` (npm, yarn) keeps that path, so the new compiler's output is never produced for an unchanged source.

## Evidence

Reproduced on `master` @ `779c60c9` (0.30.4 source build), Windows 11, Node 24.18.0, `typescript@7.0.2`, `--no-plugins`.

Dependency cache, manifest-less child:

```ts
// main.ts (the ttsx entry)
import { spawnSync } from "node:child_process";
import path from "node:path";
const env: NodeJS.ProcessEnv = { ...process.env };
delete env.TTSX_RUNTIME_MANIFEST;
const result = spawnSync(process.execPath, [path.join(__dirname, "child.ts")], { env, encoding: "utf8" });
process.stdout.write(`child said: ${result.stdout.trim()}\n`);

// child.ts
const version: string = "v1"; console.log(version);
```

`ttsx main.ts` prints `child said: v1`. Change `child.ts` to `"v2"`; `ttsx main.ts` still prints `child said: v1`.

The same state is reached without editing the environment when a process started by a ttsx program performs its first TypeScript load after the launcher has exited: the launcher deletes `runtime-manifest.json` on exit, and the runtime reads the manifest lazily on first use, so that process finds none.

Orphan cache, compiler identity: a wrapper executable at one fixed path runs the real tsgo and stamps each orphan output with `compiled by <stamp>`; `TTSC_CACHE_DIR` points at a scratch directory.

| Step | Output |
| --- | --- |
| Run with stamp A | `compiled by A` |
| Rebuild the wrapper as B at the same path, run | `compiled by A` |
[...]

PR #1410, opened 2026-09-18
fix(ttsx): exact source ownership, launcher process contract, and one-identity structure

### LINKED ISSUE #1413: fix(ttsx): Bun and Deno pass the Node gate and run without ttsx's runtime
## Problem

Bun and Deno pass ttsx's Node version gate, because both report a Node version in `process.versions.node` for compatibility. Neither implements `module.registerHooks`, so ttsx cannot run on them. It does not stop, though.

- **`bun --preload ttsc/register src/main.ts`:** dies inside ttsc with `TypeError: (0, node_module_1.registerHooks) is not a function`.
- **`bun node_modules/ttsc/lib/launcher/ttsx.js src/main.ts`:** type-checks the project, then spawns its child with `process.execPath`, which is Bun. Bun runs the entry through its own TypeScript support instead of the checked emit, and the project's transform plugins silently do not apply.

## Evidence

Windows 11, Bun 1.4.2, Deno 2.9.6, head of #1410:

| Command | Result |
| --- | --- |
| `bun ttsx.js src/main.ts` | prints the program output; no ttsx runtime hooks were installed |
| `bun --preload ttsc/register src/main.ts` | `TypeError: registerHooks is not a function` from `installRuntimeHooks.js` |
| `deno run -A ttsx.js src/main.ts` | Deno refuses the CommonJS launcher itself; `process.versions.node` reads 26.3.0 |

## Consequence surface

- The launcher's gate (`runTtsx`) and the preload's gate (`assertNodeRuntimeSupport`) share `checkNodeRuntimeSupport`.
- `bunx ttsx` runs the launcher with Node unless `--bun` is given.

## Approach

Refuse a runtime whose `process.versions` names Bun or Deno, with a message that says why and how to run under Node. Supporting those runtimes is a different feature.

## Acceptance

- Both gates refuse Bun and Deno with an actionable message; Node is unaffected.
- A regression case pins the refusal.

### LINKED ISSUE #1414: fix(ttsx): a read-only project directory fails on a bare EPERM
## Problem

ttsx fails on a project directory that refuses writes, which covers read-only checkouts, mounts, and container filesystems. Both of these failures show only a bare `EPERM`:

1. **An included entry:**
   - Without `--cache-dir`, the runtime output directory defaults to the project's `node_modules/.cache/ttsc/ttsx`.
   - Creating it fails with `EPERM: operation not permitted, mkdir '...\node_modules\.cache\ttsc\ttsx'`, before anything compiles.
2. **An entry outside `include`, and a root the program reaches at run time:**
   - These are compiled through a temporary tsconfig written beside the project's own, for the length of the build.
   - The write fails with `EPERM: operation not permitted, open '...\.ttsx-entry.<pid>.tsconfig.json'`, naming a file the user never created.

## Evidence

On Windows 11 with an ACL that denies file and directory creation on the project directory:
- `ttsx src/index.ts` failed as in case 1;
- `ttsx clear.ts` with `--cache-dir` elsewhere failed as in case 2;
- the same included entry ran when given `--cache-dir` elsewhere.

## Consequence surface

- **The runtime cache default** (`prepareExecution`), and therefore `ttsc/register` too.
- **The temporary tsconfig.**
  - TypeScript-Go's command line cannot combine a project with a file list: `-p` with file names is TS5042, and file names with a tsconfig present is TS5112 unless `--ignoreConfig` drops the config.
  - `${configDir}`, the default `typeRoots`, and `types` resolve from the directory of the config the compiler is given.
  - So a checked build of one file with the project's options needs a config file in that directory.
  - ttsc ships no compiler host that could replace the roots in memory: the platform `ttsc` binary refuses `build` and `check`.
- **Emit-only builds** (installed packages, descriptor loads) already keep their temporary tsconfig in the private cache.

## Approach

- The default runtime cache falls back below the system temp directory when the project refuses it. A `--cache-dir` the user named is never replaced.
- The temporary tsconfig write reports the directory, the cause, and the remedy: make the directory writable, or add the file to the project's `include`/`files`.
- Replacing the roots in memory would need a shipped compiler host, which is a packaging decision outside this issue.

## Acceptance

- An included entry of a read-only project runs without `--cache-dir` and leaves the project unchanged.
- An out-of-include entry there fails naming the directory and the remedy.
- Regression cases on Windows (ACL) and POSIX (mode bits).

### LINKED ISSUE #1415: fix(ttsx): a forwarded --noEmit or --emitDeclarationOnly stops the entry from running
## Problem

Forwarding a flag that suppresses output breaks the run: `ttsx --noEmit src/main.ts` and `ttsx --emitDeclarationOnly --declaration src/main.ts` both fail with `ttsx: emitted entry not found`.

Flags before the entry go to the type-check, and neither flag changes that check. But ttsx's runtime build is the same compiler pass. The forwarded flag comes after ttsx's own `--noEmit false --emitDeclarationOnly false`, so the build writes no JavaScript and nothing can run.

## Evidence

Reproduced on `779c60c9` and on the head of #1410, with a trivial CommonJS project: both commands exit 2 with `emitted entry not found`, and the same entry runs without the flag.

## Consequence surface

- Every ttsx runtime build that takes forwarded flags: the entry project build and the entry-only build of an out-of-`include` entry.
- ttsc's separate TypeScript-Go check pass, which runs before a native plugin host emits. It must stay no-emit even when the runtime build re-enables emit.
- A response file (`@file`) can carry the flag too.

## Approach

- Runtime builds switch `--noEmit` and `--emitDeclarationOnly` back off after every forwarded flag.
- The check pass's own `--noEmit` follows the forwarded flags, so it wins.

## Acceptance

- Both commands run the entry.
- A regression case pins them.

### LINKED ISSUE #1416: fix(ttsx): a forwarded --inlineSourceMap fails with TS5053
## Problem

`ttsx --inlineSourceMap src/main.ts` fails before the program starts, with `error TS5053: Option 'sourceMap' cannot be specified with option 'inlineSourceMap'`.

ttsx forces `--sourceMap` on its private runtime build when the build would otherwise carry no map to inline, which keeps stack traces and coverage mapped (#353). That decision reads the project config only, so a forwarded `--inlineSourceMap` gets a forced `--sourceMap` beside it.

## Evidence

Reproduced on the head of #1410, for an included entry and for one outside `include`. Without the flag, both run.

## Consequence surface

- The entry project build and the entry-only build.
- The effective-option reader handled a bare boolean flag (`--inlineSourceMap` last, or followed by another flag) only for `noLib`. Every other bare flag read as the next token, or as nothing.

## Approach

Decide whether to force a map from the options the build actually compiles with: the config plus the forwarded flags. A bare boolean flag reads as on.

## Acceptance

- `--inlineSourceMap` runs, and a stack trace maps to the `.ts` line through the inline map.
- A regression case covers an included entry and one outside `include`.

### LINKED ISSUE #1417: fix(ttsx): a forwarded --module silently skips the entry
## Problem

On a CommonJS project, `ttsx --module esnext src/main.ts` (or `--module preserve`) never runs the entry and exits 0 without a line of output.

The forwarded flag makes the runtime build emit ES modules. The runtime classifies each served file by the project config alone, so it serves that ESM as CommonJS, and Node's loader evaluates nothing. On `779c60c9` the same command at least failed loudly, with `SyntaxError: The requested module './helper' does not provide an export named 'helper'`.

## Evidence

Head of #1410, CommonJS project whose `src/main.ts` imports a sibling:
- `ttsx src/main.ts` prints the value;
- `ttsx --module esnext src/main.ts` prints nothing and exits 0;
- a one-line `console.log` entry behaves the same.

## Consequence surface

- The entry project build and the entry-only build receive forwarded flags, and their classification must follow the emit.
- Runtime roots and dependency builds receive no forwarded flags. They emit by the config and are classified by it, so they stay consistent.
- The effective-option reader already covers a response file through `--showConfig`.

## Approach

Classify the builds that take forwarded flags by their effective `module` and `target`: the config plus those flags, in the order the compiler applies them.

## Acceptance

- `--module esnext` and `--module preserve` run an included entry and an out-of-include entry that import a sibling.
- A regression case pins both.

samchon:
## Intent

This is the one pull request for the whole accepted cycle of the ttsx issue campaign. It covers:
- exact source ownership in every runtime and emit lane;
- private build isolation, cache validity, project discovery, and JSX execution;
- the launcher's process and argv contract;
- the one-identity structure refactor of `packages/ttsc`.

Each of the first ten issues closes from the commit that implements it. The five found by the later discovery rounds close from this description:

Closes #1413
Closes #1414
Closes #1415
Closes #1416
Closes #1417

## Cycle issues

| Issue | Resolution | Commits |
| --- | --- | --- |
| #1407 one documented public identity per source file | Every `packages/ttsc/src` module is split to one exported identity named after its file, placed first, with JSDoc. A test-evidence case runs the real `evidence/singular` and `evidence/documented` rules over the package source, with a canary. The convention is recorded in the development skill. | `d325b1f8`, `b9d87f38`, `a6917a37`, `9dafd501`, `1aef145d` |
| #1382 ttsx serves an unrelated same-stem emit | `EmitOwnershipIndex` replaces trailing-stem scoring. It works from the build's output record, a forward mirror against the pinned root, and an inverse identity match. It handles 8.3 names, symlinks, and case, and uses the source map or compiler precedence for `.ts`/`.tsx` twins. A source no build compiled is compiled alone through its owning project: type-checked in the user tree, emit-only in `node_modules` and descriptor loads. An emit-only root's temporary tsconfig lives in the private cache, with plugin config discovery anchored at the real one. `ttsc <file>` refuses a file outside the file set. | `b8ffa205`, `f94e165d`, `19f2b47a`, `d3ca97b4`, `fcc3d73d`, `4b9b4775`, `423b7a4f`, `73fae775`, `de668b2b`, `d899c10d`, `f973904f` |
| #1408 preserved JSX cannot run | Runtime builds compile a preserved `jsx` mode with the runtime the checker reads. An import source selects `react-jsx`; otherwise a declared factory selects `react`. Orphans use `react-jsx`, and the orphan cache key includes the emit policy. | `e61547b7`, `023515c5`, `71f6dad0`, `dcc6c16c`, `ebfad711` |
| #1404 private builds write into the project | Every ttsx build, including the separate type-check pass, isolates declarations, build information, `outFile`, and forwarded output locations. Emit-only lanes no longer inherit `noEmitOnError`, and no emit-only build creates a file in an installed package. | `1d1f92c3`, `2a4743b7`, `423b7a4f` |
| #1406 solution-style references ignored | The owning project follows the language service's rule: directory references, nested solutions, cycles, and the compiler's own root-file list. A file the program writes at run time is looked up again. | `3a29d96e`, `b609f7fe` |
| #1405 persistent caches outlive their inputs | Manifest-less dependency builds go to a per-process directory, and dead owners' directories are swept. The orphan cache is keyed by compiler identity and ttsc version. | `48b000bf` |
| #1401 launcher consumes the program's argv | Terminal flags are read only before the entry, and the separator is read in order. Only a `--` directly after the entry is dropped. | `5f536469`, `449a277a` |
| #1409 `--watch` hangs | Refused before any compiler starts, with alternatives. | `5f536469` |
| #1402 entry is not Node's main module | The child is `node [-r …] <entry>`, and the bootstrap is removed. | `694b0e35`, `0615df74` |
| #1403 signals and cleanup | The launcher is async and holds termination signals. SIGTERM and SIGHUP are forwarded, cleanup runs on every path, and a child's signal death is re-raised. | `5f802447`, `67e16f8f` |
| #1413 Bun and Deno pass the Node gate | The launcher and the `ttsc/register` preload refuse a runtime whose `process.versions` names Bun or Deno, and say how to run under Node. | `84748aeb` |
| #1414 a read-only project fails on a bare EPERM | Without `--cache-dir`, the runtime cache falls back below the system temp directory. The temporary tsconfig a file outside the file set needs reports the unwritable directory and the remedy. | `43bf5897`, `1c6b2d46` |
| #1415 a forwarded `--noEmit` stops the run | Runtime builds switch `--noEmit` and `--emitDeclarationOnly` back off after the forwarded flags. Check passes keep their own `--noEmit` last. | `44ca4a97` |
| #1416 a forwarded `--inlineSourceMap` fails with TS5053 | Whether to force a source map is read from the config and the forwarded flags. | `3ce6c0cf` |
| #1417 a forwarded `--module` silently skips the entry | Builds that take forwarded flags classify their emit by the effective `module`/`target`, including through a response file. | `3ce6c0cf`, `123f5aa3` |

A pre-existing Windows bug found on the way is also fixed: `readProjectConfig` returned inherited absolute path options with `/` separators.

**Change integrity:** `d899c10d` changes an existing fixture. Its fake native host wrote `outDir/main.js` for `test/main.ts`, a layout no compiler produces that only stem matching accepted. The assertion is unchanged, and `pitfalls.mdx` now states the layout a self-writing sidecar must follow.

## Review

- Every commit has an Individual Self-Review recorded as a `COMMENT` review on this pull request. Every finding was adjudicated there, and each accepted one was fixed in a follow-up commit named in that review.
- Subagent reviewers ran until the user asked for single-agent work. From `19f2b47a` on, the main agent reviewed each commit by reading the diff.
- Overall Self-Review rounds ran against the head while CI ran, and are recorded as reviews.
  - Round 2 proved the refactor range and both formatting commits equivalent mechanically, and read every behavioral declaration in full.
  - Later rounds followed each batch of fixes.
  - The final round on `123f5aa3` is clean.
- Discovery rounds 4–7 ran after the first ten issues landed. Round 7 was empty.

## Verification

- **CI:** runs on Linux, macOS, and Windows. The runtime suite (`features/ttsx-runtime`) now runs on the macOS and Windows representative rows as well as in the Linux core lane (`81d1f7a9`).
- **Local, Windows:** 172/173 runtime cases pass. The remaining one needs a file-symlink privilege this workstation lacks; CI runners have it.
- **Node 22.15.0 (the engines floor):** 166/167 runtime cases pass, with the same exception.
- **Pinned items:**
  - An ESM entry, top-level await included, runs on Node 22.15.0 and 24.
  - `node --watch --require ttsc/register` runs and restarts.
  - The POSIX signal case passes on Ubuntu.
- **Discovery rounds 4–7** probed:
  - runtimes and environments: workers, `fork`, Yarn PnP, Bun, Deno, a 3000-module project, concurrency, nodemon, read-only trees;
  - edges: ESM edges, 20 forwarded compiler flags, response files, argv quoting, and spaced install paths.
  - Their dispositions are recorded as reviews on this pull request.
- **Not verifiable here:** Windows console Ctrl+C delivery needs an attached console.

https://claude.ai/code/session_01RUNXb4ecweEo5KBbpAuNk4


samchon (commented):
## Individual Self-Review: `d325b1f8` refactor(ttsc): one documented public identity per source file

**What landed.** Every multi-export module under `packages/ttsc/src` was split so each file exports one public identity named after the file, right after its imports, with a JSDoc block. Shared internals became one exported namespace per concern. Tests, scripts, Go comments, and docs followed the new paths. A test-evidence case now runs the real `evidence/singular` and `evidence/documented` rules over a copy of the package source, and a canary proves both rules are active. The development skill records the convention. Resolves #1407.

**Reviewer result.** No behavior regression. Function bodies match old against new apart from import rewiring. The added `".."` path depths resolve correctly in the compiled `lib` layout. Public exports are unchanged, module-level state still has one owner, and the one new import cycle is safe under CommonJS.

**Adjudication**
- 127 files are not prettier-formatted. Accepted as expected: the campaign formats the unified pull request once before merge (development skill, issue-campaign development).
- Three doc-accuracy findings were verified and fixed in `b9d87f38`. See the inline comments.
- The evidence case did not account for location-less diagnostics. Verified and fixed in `b9d87f38`: it now requires every printed `error TSxxxx` code to be a located diagnostic.

packages/ttsc/src/plugin/internal/source/PluginBuildLockProtocol.ts
```diff
@@ -0,0 +1,188 @@
...
+ * held generation is `current/`, carrying a generation id and an owner, and it
+ * is freed only by renaming `current/` onto the tombstone
+ * `retired/<generation>`, so a stale releaser can never free a successor. The
+ * legacy protocol held `<lockDir>` itself; a v2 process still recognizes and
+ * reclaims an abandoned legacy lock, fenced by the legacy directory's mtime.
+ */
+export namespace PluginBuildLockProtocol {
+/**
+ * How long a waiter trusts a live-looking holder before stealing the lock. It
+ * matches the ttsx dependency-build lock (`DependencyBuildLockProtocol`), so a
```
samchon: Finding (nit): the steal-timeout reference named `DependencyBuildLockProtocol`, but `DEP_BUILD_LOCK_STEAL_MS` lives in `installRuntimeHooks.ts`. Fixed in b9d87f38.

packages/ttsc/src/launcher/internal/watch/WatchPaths.ts
```diff
@@ -0,0 +1,71 @@
+import fs from "node:fs";
+import path from "node:path";
+
+/**
```
samchon, line 4: Finding (nit): "lexical" overstated the helpers, because `isDirectory` and `nearestExistingDirectory` stat the disk. The doc now says which helpers touch the filesystem. Fixed in b9d87f38.

packages/ttsc/src/internal/IJavaScriptRuntimeCapabilities.ts
```diff
@@ -0,0 +1,29 @@
...
+ * What a JavaScript runtime executable can do, as measured by running it.
+ *
+ * Produced by {@link javascriptRuntimeCapabilities}, which asks the interpreter
+ * itself rather than inferring from the host process: a `node` found on `PATH`,
+ * a version-manager shim, and Bun can all stand in for the runtime a plugin
+ * descriptor is evaluated with, and each supports a different loader.
+ */
+export type IJavaScriptRuntimeCapabilities = {
+  /**
+   * The executable is Bun. Bun loads TypeScript descriptors natively, so the
```
samchon: Finding (nit): the Bun claim was right in effect but wrong in cause. The preload is gated on `registerHooks`, which Bun lacks, not on `bun`. The doc now states the actual gate. Fixed in b9d87f38.

tests/test-evidence/src/features/test_evidence_repository_ttsc_source_holds_one_documented_identity_per_file.ts
```diff
@@ -0,0 +1,220 @@
...
+        "  },",
+        "};",
+        "",
+      ].join("\n"),
+      files,
+      compilerOptions: readPackageCompilerOptions(packageRoot),
+    });
+    try {
+      linkPackageDependencies(packageRoot, project.directory);
+      const result: IRunResult = runCheck(project.directory);
```
samchon, line 83: Finding (minor): only located diagnostics were parsed, so an option error with no file position would pass unseen. The case now requires the number of printed error codes to equal the number of located diagnostics. Fixed in b9d87f38.

tests/test-evidence/src/features/test_evidence_repository_ttsc_source_holds_one_documented_identity_per_file.ts
```diff
@@ -0,0 +1,232 @@
...
+      const diagnostics: IDiagnostic[] = parseDiagnostics(result.output);
+      // A diagnostic with no location (an option error, say) matches no parsed
+      // line, so every code the compiler printed has to be one of them.
+      const printed: number =
+        stripColors(result.output).match(/\berror TS\d+:/g)?.length ?? 0;
```
samchon: Finding (low): `parseDiagnostics` accepts warnings, but the tally counted only errors. A warning would therefore produce a misleading "printed 1 but only 2" message ahead of the violation list. Also, a quoted `error TS1234:` in a code frame could trip the tally before the violation list. Fixed in a6917a37: the tally counts both severities and runs after the per-file violation check.

packages/ttsc/src/internal/IJavaScriptRuntimeCapabilities.ts
```diff
@@ -0,0 +1,31 @@
...
+ * What a JavaScript runtime executable can do, as measured by running it.
+ *
+ * Produced by {@link javascriptRuntimeCapabilities}, which asks the interpreter
+ * itself rather than inferring from the host process: a `node` found on `PATH`,
+ * a version-manager shim, and Bun can all stand in for the runtime a plugin
+ * descriptor is evaluated with, and each supports a different loader.
+ */
+export type IJavaScriptRuntimeCapabilities = {
+  /**
+   * The executable is Bun. Bun loads TypeScript descriptors natively and has no
```
samchon: Finding (low): the wording still read as if the `bun` flag omitted the preload. The actual gate is the separate `registerHooks` measurement, and the doc also left out `--no-install`. Fixed in a6917a37.

samchon (commented):
## Individual Self-Review: `b9d87f38` docs(ttsc): correct the structure refactor's review findings

**What landed.** Fixes for the four findings from the `d325b1f8` review: the plugin lock steal-timeout reference, the `WatchPaths` "lexical" wording, the Bun capability doc, and a tally in the repository evidence check that accounts for location-less diagnostics. It corrects #1407 and closes no issue.

**Reviewer result.** All four claims are verified against the code. The tally cannot produce a false pass: ttsc renders diagnostics through `FormatDiagnosticsWithColorAndContext`, and its summary table contains no `error TSxxxx:` text.

**Adjudication.** All three low-severity candidates were accepted and fixed in `a6917a37`. See the inline comments.

samchon (commented):
## Individual Self-Review: `a6917a37` test(evidence): report source violations before the diagnostic tally

**What landed.** The repository evidence check now tallies warning codes as well as error codes, and it does so after the canary and per-file violation checks. The Bun capability doc now names the `registerHooks` gate. This corrects #1407 and closes no issue.

**Reviewer result.** Both claims are verified. Lint warnings print the literal word `warning` (`LintDiagnostic.Category()` renders through `Category.Name()`). Each located line carries exactly one category token. The flow still fails on a missing canary diagnostic, on a non-canary diagnostic, and on a location-less diagnostic.

**Adjudication.** All three low-severity nits were accepted and fixed in `9dafd501`:
- the violation message now includes the full output, so a location-less diagnostic in the same run is visible;
- the two parser doc comments now say error or warning;
- the Bun wording states only what the measured gate does, instead of asserting something about every Bun release.

samchon (commented):
## Individual Self-Review: `9dafd501` test(evidence): show the whole check output beside source violations

**What landed.** The per-file violation error in the repository evidence check now carries the full check output. The two parser doc comments say error or warning. The Bun capability doc states only what the measured `registerHooks` gate does. This corrects #1407 and closes no issue.

**Reviewer result.** Clean. Both claims are verified. The preload gate checks only the measured `registerHooks`. The spawned runtime can be Bun. The `!bun` check only selects the Node executable for nested tools. So the doc sentence is exactly true.

**Adjudication.** Nothing to act on.

packages/ttsc/src/launcher/internal/prepareExecution.ts
```diff
@@ -326,10 +340,7 @@ function buildProject(
   if (result.status === 0) {
     linkVirtualProjectLayout(context);
```
samchon, line 408: Finding 1 (verified): `linkVirtualProjectLayout` ran before `listOutputs`, so hard-linked or copied user files in the no-`outDir` emit mirror were recorded as outputs. Fixed in f94e165d: the record is now taken first.

packages/ttsc/src/launcher/internal/buildSingleRootProject.ts
```diff
@@ -0,0 +1,192 @@
...
+}): { project: ReturnType<typeof readProjectConfig>; rootDir: string } {
+  const options = props.options ?? {};
+  // `source` already carries the one spelling the runtime decided on. tsgo
+  // compares it against `rootDir` textually — `GetCommonSourceDirectory` takes
+  // `rootDir` verbatim and `ContainsPath` is lexical — so a mismatch here is
+  // not a near miss: the file counts as outside `rootDir`, and tsgo emits it to
+  // its own source path with the extension changed instead of under `outDir`,
+  // writing a `.js` and its map beside the user's `.ts` where nothing cleans
+  // them up.
+  const rootDir = commonAncestorDirectory(
```
samchon: Findings 2 and 7a: the inherited `composite` rejected every import (TS6307), and a narrow rootDir rejected imports beyond the project (TS6059). Fixed in f94e165d with `composite: false` and a volume-root rootDir.

samchon (commented):
## Individual Self-Review: `b8ffa205` fix(ttsx): run a source only from output compiled from that same file

**What landed.** `EmitOwnershipIndex` replaces trailing-stem scoring in every ownership lane: runtime serving, entry preparation, dependency builds, isolated emits, CommonJS name scans, and `ttsc <file>`. Each build records its outputs, and the record is the authority. A source no build covered is compiled alone through its nearest project: checked in the user tree, emit-only under `node_modules`, and cached per run by content. `ttsc <file>` refuses an unowned file. Resolves #1382.

**Reviewer result.** Case and 8.3 inverse matching, extension mappings, isolated emits, root-build naming and cleanup, cache keys, and marker rejection are all verified correct. The reviewer reproduced four defects and suspected three more.

**Adjudication.** Every finding was accepted and fixed in `f94e165d`, each with a regression case where one is constructible:
1. The output record was taken after the virtual layout hard-linked user files into the emit mirror, so a stale `tool.js` ran. It is now taken first (`test_ttsx_never_runs_a_stale_javascript_file_beside_an_uncompiled_source`).
2. `composite` made every import of a compiled-alone root fail with TS6307. The synthesized config now switches it off (`test_ttsx_runs_a_root_that_imports_beyond_a_composite_project`).
3. Local `.ts` plugin descriptors outside `include` were type-gated under the consumer config. They are now emit-only during descriptor loads (`test_loadprojectplugins_evaluates_a_typescript_descriptor_outside_include_without_the_type_gate`).
4. A forwarded `--rootDir` was ignored by the mirror. A shared effective-option reader now feeds ttsx, root builds, and `ttsc <file>` (`test_ttsx_and_ttsc_follow_a_forwarded_root_dir`).
5. Per-lookup cost: `find` now memoizes by spelling before resolving, and each output's source identities are resolved once.
6. `a.ts` next to `a.tsx`: an output two TypeScript sources could produce is now claimed by neither (unit case added).
7. Root-lane edges:
   - rootDir is now the source volume, so an import beyond the project is no longer TS6059, and an emit-only build never writes beside the source.
   - An emit-only root whose build fails (for example, a read-only install) falls back to the isolated emit.
   - A failed root is memoized. No regression case exists for the read-only install, because a read-only directory is not reliably constructible across the three CI platforms.

packages/ttsc/src/compiler/internal/runSingleFileEmit.ts
```diff
@@ -31,22 +37,26 @@ export function runSingleFileEmit(options: TtscSingleFileEmitOptions): string {
     tsconfig: options.tsconfig,
   });
   const tsconfig = project.path;
-  const projectRoot = project.root;
+  // The root `pinInferredRootDir` hands tsgo below: the declared one, which
+  // `readProjectConfig` already absolutized, or the project's own directory.
+  const rootDir =
+    typeof project.compilerOptions.rootDir === "string"
+      ? project.compilerOptions.rootDir
```
samchon: Finding 4 (verified): a forwarded `--rootDir` overrides this root at the tsgo level. Fixed in f94e165d through `readEffectiveCompilerOptions`.

samchon (commented):
## Individual Self-Review: `e61547b7` fix(ttsx): compile JSX a project preserves for another tool

**What landed.** Preserved JSX modes (`preserve`, `react-native`) are replaced in every runtime build that goes through a project. The replacement is the classic transform when a factory is declared, and otherwise the automatic runtime. The orphan lane compiles `.tsx` with the automatic runtime, and the orphan cache key now includes the emit arguments. Resolves #1408.

**Reviewer result.** Several points were verified as correct:
- casing and response files work;
- inline `--jsx=preserve` is rejected by tsgo itself;
- `react-jsxdev` and forwarded executable modes are left alone;
- the isolated `--jsx react-jsx` is harmless for `.ts` and `.cts` inputs;
- only runtime lanes call `runtimeCompilerArgs`;
- the orphan test fails without the change.

One real bug and three narrow gaps were found. All four are projects that failed with a SyntaxError before this change, so none is a regression.

**Adjudication**
1. `reactNamespace` was not treated as classic, which produced TS5089. Accepted and fixed in `023515c5`.
2. A factory together with `jsxImportSource` under preserve produced TS5089. Accepted and fixed in `023515c5`: the declared factory wins and the import source is set to `null` for the runtime build.
3. A per-file `@jsx` pragma with no configured factory is ignored by the automatic runtime. Accepted as a documentation item: `@jsxRuntime classic` is now documented. A per-file pragma cannot choose a project-wide mode.
4. `jsxFactory` without `jsxFragmentFactory`, used with fragments, fails with TS17016 under `react`. Accepted as a documentation item.
5. The dependency cache key is unchanged. This only matters for the manifest-less shared temp fallback, which #1405 scopes per process. Deferred to #1405.
6. Test gaps (`react` and `react-jsxdev` unchanged, response file, `reactNamespace`, factory plus import source) are closed by `test_ttsx_keeps_executable_jsx_modes_and_honors_every_classic_declaration`.

samchon (commented):
## Individual Self-Review: `023515c5` fix(ttsx): honor every classic JSX declaration a preserved project makes

**What landed.** `reactNamespace` now selects the classic transform. A factory declared beside `jsxImportSource` kept the factory. The JSX fixture gained `jsx-dev-runtime` and `createElement`, and the docs and a modes test were added. This corrects #1408 and closes no issue.

**Reviewer result.**
- `--jsxImportSource null` is accepted by tsgo 7.0.2 and unsets the config value.
- `reactNamespace` with fragments runs.
- Forwarded and response-file import sources are handled.

One verified, medium-high finding: when a preserved project declares both a factory and `jsxImportSource`, TypeScript's checker reads the import source (automatic runtime). The commit kept the factory, so a program that `ttsc` checks cleanly failed with TS2874. The test case could not tell the two transforms apart.

**Adjudication**
- Accepted and fixed in `71f6dad0`. An import source now selects `react-jsx` and nulls the factories. The test case now has no factory in scope and no global `JSX`, so only the automatic runtime can compile it. The docs are corrected.
- The upstream `jsx: react` pragma behaviour in a script-mode file is out of scope.
- The `react`, `react-jsxdev`, and response-file cases add coverage. They are kept deliberately as unchanged-behaviour guards.

samchon (commented):
## Individual Self-Review: `f94e165d` fix(ttsx): close the ownership gaps the source ownership review found

**What landed.**
- The output record is taken before the virtual layout.
- The ownership index has a two-TypeScript-source ambiguity rule, and lookups are memoized.
- Single-root builds use a volume-root `rootDir` with `composite` off.
- Descriptor loads are emit-only.
- An emit-only root failure falls back to the isolated emit, and failed roots are memoized.
- A forwarded `--rootDir` is read through `readEffectiveCompilerOptions`.

This corrects #1382 and closes no issue.

**Reviewer result.** The following were verified correct:
- the effective-root resolution base in every lane (tsgo runs in `execution.projectRoot`);
- `rootIsChecked`, the fallback, and the content-keyed failure memo;
- deep paths (more than 260 characters) under the volume root, with correct stack traces;
- recording before the layout;
- the in-place `x.js` beside `x.ts` layout.

**Adjudication**
- **High, verified:** the ambiguity rule refused `twin.js` for both `twin.ts` and `twin.tsx`. The compiler includes and resolves `.ts` over `.tsx`, so this broke an ordinary layout. Fixed in `19f2b47a`: the build's source map names the file the compiler read, and without a map the compiler's precedence decides. Covered by a unit case and by `test_ttsx_serves_a_source_whose_twin_has_another_typescript_extension`.
- **Doc overclaim** ("import from anywhere") is not true across Windows drives. Fixed in `19f2b47a`.
- **Declarations of root builds landing in a declared `declarationDir`:** resolved by #1404 isolation (`1d1f92c3`, `2a4743b7`).
- **Low: memoized answers freeze disk state** for the life of the process. Accepted as is: a served file is looked up when it loads, and Node caches the module anyway.
- **Low: the failure memo replays a failure caused by an imported file** fixed later in the same process. Accepted, since `builtRoots` has the same scope.
- **Low: repeated `--showConfig` spawns with an `@file`.** Accepted, since it is only paid with response files.

samchon (commented):
## Individual Self-Review: `1d1f92c3` fix(ttsx): keep every output of a runtime build in the run's directory

**What landed.** The entry project build, entry-only and runtime root builds, and dependency builds now pass `isolateOutputsTo`. Resolves #1404.

**Reviewer result.** Probed and verified:
- the no-plugin lanes: `composite`, `emitDeclarationOnly`, and `incremental` without `tsBuildInfoFile`;
- dependencies with their own `declarationDir` and `tsBuildInfoFile`;
- that emitted JavaScript, source maps, and ownership lookups are unaffected;
- that all three new tests fail before the change.

`outFile` with AMD is rejected by tsgo 7 itself.

**Adjudication**
- **Medium, verified:** the separate `tsgo --noEmit` check pass, which runs with a check plugin that does not report TypeScript diagnostics or after a plugin failure, got no isolation. It wrote the project's `tsBuildInfoFile`. Fixed in `2a4743b7`: `runTsgo` appends the isolation arguments. Covered by `test_plugin_corpus_ttsx_check_plugin_pass_leaves_build_info_untouched`, which fails with the line removed (negative control run locally).
- **Low, pre-existing:** a `composite` + `declarationMap` out-of-include entry failed with TS5069 once `composite` was off. Fixed in `2a4743b7` by turning declarations off in single-root builds. The composite test now sets `declarationMap`.
- **Nit:** the forwarded-flags test compares top-level names only. Accepted, since every forwarded location in it is top-level.

samchon (commented):
## Individual Self-Review: `71f6dad0` fix(ttsx): pick the JSX runtime the type-checker already reads

**What landed.** Under a preserved `jsx` mode, a `jsxImportSource` now selects `react-jsx` and nulls any declared factories. A factory without an import source selects `react`. This corrects #1408.

**Reviewer result.** Verified against tsgo 7.0.2:
- the checker reads JSX through the import source when a factory is also declared, including with `reactNamespace`;
- `--jsxFactory null`, `--jsxFragmentFactory null`, and `--reactNamespace null` unset config values;
- the changed test case distinguishes the two transforms.

**Adjudication**
- **Low:** a file marked `@jsxRuntime classic` in a project with both a factory and an import source runs without the configured factory. No single mode serves both kinds of file (TS5089). Documented in `dcc6c16c`/`ebfad711`: such a file needs its own `@jsx` and `@jsxFrag` pragmas.
- The stale `runtimeCompilerArgs` summary is corrected. A `reactNamespace` plus import-source case is added, and the test was renamed to what it asserts (`dcc6c16c`/`ebfad711`).

samchon (commented):
## Individual Self-Review: `19f2b47a` fix(ttsx): settle a same-stem .ts/.tsx output by the file the compiler read

Following the user's instruction, this and every later Individual Self-Review in this PR is done by the main agent personally, not by a subagent.

**What landed.** When an output could come from two TypeScript sources, `EmitOwnershipIndex.owns` reads the build's source map and trusts the file it names. With no map, the compiler's precedence applies (`.ts` before `.tsx`). The docs narrow the volume-root reach to the same drive. There is a unit case and an end-to-end twin case.

**Review.** I read the diff and probed the twin layout under `preserve` and `react-jsx`, with and without `exclude`. `ttsc` and `ttsx` agree on which twin is compiled. Two findings:
1. A map whose `sourceRoot` is a URL names no file on disk, so it would be counted as proof against the real owner. Fixed in `d3ca97b4`: such a map counts as no map.
2. The forward mirror returned `null` when the first extension candidate belonged to the twin. It never tried `.jsx` for a preserved `.tsx`; the inverse lookup happened to rescue it. Fixed in `d3ca97b4`, with a unit case.

**Known limit.** `ttsc <file>` has no forced source map. If an `exclude` makes the compiler take `twin.tsx` while `twin.ts` exists, the precedence fallback refuses `twin.tsx` there. ttsx is unaffected because its builds always have maps. The layout is rare and the refusal is explicit, so this is accepted and documented in `owns`'s JSDoc.

samchon (commented):
## Individual Self-Review: `2a4743b7` fix(ttsx): isolate the type-check pass and drop root declarations

**What landed.**
- `TsgoArguments.isolatedTsgoOutputArgs` is now exported and documented, and `BuildExecution.runTsgo` appends it after the forwarded flags.
- Single-root builds set `declaration: false` and `declarationMap: false`.
- The new native-plugin case pins the check pass, and the composite case now sets `declarationMap`.

**Review.**
- `runTsgo` is only reached from `runBuild` lanes. The isolation arguments are empty unless `isolateOutputsTo` is set, so ordinary `ttsc` builds are unchanged.
- A `--noEmit` pass given `--outDir <private>` does not trigger TS5011, because tsgo skips the rootDir inference check for no-emit.
- Emit builds already pass `--emitDeclarationOnly false`, so `declaration: false` cannot conflict with an inherited `emitDeclarationOnly`. I confirmed this by running the `emitDeclarationOnly` project through the out-of-include lane.
- Negative control: with the one `runTsgo` line removed from the built lib, the new case fails with "wrote build information into the project". Restored, it passes.

No findings.

samchon (commented):
## Individual Self-Review: `dcc6c16c` and `ebfad711` (JSX case rename and content)

`dcc6c16c` was meant to carry the `reactNamespace` plus import-source case, the `@jsxRuntime classic` documentation, and the corrected `runtimeCompilerArgs` summary. Its staging step failed on a stale pathspec, so it committed only the file rename. `ebfad711` adds the content and says so in its message. I did not force-push, so both commits stay in the history.

**Review of the combined change.**
- The new case (`preserve` with `reactNamespace: "R"` and `jsxImportSource`, and no `R` in scope) can only pass if the automatic runtime is selected and `--reactNamespace null` is applied. It passes.
- The docs sentence matches the verified behaviour: `@jsx` and `@jsxFrag` pragmas make a classic-marked file run.

No findings.

samchon (commented):
## Individual Self-Review: `d3ca97b4` fix(ttsx): trust a source map only when it names a file on disk

This commit implements the two findings from the `19f2b47a` self-review:
- a map naming no file on disk counts as no map;
- the forward mirror continues to the next extension.

I re-ran the unit, twin, same-name, and JSX-extension cases, and all pass. The `preserved/doc` unit case pins the second fix at the index level.

No findings.

samchon (commented):
## Individual Self-Review: `3a29d96e` fix(ttsx): run a file with the referenced project that owns it

**What landed.**
- `compiler/internal/project/resolveOwningProjectConfig.ts`: the language service's rule for choosing a file's project from its nearest config and that config's references. Containment comes from the compiler's own `--showConfig` file list, compared by filesystem identity.
- The rule is wired into ttsx entry discovery, including `ttsc/register` through `prepareExecution`, and into the runtime's `owningModuleOptions` and project lane (`owningTsconfig`).
- New docs section "Solution-style configs", plus two cases. The first covers a Vite-shaped solution with app and node configs, register, and an uncontained script. The second covers directory references, nested solutions, a cycle, and a required source in a solution-shaped package.
- Resolves #1406.

**Review.** I reviewed the diff by reading it. As the user directed, verification runs once at the end.
- **Cost:** a config without `references` returns before anything is spawned, so ordinary projects pay one extra JSON read. With references, `--showConfig` runs once per config per process (`rootFilesCache`), and the runtime memoizes the answer per file.
- **Cycle:** the discovered config's identity is seeded into `seen`, so a reference back to it is skipped.
- **Explicit project:** `-P` keeps its existing path through `readProjectConfig`.
- **Descriptor inputs:** every config the search reads is reported through `recordPluginDescriptorTsconfigCandidates`.
- **Residual risk (accepted):** `rootFilesCache` lives for the process, so a file created after the first question about its solution is not counted as contained. It takes the out-of-include lane of the discovered config, the documented behaviour for an uncontained file.

No findings.

samchon (commented):
## Individual Self-Review: `48b000bf` fix(ttsx): scope runtime caches to what decides their content

**What landed.**
- `dependencyCacheRoot` no longer falls back to a persistent shared directory when no manifest exists. Such a process gets a private `os.tmpdir()/ttsx-dep/process-<pid>-<nonce>` directory with an owner record, removed on exit. Creating one sweeps directories whose owner is provably gone.
- `isLocalProcessGone` is shared with `inspectDependencyBuildLock` and replaces its private helpers, with identical semantics: `EPERM` means alive, and a record from another host is never gone.
- The orphan cache key now carries the compiler binary's identity (realpath, size, mtime, inode, device) and the ttsc version.
- Docs cover cache scoping and keys. Two new cases: a manifest-less child edited between runs, and an orphan cache reused until the compiler is rewritten in place.
- Resolves #1405.

**Review.** I reviewed the diff by reading it; verification runs at the end.
- **Version path:** `ownPackageVersion` resolves `packages/ttsc/package.json` from `lib/launcher/internal/runtime`, four levels up.
- **Sweep safety:** it only removes directories with a readable owner record proven gone on this host. A directory created before its record is written is left alone.
- **One run or a manifest-less process:** hits within a run still share its `depCacheDir`. A manifest-less process is its own scope, as #1405's approach requires.
- **Accepted trade-off:** an orphan cache restored across CI machines misses after a reinstall, because the binary's mtime and inode change. Correctness wins over a warm cache here.
- **Legacy entries:** `<key>` directories written by older ttsc versions under `os.tmpdir()/ttsx-dep` are never read now. I did not delete them, since a concurrently running older ttsx may still own them.

No findings.

samchon (commented):
## Individual Self-Review: `5f536469` fix(ttsx): leave the program's argv to the program and refuse --watch

**What landed.**
- `parseFlags` reads the `--` separator in order when `forwardAfterFirstPositional` is set, which is ttsx's mode:
  - a `--` before the entry ends the launcher's options, as before;
  - the one `--` directly after the entry is dropped as the documented optional separator;
  - every later `--` reaches the program.
- The pre-split path that ttsc uses is unchanged.
- `runTtsx.parseCLI` reads `--help` and `--version` from the parse result, so only tokens before the entry count. If parsing fails, it falls back to the tokens before the first entry-like token.
- `assertNoWatch` refuses `--watch` and `-w` before the entry.
- The help text and docs are updated. There are two new cases.
- Resolves #1401 and #1409.

**Review.** I reviewed the diff by reading it; verification runs at the end.
- **Terminal detection:** only dash-prefixed tokens are resolved, so a forwarded value such as `all` in `--target all` cannot read as `--all`.
- **Existing cases still hold:** `test_ttsx_forwards_argv_after_and_runs_preload_modules` and `test_runner_corpus_ttsx_executes_the_intended_entrypoint_and_side_effects` use `entry -- <args>`, which still drops the one separator.
- **Unchanged cases:** `ttsx -- entry.ts` still routes the post-separator tokens to passthrough and reports that no entry was given. The old behaviour is kept, since this is not a layout the issue covers.
- **To confirm in the final verification pass:** the refusal message recommends `node --watch --require ttsc/register <entry>`. That Node's watch mode keeps `--require` for restarts is expected but unproven here, so I will confirm it before merge.

No findings.

samchon (commented):
## Individual Self-Review: `694b0e35` fix(ttsx): run the entry as Node's own main module

**What landed.** The ttsx child is now `node [-r preload...] <entry> [argv...]`. The runtime-hook preload on `NODE_OPTIONS` installs the hooks ahead of every user preload. `registerRuntimeHooks.ts` (the bootstrap) and `runtime/entryModuleFormat.ts` (used only by it) are deleted. Stale comments are corrected, a docs paragraph is added, and one new case covers the main module for CommonJS and ESM, `argv[1]`, a handled uncaught error, `process.exit(7)`, and an unhandled throw. Resolves #1402.

**Review.** I reviewed the diff by reading it; verification runs at the end.
- **Symlinked entry:** Node resolves the main module through its real path, so the target package's `type` decides the format. That matches the compiler and `test_ttsx_classifies_a_symlinked_entry_from_the_target_package_type`.
- **Main-module check:** `isProcessEntry` still compares with `process.argv[1]`, which Node now sets itself. In direct ttsx no `prepareRuntimeEntry` exists, so it only matters under register, as before.
- **Risk, top of the final verification list:** Node now chooses the entry's loader by its own rules. A project that emits ESM (`module: esnext`) in a package without `type` reaches the CommonJS loader. There `_extensions['.ts']` compiles ESM text and relies on Node's `require(esm)` and syntax detection. That is the same path a required non-entry `.ts` file already takes, and register takes it for entries. The bootstrap used to route such an entry through `import()`. If the final pass shows a gap on the oldest supported Node (22.15), the fix belongs in the CommonJS hook, which then serves both entries and required files.

No other findings.

samchon (commented):
## Individual Self-Review: `5f802447` fix(ttsx): forward signals to the program and clean up on every exit path

**What landed.**
- `runTtsx` is async. `LauncherSignals` holds SIGINT, SIGTERM, and SIGHUP on POSIX, and SIGINT and SIGBREAK on Windows, for the whole run.
- A signal during the synchronous preparation is recorded. After preparation has cleaned up, and after the event loop turns (`settleSignals`), the run ends by that signal.
- While the program runs, SIGTERM and SIGHUP are forwarded to it. SIGINT is not, because the terminal already delivers it to the process group.
- The runtime directory is removed on every exit, and a child's signal death is re-raised on the launcher. If the platform cannot send that signal to itself, the launcher exits with 1.
- `ttsx.ts` awaits the result. There is a docs paragraph and a POSIX-only case: SIGTERM to the launcher's pid alone, with and without a handler, and SIGINT to the process group, each with runtime-directory cleanup.
- Resolves #1403.

**Review.** I reviewed the diff by reading it; verification runs at the end. One finding, fixed before commit:
- The catch path for a failed preparation ran in the same tick as the failure, before the pending signal listener could record the signal. A Ctrl-C that killed tsgo mid-check therefore exited 2 instead of 130. Both decision points now await `settleSignals()` first.

**Coverage notes**
- Leaving no synthesized `.ttsx-entry.*.tsconfig.json` after an interrupted out-of-include build now follows from the launcher surviving the signal: `buildSingleRootProject` removes it in `finally`. No automated case pins that timing, because interrupting tsgo inside the synchronous build is not deterministic across CI runners.
- A SIGINT sent to the launcher's pid alone is recorded, not forwarded. The launcher then waits for the program, as a shell does. Accepted: forwarding would deliver a terminal Ctrl-C twice.
- The `ttsc/register` lane owns no process. It still cleans up on `exit` only, and it does not install handlers that would change the program's own signal behaviour, as #1403's approach specifies.

samchon (commented):
## Overall Self-Review, round 1: `779c60c9...5f802447`

I reviewed the whole base-to-head change: the #1407 structure move, which was reviewed per commit, and every behavioural change after it. I checked it issue by issue against each Acceptance list, and read the final `execute.mdx` through.

**Findings (all fixed in `bfe24b6d`)**
1. #1401's acceptance asks for parser unit cases, but only spawned cases existed. Added `test_parseflags_reads_the_ttsx_separator_in_order`, which covers:
   - `--` read in order after the entry;
   - a post-entry `--help` kept in the tail;
   - a pre-entry `--` still ending the launcher's options;
   - ttsc's split-first handling unchanged.
2. #1402's acceptance lists a rejected top-level await, and no case covered it. Added to `test_ttsx_runs_the_entry_as_the_main_module`.
3. `execute.mdx` "Files the program reaches at run time" still said the nearest `tsconfig.json` serves a runtime file. Since `3a29d96e`, a solution-style config hands it to the referenced project that contains it. Corrected.

**Checked without findings**
- **Control flow in `runTtsx`:** help and version, the Node support gate, preparation, the signal settle point, the forwarding window, and cleanup on every path.
- **Final docs:** the pages agree with each other and with the code. `setup/ttsc.mdx` still shows the valid `entry -- args` separator.
- **Formatting:** it touched only files this PR changed (`9b88f4c5`). The full-repository Go formatting was cut short after Prettier finished, and the Go files this PR changed were formatted directly.

Local suites and CI are now running against `9b88f4c5`. Round 2 follows.

samchon (commented):
## Individual Self-Review: `67e16f8f` fix(ttsx): complete the asynchronous launcher the signal change began

I read the diff against `5f802447`, which caused the break. The final build found it: `5f802447` switched to `spawn` and `await`, but the import stayed `spawnSync` and `runPreparedEntry` was not `async`, so `packages/ttsc` did not compile. This is also why the first broad CI run failed across lanes.

**Checked without findings**
- `runPreparedEntry` is awaited on every path that reaches it.
- The only `node:child_process` binding left in `runTtsx.ts` is `spawn`.
- `packages/ttsc` compiles, and the launcher suites ran against this build.

samchon (commented):
## Individual Self-Review: `fcc3d73d` fix(ttsx): compile a runtime root even when its project builds nothing

I read `serveProjectEmit` before and after, and traced both callers down to the isolated fallback.

**Finding (fixed in `4b9b4775`)**
1. No case pinned the change. Added `test_ttsx_compiles_a_required_source_whose_project_lists_no_files`: its owning `tsconfig.json` has `files: []` and `experimentalDecorators`, and the program requires a file beside it by path. The fixed code prints `arguments=3`. With `fcc3d73d` reverted the case fails, because the isolated emit applies standard decorators.

**Checked without findings**
- A failed emit-only root build still returns `null`, and the caller then falls back to the isolated emit, as the updated JSDoc says.
- A failure of a checked root still propagates: a type error stops the run.

samchon (commented):
## Individual Self-Review: `a08c34ad` chore: settle the formatter on namespace clusters

The commit is formatting only; I checked that none of its hunks change any tokens. Prettier needed a second pass on 21 files with namespace clusters, and CI's typecheck lane caught the first pass's output. A third pass over the same files changes nothing. The earlier loop missed files because of the Windows argument-length limit, so this pass ran in batches of 40.

samchon (commented):
## Individual Self-Review: `449a277a` fix(ttsx): consume only the separator directly after the entry

The parser case round 1 added (`test_parseflags_reads_the_ttsx_separator_in_order`) failed on CI and locally, as did the spawned `test_ttsx_hands_every_token_after_the_entry_to_the_program`. The cause: `entry.ts -- -- x` dropped both separators, because the skip condition was "the tail is still empty" rather than "this is the token right after the entry".

**Checked without findings**
- `entryJustRead` is set only when the first positional arrives in forwarding mode, and it is cleared on the next token whatever that token is.
- A `--` before the entry still ends the launcher's options.
- ttsc mode (no `forwardAfterFirstPositional`) never reaches the in-order branch.
- Both cases now pass. I also probed the five documented argv shapes directly against `lib`.

samchon (commented):
## Individual Self-Review: `1aef145d` test(ttsc): load each plugin-lock function from its own module

Three POSIX lock cases failed on CI with `acquirePluginBuildLock is not a function`: their workers still destructured four functions from `buildSourcePlugin.js`, which exports only `buildSourcePlugin` after the #1407 split. `sourceBuildLibraryPath` now takes the module name, matching `dependencyCacheLibraryPath`.

**Consequence surface**
TypeScript cannot check a `require` that a worker script builds as a string at run time, so I scanned for the whole class. Every such `require` of `packages/ttsc/lib` in `tests/` is either a helper-named module or a literal `"packages", "ttsc", "lib", …` path: 22 sites, launcher CLIs excluded. I checked each destructured name against the built module's exports. The only stale ones were these three cases. `TtscCompiler` reaches `index.js` through `__exportStar`.

samchon (commented):
## Individual Self-Review: `423b7a4f` fix(ttsx): keep emit-only root builds out of the source tree

The trigger was `test_loadprojectplugins_ttsx_descriptor_tracks_runtime_inputs`, which failed on CI and locally. A descriptor fingerprints a missing resolution candidate by the metadata of its nearest existing ancestor. The root lane wrote and removed `.ttsx-root.<key>.tsconfig.json` in that directory during evaluation, so the candidates became unstable and lost their cache proof. The same write also targeted installed packages' directories, which may be read-only.

**Decision**
- An emit-only build's tsconfig moves beside its private emit directory.
- A checked build stays beside the real config, because `${configDir}`, `types`, and the default `typeRoots` resolve from it.
- `${configDir}` is the one anchor that changes emit (through `paths`), so a chain that uses it keeps the beside placement.

**Consequence surface**
- Both emit-only lanes turned an inherited `noEmitOnError: true` into an empty output. The root lane then degraded to the project-less emit; the dependency lane to one root build per reached file. Both now switch it off.

**Mutation checks** (each fix reverted alone)
- Always placing the tsconfig beside the real one fails `…_without_writing_into_the_package`.
- Ignoring `${configDir}` fails `…_resolves_config_dir_paths_…`.
- Removing the synthetic `noEmitOnError: false` fails `…_root_whose_config_sets_no_emit_on_error`. Its first version passed anyway, because the isolated fallback rescued it; the legacy decorator probe now tells the two apart.
- Removing the dependency lane's flag fails `…_dependency_whose_config_sets_no_emit_on_error`. Its first version also passed through the root-lane rescue; the program now looks for the unimported file's emit in the manifest's dependency cache.

**Finding (fixed in `73fae775`)**
1. A private tsconfig also moves native plugins' config-file discovery, which anchors at the tsconfig's directory. `pluginConfigDir` now names the real directory, as the bundler adapters' overlay already does.

samchon (commented):
## Individual Self-Review: `d899c10d` test(ttsx): mirror the fake host's output below the entry rootDir

**Change integrity: this changes an existing test fixture.** The native lane failed `test_plugin_corpus_ttsx_register_executes_source_plugin_output_end_to_end` with "emitted entry not found". The fixture's fake host wrote `outDir/main.js` for `test/main.ts`. No compiler produces that layout: the trailing-stem matching that #1382 removed used to accept it. The test's assertion, that the transformed output is what runs, is unchanged. Only the host now writes where tsgo would: the source's path relative to the synthesized config's `rootDir`, below `--outDir`.

A real sidecar that writes output itself has the same obligation, so `pitfalls.mdx` now states it. I checked every other case that runs a native host under ttsx or the register hook. The utility-plugin cases print through tsgo. The one fake compiler, in `test_ttsx_executes_javascript_emitted_by_the_consumer_local_tsgo`, writes `<outDir>/src/index.js` for `src/index.ts` with `rootDir: "."`, which is the mirrored path.

samchon (commented):
## Individual Self-Review: `73fae775` fix(ttsx): anchor plugin config discovery for a private root tsconfig

**Checked without findings**
- `pluginConfigDir` is set only for a private tsconfig. A caller's own value still wins, and a beside tsconfig leaves discovery where it was.
- The value is the real config's directory, the directory discovery would have used with a beside tsconfig.
- `test_banner_ttsx_discovers_an_installed_package_root_config` gives an installed package root `@ttsc/banner` with a `banner.config.cjs`. The program reads the root's emit from the manifest's dependency cache. With the anchor removed it prints `bannered=false`.

samchon (commented):
## Individual Self-Review: `4b9b4775` test(ttsx): pin the root lane for a project whose build emits nothing

This is the case from the `fcc3d73d` review. The fixture's owning tsconfig lists no files, so the project build fails with nothing emitted. The decorator probe tells the root lane (3 arguments) from the isolated emit (2). I reverted `fcc3d73d`'s change and the case failed; with it restored the case passes. No findings.

samchon (commented):
## Individual Self-Review: `de668b2b` fix(ttsx): attempt a failing owning project's build once per process

The Overall Self-Review round 2 found this. `fcc3d73d` sends a file on to the root lane when its owning project's build produced nothing. `ensureProjectBuilt` memoized only successes, so every file that project owned re-ran the failing project build before its own root build.

**Checked without findings**
- A failure is memoized exactly like a success: per tsconfig, per process. A later file in a project whose build failed goes straight to its root lane, which gives the same result as before.
- A lock timeout, which may be transient, is memoized too. The root lane still serves the file correctly, so the cost is a root build instead of a project build.
- Mutation check: the one-off probe counts retired lock generations of the failed project's key in the manifest's dependency cache. It shows 2 build attempts with the memo removed and 1 with it.

**Test**
`test_ttsx_builds_a_failing_project_once_for_every_file_it_owns` counts the compiler invocations for the owning tsconfig through a logging wrapper around the real compiler. A script cannot stand in for the compiler binary on Windows, so the case runs on POSIX CI. The lock-generation probe above is too tied to lock internals to be a committed assertion.

samchon (commented):
## Individual Self-Review: `b609f7fe` fix(ttsx): find the referenced project of a file written at run time

The Overall Self-Review round 2 found this, and a reproduction confirmed it before the fix. Under a Vite-style solution, a program that reached one file of `tsconfig.app.json` and then wrote `src/generated.ts` ran the generated file with the solution's empty options: its decorator probe reported 2 arguments instead of 3.

**Checked without findings**
- The cache answers every file that existed when the list was taken. Only a file whose change time is at or after `takenAt` triggers a second `--showConfig`, and that refresh replaces the cached entry, so the next generated file compares against the new time.
- `takenAt` is read before the compiler is spawned, so a file written during the spawn errs toward a refresh.
- The change time cannot be set by a program, unlike the modification time, so a file copied in with a preserved timestamp still refreshes.
- A file that does not exist returns `false` and never refreshes.

**Test**
`test_ttsx_runs_a_generated_file_with_the_referenced_project_that_owns_it` failed with `generated=2` before the fix and passes after it.

samchon (commented):
## Individual Self-Review: `0615df74` docs(ttsx): describe the launcher's child and restore a moved JSDoc

Comments only. The Overall Self-Review round 2 found:
- two stacked JSDoc blocks above `terminalRequest`;
- a summary that still described executing the compiled entry.

Each JSDoc now sits on its own function. The summary names what the child is since #1402 (the source entry as Node's main module, served by the hooks) and keeps the import-extension rewrite, which still applies. No findings.

samchon (commented):
## Overall Self-Review, round 2: `779c60c9...4b9b4775`

A fresh pass over the whole base-to-head change, run while the final CI and local suites ran.

**How the surface was covered**
- **The #1407 refactor (`779c60c9..9dafd501`).** Every declaration in `packages/ttsc/src` was matched by name before and after, with trivia, JSDoc, modifiers, and namespace qualifiers normalized.
  - 1,029 declarations are token-identical.
  - The 11 that differ are all accounted for: one more `..` in a `__dirname`-relative path where the file moved one directory deeper (each checked against its new location), the `IJavaScriptRuntimeCapabilities` rename, and semicolons.
  - Every `__dirname` join at head either sits in a file that did not move or gained exactly the depth its move added.
  - The Go, script, generator, and documentation hunks in that range are path and comment updates. The one run-time module a Go test loads still exports the name it calls.
- **The formatting commits (`9b88f4c5`, `a08c34ad`).** All 188 files have identical token streams, modulo redundant parentheses, a union's leading `|`, trailing commas, and import order.
- **Every behavioral change after `9dafd501`.** I read every changed or added declaration in its final form: 36 changed, 49 added, 21 removed. I also read `execute.mdx` in full and scanned the added and changed tests for global-state, port, and timing hazards.

**Findings (all fixed)**
1. A project build that produced nothing was attempted again for every file it owned, because only successes were memoized (`de668b2b`).
2. The root files of a referenced config were cached per process. A file the program wrote afterwards fell back to the solution's empty options; reproduced as `generated=2` (`b609f7fe`).
3. `runTtsx.ts` had a misplaced JSDoc and a summary that predated #1402 (`0615df74`).

**Found by the final suites and CI, fixed before this round's findings**
`449a277a`, `1aef145d`, `423b7a4f`, `d899c10d`, `73fae775`, `4b9b4775`. Their Individual Self-Reviews give the evidence.

**Pinned verification items**
- **ESM entry.** An ESM-emitting project with no package `type` runs through Node's own main-module path, top-level await included, on Node 22.15.0 (the engines floor) and 24. A CommonJS entry keeps `require.main === module` and its argv on both.
- **`--watch` suggestion.** `node --watch --require ttsc/register src/main.ts` runs, restarts on an edit, and keeps `require.main === module`.
- **POSIX signals.** `test_ttsx_forwards_termination_signals_and_cleans_up_on_posix` passed on Ubuntu in the core lane. The core lanes do not run on macOS, so that is the POSIX coverage CI provides.

**Local suite**
The remaining local failures are the known Windows environment set, and each one passes on CI:
- symlink creation `EPERM`;
- a fake `tsc.exe` that cannot spawn (`UNKNOWN`);
- `npm pack` spawning;
- the coverage source-map cases.

Round 3 follows on the new head.

samchon (commented):
## Individual Self-Review: `f973904f` docs(ttsx): drop a reference to the removed isWithin helper

Comment only; round 3 found it. The JSDoc now describes the rejected test ("whether the file lies under `rootDir`") instead of naming a module this pull request removed. Prettier's reflow of the paragraph is the only other change. No findings.

samchon (commented):
## Overall Self-Review, rounds 3 and 4: `779c60c9...f973904f`

**Round 3** was a fresh pass over the whole surface after round 2's fixes. I re-read these in their final form:
- `buildSingleRootProject`, including its placement and anchor rules;
- `parseFlags`;
- `resolveOwningProjectConfig`, including the new cache refresh;
- `owningModuleOptions`;
- the runtime lanes.

It also searched the repository for references to every symbol this pull request removed, and re-ran the evidence repository case on the final source (passed).
- **Finding (fixed in `f973904f`):** `owningModuleOptions` explained its rule by citing `isWithin`, a module this pull request deleted. The other hits for that name are the path-identity context's own method.

**Round 4** started fresh again:
- The equivalence proofs for the refactor range and both formatting commits were re-run and are unchanged.
- The behavioral declaration inventory since `9dafd501` differs from round 2's only by round 2's own fixes (`ensureProjectBuilt`, `failedProjects`, `RootFiles`, `createdSince`), which round 3 re-read.
- The new tests inherit the process environment through `TestProject.spawn`, so the POSIX wrapper case sees `PATH` and the rest.

Round 4 found nothing to improve.

**CI evidence on `0615df74`.** The core lane ran the new cases on Ubuntu, all passing: the failing-project build count through the compiler wrapper, the generated file under a solution, the three emit-only placement and `noEmitOnError` cases, and the POSIX signal case. `f973904f` changes one comment, and CI now runs on it as the final head.

samchon (commented):
## Individual Self-Review: `7fdb8dea`, `81d1f7a9`, `98b25fe5` (the runtime suite on every platform)

These three commits answer one question: does ttsx behave on macOS and Windows as it does on Linux? The `experimental` platform rows ran only a one-line `npx ttsx` smoke. The 167-case runtime suite ran only in the Linux core lane.

**`81d1f7a9` ci: the runtime platform task**
- `features/ttsx-runtime` now runs on the macOS and Windows representative rows. Linux already runs it in the core lane, so it is not repeated there.
- The task is selected by any `packages/ttsc` change, a change to the suite itself, the shared test utilities, and every full plan.
- The rows build with the `experimental` scope and get the system Go binary, the same setup the watch rows use.
- `validation-plan.test.cjs` pins:
  - which rows run it for a compiler change and for a suite change;
  - that other feature suites keep their Linux-only topology;
  - the workflow step and its `TTSC_TEST_DIR`.

**`7fdb8dea` test harness: correction of an earlier classification**
My round 2 record called the five local source-map failures environmental. They were a harness defect. For a path with a `~`, Node records a script URL with the `~` literal but keys the source-map cache by `pathToFileURL`, which encodes it as `%7E`; plain Node reproduces this. The harness matched the two by exact string, so it would have failed on GitHub's Windows runners too (`RUNNER~1`). Both are now matched by the path they decode to.

The consumer-local compiler case installs a Node script under the name `tsc.exe`. Windows cannot run that, so the case runs on POSIX.

**`98b25fe5` the first macOS run**
Only one case failed, and it predates this pull request. It expected the logical temp path, but Node reports the resolved `/private/var/...`, under `ttsx` exactly as under `node`. The expectation is now `fs.realpathSync(root)`. That form also keeps a Windows 8.3 name as Node reports it, and the case passes on Windows locally.

**Results**
- Locally on Windows, 166 of the 167 runtime cases pass.
- The remaining case creates a file symlink, which needs a privilege this workstation lacks. The case already fails loudly on such a host, and I left it that way rather than make it skip silently: the Windows runner has the privilege, and a silent skip would weaken what CI proves.
- On `81d1f7a9`, the Windows rows passed the suite. The macOS row failed only on the case `98b25fe5` fixes.

samchon (commented):
## Discovery rounds 4–7 and Individual Self-Reviews: `84748aeb`…`123f5aa3`

The user asked for every unexplored area to be investigated, and for this pull request to merge only after discovery saturates, a clean Overall Self-Review round passes, and CI is green. The rounds are recorded in the campaign ledger. This review summarizes them and the commits they produced.

### Discovery rounds

- **Round 4**
  - Areas: worker threads, `fork`, Yarn PnP (Yarn 4.18), Bun, Deno, the whole runtime suite on Node 22.15.0 (166/167), ESM edges, concurrency, a 3000-module project, nodemon, a read-only tree, and the in-memory root-replacement design.
  - Accepted: #1413 and #1414.
  - Rejected with evidence:
    - PnP zipped TypeScript dependencies: TypeScript 7 cannot resolve into zips.
    - A child the program starts with a bare environment: it drops `NODE_OPTIONS` on its own, and plain Node behaves the same.
    - Watcher churn: no nodemon restart could be reproduced.
  - Scale: head 19–25 s against 78 s at the base.
- **Round 5**
  - Areas: stdin, exit codes, rejection, stack traces from out-of-include roots, JSON, nested ttsx, `.mts`/`.cts`, case and `..` spellings, a JSX root, tricky argv, and 20 forwarded compiler flags.
  - Accepted: #1415, #1416, #1417, and the read-only default cache (#1414).
- **Round 6**
  - Areas: a spaced install path, an out-of-include/project require cycle, response files, and read-only `ttsc/register`.
  - Found: the response-file gap in the #1417 fix (`123f5aa3`).
- **Round 7**
  - Areas: the full axis map (CLI, discovery, build lanes, runtime, process, platforms, caches) re-walked, and a jsconfig-only project probed.
  - Found nothing.

### The in-memory design (asked for on 2026-09-19)

I implemented it and reverted it. The design parses the tsconfig where it lives and replaces its roots in `driver.LoadProgram`; the Go part and its tests worked. It needs a compiler host to run in, and the shipped platform `ttsc` binary refuses `build` and `check`, because compilation happens in the JS CLI through tsgo or in plugin-built sidecars. So adopting it means shipping a new binary, a packaging decision outside this pull request. #1414 records this, and the synthesized-config JSDoc explains why no other placement is faithful.

### Commits

| Commit | Issue | Verification |
| --- | --- | --- |
| `84748aeb` | #1413 Bun/Deno refused by the launcher and the preload | Real Bun 1.4.2 re-run: both refuse with the new message; unit case |
| `43bf5897` | #1414 read-only project | Windows ACL end-to-end cases for both halves; each fails with its fix reverted |
| `1c6b2d46` | #1414 follow-up | `43bf5897` committed the cache helper twice (a keep-index stash round trip re-applied the insertion), so the package stopped compiling. Every pushed file was then scanned for duplicated top-level declarations: only this one |
| `44ca4a97` | #1415 `--noEmit`/`--emitDeclarationOnly` | Base reproduces the failure; the case fails with the override reverted |
| `3ce6c0cf` | #1416 `--inlineSourceMap`, #1417 forwarded `--module` | Both cases fail with their decision reverted to the config |
| `123f5aa3` | #1417 follow-up | Classifies the entry-only build by the project, since the synthesized config is gone before `--showConfig` runs. The response-file case fails on the prior head and passes now |

### Checked without findings

- Check passes now put `--noEmit` after the forwarded flags, so the unconditional `--noEmit false` of runtime builds cannot turn a check pass into an emit.
- Check-stage plugin hosts only analyze.
- A bare boolean flag reads as on only when nothing, or another flag, follows it.
- Runtime roots and dependency builds take no forwarded flags, so their config-based classification matches their emit.

The local runtime suite on Windows and CI on all three operating systems run on this head.

samchon (commented):
## Overall Self-Review, final round: `779c60c9...123f5aa3`

This is a fresh pass over the whole surface on the head that merges.

- **Refactor and formatting commits.** The equivalence proofs were re-run and are unchanged.
  - Refactor range: 1,029 declarations identical, 11 accounted for.
  - Formatting commits: 167 and 21 files token-identical.
- **Behavioral change since `9dafd501`.** The declaration inventory differs from round 4's only by this cycle's later fixes:
  - `checkNodeRuntimeSupport`, `NON_NODE_RUNTIMES` (#1413);
  - `unwritableConfigDirectory`, `defaultRuntimeCacheDir` (#1414);
  - `runtimeEmitProfile`, `isOn` (#1416, #1417).

  I read each in its final form, together with the changed `runTsgo`, `readEffectiveCompilerOptions`, `runtimeCompilerArgs`, and `buildEntryProject`.
- **Tests, CI configuration, and documentation.** The validation plan's new runtime task and its pinned cases, the harness fixes, the six new runtime cases, and the `execute.mdx` additions.
- **Findings:** none.

**CI on `123f5aa3`:** 29/29.
- The runtime suite ran on darwin-x64 and win32-x64 as well as in the Linux core lane.
- All six cases added in rounds 4–6 recorded real timings on all three operating systems, the read-only cases included.

**Close lines, reconciled against `git log origin/master..HEAD`:**
- Commits: #1382 and #1401–#1409 (ten `Close` lines, no reverts in the range).
- This description: #1413–#1417.
- The in-memory root-replacement attempt was never committed.

**Merge:** the user authorized merging once discovery saturates, a clean round passes, and CI is green. Discovery round 7 was empty, this round is clean, and CI is green, so this pull request merges with the repository's squash method.

Issue #1428, opened 2026-09-19
unplugin: a watch-broker drain that times out is trusted as proof


samchon:
## Problem

A delivery settles its generation's trackers before it reads their verdict. A tracker in the watch broker settles by round-trip: the child answers `drain` after its own loop has run, and the ordered IPC channel puts every event it sent before the answer ahead of the answer (#1272).

`drainWatchBroker` also resolves after a fixed 10 ms fallback, whether or not the child has answered, and the delivery then reads the trackers exactly as if it had. When the round-trip takes longer, because the host's loop is busy, the child is forwarding a burst, or the machine is loaded, an edit whose event is still in flight is missed. The tracker's silence is taken as proof, the delivery serves the cached output, and nothing re-requests the module when the event arrives later.

## Resolution

1. A drain reports whether it was answered.
2. An unanswered drain marks the tracker unverified. Its silence then proves nothing until one delivery proves the recorded state by reading it, the rule a dropped-events gap already follows.
3. The fallback exists so that a child that never answers cannot hold a delivery. Once missing it can no longer cost correctness, it waits long enough that a busy but healthy child still answers within it.

## Acceptance

- A drain the child answers settles the tracker as before.
- A drain that times out marks the tracker unverified, and the delivery validates the generation against its recorded state.
- A broker that died releases every waiting drain at once, unanswered.

Issue #1426, opened 2026-09-19
fix(unplugin): inotify overflow is invisible to Linux watches


samchon:
## Problem

An inotify instance holds at most `fs.inotify.max_queued_events` unread events, 16384 by default. When the queue is full, the kernel drops every later event and queues one `IN_Q_OVERFLOW` event with watch descriptor `-1`.

libuv discards that event. In `src/unix/linux.c`, `uv__inotify_read` runs `w = find_watcher(loop, e->wd); if (w == NULL) continue;`. `@parcel/watcher` discards it too. A Linux watch opened through `fs.watch` can therefore lose events without any notice, and the adapter reads that watch's silence as proof.

Every `fs.watch` in the host process shares one inotify instance per event loop, including the host's own watchers. A burst can fill the queue, especially while the event loop is busy compiling: a large checkout, an install, or a generator writing thousands of files.

## Resolution

1. Add a hidden `ttsc __watch` command to the native binary.
   - On Linux it owns one inotify instance and speaks newline-delimited JSON over stdio.
   - It reports `IN_Q_OVERFLOW`.
   - It uses only `golang.org/x/sys/unix`, with no cgo.
2. Open every Linux directory watch of the adapter through that helper:
   - the directory observer's watches;
   - the exact-input trackers' non-recursive watches.

   An overflow reaches every subscriber as an unattributed mutation.
3. A subscription is live only once the helper acknowledges it:
   - trackers wait for their initial subscriptions;
   - a directory watched later is read, and its entries announced, only after its acknowledgement;
   - a Vite serve scope re-checks its entries once its initial subscriptions are live.
4. Draining a tracker makes the helper read its instance empty before it answers. An edit made before the drain has therefore reached the tracker.
5. When the helper cannot start, no Linux watch is trusted. Generations then validate against their recorded state, and Vite serve scopes fall back to the bounded poll.

## Acceptance

- Go tests cover the helper's event mapping:
  - shared watch descriptors;
  - removal;
  - self-deletion;
  - overflow;
  - sync.
- On Linux, the existing observer and tracker scenarios pass through the helper, and an overflow reaches every subscription as an unattributed mutation.
- Without the helper, no Linux watch is trusted.

Issue #1425, opened 2026-09-19
fix(unplugin): FSEvents drops are invisible to macOS watches


samchon:
## Problem

FSEvents can drop events. When it does, it marks the next event it delivers with `kFSEventStreamEventFlagMustScanSubDirs` together with `UserDropped` or `KernelDropped`. libuv discards every event that carries a dropped flag. In `src/unix/fsevents.c`, `uv__fsevents_event_cb` runs `if (flags & kFSEventsSystem) continue;`. A macOS watch opened through `fs.watch` can therefore lose events without any notice, and the adapter then reads that watch's silence as proof that its inputs are unchanged.

#1418 made macOS watches immune to the loss libuv causes by re-creating its shared stream. Drops by the FSEvents client or by the kernel remain invisible.

## Resolution

1. Watch through the `fsevents` binding inside the watch broker on macOS.
   - The binding creates one FSEventStream per watch, and `FSEventStreamStart` runs inside `watch()`.
   - It passes every event's flags through.
   - It becomes an optional dependency, which npm, pnpm, and Yarn install on macOS by default.
2. Report an event that carries `MustScanSubDirs`, `UserDropped`, `KernelDropped`, `EventIdsWrapped`, `RootChanged`, `Mount`, or `Unmount`, or names a path outside the watched directory, to the registration of that watch as a gap. A generation then stops trusting its watchers' silence until one delivery proves its recorded state by reading it, and a Vite serve scope re-checks every input it covers.
3. Each watch has its own stream, so opening or closing one disturbs no other watch. A registration is ready once its watches return, and no stream swap needs proving, so the shared-watch and probe machinery of #1418 is removed.
4. When the binding cannot be loaded, the broker cannot see a drop, so it reports its macOS registrations failed. Generations then validate against their recorded state, and Vite serve scopes fall back to the bounded poll.

## Acceptance

- A macOS registration receives events through the binding.
- A drop flag reaches the registration of the watch as a gap.
- Without the binding, no macOS registration is trusted.

Issue #1424, opened 2026-09-19
fix(unplugin): a Windows watch overflow reaches no rename-only tracker


samchon:
## Problem

On Windows, libuv reports a watch whose `ReadDirectoryChangesW` buffer overflowed as one change event with no filename. `src/win/fs-event.c` calls `handle->cb(handle, NULL, UV_CHANGE, 0)` when the completed read carries no data. Node surfaces it as `('change', null)`. Anything below the watched directory may have changed.

The adapter drops that notice on two paths:
- The watch broker forwards an event only when the registration asked for every event, the event is a rename, or the location is recursive. A rename-only registration, such as the candidate tracker, never hears an overflow.
- The shared classifier in `createHostInputMutationTracker` drops every non-rename event for a rename-only tracker before it looks at the filename. For a tracker that hears every event, it records an unattributed change as a content change, not as a possible new entry.

A candidate that appears during an overflow is therefore never heard, and the generation keeps serving the resolution it made before.

## Resolution

1. Treat an event without a filename as an unattributed mutation on every backend: anything below the watched directory may have been created, removed, or changed.
2. Forward it from the broker child for every registration, whichever events the registration asked for.
3. Classify it as a mutation before the rename-only filter runs.

## Acceptance

- A rename-only brokered registration receives an unattributed change event.
- The classifier records an unattributed event of either type as a mutation, for rename-only trackers and trackers that hear every event alike.

PR #1429, opened 2026-09-19
fix(unplugin): close every residual host-observation gap

samchon:
## Intent

#1400 made every host observe exactly what the compiler read. This PR closes the residual gaps in that observation, where an input could change without any host re-running the module. Each gap is its own commit and issue:

| Issue | Gap | Resolution |
| --- | --- | --- |
| #1424 | Only an all-events broker registration heard a Windows `ReadDirectoryChangesW` overflow, which libuv reports as an unattributed change; rename-only trackers dropped it. | An event without a name is a mutation for every tracker and registration. |
| #1422 | Turbopack fails a whole module whose dependency lies outside its project filesystem root, so a `typeRoots` entry or plugin dependency beside the workspace answered the page with an error. | The loader resolves Turbopack's root as Next does and hands Turbopack only the inputs inside it. `withTtsc` passes the configured root. A module with an input left out depends on a per-process marker, so a persisted result is never reused in a later process. |
| #1423 | Turbopack takes a dependency's baseline only when it processes the loader's result, so a change between the compile's read and that moment was absorbed. | Under `next dev`, every input goes to the bridge as well. It proves registration against the compile's read, and rewrites a stale module's sentinel with growing delays until the module runs again. The run is acknowledged across the worker pool through a shared run record. |
| #1425 | libuv discards FSEvents' dropped-events flags on macOS. | The broker watches through the `fsevents` binding (a new optional dependency), one stream per watch. A drop, wrapped id, changed root, mount, or unplaceable path reaches the registration as a gap. The #1418 shared-stream probe machinery is removed, since no stream is shared any more. Without the binding, macOS registrations are reported failed. |
| #1426 | libuv discards inotify's `IN_Q_OVERFLOW` on Linux. | A hidden `ttsc __watch` command in the native binary owns one inotify instance and reports overflow. Every Linux watch of the adapter lives in it. Watches go live on acknowledgement, directories are read after that, trackers drain through a `sync` that reads the queue empty, and Vite scopes re-check once live. Without the helper, no Linux watch is trusted. |
| #1428 | A broker drain that timed out after 10 ms resolved as if answered, so a busy host could serve an edit's stale output. | A drain reports whether it held. An unheld one leaves the tracker unverified, and the delivery proves its recorded state by reading it. |

## Scope

- `packages/unplugin`: the tracker, broker, bridge, Turbopack, Next, and Vite serve watch paths, plus `fsevents` as an optional dependency.
- `packages/ttsc`: the new `internal/watchhelper` package and the hidden `__watch` dispatch in `cmd/ttsc`. It uses `golang.org/x/sys/unix` only, with no cgo, and exits 2 on non-Linux.
- Tests:
  - Features: 45 → 46.
  - Native: 221 → 226.
  - Go tests for the helper, registered in `test-owners.cjs` and `test-go-ttsc.cjs`.
  - The three scenarios that open real Linux watches moved to `native-plugins`, since the `features` lane builds no native binary.
- Docs: the unplugin website page (Turbopack root and baseline, macOS drops, Linux helper, drain barrier) and the README Turbopack paragraph.

- `tsc` for `@ttsc/unplugin` and `tests/test-unplugin` passes.
- The `features` tree passes. The `native-plugins/adapters` tree passes, except the Bun scenario, which needs `bun`.
- The new and moved scenarios run individually: Turbopack baseline, bridge confirmation, broker ready/gap/fake-fsevents, drain timeout, and the Vite subscription scenario.
- The 75 CI script tests pass, including ownership, the test contract, and structure, and `scripts/ci/format-check.cjs` passes.
- `go vet` and `go test -c` for `internal/watchhelper` pass with `GOOS=linux`.
- Not run locally: the Linux helper's runtime scenarios and the real FSEvents scenario. No Linux or macOS host is available here, so they are left to the Linux lanes and the `bundler-defenses-macos` lane.

## Individual Self-Review

Each implementation commit has a review record, posted below as formal reviews.

https://claude.ai/code/session_01491ZRN4aGMhrJFdvHj4gHp


samchon (commented):
Individual Self-Review (solo).

Scope: `fix(unplugin): count an unattributed watch event as a mutation for every tracker` (#1424).

Checked:
- Source evidence: libuv 1.52.1 `src/win/fs-event.c` calls `handle->cb(handle, NULL, UV_CHANGE, 0)` when a completed read carries no data.
- Every consumer of an event without a name:
  - Broker child filter: now forwards it.
  - `routeWatchBrokerMessage`: a missing directory is a membership change. A present directory with a null name goes either to `classify` or to the project filters' fall-through, which records a mutation.
  - `createHostInputMutationTracker.classify`: returns mutation before the rename-only filter.
  - In-process project tracker: records a mutation on the root.
  - Linux observer: passes the null through.
  - Vite serve scope: already re-checked every entry for a null file of any type.
- Cost: an unattributed event is rare, since a backend emits one only when it cannot name what changed. Counting it as a mutation re-proves the generation, which is the conservative answer.
- Tests:
  - The new harness runs the child program in-process, and its scenario fails on the old filter.
  - The relevance scenario's new rows fail on the old classifier.
  - The routing scenario still passes. The inventory is 44.

Verdict: clean.

samchon (commented):
Individual Self-Review (solo).

Scope: `fix(unplugin): keep the Turbopack loader's dependencies inside Turbopack's root` (#1422).

Checked:
- Root resolution matches Next 16.3.4's `dist/server/config.js`:
  - explicit `outputFileTracingRoot` first, then `turbopack.root`;
  - `NEXT_PRIVATE_OUTPUT_TRACE_ROOT`;
  - `findRootDirAndLockFiles(dir).rootDir`, required from the project's own `next`.
- Fallbacks:
  - An absent `turbopackRoot` (hand-written rule) keeps to the project directory.
  - If Next's function throws, the project directory is used too. Every Turbopack root contains it, so no path Turbopack rejects is ever registered.
  - `withTtsc` always passes the explicit root or `null`. A relative root is resolved against cwd, as Next does.
- Acceptance uses lexical containment (`pathIsWithin`), the same judgement Turbopack's runtime makes: `relative(cwd, file)` joined onto the project path. A path on another drive is never within the root.
- Refused inputs:
  - They never reach the channels.
  - With a bridge they are observed there.
  - `untracked` fires once per registration, after the accepted inputs.
  - The marker's directory uses the bridge prefix and pid. So it is removed at exit and swept after a crash, and a later process always sees a missing dependency and re-runs the module.
- Bridged kinds are unaffected: listing and membership still go to the bridge.
- The sentinel and marker live under `rootContext/node_modules/.cache/ttsc`, which is always inside the root.
- Cost: root resolution is memoized per worker by project and configured root.
- Tests:
  - resolution rows, including a fake `next/dist/lib/find-root`;
  - channel and bridge rows;
  - the marker identity;
  - the native scenario, where a plugin-reported sibling dependency is kept out and the marker registered;
  - the wiring scenario's configured-root precedence.

  All 10 Turbopack loader scenarios, all `features` scenarios, and the 75 CI script tests pass locally.

Verdict: clean.

samchon (commented):
Individual Self-Review (solo).

Scope: `fix(unplugin): re-run a Turbopack module whose input changed before Turbopack's baseline` (#1423), with the follow-up `3998f974d`.

Checked:
- Windows covered:
  - From the compile's read to registration: `replace(..., startedAt)` re-checks any entry with a change recorded after `begin()`, a scope opened after it, or a fallback poll. The broker's readiness delivers an unattributed recheck for its scope.
  - An event that is still in flight at registration arrives later and signals then, and the confirmed schedule covers it the same way.
  - From registration on: native events or the bounded poll, then `check`, then `invalidateModule`, then `signal`.
- Confirmed schedule:
  - Rewrites land at 50, 250, 1,000, 4,000, and 16,000 ms.
  - A repeated signal keeps the running schedule.
  - `acknowledge` clears it, `close` settles every schedule before disposing, and a stale timer checks that it is still the current schedule before writing.
- Cross-worker acknowledgement:
  - A run token (a random UUID) lives in `ttsc-watch-bridge-<ppid>-runs`. The sweep removes that directory once the host process is gone.
  - The comparison is by token difference, taken at signal time, so it does not depend on clock order.
  - A run starting after the read necessarily read the signalled change.
  - An unreadable or unwritable record keeps every rewrite, so only efficiency depends on it.
- Non-Turbopack hosts are untouched: `confirmDelivery` defaults to false, so they keep the immediate write.
- Dual registration (`ignores: () => true`) still hands every input to Turbopack's channel, so its own watcher and persistent cache are unchanged.
- Finding, fixed in `3998f974d`: `signal` was exposed on `HostWatchBridge` with no external caller once the post-return recompile turned out to be subsumed by the registration proof. It is internal now. The issue body was updated to the implemented resolution.
- Tests:
  - the features scenario covers an immediate bridge, confirmed repetition, acknowledgement on the same bridge and on another sharing the directory, and close with rewrites still owed;
  - the native Turbopack scenario edits the declaration after the loader returns, and asserts a rewrite, a repeat, and a stop after the next run. Both fail without the change.
  - The features and native adapter trees pass locally, except the Bun scenario, which needs `bun`.

Verdict: clean after `3998f974d`.

samchon (commented):
Individual Self-Review (solo).

Scope: `fix(unplugin): hear FSEvents' dropped-events notice on macOS` (#1425).

Checked:
- Binding resolution:
  - `createRequire(import.meta.url).resolve("fsevents")`. The CommonJS build rewrites `import.meta.url` through `pathToFileURL(__filename)`, as verified in `lib/.../fseventsBindingPath.js`, so either build resolves from its own copy of the package.
  - Under pnpm the optional dependency is linked into `packages/unplugin/node_modules` on macOS, so the macOS lane resolves it. The new real-backend scenario asserts that it does.
- Child program:
  - Registrations own their watches. `ready` follows the synchronous `watch()` calls, whose `FSEventStreamStart` runs inside `watch()`.
  - `realpathSync.native` places events below the real root. The root's own events are skipped, as libuv skips them.
  - A non-recursive registration drops nested paths. The flag mapping is libuv's: CHANGE only for a modification with no create, remove, or rename.
  - `DROPPED` (MustScanSubDirs, UserDropped, KernelDropped, EventIdsWrapped, RootChanged, Mount, Unmount) and unplaceable paths send `gap`. `HistoryDone` is excluded, since it only follows a `since` replay, which is never requested.
- Failure paths:
  - A `null` or unloadable binding reports `failed` with `ready` and opens nothing.
  - A throwing `realpath` or `watch` fails the registration.
- Parent: `gap` keeps its route. Trackers become unverified, and the Vite scope rechecks through `openIsolatedRecursiveWatch`. Comments that described the swap proof now describe the per-stream backend.
- Removed: the shared handles, `HANDLE_BOUND`, and the probe, swap, and sweep. Nothing references them. Probe directories that an older broker left after a crash are no longer swept. They are empty temp directories, which the OS temp policy cleans.
- Windows and every other non-macOS path is unchanged: `fs.watch` per registration, as the former non-shared mode.
- Tests:
  - the fake-binding scenario: ready, the real root, mapping, recursion filtering, gaps, stop on remove, and failure without a binding;
  - the real-backend scenario: after-ready events for two registrations on one directory, nested events only to the recursive one, and no gaps. It runs in the macOS lane through `--include=watch_broker`.
  - The features tree and the 75 CI script tests pass locally.

Verdict: clean.

samchon (commented):
Individual Self-Review (solo).

Scope: `fix(unplugin): trust no tracker silence behind an unanswered drain` (#1428).

Checked:
- Every release path reports its cause:
  - `drained` routed → `true`;
  - fallback timer, a failed `send`, or broker death (`fail`) → `false`.
  - A dead broker has already failed its trackers, so `false` there changes nothing beyond releasing the waiters.
- `broker.draining` shares one boolean among concurrent waiters, and each settling tracker applies it to itself. `settleMutationTrackers` marks `unverified` only on `false`. The in-process drain resolves `true` after its two turns, as before.
- Every reader of silence already honours `unverified`:
  - `trackerProvesInputUnchanged`;
  - `notificationsProveMembership` (project, host-input, and candidate trackers).

  `matchesCompleteInputSnapshot` clears it once a delivery has proven the recorded state.
- Reference counting: `pendingDrains` and the channel's ref/unref are unchanged, and the release is idempotent (`settled`).
- Cost: a healthy child answers in about a millisecond. The one-second bound applies only to a child that stopped answering, and missing it now costs one validation, not a stale delivery.
- Tests:
  - an answered drain, an unsendable one, and an unanswered one (which waits about 1 s and cleans its bookkeeping);
  - `settleMutationTrackers` marking only the unheld tracker.
  - The routing test still passes with the new release signature.

Verdict: clean.

samchon (commented):
Individual Self-Review (solo).

Scope: `fix(unplugin): watch Linux directories through a helper that reports overflow` (#1426).

Checked:
- Go helper (`internal/watchhelper`):
  - It uses libuv's mask and typing, where anything beyond ATTRIB|MODIFY, IN_ISDIR included, is a rename.
  - Nameless self events are dropped. DELETE_SELF, MOVE_SELF, UNMOUNT, and IGNORED report `gone` once per subscription and forget the descriptor, and a moved directory's watch is removed.
  - IN_Q_OVERFLOW (wd -1) is broadcast.
  - Shared descriptors map to sets of ids, and a descriptor is removed with its last id.
  - One poll loop covers the instance and a wake pipe. Commands are processed after the instance drains, and `sync` drains again before answering, so stdout order is event order.
  - Stdin EOF exits 0, and a flush error (client gone) exits.
  - Built with `x/sys/unix` only, no cgo. Non-Linux builds answer exit 2. It vets and cross-compiles for Linux.
- Go tests (Linux): shared descriptors and removal, sync ordering over 200 creations, self-deletion, an unwatchable path, stdin close, and the byte-level decoder (overflow, ISDIR rename, nameless, unknown descriptor, end). The package is claimed in `test-owners.cjs` and run by `test-go-ttsc.cjs`.
- Client:
  - The binary resolves as ttsc resolves its own: absolute `TTSC_BINARY`, then the platform package from the `ttsc` manifest, then the dev layout.
  - A binary that exits before answering is refused once per process.
  - On death, every sync is released unanswered and every subscription ends, so trackers fail and scopes poll.
  - References are counted, so outstanding replies keep the loop alive and idle helpers never do.
- Shared watch: one helper subscription per directory. `ready` is shared. An error, gone, or death ends it for every subscriber through `errors`, the old `fs.watch` error path. A close before the answer releases its reference, and the late answer is ignored.
- Observer:
  - A directory is read only after its watch is live, so entries created between the subscription and the read are read or heard.
  - The initial tree and each `track` are batches. A batch joins directories still opening for another caller, and a `track` batch reports the tracked path changed once live. This closes the compile-to-track window for directories outside the admitted tree.
  - Overflow notices are collapsed once per burst per observer.
- Trackers:
  - `settleOpenedDirectoryWatches` awaits readiness, bounded at 10 s, and fails the tracker on refusal or timeout.
  - The drain becomes the helper's sync; an unanswered sync makes the tracker unverified (#1428).
- Vite: a Linux scope re-checks all entries (unattributed event) once live. A thrown observer (no helper) fails the scope into the poll.
- Moves: the three scenarios that open real Linux watches go to `native-plugins`, since the `features` lane builds no binary. The features tests that remain all inject watch seams.
- Tests: the routing table (features) and the Linux native scenario (live trackers draining through sync, overflow mutating both trackers, helper death failing both, a missing binary failing a new tracker).
- Local: tsc, the features tree, the moved Vite scenario on Windows, the structure, contract, and ownership tests, and the format check all pass. Linux runtime behaviour is left to CI, since no Linux host is available locally.

Verdict: clean.

samchon (commented):
Overall Self-Review, round 1 (solo), on `c19d471c1`.

Surface: the whole diff from `d6d666ea1` (#1400) to the head, 77 files. I re-read it top to bottom, not per commit, with these cross-commit consequences in view:

- Option flow: `turbopackRoot` rides the rule options into the loader, and `resolveOptions` copies only its own three fields, so it reaches neither the compile nor its cache key.
- Turbopack:
  - Refused inputs, dual registration, and confirmed signals compose. A refused input reaches only the bridge. An accepted one reaches Turbopack's channel and the bridge.
  - Sentinels, the process marker, and the run record all live under the project's `node_modules/.cache/ttsc`, inside every Turbopack root.
  - The sweep removes only dead owners. The run record is keyed by the host's pid, which is alive while its workers are.
- Other build hosts: `openHostWatchBridge`'s new parameter defaults to immediate signals. `HostWatchBridge.acknowledge` exists on the one implementation and both test fakes.
- Linux Vite scopes:
  - A scope counts as live when opened, but the readiness recheck bumps the history floor and re-checks every owned entry. A registration whose compile started before the scope went live is therefore proven.
  - An observer that throws for want of a helper fails its scope into the poll.
- Linux trackers:
  - Readiness is awaited, bounded at 10 s.
  - The drain is the helper's sync. An unanswered sync marks the tracker unverified (#1428).
  - Helper death fails every tracker through the shared watch's `errors`.
- macOS and Windows broker: Windows keeps `fs.watch` per registration and its failure semantics. The macOS failure without a binding is reported with `ready`, so no registration waits.
- Packaging:
  - `fsevents` is optional and resolved lazily, so non-macOS installs skip it.
  - The helper is a subcommand of the binary the compile already resolves, through the same order (absolute `TTSC_BINARY`, then the platform package beside `ttsc`, then the dev layout). A binary without the command is refused once.
  - `check:dependencies` passes.
- Suites: the features tree needs no native binary on any platform, since every remaining watch-opening scenario injects a seam. The native tree and the macOS lane carry the real-backend scenarios.

Candidate considered and dismissed: `scripts/test-go-ttsc.cjs`'s header counts "six packages". It narrates the #622 history of the original six, and each later package carries its own comment in the list.

Result: no sound improvement found. Clean.

samchon (commented):
CI diagnosis on `47d3348cd`, fixed in `e574f4b74`.

`bundler defenses` (Linux) failed. The failing scenarios:

- the directory observer, the Linux trackers, and the helper scenario;
- the real-backend tracker;
- five `transformTtsc` persistent-validation scenarios, which expect live notifications.

All of them had one cause: `no Linux watch helper can watch …`. The platform packages and the dev layout are both built from `cmd/platform` (`scripts/build-platform-package.cjs`, `scripts/go.cjs build-native`), not `cmd/ttsc`. So the adapter resolved a binary with no `__watch`. The binary answered "unknown command", was refused, and no Linux watch was trusted.

The failure mode was the designed one: trackers failed and generations validated from recorded state. Nothing crashed and nothing served stale output. It only defeated the notification proofs those scenarios measure.

The hidden command now lives in `cmd/platform`, and it is removed from `cmd/ttsc`, which the adapter never resolves.

The macOS lane (real FSEvents through the binding) and the Windows lane passed on this head.

samchon (commented):
CI diagnosis on `1223fa98d`, fixed in `6f489dd2c`.

With the helper now served from `cmd/platform`, the Linux watch scenarios passed:

- the directory observer;
- the Linux trackers;
- the real-backend tracker, now through the helper.

The runner then exited during `test_linux_watch_helper_keeps_every_watch_honest` ("The runner exited before finishing"). The scenario stops the helper and awaits its `exit` event. An idle helper is unreferenced by design, so it never keeps a host alive, and nothing else held the loop. The loop emptied, and Node exited mid-suite.

The cause is in the test only. The adapter itself holds a reference exactly while a reply is outstanding (`referenceLinuxWatchHelper`), so a helper that dies while a registration or sync waits on it is always heard. The scenario now references the helper before stopping it.

samchon (commented):
Individual Self-Review of `22ba243cc`, `1f00155cd`, and `3d467c1d7`: removing every assumption about another program's internals or timing.

Audited the whole PR for hardcoded knowledge that stood in for a contract. Four places were not principled, and each is fixed:

1. **Turbopack's re-signal schedule** (`22ba243cc`). It stopped after a fixed 50 ms…16 s schedule, so it relied on Turbopack taking its baseline within 16 s. Each rewrite now waits four times as long as the one before, without end, until the module runs again. For any baseline delay, a later rewrite lands after it. A module never run again costs rewrites that grow only with the logarithm of the time it stays stale. The delay is capped at Node's maximum timer.
2. **Binary resolution** (`1f00155cd`, documented in `3d467c1d7`). `@ttsc/unplugin` carried a copy of ttsc's `resolveBinary` order. ttsc now shares it through a `ttsc/binary` subpath, on the precedent of `ttsc/path-identity`, and the adapter imports it. The copy is deleted. Both the CommonJS and ES module builds resolve it.
3. **Turbopack root** (`1f00155cd`). The loader read Next's internal `next/dist/lib/find-root` and `NEXT_PRIVATE_OUTPUT_TRACE_ROOT`. It now uses only the roots the configuration names: `turbopack.root` and `outputFileTracingRoot`, passed as `turbopackRoots`.
   - Every root that contains the project is an ancestor of it, so the deepest one lies inside whichever Turbopack uses.
   - With none named, the loader takes the project directory, which every Turbopack root contains.
   - Inputs beyond it reach the bridge and the per-process marker, so this costs caching, never correctness. The docs tell monorepos to set `turbopack.root`.
4. **Cross-worker run record** (`1f00155cd`). It was keyed by `process.ppid`, which assumed pool workers are children of their host, and let two unrelated hosts started from one shell answer each other. It now lives in the transform session `withTtsc` opens, which is exactly what one host's workers share. Without a session, each worker answers only its own signals.

Reviewed and kept, since each is a contract rather than an assumption:

- the drain and sync bound (1 s) and the readiness bound (10 s). Missing either proves nothing, and the tracker falls back to validation.
- the FSEvents flag values and libuv's inotify mask and typing. These are OS and library API constants.
- `FIRST_SIGNAL_DELAY_MS` (50 ms), which affects efficiency only.

There is no monkeypatching anywhere. Nothing overrides host objects, prototypes, globals, or module loading. The test harness runs the broker program with injected modules, which is dependency injection, not patching.

Tests:
- The root rows cover none named, one, two in either order, a relative root, and one that does not contain the project.
- The Next wiring asserts `turbopackRoots`.
- A bridge of another host, with its own runs directory, does not stop the rewrites.
- Local runs pass: the features tree, the bridge and Turbopack scenarios, and the 75 CI script tests.

samchon (commented):
Overall Self-Review, round 4 (solo), on `3d467c1d7`. CI is green on this head (29/29).

Delta since round 3: the unbounded re-signal schedule, `ttsc/binary`, public-configuration Turbopack roots, and the session-keyed run record (reviewed individually above).

Fresh pass over the whole surface:

- No reference remains to the removed concepts: `ppid`, the singular `turbopackRoot`, `find-root`, `NEXT_PRIVATE_*`, the fixed signal list, the copied binary resolver. The only `turbopackRoot` left is the loader's local variable.
- `ttsc/binary` appears in both `exports` maps (source and publish), is documented as public surface, and resolves from both the CommonJS and ES module builds of the adapter.
- The bridge's `localRuns` is bounded by the importer count. Without a session, a worker answers only its own signals, so two hosts can never cross-acknowledge.
- A relative configured root is resolved against the working directory, as Next resolves it. A named root that does not contain the project is ignored, so the loader never widens past the project directory on a misconfiguration.
- The Linux, macOS, and Windows lanes all pass on this head, including the helper, FSEvents, broker, bridge, and Turbopack scenarios.

Result: no sound improvement found. Clean. The merge gate is met.

Issue #1434, opened 2026-09-19
ttsc: the watch helper's overflow report is proven only on synthetic bytes


samchon:
## Problem

The watch helper (#1426) reports inotify's `IN_Q_OVERFLOW`. Its tests prove the decoding only on hand-built event bytes (`TestDispatchMapsEventsAsLibuvDoes`). No test makes the kernel overflow a queue, so nothing shows that the helper's real read path receives and reports the event the kernel actually queues.

## Resolution

Fill an unread inotify instance past `fs.inotify.max_queued_events` with uniquely named creations, so no events coalesce. Then drain it through the helper's own read and decode, and assert an overflow is reported.

## Acceptance

- On Linux CI, the Go test observes a real kernel overflow through `drain`.

Issue #1433, opened 2026-09-19
unplugin: the Turbopack baseline race is reproduced only against a fake loader


samchon:
## Problem

#1423's premise is that Turbopack takes a loader's dependencies as its baseline only when it processes the loader's result. That premise comes from reading the runtime embedded in Next's native binary. The fix, dual registration with confirmed re-signals, is proven only against a fake loader context (`test_turbopack_loader_signals_a_change_before_turbopacks_baseline`). No test reproduces the race on real `next dev`. So nothing shows that the fix closes it there, or that the premise holds as stated, for example whether Turbopack's own watcher already catches an input the module depended on before.

## Resolution

Reproduce the race in the real-framework contract (`experimental/test-unplugin`), under `next dev` with both Turbopack and webpack:

- The fixture plugin rewrites a `RACE_`-prefixed contract value without the prefix right after reading it. The edit then lands between the compile's read and the host taking the module's dependencies, deterministically.
- Once for an input the module already depended on, and once for one it depends on for the first time (`FROM_LATE` reads `src/late-input.server.ts`).
- Assert the page eventually renders the rewritten value.

## Acceptance

- Both bundlers render the rewritten value after each race, on the Linux e2e lane.

Issue #1423, opened 2026-09-19
fix(unplugin): Turbopack takes a loader's dependencies at return, so an earlier change is never seen


samchon:
## Problem

Turbopack takes a loader's dependencies only when the loader returns. Its loader runtime sends `fileDependencies` from the `runLoaders` completion callback, and Turbopack reads their state at that moment as its baseline. The runtime is `transforms/webpack-loaders.ts`, embedded in Next 16.3.4's native binary.

A change can land after the transform read an input but before Turbopack takes that baseline. The change is then part of the baseline: nothing ever re-runs the module, and the page keeps the output of the older state.

#1400 closed this window for failed compiles: a failure whose project moved while it ran is compiled again. What remains is short but real:
- from the capture's last read to the loader's return, including registration;
- from the return to Turbopack processing the dependency message, which grows when the host is busy.

webpack and Rspack compare file timestamps with the compilation's start, so they are not affected.

## Resolution

In a development session, where the bridge exists:

1. Hand every input the loader registers to the bridge as well. The bridge proves at registration that nothing changed since the compile read each input (its `startedAt` check), and observes each one from then on, independently of Turbopack's baseline.
2. When the bridge finds the module stale, rewrite its sentinel after a short delay, and keep rewriting it with a growing delay until the module runs again. A rewrite that lands before Turbopack took the sentinel's own baseline is then followed by one that lands after it. The module's next run is the acknowledgement, and it stops the rewrites.

## Acceptance

- A change made after the compile read an input, and before the loader returned, re-runs the module.
- A change the bridge hears after registration rewrites the sentinel until the module is delivered again, and stops then.
- An edit that Turbopack's own watcher already acted on causes no rewrite once its delivery arrives.

PR #1435, opened 2026-09-19
fix(unplugin): prove host observation on real hosts and surface its costs

samchon:
## Intent

#1429 closed the host-observation gaps in logic, but two of them were proven only against stand-ins. This PR adds real-host evidence for them:

| Issue | Was proven on | Now proven on |
| --- | --- | --- |
| #1434 | hand-built inotify event bytes | a real kernel `IN_Q_OVERFLOW`: an unread instance filled past `fs.inotify.max_queued_events`, drained through the helper's own read and decode |
| #1433 | a fake Turbopack loader context | real `next dev`, Turbopack and webpack, where the fixture plugin rewrites the input between the compile's read and the host taking the module's dependencies. It does this once for an input the module already depended on, and once for one it depends on for the first time. |

## Known limit

A real FSEvents dropped-events notice (#1425) cannot be provoked on demand, so that path remains proven with a stand-in binding. The `fsevents` binding's pass-through of every flag was verified by reading its source.

- `go vet` passes for the helper and the fixture plugin (the helper with `GOOS=linux`).
- Both new tests run only where they can: the Go test on Linux, and the e2e on the `platform (linux-x64)` lane. This PR's CI is their first run.

https://claude.ai/code/session_01491ZRN4aGMhrJFdvHj4gHp


samchon (commented):
Individual Self-Review (solo) of `4472fdd1b` (#1434) and `aeeb20f98` (#1433).

`4472fdd1b`:
- The test reads the live `max_queued_events` and creates one more uniquely named file than it allows, so no event coalesces and the kernel must overflow.
- It then drains through `h.drain()`, the loop's own read path, over the real descriptor.
- It skips only when the limit exceeds 2^20, where filling would take unreasonably long, and says so. The GitHub runner's default is 16384.

`aeeb20f98`:
- The race is deterministic. The plugin's write happens inside the compile, after its read, so it always precedes the loader's return and Turbopack's processing of the result.
- Both variants are covered: the contract input, which the module already depended on, and `src/late-input.server.ts`, which it depends on for the first time.
- The rewrite drops the `RACE_` prefix, so the next compile is stable and cannot race itself.
- The webpack run of the same contract checks that the race resolves there too, where timestamps decide.
- Waiting uses the contract's 30 s `eventually`. The re-signal schedule's rewrites at 0.05, 0.25, 1.05, 4.25, and 17 s all land inside it.

Verdict: clean. Both tests first run on this PR's CI.

samchon (commented):
Individual Self-Review (solo) of `069acb7db` (#1438), `c41d1f65e` (#1439), and `ca8fad3bd` (#1440). These close the residuals the PR body listed as known limits.

- `069acb7db`:
  - The macOS gap path now has real-host evidence. The installed binding opens streams with `WatchRoot`, so moving the watched directory delivers a real `RootChanged`, which `DROPPED` includes. The registration must receive a gap.
  - It runs in the `bundler-defenses-macos` lane through `--include=watch_broker`, and returns at once elsewhere.
  - A dropped-events notice itself still cannot be provoked, but it takes the same branch in the same stream callback.
- `c41d1f65e`: the caching cost of an unnamed Turbopack root is no longer silent.
  - It emits one Node process warning per project (`TTSC_TURBOPACK_UNTRACKED_INPUTS`), only when a module left an input untracked and the configuration names no root. It names `turbopack.root` and the hand-wired `turbopackRoots`.
  - The loader test observes it through `process.on("warning")`, replacing nothing: none for a named root, and one for two runs on an unnamed project.
- `ca8fad3bd`: a missing `fsevents` binding is named once per process (`TTSC_FSEVENTS_MISSING`) when the broker starts on macOS without it. The same observation style is used in its features test.
- Both warnings use Node's standard warning channel, so a host that silences warnings (`NODE_NO_WARNINGS`) silences these too, by its own choice.

Verdict: clean. The body's "Known limit" section is superseded by these commits.

samchon (commented):
CI diagnosis on `431ee4e23` (run 35453556672).

`platform (linux-x64)`, the lane that runs the real-host contracts, passed. Every contract passed, including:

- webpack and Rspack: an edit landing after ttsc returned and before the host recorded the module's inputs, to a tracked input and to a first-time input, in a fresh watcher after the compile-count contracts;
- Next Turbopack: the same two races through a loader placed after ttsc in the rule chain, plus the plugin-driven races during the compile;
- the earlier heads' failures were defects of my scenarios, not of the adapter: the output check did not know the new values, and the race steps ran before contracts that assumed the previous state.

`go` failed on `TestLSPProxyDidSavePublishesVersionlessPluginDiagnostics` in `packages/ttsc/test/driver`: `upstream frame did not arrive in 10s`. The PR does not touch the LSP proxy, `internal/watchhelper` passed in the same job (including the real kernel overflow), and the test passed on the six preceding master runs. It is filed as #1441 with the analysis of the `didSave` path. The failed job is re-running; this PR's verdict uses that rerun, and #1441 is left to its own fix because its cause needs reproduction under load rather than a guess.

Still open before merge: the Turbopack `THIRD` recovery that stayed on the failed compile for over 30 s on head `ae0b651d1`. The repeat experiment (12 × next-turbopack, 12 × rspack, 6 × webpack) measures its frequency.

samchon (commented):
Overall Self-Review (solo) on `431ee4e23`. CI is green on this head (28/28, after the unrelated `go` flake re-ran; see #1441).

Surface: the whole diff from master, 14 files, read top to bottom.

Measured facts this PR rests on:
- With #1423's fix off (experiment branch, same contracts), real `next dev --turbopack` never reflected an edit landing after ttsc returned when the edited file became a dependency for the first time; the unit scenario failed the same way. With the fix on, both are reflected. The premise and the fix are proven on the real host, and #1423's body now states the measured scope.
- webpack and Rspack reflect the same races without any fix, as their timestamp comparison predicts.
- A real kernel `IN_Q_OVERFLOW` reaches the helper's `drain`; a real FSEvents `RootChanged` reaches the broker as a gap through the installed binding.
- 30 repeated runs of the three loader-host contracts passed. The two one-off failures seen earlier are recorded in #1442 with the diagnostics now built into the contract.

Code:
- The two warnings are Node process warnings with stable codes, emitted once per project and once per process, on the exact conditions that cost caching or notifications. Neither changes behaviour.
- `resolveTurbopackRoot`, `fseventsBindingPath`, and the broker spawn are unchanged in logic; the spawn only names the binding once.
- Tests observe warnings through `process.on("warning")`; no monkeypatching anywhere. The e2e race loader is a plain webpack-style loader using only the public loader context.
- Go: the overflow test skips only above 2^20 queued events, which no default system sets.

PR #1437, opened 2026-09-19
fix(ttsc): report a missing Windows Go wrapper as not found, and fix two environment-bound tests

### LINKED ISSUE #1432: A missing Windows .cmd Go toolchain reports EINVAL instead of the install hint on current Node
## Summary

When `TTSC_GO_BINARY` names a Windows `.cmd` or `.bat` wrapper that does not exist, ttsc should report "the Go toolchain was not found" and name `TTSC_GO_BINARY`. On Node 24 it reports instead:

```
ttsc: reading go.mod for plugin "go-source-plugin" failed to spawn missing-go.cmd: spawnSync missing-go.cmd EINVAL
```

## Cause

`spawnGoTool` preserves the native ENOENT contract for a missing wrapper by spawning the `.cmd` path directly without a shell. Since the CVE-2024-27980 fix (Node 18.20.2, 20.12.2, 21.7.3 and later), Node refuses to launch a `.cmd` or `.bat` without a shell and answers EINVAL before looking for the file. The callers map only ENOENT to the install hint.

## Reproduction

`test_plugin_corpus_missing_go_toolchain_points_users_at_the_install_hint` and the missing-wrapper cases of `test_spawngotool_preserves_windows_command_wrapper_arguments`, on Windows with Node 24.

## Expected

The wrapper resolution already knows the file does not exist, so the missing tool reports the ENOENT result Node gives for any missing executable, on every Node version.

### LINKED ISSUE #1436: Two tests fail on Windows machines with an 8.3 temp path or without the symlink privilege
## Summary

Two tests fail for reasons unrelated to the behavior they pin. Both show up on a Windows developer machine; the CI runners have neither condition.

1. `test_project_rule_lifecycle_surfaces_through_package_discovery_cli_api_and_watch` compares the URI the server publishes for the project's tsconfig with `pathToFileURL(...).href` as strings. When the temp directory is an 8.3 short path (`C:\Users\SAMCHO~1\...`), Node's `pathToFileURL` spells `~` as `%7E` while the server keeps the unreserved `~`. The two URIs are equivalent (RFC 3986 section 6.2.2.2), and the server names this file itself, because no client opened it. Other LSP tests compare a URI the server echoes back from `didOpen`, where string equality is correct.
2. `test_ttsx_virtual_layout_mirrors_a_file_symlink_project_entry` creates its file-symlink fixture unconditionally, so on Windows without `SeCreateSymbolicLinkPrivilege` the fixture itself throws EPERM. The peer symlink tests return early when the fixture cannot be created. The product's fallback for a refused mirror is pinned separately by `test_linkvirtualentry_copies_a_file_symlink_entry_when_symlink_creation_fails`.

## Expected

The first test compares the server-named URI as the file it names. The second skips when its fixture cannot exist, like its peers.

samchon:
## Intent

Fix the remaining defects found during the ttsx work: one product defect on current Node for Windows, and two tests that failed for reasons unrelated to the behavior they pin.

Closes #1432
Closes #1436

## Changes

- **`spawnGoTool` (#1432).** A missing Windows `.cmd` or `.bat` Go wrapper was spawned directly to preserve the native ENOENT contract. Since the CVE-2024-27980 fix, Node refuses that spawn with EINVAL before looking for the file, so the "Go toolchain was not found" install hint never appeared. The wrapper resolution already knows the file does not exist, so nothing is spawned: the result is the one Node returns for a missing executable. The ENOENT errno comes from `util.getSystemErrorMap()`.
- **Project-rule lifecycle test (#1436).** It compared the URI the server publishes for a config no client opened with `pathToFileURL(...).href` as strings, and Node spells the unreserved `~` of a Windows 8.3 path as `%7E`. It now compares the file each URI names.
- **File-symlink virtual-layout test (#1436).** It returns early when its symlink fixture cannot be created, as its peer symlink tests do. The product fallback for a refused mirror stays pinned by `test_linkvirtualentry_copies_a_file_symlink_entry_when_symlink_creation_fails`.

## Verification (local, win32-x64, Node 24.18.0)

- `test_plugin_corpus_missing_go_toolchain_points_users_at_the_install_hint`: failed with EINVAL before the change, passes after.
- `test_spawngotool_preserves_windows_command_wrapper_arguments`: pass.
- `test_project_rule_lifecycle_surfaces_through_package_discovery_cli_api_and_watch`: failed on `~` versus `%7E` before, passes after.
- `test_ttsx_virtual_layout_mirrors_a_file_symlink_project_entry`: threw EPERM before, passes after.
- `packages/ttsc` `tsc` build: clean.

https://claude.ai/code/session_01RUNXb4ecweEo5KBbpAuNk4


samchon (commented):
CI on 2cb5916c: 28/28 green, including every platform lane, both Windows lanes, and the ttsc core and native defenses.

samchon (commented):
Self-review round on 2cb5916c: fileURLToPath throws on a non-file URI, and the lifecycle test called it inside a notification predicate, where the throw would escape the listener. Fixed on bc859fc3 with a non-throwing namesFile helper that answers false for any other scheme, as the replaced string comparison did. The test passes locally.

samchon (commented):
Coverage round on 4d2ce3dd: added test_spawngotool_reports_every_missing_go_tool_as_native_enoent. It covers every missing spelling (bare, .cmd/.bat in either case, relative, absolute) and compares each result field by field with Node's own ENOENT. Mutation check: with the fix reverted in the built lib it fails with EINVAL; with the fix it passes.

samchon (commented):
CI on 4d2ce3dd: 28/28 green, nestia consumer run included. Merging with the repository's squash method.

Issue #1444, opened 2026-09-19
unplugin: a run that started before the change silenced Turbopack's re-signal


samchon:
## Problem

Reproduced in a 100-run repeat of the real `next dev --turbopack` contract (run 35457197160): after an edit landing after ttsc returned, to an input the module depended on for the first time, Turbopack rendered `SEVENTH SEVENTH SEVENTH LATE_RACE_SEVENTH`. Three modules delivered the new state; one kept the old.

The confirmed re-signal of #1423 stopped on the wrong condition. A loader start (`acknowledge`) ended the rewrites, on the reasoning that a run in progress would read the changed state. It need not: a run that started before the change, or one served from a generation captured before it, delivers the old state, and the registration of that delivery then carried a `startedAt` after the change had already been recorded, so the bridge saw nothing racing and never signalled again. The run record shared through the transform session widened the same mistake across the pool: any worker's run silenced every worker's signal.

## Resolution

Only a registration answers a signal. The registration carries what the delivery read and when its compile began, and the bridge's replacement re-checks every input changed since. A delivery that read the old state is signalled again at once; one that read the current state ends the rewrites. `acknowledge` and the shared run record are removed. A pool worker that never sees the registration keeps rewriting, at intervals growing fourfold, until the host runs and registers the module in that worker; such a run is a cache hit.

## Acceptance

- The bridge scenario: a stale re-registration is signalled again at once, and a current one stops the rewrites.
- The real-host contract passes its Turbopack races, including the first-time input, across repeated runs.

Issue #1443, opened 2026-09-19
unplugin: Windows re-spawns fsutil on every registration that removes an entry


samchon:
## Problem

On Windows, `createViteServeInputWatch` clears its per-directory case-sensitivity memo whenever an entry is removed (`pathIdentityMemosDirty`), and `createFilesystemPathIdentityContext` answers each directory's sensitivity by spawning `fsutil.exe file queryCaseSensitiveInfo` synchronously. Every registration that removes an entry therefore re-spawns `fsutil` for the directories it touches. Measured: a bridge `register` over one input costs about 245 ms on Windows, 210 ms of it in `spawnSync`, so the bridge's first confirmed rewrite lands at about 300 ms instead of 50 ms, and every Vite dev-server registration pays the same.

The memo is cleared because a rename or removal can change which directory a path resolves through, so a cached answer for a directory that no longer exists could be wrong. But the answer for a directory that still exists at the same spelling cannot change through a rename elsewhere; Windows sets the flag per directory.

## Resolution

Invalidate the memo per directory, not wholesale: drop the entries whose directory (or an ancestor) was renamed or removed, and keep the rest. Keep the `fsutil` probe for a directory the memo does not know.

## Acceptance

- A registration that removes an entry re-probes only directories that changed.
- The bridge's first confirmed rewrite lands near `FIRST_SIGNAL_DELAY_MS` on Windows.

Issue #1442, opened 2026-09-19
unplugin: one unreproduced Rspack bundle mixing two states


samchon:
## Problem

Two failures appeared once each on the `platform (linux-x64)` e2e lane during PR #1435, and did not reproduce in a 30-run repeat (12 × next-turbopack, 12 × rspack, 6 × webpack, run 35453565882):

1. **Rspack** (run 35447687865, head `1223fa98d` + e2e changes): after the `SECOND` edit to the contract input, the reported bundle carried `SECOND` for three modules and `FIRST` for the fourth. The contract now records which modules each build rebuilt and whether a later build corrects a mixed one (`431ee4e23`), so the next occurrence names the module Rspack did not rebuild.
2. **Next Turbopack** (run 35451178181, head `ae0b651d1`): after `project.break()` and the `THIRD` repair, the page kept answering the failed compile's error (`invalid contract type`, from `selectTransformedSource`) for over 30 s. The dev server log shows no compile after the repair.

Both are consistent with an edit arriving while the host has the previous state in flight, which is the class #1423 addresses for Turbopack. Neither is proven.

## Update

The Turbopack failure reproduced in the 100-run repeat (run 35457197160) and is a real defect, fixed as #1444: the re-signal stopped on a loader start instead of on a registration that read the change. The Rspack failure did not reproduce in 52 runs (12 + 40) and Rspack does not use the confirmed bridge, so it stays open here with the diagnostics of `431ee4e23` in place.

## Resolution

- Keep the diagnostics of `431ee4e23` in the contract.
- On the next occurrence of either, read the recorded rebuild sets (Rspack) or the loader run count and sentinel writes (Turbopack) and fix the cause. If the Turbopack one recurs, check whether the failed generation's recovery inputs (`failed: true` registration) reached the bridge before the repair, and whether the repaired compile was claimed by another worker's shared session.

## Acceptance

- 100 consecutive runs of the three contracts pass, or the cause is found and fixed with a deterministic reproduction.

samchon:
Closing on evidence, not on a fix.

- The one occurrence was on head `1223fa98d`, before #1444 and #1446 landed.
- Since then Rspack has passed 12 + 40 + 40 + 150 = 242 repeated runs of the same contract on Linux CI (runs 35453565882, 35466387621, 35470037598), with the contract's diagnostics in place (`431ee4e23`: the modules each build rebuilt, and whether a later build corrects a mixed one).
[...]

Issue #1441, opened 2026-09-19
ttsc: LSP proxy didSave forward can miss its 10 s window (flaky test)


samchon:
## Problem

`TestLSPProxyDidSavePublishesVersionlessPluginDiagnostics` (`packages/ttsc/test/driver`) failed once on the `go` lane of PR #1435 (run 35453556672) with `upstream frame did not arrive in 10s`, after the same test passed on the six preceding `master` runs. The PR touches nothing the LSP proxy reads, so the test is flaky.

The test sends an upstream `publishDiagnostics` (version 7), reads the editor frame it produces, then sends a versionless `didSave` and expects the proxy to forward it upstream within 10 s. The `didSave` path runs `invalidateSymbolProvider`, `invalidateResidentPlugins`, `refreshPluginCompletionHints`, `publishPluginDiagnosticsForDocumentNotification`, and `resumePendingProjectDiagnosticRefresh` synchronously before `writeUpstreamFrame`. With the test's `stubSource`, the first three return at once, so the wait is in the last two or in the pipes.

## Resolution

Reproduce under load (`go test -count=200 -run TestLSPProxyDidSave ./test/driver` with `-race` and CPU limits), find the step that can hold the editor pump for 10 s, and remove the dependency of the upstream forward on it. The forward of an editor notification must not wait on plugin work.

## Acceptance

- The test passes 200 consecutive runs under `-race` with `GOMAXPROCS=1`.
- The editor-to-upstream forward of `didSave` does not block on diagnostics publication.

PR #1445, opened 2026-09-19
fix(unplugin): end a Turbopack re-signal only on a registration that read the change

samchon:
## Intent

The 100-run repeat of the real `next dev --turbopack` contract (run 35457197160, on master after #1435) reproduced a defect in #1423's confirmed re-signal: one module of four kept the old value after an edit landing after ttsc returned. This PR fixes it (#1444).

## Cause

The re-signal stopped when a loader started (`acknowledge`), on the reasoning that a run in progress would read the changed state. A run that began before the change, or was served from a generation captured before it, delivers the old state. Its registration then carried a `startedAt` after the change had already been recorded, so the bridge saw nothing racing and never signalled again. The run record shared through the transform session let any worker's run silence every worker's signal.

## Change

- Only a registration answers a signal. It carries what the delivery read and when its compile began; the bridge's replacement re-checks every input changed since. A delivery that read the old state is signalled again at once, one that read the current state ends the rewrites.
- `HostWatchBridge.acknowledge` and the shared run record are removed. The confirm option is a boolean again.
- A pool worker that never sees the registration keeps rewriting at fourfold intervals until the host runs and registers the module in that worker, from its cache.

- The bridge scenario covers a stale re-registration (signalled again at once) and a current one (rewrites stop); the Turbopack loader scenario still stops after a run that delivered the change.
- The features tree and the 75 CI script tests pass.
- The real-host proof is CI's: the `platform (linux-x64)` contracts, whose Turbopack first-time-input race is the reproduction.

## Also found

- #1443: on Windows, every registration that removes an entry re-spawns `fsutil`, which delays the first confirmed rewrite to about 300 ms. Efficiency only; left to its own fix.


samchon (commented):
Individual Self-Review (solo) of `9506aeecc` (#1444).

- The stop condition is now the one that carries evidence. `register` settles the owed rewrite, then `replace` re-checks every input changed since the delivery's `startedAt`. A stale delivery reaches `check → reloadImporters → invalidateModule → signal` synchronously and starts a new schedule; a current one leaves no signal.
- Removed: `acknowledge`, the shared run record, the session lookup in the loader. Nothing else read them.
- Cross-worker: a worker that never sees the registration keeps rewriting at fourfold intervals. The host's extra run is a cache hit, and that run's registration stops it there. Correctness no longer depends on any inter-worker state.
- Turbopack still opens the bridge with `confirmDelivery = true`; every other host is unchanged (`false`).
- Tests: the bridge scenario asserts both branches (stale re-registration signalled again at once; current one stops), and the native loader scenario still stops after a run that delivered the change. Two fakes dropped `acknowledge`.
- Measured on Windows: a re-registration's first rewrite lands at ~300 ms instead of 50 ms because `register` re-spawns `fsutil` (#1443). Efficiency only.

Verdict: clean.

samchon (commented):
Individual Self-Review (solo) of `ba7844d39` (#1446) and the two diagnostic commits before it.

- Every path that surfaces a rejected generation now registers first: the replayed terminal verdict, and both `awaitOrEvict` calls. The registration runs before the throw, synchronously, so a loader's `callback(error)` follows every `addDependency`, which Turbopack's runtime sends before it reports the error.
- `TtscUnstableGenerationError.validation.cached` is the last attempt's generation, so `notifyFailedGenerationInputs` registers what a failed envelope of it would: the walk's files, the out-of-walk inputs, and the membership input. Any other rejection registers `[]` with `failed: true`, which `withSelectionInputs` fills with the consulted configuration files, and the bridge takes as a failed registration that keeps the importer's earlier inputs observed.
- The hook is only called when present, so one-shot hosts and callers without hooks are unchanged.
- Test: the readdir-failure reproduction of the existing unstable-generation scenario, with a hook, asserting the failed registration on the first rejection and on the replayed verdict, and no failed flag on the recovery. It fails without the change at the first assertion.
- The diagnostics that found this stay in the contract (compile counts and sentinel states around the break and repair).
- Not proven by this PR: whether the same class explains the one Rspack mixed bundle (#1442). Rspack keeps a module's previous dependencies across a failed run, so it is not this defect; the repeat after this fix includes 40 Rspack runs.

Verdict: clean.

samchon (commented):
Overall Self-Review (solo) on `ba7844d39`. CI is green on this head (the planner's 10 lanes for this change set, `platform (linux-x64)` with the real-host contracts included).

Measured on real `next dev --turbopack`, same contracts, before and after:
- before #1444: 1 of 100 runs kept one module on the old value after a post-return edit;
- before #1446: 2 of 60 and 4 of 80 runs kept a page's failure after its repair;
- after both: 80 of 80 Turbopack runs and 40 of 40 Rspack runs pass (run 35466387621).

Surface: the whole diff from master, 13 files.

- #1444: the confirmed re-signal ends only on a registration, which the bridge re-checks against the changes since the delivery's compile began. `acknowledge` and the shared run record are gone; nothing else read them. Turbopack keeps `confirmDelivery = true`, every other host `false`.
- #1446: every rejection path in `transformTtsc` registers recovery inputs before it throws: the replayed terminal verdict and both `awaitOrEvict` awaits. The `catch` re-throws, so the retry `continue` paths are unchanged. An unstable generation registers its last attempt's inputs; any other rejection registers the selection inputs through `withSelectionInputs`. The hook is optional, so one-shot and hookless callers are unchanged.
- The bridge takes `failed: true` with an empty list as a failed registration that keeps the importer's earlier inputs observed, and Turbopack takes the sentinel it returns as a dependency, so a later change reaches the module through either channel.
- Tests: the bridge scenario asserts both stop conditions; the rejected-generation scenario asserts registration on the first rejection and on the replay, and fails without the change; the e2e contract keeps the diagnostics that found #1446.
- Docs describe the registration-only stop condition.

Open, outside this PR: #1442 (one unreproduced Rspack mixed bundle, now 0 of 92 further Rspack runs), #1443 (Windows `fsutil` re-spawn, efficiency), #1441 (LSP proxy flake, outside unplugin).

PR #1447, opened 2026-09-19
fix(unplugin): keep the case-policy memo across a registration that replaces its inputs

### LINKED ISSUE #1443: unplugin: Windows re-spawns fsutil on every registration that removes an entry

samchon:
## Intent

Fix #1443: on Windows, every registration that removed an entry reset the per-directory case-sensitivity memo, so the next registration re-spawned `fsutil` per directory. Measured: a bridge re-registration cost about 240 ms, 1 ms after this change.

## Change

Only a rename can make a remembered path fact wrong, so a removal marks the memos dirty only once a rename has been heard since the last reset. A registration replacing its inputs keeps them; the removal a rename causes still resets, and a replacement server still rediscovers the policy.

- The Vite subscription scenario gains both cases: a replaced registration does not re-probe, a removal after a rename does. It passes with the features tree.
- Re-registration measured at 1 ms after the first (which still probes once).


samchon (commented):
Individual and Overall Self-Review (solo) on the one commit of this PR. CI is green on this head.

- The change is one flag: `topologyChanged` set by a rename in `recordChange`, consumed by `remove`, cleared by `resetPathIdentityMemos`. Every reset site is unchanged, so an unattributed event, `dispose`, `forget`, and `replace` still reset when a removal followed a rename.
- Correctness argument: a path-identity memo can only be wrong after a rename, and a rename always sets the flag before any removal it causes, since `check` runs after `recordChange`. A removal with no rename since the last reset removes an entry whose directories are unchanged, so its memo stays right.
- The memo's size is still bounded by a rename-caused removal, a dispose, or a forget; a session with no renames keeps one answer per directory it touched, which is what a long session needs anyway.
- Test: the existing case-policy scenario gains both branches; the Windows measurement (240 ms to 1 ms per re-registration) is in the PR body.

PR #1448, opened 2026-09-19
test(unplugin): observe convergence in the watch-host contracts, not one build

samchon:
## Intent

A test must decide in one run. The watch-host contracts (webpack, Rspack, Rollup, Rolldown, esbuild) took the next watch event as the edit's build and asserted its values, which made them depend on how many builds the host emitted for one edit. That is the host's choice, not the adapter's, and it is what the one unreproduced Rspack failure (#1442) most likely was: a partial build judged as the final one.

## Change

Every value assertion in those contracts now waits for the build in which every consumer carries the value, skipping failed and partial builds on the way. A host that never converges runs out the deadline, and the error lists every build seen while waiting, failed ones included, so a host that converged on a mix is told apart from one that never rebuilt. Compile-count assertions are unchanged, so cache reuse is still proven. The Next, Vite, Farm, and Bun contracts already observed converged state and are untouched.

The earlier Rspack diagnostic block, which recorded rebuilt module sets after a mixed build, is removed: the convergence wait makes a mixed build a non-event, and the deadline error carries the trace.

## Verification

- Syntax-checked locally; the contracts run only on the `platform (linux-x64)` lane, which this PR's CI runs.


samchon (commented):
Individual and Overall Self-Review (solo) on the one commit of this PR. CI is green on this head, and every host contract passed on the Linux e2e lane.

- `settledOutput` decides by state, not by event count: it returns the first build in which every consumer carries the value, treats a failed build as not yet converged, and re-throws the queue's own deadline so a stalled host cannot loop. The deadline error lists every build seen, so a host that converged on a mix and one that never rebuilt are told apart.
- Compile-count assertions stay beside each convergence wait, so a host that converged by recompiling more than the contract allows still fails.
- `changedOutput` and the Rspack diagnostic block are removed; nothing else referenced them. The `unchanged rebuild` step after `watcher.invalidate()` also waits for convergence and still asserts no new compile.
- Kept as single-event waits: the initial failure before any subscription, and the race watcher's first build, where the queue is provably empty.
- The Next, Vite, Farm, and Bun contracts already observed converged state through HTTP polling, module-graph state, or direct calls.
- No hardcoding or monkeypatching: one contract helper generalized, the fixture's own quoting relied on for the value regex.

PR #1449, opened 2026-09-20
ci: run the whole adapter suite and its real hosts on every representative OS

samchon:
## Intent

Verify the adapter on every OS it supports, not on Linux plus a list of OS-specific cases.

## What was measured before this change

| Check | Linux x64 | Windows x64 | macOS |
| --- | --- | --- | --- |
| adapter `features` tree | all | none | none |
| adapter `native-plugins` tree | all | 23 selected cases | 5 selected cases |
| real hosts (Next, Vite, webpack, Rspack, Rollup, Rolldown, esbuild, Farm, Bun) | all | none | none |

The adapter watches each OS through different code (the Linux inotify helper, the Windows and macOS brokers with `fs.watch` and `fsevents`), so a case absent from a list ran nowhere on that OS.

## Change

- The Windows and macOS bundler lanes run both adapter trees with no selection. Bun is set up on both; macOS keeps its descriptor limits.
- The packed E2E, which drives the real hosts, runs on every representative OS instead of `linux-x64` only.
- The planner's tests state both contracts and no longer enumerate cases.

## Expectation

This is the first run of most of the suite on Windows and macOS, and of the real hosts there. Failures are findings: each is either an OS defect in the adapter, to fix here, or an OS assumption in a test, to fix here too.


samchon:
Superseded by the PR that carries this CI change together with the two defects its first run found (#1450, #1451).

Issue #1460, opened 2026-09-20
unplugin: a sentinel rewritten while Rollup is building is lost, and the importer stays on its old output


samchon:
## Symptom

In `rollup --watch`, editing a compiler input that a module depends on only through the plugin (a type-only declaration, a config) while a build is in progress leaves that module on its previous output. The next builds keep serving it from Rollup's cache, and nothing re-runs it until its input changes again.

Measured on the Windows package contract (`platform (win32-x64)`), where the failure showed as `rollup second rebuild` timing out with `[["THIRD","THIRD","THIRD","SECOND"], ["THIRD","THIRD","THIRD","SECOND"]]`: three consumers rebuilt on the new value while the entry stayed on the old one, through two builds.

## Cause

A watching Rollup build hears the bridge's sentinel through a per-path chokidar watcher and applies the invalidation to `this.cache.modules` (`Task.invalidate`: the module's `originalCode` is set to `null`). While a build is running, `this.cache` is the previous build's cache; the running build already took the module from it, and when it finishes, its result replaces `this.cache` with the module's `originalCode` intact. The invalidation is gone, and the rerun Rollup schedules serves the module from that cache.

The window is the whole build after the module was read from the cache. It opened routinely on Windows because Rollup's watcher delivered the previous build's sentinel events only after that build ended, which started a leftover build, during which the next edit's sentinels were rewritten.

## Fix

The bridge remembers which importers it signalled and which registered since (`owes`). Rollup asks `shouldTransformCachedModule` before it serves a module from its cache, and the adapter answers `true` for an importer the bridge still owes a run, so the module is transformed instead, registers, and settles the signal. No timer and no second build: the signal is answered on Rollup's own channel, in the build that would otherwise have lost it.

samchon:
A second half, found by the contract's new mid-build edit: a registration must not answer the signal with a stale delivery. A build pass proves the generation once and serves every later module of the pass from it, so a module first delivered after the edit, in the same pass, carries the state before it. Its registration was checked against changes since its own transform began, which was after the edit, so nothing was raced and the registration settled the signal; the module then came from Rollup's cache. The registration is now checked against changes since the pass opened.

samchon:
Fixed in #1452 (merged as 52208a288).

PR #1464, opened 2026-09-20
test(unplugin): run one scenario matrix on every host and root, and speak esbuild's and Farm's channels

### LINKED ISSUE #1461: unplugin: a new root file under a linked project root is never heard where the backend reports physical paths
## Symptom

On macOS, a watching Rollup session on a project named through a link (every macOS temporary directory is one) never recompiles when a root file the tsconfig includes appears; on Linux and Windows it does. Measured on the host matrix (run 35519818920): `rollup (linked root): new root file` timed out with the compile count unchanged.

## Cause

The serve watcher's membership check compared the event path with the membership entry's root by string (`pathIsWithin(absolute, entry.file)`). FSEvents reports the physical path of what changed, so an event under `/private/var/.../project/src/x.d.ts` was never below the root registered as `/var/.../project`; inotify and ReadDirectoryChangesW report names under the watched path as it was given, which is why the other platforms heard it.

## Fix

A membership entry keeps the root's physical spelling beside the one it was registered under, and an event is placed under the root's own name whichever spelling names it, before its membership policy is asked (`relativeToProject`), in both the event check and the directory-level admission.

### LINKED ISSUE #1462: unplugin: Farm is handed a linked project's inputs under the physical spelling, which its watcher cannot relate to its root
## Symptom

Under a Farm root named through a junction (Windows), an edit to a compiler-only input no longer rebuilds the modules that read it, and `package.json` and `tsconfig.json` are registered twice, under both spellings. Measured on the host matrix (run 35519818920): `farm (linked root): edit between builds`, with the input registered as `farm-linked\src\contract-input.server.ts` while Farm's root is `farm-linked-link`.

## Cause

Inputs are handed to a host under the spelling of the module the host delivered (`hostSpelling`, samchon/ttsc#1451). Farm's resolver delivers the module physically, but Farm relates every watch file to its configured root, so a physical path is not the module's dependency as far as its watcher is concerned.

## Fix

A host whose channel relates inputs to a root of its own, rather than to the module, says so: the Farm adapter hands the transform its resolved root (`hooks.spelling`), and every input of the delivery, the failed and rejected registrations included, is spelled under it.

Farm cannot resolve the entry from a root named through a symbolic link at all ("Can not resolve `./src/main.ts`", measured on Linux and macOS), so the matrix runs Farm's linked root on Windows only.

### LINKED ISSUE #1463: unplugin: esbuild loses an edit landing between one module's loader result and another's
## Symptom

Under `esbuild` watch, editing a compiler-only input after ttsc returned one module and before the build ended leaves that module on its old output for good. Measured on the host matrix on Linux, macOS and Windows (run 35519818920): `esbuild: edit after ttsc returned`, one build with the old value and no rebuild in 60 s.

## Cause

esbuild keeps one watch state per path for a whole build, taken from the last loader result that named the path. Measured on plain esbuild with no ttsc: a module whose loader result named the input after the edit sets the baseline to the edited state, so the earlier module's edit is masked and nothing rebuilds. The adapter handed every module the same inputs through `watchFiles`, so the modules served from the cache right after the edit masked it for the module compiled before it.

## Fix

Every compiler input goes to the adapter's bridge, which observes it and rewrites one sentinel per module; esbuild watches only the module and its sentinel, a path no other module's result names, so no later result can mask it. esbuild tells a plugin nothing about whether its context watches, so the bridge opens for a one-shot `build()` too and closes when the last context is disposed, which `build()` reports at its end.

### LINKED ISSUE #1465: unplugin: Rolldown drops a change to a watched input that lands while it is building
## Symptom

Under `rolldown` watch, an edit to a compiler-only input that lands after ttsc returned a module and before the build ended sometimes leaves the module on its old output, with no rebuild. Measured on the host matrix, intermittently, on Linux, macOS and Windows (run 35531207637: `rolldown: edit after ttsc returned` and `rolldown: new input after ttsc returned`, one build carrying the old value and none after in 60 s), while two earlier runs of the same head passed.

## Cause

Rolldown watched the input through its own `addWatchFile` channel, and a change to a watched file landing while Rolldown is building is not always turned into a rebuild. The window is the build; whether the edit is lost depends on where in the build it lands.

## Fix

Every input goes to the bridge, as it does for Rollup, and the bridge repeats the signal until the module registers again with a delivery that read the current state, as it does for Turbopack: a rewrite that lands during a build is lost, the next lands after it. Rolldown watches only each module and its sentinel.

samchon:
## Intent

Make the host contracts the safety net first, then fix what they find by class. Every host contract was hand-written with its own steps, so an edit landing while a host was building was exercised on none of them until Rollup lost one (#1460). The contracts now share one list of scenarios, run on a project named directly and on one named through a link, on Linux, macOS, and Windows; each host declares only how its output and failures are read and which seams it offers after ttsc.

## The matrix

15 scenarios per session: initial failure and repair; edit between builds (one compile shared); edit again; break and recover; unchanged rebuild; edit during the compile; new input during the compile; edit after ttsc returned, before the host's build ended; new input after ttsc returned; two edits in one tick; edit during the host's own build; saved the way an editor saves (write beside, rename over); deleted and recreated; new root file; final edit. Hosts: vite7, vite8, react-router, rollup, rolldown, esbuild, webpack, rspack, farm, next-webpack, next-turbopack, bun. Roots: direct and linked (junction on Windows, symbolic link elsewhere). esbuild and Bun offer no seam after a module's loader, so theirs is the resolution of the module's imports.

## What it found, and the fixes by class

The complete list came from the matrix on all three OS before anything was fixed (run 35519818920), plus one finding that reproduced only on the fourth run (35531207637).

| Issue | Host | Root | OS | Class | Defect |
| --- | --- | --- | --- | --- | --- |
| #1463 | esbuild | direct | all | host signal | esbuild keeps per watch file the state of the last loader result naming it, so a module loaded after an edit masks the edit for a module loaded before it; no rebuild ever. Measured on plain esbuild. Every input now goes to the bridge; esbuild watches each module and its own sentinel. The bridge opens for a one-shot build too, since esbuild cannot say whether it watches, and closes on dispose. |
| #1465 | rolldown | both | all, intermittent | host signal | A change to a watched file that lands while Rolldown is building is not always turned into a rebuild. Every input now goes to the bridge, and the bridge repeats the signal until the module registers again, as it does for Turbopack. |
| #1462 | farm | linked | Windows | path spelling | Farm was handed the input under the physical spelling while it relates watch files to its configured root. A rooted channel now names its root (`hooks.spelling`), and every registration of a delivery is spelled under it. |
| #1461 | rollup | linked | macOS | path spelling | The serve watcher compared FSEvents' physical event path with the membership root's link spelling by string, so a new root file under a linked root never reached its importers. The entry keeps both spellings; the event is placed under the root's name before its policy is asked. |
| — | farm | linked | Linux, macOS | host limit | Farm cannot resolve the entry from a root named through a symbolic link ("Can not resolve `./src/main.ts`"); the matrix runs Farm's linked root on Windows only. |
| — | bun | direct | all | harness | The resolution seam read only the main input for the late-race marker. |

One observation at the pre-fix head did not reproduce in the three later runs and is not claimed fixed: `webpack (linked root): deleted and recreated` and `rspack (linked root): deleted and recreated` on macOS x64 (no build after recreation within 60 s, run 35519818920).

## Verification

- Test-workflow runs at 34a4405da: two green (35525260738, 35527517685); the third (35531207637, the PR's own) failed only on Rolldown's late edits on all three OS, which f153766c4 addresses; its run is the verification of this head.
- Locally (Windows): rollup, rolldown, webpack, rspack, vite7, vite8, esbuild, farm matrices on both roots; the new unit scenarios. The Next and Bun contracts, and the cold native envelope tests, were verified on CI only: this machine was too loaded to run them within their budgets.
- The Farm session models Farm as measured: a failed `compile()` leaves its compiler compiling for good and `farm start` fails outright, so the initial repair is a restart; the dev server's watcher reports every resolved watch path; a failed update surfaces as the unhandled rejection Farm's HMR engine re-throws.

Closes #1461, closes #1462, closes #1463, closes #1465.


samchon (commented):
Overall Self-Review (solo), head 34a4405da: rounds over the harness (every changed contract file, two rounds, three findings applied in the first: a leftover re-export, an unguarded read in the esbuild seam during the deletion scenario, a Farm session removing every unhandledRejection listener), and over the source diff (8 files) plus the harness delta since those rounds: no findings. Two full workflow runs green at this head (35525260738, 35527517685).

samchon (commented):
Round over the Rolldown delta (f153766c4: bridge kinds and confirmation for Rolldown in unplugin.ts, the channel note in registerBuildWatchInputs, the predicate test's Rolldown section, the page): no findings. Every check passes on f153766c4: the three OS package contracts, the three bundler-defense suites, typecheck, build.

Issue #1459, opened 2026-09-20
unplugin: a project below a link, every macOS temporary directory among them, gets no notification coverage at all


samchon:
## Problem

A tracker refuses notification coverage for every input whose lexical path traverses a symbolic link or junction anywhere above it, up to the volume root, and the project tracker withdraws its whole coverage when the project root itself lies below one. Every macOS temporary directory lies below such a link (`/var/…` to `/private/var/…`), and a linked workspace is one anywhere, so on such a project no input is ever proven by the watchers' silence: every delivery re-reads its candidates and inputs (`a candidate the generation watches must cost no filesystem call per delivery: 27 !== 0`, `persistent validation statted 16.0 paths per module (bound: 12)` on the macOS lane, run 35495486710; the diagnostic run 35496965177 shows the candidate tracker's coverage empty).

The exclusion guards against a link retargeted between the input and the directory watched for it, which moves the input without an event on the old target. A link at or above the watched directory is a different case: every delivery re-checks the watched directory's identity before it reads the tracker (`verifyLocations`), so retargeting such a link withdraws the tracker rather than moving an input silently.

## Resolution

The walk stops before the watched directory: only components between the input and the tracker's root are examined, and the project tracker's own root-level check is dropped, since the root's identity check already covers it. An input outside the root, and a tracker with no root, keep the whole walk.

## Acceptance

- A project reached through a link (a junction on Windows) keeps its inputs and candidates covered: a delivery after the first reads none of them.
- A link inside the project still leaves the inputs below it on metadata validation.

samchon:
Fixed in #1452 (merged as 52208a288).

PR #1466, opened 2026-09-21
test(unplugin): edit the tsconfig and rename a dependency in the host matrix, cover external inputs below a link, and measure the Windows drain

samchon:
## Summary

Two more edits every host must converge on in the host matrix, one coverage gap closed, one Windows ordering claim measured, and one cost stated.

- **Matrix: edit to the tsconfig.** The fixture plugin honours a `fixed` entry in its own tsconfig config, so adding and removing it has an observable effect on every consumer. Every host, both roots, three OS, converge on the value the tsconfig sets and back.
- **Matrix: dependency renamed away and back.** The late input is renamed to `.moved`, which a watcher hears as a rename, not a write; every host reports the failure and finds the dependency again once it is renamed back.
- **Fix: an external input below a link is covered by its own watched directory.** The coverage walk stopped only at the project root, so an input outside the project below any link (every macOS temporary directory among them) was examined up to the volume root and never covered, and each delivery proved it by reading. The directory watched for an input is re-checked by identity on every delivery, so the walk now stops there: the root for an internal input, the input's own nearest existing directory for an external one. Only a link between the input and its watched directory can move it silently, and those stay on metadata validation (follow-up to #1459).
- **Measured: the Windows drain's order.** The broker answers a drain after two turns on the ground that the kernel queues a directory change's completion before the write returns and hands completions out in order. A new scenario writes and drains two hundred times in a row on the real backend and asserts the write is delivered before every drain's answer. The doc states the ground instead of the mechanism alone.
- **Docs: what a one-shot esbuild `build()` pays for the bridge.** One process start per ancestor of the project plus the broker on Windows (the `fsutil` case-sensitivity probes), only the watch helper on Linux and macOS, once per build.

## Verification

- Test workflow run [35561237692](https://github.com/samchon/ttsc/actions/runs/35561237692) at `0593c513a`: every lane green on the first run, both new scenarios on all 12 hosts, both roots, three OS, plus the Windows drain measurement and the link-coverage unit scenario (junction roots on Windows, symbolic links elsewhere).
- Overall Self-Review: a round over the source delta and a round over the harness, test, and docs delta; no findings.

## Not claimed

- webpack/rspack, linked root, macOS, "deleted and recreated" (seen once at 5bc219776 in #1464) did not reproduce in this run either; a further dispatch samples it after merge.
- Rolldown on macOS, an existing input edited after ttsc returned: covered by the acknowledged-signal fix of #1465, not measured separately on macOS.


samchon:
Overall Self-Review (solo): one round over the source delta (`createHostInputMutationTracker.ts`, `watchBrokerSource.ts`), one over the harness, test, and docs delta (fixture plugin's `fixed` entry, both new scenarios, the Windows drain measurement, the link-coverage scenario, the esbuild cost paragraph). No findings in either round. The run at `0593c513a` (35561237692) is green on every lane.

PR #1467, opened 2026-09-21
fix(unplugin): hold the host matrix and the restart contract on every host, and keep persistent caches valid across restarts

samchon:
## Summary

Tests first, then three fixes by class, and a maintainer map of the invalidation model.

### Tests

- Host matrix: five more edits every host must converge on (a declaration no bundler loads, a declaration outside the root, the extended config, a dependency's directory renamed away and back, a missing import that appears), and every host run with two transform plugins: the standalone source plugin, whose envelope has no compiler graph, and a linked contributor (`assets/linked/contract`), whose compile runs the program.
- Restart contract (`restarts.mjs`, `restart-cycle.mjs`): webpack, Rspack, Farm, and Next with both compilers, each session in a process of its own over the host's persistent cache; a restart over an unchanged project must compile nothing, and every offline edit (input, tsconfig, root file added and removed, new dependency, dependency renamed away and back, external input) must change what the next session serves.

### Fixes

1. **`handWatchInputs` keeps the registration that says most about a path.** A graph lists the project root as a resolution input, so a failed generation's recovery list named the root plainly before its membership, and the membership was dropped: nothing observed a root file appearing, and a recreated input under a watching Next session was never heard again.
2. **Persistent caches stay valid across restarts.** The bridge's sentinels live in one directory for every process and survive the session (a per-process directory made every restored module stale on every restart), and the root-file membership reaches the host's file channel as a record file written only when its digest moves, refreshed at every build start.
3. **A pooled session publishes only a reusable compile.** A success whose graph proof failed on its worker was published under an unchanged key and adopted back by the retry and every other worker, failing a fresh project's first compile everywhere.

Each fix has a unit scenario that fails without it; the matrix and the restart contract are the proof on real hosts.

### Map

`website/src/content/docs/development/reference/unplugin-invalidation.mdx`: objects, one delivery's decisions, capture and stability, validation paths, registration, session publication, persistent caches.

Supersedes #1468.

https://claude.ai/code/session_01K5v8A5gSmQvAB7BdkZtgp6


samchon (commented):
Overall Self-Review round on `35abfd8a6` (CI green, 28 required checks). The round mapped the record model's writers, readers and registrations, every host's tool root, every per-process file and its lifetime, the retry bound, and every public claim the docs make about them. It found the following, which land in a follow-up pull request rather than holding this merge:

- `notifyProjectRecord` handed nothing over when a record write failed, reasoning that the host would run the module again at its next start. A persistent cache does not: a module handed over without the record is restored on its own bytes. The record is now handed over while it exists, the generation's inputs are read once, and the write is retried per delivery. A new native test pins it.
- `probeDirectory` added one exit listener per probed project root, past Node's leak warning in a process watching many projects; one listener now removes them all, as one does for session stores.
- The setup guide still said build hosts are handed inputs under a module's spelling (webpack, Turbopack) and Farm under its root's; that failure registrations claim no input state; `hostToolDirectory` said Farm needs the record inside its root, which the host matrix contradicts; the README listed five of seven record exports; the setup guide now also says the tool directory must be writable.
- The unplugin and source-map harnesses left a packed tarball in each package directory; they pack with `--out` as `pnpm package:tgz` does.

PR #1468, opened 2026-09-21
test(unplugin): stop each host with a persistent cache, edit under it, and start it again

samchon:
## Summary

Draft: a host with a persistent cache is stopped, the project is edited while nothing runs, and the host is started again over the cache its last session stored. Only the compiler's verdict on the current state may be served.

Steps, each across a restart: an input edit, a tsconfig edit and its restoration, a broken root file added and removed, a new dependency, a dependency renamed away and back, an external input broken and repaired. Each session proves its cache was stored before the next starts, so a step never passes by rebuilding from nothing.

Hosts: webpack (`cache.type: filesystem`), Rspack (`cache.type: persistent`), Farm (`persistentCache`), Next with webpack and with Turbopack (both caches on by default in Next 16.3).

Stacked on #1467, whose fixture it uses. Tests first: the run collects the failure list; fixes follow by class.

## Expected from the code

Membership is registered only with a bridge (`registerBuildWatchInputs`), so a root file that appears or vanishes while the host is stopped is predicted to leave a stale cached module in webpack, Rspack and Next.


samchon:
Rebased onto #1467's head (linked plugin mode, type-stripping stage). The restart contract now runs with the linked plugin, so its verdict steps (a broken root file, a broken external input) are observable; a source plugin's envelope carries no compiler verdict. Run 35573221013 collects the failure list; the earlier run on the source-plugin head was cancelled.

samchon:
Folded into #1467: its branch now carries the restart contract and the persistent-cache fix on top of the matrix extension, so one PR is reviewed and merged.

